PR #12, opened 2017-04-09
Plasma class (WIP)

SolarDrew:
First attempt at a basic Plasma class with physical coordinates and a few core and derived variables. Useful tests are difficult to write at the moment since it doesn't really do anything yet, but I've included some anyway.

Things I think still need to be included:
- [ ] Better docstrings, covering the derived variables
- [ ] Getters and setters for core variables
- [ ] (Dummy?) methods for reading/writing data
- [ ] Easier access to vector components, e.g. Plasma.magnetic_field.x
- [ ] Unit verification on setters

Also, I realise we didn't conclusively decide the units issue but I've used astropy for now because that's what I'm used to.


namurphy:
Thank you!  I had forgotten to share my prototype code which for simplicity only calculates the Alfven speed.  A difference (I think) is that this code stores the quantities as private variables (e.g., `self.__B`) and uses setters (`self.set_B(42)`) and getters (`self.B`).  This strategy (which I only looked up recently!) would allow `self.set_B` to make sure that the magnetic field is in the right units, for example, or that the dimensions are correct with other arrays.  I don't know if this is the best strategy but I figured I'd try it out and we can discuss it.  I also added `ensure_unit` as a way of making sure that a quantity has the right units (though this function is still a bit tempermental...).  

`from astropy import units as u
import numpy as np
from astropy.constants import mu0, m_p

def ensure_unit(quantity, unit):
    """Makes sure that a quantity is attached to a particular unit
    """

    if quantity is None:
        return None
    else:
        quantity = quantity * u.dimensionless_unscaled
        quantity = quantity.si
        if quantity.si.unit == u.dimensionless_unscaled:
            return quantity * unit
        elif quantity.si.unit == 'T':
            return quantity
        else:
            # Should return an error with a useful error message here
            print("Incompatible dimensions")
            return None

    if not (quantity is None):
        assert quantity.unit == unit, 'Error in units'
        

class PlasmaSI:
    """ Prototype class to represent a two component plasma with ions
    and electrons.
    """
    def __init__(self, 
                 B = None, # Magnetic field in Tesla
                 n = None, # Number density in particles per cubic meter
                 species = None):
       
        self.__B = ensure_unit(B, u.T)
        self.__n = ensure_unit(n, u.m**(-3))
        self.__species = species

        if self.__species == 'H':
            self.__m_i = m_p

    @property
    def B(self):
        return self.__B
        
    @property
    def n(self):
        return self.__n

    @property
    def species(self):
        return self.__species
        
    def set_B(self, B):
        self.__B = ensure_unit(B, u.T)

    def set_n(self, n):
        self.__n = ensure_unit(n, u.m**(-3))

    def set_species(self, species):
        self.__species = species

    def Alfven_speed(self):
        """Calculate Alfven speed
        """
        if self.__B is None or self.__n is None:
            return None
        else:
            return (np.abs(self.__B)/np.sqrt(mu0*self.__m_i*self.__n)).si
`

SolarDrew:
Yes, I agree we should use getters and setters for the core variables. I actually have in the code that I've adapted this from, and I was going to put them in here as well but I wanted to have tests in place first.

As for verifying the units, you may be pleased to hear (depending on how long you spent writing ensure_unit) that there's an astropy decorator for that:
```python
import astropy.units as u

@u.quantity_input
def speed(distance: u.m, time: u.s):
    return distance / time
```
Again, I wanted to include this but hadn't got as far as testing it.

plasmapy/plasma.py
```diff
@@ -0,0 +1,79 @@
...
+        self.y = domain_y
+        self.z = domain_z
+
+        x, y, z = self.x.si.value, self.y.si.value, self.z.si.value
+        self.grid = np.meshgrid(x, y, z, indexing='ij') * u.m
+        self.domain_shape = (len(self.x), len(self.y), len(self.z))
+
+        # Initiate core plasma variables
+        self.density = np.zeros(self.domain_shape) * u.kg / u.m**3
+        self.momentum = np.zeros((3, *self.domain_shape)) * u.kg / (u.m**2 * u.s)
```
colbych: This line is causing a problem when I try to use this object.

`import numpy as np
print np.zeros((3, *(2,2)))`
and the output is:
`  File "<stdin>", line 1
    print np.zeros((3, *(2,2)))
                       ^
SyntaxError: invalid syntax`

yopology: That appears to be a python2 vs python3 issue. 
(3, *dms) is a valid syntax in python3 but not in python2.

colbych: Yeah I see that now. Is this code being developed in python3.x or python2.x?

SolarDrew: Yes, that example works for me in Python3. @colbych , didn't you say you tried it in py3 and it still failed?

colbych: I did but that was apparently a lie! It crashed when I tried to import the plasma class in python 2.x and when I tried with python 3.x, but I realize now the python 3.x crash was because I don't have numpy working for python 3.x ¯\_(ツ)_/¯

Sorry for the confusion.

SolarDrew: Aha! No worries, it's easily done :)

plasmapy/plasma.py
```diff
@@ -0,0 +1,79 @@
...
+
+Defines the core Plasma class used by PlasmaPy to represent plasma properties.
+"""
+
+import numpy as np
+import astropy.units as u
+
+mu0 = np.pi * 4.0e-7 * (u.newton / (u.amp**2))
+
+class Plasma():
```
Cadair: you don't need the brackets in 3.x

plasmapy/plasma.py
```diff
@@ -0,0 +1,79 @@
...
+    """Core class for describing and calculating plasma parameters.
```
Cadair: you should agree (if you haven't already) on a convention for this, I prefer to always have a line break after and before the triple quotes.

plasmapy/plasma.py
```diff
@@ -0,0 +1,79 @@
...
+
+
```
Cadair: Only one blank line?

plasmapy/plasma.py
```diff
@@ -0,0 +1,79 @@
...
+    Parameters
+    ----------
+    domain_x : `astropy.units.Quantity`
+        1D array of x-coordinates for the plasma domain. Must have units convertable to length.
+    domain_y : `astropy.units.Quantity`
+        1D array of y-coordinates for the plasma domain. Must have units convertable to length.
+    domain_z : `astropy.units.Quantity`
+        1D array of z-coordinates for the plasma domain. Must have units convertable to length.
+    """
+    @u.quantity_input(domain_x=u.m, domain_y=u.m, domain_z=u.m)
```
Cadair: This should be:
```python
@u.quantity_input
def __init__(self, domain_x: u.m, domain_y: u.m, domain_z: u.m):
```

SolarDrew: They're the same, no? Or is this a purely aesthetic issue?

plasmapy/plasma.py
SolarDrew: The second one is supposed to have a more detailed description on it, that I haven't written yet....

plasmapy/plasma.py
Cadair: They have the same effect. Specifying them as annotations is a much better idea if you are py 3 only. Mainly you can access them at anytime using the `__annotations__` attribute.

SolarDrew: Ah, ok. Interesting.

plasmapy/plasma.py
Cadair: ...

SolarDrew: I left the blank line in to remind me to fill it with something, but time constraints prevented this. I know, I'm a terrible person.

PR #17, opened 2017-04-13
Creating a constants subpackage

namurphy:
This pull request adds a constants subpackage to PlasmaPy using the Astropy.constants.Constant class.  This includes physical constants needed for PlasmaPy.  I used some lines from si.py in Astropy's constants directory, so see the Astropy license which is copied in the licenses directory.

Potential issues for the future:
- Astropy uses the CO2010 recommended values and uncertainties for most physical constants.  CO2014 is already out, but we probably want to update these to the CO2018 recommended values and certainties when they come out.  For updates after that, we will probably want to discuss with Astropy developers their philosophy on updating constants since unit tests and other applications may depend on them giving a consistent value.
- I needed to suppress a warning message from Astropy saying that constants were already defined within Astropy.  However, it seems reasonable to fork our own constants package.
- We would need considerable remodifications if we do not want Astropy as a dependency.

[The commit history also shows me gettings confused when dealing with branches in git/GitHub even after using them for ten billion years...sigh!]


Cadair:
May I suggest that rather than re-defining the constants here you just import them from Astropy unless you need to modify them for some reason. To the users it will have the same effect but will mean you don't get the warning and you don't have to port patches from Astropy.

namurphy:
Yeah, I was wondering about doing it that way.  My motivation for redefining the constants is that Astropy is sticking with the CO2010 data for a variety of reasons, and we might want to update it to CO2014 or CO2018 to have the most accurate results.  On the other hand, if someone ends up mixing Astropy and PlasmaPy, then it would make the most sense to have consistent definitions.  I'll give your suggestion a go.  Thanks!

namurphy:
I followed the suggestion from @Cadair for now, and we can think about updating to the CO2018 values later on if it ends up becoming an issue.

plasmapy/constants/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Physical and mathematical constants for use within PlasmaPy"""
+
+from numpy import pi
+
+from astropy.constants.si import h, hbar, k_B, c, G, g0, m_p, m_n, m_e, \
```
Cadair: Most codebases I have seen use brackets to line-break long imports, I think it looks a little nicer.

namurphy:
Is this the syntax you're recommending?
```
from astropy.constants import (h, hbar, k_B, c, G, 
                                                  ...
                                                  kessel_run)
```

Cadair:
yeah that's it, indented to the opening bracket normally.

Cadair:
also ❤️ the `kessel_run` unit.

namurphy:
Fantastic, thanks!  I'll go ahead and make the change unless I forget in the next 17 seconds.  

The kessel_run unit came from this pull request for Astropy.  Unfortunately, I don't think it was accepted...  https://github.com/astropy/astropy/pull/5927

Cadair:
yeah I saw :D

SolarDrew:
:+1:

PR #20, opened 2017-04-19
Creating CONTRIBUTE.md as a guide to contributing to PlasmaPy

namurphy:
This is partially based off of the instructions from Astropy which are extremely useful. I still have some changes to push to this pull request, but I am creating a pull request while writing instructions on how to make a pull request. This still requires proofreading.  Given how confused I still get about using GitHub (sigh), it would probably be helpful for someone else to follow these instructions to see if there are ways to improve them.


namurphy:
Oh...would this information be better placed in the [wiki](https://github.com/PlasmaPy/PlasmaPy/wiki)?

SolarDrew:
It depends - I think either way we want to focus these things in one place. So if we plan on using the wiki at all then we should put it there, if not then we shouldn't.

namurphy:
I have a slight preference to put it in the repository for the time being, and then revisit the question of whether or not to put this (and other things) in the wiki at a later date.

CONTRIBUTE.md
```diff
@@ -0,0 +1,154 @@
...
+physics community.  
+
+## Sharing ideas
+
+There are several methods of communication that are being used in the
+early stages of PlasmaPy development:
+
+* [Signing up for the PlasmaPy email
+  list](https://groups.google.com/forum/#!forum/plasmapy) will allow
+  you to participate in broader discussions and keep up with .
```
SolarDrew: keep up with...?

CONTRIBUTE.md
```diff
@@ -0,0 +1,154 @@
...
+GitHub**](https://github.com/join?source=header-home).  A free account
+will suffice for you to have unlimited public repositories.  If you
+are new to [git](https://git-scm.com/), helpful resources include
+documentation on [git
+basics](https://git-scm.com/book/en/v2/Getting-Started-Git-Basics) and
+an [interactive git
+tutorial](https://try.github.io/levels/1/challenges/1).  You must also
+[install
+git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+locally on your computer.
```
SolarDrew: Maybe worth emphasising here that it really is worth getting familiar with git before trying to make code contributions? I suspect a lot of people will be tempted to skip the tutorials.

CONTRIBUTE.md
```diff
@@ -0,0 +1,154 @@
...
+following commands (after changing *your-username* to your username).
+If you would like to use HTTPS, then run:
+```
+git clone https://github.com/*your-username*/PlasmaPy.git
+```
+If you have previously [set up an SSH
+key](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/), then you may run:
+```
+git clone git@github.com:*your-username*/PlasmaPy.git
+```
```
SolarDrew: I'd add here that https is the default (and easier to set up) but that ssh is more secure, just to give people a way of deciding between the two if they're not familiar with git.

CONTRIBUTE.md
```diff
@@ -0,0 +1,154 @@
...
+
+To check on which remotes exist, run `git remote -v`.  You should get
+something like this:
+```
+namurphy	git@github.com:namurphy/PlasmaPy.git (fetch)
+namurphy	git@github.com:namurphy/PlasmaPy.git (push)
+plasmapy	git@github.com:PlasmaPy/PlasmaPy.git (fetch)
+plasmapy	git@github.com:PlasmaPy/PlasmaPy.git (push)
+```
```
SolarDrew: I like the idea of having clearer remote names but it might cause less confusion to stick to the convention of origin and upstream, since pretty much every other resource uses those.

CONTRIBUTE.md
```diff
@@ -0,0 +1,154 @@
...
+git config --global push.default simple
+```
+
+Once you have set up your fork and created a branch, you are ready to
+make edits to PlasmaPy.  Switch to your new branch by running:
+```
+git checkout *your-new-feature*
+```
+Go ahead and modify files with your favorite text editor.  Be sure to
+include tests and documentation with any new functionality.  We also
```
SolarDrew: We probably want to link to some more detailed instructions on writing good tests and docs, because it's not trivial.

namurphy: Yes, definitely.  It would be great if we could have an example set of tests and docs that we could point to in our code.  The example tests should also include how to test to make sure that exceptions come up when needed (as that is something that I need to start doing).

CONTRIBUTE.md
namurphy: That's a good point.  Th idea to change origin and upstream came from the Astropy instructions, but you're right that it could get confusing.  We could go with origin and upstream for now, and perhaps revisit this later if it becomes an issue.

SolarDrew: Ah, fair enough. Yeah, that works. I'd be happy to change it if we do get a lot of people saying that origin and upstream are confusing.

CONTRIBUTE.md
SolarDrew: Good idea.

CONTRIBUTE.md
namurphy: I suspect "a lot of people" includes me!  ...though eventually I did go to a Software Carpentry workshop which helped a lot.

SolarDrew: True, the software carptentry material is really good. Perhaps we should link to that as well?

CONTRIBUTE.md
```diff
@@ -0,0 +1,147 @@
...
+
+Next you must **clone your fork to your computer**.  Go to the
+directory that will host your PlasmaPy directory, and run one of the
+following commands (after changing *your-username* to your username).
+If you would like to use HTTPS (which is the default and easier to set
+up), then run:
+```
+git clone https://github.com/*your-username*/PlasmaPy.git
+```
+SSH is a more secure option, but requires to to [set up an SSH
```
SolarDrew: requires you to

smithsp:
Since one method of contributing is by way of pull request, I point out that you can have templates for pull requests and issues in a `.github` folder of the repo.
```
~/OMFIT-source : ls .github/
ISSUE_TEMPLATE.md        PULL_REQUEST_TEMPLATE.md
```

PR #21, opened 2017-04-21
Re-adding and PEP8ifying plasma dispersion functionality

namurphy:
One of my recent pull requests accidentally removed the plasma dispersion function and derivative.  Here I am re-adding them, shortening the analytic_functions directory to analytic, and modifying them to be consistent with PEP8 standards.


namurphy:
I just retested this branch and the added functionality seems to be working satisfactorily again.  I also added a temporary test script at the top level of the repository so that we can organize our tests until we get continuous integration testing set up.  This script is not intended to be a long-term part of the repository and can be deleted once we get Travis CI or something similar set up.

PR #24, opened 2017-05-02
Atomic physics

namurphy:
I wrote up a bunch of routines to access atomic data, including elements and isotopes.  I figured that plasma physicists occasionally care about atomic physics, and then I might have gotten just a little bit ridiculous.  For plasma parameters calculations, we'll need ion_mass.  I had fun writing the tests too, as there were a bunch of special cases and I wanted to test the exceptions too.

The routines are:
- `element_symbol`
- `isotope_symbol`
- `atomic_number`
- `is_isotope_stable`
- `half_life`
- `mass_number`
- `element_name`
- `standard_atomic_weight`
- `isotope_mass`
- `ion_mass`
- `nuclear_binding_energy`
- `energy_from_nuclear_reaction`
- `known_isotopes`
- `common_isotopes`
- `stable_isotopes`

Remaining issues are:
- The inputs for these routines are quite flexible.  `element_symbol` for example can accept inputs like `'hydrogen'`, `'H-1'`, `'D'`, `42`, `'42'`, `'protium'`, etc.  These are case insensitive except where case makes a difference: `'p'` refers to protons but `'P'` refers to phosphorus.  (Also apparently are are elements besides hydrogen.  Who knew?)  I included neutrons in a bunch of these too, though we might want to adjust whether we want functionality like `element_name('n')` returning `neutron`.
- I need to do a better job of documenting where the atomic data came from.  Most of it came from NIST.
- I am not sure what the best way to store atomic data is.  For this pull request, I wrote python code to write python code for dictionaries of dictionaries called `Elements` and `Isotopes` that contain rudimentary atomic data.  I also don't know where the best place is to deposit helper code that I used to write the code for the Isotopes data structures.
- The half-life data is woefully incomplete, and I haven't found a good source.  There are also complications that some isotopes have more than one half-life if the nucleus has multiple configurations.  This made my brain hurt.
- I should write a quick routine `isotopic_composition()` to get the terrestrial relative abundances.
- I am wondering if these routines should be put in a different directory like `physical_data` instead of `constants`.
- Probably the best way to make sure code is most readable is to use atomic and isotopic symbols consistently, like atomic symbol for an element and `'Pb-209'` for an isotope.
- I should look through this again in case there are any sections of the code that require some comments explaining what is going on (like `# magic` or `# here be dragons`), or if I should do some refactoring.
- There's a huge amount of atomic data that plasma physicists need.  We will may want to partner with databases like CHIANTI and AtomDB eventually.
- My cat says hi.


namurphy:
I just had a thought.  For several of these methods, I'm calling element_symbol or isotope_symbol at the beginning to get the element or isotope into a standard format before doing whatever else inside the routine.  I'm wondering if a more elegant/understandable way to go about this would be to have a decorator that converts the input into the standard form.

I'm also not sure what the best way is to have a standard format for ionization state.  Should it be something like `'He-4 2+'` or `'He-4 +2'` for a string representation of an alpha particle/doubly charged He-4?  Or should it be a tuple along the lines of `(element, mass_number, Z)`?

plasmapy/constants/atomic.py
```diff
@@ -0,0 +1,1187 @@
...
+    >>> element_symbol('79')
+    'Au'
+    >>> element_symbol('N'), element_symbol('n')  # Nitrogen, neutron
+    ('N', 'n')
+    >>> element_symbol('P'), element_symbol('p')  # Phosphorus, proton
+    ('P', 'p')
+
+    """
+
+    if type(argument) not in [str, int]:
```
smithsp: @namurphy `isinstance` is preferred over `type in`

plasmapy/constants/atomic.py
```diff
@@ -0,0 +1,1187 @@
...
+    symbol = isotope_symbol(argument, mass_numb)
+    atomic_mass = Isotopes[symbol]['atomic_mass']
+
+    return atomic_mass
+
+
+def ion_mass(argument, Z=1, mass_numb=None):
+    """Returns the mass of an ion.
+
+ by finding the atomic mass of a neutral
```
smithsp: Is the indentation wrong here?  Is part of your travis build a documentation build?

plasmapy/constants/elements.py
```diff
@@ -0,0 +1,304 @@
...
+                       "lawrencium": "Lr", "rutherfordium": "Rf",
+                       "dubnium": "Db", "seaborgium": "Sg", "bohrium": "Bh",
+                       "hassium": "Hs", "meitnerium": "Mt",
+                       "darmstadtium": "Ds", "roentgenium": "Rg",
+                       "copernicium": "Cn", "nihonium": "Nh",
+                       "flerovium": "Fl", "moscovium": "Mc",
+                       "livermorium": "Lv", "tennessine": "Ts",
+                       "oganesson": "Og", "neutron": "n"}
+
+Elements = {
```
smithsp: I am surprised this type of information is not tabulated elsewhere in scipy.

plasmapy/constants/tests/test_atomic.py
```diff
@@ -0,0 +1,490 @@
...
+        isotopes = common_isotopes(element)
+
+        if len(isotopes) > 0:
+
+            sum_of_iso_comps = 0
+            for isotope in isotopes:
+                sum_of_iso_comps += isotopic_composition(isotope)
+
+            assert np.isclose(sum_of_iso_comps, 1, atol=1e-6), \
+                "Problem with: " + element
```
smithsp: 👍 For good testing.

namurphy:
Thank you, @smithsp!  This is really helpful.  I'll make the change and fix the docstring which I started editing but forgot to finish.

I was surprised too that basic atomic data is not available in SciPy, and that it isn't in Astropy either.  Thinking more, elemental and isotopic data would be most useful as an upstream feature.  I'll try to remember to submit this as an issue to the SciPy repository, unless anyone else would like to do so.

namurphy:
There was a [conversation on the scipy-dev email list](http://scipy-dev.scipy.narkive.com/okrqD8xj/period-table) ten years ago about this very issue.  However, it doesn't seem like they ended up completing the implementation.

SolarDrew:
👍

PR #32, opened 2017-05-16
Expanding functionality and tests for atomic data functions

namurphy:
The purpose of this pull request is to allow string inputs to various atomic data functions to include ionization state information, and to add a function to get the charge state.  I included a bunch of tests also.  This is useful functionality to have available for the plasma parameters functions that I'm working on now.  Some examples:

```Python
>>> charge_state('Fe +26')
26
>>> charge_state('Fe++')
2
>>> charge_state('H-')  # important for stellar opacities!
-1
>>> element_name('H++')
ValueError: Cannot have an ionization state greater than the atomic number in element_name.
```

Once the atomic data routines have stabilized (which hopefully isn't too much to ask of unstable isotopes), it would be good for me to write up some overall documentation on the different routines and the different formats that these functions are compatible with.  The data on isotopic half-lives is also pretty incomplete so that will eventually need to be updated too.


namurphy:
The special cases are cluttering up the code for a few of these functions, but I think it's worth it so that the input is much more flexible.  Even so, it may be worthwhile to eventually refactor this a bit more so that the code is cleaner and more readable.

Also, trying to get the test coverage as high as possible with coveralls is addictive!

plasmapy/constants/atomic.py
```diff
@@ -687,11 +709,13 @@ def ion_mass(argument, Z=1, mass_numb=None):
     Parameters
     ----------
     argument: string or integer
-        A string representing an element or isotope, or an integer representing
-        the atomic number of an element.
+        Returns the mass of an ion by finding the standard atomic
+        weight of an element or the atomic mass of an isotope, and then
+        accounting for the change in mass due to loss of electrons from
+        ionization.
```
StanczakDominik: This is a copy of `ion_mass()`'s docstring. It might have been a mistake.

plasmapy/constants/atomic.py
```diff
@@ -746,9 +782,56 @@ def ion_mass(argument, Z=1, mass_numb=None):
...
+                              "being returned from ion_mass is outside of the "
+                              "range of known isotopes or electrons/ions.")
+
+    if isinstance(argument, str) and \
+            str(argument).lower() in ['e+', 'positron']:
+        return const.m_e
+
+    if isinstance(argument, str) and \
+            str(argument).lower() in ['e+', 'positron']:
+        return const.m_e
```
StanczakDominik: Lines 810-812 are duplicated.

plasmapy/constants/atomic.py
```diff
@@ -729,11 +757,19 @@ def ion_mass(argument, Z=1, mass_numb=None):
     uses hydrogen's standard atomic weight based on terrestrial values.  To
     get the mass of a proton, use ion_mass('p').
 
+    This function can accept a Quantity in units of mass.  If the Quantity
+    is close to the mass of an electron or positron, it will return the
+    mass of an electron or positron to full known precision.  If the
+    Quantity is within the mass range of known isotopes, it will return the
+    mass that is inputted to it.
```
StanczakDominik: Okay, that's just amazing.

plasmapy/constants/atomic.py
namurphy: Thanks for finding this!  One of the things I still need to do is to go through the docstrings and make sure everything they are consistent with what the functions are actually doing, so this is very helpful.

plasmapy/constants/atomic.py
namurphy: Another good catch.  A `for` loop would be a much more pythonic way to duplicate lines:

```Python
    for i in range(0,2):
        if isinstance(argument, str) and \
                         str(argument).lower() in ['e+', 'positron']:
            return const.m_e
```

(Just kidding!  I'll delete the duplication.)

namurphy:
Would one more person be able to review this pull request to see if it's ready to be accepted?  Kind thanks!

namurphy:
The main addition to the functionality is that doing `known_isotopes()` or `common_isotopes` or `stable_isotopes()` will provide a full list of all known isotopes, all common isotopes, or all stable isotopes across all elements.  Also `common_isotopes('Fe', most_common_only=True)` returns the isotope of iron with the highest terrestrial isotopic abundance.

We can also do `stable_isotopes('Fe', unstable_only=True)` to get a list of unstable isotopes instead of stable isotopes.  I'm wondering if it would be better to have a separate function called `unstable_isotopes()`, though I'm not sure how often that functionality will be needed.

I had forgotten what I had actually done before my short vacation, so I'm glad I put that information in the commit!

StanczakDominik:
If there is demand for `unstable_isotopes()` we could always do
```unstable_isotopes = functools.partial(stable_isotopes, unstable_only=True)```
 However, I don't know how useful or not that would be either.

PR #33, opened 2017-05-16
Basic simulation capability [WIP]

### LINKED ISSUE #7: Implement a basic 2D MHD code
One possibility for future development in PlasmaPy is adding basic simulation capabilities.  At first, the codes can be simple (e.g., 2D resistive MHD on a rectangular grid using finite differences and a straightforward time-advance).  Later, the project could include more sophisticated options and numerical methods (finite element, spectral element, finite volume, etc.).  Some thoughts on 2D MHD simulation capabilities are:
- The code should be easy to set up and use so that it can be used for educational purposes
- The code should be extremely well documented, with unit tests and test cases
- The output should be in a standard format, probably HDF5
- Following best practices for scientific computing, the code should work before it is optimized. Possibilities for optimization include:
  - Using numpy arrays and cython/numba
  - Developing it in Fortran or C/C++
- When figuring out how to develop this, we should keep in mind possibilities for longer-term development
- Related possibilities for development include
  - A basic 2D particle-in-cell code for kinetic simulations
  - Reduced MHD simulation capabilities

SolarDrew:
Closes #7 and provides a simple-to-use interface for setting up and running simulations. What's here at the moment is based largely on a similar previous project, which can be found [here](https://gitlab.com/SolarDrew/ParaSol), but I'm happy to deviate from it based on input from others.

Below are some snapshots of density from a demo 2D simulation with a plasma blob initally in the middle and diffusing outwards. It's technically solving the MHD equations but with B = 0. As you can see, I'm getting some instabilities even without magnetic field, and the less said about the boundary conditions the better.
Also speed is obviously an issue at the moment.

Density, i=0
Density, i=1000
Density, i=2000
Density, i=3000
Density, i=4000


plasmapy/classes/plasma.py
```diff
@@ -58,22 +58,152 @@ def __init__(self, domain_x, domain_y, domain_z):
...
+    @property
+    def density(self):
+        return self._density
+
+    @density.setter
+    @u.quantity_input
+    def density(self, density: u.kg/u.m**3):
+        """Sets the simulation's density profile to the specified array.
+        Other arrays which depend on the density values, such as the kinetic pressure,
+        are then redefined automatically.
```
StanczakDominik: That's clever, I like it!

plasmapy/classes/plasma.py
```diff
@@ -58,22 +58,152 @@ def __init__(self, domain_x, domain_y, domain_z):
...
+
+        density : ndarray
+            Array of density values. Shape and size must be equal to those of the simulation grid.
+            Must have units of density.
+
+        """
+
+        assert density.shape == self.grid_size, \
+            'Specified density array shape does not match simulation grid.'
+        self._density = density
```
StanczakDominik: Doesn't the previous necessitate `self._pressure = self.pressure()` after this line here?

SolarDrew: I think it doesn't, but because the comment is wrong rather than the implementation.

plasmapy/classes/plasma.py
```diff
@@ -1,14 +1,15 @@
...
+=======================
 
 Defines the core Plasma class used by PlasmaPy to represent plasma properties.
 """
 
 import numpy as np
 import astropy.units as u
-
-mu0 = np.pi * 4.0e-7 * (u.newton / (u.amp**2))
+from astropy.utils.console import ProgressBar
```
StanczakDominik: Right, because **of course** astropy would have a built-in progress bar for all things astronomical.

plasmapy/classes/plasma.py
```diff
@@ -111,6 +112,9 @@ def __init__(self, domain_x, domain_y, domain_z, gamma=5/3):
         self._energy = np.zeros(self.domain_shape) * u.J / u.m**3
         self._magnetic_field = np.zeros((3, *self.domain_shape)) * u.T
 
+        # Connect a simulation object for simulating
+        self.simulation_physics = MHDSimulation()
```
StanczakDominik: Love that you're keeping this generic!

plasmapy/classes/plasma.py
```diff
@@ -306,3 +310,48 @@ def alfven_speed(self):
...
+        >>> myplasma.simulate(max_time=1000)
+
+        >>> # Run a simulation for up to half an hour of simulation time.
+        >>> myplasma.simulate(max_time=30*u.minute)
+        """
+        if np.isinf(max_its) and np.isinf(max_time.value):
+            raise ValueError("Either max_time or max_its must be set.")
+
+        if np.isinf(max_time):
+            pb = ProgressBar(max_its)
```
StanczakDominik: Perhaps we could raise a warning here if both `max_time` and `max_its` are set?

plasmapy/classes/simulation.py
```diff
@@ -0,0 +1,100 @@
...
+        self.current_time = 0 * u.s
+        # Domain size
+        # self.grid_size = grid_size
+
+        # Physical parameters
+        # self.gamma = gamma
+
+        # Collect equations into a nice easy-to-use list
+        # self.equations = [self._ddt_density, self._ddt_momentum,
+        #                  self._ddt_energy, self._ddt_magfield]
```
StanczakDominik: Wow, I love this idea.

plasmapy/classes/plasma.py
SolarDrew: Yeah, astropy is amazing :D

plasmapy/classes/plasma.py
SolarDrew: Could do, but the question then becomes how to procede in a predictable way if there are two possible end 'times'. Stop at whichever one is sooner I suppose, but I feel like that's going to lead to unexpected behaviour if users aren't careful. I'm happy to be convinced otherwise though.

plasmapy/classes/simulation.py
SolarDrew: Thanks :) I found in the earlier version of this code that it's an easy and generalisable way to get at the functions you need.

plasmapy/classes/plasma.py
StanczakDominik: Sure, an exception does make more sense than a warning.

plasmapy/classes/simulation.py
```diff
@@ -0,0 +1,419 @@
...
+"""
+
+import numpy as np
+import astropy.units as u
+from .spatial_solvers import Solver
+from ..constants import mu0
+
+
+class MHDSimulation():
+    """Physics class for magnetohydrodynamics.
```
SolarDrew: r"""

plasmapy/classes/plasma.py
```diff
@@ -1,30 +1,31 @@
...
 
 import numpy as np
 import astropy.units as u
+from astropy.utils.console import ProgressBar
+from .simulation import MHDSimulation, dot
+from ..constants import mu0
 
-mu0 = np.pi * 4.0e-7 * (u.newton / (u.amp**2))
 
 class Plasma():
```
Cadair: You don't need the `()` in Python 3.

plasmapy/classes/plasma.py
```diff
@@ -34,46 +35,282 @@ class Plasma():
...
+        print(self.domain_shape)
+        self.gamma = gamma
 
         # Initiate core plasma variables
-        self.density = np.zeros(self.domain_shape) * u.kg / u.m**3
-        self.momentum = np.zeros((3, *self.domain_shape)) * u.kg / (u.m**2 * u.s)
-        self.pressure = np.zeros(self.domain_shape) * u.Pa
-        self.magnetic_field = np.zeros((3, *self.domain_shape)) * u.T
+        self._density = np.zeros(self.domain_shape) * u.kg / u.m**3
+        self._momentum = np.zeros((3, *self.domain_shape)) * u.kg \
```
Cadair: Why Break the line like this?

plasmapy/classes/plasma.py
```diff
@@ -34,46 +35,282 @@ class Plasma():
...
+            / (u.m**2 * u.s)
+        self._energy = np.zeros(self.domain_shape) * u.J / u.m**3
+        self._magnetic_field = np.zeros((3, *self.domain_shape)) * u.T
+
+        # Collect core variables into a list for usefulness
+        self.core_variables
+
+        # Connect a simulation object for simulating
+        self.simulation_physics = MHDSimulation(self)
```
Cadair: A better way to hold this on the `Plasma` object might be to have a method like `get_simulation`.

plasmapy/classes/plasma.py
```diff
@@ -34,46 +35,282 @@ class Plasma():
...
+    @u.quantity_input
+    def energy(self, energy: u.J/u.m**3):
+        """Sets the simulation's total energy density profile to the specified array.
+        Other arrays which depend on the energy values, such as the kinetic
+        pressure, are then redefined automatically.
+
+        Parameters
+        ----------
+
+        energy : ndarray
```
Cadair: `numpy.ndarray` here and elsewhere.

plasmapy/classes/plasma.py
```diff
@@ -34,46 +35,282 @@ class Plasma():
...
+
+    # Magnetic field
+    @property
+    def magnetic_field(self):
+        return self._magnetic_field
+
+    @magnetic_field.setter
+    @u.quantity_input
+    def magnetic_field(self, magnetic_field: u.Tesla):
+        """
```
Cadair: You are inconsistent with line breaks after `"""`

plasmapy/classes/simulation.py
```diff
@@ -0,0 +1,419 @@
...
+        self.plasma = plasma
+        # Domain size
+        # self.grid_size = grid_size
+
+        # Physical parameters
+        # self.gamma = gamma
+
+        grids = (self.plasma.x.si, self.plasma.y.si, self.plasma.z.si)
+        ranges = [grid for grid in grids if len(grid) > 1]
+        stepsize = [range[1] - range[0] for range in ranges] * grids[0].unit
```
Cadair: should try to not use `range` maybe use `range_` instead.

plasmapy/classes/simulation.py
```diff
@@ -0,0 +1,419 @@
...
+        3D grid of scalar values which are the dot products of specified
+        vectors,
+
+        .. math::
+
+           a = \\vec{v_1} \\cdot \\vec{v_2}
+
+    """
+
+    assert vec1.shape[0] == 3, "First argument provided is not a vector field"
```
Cadair: assert statements make poor error messages.

plasmapy/classes/plasma.py
StanczakDominik: Doesn't that conflict with the idea of having multiple simulation engines (MHD, PIC etc) that can interact with the Plasma class?

StanczakDominik: Oh right, yeah, I see how you could do this. Nevermind, thanks!

plasmapy/classes/plasma.py
SolarDrew: But it looks so strange without them....

plasmapy/classes/plasma.py
SolarDrew: I was probably blindly obeying Emacs when it told me the line was too long.

plasmapy/classes/plasma.py
SolarDrew: Good call, thanks.

plasmapy/classes/simulation.py
SolarDrew: You'd suggest `if things: raise SomeException` instead I assume?

Cadair: indeed.

SolarDrew:
Thanks, @StanczakDominik . If only it worked!

plasmapy/classes/simulation.py
namurphy: Good catch.  I am pretty sure I am doing things like this elsewhere, which I've been needing to go back and fix.

[For my own future reference in case I forget (hi future Nick!), I'm thinking of atomic.py with charge_state being a function in some places and an integer elsewhere]

plasmapy/classes/plasma.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, domain_x, domain_y, domain_z, gamma=5/3):
 
         x, y, z = self.x.si.value, self.y.si.value, self.z.si.value
         self.grid = np.meshgrid(x, y, z, indexing='ij') * u.m
+        self.grid = np.squeeze(self.grid)
```
StanczakDominik: So this would turn a grid of shape (1, 1, 5) into (5,), right? We should remember to make sure this won't break things like 1D simulations. I'm not saying it will, just saying it looks like it might.

plasmapy/classes/simulation.py
```diff
@@ -45,7 +46,7 @@ class MHDSimulation():
     def __init__(self, plasma):
         """
         """
-        self.dt = 0.01 * u.s
+        self.dt = 1e-4 * u.s
```
StanczakDominik: Do we intend to allow setting this via init arguments later on? Seems to me it would be applicable to only a few selected regimes otherwise, right?

plasmapy/classes/plasma.py
SolarDrew: You're right, we need to keep an eye on that. At the moment I'm just making sure that I use dimensions in the order x, y, z so that it doesn't break, but at some point I'll try and make it a bit more robust.

plasmapy/classes/simulation.py
SolarDrew: Yeah, at this point it's just a value that makes it run. Ideally I'd like to have initial dt set by the user, with the option of using an adaptive timestep.

plasmapy/examples/basic_simulation.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def gaussian(x, mean=0.0, std=1.0, amp=1.0):
+    """Simple function to return a Gaussian distribution"""
+    if isinstance(x, list):
+        x = np.array(x)
+    power = -((x - mean) ** 2.0) / (2.0 * (std ** 2.0))
+    f = amp * np.exp(power)
+    if amp == 1:
+        f = f / max(f)
```
StanczakDominik: It seems better to use `np.max` here, it's better optimized for arrays. I'm not particularly sure why do we need this `if` statement, though.

plasmapy/examples/basic_simulation.py
```diff
@@ -0,0 +1,56 @@
...
+        waves.simulate(max_its=max_i)
+        print("Simulation complete")
+
+        fig, ax = plt.subplots()
+        plt.imshow(waves.density.value, cmap='plasma')
+        plt.colorbar()
+        savedir = Path("/home/drew/PlasmaPy/plasmapy/tests/test_output")
+        if not savedir.exists():
+            savedir.mkdir()
+        plt.savefig(str(savedir/"mhd_waves_{}".format(max_i)))
```
StanczakDominik: We could probably plug `matplotlib.animation` into this. I could give this a go and put it into `plasmapy.visualization`.

plasmapy/examples/basic_simulation.py
SolarDrew: Because I've lifted this function entirely from a completely different project 😉 Good call, I'll change that or remove it, thanks.

plasmapy/examples/basic_simulation.py
SolarDrew: Yeah, I was intending to add a keyword or two to the simulation function to have it plot or animate, bit I never got round to figuring out quite how to structure it. Feel free to hack away at it 😃

StanczakDominik: https://github.com/StanczakDominik/PlasmaPy/tree/animation and hacked away it has been. This is also a work in progress, of course :smile:

StanczakDominik:
I've done a few minor changes to make checking your work out easier on multiple configurations [on my version of your branch](https://github.com/StanczakDominik/PlasmaPy/commit/edbce1f2f35785d841a278cbc43af1c53faf1d85). Might come in handy later!

namurphy:
The failing test is because of `astropy.units.atmosphere` being changed into `astropy.units.atm`, which is unrelated to today's commits.

Again, thank you!

SolarDrew:
Update: so before NAM I managed to get _some_ of the hyperviscosity implemented and running. With this in place, I've managed to run the Riemann shock tube test and the output looks like this:
riemann_shock_final
For those who aren't familiar with it, it should look something like this:
image
(image from [here](http://num3sis.inria.fr/blog/eulerian-flows-approximate-riemann-solvers-validation-on-1d-test-cases/)).

So there are still  definitely some issues that need fixing, mostly with stability, but it's getting there.

namurphy:
I haven't looked into this much, but [sympy.utilities.codegen](http://docs.sympy.org/latest/modules/utilities/codegen.html) is probably useful (for some definition of useful).  There was a [tutorial on automatic code generation with SymPy](https://www.youtube.com/watch?v=5jzIVp6bTy0) at SciPy 2017.  (These are what I mentioned in our last telecon.)

In the intermediate future we may also wish to have some finite element modeling capabilities in here too, such as with packages like [FEniCS](https://fenicsproject.org/).  This could be really helpful with non-trivial geometry like tokamaks with non-circular cross sections.

plasmapy/classes/simulation.py
```diff
@@ -476,3 +486,36 @@ def tensordiv(tensor, solver):
...
+
+    assert vec.shape[0] == 3, "First argument provided is not a vector field"
+    assert tensor.shape[:2] == (3, 3), \
+        "Second argument provided is not a tensor field"
+
+    dims = range(len(vec.shape[1:]))
+    dot = np.array([sum([vec[i] * tensor[i, 0, ...] for i in dims]),
+                    sum([vec[i] * tensor[i, 1, ...] for i in dims]),
+                    sum([vec[i] * tensor[i, 2, ...] for i in dims])]) \
+        * vec.unit * tensor.unit
```
StanczakDominik: Perhaps, if speed is an issue, this part could be optimized somehow. Two possibilities that come to mind are some crazy numpy broadcasting tricks (I'm not saying I see how to do this, though!) and JITting this part via Numba.

Numba doesn't play nice with Coveragepy, though. :disappointed:

plasmapy/examples/basic_simulation.py
```diff
@@ -53,24 +54,25 @@ def riemann_shock():
...
-    #     ax[2].plot(x, riemann.energy)
-    #     ax[3].plot(x, riemann.pressure)
-    #     ax[3].set_ylim(0, 1.2)
-    #     plt.savefig(
-    #         str(savedir/"riemann_shock_{:.4f}".format(maxt)).replace('.', '_'))
-    #     plt.close()
     t0 = datetime.datetime.now()
+    for maxt in np.arange(0, 0.21, 0.02):
+        print('- Running Simulation... ', end='')
+        riemann.simulate(max_time=maxt*u.s)
```
StanczakDominik: As I see it, we're running all pictures from 0 to t where t increases in the `arange`, right? Perhaps it would be easier on the computation to rewrite `simulation` to avoid that - go from 0 to `t_1`, take a snapshot.

In all likelihood this is something you've already thought of, but I just thought I'd raise it out in the open if anyone wants to hack away at implementing that.

plasmapy/examples/basic_simulation.py
```diff
@@ -53,24 +54,25 @@ def riemann_shock():
...
+        print("Simulation complete")
+
+        fig, ax = plt.subplots(2, 2)
```
StanczakDominik: I always liked doing it like this:

    fig, ((density_ax, velocity_ax), (energy_ax, pressure_ax)) = plt.subplots...

plasmapy/examples/basic_simulation.py
StanczakDominik: Huh, this is outdated now. Okay, git trees are magic.

plasmapy/classes/simulation.py
SolarDrew: Yeah, I feel like this could be speeded up by doing it the 'proper' numpy way. At the moment most of the viscous tensor stuff is still a bit fluid (boom boom) because I don't really fully understand and I haven't figured out how to do it with broadcasting. Premature optimisation, right...?

plasmapy/examples/basic_simulation.py
SolarDrew: Yeah, this is a bit nicer. Good call, thanks.

plasmapy/classes/simulation.py
StanczakDominik: Of course! Still, let's keep it in mind for later :)

I think I might have done something like this with `numpy.einsum` or whatever. I'll try to figure out how to do this via broadcasting a bit later.

SolarDrew: Definitely :) Cool, thanks :D

plasmapy/examples/basic_simulation.py
SolarDrew: Yep, they are :P You're right, in the long term I'd like to have the Simulation object itself be able to save a snapshot so you don't need this sort of thing, but I haven't got around to it yet so I'm still using this hack.

SolarDrew:
:sob: Fair enough. Some day I'll come back to this and make it work!

PR #34, opened 2017-05-18
Calculating plasma parameters

namurphy:
This pull request is a work in progress for plasma parameter calculations, including:
- Alfven speed
- Ion sound speed
- Electron/ion thermal speed
- Electron/ion gyrofrequency
- Electron/ion gyroradii
- Electron/ion plasma frequencies
- Debye length and Debye number
- Electron/ion inertial lengths
- Magnetic pressure/magnetic energy

**Comments**
- I am trying to have the documentation be as thorough as possible.  In particular, the docstrings should have a description of the phenomena in question that is ideally clear enough for undergraduate physics majors to understand while simultaneously being useful to people who have been doing plasma physics for about a gigasecond.
- I still need to finish some of the functions and a lot of the tests.
- I have been putting in lots of exceptions as a way for users to get as useful error messages as possible.
- I have been using a naming convention that has the full name of the quantity being calculated (e.g., `ion_inertial_length` rather than `d_i`.
- I have been trying to use names that provide the most physical intuition (e.g., `ion_gyroradius` hints at gyration, while `ion_Larmor_radius` hints that it was discovered by someone named Larmor).
- I have been trying to put everything in SI units, while allowing temperatures to be included in units of eV.
- This pull request uses functionality from https://github.com/PlasmaPy/PlasmaPy/pull/32 and will therefore fail tests.
- I am hoping that transport parameters (e.g., resistivities, viscosities, Braginskii everything and beyond, collision frequencies) can be included in a new file called `transport.py`
- There is a discrepancy of a factor of a `sqrt(2)` in different books for the calculation of electron/ion thermal velocities.  The NRL Plasma Formulary has `v_Ti = sqrt(k_B*T_i/m_i)` while some textbooks I've found have `v_Ti = sqrt(2*k_B*T_i/m_i)`.
- I'd like to put include some references for these quantities too, in particular if there are resources that are available online.

**Questions**
- What is the most accepted or robust way of calculating the Coulomb logarithm for a two-component plasma?
- Is it okay to have these functions assume a proton-electron plasma if no ion input is provided?
- Are there any important plasma parameters that this file is missing?  [I plan to add the upper and lower hybrid frequencies.  Maybe the speeds of slow and fast magnetosonic waves as functions of angles too, though that might be better elsewhere.  Suggestions for `transport.py` are welcome too!]

Thank you!  (I'll be mostly on vacation the next few days, but will probably have a chance to follow up here while taking various trains.)


plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+"""
+
+
+def Alfven_speed(B, density, ion="p"):
+    """Returns the Alfven speed.
+
+    Parameters
+    ----------
+    B : Quantity
+        The magnetic field in units such as Tesla or Gauss
```
StanczakDominik: This is probably obvious but I don't think it'd hurt to say
> The magnetic field **magnitude** in units such as Tesla or Gauss

We'll have to be careful attaching this to the Plasma class, as that one's B variable is a vector field.

StanczakDominik: Oh wait, unless we want this to be able to return "directional" Alfven speeds, say, for every point on a `Plasma`'s grid? Then that's probably fine up to possible issues with `Quantity` I wouldn't know about.

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+    >>> print(V_A2.cgs)
+    4318562.686969444 cm / s
+    >>> print(V_A2.to(u.km/u.s))
+    43.18562686969444 km / s
+
+    """
+
+    try:
+        if B.si.unit in ['1 / m3', 'kg / m3'] and density.si.unit in ['T']:
+            B, density = density, B
```
StanczakDominik: This is interesting - the intention is to have interchangeable parameter ordering, right? I suppose it's a possibility now that we have `Quantities`, but this might mess up with static type checkers or something. Still, I'd keep it in - gotta experiment with them new possibilities!!

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+
+
+def ion_thermal_speed(T_i, ion='p'):
+    """Returns the most probable speed for an ion within a Maxwellian
+    distribution.
+
+    Parameters
+    ----------
+    T_i : Quantity
+        The ion temperature
```
StanczakDominik: `The ion temperature **in either Kelvin or electron-volts**`

For consistency with other docstrings.

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+
+    Returns
+    -------
+    omega_ce: Quantity
+        Electron gyrofrequency in radians per second.
+
+    Notes
+    -----
+    The electron gyrofrequency is the angular frequency of electrons
+    gyration around magnetic field lines.
```
StanczakDominik:
```markdown
and is given by:

 .. math::
 omega_{ce} = \frac{e B}{m_e}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+
+    Returns
+    -------
+    omega_ci: Quantity
+        The ion gyrofrequency in units of radians per second
+
+    Notes
+    -----
+    The ion gyrofrequency is the angular frequency of ion gyration
+    around magnetic field lines.
```
StanczakDominik:
```markdown
and is given by:

 .. math::
 omega_{ce} = \frac{e B}{m_i}
```

I know these things are obvious to most, but hey, this was supposed to be readable for **undergrads**. ;)

StanczakDominik:
> I am trying to have the documentation be as thorough as possible. In particular, the docstrings should have a description of the phenomena in question that is ideally clear enough for undergraduate physics majors to understand while simultaneously being useful to people who have been doing plasma physics for about a gigasecond.

Awesome, I should be able to check the docs for understanding for physics undergrads then.

> I have been trying to use names that provide the most physical intuition (e.g., ion_gyroradius hints at gyration, while ion_Larmor_radius hints that it was discovered by someone named Larmor).

That seems reasonable. While it risks increasing clutter, it would be a possibility to provide an alias with a single line such as `ion_Larmor_radius = ion_gyroradius`. However, that's probably non-pythonic, as it means there's more than one obvious way to do it. Otherwise, just putting the AKA's in docstrings seems fine.

Hope you have a great vacation :)

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+
+    Returns
+    -------
+    r_L : Quantity
+        Electron gyroradius
+
+    Notes
+    -----
+    The electron gyroradius is also known as the Larmor radius for
+    electrons.
```
StanczakDominik:
```markdown
and is given by:
 .. math::
 r_L = \frac{V_{perp}}{omega_{ce}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+        Representation of the ion species.  If not given, then the ions
+        are assumed to be protons.
+
+
+
+    Notes
+    -----
+
+
+    The ion gyroradius is also known as the Larmor radius for ions.
```
StanczakDominik: Tentatively, though I see this is a work in progress:
```markdown
and is given by:
 .. math::
 r_L = \frac{V_{perp}}{omega_{ci}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+        Electron number density
+
+    Returns
+    -------
+    omega_pe: Quantity
+        Electron plasma frequency in radians per second
+
+    Notes
+    -----
+    The electron plasma frequency is
```
StanczakDominik:
```markdown
... math::
 \omega_{pe} = e \sqrt{\frac{n_e}{\epsilon_0 m_e}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+    lambda_D: Quantity
+        The Debye length
+
+    Notes
+    -----
+
+    The Debye length is the exponential scale length for charge
+    screening and is given by
+
+    MATH MATH MATH MATH MATH
```
StanczakDominik:
```markdown
... MATH::
 \lambda_D = \sqrt{\frac{\epsilon_0 k_b T_e}{n_e e^2}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+        Representation of the ion species.  If not given, then the ions
+        are assumed to be protons.
+
+    Returns
+    -------
+
+
+    Notes
+    -----
+    The ion inertial length is also known as an ion skin depth.
```
StanczakDominik:
```markdown
and is given by:
... math::
 d_i = \frac{c}{\omega_{pi}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+        Electron number density
+
+    Returns
+    -------
+    d_e : Quantity
+        Electron inertial length
+
+    Notes
+    -----
+    The electron inertial length is also known as an electron skin depth.
```
StanczakDominik:
```markdown
and is given by:
... math::
 d_e = \frac{c}{\omega_{pe}}
```

plasmapy/analytic/parameters.py
```diff
@@ -0,0 +1,852 @@
...
+    B: Quantity
+        The magnetic field in units such as Tesla or Gauss
+
+    Returns
+    -------
+    p_B: Quantity
+        The magnetic pressure in units Pascals (Newtons per square meter)
+
+    Notes
+    -----
```
StanczakDominik:
```markdown
The magnetic pressure is given by:
... math::
 p_B = \frac{B^2}{2 \mu_0}
```

plasmapy/analytic/tests/test_parameters.py
```diff
@@ -0,0 +1,126 @@
...
+        Alfven_speed(5*u.A,n_i)
+
+    with pytest.raises(u.UnitConversionError):
+        Alfven_speed(B, 5)
+
+    with pytest.raises(u.UnitsError):
+        Alfven_speed(B, 5*u.m**-2)
+
+    with pytest.raises(ValueError):
+        Alfven_speed(B, n_i, ion='spacecats')
```
StanczakDominik: :heart:

plasmapy/analytic/parameters.py
namurphy: That's a good point about directional Alfven speeds once in a while people will want that functionality.  Perhaps being able to set `directional=True` as a keyword argument would be a way to allow this.

plasmapy/analytic/parameters.py
namurphy: Correct, I was thinking of allowing interchangeable parameter ordering.  I hadn't thought about the effect there might be on static type checkers, though I have been wondering if there are other problems that could arise from this in the future.  I'm happy to try this out for a while, and modify this later on if it becomes a problem.

plasmapy/analytic/parameters.py
```diff
@@ -263,13 +294,25 @@ def electron_thermal_speed(T_e):
...
     except Exception:
         raise ValueError("Cannot find electron thermal speed")
 
     if V_Te > c:
         raise UserWarning("Electron thermal speed is greater than speed "
-                          "of light")
+                          "of light.")
+    elif V_Te > 0.1*c:
+        raise UserWarning("Electron thermal speed is greater than 10% of "
+                          "speed of light.")
```
StanczakDominik: Out of curiosity, how unusual and warning-worthy is this?

namurphy: This is one of the things I still am uncertain about.  I personally think a warning like this would be worthwhile since most of these expressions lose validity as relativistic effects become important, but I am not certain what fraction of the speed of light would be most appropriate.  For 0.1c, the Lorentz factor is ~1.005, for 0.2c, it's ~1.02, and for 0.3c it's ~1.05.  I doubt that we would have accuracy better than ~1-10% anyway in most plasma situations anyway so maybe 0.2c or 0.3c would be appropriate as a cutoff for the warning, but 0.1c is what I saw most often elsewhere so I defaulted to that.  I'm flexible on this in case anyone else has other opinions!

Come to think of it, somewhere we should have a function for the Lorentz factor.  It may be useful to include that information in the warning message.

namurphy:
In case you were wondering why I haven't worked much on this in the last week, it is because I have abandoned you and started working on PlasmaFortran instead.

namurphy:
The main things I need to do are update the docstrings and add more tests.  I will probably want to play around with exception handling too for some of the functions.  This is happily closer to done than I remembered.

plasmapy/constants/atomic.py
```diff
@@ -1349,7 +1349,7 @@ def isotopic_abundance(argument, mass_numb=None):
 
 
 def charge_state(argument):
-    """Returns the charge state of an ion.
+    """Returns the charge state of an ion or other particle.
```
StanczakDominik: Just wanted to say that this is impressively meticulous.

plasmapy/utils/checks.py
```diff
@@ -0,0 +1,168 @@
...
+    TypeError
+        If the argument is not a Quantity or units is not entirely units.
+
+    UnitConversionError
+        If the argument is not in acceptable units.
+
+    ValueError
+        If the argument contains NaNs or other invalid values as
+        determined by the keywords.
+
```
StanczakDominik: A nice touch that this doesn't simply return `True` but does all the testing inside.

plasmapy/physics/parameters.py
```diff
@@ -0,0 +1,1366 @@
...
+
+    omega_ci = ion_gyrofrequency(B, ion)
+
+    r_L = (Vperp/omega_ci).to(units.m,
+                              equivalencies=units.dimensionless_angles())
+
+    return r_L
+
+
+def electron_plasma_frequency(n_e=None):
```
StanczakDominik: Why is the default here `None`? Should this have a default argument at all?

plasmapy/physics/parameters.py
```diff
@@ -1243,3 +1225,142 @@ def magnetic_energy_density(B):
     E_B = (B**2/(2*mu0)).to(units.J/units.m**3)
 
     return E_B
+
+
+def upper_hybrid_frequency(B, n_e=None):
```
StanczakDominik: I don't quite understand why this allows `None` as a default `n_e` either.

SolarDrew:
I think the point about barrier to entry is a good one. We should definitely make sure that units are properly described in the docs if we're going to use them heavily throughout the code. As you say, we don't want to put people off when they're just starting out.

That said, I think enouraging new users to properly think about physical units when they're programming for phyics is going to benefit them significantly later on. Also, in theory new users shouldn't have much conceptual problem with what units are and what they're for, so it's then just a case of learning how to use them in this context. I appreciate that's more difficult if they're also still getting to grips with Python more generally.

Perhaps a reasonable compromise would be to continue requiring units but to allow unitless input to default to SI (or whatever system is most sensible)?

namurphy:
Thank you for the thoughtful suggestions!  

Yeah...we definitely do not have enough internal documentation about units and Quantities.  We'll probably need a discussion about `astropy.units` somewhere around our getting started page.  Following @StanczakDominik's suggestion and changing `Quantity` to `astropy.units.Quantity` to have the full namespace would be a step in the right direction.

@SolarDrew's suggestion about allowing unitless inputs and having them default to SI would be a very valid way to go.  We would want to be careful to prevent people from getting unintended results.  Here are some possible syntaxes for one function with some being more readable and some less:
```Python
>>> Alfven_speed(0.1*units.T, 8e-9*u.kg/u.m**3)  # very readable, my favorite
>>> Alfven_speed(0.1, 8e-9)  # ??????
>>> Alfven_speed(B=0.1, rho=8e-9)  # More readable, but can get confusion about units
>>> Alfven_speed(B_SI=0.1, rho_SI=8e-9)  # clear, somewhat readable
>>> Alfven_speed(B_T=0.1, rho_kg_per_m3=8e-9)  # unambiguous and readable
>>> Alfven_speed(B=0.1*units.T, rho=8e-9*units.kg/units.m**-3)  # very readable, but redundant
```
When the inputs are floats, would it be a good idea to spit out a `UserWarning` that says that SI units were assumed?  Or would it be enough to just spell this out clearly in the documentation?

For ion and electron number densities, I've been mostly trying to have them be keyword inputs so that they can be better distinguished.

Hope you're having a good weekend!

StanczakDominik:
You don't need to put 'None' or any default arguments to turn arguments into keyword args, Python is awesome like that. If you just have

    def Alfven_speed(B, rho):

You can call that equally well with `Alfven_speed(1e-1, 8e-9)`, `Alfven_speed(B = 1e-1, rho = 8e-9)` and `Alfven_speed(rho = 1e-1, B = 8e-9)`. I believe this covers all the use cases. except for passing in quantities without referring to them as keywords:

     Alfven_speed(1e-1 * u.T, 8e-9 * u.kg / u.m**3)
     Alfven_speed(8e-9 * u.kg / u.m**3, 1e-1 * u.T)

And while this would be nice, I would say it's non-critical as it kind of breaks a few norms in Python.

`UserWarning`s on float input are an awesome idea - especially given that they only pop up once, so we don't get excessive nagging.

As for SI defaulting, I believe we could do this with @namurphy's `_check_quantity` function:
* rework the `units` argument to be only one `Unit`, preferably the SI one, letting astropy check whether the `Quantity` supplied can be 
* have the function check if the number is a dimensionless number, not a quantity - a quick hack using `numbers` from the standard library looks like this, though Astropy may have something cleaner for this:

        if isinstance(arg, numbers.Number) or isinstance(arg, np.ndarray):

and if it is, raise a `UserWarning` and multiply `arg` by the one preferred SI `unit`. I'm kind of torn on whether to do this in-place (using `globals`) or to change the signature to `B = _check_quantity(B...)`.

Either way, I could probably take a shot at getting that done if you'd like.

namurphy:
Thank you for helping me learn more Python and unlearn IDL!  (I used to be a really good IDL programmer, which has given me some habits that have not been entirely helpful after switching to Python...)

I just dug up some code from a few months ago that was intended to make sure that a number thing would have a particular set of units attached to it:
```Python
from astropy import units
import warnings

def ensure_unit(argument, unit):
    if argument is None:
        return None
    elif isinstance(argument, units.Quantity):
        try:
            quantity = argument.to(unit)
        except Exception:
            raise units.UnitConversionError("The argument to ensure_unit cannot be converted to the requested unit.")
    else:
        try:
            quantity = argument*unit
        except Exception:
            raise units.UnitsError("The argument to ensure_unit cannot be given a unit.")
        else:
             warnings.warn("Assuming SI units.", UserWarning, stacklevel=2)

     return quantity
```
This was for an instance where I wanted `None` to just go through, though perhaps this would benefit from an optional keyword called `None_shall_pass`.  I have not checked that this is entirely robust.

I would vote for changing the signature to `B = _check_quantity(B...)` (assuming that my opinion is not coming from bad habits that I learned using Fortran).  I am thinking that all of this would be most elegant as a decorator like `astropy.units.quantity_input`, though that does not do everything that we want to do.  

@StanczakDominik - If you would like to implement this, please feel free to do so!  (If you're busy, no worries, as I should have a chance to get to it in a megasecond or two.)  Thank you!

namurphy:
Oh...if `_check_quantity` is changed to allow only one unit and not one of the units in a tuple, we would need to make sure that temperatures can be inputted in eV (using the `astropy.units.temperature_energy` equivalency).  

We would also need to be careful with `ion_gyroradius` and `electron_gyroradius` since I put in arguments like `Vperp_or_Te` that presently allows units of velocity, temperature, and energy (per particle).  I've been wondering if it would be better to have `Vperp` and `T_e` as separate keyword arguments.

namurphy:
I was thinking about the best ways to structure the docstrings for these functions.  Some of the suggestions that came up at SciPy 2017 include keeping things short, prioritizing, and avoiding redundancy.  Concepts should be explained in easy to understand language.  At the same time, it is important that there be a useful discussion of the physical concepts behind the ideas and to document exceptions.  It should also be easy to skim through with the most immediately needed information at the top.

How's this for an outline?
- A one or two line description of the function
- Parameters (a list of the inputs along with types accepted)
- Returns (a list of the outputs along with the output type)
- Notes (specific details needed for calling the function, e.g., that one and only one of two keywords must be included)
- Examples
- Formula (with each of the variables defined)
- Concept (a discussion in words of the physics involved)
- Raises (a list of exceptions, kept near the bottom?)
- See also
- References

Some sections will only sometimes be needed, like Notes, See also, References, Formula, and Concept.

namurphy:
This is ready for review again.  I made a bunch of changes in response to @StanczakDominik's thoughtful suggestions.  The biggest change is that floats, integers, and NumPy arrays as inputs will be interpreted as having SI units, but a UserWarning will be raised so that it will not happen silently.  

The main two things to be done in subsequent pull requests are:
- Update the docstrings so that all of them have good discussions of the physical concepts behind them.  I started to do this, but it involved thinking, so I decided to postpone much of this until later.  This sort of thing would probably be a good starter pull request for new contributors.
- Change `_check_quantity()` and `_check_relativistic()` to decorators.  
I can raise these as issues.

Later on we can try adding these functions to the base plasma class(es).

plasmapy/physics/parameters.py
```diff
@@ -0,0 +1,1486 @@
...
+    B : Quantity
+        The magnetic field magnitude in units convertible to tesla.
+
+    density: Quantity
+        Either the ion number density in units convertible to 1 / m**3,
+        or the mass density in units convertible to kg / m**3.
+
+    ion : string, optional
+        Representation of the ion species.  If not given, then the ions
+        are assumed to be protons.
```
SolarDrew: Maybe worth specifying valid inputs here? Basic, I know, but explicit is better than implicit, right?

plasmapy/physics/parameters.py
```diff
@@ -0,0 +1,1486 @@
...
+    <Quantity 4.317387185721358 cm / us>
+
+    """
+
+    if isinstance(B, units.Quantity) and isinstance(density, units.Quantity):
+        if B.si.unit in [units.m**-3, units.kg/units.m**3] and \
+                density.si.unit in [units.T]:
+            B, density = density, B
+
+    _check_quantity(B, 'B', 'Alfven_speed', units.T)
```
SolarDrew: We may have been over this, but why are you using a `_check_quantity` function rather than the astropy `@quantity` decorator?

plasmapy/physics/parameters.py
```diff
@@ -0,0 +1,1486 @@
...
+        V_A = (np.abs(B)/np.sqrt(mu0*rho)).to(units.m/units.s)
+    except Exception:
+        raise ValueError("Unable to find Alfven speed")
+
+    _check_relativistic(V_A, 'Alfven_speed')
+
+    return V_A
+
+
+def ion_sound_speed(*ignore, ion='p', T_e=0*units.K, T_i=0*units.K,
```
SolarDrew: Don't think I've seen this `*ignore` before, what does that do?

plasmapy/physics/parameters.py
namurphy: Yes, this is something I was worrying about.  Maybe: "Representation of the ion species (e.g., 'p', 'D+', 'He-4 1+', or 'alpha')."?

This is going to come up everywhere, so we will definitely need a full documentation page on this.

plasmapy/physics/parameters.py
namurphy: `@quantity_input` doesn't check if there were infinities or NaNs or complex numbers, if I understand correctly, and it doesn't allow for the case where there was more than one possible set of units for an input.  `_check_quantity` will also assume SI units for a number while spitting out a `UserWarning`.  

Now that you mention it, I should look over `@quantity_input` for Astropy 2.0 to see what has changed.  We could possibly propose a pull request to Astropy if there are only a few small things that we would like to add.

plasmapy/physics/parameters.py
namurphy: My motivation here is to make sure that everything gets called with a keyword, since there's both `T_e` and `T_i`, which means that a call like `ion_sound_speed(5*units.K, 10*units.K)` is ambiguous and confusing.  If there are any positional arguments, then this raises an exception.  Hm...do you think the variable name `ignore` could be changed to something where the motivation is more apparent?

plasmapy/physics/parameters.py
SolarDrew: Fair enough. It doesn't seem to be on the "what's new in Astropy 2.0" page. So yeah, let's go with this for now then and see if we can get this functionality into Astropy, since it'll neaten up a lot of this code :)

SolarDrew:
Looks good to me overall, I'm happy to merge it. Would be great not to have that 0.3% test coverage decrease though :P

plasmapy/physics/parameters.py
SolarDrew: Good point. Maybe an internal link to the more detailed docs then, once they exist?

plasmapy/physics/parameters.py
SolarDrew: Maybe. Although come to think of it I think keyword-only arguments are just a thing in Python 3.6, so you may not need it at all.

One moment while I investigate this...

SolarDrew: Ok, looks like you can just specify `*` as one of the arguments and anything after that is keyword-only. So:
```
def ion_sound_speed(*, ion='p', T_e=etc)
```

Whether we think that's clearer or not is another issue I guess.

StanczakDominik: I haven't seen anything like this, but I like it! If called like `ion_sound_speed(30*units.K)`, it's going to throw a `Takes 0 positional arguments but 1 was given`, which is pretty intuitive. I agree with @SolarDrew as to removing `ignore` and going with just `*`.

StanczakDominik:
I went ahead and removed the `ignore` part.

I'd say merge away, the 0.3% coverage decrease seems to be due to astropy maturing into 2.0.

plasmapy/physics/parameters.py
namurphy: Yes, definitely.  I put in some example ion strings for the time being.

namurphy:
@StanczakDominik - Breaking code is the first step towards making code better!  Feel free to try out changes like this on my branches.  

The `*` syntax is cleaner, but an advantage of having `*ignore` is that it allows a more specific exception to be raised that says what went wrong and why it is the way it is.  I suppose I'll leave it as is for now, but I don't know which is more Pythonic.  If `*ignore` is changed to `*` then some tests may need to be updated, in case a different exception is raised.

Thank you for the code review!  This discussion is leaving a good record for whoever next works on `parameters.py`.

PR #39, opened 2017-05-25
Improved exception handling for plasma dispersion function

namurphy:
This pull request has more detailed exception handling for `plasma_dispersion_function()` and `plasma_dispersion_function_deriv`.  These functions can yield a TypeError, UnitsError, or ValueError.  These exceptions are now tested.

I also added the ability for these functions to accept a `Quantity` as an input, as long as it is dimensionless.  However, the functions will not return a quantity but rather a `complex`, `float`, or `ndarray`.

### DIFF
--- a/plasmapy/analytic/analytic.py
+++ b/plasmapy/analytic/analytic.py
@@ -1,23 +1,31 @@
 """Functions related to the plasma dispersion function"""
 
+import numpy as np
+from scipy import special
+from astropy import units as u
+
 
 def plasma_dispersion_func(zeta):
     """Calculate the plasma dispersion function
 
     Parameters
     ----------
-    zeta : complex number or array
+    zeta : complex, int, float, ndarray, or Quantity
         Argument of plasma dispersion function.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


namurphy:
Thinking about this more, I am wondering if I have been using Python best practices for exception handling.  My current style is to put in detailed exceptions to maximize the likelihood of having useful error messages, though this does clutter up the code a bit more and makes the code less simple.  Also, different functions have similar exceptions, so to some extent I'm repeating code which is generally not ideal.  Are there better ways to do this?

The main recommendations on what not to do are to not do bare exceptions (e.g., use `except Exception:` instead of `except:`) and to not let exceptions silently pass:
```Python
try:
    You_shall_not()
except GandalfException:
    pass    # never do this!
```

SolarDrew:
Explicit is better than implicit. I'd say your exceptions are a sensible amount of detailed, they don't bog the code down excessively but they give a decent description of why there's an error.
Will try and get time to have a play with this code tonight, but it looks mergeable enough to me :+1:

PR #41, opened 2017-05-30
Exception docstrings and negative charge states in atomic routines

namurphy:
The changes to the docstrings are to make sure that the exceptions that are listed correspond to the exceptions that are called in the code.  The other change is that some of these functions now raise a UserWarning when the charge state is -4 or below, though we could potentially change this to -3 or below based on advice from my upstairs neighbor who is a chemist.

[These are fairly minor changes.]


namurphy:
Thanks!  I'll merge this since one review is probably enough for a relatively minor change.

PR #43, opened 2017-05-31
Docs [WIP]

colbych:
Closes #43.

Used Sphinx to generate some docs in the docs dir. This is not expected to be pulled anytime soon, just putting it up here so other people can see the direction I was going with this and give feedback.

Right now you can look at the index.rst and you can go through a web and look at the auto documentation.


docs/conf.py
```diff
@@ -0,0 +1,154 @@
...
+version = '0.0'
+# The full version, including alpha/beta/rc tags.
+release = '0.0.1'
+
+# The language for content autogenerated by Sphinx. Refer to documentation
+# for a list of supported languages.
+#
+# This is also used if you do content translation via gettext catalogs.
+# Usually you set "language" from the command line for these cases.
+language = None
```
StanczakDominik: Huh, I wasn't aware `sphinx` could do multiple language versions of docs. That's cool! I could help out with translating this to Polish if this ever gets that big. :)

SolarDrew:
I say we merge it now and add the actual contents of the docs in new PRs. I'll merge it tomorrow morning if there are no objections from @colbych (or anyone else) before then.

PR #53, opened 2017-07-18
Updates to licenses and removing README files

namurphy:
This is a pull request for a few minor organizational changes.
- I moved our license to the top-level directory to make it more obvious.
- I updated the dates of Astropy's license.
- I removed some nearly blank `README.md` files in subdirectories since we haven't been using them, and I would vote to have the equivalent documentation kept in the `__init__.py` files instead (e.g., embedding the documentation in the code).  I've usually been 
- I added a few placeholder `__init__.py` files to directories that do not yet have much in them.  These are in place of the `README.md` files.
Do these changes seem helpful?


namurphy:
The failing test is happening because `astropy.contants.atmosphere` changed to `astropy.constants.atm` when going from version 1.3.1 to 2.0.0.  I have a fix for this in pull request https://github.com/PlasmaPy/PlasmaPy/pull/34 so we shouldn't have to worry about it here.  

With all that said, if Astropy were to bar changes like this, it sure would take a weight off of our shoulders.

Issue #52, opened 2017-07-03
What are our atomic data needs?


namurphy:
Over the last few years, I have gradually learned that plasmas tend to contain atoms.  Hence, we will need to include some degree of atomic physics within PlasmaPy.  The atomic data needs in plasma physics are very broad.  For example, temperatures of interest range from ~10 K in molecular clouds to ~30-40 MK in solar flares.  Astronomers tend to care about the dozen or so most abundant elements in the universe, while laboratory plasma physicists care a lot about impurities leeching in from the plasma-facing wall which is often from different materials.  Dust also plays important roles in laboratory and astrophysical plasmas.

Including the ability to access all of the information any plasma physicist might ever need would be a never-ending project that would increase the size of this package by orders of magnitude.  However, the ability to access important atomic data will be extremely useful for numerous plasma physicists across sub-disciplines.  Hence we will need to do a lot of strategizing about what the most important needs are and how best to fill them.  Some questions are:
* What atomic data sets are most broadly needed by plasma physicists?
* How and to what extent should PlasmaPy fulfil atomic physics needs?
* What are the existing databases, and how would it be best to partner with them?
* Are there atomic data needs that are not being fulfilled by any databases?
* What should we include in the core package in our version 0.1 release?

A subset of the atomic data needs are:
* Ionization and recombination rates
  - Collisional ionization, photoionization, etc.
  - Radiative recombination, dielectronic recombination, etc.
* Cooling rate coefficients (e.g., for a given ion at a given temperature)
* Collisional cross sections
* Charge exchange cross sections
* Photoionization cross sections
* Everything associated with plasma spectroscopy and radiative transfer (which would be huge amounts of data and would undoubtedly be too much)
* Ability to work with both Maxwellian and non-Maxwellian distribution functions

Occasionally there are nicely behaved functions that would be easy to implement such as the radiative recombination rate.  Things like that are definitely worth including.

I am not sure what the best ways to proceed are.  Some of our many options are:
* Include the most important atomic data within the core package (or not)
* Have an affiliate package that would be responsible for atomic data needs
* Have query or bridge routines within the core package that would get the requested atomic data from websites or the local directories that contain the databases
* Or instead, not include much in the way of atomic data within PlasmaPy and instead just put up some documentation that lets users know how/where to access atomic data using other databases/Python packages.

I am imagining that this will be a long-term conversation, but I figured it would be worth bringing up sooner rather than later.  Any thoughts or ideas?

[...]

Issue #48, opened 2017-06-08
Improve functionality of atomic subpackage


namurphy:
Thus far in the atomic routines, there are a few ways to represent ions in the form of a string: `'p'`, `'proton'`, `'H-1+'`, `'H-1 1+'`, and `'H-1 +1'` all represent protons.  Another possibility would be to represent them as a tuple or class that contains the ion species and the ionization state: e.g., `('H-1', 1)` would also represent a proton.  Perhaps `(1, 'H-1')` could be treated as equivalent to `('H-1', 1)`, if we wanted to maximize flexibility.  This is similar to but much simpler than the `ParticleSpecies` class that was thoughtfully suggested in issue https://github.com/PlasmaPy/PlasmaPy/issues/40 for a particle stepper.  

Also, I found another issue for isotopes with plus or minus signs to indicate the charge state:
```Python
>>> charge_state('H-1+')  # this works fine
1
>>> charge_state('H-1-')  # this should be -1 but instead just makes my computer sad
ValueError: The following input does not have valid charge state information: H-1-
```
If the first syntax works, then I was thinking that the second syntax should also work, except if there is a chance of ambiguity.  An additional like in `test_atomic.py` to test that would be:
```Python
assert charge_state('H-1-') == -1
```
I'm still thinking about this, as there is a chance of ambiguity.  Another possibility would be to allow syntax like `'H-1 +'` where there is a space between the isotope and the plus sign.  

Now to see if I can figure out a checklist:
- [ ] Allow ions to be represented as a tuple like `('H-1', 1)`.
- [ ] Allow ions to be represented like `'H-1-'` and act the same way as `'H-1+'`
- [ ] Allow ions to be represented like `'H-1 +'` or `'H-1 -'`

namurphy:
Here are some antimatter matters with inconsistent or undesirable behavior:
```Python
>>> charge_state('p-')
-1
>>> charge_state('e+')
ValueError: Invalid element or isotope information in charge_state
>>> charge_state('positron')
ValueError: Invalid element or isotope information in charge_state
>>> charge_state('antiproton')
ValueError: Invalid element or isotope information in charge_state
```

namurphy:
We may want `charge_state` to allow electrons, even though it was intended for ions.
```Python
>>> charge_state('electron')
ValueError: Invalid element or isotope information in charge_state
>>> charge_state('e-')
ValueError: Invalid element or isotope information in charge_state
```

[...]

PR #54, opened 2017-07-18
Particle stepper for #40 [WIP]

### LINKED ISSUE #40: Add a particle stepper for given external fields
As discussed on HipChat recently, I'd like to give adapting my particle stepper a try. The way I see this working would be:

- [x] Add a `ParticleSpecies` class containing their basic information such as position and velocity (as numpy `ndarray`s), charge and mass. This seems like a minimal necessary set of information. I would also like to save historical data on their positions and velocities every few timesteps into a larger ndarray to avoid frequent file IO (the iteration step between saving could be given by user, according to the resolution required).
- [x] Give them a `push` method that takes electric and magnetic fields (`ndarray`s) at particle locations (so they're the same shape) and updates their positions and velocities. I have two Boris pushers, one nonrelativistic and one relativistic - I can port both. Of course, we want velocities at half timesteps to help with energy conservation - adding another method `init_push` (without updating positions) to start that is trivial.
- [x] Add functionality for interpolating fields from - initially - Cartesian grids to particle positions. I've done this in 1D. Given that we want 3D, I'll try to find an algorithm that implements that - I have a few papers with those lying around.
- [ ] I have to say that I have absolutely no idea on how to do this in non-cartesian geometries, so that's something I'd like to postpone until later. As far as I can tell our current idea for the `Plasma` class is to stay Cartesian for a bit, then branch out a little, so that seems reasonable to me.

I should be able to start working on this by about June 7th or so. Any help brainstorming before I get cracking on this would be much appreciated!

StanczakDominik:
Closes #40 .

I've got the basics going, going to keep adding tests to this. This is heavily work in progress, probably not really PEP8-compliant et cetera.

Here's what I have right now, for a very basic test in a uniform magnetic field:

figure_1

- [x] Need to start fitting sines to trajectories in the XY plane for this test
- [x] Add functionality for visualizing trajectories for multiple particles
- [x] Need to test in non-uniform fields
- [x] Need to figure out electric fields
- [x] Need to incorporate @namurphy 's atomic data system for charge and mass information

There's a conflict in `plasmapy/constants/__init__.py` that seems to be caused by the switch to Astropy 2.0.

I'm adding this in as a reference for further tests:
https://en.wikipedia.org/wiki/Guiding_center


namurphy:
@StanczakDominik - While you're incorporating the atomic data functionality, I'll be very interested if you have any ideas on how to improve it!  

[linking in issues related to atomic functionality: https://github.com/PlasmaPy/PlasmaPy/issues/48 and https://github.com/PlasmaPy/PlasmaPy/issues/52]

StanczakDominik:
figure_1
Here's a trajectory for the ExB drift test case.

StanczakDominik:
All right, I believe this is pretty much done up to docstring and refactoring. I was planning on testing the stepper in non-uniform electric fields, but I'm not particularly sure about how to actually combine that with the interpolated field! The drift I meant was (copied from Wikipedia):

image

So yeah, I'm not particularly sure how to test this, but it seems to me it's working decently-ish. I'd appreciate any nitpicking!

namurphy:
Thank you, this looks great!  I'll take a quick look through it now and see if I can come up with any suggestions.

antoinelpp:
It's great !! :+1: 
I've try to use it by checking the drift velocity (by fixing **E**x**B** and comparing the velocity), and it works well.
The Boris scheme is nice, but if the magnetic field is only along 1 direction, or if there is not magnetic field, it if faster to use a simple leap-frop scheme. Do you this we should implement it ?

I have one remark concerning the `species` classe : including dT, NT and the history as attribute might complicate the use of the class latter. 



There is the code I made (I'm not sure it's the best way to use it, feedback welcome ! :) 

```python
import numpy as np
from plasmapy.classes import Plasma, Species
```


```python
#init plasma class
#Domain 2x2x2 m centered in (0,0,0) 
plasma = Plasma(domain_x=np.linspace(-1, 1, 10)*u.m,
                domain_y=np.linspace(-1, 1, 10)*u.m,
                domain_z=np.linspace(-1, 1, 10)*u.m)

#initialing the fields
#B = Bx
B0 = 4 * u.T
plasma.magnetic_field[0,:,:,:] = np.ones((10,10,10)) * B0
#E = Ey  perp to B
E0 = 2 * u.V/u.m
plasma.electric_field[1,:,:,:] = np.ones((10,10,10)) * E0 

#init particle
species = Species(plasma,'p',1,1,1e-10*u.s,10000)

#init  velocity
V0 = 0 * (u.m / u.s)
species.v[0][1] = V0
species.v[0][2] = V0

#run the pusher for NT dt
species.run()

#priintout to check
#print(species.v,species.x,species.eff_m,species.eff_q)

#process data
species.plot_time_trajectories()

#calculate the drift velocity :
#mean velocity 
vmean = np.sqrt(np.sum(np.mean(species.velocity_history,axis=0)**2))


print(f"calculated drift velocity is {vmean} to compare with the theoric E0/B0 = {E0/B0}")

#and from position :
Vdrift = species.x/(species.NT*species.dt)
normdrift = np.sqrt(np.sum(Vdrift**2))
print(f"calculated drift velocity from position is {normdrift}")
```


    calculated drift velocity is 0.5002180366502276 m / s to compare with the theoric E0/B0 = 0.5 V / (m T)
    calculated drift velocity from position is 0.5002180366502293 m / s



```python

```

StanczakDominik:
Awesome that it works! I had something similar in a test, but your execution is neat as well. Would you mind if I put it in a jupyter notebook and used it as an example of the usage while I'm fixing up docstrings for this one?

antoinelpp:
Sure go ahead ! :smile:

StanczakDominik:
All right, the docs are done, there's an example notebook, if anyone wants to review this, I'd be much obliged :)

jhillairet:
Hi,

Few comments after trying the stepper on a different project:
- calling Species() with nt=1e3 leads to a TypeError in line 78 of species.py, while nt=1000 is OK
- I didn't found how to define an electron instead of a proton when calling Species()... Seems not exist in atomic?
- using a high number of point for the domain will rapidly leads to memory problem (notably because of the use of meshgrid). 

- I think there is problem with the Plasma magnetic and electric field. If ones try to import field map, for example having only Bx and Bz (By assumed to be 0), when creating the Species(), the RegularGridInterpolator interpolates in the first N dimensions. Indeed, from the documentation of [RegularGridInterpolator ](https://docs.scipy.org/doc/scipy-0.18.0/reference/generated/scipy.interpolate.RegularGridInterpolator.html#scipy.interpolate.RegularGridInterpolator), the interpolated value is an array of shape (m1, ..., mn, **...**). The important point is than supplementary dimensions are at "the end", not at the beginning. So if B is a vector (Bx,By,Bz), then instead of being of shape (3,nx,ny,nz) is should be defined as (nx,ny,nz,3).

I guess I'm not very clear, so here is an example:
the following will work :
`
B = np.zeros((171,1,901,3))
B[:,0,:,0] = B_x*u.T
B[:,0,:,1] = 0*u.T
B[:,0,:,2] = B_z*u.T

B_inter = RegularGridInterpolator(
    (domain_x, domain_y, domain_z), B
)
`
while this one (the present implementation) will fail:
`
B = np.zeros((3,171,1,901))
B[0,:,0,:] = B_x*u.T
B[1,:,0,:] = 0*u.T
B[2,:,0,:] = B_z*u.T

B_inter = RegularGridInterpolator(
    (domain_x, domain_y, domain_z), B
)
`

StanczakDominik:
@jhillairet thanks for the heads up!

> - calling Species() with nt=1e3 leads to a TypeError in line 78 of species.py, while nt=1000 is OK

The `TypeError` is due to me forgetting to force integer timesteps. That should be fixed already.

> - I didn't found how to define an electron instead of a proton when calling Species()... Seems not exist in atomic?

I didn't want to duplicate functionality from @namurphy's #60 pull request, that change is in there. It should work as `'e'` once that's pulled into master.

> - using a high number of point for the domain will rapidly leads to memory problem (notably because of the use of meshgrid). 

I'm not certain as to the best course of action here - `meshgrid` has a `sparse` keyword which I thought could alleviate this, but it still errors.  @SolarDrew, how large `plasma` instances have you been using in the MHD solver ?

For now, I cleaned the code calling `meshgrid` up a bit.

> - I think there is problem with the Plasma magnetic and electric field for non-scalar field, since  RegularGridInterpolator interpolates in the first N dimensions. Indeed, from the documentation of [RegularGridInterpolator ](https://docs.scipy.org/doc/scipy-0.18.0/reference/generated/scipy.interpolate.RegularGridInterpolator.html#scipy.interpolate.RegularGridInterpolator), the interpolated value is an array of shape (m1, ..., mn, **...**). The important point is than supplementary dimensions are at "the end", not at the beginning. So if B is a vector (Bx,By,Bz), then instead of being of shape (3,nx,ny,nz) is should be defined as (nx,ny,nz,3).
> 
> I guess I'm not very clear, so here is an example:
> the following will work :
> ````
> B = np.zeros((171,1,901,3))
> B[:,0,:,0] = B_x*u.T
> B[:,0,:,1] = 0*u.T
> B[:,0,:,2] = B_z*u.T
> 
> B_inter = RegularGridInterpolator(
>     (domain_x, domain_y, domain_z), B
> )
> ````
> 
> while this one (the present implementation) will fail:
> 
> ```
> B = np.zeros((3,171,1,901))
> B[0,:,0,:] = B_x*u.T
> B[1,:,0,:] = 0*u.T
> B[2,:,0,:] = B_z*u.T
> 
> B_inter = RegularGridInterpolator(
>     (domain_x, domain_y, domain_z), B
> )
> ```

That's a good point. The `magnetic_field` attribute of the `Plasma` class was in the shape `(3, n_x, n_y, n_z)` when I started extending that, so I went with that convention. While `RegularGridInterpolator` does require the vector's directions to be located at the last entry in the `shape` array, the array I'm passing to it is transposed:

```
self.B_interpolator = RegularGridInterpolator(
            (self.plasma.x.si.value,
             self.plasma.y.si.value,
             self.plasma.z.si.value),
            self.plasma.magnetic_field.T.si.value,
            method="linear",
            bounds_error=True)
```

Note especially the `T` 

    self.plasma.magnetic_field.T.si.value,

I kind of didn't think we'd need to expose the interpolator to the outside world. But for all I know this could be simply a terribad implementation by me. Given we wanted to have interoperability between the MHD solvers, I guess @SolarDrew could have a say in this one.

All in all, thanks for the helpful feedback @jhillairet , I'd really appreciate further discussion on this one!

jhillairet:
> That's a good point. The magnetic_field attribute of the Plasma class was in the shape (3, n_x, n_y, n_z) when I started extending that, so I went with that convention. While RegularGridInterpolator does require the vector's directions to be located at the last entry in the shape array, the array I'm passing to it is transposed:
>
>```
>self.B_interpolator = RegularGridInterpolator(
>            (self.plasma.x.si.value,
>             self.plasma.y.si.value,
>             self.plasma.z.si.value),
>            self.plasma.magnetic_field.T.si.value,
>            method="linear",
>            bounds_error=True)
>```
Concerning that point, the transpose trick may work only for field having same spatial dimensions. For example, in your notebook, using the following will lead to an error, since `RegularGridInterpolator `will expect `value `to be shaped as (nx, ny,nz,...), while the transpose of the B/E fields leads to a shape of (nz,ny,nx,3) instead:

```
plasma = Plasma(domain_x=np.linspace(-1, 1, 10)*u.m,
                domain_y=np.linspace(-1, 1, 20)*u.m,
                domain_z=np.linspace(-1, 1, 30)*u.m)

B0 = 4 * u.T
plasma.magnetic_field[0,:,:,:] = np.ones((10,20,30)) * B0

E0 = 2 * u.V/u.m
plasma.electric_field[1,:,:,:] = np.ones((10,20,30)) * E0 

# plasma.magnetic_field.shape : (3, 10, 20, 30) !
species = Species(plasma,'p',1,1,1e-10*u.s,10000)
```
`--> ValueError: There are 10 points and 30 values in dimension 0`

Here are some solutions I see:
- interpolating each component independently and re-combining
- re-aranging position vector to (nx*ny*nz, 3), as well as field (nx*ny*nz,3), then using interp1d. 
- put the field components at last ?

SolarDrew:
> I'm not certain as to the best course of action here - meshgrid has a sparse keyword which I thought could alleviate this, but it still errors. @SolarDrew, how large plasma instances have you been using in the MHD solver ?

Up to around 128^3, if I remember rightly, though only for testing purposes of course. I've wondered (and metioned a few times) about using yt for representing some of the core plasma stuff, as it has a nice way of structuring data fields that would actually probably work well for us. I think it can also read data directly from a file rather than loading it into RAM, so that might be a way around memory issues. I'd have to check that though, I could be completley wrong.

As for the magnetic field, personally I think it should be ordered (3, nx, ny, nz). This is mostly for ease-of-use for the user - `bx = plasma.magnetic_field[0]` is more intuitive than `bx = plasma.magnetic_field[:, :, :, 0]`. That said, we can always just have extra attributes like `magnetic_field_x` and abstract it away from the user. Similiarly I think @StanczakDominik's point about the user not seeing the interpolator is reasonable - the user-facing code is the part that most needs to be clean and intuitive, and as long as everything underneath works, it can be a bit less polished.

> Concerning that point, the transpose trick may work only for field having same spatial dimensions.

Valid point. The easiest fix I see is to transpose the x, y, and z arrays as well, since you're transposing the magnetic field.

StanczakDominik:
The test coverage is now fixed and I believe that if we can live with this failing at high resolutions, which is a more general issue that could be fixed later, this could reasonably be merged for now - if nobody has any objections, that is! I think merging my own code on something like this one may be unwise.

SolarDrew:
Don't wait for me if other people approve it first but I'll try and have  a proper look over this tomorrow.

plasmapy/__init__.py
```diff
@@ -15,5 +15,5 @@
 import sys
 import warnings
 
-if sys.version_info[:2] < (3, 6):  # coveralls: ignore
+if sys.version_info[:2] < (3, 6):  # pragma: no cover
```
SolarDrew: You should pull in changes from upstream so this doesn't undo all my hard work :P

StanczakDominik: Oh. I seem to have been editing those files in parallel. I'll do some rebasing here and limit my changes to Species and minor changes in Plasma.

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+        self.scaling = scaling
+        self.eff_q = self.q * scaling
+        self.eff_m = self.m * scaling
+
+        self.plasma = plasma
+        if np.isinf(dt) and np.isinf(nt):
+            raise ValueError("Both dt and nt are infinite.")
+
+        self.dt = dt
+        assert self.dt.si.unit == u.s
```
SolarDrew: I think we should use the astropy `@u.quantity_input` decorator for this.

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+        self.NT = int(nt)
+        self.t = np.arange(nt) * dt
+
+        self.x = np.zeros((n, 3), dtype=float) * u.m
+        self.v = np.zeros((n, 3), dtype=float) * (u.m / u.s)
```
SolarDrew: Would it be practical (or useful) to make these into properties that just return the most recent values from `position_history` and `velocity_history`?

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+        --------
+        `astropy.units.Quantity`
+            Array of kinetic energies, shape (nt, n).
+        """
+        return (self.velocity_history ** 2).sum(axis=-1) * self.eff_m / 2
+
+    def boris_push(self, init=False):
+        """
+        Implements the Boris algorithm for moving particles and updating their
+        velocities.
```
SolarDrew: It might be worth having a little more description, some maths and/or a link in the dosctring, for those who aren't familiar with the algorithm.

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        self.boris_push(init=True)
+        for i in range(self.NT):
+            self.boris_push()
+            self.position_history[i] = self.x
+            self.velocity_history[i] = self.v
+
+    def __repr__(self, *args, **kwargs):
+        return f"Species(q={self.q:.4e},m={self.m:.4e},N={self.N}," \
+               f"name=\"{self.name}\",NT={self.NT})"
```
SolarDrew: +1 for format strings. Format strings are the best thing.

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def run(self):
+        """
+        Runs a simulation instance.
+        """
+        self.boris_push(init=True)
+        for i in range(self.NT):
+            self.boris_push()
+            self.position_history[i] = self.x
+            self.velocity_history[i] = self.v
```
SolarDrew: for i=0 this will overwrite the initial values that you're defining before the loop, no? (Not familiar with the algorith myself, maybe that's fine and I'm just showing my ignorance.)

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+        return f"Species(q={self.q:.4e},m={self.m:.4e},N={self.N}," \
+               f"name=\"{self.name}\",NT={self.NT})"
+
+    def __str__(self):
+        return f"{self.N} {self.scaling:.2e}-{self.name} with " \
+               f"q = {self.q:.2e}, m = {self.m:.2e}, " \
+               f"{self.saved_iterations} saved history " \
+               f"steps over {self.NT} iterations"
+
+    def plot_trajectories(self):  # pragma: no cover
```
SolarDrew: This should change to `#coveralls: ignore` if you pull in upstream. Also I think there is a plugin or some such for testing plot outputs with pytest so you may just be able to test this function anyway. I'll double-check this and send you a link.

SolarDrew: This should change to `#coveralls: ignore` if you pull in upstream. Also I think there is a plugin or some such for testing plot outputs with pytest so you may just be able to test this function anyway. I'll double-check this and send you a link.

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -67,10 +69,11 @@ def test_Plasma_derived_vars():
...
     # Set an arbitrary uniform values throughout the plasma
-    test_plasma.density = 2.0 * np.ones((8, 8)) * u.kg / u.m**3
-    test_plasma.momentum = 10.0 * np.ones((3, 8, 8)) * u.kg / (u.m**2 * u.s)
-    test_plasma.pressure = np.ones((8, 8)) * u.Pa
-    test_plasma.magnetic_field = 0.01 * np.ones((3, 8, 8)) * u.T
+    test_plasma.density[...] = 2.0 *  u.kg / u.m**3
+    test_plasma.momentum[...] = 10.0 * u.kg / (u.m**2 * u.s)
+    test_plasma.pressure[...] = 1 * u.Pa
+    test_plasma.magnetic_field[...] = 0.01 * u.T
+    test_plasma.electric_field[...] = 0.01 * u.V / u.m
```
SolarDrew: Nice. Hadn't thought of doing it like that but I definitely prefer it to my version.

plasmapy/classes/plasma.py
```diff
@@ -53,15 +53,16 @@ def __init__(self, domain_x, domain_y, domain_z):
         self.y = domain_y
         self.z = domain_z
 
-        x, y, z = self.x.si.value, self.y.si.value, self.z.si.value
-        self.grid = np.meshgrid(x, y, z, indexing='ij') * u.m
+        self.grid = np.array(np.meshgrid(self.x, self.y, self.z,
```
StanczakDominik: Yeah, this seems to be working nicely now.

plasmapy/classes/species.py
```diff
@@ -0,0 +1,231 @@
...
+    a group of particles with a particular initial velocity distribution.
+
+    Parameters
+    ----------
+    plasma : `Plasma`
+        plasma from which fields can be pulled
+    type : str
+        particle type. See `plasmapy.constants.atomic` for suitable arguments.
+        The default is a proton.
+    n: int
```
StanczakDominik: Sure, done already.

plasmapy/classes/species.py
StanczakDominik: I'm not quite sure. The way I'm doing this is that `x` and `v` do all the live calculation, so they're small arrays that can be kept in memory, while `position or velocity_history` do all the memory storage. This way, it's easier to convert the history into, say, a HDF5 dataset. I'll port that code a bit later.

StanczakDominik:
There were a lot of commits here and the git history was particularly tangled, so I squashed them. I squashed them all. And not just the commits, but the merges and rebases too.

I also limited the tests to running on a single species for each one - this cuts down on the testing duration.

This may be good to go!

SolarDrew:
You monster!

Are the tests still valid if you're only testing one species? The code doesn't do spectacularly different things for each one?

StanczakDominik:
Yeah, it's basically the same thing, I just pull charge and mass information from species. Besides, I've got two different tests in here, `test_particle_uniform_magnetic` and `test_particle_exb_drift`. I used to run the first one thrice with three different species to check they were doing the same thing, which they were, and set the third one to a proton. Now I have one species for the first test and one for the latter. It seems to be working nicely.

If we ever pull out simulation into a subpackage, the number of tests can be easily increased. Otherwise, I wouldn't want to slow down tests for, say, `atomic`, as those are lightning fast.

SolarDrew:
Seems reasonable. @PlasmaPy/plasmapy-community if there are no more objections I'll merge this.

jhillairet:
I think you can merge as a good starting point. We will see later for solving problems such as non-regular grid etc.

tulasinandan:
Great work @StanczakDominik ! I think we should include some more pushers, specifically the Vay pusher [1] and RK4. The reason to include RK4 is that in our work we see regularly that RK4 is better at correctly tracing the trajectory while Boris is better at conserving energy. So if someone is interested in following accurate trajectories, they might want to go with RK4. Let me know if you need help with that.

tulasinandan:
Oh for sure! Let's push the 0.1 release first!

PR #58, opened 2017-07-19
Omitting __init__.py in coverage tests

StanczakDominik:
Given that `__init__.py` files don't really do much except exposing, and with Astropy graduating to 2.0 there's a bunch of trouble with their imports, it seems to make little sense to test them. This is a small commit that removes them from Coveralls testing.

I'm not sure whether removing those is good or bad for our testing (probably "neutral without annoyances"), but it's something to consider. Feel free to close this if it's unhelpful!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,8 @@
 [report]
-omit = 
-	ci-helpers/*
-	*/tests/*
+omit =
+ 	ci-helpers/*
+ 	*/tests/*
+	*__init__.py
+exclude_lines =
+    def __repr__
+    def __str__


SolarDrew:
👍 I'm not completely convinced that "stop testing the things that are failling the tests" is particularly sound as an idea in itself :P But I agree not much is happening in the `__init__.py`'s, so I'm happy to ignore them.

namurphy:
An interesting idea!  I'm definitely in favour of further customizing coverage reports.  I'm slightly hesitant to exclude all `__init__.py` files if we can more specifically target certain parts of the code instead that are not expected to run in the tests.  I'm not that familiar with coveralls so I looked up some documentation on [excluding code from coverage.py](http://coverage.readthedocs.io/en/coverage-4.2/excluding.html).

We have this code in our main `__init__.py`:

```Python
if sys.version_info[:2] < (3, 6):
    warnings.warn("PlasmaPy does not support Python 3.5 and below")
```

If we only test Python version 3.6 and above, then there's no point in including this in the coverage report.  Adding the comment `# pragma: no cover` should prevent a line or code block from being included in the coverage reports:

```Python
if sys.version_info[:2] < (3, 6):  # pragma: no cover
    warnings.warn("PlasmaPy does not support Python 3.5 and below")
```

It would probably also be useful for some defensive exceptions that are unnecessary to test:

```Python
def print_something(something):
    try:
        print(something)
    except Exception:  # pragma: no cover
        raise RunTimeError("Really weird and unexpected error in print_something")
```
We can also use regular expressions for code blocks.  If we do something like this:
```
[report]
exclude_lines =
    *
```
I wonder if we'll get really good coverage reports by excluding all the lines, or a zero division error!

namurphy:
Thinking a little more, I'm not sure that `# pragma: no cover` is very readable.  Would it be better to use something like `# exclude from coverage reports`?

StanczakDominik:
I like the suggestion for a custom `# coveralls: ignore` directive! I hadn't considered that.

SolarDrew:
So to summarise: the plan is to implement a coveralls ignore directive and leave the `__init__.py`s in the tests. If nobody objects before tomorrow morning I'll close this PR and make a new issue to that effect.

PR #62, opened 2017-07-29
Inherit yt's Impostor Syndrome Disclaimer

StanczakDominik:
Can we all just agree that this is lovely enough to include?


namurphy:
I definitely support this!  Thank you for thinking of it!

PR #63, opened 2017-07-29
Add note about requirements

GrahamGoudeau:


StanczakDominik:
Oh yeah, true. Thanks for that!

PR #64, opened 2017-07-30
WIP: Decorators

### LINKED ISSUE #55: Change _check_quantity() and _check_relativistic() to decorators
When working on `parameters.py`, I created two temporary methods in `utils/checks.py`: `_check_quantity` and `_check_relativistic`.  `_check_quantity` does a lot of checking to make sure that a Quantity has the right units or gets the right units attached to it and has valid values.  `_check_relativistic` makes sure that the output of a function is not relativistic by raising warnings if the velocity is greater than 10% of the speed of light.  These methods would be much more elegant as decorators so that they can be called like:
```Python
from astropy import units

@check_relativistic
@check_quantity
def ion_sound_speed(whatever: units.T, ... ):
    ...
    return c_s.to(units.m/units.s)
```
Your mission, if you choose to accept it, is to change `_check_quantity` and `_check_relativistic` to decorators.  After that, the next step would be to modify `parameters.py` to use the decorators instead of the original methods.  Thank you in advance!

GrahamGoudeau:
Closes #55 

This is a work-in-progress. Still to do:
* figure out how to handle the cases in `parameters.py` where computation happens before the original calls to `_check_quantity`
* once that is done, clean up the remaining export of `_check_quantity` in `utils/__init__.py` and nest that method inside of the `check_quantity` decorator to keep the namespace clean
* clean up documentation, including providing a documentation of the options provided in the validation dicts: `units`, `can_be_negative`, `can_be_complex`, `can_be_inf`, and `ignore`

Some features of the decorators:
* eliminates the need to repeat the function name- it's pulled from the Python function metadata
* `@check_quantity` works fine with default parameters, and also verifies that the given default is correct (if forced to fall back to the default)
* if a function is marked `@check_quantity`, its params must match the given validation dict exactly. If it does not, a `TypeError` is thrown. This includes if a parameter gets renamed in the method definition, but not in the validation dict, and vice versa. This will stop parameters from getting silently skipped from validation if someone carelessly renames a param.
* if a function is marked `@check_quantity`, then all of its parameters that are of type `units.Quantity` must be validated or explicitly ignored by providing a `{ 'ignore': True }` dict for that param

This also cleans up the tests nicely, since we don't have to provide a fake function name and arg name anymore.

### DIFF
--- a/plasmapy/utils/tests/test_checks.py
+++ b/plasmapy/utils/tests/test_checks.py
@@ -5,108 +5,124 @@ from astropy import units as u
 import pytest
 
 from ...constants import c
-from ..checks import (_check_quantity, _check_relativistic)
+from ..checks import (check_quantity, check_relativistic)
 
 
 def test__check_quantity():
+    def run_test(arg_value, units, can_be_negative=True, can_be_complex=False, can_be_inf=True):
+        """ Tests the check_quantity decorator
+        Tests that the given value is an astropy Quantity with correct units
+        and valid numerical values
+        Builds a method, decorates it, and immediately calls it
+        """
+        @check_quantity({
+            'arg': {'units': units, 'can_be_negative': can_be_negative, 'can_be_complex': can_be_complex, 'can_be_inf': can_be_inf}
+        })
+        def test_method(arg): pass
+        test_method(arg_value)
 
     # exceptions associated with the units keyword
+    with pytest.raises(TypeError):
+        run_test(5*u.T, 5*u.T)
 
     with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', 5*u.T)
+        run_test(5*u.T, 5)
 
     with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', 5)
+        run_test(5*u.T, [u.T, 1])
 
     with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', [u.T, 1])
+        run_test(5*u.T, [1, u.m])
 
     with pytest.raises(TypeError):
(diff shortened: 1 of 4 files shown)


StanczakDominik:
Holy wow, a race condition on pull requests was something I sure didn't expect. Thanks a lot for this! However, it looks like you're doing this simultaneously with 
@thomasjpfan over at #65 ! Could you two try to collaborate on this to avoid duplication of work? I'll try to help however I can, but I'm swamped with work today :(

plasmapy/utils/tests/test_checks.py
```diff
@@ -5,108 +5,124 @@
...
+    def run_test(arg_value, units, can_be_negative=True, can_be_complex=False, can_be_inf=True):
+        """ Tests the check_quantity decorator
+        Tests that the given value is an astropy Quantity with correct units
+        and valid numerical values
+        Builds a method, decorates it, and immediately calls it
+        """
+        @check_quantity({
+            'arg': {'units': units, 'can_be_negative': can_be_negative, 'can_be_complex': can_be_complex, 'can_be_inf': can_be_inf}
+        })
+        def test_method(arg): pass
```
SolarDrew: Why is `run_test` defining a wrapped function and then calling it? Could you not achieve the same thing by decorating `run_test` instead?

plasmapy/utils/tests/test_checks.py
```diff
@@ -5,108 +5,124 @@
...
+        test_method(arg_value)
 
     # exceptions associated with the units keyword
-
     with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', 5*u.T)
+        run_test(5*u.T, 5*u.T)
```
SolarDrew: I like the improvement in readability here :+1:

StanczakDominik:
Since #55 is closed, I'm going to go ahead and close this one as well.

PR #65, opened 2017-07-30
Check decorators

thomasjpfan:
Attempt to resolve issue: https://github.com/PlasmaPy/PlasmaPy/issues/55

Refactors `test_checks` to use `pytest.mark.parametrize` so the new decorators can use the same examples as `_check_quantity` and `_check_relativistic`.

Using `check_relativistic`:
```python
@check_relativistic
def speed():
    return 1*u.m/u.s
```
or with keywords:
```python
@check_relativistic(betafrac=0.01)
def speed():
    return 1*u.m/u.s
```

Using `check_quantity`:
```python
@check_quantity("x")
@check_quantity("y", can_be_negative=False)
def func(x: u.m, y: u.s=1*u.s):
    return x
func(1*u.m, 2*u.s)
```

I was able to convert most parameter functions. Parameter functions `electron_gyroradius`, `ion_gyroradius`, and `Alfven_speed` could not be converted because it does some preprocessing before checking the units.

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -1321,6 +1321,7 @@ def magnetic_energy_density(B: units.T):
 
     return E_B
 
+
 @check_quantity("B")
 @check_quantity("n_e", can_be_negative=False)
 def upper_hybrid_frequency(B: units.T, n_e: units.m**-3):
@@ -1384,7 +1385,7 @@ def upper_hybrid_frequency(B: units.T, n_e: units.m**-3):
 
 @check_quantity("B")
 @check_quantity("n_i", can_be_negative=False)
-def lower_hybrid_frequency(B: units.T, n_i:units.m**-3, ion='p'):
+def lower_hybrid_frequency(B: units.T, n_i: units.m**-3, ion='p'):
     r"""Returns the lower hybrid frequency.
 
     Parameters
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -13,7 +13,7 @@ import numpy as np
 # For future: change these into decorators.  _check_quantity does a
 # bit more than @quantity_input as it can allow
 
-from ..utils import _check_quantity, _check_relativistic
+from ..utils import _check_quantity, check_relativistic, check_quantity
 
 
 r"""Values should be returned as an Astropy Quantity in SI units.
@@ -72,6 +72,7 @@ by an angular frequency to get a length scale:
 """
 
 
+@check_relativistic
 def Alfven_speed(B, density, ion="p"):
     r"""Returns the Alfven speed.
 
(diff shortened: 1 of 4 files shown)


StanczakDominik:
Hey, thanks a lot for that! I'm kind of pressed for time, but if no one gets on it I'll look into this ASAP

plasmapy/physics/parameters.py
```diff
@@ -178,12 +179,13 @@ def Alfven_speed(B, density, ion="p"):
...
 
-    _check_relativistic(V_A, 'Alfven_speed')
-
     return V_A
 
 
-def ion_sound_speed(*ignore, T_e=0*units.K, T_i=0*units.K,
+@check_relativistic
+@check_quantity("T_e", can_be_negative=False)
+@check_quantity("T_i", can_be_negative=False)
```
SolarDrew: Is there a way this could be implemented so that you could specify both in one check_quantity call? I feel like that might be more readable than decorating the function three times.

plasmapy/utils/tests/test_checks.py
```diff
@@ -5,108 +5,294 @@
...
+    (5j*u.m, u.m, True, True, True)
+]
 
-    with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', 5)
 
-    with pytest.raises(TypeError):
-        _check_quantity(5*u.T, 'arg', 'funcname', [u.T, 1])
+# Tests for _check_quantity
+@pytest.mark.parametrize(
```
StanczakDominik: Agreed, `parametrize` is awesome!

plasmapy/physics/parameters.py
thomasjpfan: I think the best way to make it one call is to pass in a dictionary as shown in https://github.com/PlasmaPy/PlasmaPy/pull/64.

thomasjpfan: Case 1:
```python
@check_quantity({
    "T_e": { "units": units.K, "can_be_negative": False},
    "T_i": { "units": units.K, "can_be_negative": False},
    "B": {"units": units.T}
})
def func(T_e, T_i, B):
    pass
```
Case 2:
```python
@check_quantity({
    "T_e": { "can_be_negative": False},
    "T_i": { "can_be_negative": False}
})
def func(T_e:  units.K, T_i: units.K, B: units.T):
    pass
```
It all comes down to a style choice. Which one do you prefer?

GrahamGoudeau: In case 2, I'm not immediately sure when those types would be checked (at each runtime call? haven't used that feature before), but what would happen if someone leaves the type off? Does it miss validation in that case? I think I prefer case 1, so that there's a single point of truth for all the validation info.

thomasjpfan: If you prefer Case 1,
```python
@check_quantity({
    "T_e": { "units": units.K, "can_be_negative": False },
    "T_i": { "units": units.K, "can_be_negative": False },
    "B": {"units": units.T}
})
def func(T_e, T_i, B, n_i):
    pass
```
If `n_i` happens to be a Quantity, should that raise an error?

If you prefer Case 2,
```python
@check_quantity({
    "T_e": { "can_be_negative": False },
    "T_i": { "can_be_negative": False }
})
def func(T_e: units.K, T_i: units.K, B: units.T, n_i):
    pass
```
If `n_i` happens to be a Quantity, should that raise an error?

thomasjpfan:
The latest commit simplifies the logic when dealing default arguments by using `inspect.BoundArguments`. For now, `check_quantity` ignores arguments that are not included in the dictionary:
```python
@check_quantity({
    "B": {"units": unit.T}
})
def func(B, z):
    pass
```
will only check `B` and not `z`. `check_quantity` only type checks arguments explicitly.

SolarDrew:
Looks good to me. @StanczakDominik , @namurphy , any objections to merging this?

namurphy:
This looks good to me too.  Thank you!

PR #84, opened 2017-08-08
New coveralls directive

SolarDrew:
As per the discussion on #58 , here is a new coveralls directive that we can use to discount things in the tests.


StanczakDominik:
Works nicely. Dat 98.6% coverage :sunglasses:

PR #85, opened 2017-08-08
Small tweak for dropping astropy < 2.x

SolarDrew:
Spotted this while I was doing something else. Since we decided a few weeks ago only to support astropy >2.x, checking for version seems redundant to me.


StanczakDominik:
And redundant it is! Tests are passing and this is a small fix.

PR #94, opened 2017-08-25
Reorganize atomic physics functionality

namurphy:
This pull request moves the functions associated with atomic physics from the `constants` subpackage into a new `atomic` subpackage.  The purpose is that `atomic` is a more intuitive name for the location of these functions.

I might include some bug fixes here as well as described in https://github.com/PlasmaPy/PlasmaPy/issues/48.


plasmapy/physics/parameters.py
```diff
@@ -5,8 +5,8 @@
 from astropy.units import (UnitConversionError, UnitsError, quantity_input,
                            Quantity)
 
-from ..constants import (m_p, m_e, c, mu0, k_B, e, eps0, pi, ion_mass,
-                         charge_state)
+from ..constants import (m_p, m_e, c, mu0, k_B, e, eps0, pi)
+from ..atomic import (ion_mass, charge_state)
```
StanczakDominik: Come to think about it that's a funny import. Are the `( )` in thee to keep things working without `\` as it used to go multiline, like in line 5 up there? If so, that's cool, never thought about it!

namurphy: Yep, the [parentheses help with making multiline imports cleaner](https://www.python.org/dev/peps/pep-0328/).  Even on a single line, I sometimes use them since I personally find the code to be slightly more readable that way.

StanczakDominik: Cool, today I learned. Thanks!

PR #95, opened 2017-08-25
Stepper: Allowing unequal spatial dimensions grid

jhillairet:
Modified the interpolation routine to allow unequal spatial dimension grid (nx,ny,nz) with nx!=ny!=nz. 
Also added a test case.

### DIFF
--- a/plasmapy/classes/species.py
+++ b/plasmapy/classes/species.py
@@ -85,9 +85,9 @@ class Species:
                                          dtype=float) * (u.m / u.s)
         # create intermediate array of dimension (nx,ny,nz,3) in order to allow
         # interpolation on non-equal spatial domain dimensions
-        _B = np.moveaxis(self.plasma.magnetic_field.si.value,0,-1)
-        _E = np.moveaxis(self.plasma.electric_field.si.value,0,-1)
-        
+        _B = np.moveaxis(self.plasma.magnetic_field.si.value, 0, -1)
+        _E = np.moveaxis(self.plasma.electric_field.si.value, 0, -1)
+
         self._B_interpolator = RegularGridInterpolator(
             (self.plasma.x.si.value,
              self.plasma.y.si.value,
@@ -200,7 +200,7 @@ class Species:
         return f"Species(q={self.q:.4e},m={self.m:.4e},N={self.N}," \
                f"name=\"{self.name}\",NT={self.NT})"
 
-    def __str__(self): # coveralls: ignore
+    def __str__(self):  # coveralls: ignore
         return f"{self.N} {self.scaling:.2e}-{self.name} with " \
                f"q = {self.q:.2e}, m = {self.m:.2e}, " \
                f"{self.saved_iterations} saved history " \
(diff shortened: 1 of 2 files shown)


StanczakDominik:
Wait, THAT didn't work? Huh. I can't check right now, but I'll get to that later. Well spotted!

plasmapy/classes/tests/test_Species.py
```diff
@@ -189,3 +189,15 @@ def test_particle_exb_nonuniform_drift():
...
+    '''
+        Test the particle stepper when the spatial domain dimensions are not equal
+    '''   
+    x = np.linspace(0,1,10)*u.m
+    y = np.linspace(0,1,20)*u.m
+    z = np.linspace(0,1,30)*u.m
+    
+    plasma = Plasma(x,y,z)
+    
+    Species(plasma, 'e', dt=1e-14*u.s, nt=1e3)
```
StanczakDominik: I'm not sure this will actually check the interpolator without using the `run` method. How about something like this?

    Species(..., nt = 5).run()

That way we don't waste much time on this test - if it's going to glitch out, 5 iterations are plenty enough of time.

plasmapy/classes/species.py
```diff
@@ -237,7 +236,6 @@ def plot_time_trajectories(self, plot="xyz"):  # coveralls: ignore
         """
         from astropy.visualization import quantity_support
         import matplotlib.pyplot as plt
-        from mpl_toolkits.mplot3d import Axes3D
```
StanczakDominik: An oversight on my part that I didn't automate testing visualization, my bad. But this will break the plots unless we keep the `import` or move it to the beginning of the file - I wasn't sure we should be doing that as it'll be only necessary for visualization, but that's probably way too premature optimization.

Pep8 won't catch that breaking either, matplotlib is weird like that on the 3d plotting end.

plasmapy/classes/species.py
```diff
@@ -83,23 +83,29 @@ def __init__(self, plasma, particle_type='p', n_particles=1, scaling=1,
...
             (self.plasma.x.si.value,
              self.plasma.y.si.value,
              self.plasma.z.si.value),
-            self.plasma.electric_field.T.si.value,
+            _E,
             method="linear",
             bounds_error=True)
 
     def _interpolate_fields(self):
+        print(self.x.si.value)
```
SolarDrew: Won't this print for every iteration? Possibly not what the user wants, particularly for large grids.

StanczakDominik: Oh yeah, that it will. Missed this one.

jhillairet: Yes, it was for debug purpose. I'll remove it.

plasmapy/classes/species.py
jhillairet: OK, I'll re-add it.

plasmapy/classes/tests/test_Species.py
jhillairet: The Species() constructor does setup the interpolator and the bug was here. I can add a run() method call, does not cost a lot. Btw is also tested in previous existing tests.

PR #97, opened 2017-09-05
Revise vision statement and create roadmap document

namurphy:
This pull request contains updates to the vision statement after a few months into the project.  

The main change is to absorb the Code Development Committee into the Coordinating Committee.    The responsibilities for both of these committees are blurred, and this pull request combines the two committees into a single committee.  For the first few months of active code development, the organization has been very informal.  This pull request is in advance of solidifying committee membership.

Are there any other changes that we should make while we're revisiting the vision statement?  I'll read through this page again in case there are any other places we'd like to update this.

### DIFF
--- a/vision_statement.md
+++ b/vision_statement.md
@@ -114,8 +114,8 @@ conduct](https://www.python.org/psf/codeofconduct/).
 
 The Coordinating Committee (CC) will oversee the PlasmaPy project and
 code development.  The CC will ensure that roles are being filled,
-facilitate intergroup communication, coordinate and delegate tasks,
-manage the project repository, oversee the code review process,
+facilitate community-wide communication, coordinate and delegate
+tasks, manage the project repository, oversee the code review process,
 regulate intercompatibility between different subpackages, seek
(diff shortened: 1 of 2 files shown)


vision_statement.md
```diff
@@ -112,22 +112,13 @@ conduct](https://www.python.org/psf/codeofconduct/).
...
-The Code Development Committee (CDC) will directly oversee the
-development of the codebase.  This committee will be responsible for
-managing the project repository, reviewing and accepting pull requests,
-ensuring intercompatibility between different subpackages, and
-coordinating automated testing.  The CDC will be responsible for
-ensuring that the PlasmaPy codebase adheres to the standards defined
-in this document.
+The Coordinating Committee (CC) will oversee the PlasmaPy project and
+code development.  The CC will ensure that roles are being filled,
+facilitate intergroup communication, coordinate and delegate tasks,
```
SolarDrew: What groups are intercommunicating if we're cutting down to one committee?

namurphy: Good catch - I changed it to "facilitate community-wide communication".

PR #98, opened 2017-09-07
Include versioning in vision statement and add change log/release notes

namurphy:
I added new files for the change log and release notes for our 0.1 release.  The content of these files is temporary and will need to be revised as we approach the release date.  I am not sure if it is best to store the change log and release notes here or elsewhere (such as in the docs subdirectory).

I edited the vision statement and change log to say that PlasmaPy uses [Semantic Versioning](http://semver.org/), which a few of us had informally decided upon in the past but I don't think we ever officially decided it.  Is it okay if we officially say that we're using Semantic Versioning?


SolarDrew:
👍

PR #117, opened 2017-09-21
Implement astropy.units.quantity_input to physics

StanczakDominik:
This is heavily work in progress, tests will fail, coverage will decrease and the PEP8 bot shall grit teeth. But we shall endure, and we shall get there, and this code shall work without exception.

Eventually.

Some of the stuff I've done alongside introducing quantity_input:
* Separate number density and mass density as keywords in Alfven speed to bypass issues with the previous approach not playing well with 
* Add check_positive decorator (work in progress as I figure out how to do this)
* Switch UserWarning from exception to warning

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -13,7 +13,7 @@ import numpy as np
 # For future: change these into decorators.  _check_quantity does a
 # bit more than @quantity_input as it can allow
 
-from ..utils import _check_quantity, check_relativistic, check_quantity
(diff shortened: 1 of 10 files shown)


StanczakDominik:
Kinda stuck on unit conversions between rad/s and 1/s. Still got some energy (eV to K) conversion issues I thought I figured out, but I guess not everywhere.

StanczakDominik:
This PR has gone wrong enough that I'll be closing this and possibly rewriting from scratch in the future (likely after 0.1). I might reopen this then.

Issue #118, opened 2017-09-22
Consider splitting up tests into multiple ones per function


StanczakDominik:
Let's say I'm debugging `physics.Alfven_speed` after making some changes. I run pytest in my trusty old PyCharm and I run into an error on line 69. I fix that one and run pytest again. Now there's an error on line 79, which the previous test run didn't reach as `assert` failure means executing the code within the test function gets stopped. I fix that one and proceed to repeat until there are no tests failing in `test_parameters`.

That's a lot of test runs, and as our tests increase in number and size, it might not be a scalable solution.

Another approach would be to split the tests into multiple test functions (I'm not 100% sure what's the best approach here, maybe you group those into classes or some such thing). That way, we get instant visual feedback on which tests are working, which are failing and where we need to aim our fixes before running the test suite again. The testing process becomes much more granular\incremental that way.

I believe it would also be handy for automated Travis CI tests, decreasing the number of commit pushes necessary.

There is also a little thing called [pytest fixture](https://docs.pytest.org/en/latest/fixture.html) that'd I'd like to implement later wherever it ends up being applicable. Most of our tests check qualitatively different behaviour, but that may also be useful.

One possible disadvantage of the proposed approach is that the size of our test suite will grow by a bunch, but I think it's only a visual increase - we'd essentially be doing the same number of tests, just with separate labels and in parallel.

Here's a reference to Stack: https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test

thomasjpfan:
`@pytest.mark.parametrize` can be used as well. For example, this snippet:
```python
def test_Alfven_speed():
    with pytest.raises(ValueError):
        Alfven_speed(B_nanarr, rho_arr)

    with pytest.raises(ValueError):
        Alfven_speed(B_arr, rho_negarr)

    with pytest.raises(u.UnitConversionError):
        Alfven_speed(5*u.A, n_i, ion='p')
```
can be refactored to:
```python
# B, density, kwargs, error
alfven_speed_error_test_cases = [
   (B_nanarr, rho_arr, {}, ValueError),
   (B_arr, rho_negarr, {}, ValueError),
   (5*u.A, n_i, {'ion':'p'}, u.UnitConversionError)
]

@pytest.mark.parametrize(
   "B,density,kwargs,error", alfven_speed_error_test_cases)
def test_Alfven_speed_errors(B, density, kwargs, error):
[...]

PR #119, opened 2017-09-23
Split tests in atomic

thomasjpfan:
Works on issue #118 for `atomic` module. There are some style changes, for example: `9.11e-31*u.kg` to `9.11e-31 * u.kg`.

### DIFF
--- a/plasmapy/atomic/tests/test_atomic.py
+++ b/plasmapy/atomic/tests/test_atomic.py
@@ -481,7 +481,7 @@ nuclear_binding_energy_table = [
 
 @pytest.mark.parametrize("argument, expected_error",
                          nuclear_binding_energy_table)
-def test_nuclear_binding_energy(argument, expected_error):
+def test_nuclear_binding_energy_error(argument, expected_error):
     with pytest.raises(expected_error):
         nuclear_binding_energy(argument)
 
--- a/plasmapy/atomic/tests/test_atomic.py
+++ b/plasmapy/atomic/tests/test_atomic.py
@@ -83,137 +83,118 @@ def test_element_symbol_error(argument, expected_error):
         element_symbol(argument)
 
 
-def test_isotope_symbol():
-    assert isotope_symbol('He', 4) == 'He-4'
-    assert isotope_symbol('helium-4') == 'He-4'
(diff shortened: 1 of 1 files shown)


plasmapy/atomic/tests/test_atomic.py
```diff
@@ -10,398 +11,416 @@
...
+# (argument, expected_error)
+mass_number_error_table = [
+    ('H-359', ValueError),
+    ('C-12b', ValueError),
+    (-1.5, Exception),
+    ('N-13+-+-', ValueError)]
+
+
+@pytest.mark.parametrize(
+    "argument, expected_error", mass_number_error_table)
```
StanczakDominik: Oh the things we do for love and PEP8 compliance. :smile:

plasmapy/atomic/tests/test_atomic.py
```diff
@@ -660,60 +714,65 @@ def test_isotopic_abundance():
...
-        isotopes = common_isotopes(element)
+isotopic_abundance_elements = (
+    atomic_number(atomic_numb) for atomic_numb in range(1, 119))
+isotopic_abundance_isotopes = (
+    common_isotopes(element) for element in isotopic_abundance_elements)
+isotopic_abundance_sum_table = [
+    (element, isotopes)
+    for element, isotopes in
+    zip(isotopic_abundance_elements, isotopic_abundance_isotopes)
+    if isotopes]
```
StanczakDominik: I'm curious whether this wouldn't run faster with a generator instead of a list, given that at no point here do we use more than a single element from this list, so there's no need to keep the whole list in memory. I wasn't able to show an explicit variation in speed over noise this way, though. The code is quite fine, I'm just thinking aloud (well, the keyboard is loud).

thomasjpfan: Good suggestion!

StanczakDominik: Haha, I wasn't really suggesting changing this one, but yeah, we might as well go with generators here :laughing:

namurphy:
This is great!  Thank you, @thomasjpfan!  I'm going to try to write tests using these strategies in the future.

PR #125, opened 2017-09-28
Old version errors

### LINKED ISSUE #112: Raise ImportError when installing/importing with old versions of Python, Astropy, & NumPy
We decided that PlasmaPy will only be supported for:
- Python version > 3.6
- Astropy version > 2.0
- NumPy version > 1.13

However, when I try to run:
```ShellSession
python setup.py install
```
from the command line with Python 3.5 then I get a `SyntaxError` for syntax that is new in version 3.6.

When I try to run
```Python
import plasmapy
```
in Python 3.6 with Astropy 1.3.1, then I get an exception since one of the constants imported from Astropy was renamed in 2.0.

We should raise an appropriate exception (probably an `ImportError`) when we try to install or import PlasmaPy with any of the unsupported versions above.  We should also have appropriate and useful error messages for any of these situations.  The pull request to close this issue would involve editing `setup.py`, `requirements/base.txt`, and `plasmapy/__init__.py`.

Thank you!
Nick

colbych:
Closes #112 

I tested this but if anyone else wants to double check, it is always appreciated.


plasmapy/__init__.py
```diff
@@ -5,16 +5,79 @@
...
+if sys.version_info < _split_version(__minimum_python_version__):
+    warnings.warn("PlasmaPy does not support Python 3.5 and below")
+
+_check_numpy_version()
+_check_astropy_version()
+
 from .classes import Plasma
 from . import classes
 from . import constants
 from . import atomic
```
namurphy: I'm inclined to leave these import statements here, even though PEP 8 doesn't like them.  I think it makes sense to have these after all of the checking for version numbers and all that.  For this section of code, I'm wondering if we could do something like this:
```Python
try:
    from .classes import Plasma
    from . import classes
    from . import constants
    from . import atomic
    from . import math
    from . import physics
    from . import utils
except Exception:
    raise ImportError("Unable to load PlasmaPy subpackages.")
```

plasmapy/__init__.py
```diff
@@ -5,16 +5,78 @@
...
+
+def _split_version(version):
+    return tuple(int(ver) for ver in version.split('.'))
+
+
+def _min_required_version(required, current): # coveralls: ignore
+    """ Return `True` if the current version meets the required minimum
+        version and `False` if not/ if not installed.
+
+        Right now `required` and `current` are just . seperated strings
```
namurphy: Minor edit: `seperated` should be changed to `separated`

plasmapy/__init__.py
colbych: I'll give it a go. Thanks Nick

plasmapy/__init__.py
```diff
@@ -5,16 +5,78 @@
...
+        but It would be good to make this more general and accept modules.
```
namurphy: Minor edit: uncapitalize `It`

plasmapy/__init__.py
```diff
@@ -5,16 +5,78 @@
...
+
+    if not required_version:
+        ver_error = ("Numpy {} or above is required for PlasmaPy. The "
+                     "currently installed version is {}"
+                     ).format(__minimum_numpy_version__, np_ver)
+        raise ImportError(ver_error)
+
+
+def _check_astropy_version(): # coveralls: ignore
+    """ Make sure astropy in installed and meets the minimum version requiremnets
```
namurphy: Minor edit: `requiremnets` to `requirements`

colbych: 😳  Sorry! I have a real problem with spelling. I'll add a spell checker to vim.

plasmapy/__init__.py
```diff
@@ -5,16 +5,78 @@
     author as __author__,
 )
 
+import sys
+import warnings
+
+__minimum_python_version__ = '3.6'
+__minimum_numpy_version__ = '1.10.0'
```
namurphy: Minor edit: I think we want `'1.10.0'` to `'1.13.0'`

namurphy:
@colbych - Thank you for working on this!  It's looking good and I think it's almost ready to merge.

PR #139, opened 2017-10-03
Removed auto-referencement in importation

antoinelpp:
I add some issue during the developpement of some feature.
I think that this is better.


namurphy:
This looks good to me, and is a straightforward change so I'll merge it.  Thank you!

Issue #100, opened 2017-09-08
Functions to get magnetostatic fields from known solutions (infinite wires, circular loops)


namurphy:
In addition to the Biot-Savart law solver proposed in #99, we should have a way of calculating the magnetic fields resulting from simple current-carrying wire setups.  This would include the magnetic field from an infinite wire (which has a simple exact solution) and the magnetic field resulting from a circular current-carrying loop (which has a complicated exact solution) for boundary conditions at infinity.  There might be other analytical solutions worth including, but these two are probably the most important ones.

This could perhaps go in the `physics` subpackage in a file called `magnetostatics.py`.

namurphy:
Happy issue/pull request #100, PlasmaPy!

golanor:
Hacktoberfest TLV
Shotgun.
Now that the mandatory announcement is over - I'll start by adding the infinite wire case.

SaravananOffl:
I can work on this!
@namurphy you can assign it to me

hzxusx:
@SaravananOffl @namurphy 
Is any one still working on this feature? I have writen a [Julia script](https://github.com/hzxusx/active-mag-design/blob/master/wires.jl) for this last semester as part of my class project. I can adapt it to python.

I implemented circular wire, (finite) straight wire, general wire(described by parametric equation) types and methods to calculate their magnetic field at any position in 3D Cartesian coordinate. Those non-analytical integrals are done by Gauss-Legendre integration.

StanczakDominik:
As far as I can tell, this is completely stalled. Go for it, @hzxusx!

hzxusx:
Great! I'm happy to contribute. 

A question on your development habit, shall I open a feature branch? You seems use little of branches.

StanczakDominik:
We use a bunch of them, but mostly on our personal forks. Take a look at the code development guide over at docs.plasmapy.org, and if that doesn't clear things up, please say so - I'll be happy to improve those docs!

PR #147, opened 2017-10-19
working on issue #100

golanor:
added a function to calculate the magnetic field created by an infinite, current-carrying wire, to the file magnetostatics.py


plasmapy/physics/magnetostatics.py
```diff
@@ -0,0 +1,62 @@
...
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> distance = 1.5*u.m
+    >>> current = 2.45*u.ampere
+    >>> infinite_wire(current, distance)
+    <Quantity 3.2666666666666673e-07 T>
+    """
+    if r > 0:
+        B = np.divide(mu0 * I,2*pi*r)
```
StanczakDominik: Why not straight up `(mu0 * I) / (2 * pi * r)`? Is there any difference in speed between the two?

plasmapy/physics/magnetostatics.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> distance = 1.5*u.m
+    >>> current = 2.45*u.ampere
+    >>> infinite_wire(current, distance)
+    <Quantity 3.2666666666666673e-07 T>
+    """
+    if r > 0:
```
StanczakDominik: If doing this in 3D on Numpy arrays, this part will crash.

plasmapy/physics/magnetostatics.py
golanor: Shouldn't be, just a force of habit to avoid integer divides in python 2.

plasmapy/physics/magnetostatics.py
golanor: Of course. Hadn't really thought about it being 3D. Hopefully I'll have time to correct this later this week.

plasmapy/physics/magnetostatics.py
SolarDrew: I think `np.divide` is slightly faster if you do it in-place rather than returning a new variable. But it's basically the same unless you're iterating over the calculation a lot.

StanczakDominik:
We're probably going to merge the more complete #306 in favor of this one :( going to close this one for now.

PR #148, opened 2017-10-19
Fix #146 - Convert doc string to raw string

### LINKED ISSUE #146: Turn docstrings into raw strings
As seen in #145, it turns out you need to use raw strings (`"this is not a raw string", r"but this is because of the prepended r"`) for docstrings if you want to have LaTeX formulas rendering properly via Sphinx. We could probably use some help with this one - basically, this means going through all the files with some clever regex substitution or something.

raajitr:
Fixes #146,

Hi I've made a python script for this which I can share with you on gist. Its not a super efficient script, as I haven't used RegEx as I am not that proficient in it.

Please let me know if any changes are required.

Thanks

### DIFF
--- a/plasmapy/__init__.py
+++ b/plasmapy/__init__.py
@@ -18,7 +18,7 @@ def _split_version(version):
 
 
 def _min_required_version(required, current):  # coveralls: ignore
-    """ Return `True` if the current version meets the required minimum
+    r""" Return `True` if the current version meets the required minimum
         version and `False` if not/ if not installed.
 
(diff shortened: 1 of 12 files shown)


raajitr:
[Here's](https://gist.github.com/raajitr/7e52b1e4d6d2597637b7147b2cff3658) the script for that.
Feel free to comment on this for suggestions and improvement for me.

Thanks,

StanczakDominik:
Nicely done! Thanks for sharing the script, it's neatly written :)

Merging this one!

PR #149, opened 2017-10-29
Added functions to quantum.py

lemmatum:
Added functions for thermal deBroglie length, Thomas-Fermi screening length, and Fermi energy. Also added corresponding tests and imports in __init__.py. Added check to quantum.py and test_quantum.py for if module is called __main__, which then swaps from relative to absolute path imports. This allows for individual execution/testing of these modules.

### DIFF
--- a/plasmapy/physics/quantum.py
+++ b/plasmapy/physics/quantum.py
@@ -2,20 +2,12 @@
 Functions for quantum parameters, including electron degenerate
 gases and warm dense matter.
 """
-if __name__ == "__main__":    
-    import numpy as np
-    from astropy import units
-    from plasmapy.constants import c, h, hbar, m_e, eps0, e, k_B
-    from plasmapy.atomic import ion_mass
-    from plasmapy.utils import _check_quantity, _check_relativistic, check_quantity
-    from plasmapy.physics.relativity import Lorentz_factor
-else:
-    import numpy as np
-    from astropy import units
-    from ..constants import c, h, hbar, m_e, eps0, e, k_B
-    from ..atomic import ion_mass
-    from ..utils import _check_quantity, _check_relativistic, check_quantity
-    from .relativity import Lorentz_factor
+import numpy as np
+from astropy import units
+from ..constants import c, h, hbar, m_e, eps0, e, k_B
+from ..atomic import ion_mass
+from ..utils import _check_quantity, _check_relativistic, check_quantity
+from .relativity import Lorentz_factor
 
 
 def deBroglie_wavelength(V, particle):
--- a/plasmapy/physics/quantum.py
+++ b/plasmapy/physics/quantum.py
@@ -1,9 +1,21 @@
-import numpy as np
-from astropy import units
-from ..constants import c, h
-from ..atomic import ion_mass
-from ..utils import _check_quantity, _check_relativistic
-from .relativity import Lorentz_factor
+"""
+Functions for quantum parameters, including electron degenerate
+gases and warm dense matter.
+"""
+if __name__ == "__main__":    
+    import numpy as np
+    from astropy import units
+    from plasmapy.constants import c, h, hbar, m_e, eps0, e, k_B
+    from plasmapy.atomic import ion_mass
+    from plasmapy.utils import _check_quantity, _check_relativistic, check_quantity
+    from plasmapy.physics.relativity import Lorentz_factor
+else:
+    import numpy as np
+    from astropy import units
+    from ..constants import c, h, hbar, m_e, eps0, e, k_B
+    from ..atomic import ion_mass
+    from ..utils import _check_quantity, _check_relativistic, check_quantity
+    from .relativity import Lorentz_factor
 
 
 def deBroglie_wavelength(V, particle):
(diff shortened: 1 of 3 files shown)


plasmapy/physics/tests/test_quantum.py
```diff
@@ -1,9 +1,21 @@
-import numpy as np
-import pytest
-import astropy.units as u
-from ...constants import c, h
-from ..quantum import deBroglie_wavelength
-
+if __name__ == "__main__":
```
StanczakDominik: As before, this looks weird-ish to me. It seems to me like with properly written and properly used code you shouldn't need to check for this... I could really use a second opinion on this. @PlasmaPy/plasmapy-community?

plasmapy/physics/tests/test_quantum.py
```diff
@@ -48,3 +60,37 @@ def test_deBroglie_wavelength():
...
+n_e = 1e23 * u.cm**-3
+# should probably change this to use unittest module
+# add tests for numpy arrays as inputs
+# add tests for different astropy units (random fuzzing method?)
+def test_thermal_deBroglie_wavelength():
+    r"""Test the thermal_deBroglie_wavelength function in quantum.py."""
+    lambda_dbTh = thermal_deBroglie_wavelength(T_e)
+    # test a simple case for expected value
+    assert np.isclose(lambda_dbTh.value,
+                      6.919367518364532e-10)
```
StanczakDominik: Related to #150: doing it like this:

    assert np.isclose(lambda_dbTh.value, 6.919367518364532e-10), "Thermal deBroglie wavelength is off expected result for 1eV"

... or with some other short explanation could be much more readable. The string would be thrown on assertion failure in PyTest.

plasmapy/physics/tests/test_quantum.py
```diff
@@ -48,3 +60,37 @@ def test_deBroglie_wavelength():
...
+    # testing returned units
+    assert lambda_dbTh.unit == u.m
+
+def test_Fermi_energy():
+    r"""Test the Fermi_energy function in quantum.py."""
+    energy_F = Fermi_energy(n_e)
+    # test a simple case for expected value
+    assert np.isclose(energy_F.value,
+                      1.2586761116196002e-18)
```
StanczakDominik: As before, this could use an explanation!

plasmapy/physics/tests/test_quantum.py
```diff
@@ -48,3 +60,37 @@ def test_deBroglie_wavelength():
...
+    # testing returned units
+    assert energy_F.unit == u.J
+
+def test_Thomas_Fermi_length():
+    r"""Test the Thomas_Fermi_length function in quantum.py."""
+    lambda_TF = Thomas_Fermi_length(n_e)
+    # test a simple case for expected value
+    assert np.isclose(lambda_TF.value,
+                      5.379914085596706e-11)
```
StanczakDominik: And likewise :) Basically, a "this is the expected value because..." either in the assert statement or in the docstring would be perfect.

plasmapy/physics/quantum.py
```diff
@@ -1,9 +1,21 @@
...
-from astropy import units
-from ..constants import c, h
-from ..atomic import ion_mass
-from ..utils import _check_quantity, _check_relativistic
-from .relativity import Lorentz_factor
+"""
+Functions for quantum parameters, including electron degenerate
+gases and warm dense matter.
+"""
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
lemmatum: It's not about importing, but rather about running directly. Can you run `python test_quantum.py` directly from the command line when this split is not there? This probably isn't necessary for all modules, but I think it could be valuable for running tests individually.

plasmapy/physics/tests/test_quantum.py
lemmatum: This [link](https://stackoverflow.com/questions/419163/what-does-if-name-main-do#419185) may explain it better.

plasmapy/physics/quantum.py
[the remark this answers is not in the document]
lemmatum: This is just my ignorance of how PyTest works. I'll remove this split and push an update.

plasmapy/physics/tests/test_quantum.py
```diff
@@ -71,26 +61,29 @@ def test_thermal_deBroglie_wavelength():
     r"""Test the thermal_deBroglie_wavelength function in quantum.py."""
     lambda_dbTh = thermal_deBroglie_wavelength(T_e)
     # test a simple case for expected value
+    expectStr = "Thermal deBroglie wavelength is off expected result for 1eV"
```
StanczakDominik: We could use the python 3.6 technique I mentioned in #152 here! Then again, pytest should show the value of `lambda_dbTh` within `np.isclose` so I guess that's not absolutely necessary.

plasmapy/physics/tests/test_quantum.py
```diff
@@ -71,26 +61,29 @@ def test_thermal_deBroglie_wavelength():
...
-                      6.919367518364532e-10)
+                      6.919367518364532e-10), expectStr
     # testing returned units
     assert lambda_dbTh.unit == u.m
 
 def test_Fermi_energy():
     r"""Test the Fermi_energy function in quantum.py."""
     energy_F = Fermi_energy(n_e)
     # test a simple case for expected value
+    expectStr = "Fermi energy is off expected result for 1e23 cm^-3"
```
StanczakDominik: Actually, is this an `expected value string` or an `exception string`?

plasmapy/physics/tests/test_quantum.py
StanczakDominik: Sure, I'm aware of the practice of doing that! However, what I'm trying to figure out is what's the optimal way of achieving the ability to execute a script visibly testing some library functions while debugging it, while also not having to deal with this sort of visibility mess.

Here's a few cases of this that I've found:
* https://github.com/matplotlib/matplotlib/blob/e6ef481a32ed3ec982ce4dabdc627334fb1db2e4/lib/mpl_toolkits/axisartist/clip_path.py#L140
* https://github.com/matplotlib/matplotlib/blob/7131876063f933bc5e5c5d768722dd81bb69038e/lib/mpl_toolkits/axisartist/angle_helper.py#L439
* https://github.com/numpy/numpy/blob/c90d7c94fd2077d0beca48fa89a423da2b0bb663/numpy/lib/tests/test_arrayterator.py#L50

So while I completely agree it's nice to be able to run `python whatevermodulewereworkingon.py` and see how it's coming along while development, we should probably not need to do this kind of visibility double-checking to achieve that.

plasmapy/physics/tests/test_quantum.py
lemmatum: Well it's telling you that the expected value is off, but it's thrown as an exception, so both? 😖

StanczakDominik: Hah :smile:

plasmapy/physics/quantum.py
```diff
@@ -99,3 +103,187 @@ def deBroglie_wavelength(V, particle):
...
+        lambda_dbTh = h / np.sqrt(2 * np.pi * m_e * k_B * T_e)
+    except Exception:
+        raise ValueError("Unable to find thermal deBroglie wavelength.")
+    return lambda_dbTh.to(units.m)
+
+@check_quantity({
+    'n_e': {'units': units.m**-3, 'can_be_negative': False}
+})
+def Fermi_energy(n_e):
+    r"""Calculate the Thomas-Fermi screening length.
```
namurphy: Should this be `"Calculate the Fermi energy"` instead?

lemmatum: good catch!

lemmatum:
Any idea why travis-ci doesn't like things like:
```
except Exception: 
    raise ValueError("Blah")
```
?

namurphy:
> Any idea why coveralls doesn't like things like:
> ```
> except Exception: 
>    raise ValueError("Blah")
> ```
> ?

Unless exceptions are tested explicitly, tests will not cover exceptions, so lines like this will be flagged by coveralls as not being tested.  What I usually do is:

```Python
import pytest

def test_something():
    with pytest.raises(ValueError):
        a_function(input_that_should_raise_ValueError)
```
so that I can test to make sure that exceptions are being raised when they should be.  I tend to be very meticulous about testing exceptions, but realistically only relatively important exceptions should have tests.  You can also add `# coveralls: ignore` to the end of a line to indicate that a line should not be covered in coverage reports, though I try to use this sparingly.

Thank you for working on this!

lemmatum:
Thanks for the tips! I opted to remove the try-except blocks in the functions I wrote since they weren't actually checking anything important. I think the things that would make the functions fail are already caught by the unit checks and preventing negative values as inputs.

StanczakDominik:
And that's that, thanks a bunch for this one!

lemmatum:
Woohoo! My first approved pull request! 🎆 
Any tips on branch deletion and synchronizing my repos with the updated one?

StanczakDominik:
What I tend to do is to remove the feature branch and pull from the
upstream master branch into my own master branch ( git pull upstream/master
should do the trick).

Congratulations! :)

I need to figure out if GitHub has a way to rebase without changing commit
hashes, which I think it does...

jasperbeckers:
FYI; from now on we should either indent the equation line in the docstring after :: math:: for display mode or use the inline version ::math::`x^2`. Otherwise it will not display correctly on our Read the Docs.

lemmatum:
[This](https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository) seems to have worked for me. It's basically `fetch upstream`, `checkout master`, `rebase upstream/master` and then `push -f origin master`. Don't know if it has changed the commit hashes or not.

PR #141, opened 2017-10-04
Distribution function

antoinelpp:
First contribution for me !
To start with #101, I added a 1D Maxwellian distribution function calculation.
Please tell me if you like the interface, numbre of arguments and so on.

Needs to be done :
- [ ] Do the same with Kappa
- [ ] implement a 2D and/or 3D version of it.
- [ ] add particle generator
- [ ] maybe : add symbolic formulation

Issue #101, opened 2017-09-08
Functions for Maxwellian and kappa distribution functions


namurphy:
Particle distribution functions are used pretty commonly in plasma physics, and there are analytical expressions for [Maxwell-Boltzmann](https://en.wikipedia.org/wiki/Maxwell-Boltzmann_distribution) and [kappa](https://arxiv.org/pdf/1003.3532.pdf) distributions.  It would be great to create functions for these distributions in the `physics` subpackage.

antoinelpp:
Hi, I'd like to take care of this one.
I'll try to implement :
* functions with arg {T, V_dift, v} which return f(v) (add $\kappa$ for the Kappa function)
* A generator, creating a particle which velocity corespond to the discribution function

I'm not sure if I can do it all by myself... We'll see ! :)

StanczakDominik:
Awesome! For the particles, could you also try to tie it to that PIC
`species` class and try to make it return velocities for multiple
particles? I'll try to help out as I can :)

12.09.2017 09:51 "antoinelpp" <notifications@github.com> napisał(a):

lemmatum:
Anyone want to add Kappa distributions so we can close this one out? Or got any tips about Kappa distributions and what they're supposed to do? (I've never used them).

lemmatum:
I sorted out the kappa velocity distribution functions (both 1D and 3D). This issue is nearly done.

lemmatum:
I believe #141 and #171 close this issue. Subsequent testing and verification for the kappa distribution should fall under #186

Issue #131, opened 2017-10-02
Add eV_temp unit in astropy.units


namurphy:
A common convention in plasma physics is to use the electron-volt (eV) as a unit of temperature.  Strictly speaking, the eV is not a unit of temperature but is rather a measure of the typical thermal energy per particle (using the relation E = k_B * T).  The duality of eV as a unit of temperature and energy complicates the usage of units.

At present, several functions in `parameters.py` permit eV as a unit of temperature.  The conversion from eV to K is done inside of the function using an equivalency as follows:
```Python
>>> from astropy import units as u
>>> (1*u.eV).to(u.K, equivalencies=u.temperature_energy())
<Quantity 11604.522060401008 K>
```
A problem with this approach is that this non-standard unit conversion is done silently and implicitly, and I am having doubts about this design decision.   

Some considerations are:
* Within PlasmaPy, we are maintaining a strict adherence to SI units as the recognized international standard.  
* Departures from SI units can lead to confusion among newcomers to plasma physics who have almost always learned to use the kelvin as the unit of temperature. 
* "Explicit is better than implicit."  
* "There should be one-- and preferably only one --obvious way to do it."  (in this case, one way to represent temperature)
* Plasma physicists widely use eV as a unit of temperature, and will probably want to continue to be able to do so.
* Expressions such as T_e = 1 keV leave out a silent factor of k_B.
* The eV is also used as [a measure of momentum](https://en.wikipedia.org/wiki/Electronvolt#Momentum), often with a silent factor of c.
* Confusions may arise in the code about whether a variable named T has units of temperature or energy.

Our options include:
1. Allow functions and methods to accept temperatures in units of eV.  This represents the status quo.
2. Allow eV to be inputted as temperature, but raise a warning message that a non-standard unit conversion is being made.
3. Do not allow temperature variables to accept eV, but rather enforce that temperature inputs require units of temperature.  Two possibilities to take this route would be
  - Create a shortcut function (perhaps called `kelvinify`) to explicitly do the temperature conversion.
  - Have users manually convert to Kelvin in the arguments to functions:
```Python
>>> from astropy.units import keV
>>> from plasmapy.constants import k_B
>>> thermal_velocity(T=1.4*keV/k_B, particle='p')
```

The more I think about it, the more I am in favor of option 3.  What does everyone else think?

StanczakDominik:
[...]

PR #152, opened 2017-10-30
Added/modified 1D and 3D Maxwellian speed/velocity distributions

lemmatum:
Added and modified 1D and 3D Maxwellian speed and velocity distribution functions. These functions also include optional drift velocities/speeds. I've included tests for these new functions, but the 3D tests fail because scipy's QUADPACK based triple integral doesn't play nice with units. I'm open to suggestions on how to work around this!

Contributions towards #101  and #141

### DIFF
--- a/plasmapy/physics/distribution.py
+++ b/plasmapy/physics/distribution.py
@@ -14,11 +14,12 @@ import numpy as np
 from ..utils import _check_quantity, check_relativistic, check_quantity
 
 
-@u.quantity_input
-def Maxwellian_1D(v: u.m/u.s,
+def Maxwellian_1D(v,
                   T, 
                   particle="e",
-                  V_drift=0*u.m/u.s):
+                  V_drift=0,
+                  vTh=np.nan,
+                  units="units"):
     r"""Returns the probability at the velocity `v` in m/s
      to find a particle `particle` in a plasma of temperature `T`
      following the Maxwellian distribution function.
@@ -38,6 +39,17 @@ def Maxwellian_1D(v: u.m/u.s,
         
     V_drift: Quantity, optional
         The drift velocity in units convertible to m/s
+        
+    vTh: Quantity, optional
+        Thermal velocity (most probable) in m/s. This is used for
+        optimization purposes to avoid re-calculating vTh, for example
+        when integrating over velocity-space.
+        
+    units: string, optional
+        Selects whether to run function with units and unit checks (when
+        equal to "units") or to run as unitless (when equal to "unitless").
+        The unitless version is substantially faster for intensive
+        computations.
 
     Returns
     -------
@@ -75,10 +87,29 @@ def Maxwellian_1D(v: u.m/u.s,
     >>> Maxwellian_1D(v=v, T= 30000*u.K, particle='e',V_drift=0*u.m/u.s)
     <Quantity 5.916329687405703e-07 s / m>
     """
-    # convert temperature to Kelvins
-    T = T.to(u.K, equivalencies=u.temperature_energy())
-    # get thermal velocity and thermal velocity squared
-    vTh = thermal_speed(T, particle=particle, method="most_probable")
+    if units == "units":
+        # unit checks and conversions
+        # checking velocity units
+        v = v.to(u.m/u.s)
+        # catching case where drift velocities have default values, they
+        # need to be assigned units
+        if V_drift == 0:
+            if not isinstance(V_drift, astropy.units.quantity.Quantity):
+                V_drift = V_drift * u.m/u.s
+        # checking units of drift velocities
+        V_drift = V_drift.to(u.m/u.s)
+        # convert temperature to Kelvins
+        T = T.to(u.K, equivalencies=u.temperature_energy())
+        if np.isnan(vTh):
+            # get thermal velocity and thermal velocity squared
+            vTh = (thermal_speed(T, particle=particle, method="most_probable"))
+        elif not np.isnan(vTh):
+            # check units of thermal velocity
+            vTh = vTh.to(u.m/u.s)
+    elif np.isnan(vTh) and units == "unitless":
+        # assuming unitless temperature is in Kelvins
+        vTh = (thermal_speed(T*u.K, particle=particle, method="most_probable")).si.value
+    # Get thermal velocity squared
     vThSq = vTh ** 2
     # Get square of relative particle velocity
     vSq = (v - V_drift) ** 2
@@ -86,7 +117,10 @@ def Maxwellian_1D(v: u.m/u.s,
     coeff = (vThSq * np.pi) ** (-1 / 2)
     expTerm = np.exp(-vSq / vThSq)
     distFunc = coeff * expTerm
-    return distFunc.to(u.s/u.m)
+    if units == "units":
+        return distFunc.to(u.s/u.m)
+    elif units == "unitless":
+        return distFunc
 
 def Maxwellian_velocity_3D(vx,
                            vy,
@@ -226,8 +260,6 @@ def Maxwellian_velocity_3D(vx,
         vTh = (thermal_speed(T*u.K, particle=particle, method="most_probable")).si.value
     # accounting for thermal velocity in 3D
     vThSq = 3 * vTh ** 2
-#    vTh3D = np.sqrt(3) * vTh
-#    vThSq = vTh3D ** 2
     # Get square of relative particle velocity
     vSq = ((vx-Vx_drift) ** 2 + (vy-Vy_drift) ** 2 + (vz-Vz_drift) ** 2)
     # calculating distribution function
--- a/plasmapy/physics/distribution.py
+++ b/plasmapy/physics/distribution.py
@@ -1,6 +1,6 @@
 """Functions to deal with distribution : generate, fit, calculate"""
 
-from astropy import units
+from astropy import units as u
 
 from astropy.units import (UnitConversionError, UnitsError, quantity_input,
                            Quantity)
@@ -14,14 +14,14 @@ import numpy as np
 from ..utils import _check_quantity, check_relativistic, check_quantity
 
 
-@units.quantity_input
-def Maxwellian_1D(v: units.m/units.s,
+@u.quantity_input
+def Maxwellian_1D(v: u.m/u.s,
                   T, 
                   particle="e",
-                  V_drift=0*units.m/units.s):
+                  V_drift=0*u.m/u.s):
     r"""Returns the probability at the velocity `v` in m/s
-    to find a particle `particle` in a plasma of temperature `T`
-    following the Maxwellian distribution function.
+     to find a particle `particle` in a plasma of temperature `T`
+     following the Maxwellian distribution function.
 
     Parameters
     ----------
@@ -76,7 +76,7 @@ def Maxwellian_1D(v: units.m/units.s,
     <Quantity 5.916329687405703e-07 s / m>
     """
     # convert temperature to Kelvins
-    T = T.to(units.K, equivalencies=units.temperature_energy())
+    T = T.to(u.K, equivalencies=u.temperature_energy())
     # get thermal velocity and thermal velocity squared
     vTh = thermal_speed(T, particle=particle, method="most_probable")
     vThSq = vTh ** 2
@@ -89,17 +89,17 @@ def Maxwellian_1D(v: units.m/units.s,
         distFunc = coeff * expTerm
     except Exception:
         raise ValueError("Unable to get distribution function.")
-    return distFunc.to(units.s/units.m)
+    return distFunc.to(u.s/u.m)
 
-@units.quantity_input
-def Maxwellian_velocity_3D(vx: units.m/units.s,
-                           vy: units.m/units.s,
-                           vz: units.m/units.s,
+@u.quantity_input
+def Maxwellian_velocity_3D(vx: u.m/u.s,
+                           vy: u.m/u.s,
+                           vz: u.m/u.s,
                            T, 
                            particle="e",
-                           Vx_drift=0*units.m/units.s,
-                           Vy_drift=0*units.m/units.s,
-                           Vz_drift=0*units.m/units.s):
+                           Vx_drift=0*u.m/u.s,
+                           Vy_drift=0*u.m/u.s,
+                           Vz_drift=0*u.m/u.s):
     r"""Return the probability of finding a particle with velocity components
     `v_x`, `v_y`, and `v_z`in m/s in an equilibrium plasma of temperature 
     `T` which follows the 3D Maxwellian distribution function. This 
@@ -184,7 +184,7 @@ def Maxwellian_velocity_3D(vx: units.m/units.s,
     
     """
     # convert temperature to Kelvins
-    T = T.to(units.K, equivalencies=units.temperature_energy())
+    T = T.to(u.K, equivalencies=u.temperature_energy())
     # get thermal velocity and thermal velocity squared
     vTh = thermal_speed(T, particle=particle, method="most_probable")
     # accounting for thermal velocity in 3D
@@ -199,13 +199,13 @@ def Maxwellian_velocity_3D(vx: units.m/units.s,
         distFunc = coeff * expTerm
     except Exception:
         raise ValueError("Unable to get distribution function.")
-    return distFunc.to((units.s/units.m) ** 3)
+    return distFunc.to((u.s/u.m) ** 3)
 
-@units.quantity_input
-def Maxwellian_speed_1D(v: units.m/units.s,
+@u.quantity_input
+def Maxwellian_speed_1D(v: u.m/u.s,
                         T, 
                         particle="e",
-                        V_drift=0*units.m/units.s):
+                        V_drift=0*u.m/u.s):
     r"""Return the probability of finding a particle with speed `v` in m/s
      in an equilibrium plasma of temperature `T` which follows the 
      Maxwellian distribution function.
@@ -266,7 +266,7 @@ def Maxwellian_speed_1D(v: units.m/units.s,
     
     """
     # convert temperature to Kelvins
-    T = T.to(units.K, equivalencies=units.temperature_energy())
+    T = T.to(u.K, equivalencies=u.temperature_energy())
     # get thermal velocity and thermal velocity squared
     vTh = thermal_speed(T, particle=particle, method="most_probable")
     vThSq = vTh ** 2
@@ -280,17 +280,17 @@ def Maxwellian_speed_1D(v: units.m/units.s,
         distFunc = coeff1 * coeff2 * expTerm
     except Exception:
         raise ValueError("Unable to get distribution function.")
-    return distFunc.to(units.s/units.m)
+    return distFunc.to(u.s/u.m)
 
-@units.quantity_input
-def Maxwellian_speed_3D(vx: units.m/units.s,
-                        vy: units.m/units.s,
-                        vz: units.m/units.s,
+@u.quantity_input
+def Maxwellian_speed_3D(vx: u.m/u.s,
+                        vy: u.m/u.s,
+                        vz: u.m/u.s,
                         T, 
                         particle="e",
-                        Vx_drift=0*units.m/units.s,
-                        Vy_drift=0*units.m/units.s,
-                        Vz_drift=0*units.m/units.s):
+                        Vx_drift=0*u.m/u.s,
+                        Vy_drift=0*u.m/u.s,
+                        Vz_drift=0*u.m/u.s):
     r"""Return the probability of finding a particle with speed components
     `v_x`, `v_y`, and `v_z`in m/s in an equilibrium plasma of temperature 
     `T` which follows the 3D Maxwellian distribution function. This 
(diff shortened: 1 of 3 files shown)


plasmapy/physics/distribution.py
```diff
@@ -60,6 +65,8 @@ def Maxwellian_1D(v: units.m/units.s,
 
     .. math::
     f = \sqrt{\frac{m}{2 \pi k_B T}} \exp(-\frac{m}{2 k_B T} (v-V)^2)
+    f = (\pi * v_Th^2)^{-1/2} \exp(-(v - v_{drift})^2 / v_Th^2)
```
StanczakDominik: Could we add something like "or equivalently," or "two equivalent forms:" here?

That might break with the `.. math::` .rst tag. We'll have to keep an eye on that.

plasmapy/physics/distribution.py
```diff
@@ -70,25 +77,286 @@ def Maxwellian_1D(v: units.m/units.s,
...
+    
+    """
+    # convert temperature to Kelvins
+    T = T.to(units.K, equivalencies=units.temperature_energy())
+    # get thermal velocity and thermal velocity squared
+    vTh = thermal_speed(T, particle=particle, method="most_probable")
+    # accounting for thermal velocity in 3D
+    vTh3D = np.sqrt(3) * vTh
+    vThSq = vTh3D ** 2
+    # Get particle velocity squared
```
StanczakDominik: This isn't a big deal, but, "relative" particle velocity, maybe, given it's shifted by the drift one?

plasmapy/physics/distribution.py
```diff
@@ -70,25 +77,286 @@ def Maxwellian_1D(v: units.m/units.s,
...
+    is given by:
+
+    .. math::
+    f(v) = 4 \pi v^2 (\pi * v_Th^2)^{-3/2} \exp(-(v - v_{drift})^2 / v_Th^2)
+    where v_Th = \sqrt(2 k_B T / m) is the thermal speed
+
+
+    Example
+    -------
+
```
StanczakDominik: Could definitely use some doctests here :)

plasmapy/physics/distribution.py
```diff
@@ -70,25 +77,286 @@ def Maxwellian_1D(v: units.m/units.s,
...
+    f = 4 \pi \vec{v}^2 (\pi * v_Th^2)^{-3/2} \exp(-(\vec{v} - \vec{v_{drift}})^2 / v_Th^2)
+    where v_Th = \sqrt(2 k_B T / m) is the thermal speed
+
+    See also
+    --------
+    Maxwellian_speed_1D
+
+    Example
+    -------
+
```
StanczakDominik: As previously, a doctest would be nice! We could go with a single particle for simplicity.

plasmapy/physics/tests/test_distribution.py
```diff
@@ -42,3 +47,84 @@ def test_Maxwellian_1D():
...
+    # get thermal velocity and thermal velocity squared
+    vTh = thermal_speed(T, particle=particle, method="most_probable")
+    # setting up integration from 0 to 10*vTh
+    xData1D = np.arange(0, 10.01, 0.01) * vTh
+    yData1D = Maxwellian_speed_1D(v=xData1D,T=T,particle=particle)
+    # integrating, this should be close to 1
+    integ = spint.trapz(y=yData1D, x=xData1D)
+    print("1D integral {0}".format(integ))
+    exceptStr = "Integral of distribution function should be 1."
+    assert np.isclose(integ.value, 1), exceptStr
```
StanczakDominik: How about this, using the power of python 3.6 (though a `.format` would of course work as well, except for being less _awesome_):

    # cut the print
    exceptStr = f"Integral of distribution function should be 1 and not {integ}."
    assert # as previously except it returns the value now

plasmapy/physics/tests/test_distribution.py
```diff
@@ -42,3 +47,84 @@ def test_Maxwellian_1D():
...
+    # defining a closure with partially applied arguments
+    def velDist3D(vx, vy, vz):
+        velDist = Maxwellian_velocity_3D(vx=vx,
+                                         vy=vy,
+                                         vz=vz,
+                                         T=T, 
+                                         particle=particle)
+        return velDist
+    # setting up integration from -10*vTh to 10*vTh, which is close to Inf
+    infApprox = 10 * vTh
```
StanczakDominik: I'm trying to think of a way we could check whether this is a good approximation (aside from calculating this again for 20 vTh) and raise a warning if it turns out not to be one.

plasmapy/physics/tests/test_distribution.py
```diff
@@ -42,3 +47,84 @@ def test_Maxwellian_1D():
...
+    # if I recall correctly, astropy.units may have trouble with
+    # accounting for units in integrals, so this step currently fails when
+    # tplquad tries to convert a quantity with units to a scalar.
+    # integrating, this should be close to 1
```
StanczakDominik: Turn this into a simple float via `velDist3D.si.value`, then multiply the result by the proper units? `infApprox` would also have to be changed.

plasmapy/physics/tests/test_distribution.py
```diff
@@ -42,3 +47,84 @@ def test_Maxwellian_1D():
...
+    integ = spint.tplquad(velDist3D,
+                          -infApprox,
+                          infApprox,
+                          lambda z: -infApprox,
+                          lambda z: infApprox,
+                          lambda z, y: -infApprox,
+                          lambda z, y: infApprox,
+                          epsabs=1.49e-08*u.m/u.m, 
+                          epsrel=1.49e-08*u.m/u.m)
```
StanczakDominik: I'm curious, why this number in particular? Also, the units seem off, could this perhaps be why `tplquad` failed here? If we're integrating velocities, don't we need `u.m/u.s`?

plasmapy/physics/distribution.py
lemmatum: Oh, I was intending to remove the first form listed there. I think it makes more sense to look at the function as a comparison of velocities to obtain a probability with respect to the most probably velocity rather than having temperatures, and masses floating around. We can alway add the thermal velocity function under "see also"

plasmapy/physics/distribution.py
lemmatum: How do you guys run doctests?

plasmapy/physics/tests/test_distribution.py
lemmatum: oops, I'll delete the `print` line, that was mainly for my own testing. I like the idea of including the value in the exception string though!

plasmapy/physics/tests/test_distribution.py
lemmatum: I think this ends up being something like 50 or 100 e-foldings of decay, since we have roughly `exp(-v**2 / vTh**2) => exp(-(10*vTh)**2 / vTh**2) => exp(-100)`, times some coefficients. In any case, this test works for the 1D case and passes the `isclose` to 1 test with default errors.

plasmapy/physics/tests/test_distribution.py
lemmatum: Those are just the default values in tplquad 😉 , I put them in explicitly to check if adding dimensionless units would help the test behave correctly. Sometimes this works according to the "known issues" with astropy.units pages I linked below.

The final results from integration should be a probability, which is dimensionless, therefore the errors on the final value are dimensionless. `u.m/u.m` was just a shortcut for adding a dimensionless unit.

Here are some references for why the integration doesn't behave well with astropy.units. [Here](http://docs.astropy.org/en/stable/units/quantity.html#known-issues-with-conversion-to-numpy-arrays) and [here](http://docs.astropy.org/en/stable/known_issues.html#quantity-issues).

plasmapy/physics/distribution.py
StanczakDominik: That documentation update really can't come soon enough, can it? :D I'll try to automate and simplify the process a bit while I'm at it.

If you clone the repo to a directory called PlasmaPy (so you have a `PlasmaPy/plasmapy` directory where the actual repository is located, you should be able to run `pytest` or more specifically `pytest plasmapy/physics/distribution.py` from the PlasmaPy  directory and that should, due to seeing the `setup.cgg` file, include doctests.

plasmapy/physics/tests/test_distribution.py
StanczakDominik: ... 50 e-foldings? Okay. Nevermind the warning. :laughing:

plasmapy/physics/tests/test_distribution.py
StanczakDominik: Oh right, it's an error on the final value! I forgot about that, you're quite right.

Well, we can still go for floats on the internal side!

plasmapy/physics/distribution.py
StanczakDominik: It probably wouldn't hurt too much to have both of those available, but it's not critical to have both here, I suppose!

lemmatum:
I've fixed the issues with astropy.units and tplquad not playing nicely. The fix was relatively simple, but pytest now runs really slowly or stalls. When it does work, just one of the 3D tests takes about 150 seconds. I've tried profiling with cProfile, but it also stalls. I'm attaching a partial cProfile which may give some insights on how to optimize.
[test_plasmaPy_cProfile.txt](https://github.com/PlasmaPy/PlasmaPy/files/1429746/test_plasmaPy_cProfile.txt)

plasmapy/physics/tests/test_distribution.py
```diff
@@ -90,10 +93,16 @@ def velDist3D(vx, vy, vz):
                           lambda z: infApprox,
                           lambda z, y: -infApprox,
                           lambda z, y: infApprox,
-                          epsabs=1.49e-08*u.m/u.m, 
-                          epsrel=1.49e-08*u.m/u.m)
-    exceptStr = "Integral of distribution function should be 1."
-    assert np.isclose(integ.value, 1), exceptStr
+                          epsabs=0.0,
```
lemmatum: probably should change this from 0.0 to some larger value as I'm not sure how QUADPACK handles 0.0 absolute error (whether infinitely small or just ignored).

lemmatum:
I just tested out a version of Maxwellian_velocity_3D() with all the unit checks and conversions stripped out and passing additional parameters under args=() in tplquad rather than creating a partially-applied function. It runs in 1.6 seconds.... and gives 1.0000000181399045 as the results of the triple integral, which is pretty damn close.

So perhaps the speed of unit conversions, mentioned in  #131 , is a valid concern. Need to dig in and figure out what was killing the optimizations in tplquad.

plasmapy/physics/tests/test_distribution.py
```diff
@@ -91,10 +93,16 @@ def velDist3D(vx, vy, vz):
                           lambda z: infApprox,
                           lambda z, y: -infApprox,
                           lambda z, y: infApprox,
-                          epsabs=1.49e-08*u.m/u.m, 
-                          epsrel=1.49e-08*u.m/u.m)
-    exceptStr = "Integral of distribution function should be 1."
-    assert np.isclose(integ.value, 1), exceptStr
+                          epsabs=0.0, 
+                          epsrel=5.0e-1)
+    # value returned from tplquad is (integral, error), we just need the 1st
```
StanczakDominik: Or we could use the returned error in line 102-105, as `atol`? Does that make sense?

StanczakDominik:
I'd have to bet (70% of my confidence) on unit conversions. @cadair mentioned them being expensive operations, to be used sparingly. Still... it might be the partial function application.

plasmapy/physics/tests/test_distribution.py
lemmatum: I'm not sure. I don't think the error here is the error on the final value, but rather the error in how the function slices up the bounds into individual units to be calculated, or the error on these local approximate integrals.

lemmatum: In any case using `epabs=5.0e-1` gave me quite a reasonable value on the final integral value when I stripped the Maxwellian function of units. This was the aforementioned 1.0000000181399045.

lemmatum:
Here are some ideas on how to resolve this
* Write the Maxwellian without units
* Write two Maxwellian functions, one with units and the other without
* Write the Maxwellian with a flag to disable unit checks, and move unit checks and conversions to an if statement inside the function
* Write a function that takes the proposed inputs (with units), and returns a scaled version of the Maxwellian function but without units.

Some further optimization could be done by passing vThermal as an argument. For any task, such as integration, where the temperature is the same, vThermal will be the same. The way the function is written right now it needlessly calculates vThermal on each call.

Some of this will depend on what part of the code is actually slowing down tplquad. My inclination is towards option 3.

StanczakDominik:
A reasonable inclination, that. Avoiding recalculating the thermal velocity
is also a good idea!

On Nov 1, 2017 02:31, "Pawel Kozlowski" <notifications@github.com> wrote:

Here are some ideas on how to resolve this

   - Write the Maxwellian without units
   - Write two Maxwellian functions, one with units and the other without
   - Write the Maxwellian with a flag to disable unit checks, and move unit
   checks and conversions to an if statement inside the function
   - Write a function that takes the proposed inputs (with units), and
   returns a scaled version of the Maxwellian function but without units.

Some further optimization could be done by passing vThermal as an argument.
For any task, such as integration, where the temperature is the same,
vThermal will be the same. The way the function is written right now it
needlessly calculates vThermal on each call.

Some of this will depend on what part of the code is actually slowing down
tplquad. My inclination is towards option 3.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/PlasmaPy/PlasmaPy/pull/152#issuecomment-340953150>, or mute
the thread

lemmatum:
Apologies for all the pushes. I was learning how to rebase since upstream's distribution.py was updated with edits to the docstrings.
Anyway this version works now! Tests pass in about 1 second when using "unitless" flag in distribution functions.

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+    vTh = thermal_speed(T, particle=particle, method="most_probable")
+    vThSq = vTh ** 2
+    # Get square of relative particle velocity
+    vSq = (v - V_drift) ** 2
+    # calculating distribution function
+    try:
+        coeff = (vThSq * np.pi) ** (-1 / 2)
+        expTerm = np.exp(-vSq / vThSq)
+        distFunc = coeff * expTerm
+    except Exception:
```
StanczakDominik: Code coverage whines about this one, so either add some test with `pytest.raises(ValueError)`  or simply plot down a `# coveralls: ignore` after the `Exception:` to skip testing this one!

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+        # unit checks and conversions
+        # checking velocity units
+        vx = vx.to(u.m/u.s)
+        vy = vy.to(u.m/u.s)
+        vz = vz.to(u.m/u.s)
+        # catching case where drift velocities have default values, they
+        # need to be assigned units
+        if Vx_drift == 0:
+            if not isinstance(Vx_drift, astropy.units.quantity.Quantity):
+                Vx_drift = Vx_drift * u.m/u.s
```
StanczakDominik: The three `Vdirection_drift = ... ` lines don't get covered by test. Adding another small one with a floating point value would solve that!

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+        # catching case where drift velocity has default value, and
+        # needs to be assigned units
+        if V_drift == 0:
+            if not isinstance(V_drift, astropy.units.quantity.Quantity):
+                V_drift = V_drift * u.m/u.s
+        # checking drift velocity units
+        V_drift = V_drift.to(u.m/u.s)
+        # convert temperature to Kelvins
+        T = T.to(u.K, equivalencies=u.temperature_energy())
+        if vTh == -1:
```
StanczakDominik: This is a neat idea! Would adding `np.nan` instead of `-1` make sense here, to signify that it's a nonsensical value?

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+            # get thermal velocity and thermal velocity squared
+            vTh = (thermal_speed(T, particle=particle, method="most_probable"))
+        elif vTh != -1:
+            # check units of thermal velocity
+            vTh = vTh.to(u.m/u.s)
+    elif vTh == -1 and units == "unitless":
+        # assuming unitless temperature is in Kelvins
+        vTh = (thermal_speed(T*u.K, particle=particle, method="most_probable"))
```
StanczakDominik: This one doesn't get tested :(

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+        # checking drift velocity units
+        V_drift = V_drift.to(u.m/u.s)
+        # convert temperature to Kelvins
+        T = T.to(u.K, equivalencies=u.temperature_energy())
+        if vTh == -1:
+            # get thermal velocity and thermal velocity squared
+            vTh = (thermal_speed(T, particle=particle, method="most_probable"))
+        elif vTh != -1:
+            # check units of thermal velocity
+            vTh = vTh.to(u.m/u.s)
```
StanczakDominik: If the velocity variable can be used as a velocity (as the tests are passing later), is this check (elif block from line 338) necessary? Not 100% sure.

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+    # get square of relative particle speed
+    vSq = (v - V_drift) ** 2
+    # calculating distribution function
+    try:
+        coeff1 = (np.pi * vThSq) ** (-3 / 2)
+        coeff2 = 4 * np.pi * vSq
+        expTerm = np.exp(-vSq / vThSq)
+        distFunc = coeff1 * coeff2 * expTerm
+    except Exception:
+        raise ValueError("Unable to get distribution function.")
```
StanczakDominik: This block doesn't get tested either. Could `ignore` it.

plasmapy/physics/distribution.py
```diff
@@ -58,35 +63,460 @@ def Maxwellian_1D(v: units.m/units.s,
...
+        # unit checks and conversions
+        # checking velocity units
+        vx = vx.to(u.m/u.s)
+        vy = vy.to(u.m/u.s)
+        vz = vz.to(u.m/u.s)
+        # catching case where drift velocities have default values, they
+        # need to be assigned units
+        if Vx_drift == 0:
+            if not isinstance(Vx_drift, astropy.units.quantity.Quantity):
+                Vx_drift = Vx_drift * u.m/u.s
```
StanczakDominik: These 3 lines don't get tested either. Same situation as before, it seems!

StanczakDominik:
Oh yeah, I forgot to mention. Mad props on getting that test down to 1s! That's really amazing.

lemmatum:
It wasn't too difficult to optimize! Thanks those poor FORTRAN programmers who wrote QUADPACK 😜

StanczakDominik:
Their sacrifice shan't be forgotten. [*]

plasmapy/physics/distribution.py
lemmatum: Sure. Just tested it and it does work. Just had to switch from `==` to `np.isnan(vTh)` as the Truthiness test.

plasmapy/physics/distribution.py
lemmatum: Do you mean that if vTh gets added or multiplied to some other value that has units then astropy will throw an error? I was just trying to be explicit, but that might be a better way to go, less clutter.

plasmapy/physics/distribution.py
lemmatum: Or just remove the try-except block? I don't think it's doing much here.

plasmapy/physics/distribution.py
lemmatum: This refers to tests in `test_distribution.py`, correct?

plasmapy/physics/distribution.py
StanczakDominik: Well, if it's convertible to `u.m/u.s`, it's not going to throw any errors later even if you don't do the explicit check here. But explicit is neat, I was just asking out of curiosity. Let's keep it this way :)

plasmapy/physics/distribution.py
StanczakDominik: Yeah, that it does.

plasmapy/physics/distribution.py
StanczakDominik: We could do that as well. I'm not 100% sure why they pop up everywhere.

lemmatum: I just started by imitating other code I've seen in this package, but it seems like these try-except blocks are mostly useless for such simple functions.

lemmatum:
I added a bunch of tests under `Test_Maxwellian_speed_3D`, it coveralls is happy then I'll add similar tests for the other distribution functions.

lemmatum:
Looks like coveralls are happy with the 3D Maxwellian speed function tests. I'll make similar tests for the other functions.

plasmapy/physics/tests/test_distribution.py
```diff
@@ -74,6 +82,102 @@ def test_valErr(self):
...
                           particle='XXX')
+    def test_units_no_vTh(self):
+        """
+        Tests distribution function with units, but not passing vTh.
+        """
+        distFunc = Maxwellian_1D(v=self.v,
+                                 T=self.T_e, 
+                                 particle=self.particle,
+                                 units="units")
+        errStr = (f"Distribution function should be {self.distFuncTrue} "
```
StanczakDominik: Bonus points for f-string usage

lemmatum: 😸

Issue #150, opened 2017-10-29
Migrate tests to using unittest module?


lemmatum:
I think it'd be better to use a proper unit testing module like the built-in unittest, rather than using asserts with no feedback on tests passed. I have some familiarity with this module and could readily write some example tests that you can then build off of.

Thoughts?

StanczakDominik:
I don't think that's a flaw in pytest - astropy and matplotlib both use it, so I don't think it's necessary to switch to unittest (I'd be happy to be convinced otherwise, though!) I do think it's likely that our tests can be badly written! Can you point out some of the worse ones?

Perhaps instead of doing tests like this:

    assert result == 42

it would be enough to pay attention to write them in this way everywhere:

    assert result == 42, "result is not the answer to everything"

That way the string message is printed out whenever this test fails. Would that be okay?

lemmatum:
There are a few issues with assert. One big one is not being able to continue the testing after failure of an assert statement. Modules like pyTest and unittest will crank through all the tests and tell you how many successes, how many failures and where.
There are other advantages to modules like pyTest and unittest, such as being able to setup and teardown tests (very useful for testing classes/objects), and integrate tests for a package under a single script without having to explicitly call each function in every module in every subdirectory.

I agree that adding comments for test failure is a must, but I don't see a way to resolve the above issues (especially inability to elegantly fail a test) with assert statements. I'm not sure what the people at astropy and matplotlib are doing, but a quick search brings up modules that use [pyTest](https://github.com/matplotlib/matplotlib/blob/e69ea36995a76c37b928f9cab9af1161f8d36e05/lib/matplotlib/tests/test_marker.py) in matplotlib.

lemmatum:
btw, this ties into the whole `if __name__ == "__main__":` issue, which is that using these modules (e.g. pyTest), when you run a test script directly they'll tell you the successes and failures within that test script. At the moment the test scripts are just a bunch of functions that don't get run (at least in test_quantum.py).

lemmatum:
I think this and the `if __name__ == "__main__":` issues are mainly my ignorance of how pyTest works. With unittest you run the module directly using `python test_quantum.py`, whereas with pyTest you call `pyTest` instead. So yea, I agree pyTest is good, just need to cleanup and add more informative messages when tests fail.

StanczakDominik:
Did you remove a bunch of your posts in this discussion? I had email notifications for them and can't find them here right now :smile:

[...]

PR #153, opened 2017-10-30
Improve documentation on testing and doing PRs

StanczakDominik:
Here's a bunch of improvements to documentation on running tests and building docs locally.

I'd appreciate all input into style, etc. I would especially appreciate attempts to reproduce these by following the steps outlined. I may have been pretty broad there.

The changes are mostly in the [Development guide over here](http://plasmapy-stanczakdominik-test.readthedocs.io/en/testing-documentation/).

The section on pytest parametrization may be a little bit too "out there", but it was the one example that came to me at the time. Thanks to @lemmatum in #150 for helping me write this one!


StanczakDominik:
Oh yeah, @lemmatum, I haven't added your idea about using test classes yet because I haven't read up on that yet. I figured it's better to have something that works now rather than something that's optimal rather.

docs/development/code_guide.rst
```diff
@@ -77,17 +77,116 @@ Documentation
...
+       assert proof_by_riemann(truth) == expected
+
+Code coverage
+-------------
+
+PlasmaPy uses the coverage.py addon via Coveralls.io. At the end of every
+Travis CI testing session, information on which lines were executed in the test
+is sent to Coveralls.io. At the very least, try to avoid test coverage
+decreasing if possible.
+
```
namurphy: Perhaps add the following paragraph somewhere around here?

> Occasionally there will be some lines that do not require testing.  For example, testing exception handling for an `ImportError` when importing an external package would usually be impractical.  In these instances, we may end a line with `# coveralls: ignore` to indicate that these lines should be excluded from coverage reports (or add a line to `.coveragerc`).  This strategy should be used sparingly, since it is often better to explicitly test exceptions and warnings and to show the lines of code that are not tested.

docs/development/code_guide.rst
```diff
@@ -77,17 +77,116 @@ Documentation
...
+Assert statements
+-----------------
+* Pytest runs tests by checking ``assert`` statements, so this is sufficient:
+
+.. code-block:: python
+
+  def test_universe_is_sane():
+      assert 2 + 2 == 4
+
+However, making assertions descriptive is better in most cases:
```
namurphy: Oops.  I seem to have neglected to do this frequently. 😸

docs/development/code_guide.rst
StanczakDominik: Done :smile:

docs/development/code_guide.rst
StanczakDominik: You might have. A bit. :laughing: Not to worry though! I might be able to break those up a little.

docs/development/code_guide.rst
```diff
@@ -77,18 +77,125 @@ Documentation
...
+  def test_proof_if_riemann_false():
+       assert proof_by_riemann(False)
+  def test_proof_if_riemann_true():
+       assert proof_by_riemann(True)
+
+but that's a lot of typing so what you actually do is use pytest parametrization:
+
+.. code-block:: python
+
+  @pytest.mark.parametrize("truth", [True, False])
```
namurphy: I am wondering how bad of an idea it would be to change the variable name `"truth"` here to [`"truthiness"`](https://en.wikipedia.org/wiki/Truthiness).

namurphy:
This looks just about ready to merge.  One possibility for the future would be to have a separate documentation file for everything about testing, though I don't think that's necessary yet.  Thank you!

StanczakDominik:
The code guidelines did get a bit bloated so I ended up separating this one.

Issue #151, opened 2017-10-29
Improve CONTRIBUTING.md documentation


lemmatum:
This is mostly a note to myself to add the notes I've taken on this process to the documentation. Things to include:
* How to setup virtualenv
  * export LD_LIBRARY_PATH when setting up virtualenv on Linux
  * No spaces allowed in path names
  * How to activate virtualenv
  * Installation of plasmapy dependencies in virtualenv
  * `python setup.py develop` command
* How to test/run newly written functions
  * import plasmapy.blah.blah in the virtualenv

lemmatum:
Looks like we have a minor issue in #160 , and then this will be ready to close soon.

StanczakDominik:
Closed by #160, #161 and possibly some more

PR #160, opened 2017-11-01
Add notes on setting up conda environment for development

StanczakDominik:
Makes way for #151 to add virtualenv instructions by adding those for conda. The changes can be seen [here](http://plasmapy-stanczakdominik-test.readthedocs.io/en/conda-setup-docs/).


docs/development/code_guide.rst
```diff
@@ -5,14 +5,61 @@ Code Development Guidelines
...
+To set up a development environment for PlasmaPy, `the Anaconda
+distribution <https://www.anaconda.com/download/>`_ is strongly
+recommended.
+
+After installing Anaconda, launching any conda environment and entering
+the PlasmaPy's repository root directory, execute the following:
+
+.. code-block:: bash
+
+    conda create -n plasmapy-dev python=3 --file requirements/base.txt
```
lemmatum: When I try to run this on my computer I get the following error:
`CondaValueError: could not parse 'numpy (>= 1.13)' in: requirements/base.txt`

Might be worth testing this out on a couple different operating systems or at least noting the ones this has been tested on.

lemmatum: interestingly enough, my computer was perfectly happy with `pip install -r requirements/base.txt` from inside virtualenv.

SolarDrew:
I think it would be good for these instructions to include how to set up conda and virtualenv on Windows, since at the moment they seem to be assuming bash.

StanczakDominik:
Hey @lemmatum, would you mind if I fixed this one, merged it, then rebased yours on top of these changes? I might have to do some moving stuff around to make the doc make sense.

docs/development/code_guide.rst
StanczakDominik: I finally got to it and I found a solution - replacing the call to requirements.txt to one to environment.yml via conda env create. It should work. I'll test this alongside writing Windows instructions in a couple of minutes.

StanczakDominik:
This is reviewable once again. @SolarDrew, I took both Windows scientific python developers into consideration. @lemmatum, that command that didn't work was pretty much based on me freestyling and guessing. I've verified that this approach works.

Let's get this merged, finally, shall we?

lemmatum:
@StanczakDominik #161  was already merged, so you probably want to rebase your branch and copy in the changes detailed setting up a conda environment, no?

docs/development/code_guide.rst
```diff
@@ -5,14 +5,73 @@ Code Development Guidelines
...
+.. code-block:: bash
+
+  git clone https://github.com/PlasmaPy/PlasmaPy.git 
+
+To be able to import PlasmaPy from your source version:
+
+.. code-block:: bash
+
+  cd PlasmaPy
+  python setup.py develop
```
lemmatum: I think it makes more sense to talk about setting up the environment prior to this step. My inclination reading through this is to run the command immediately, and then when I read the next few lines I get a warning about needing to work in a virtual environment. That warning should come earlier.

docs/development/code_guide.rst
```diff
@@ -5,14 +5,73 @@ Code Development Guidelines
...
+PlasmaPy's repository root directory.
+
+On Windows, the way to do this is via running `Anaconda Prompt` from the
+Start Menu. `Git Bash` may also work if you have added Anaconda to `PATH`.
+
+Afterwards, execute the following:
+
+.. code-block:: bash
+
+    conda env create
```
lemmatum: When I run this on the command line in Linux I get:
```
Using Anaconda API: https://api.anaconda.org

SpecNotFound: Can't process without a name
```

Is this command explicitly Windows specific?

docs/development/code_guide.rst
```diff
@@ -5,14 +5,73 @@ Code Development Guidelines
...
+
+.. code-block:: bash
+
+    source activate plasmapy-dev
+  
+
+On Windows, skip the `source` part of the previous command.
+Virtualenv
+----------
+.. note :: TODO add virtualenv notes
```
lemmatum: We need to sort out how to combine #161 and your changes. I think there will invariably be collisions as we were working on the same thing, so question is how to handle them. The way I handled this situation in a previous merge on a different pull request, was to rebase my own branch with upstream and then interactively do a 3-way merge using [meld](http://meldmerge.org/) for viewing diffs and handling conflicts. I think doing it this way keeps the branch history clean for upstream because it effectively makes it so that your branch is shifted up in time and comes after the merge of #161 . How to visualize the [rebase](https://www.atlassian.com/git/tutorials/merging-vs-rebasing). Let me know if I am wrong on this!

docs/development/code_guide.rst
StanczakDominik: Nah, it's on Linux. I code badly when under thesis delivery deadline stress, though. It should be fixed now.

StanczakDominik:
Re: #161: whoa, I AM out of the loop, aren't I? I'm sorry, the last few days/weeks have been a torrent of frantically trying to get my thesis in by the deadline (assuming everything goes as expected tomorrow, I'm finally done with that) and PlasmaPy has been on the sidelines. I'll get to this with a clearer head over the weekend.

lemmatum:
@StanczakDominik no worries! Thesis is way more important, so focus on that!

docs/development/code_guide.rst
StanczakDominik: Having thought about that some more - sure, that makes sense. However, we now have the `requirements/base.txt` and `environment.yml`, and those two can be used to facilitate setting up the virtual environment - and the easiest way to get those is going to be via cloning the repo.

StanczakDominik:
As to combining our ~~powers~~ changes, I think the fundamental question here is whether we want to duplicate those notes in `CONTRIBUTING.md` and in the sphinx doc (thus keeping a fatter `CONTRIBUTING`), or whether we want to point `CONTRIBUTING.md` to the sphinx doc and keep that one lean. As of now I tihnk the latter makes more sense. @lemmatum would you mind me moving your writeup and adapting both of them to sort of conform in style?

docs/development/code_guide.rst
Cadair: You should avoid using `python setup.py develop` in favor of `pip install -e ./` using `develop` can lead to having (many) multiple copies of the library installed at once.

environment.yml
```diff
@@ -0,0 +1,9 @@
+name: plasmapy-dev
+dependencies:
+  - numpy
+  - scipy
+  - astropy
+  - numpydoc
+  - pytest
```
Cadair: add sphinx for documentation?

lemmatum:
@StanczakDominik the sphinx option sounds good to me. And of course you are welcome to move things around as you see fit!

environment.yml
StanczakDominik: definitely, will do

docs/development/code_guide.rst
StanczakDominik: huh, today I learned! Thanks a bunch!

StanczakDominik:
I finally got the rebase done. Let's get this done quickly with a couple of quick review iterations.

I'm now also slightly convinced that for scientific Python related work, virtualenv is rather inferior to conda :laughing:

lemmatum:
@StanczakDominik heyyyyy, I'm using virtualenv and it works out just fine for me! (And every time I've tried to run one of your conda environment setup commands it invariably fails 👿 ).

docs/development/code_guide.rst
```diff
@@ -5,14 +5,184 @@ Code Development Guidelines
...
+
+SSH is a more secure option, but requires you to [set up an SSH
+key](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/)
+beforehand.  The equivalent SSH command is:
+
+.. code-block:: bash
+
+    git clone git@github.com:your-username/PlasmaPy.git
+
+After cloning, we must tell git where the development version of
```
Cadair: you might want to explictly say to `cd` into the clone here.

docs/development/code_guide.rst
```diff
@@ -5,14 +5,184 @@ Code Development Guidelines
...
+
+  origin		git@github.com:namurphy/PlasmaPy.git (fetch)
+  origin		git@github.com:namurphy/PlasmaPy.git (push)
+  upstream	git@github.com:PlasmaPy/PlasmaPy.git (fetch)
+  upstream	git@github.com:PlasmaPy/PlasmaPy.git (push)
+
+Setting up an environment for development
+=========================================
+
+To access Git commands on Windows, try `Git Bash <https://git-scm.com/downloads>`_.
```
Cadair: would this not be better placed above the git section above?

StanczakDominik:
@lemmatum yeah, that's because the commands I wrote were pretty bad before, and where they were okay, my description of their usage was lacking :laughing: Don't blame the distribution, blame the documenter. I'd appreciate it if you could test the current iteration (once I push a fix to @Cadair's notes, within a few minutes).

docs/development/code_guide.rst
StanczakDominik: Yup. Done.

docs/development/code_guide.rst
StanczakDominik: Yeah, I missed this while copypasting.

lemmatum:
So after I copied over the environment.yml file `conda env create -f environment.yml` worked, but now `source activate plasmapy-dev` gives me the error `pyenv-virtualenv: version 'plasmapy-dev' is not a virtualenv`... weird...

StanczakDominik:
@lemmatum and this is as it should be because it isn't a virtualenv. Conda has its own special version of `activate`, which you launch by going into an anaconda environment (not a virtualenv), usually by something along the lines of (an alias to) `source /home/dominik/.anaconda3/bin/activate root`. Conda installs a `root` (if I remember correctly) environment by default; you can probably also drop the `root` name and just use their `activate`, it'll drop you into some conda env.

I guess this wasn't clear enough; I'll add a note on this. Thanks! Could you try again?

docs/development/code_guide.rst
```diff
@@ -5,14 +5,185 @@ Code Development Guidelines
...
+.. code-block:: bash
+
+    conda env create -f environment.yml
+
+The `environment.yml` file refers to the one located in PlasmaPy's root.
+You may now enter the environment via
+
+.. code-block:: bash
+
+    source activate plasmapy-dev
```
lemmatum: This part needs to be changed to be specific to conda.

StanczakDominik: I took a slightly different approach as going first into any conda env, creating the environment via the .yml file and then activating `plasmapy-dev` should let you skip all the library linking... I think. I'm not sure how that part even worked given you weren't in a conda env.

Oh yeah, if you'd like to test this again you may wanna run `conda env remove -n plasmapy-dev` from conda's root to prepare. Sorry for the trouble.

lemmatum:
When I do `source /home/user/anaconda3/bin/activate` I do indeed get the (root) virtual environment in conda. Should this already have all the relevant plasmapy dependencies installed due to the environment.yml file?
Anything else I need to test?

StanczakDominik:
It should, but that's on a technicality as most of the stuff plasmapy needs is a subset of packages provided with the default Anaconda distribution. I'd appreciate it if you could still check 

    (possibly necessary to) conda env remove -n plasmapy-dev
    conda env create -f environment.yml
    source activate plasmapy-dev
    cd PlasmaPy
    pip install -e .
    python
    >>> import plasmapy

lemmatum:
> conda env remove -n plasmapy-dev

This works
> conda env create -f environment.yml

Works
> source activate plasmapy-dev

Doesn't work, have to go via `source /home/user/anaconda3/envs/plasmapy-dev/bin/activate` or `source /home/user/anaconda3/bin/activate` to activate (root) virtual env.
> cd PlasmaPy

I was already under PlasmaPy directory for the previous commands... shouldn't this be after git cloning but before creating the conda environment as the environment.yml file is under the PlasmaPy directory?
> pip install -e .

Seems to work.
> python

yup
> import plasmapy

and works.

StanczakDominik:
> Doesn't work, have to go via source /home/user/anaconda3/envs/plasmapy-dev/bin/activate or source /home/user/anaconda3/bin/activate to activate (root) virtual env.

Well, it should work, as it's always worked for me. I think there may be some issue with all the library linking you did messing with this somehow... but I'm not sure. Do you get any particular tracebacks or errors or whatever, or does it just simply not go into `plasmapy-dev`? @namurphy @SolarDrew could I get a second opinion on this?

> I was already under PlasmaPy directory for the previous commands... shouldn't this be after git cloning but before creating the conda environment as the environment.yml file is under the PlasmaPy directory?

Yeah, probably should. I removed the call to `cd` and replaced it with just pointing `pip install` properly.

lemmatum:
Weird, now `source activate plasmapy-dev` magically works and I get a (plasmapy-dev) environment. Previously it was throwing some errors about the environment not even existing.

Cadair:
@lemmatum I would guess looking at that message that you have a conflicitng activate script from `pyenv` and `conda`. Depending on which one is higher in your path it's either working or erroring.

lemmatum:
@Cadair 🔔 ding! ding! ding! You win the prize!
pyenv is indeed higher up in my path than anaconda, when I put anaconda first then everything works as expected!

Tangential question: is it possible to run Spyder interactively within a conda virtual environment? If so that would be a handy bit of functionality, especially for plots.

lemmatum:
Any other changes necessary for this pull request?

StanczakDominik:
I'd still like to wait for  @namurphy 's opinion on this as IIRC contributing.md was mostly his work and if I'm mangling it I'd prefer to mangle it well.

Issue #158, opened 2017-11-01
Adding Langmuir probe data analysis


jasperbeckers:
The current direction of PlasmaPy seems to be mostly theoretical/computational. I think it would be good to cater to experimentalists too, for example by providing data analysis functionality. There have been previous discussions about this (eg. https://github.com/PlasmaPy/PlasmaPy/issues/86) but I don't know if those ever went anywhere.

In any case, I'd like to begin implementing some basic Langmuir probe data analysis functionality as a start. **It is one of the most common diagnostics and therefore integral to plasma physics as a whole.**

The overall functionality would be rather simple; taking potential and current measurements along with some probe and species properties and returning the obtained plasma parameters. This could look something like `[T_e, n_e, V_B, V_P, ...] = plasmapy.langmuir.analyze(potential, current, species, probe_properties)`. Of course there is a large amount of theory written on the subject and many different methods, depending on ie. plasma regime and magnetization. Implementation of the most common methods, along with good documentation, would probably be sufficient initially.

Thoughts?

StanczakDominik:
* agreed 100%
* Pandas is neat for experimental data and it's probably as close as python
comes to a standard on that, does it make sense to accept dataframes and
have the function logic separate the columns as need be?
* How about uncertainties? The first version probably doesn't need to
include them but it'd be nice to have
* I'd love to have some sample data and expected parameters for that, it'd
make testing way better!
* plasmapy.langmuir sounds neat!


On Nov 1, 2017 06:07, "jasperbeckers" <notifications@github.com> wrote:

The current direction of PlasmaPy seems to be mostly
theoretical/computational. I think it would be good to cater to
experimentalists too, for example by providing data analysis functionality.
There have been previous discussions about this (eg. #86
<https://github.com/PlasmaPy/PlasmaPy/issues/86>) but I don't know if those
ever went anywhere.

In any case, I'd like to begin implementing some basic Langmuir probe data
analysis functionality as a start. *It is one of the most common
diagnostics and therefore integral to plasma physics as a whole.*

The overall functionality would be rather simple; taking potential and
current measurements along with some probe and species properties and
returning the obtained plasma parameters. This could look something like [T_e,
n_e, V_B, V_P, ...] = plasmapy.langmuir.analyze(potential, current,
species, probe_properties). Of course there is a large amount of theory
written on the subject and many different methods, depending on ie. plasma
regime and magnetization. Implementation of the most common methods, along
[...]

PR #165, opened 2017-11-03
Work on Langmuir diagnostic submodule

jasperbeckers:
First progress towards https://github.com/PlasmaPy/PlasmaPy/issues/158. Creation of the diagnostics submodule and `langmuir.py` file.

I've created a function for obtaining the plasma parameters such as energy density and temperature from a given Langmuir V-I profile, as well as for obtaining the plasma EEDF and EEPF.


plasmapy/diagnostics/langmuir.py
```diff
@@ -22,29 +23,30 @@ class Probe():
...
     def __init__(self, area=u.m**2, config='single'):
         # Initialize probe properties
         self.area = area
         self.config = config
 
-def langmuir_analysis(potential_sweep, measured_current, probe, ion='H',
-                      method='maxwellian', smoothing=None, polyorder=3):
+
+def swept_probe_analysis(potential_sweep, measured_current, probe, ion='H',
+                         method='maxwellian', smoothing=None, polyorder=3):
```
lemmatum: Put each argument on its own line.

plasmapy/diagnostics/langmuir.py
```diff
@@ -22,29 +23,30 @@ class Probe():
...
     r"""Performs a Langmuir analysis of a given probe V-I profile in order to
     obtain various plasma parameters.
 
     Parameters
-    ----------    
+    ----------
     potential_sweep : Quantity
         The potential sweep applied to the probe in units convertible to V.
```
lemmatum: I think "The swept electric potential applied..." would be clearer.

plasmapy/diagnostics/langmuir.py
```diff
@@ -22,29 +23,30 @@ class Probe():
...
     """
     def __init__(self, area=u.m**2, config='single'):
         # Initialize probe properties
         self.area = area
         self.config = config
 
-def langmuir_analysis(potential_sweep, measured_current, probe, ion='H',
-                      method='maxwellian', smoothing=None, polyorder=3):
+
+def swept_probe_analysis(potential_sweep, measured_current, probe, ion='H',
```
lemmatum: I think this should be particle instead of ion. Depending on how you set the potential you can measure electrons as well, and in many cases you want to just that!

plasmapy/diagnostics/langmuir.py
```diff
@@ -22,29 +23,30 @@ class Probe():
...
+
     measured_current : Quantity
         The corresponding measured current in units convertible to A.
-        
+
     probe : Probe
         An instance of the Probe class containing probe properties.
-    
+
     ion : string, optional
```
lemmatum: particle (see prior comment)

plasmapy/diagnostics/langmuir.py
```diff
@@ -67,39 +69,40 @@ def langmuir_analysis(potential_sweep, measured_current, probe, ion='H',
     -------
     T_e : Quantity
         Best estimate of the electron temperature in units of eV.
```
lemmatum: Clarification question - is this a best estimate or an initial guess for some iterative optimization?

plasmapy/diagnostics/langmuir.py
```diff
@@ -67,39 +69,40 @@ def langmuir_analysis(potential_sweep, measured_current, probe, ion='H',
...
-        
+
     n_e : Quantity
         Best estimate of the electron density in units of m^-3.
-        
+
     n_i : Quantity
         Best estimate of the ion density in units of m^-3. Equal to n_e
         unless the OML method is used.
```
lemmatum: This could be replaced with the average ionization. You then set a default ionization of 1 so that n_e=n_i for most cases.

plasmapy/diagnostics/langmuir.py
```diff
@@ -67,39 +69,40 @@ def langmuir_analysis(potential_sweep, measured_current, probe, ion='H',
...
 def obtain_EEDF(potential_sweep, measured_current, probe):
     r"""Obtains the Electron Energy Distribution Function of the plasma
     from the second derivative of the measured current response using
     Druyvestein's equations.
 
     Parameters
-    ----------    
+    ----------
     potential_sweep : Quantity
         The potential sweep applied to the probe in units convertible to V.
```
lemmatum: see prior comment

plasmapy/diagnostics/langmuir.py
```diff
@@ -108,22 +111,23 @@ def obtain_EEDF(potential_sweep, measured_current, probe):
...
 def obtain_EEPF(potential_sweep, measured_current, probe):
     r"""Obtains the Electron Energy Probability Function of the plasma
     from the second derivative of the measured current response using
     Druyvestein's equations.
 
     Parameters
-    ----------    
+    ----------
     potential_sweep : Quantity
         The potential sweep applied to the probe in units convertible to V.
```
lemmatum: see prior comment

plasmapy/diagnostics/langmuir.py
jasperbeckers: Well, the electron mass is known. What we need to know is the ion mass in order to obtain the ion density from the ion saturation current, right?

lemmatum: Fair point.

It may also depend on how you view the sweep, or rather what you assume as input (this should be documented in the docstring). Is the sweep across the entire voltage from ion saturation to electron saturation, is it just the ion saturation region, or just the electron dominated region before saturation, or does the user decide by setting the particle? My initial comment was assuming this last option, but that may not be the best way to do it.

plasmapy/diagnostics/langmuir.py
```diff
@@ -22,29 +23,30 @@ class Probe():
         Electrode configuration. Possible options are 'single' (default),
         'double', 'double-asym', 'triple', 'tetra', 'penta'. The electrode
         configuration must correspond to the number of areas provided.
-    
+
     """
     def __init__(self, area=u.m**2, config='single'):
         # Initialize probe properties
         self.area = area
         self.config = config
```
lemmatum: Coveralls mentions that these two lines don't get tested. Probably not so important now, but will be important once you start writing tests.

plasmapy/diagnostics/langmuir.py
```diff
@@ -108,22 +111,23 @@ def obtain_EEDF(potential_sweep, measured_current, probe):
     """
     return
```
lemmatum: Coveralls complains that none of the returns get tested. Shouldn't be a problem once there is actually a value to return and a simple test for the function, but just a heads up. Same goes for all the other empty returns.

jasperbeckers:
@lemmatum Ideally the analysis function would iteratively converge on an optimal estimate starting from a rough guess.

plasmapy/diagnostics/langmuir.py
jasperbeckers: Ideally it would be a best approximation based on converging iterations.

plasmapy/diagnostics/langmuir.py
jasperbeckers: Yes, that's something we'll have to decide. Fact is that in order to obtain some parameters (i.e. ion density, for a single probe) we need properties provided by multiple parts of the trace (ion saturation current, in the ion current dominated region, and electron temperature, which may follow from the exponential electron current region). 

Another thing to consider is that the sweep might be split up, for example due to amplification of the ion saturation current. We might need to add a function to "stitch" these together, or require the user to do that.

plasmapy/diagnostics/langmuir.py
jasperbeckers: Well, there might be a large difference in errors of the obtained parameters. For example, the ion density obtained with OML theory is often significantly more accurate than the electron density obtained from the electron saturation current. So cases exist where there is a large difference between `n_i` and `n_e`, but only because one method is simply better than the other. If then a strange ionization value is returned this might sent the wrong message to users.

lemmatum: Ah, yes I forgot that each of these are just initial values that are supposed to converge. Then this is fine, just make sure note this point and that the n_e and n_i that you get won't necessarily give you an accurate average ionization.

plasmapy/diagnostics/langmuir.py
lemmatum: I think initial guess or initial estimate would be a more appropriate label.

plasmapy/diagnostics/langmuir.py
lemmatum: Not sure how to deal with the former. The latter can be done with a pre-processing function where the user inputs the current trace and scale. They run this pre-processing function manually and then pass the current trace (which has now been standardized) into the `swept_probe_analysis` function.

jasperbeckers: Regarding the former, I think we should provide different functions for each method / section of the analysis, so that they can be called independently. The `swept_probe_analysis` function would then call upon these functions to provide a standard "off the shelf" analysis. That way a user can either use our quick and easy analysis function, or incorporate the subfunctions in a custom analysis.

jasperbeckers:
@lemmatum I pushed all changes you requested. What are your thoughts on my last comment?

> Regarding the former, I think we should provide different functions for each method / section of the analysis, so that they can be called independently. The `swept_probe_analysis` function would then call upon these functions to provide a standard "off the shelf" analysis. That way a user can either use our quick and easy analysis function, or incorporate the subfunctions in a custom analysis.

plasmapy/diagnostics/langmuir.py
```diff
@@ -83,6 +87,10 @@ def swept_probe_analysis(potential_sweep, measured_current, probe, ion='H',
     V_P : Quantity
         Best estimate of the plasma potential in units of V.
 
+    Notes
+    -----
+    Due to the use of different methods to obtain n_e and n_i, the ratio
+    between these does not necessarily give an accurate average ionization.
```
lemmatum: looks good

plasmapy/diagnostics/langmuir.py
```diff
@@ -31,15 +31,20 @@ def __init__(self, area=u.m**2, config='single'):
...
 
-def swept_probe_analysis(potential_sweep, measured_current, probe, ion='H',
-                         method='maxwellian', smoothing=None, polyorder=3):
+def swept_probe_analysis(potential_sweep, 
+                         measured_current, 
+                         probe, 
+                         ion='H',
+                         method='maxwellian', 
+                         smoothing=None, 
+                         polyorder=3):
```
lemmatum: much cleaner, I like!

plasmapy/diagnostics/langmuir.py
```diff
@@ -120,7 +128,7 @@ def obtain_EEPF(potential_sweep, measured_current, probe):
     Parameters
     ----------
     potential_sweep : Quantity
-        The potential sweep applied to the probe in units convertible to V.
+        The swept electric potential applied to the probe in units convertible to V.
```
lemmatum: looks good.

lemmatum:
@jasperbeckers Looks good. Have you changed the "best estimate" terminology to "initial guess" or have you decided to keep it as "best estimate"?

I think your idea for having a separate function for each section/type of analysis is sensible.

Can you write a simple `test_langmuir.py` file to make coveralls happy? You can use my `test_distribution.py` as a template for how to do tests in a classy way 😉

StanczakDominik:
Hey, @jasperbeckers, how's those tests coming along? If you're too busy, can I contribute those? :)

jasperbeckers:
@StanczakDominik Yes, you can go ahead! I've been traveling so I've only been passively following the project.

plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -0,0 +1,37 @@
+"""Tests for functions that use Langmuir analysis functions."""
+
+import numpy as np
+
+from ..langmuir import (swept_probe_analysis,
```
StanczakDominik: `from plasmapy.langmuir import...` would probably fix Travis screaming about this line

plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -0,0 +1,37 @@
...
+                        obtain_EEDF,
+                        obtain_EEPF)
+
+class Test_swept_probe_analysis(object):
```
StanczakDominik: You don't need to inherit `(object)` in python 3, that's the default!

plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_no_signal():
+        """
+        Checks whether an error is raised when zero signal is passed to the
+        function.
+        """
+        errStr = (f"Performing a swept probe analysis should return an error "
```
StanczakDominik: No need for f-strings if you're not passing in any variables into them. Also, `with pytest.raises(whatever_error)` will probably be handy here once we have the actual errors!

plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -0,0 +1,37 @@
...
+                  f"when zero signal is passed, but doesn't.")
+        assert True, errStr
```
StanczakDominik: At the very least `assert False`. /s :laughing:

plasmapy/diagnostics/tests/test_langmuir.py
jasperbeckers: Thanks, I wasn't aware!

StanczakDominik:
Oh right, I think we'd need to change __init__.py in the main code directory to add a line importing langmuir so that it's visible. We may also need an init py in the langmuir directory.

If need be, I'll fix this and push to this branch by Saturday.

jasperbeckers:
@StanczakDominik Please, go ahead. I'd like to get this PR out of the way so I can start writing some actual code :)

StanczakDominik:
@jasperbeckers your branch was preeeetty far behind master, so I think I'm going to have to rebase it on top of master to get this done. Is that okay with you?

StanczakDominik:
@jasperbeckers I took your "please go ahead" for a "yes"! You might want to do

   git fetch --all
   git reset --hard origin/langmuir_analysis

Anyway this should work now, methods in the test classes were missing calls to `self` (they'd work as just plain functions) and I forgot we're going to have a `diagnostics` directory instead of a `langmuir` directory, so that absolute import was off.

I also added the `diagnostics` subpackage to our documentation.

jasperbeckers:
Can we close this?

PR #122, opened 2017-09-26
Reorganize and refactor atomic subpackage

namurphy:
This pull request makes the following changes to the `atomic` subpackage:
- Rename `element_symbol` to `atomic_symbol`
- Rename `energy_from_nuclear_reaction` to `nuclear_reaction_energy`
- Move `nuclear_reaction_energy` and `nuclear_binding_energy` to new file called `nuclear.py`
- Clean up and reorganize docstrings in `atomic.py`
- Improve variable name clarity and consistency in `atomic.py`
- Change `atomic_symbols_list` to a dictionary called `atomic_symbols` in `elements.py`
- Use `warnings.warn` when there is a `UserWarning` instead of `raise`
- Disallow neutrons in `atomic_symbol` and `element_name`, but continue to allow them in `isotope_symbol`, `isotope_mass`, and `mass_number` because neutrons do count as nuclides
- Add helper functions to identify atomic special cases, including:
  - `__is_hydrogen`
  - `__is_positron`
  - `__is_electron`
  - `__is_antiproton`
  - `__is_neutron`
  - `__is_alpha`

Issue #110, opened 2017-09-13
Add functionality to find fusion reaction rates, cross sections, and energy generation rates


namurphy:
In laboratory fusion devices and in stellar astrophysics, it's important to be able to calculate quantities associated with different fusion reactions.  We will need a way to calculate the following quantities as functions of temperatures and densities:
- Reaction rates (the number of reactions per cubic meter per second)
- Reaction cross sections (in square meters)
- Energy generation (in joules released per cubic meter per second)

We have some of the capability already in the atomic package, including `nuclear_binding_energy`, `energy_from_nuclear_reaction` (which might be worth renaming `nuclear_reaction_energy`), and `half_life`.  In particular, `energy_from_nuclear_reaction` has code that parses reactions. Some of the reactions will be radioactive decay, so we will need the capability to calculate rates for that too.  
It will probably be reasonable for the first pull request to be able to do only `'D + T --> He-4 + n'` since getting that will be tricky enough!

There is a huge number of reactions that happen, so it will probably be worth working with applications in this order:
1. Laboratory fusion devices
2. Stellar interiors (especially the CNO cycle and the proton-proton chain)
3. Supernovae nucleosynthesis

Reactions that are important in the laboratory include:
- `'D + T --> He-4 + n'`
- `'D + D --> T + p'`
- `'D + D --> T + p'`
- `'D + He-3 --> He-4 + p'`
- `'T + T --> He-4 + 2n'`
- `'He-3 + He-3 --> He-4 + 2p'`
- `'He-3 + T --> He-4 + p + n'`
- `'He-3 + T --> He-3 + D'`
- `'D + Li-6 --> 2He-4'`
- `'D + Li-6 --> He-3 + He-4 + n'`
- `'D + Li-6 --> Li-7 + p'`
- `'D + Li-6 --> Be-7 + n'`
- `'He-3 + Li-6 --> He-4 + He-3'`
- `'p + B-11 --> 3He-4'`

I got these from the [Wikipedia article on nuclear fusion](https://en.wikipedia.org/wiki/Nuclear_fusion#Important_reactions), it turns out.  The articles on the [CNO cycle](https://en.wikipedia.org/wiki/CNO_cycle) and [proton-proton chain](https://en.wikipedia.org/wiki/Proton%E2%80%93proton_chain_reaction) should also be helpful.  There's even more in the article on [stellar nucleosynthesis](https://en.wikipedia.org/wiki/Stellar_nucleosynthesis).  

[...]

PR #169, opened 2017-11-23
Refactor nuclear_binding_energy and nuclear_reaction_rate

namurphy:
Prior to now, `nuclear_binding_energy` used `isotope_mass` to get the mass of the nuclide.  This is incorrect since `isotope_mass` gives the mass including the bound electrons.  Now `nuclear_binding_energy` uses `ion_mass` instead, and assumes that isotope is fully ionized to get the mass of the nucleus.

This pull request also implements the great suggestion by @jessicalostinspace in #110 to represent the reactants and products in a nuclear reaction as lists:
```Python
>>> nuclear_reaction_energy(reactants=['He-4', 'He-4'], products=['Be-8'])
```
For the time being, I decided to still allow the old API which has the argument being a string representing the reaction (e.g., `'He-4 + He-4 -> Be-8'`).  The advantage of allowing the old API is that it is more human readable.  The disadvantages are that it makes the implementation much more complex.  Do we want to continue to allow having a string representing the reaction as an input to `nuclear_reaction_energy`?  Also, should we require that `reactants` and `products` be keyword-only arguments?  I did enforce keyword-only arguments to make sure which is which is explicit, and because the minus sign really does matter, though maybe this isn't needed.  Suggestions, opinions, and cat emoji would be greatly appreciated! :cat2: :cat2: :cat2: :cat2: :cat2: :cat2:

`nuclear_reaction_energy` also now allows for electrons and positrons, which are important for beta decays.  It's also updated to use the masses of nuclei rather than the isotopic masses, as described above.

Cross-references: #24, #110, #122

### DIFF
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -1525,7 +1525,7 @@ def _is_electron(arg):
 
     if not isinstance(arg, str):
         return False
-   
+
     return arg in ['e', 'e-'] or arg.lower() == 'electron'
 
 
@@ -1542,7 +1542,7 @@ def _is_positron(arg):
 def _is_antiproton(arg):
     r"""Returns True if the argument corresponds to an antiproton, and
     False otherwise."""
-    
+
     if not isinstance(arg, str):
         return False
 
@@ -1552,7 +1552,7 @@ def _is_antiproton(arg):
 def _is_antineutron(arg):
     r"""Returns True if the argument corresponds to an antineutron, and
     False otherwise."""
-    
+
     if not isinstance(arg, str):
         return False
 
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -692,6 +692,7 @@ def isotope_mass(argument, mass_numb=None):
     <Quantity 4.00260325413 u>
     >>> isotope_mass(2, 4)
     <Quantity 4.00260325413 u>
+
     """
 
     argument, charge_state = _extract_charge_state(argument)
@@ -1147,6 +1148,7 @@ def stable_isotopes(argument=None, unstable=False):
 
     >>> stable_isotopes('U', unstable=True)[:5] # only first five
     ['U-217', 'U-218', 'U-219', 'U-220', 'U-221']
+
     """
 
     def stable_isotopes_for_element(argument, stable_only):
@@ -1568,7 +1570,8 @@ def _is_antiproton(argument):
 
 def _is_proton(argument):
     r"""Returns True if the argument corresponds to a proton, and
-    False otherwise."""
+    False otherwise.  This function returns False for 'H-1' if no
+    charge state is given."""
 
     if not isinstance(argument, str):
         return False
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -841,12 +841,6 @@ def ion_mass(argument, Z=None, mass_numb=None):
     else:
         Z_from_arg = None
 
-    if atomic_number(argument) == 1:
-        if isinstance(argument, str) and 'H-1' in str(argument) and Z is None:
-            return const.m_p
-        if mass_numb == 1 and Z == 1:
-            return const.m_p
-
     if Z is None and Z_from_arg is None:
         Z = 1
     elif Z is not None and Z_from_arg is not None and Z != Z_from_arg:
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -696,6 +696,9 @@ def isotope_mass(argument, mass_numb=None):
 
     argument, charge_state = _extract_charge_state(argument)
 
+    if _is_proton(argument) or _is_antiproton(argument):
+        return const.m_p
+
     if charge_state is not None and charge_state != 0:
         raise ValueError("Use ion_mass instead of isotope_mass for masses of "
                          "charged particles")
@@ -1558,6 +1561,19 @@ def _is_antiproton(argument):
         return False
 
     if argument == 'p-' or argument.lower() == 'antiproton':
+        return True
+    else:
+        return False
+
+
+def _is_proton(argument):
+    r"""Returns True if the argument corresponds to a proton, and
+    False otherwise."""
+
+    if not isinstance(argument, str):
+        return False
+
+    if argument in ['p', 'p+'] or argument.lower() == 'proton':
         return True
     else:
         return False
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -790,24 +790,22 @@ def ion_mass(argument, Z=None, mass_numb=None):
     --------
     >>> print(ion_mass('p').si.value)
     1.672621898e-27
-    >>> ion_mass('H')  # assumes terrestrial abundance of D
+    >>> ion_mass('H+')  # assumes terrestrial abundance of D
     <Quantity 1.672912413964e-27 kg>
-    >>> ion_mass('H') == ion_mass('p')
+    >>> ion_mass('H', Z=1) == ion_mass('p')
     False
-    >>> ion_mass('P')  # phosphorus
+    >>> ion_mass('P', Z=1)  # phosphorus
     <Quantity 5.14322300749914e-26 kg>
-    >>> ion_mass('He-4', 2)
+    >>> ion_mass('He-4', Z=2)
     <Quantity 6.644657088401906e-27 kg>
-    >>> ion_mass('T')
+    >>> ion_mass('T+')
     <Quantity 5.007356665e-27 kg>
     >>> ion_mass(26, Z=1, mass_numb=56)
     <Quantity 9.288123453752331e-26 kg>
-    >>> ion_mass('Fe-56')
+    >>> ion_mass('Fe-56', Z=1)
     <Quantity 9.288123453752331e-26 kg>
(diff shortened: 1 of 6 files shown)


plasmapy/atomic/atomic.py
```diff
@@ -696,6 +696,9 @@ def isotope_mass(argument, mass_numb=None):
 
     argument, charge_state = _extract_charge_state(argument)
 
+    if _is_proton(argument) or _is_antiproton(argument):
+        return const.m_p
```
namurphy: Test proton and antiproton masses in `isotope_mass`.  Make sure that special cases are documented.

plasmapy/atomic/atomic.py
```diff
@@ -1560,6 +1566,19 @@ def _is_antiproton(argument):
         return False
 
 
+def _is_proton(argument):
```
namurphy: Add tests for _is_proton

plasmapy/atomic/nuclear.py
```diff
@@ -88,95 +97,150 @@ def nuclear_reaction_energy(reaction):
...
-    reactants = _get_isotopes_list(LHS)
-    products = _get_isotopes_list(RHS)
+    try:
+        if _nucleon_number(reactants) != _nucleon_number(products):
+            raise ValueError("The number of nucleons is not conserved")
 
-    mass_num_reactants = _mass_number_of_list(reactants)
-    mass_num_products = _mass_number_of_list(products)
+        if _total_charge(reactants) != _total_charge(products):
+            raise ValueError("Total charge is not conserved")
```
namurphy: Add tests to `nuclear_reaction_energy` on charge conservation, including one that would raise this as a `ValueError`

plasmapy/atomic/nuclear.py
```diff
@@ -88,95 +97,150 @@ def nuclear_reaction_energy(reaction):
...
+                     "'products' as lists with the nucleons or particles "
+                     "involved in the reaction.")
+
+    if kwargs and not args and len(kwargs) == 2:  # keyword inputs
+
+        try:
+            reactants = _get_species(list(kwargs['reactants']))
+            products = _get_species(list(kwargs['products']))
+        except Exception:
+            raise ValueError(input_err_msg)
```
namurphy: Test this exception in `nuclear_reaction_energy`

samurai688:
I compared the updated binding energies that this calculates to [this official-looking table](https://www-nds.iaea.org/amdc/ame2016/mass16.txt) from [this official-looking website](https://www-nds.iaea.org/amdc/), and they agree almost exactly. Sweet!

I would vote for keeping both methods of input. In fact I think we could even have more generous options for interpreting what the input might be, for example I feel like I've sometimes seen a format where "6Li" would be parsed to mean 1 Lithium-6 atom, rather than 6 Lithium atoms of unknown isotope.

I think the implementation of multiple input formats could be kept reasonably clean. There could be a first "interface" part of the code that reduces any valid input to the reactants / products list form. From then on the function would operate exclusively from the list form.

namurphy:
Thank you, @samurai688!  I should add some tests with data from the links you provided.

One complication we'll have to be careful about is that the `'6'` in `'6Li'` could be interpreted as either a multiplicative factor or as a mass number.  If I'm thinking about this right, then there _shouldn't_ be ambiguity since we would know that `'Li'` is not an isotope by itself so we would need to use it as the mass number.  This would also require changes in `isotope_symbol`.  Does adding this notation for isotopes sound like a good idea for everyone?

namurphy:
I haven't implemented the suggestion to allow inputs like `'Be-8 --> 2He-4'`, which might need to wait for another pull request since I'm not sure what the best way is to deal with the potential ambiguity between the `2` being a multiplicative factor versus a mass number.

I kind of went down a rabbit hole on this pull request when I tried to stop `ion_mass` from assuming that an ion is singly ionized in no charge state information is given.

namurphy:
> In fact I think we could even have more generous options for interpreting what the input might be, for example I feel like I've sometimes seen a format where "6Li" would be parsed to mean 1 Lithium-6 atom, rather than 6 Lithium atoms of unknown isotope.

> One complication we'll have to be careful about is that the '6' in '6Li' could be interpreted as either a multiplicative factor or as a mass number.

After thinking about this, a good way to remove the ambiguity would be to require an asterisk for multipliers and no asterisk for when it is a mass number:
```Python
>>> nuclear_reaction_energy('Be-8 --> 2*4He')
```

plasmapy/atomic/nuclear.py
```diff
@@ -64,119 +65,235 @@ def nuclear_binding_energy(argument, mass_numb=None):
...
+
+        try:
+            LHS_string, RHS_string = re.split('-+>', reaction)
+            LHS_list = re.split(' \+ ', LHS_string)
+            RHS_list = re.split(' \+ ', RHS_string)
+            reactants = _get_species(LHS_list)
+            products = _get_species(RHS_list)
+        except Exception as ex:
+            raise ValueError(f"{reaction} is not a valid nuclear reaction.") \
+                from ex
```
StanczakDominik: I haven't seen this `from ex` syntax before. What's that?

plasmapy/atomic/tests/test_atomic.py
```diff
@@ -619,7 +624,9 @@ def test_half_life_u_220():
             atomic_TypeError_bad_arguments))
 def test_atomic_TypeErrors(function, argument):
     with pytest.raises(TypeError):
-        function(argument)
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore')
```
StanczakDominik: Hmm. What's the goal here? Perhaps this could be simply replaced with two layers of `pytest.raises` and `pytest.warns`?

plasmapy/atomic/nuclear.py
lemmatum: I've never seen `from` being used for anything other than imports. This looks weird.

plasmapy/atomic/nuclear.py
```diff
@@ -32,28 +33,28 @@ def nuclear_binding_energy(argument, mass_numb=None):
...
     Examples
     --------
-    >>> nuclear_binding_energy('Fe-56')
-    <Quantity 7.674003137600576e-11 J>
+    >>> from astropy import units as u
+    >>> nuclear_binding_energy('Fe-56').to(u.MeV)
+    <Quantity 492.25957875689096 MeV>
     >>> nuclear_binding_energy(26, 56)
-    <Quantity 7.674003137600576e-11 J>
+    <Quantity 7.88686788449147e-11 J>
```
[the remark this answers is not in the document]
namurphy: When calculating the binding energy before, I used the mass of the isotope rather than the mass of the nucleus.  That difference is due to the mass of the electrons.

plasmapy/atomic/nuclear.py
namurphy: I just learned this!  If you run this code snippet...
```Python
try:
    raise TypeError
except Exception as e:
    raise ValueError from e
```
then you get the following
```
<ipython-input-5-42a31f7f16fe> in <module>()
      1 try:
----> 2     raise TypeError
      3 except Exception as e:

TypeError: 

The above exception was the direct cause of the following exception:

<ipython-input-5-42a31f7f16fe> in <module>()
      2     raise TypeError
      3 except Exception as e:
----> 4     raise ValueError from e
      5 

ValueError: 
```
Without the `as ex` and `from ex`, the 

```The above exception was the direct cause of the following exception:```

would instead be

```During handling of the above exception, another exception occurred:```

This is a way to help with tracebacks by keeping track of the root causes of exceptions.

File: plasmapy/atomic/atomic.py [hunk 8]
```diff
@@ -1560,6 +1563,38 @@ def _is_antiproton(argument):
         return False
 
 
+def _is_antineutron(argument):
+    r"""Returns True if the argument corresponds to an antineutron, and
+    False otherwise."""
+
+    if argument.lower() == 'antineutron':
```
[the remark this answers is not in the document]
namurphy: Good catch!  This also raises an exception is `argument` is not a string, so perhaps:

```Python
return isinstance(argument, str) and argument.lower() == 'antineutron'
```

plasmapy/atomic/tests/test_atomic.py
namurphy: This is a remnant from the rabbit hole I went down 🐰 🕳 , so I removed it.  (I actually went down a rabbit hole in real life once, so that might be where I got some of my coding habits, perhaps?)

plasmapy/atomic/nuclear.py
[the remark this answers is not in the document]
StanczakDominik: Ah, I see. That's cool, then :)

plasmapy/atomic/atomic.py
```diff
@@ -1560,6 +1563,38 @@ def _is_antiproton(argument):
...
+    charge state is given."""
+
+    try:
+
+        isotope = isotope_symbol(argument, mass_numb)
+
+        if Z is None:
+            Z = charge_state(argument)
+
+        if isotope == 'H-1' and Z == 1:
```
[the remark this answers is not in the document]
namurphy: I think I made it less bad. 😹

plasmapy/atomic/nuclear.py
StanczakDominik: Well, that's pretty darn cool, then!

namurphy:
I'm merging this in, with remaining work to be addressed in #174, #178, and #179.

PR #170, opened 2017-11-23
Issue #168

### LINKED ISSUE #168: Create physics exceptions and warnings
Several of the exceptions and warnings that come up in PlasmaPy result from when we try to violate the laws of physics.  Most of the time these violations cause a `UserWarning` or raise a `ValueError`, but these are pretty generic and don't give much insight into the cause of the problem.  To help with exception handling, I propose that we create new exceptions and warnings, such as `PhysicsError` and `PhysicsWarning`.  We could additionally be more specific and also have `RelativityError` and/or `RelativityWarning`.  One possibility would be to put these in a new file called `plasmapy/utils/exceptions.py`.

This would be a good first contribution for someone who is somewhat familiar with Python.  Here's documentation on [user-defined exceptions](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions.  Thank you!

samurai688:
Hi all!

This is the first contribution to a github / open source project I've attempted, so I'm sure I've messed some stuff up... :) let me know and I'll try to fix my mistakes!

I've attempted to address issue #168, re custom Warnings and Exceptions:
- added utils/exceptions.py with some new custom-named exceptions and warnings
- changed checks to raise the new warnings in the warnings.warn soft fashion rather than raise UserWarning as an Exception
- parameters/ion_sound_speed to raise the new PhysicsError for gamma < 1
- hopefully updated all of the tests to catch all of the assorted things that are now the new types

Closes #168


StanczakDominik:
Hey, thanks a bunch for this one! I'll take a closer look at this soon (within a few hours).

plasmapy/physics/parameters.py
```diff
@@ -228,6 +228,9 @@ def ion_sound_speed(*ignore, T_e=0*units.K, T_i=0*units.K,
 
     ValueError
         If the ion mass, adiabatic index, or temperature are invalid.
+        
+    PhysicsError
+        If an adiabtic index is less than one.
```
jasperbeckers: Typo on "adiabtic"

StanczakDominik:
Hey, I'm terribly sorry I haven't gotten to this yet, got kind of caught in debugging my thesis code (bloody C++) :( If someone else could review this PR, I would be very grateful, as that debacle has still not finished :(

plasmapy/utils/tests/test_checks.py
```diff
@@ -224,7 +224,7 @@ def func(x, y, z=10*u.eV):
 ]
 
 # (speed, betafrac, error)
-relativisitc_error_examples = [
+relativistic_error_examples = [
```
lemmatum: hah! Good catch!

plasmapy/utils/tests/test_checks.py
```diff
@@ -252,6 +255,12 @@ def test__check_relativisitc_valid(speed, betafrac):
 def test__check_relativistic_errors(speed, betafrac, error):
     with pytest.raises(error):
         _check_relativistic(speed, 'f', betafrac=betafrac)
+        
+        
+@pytest.mark.parametrize("speed, betafrac, warning", relativistic_warning_examples)
+def test__check_relativistic_warnings(speed, betafrac, warning):
+    with pytest.warns(warning):
+        _check_relativistic(speed, 'f', betafrac=betafrac)
```
lemmatum: What was your reasoning for passing the type of warning as an argument? Is this just for testing or is it part of the custom warning structure?

plasmapy/utils/exceptions.py
```diff
@@ -5,6 +5,14 @@
...
 """
 
+class PhysicsError(Exception):
+    """Error for using a deeply troubling physics value"""
+    pass
+
+class PhysicsWarning(Warning):
+    """Warning for using a mildly worrisome physics value"""
+    pass
+
```
lemmatum: I like these descriptions and the idea of a program being deeply worried by your physics 😛 
I'd like to see a more precise docstring describing when PhysicsWarning is to be used. Does one use it directly or is it just a class to inherit from when generating more specific warnings?

plasmapy/utils/exceptions.py
```diff
@@ -6,13 +6,13 @@
...
     pass
 
 class PhysicsWarning(Warning):
     """Warning for using a mildly worrisome physics value"""
     pass
 
-class RelativityWarning(Warning):
+class RelativityWarning(PhysicsWarning):
     """Warning for approaching or exceeding the speed of light"""
     pass
```
lemmatum: fancy inheritance! Is the idea that PhysicsWarning is the general class and more specific warnings will inherit from it? Do you forsee any cases where PhysicsWarning itself would be used outside of inheritance? If so, then you might need a more precise description of the use cases in the dosctring for PhysicsWarning.

plasmapy/physics/parameters.py
```diff
@@ -230,7 +230,7 @@ def ion_sound_speed(*ignore, T_e=0*units.K, T_i=0*units.K,
         If the ion mass, adiabatic index, or temperature are invalid.
         
     PhysicsError
-        If an adiabtic index is less than one.
+        If an adiabatic index is less than one.
```
lemmatum: Good catch!

plasmapy/utils/exceptions.py
namurphy: I'd also suggest extending the docstring.  I have never documented an exception before (or even written one!) so I looked up an [example in the numpydoc format](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html):

```Python
class ExampleError(Exception):
    """Exceptions are documented in the same way as classes.

    The __init__ method may be documented in either the class level
    docstring, or as a docstring on the __init__ method itself.

    Either form is acceptable, but the two should not be mixed. Choose one
    convention to document the __init__ method and be consistent with it.

    Note
    ----
    Do not include the `self` parameter in the ``Parameters`` section.

    Parameters
    ----------
    msg : str
        Human readable string describing the exception.
    code : :obj:`int`, optional
        Numeric error code.

    Attributes
    ----------
    msg : str
        Human readable string describing the exception.
    code : int
        Numeric error code.

    """

    def __init__(self, msg, code):
        self.msg = msg
        self.code = code
```

plasmapy/utils/tests/test_checks.py
namurphy: Oh, this is a good thing to bring up.  We've been having the warning or exception be an argument for cases where we are testing multiple different warnings or exceptions in one test.  Since this is only testing one warning, we could probably change `with pytest.warns(warning)` to `with pytest.warns(RelativityWarning)` and then not need to have `warning` as an argument to `test__check_relativistic_warnings`.  

We've refactored the tests quite a bit lately, so I'm wondering if we've been having the warning or exception as an argument when only one warning or exception is being tested.  I do really like `pytest.mark.parametrize` even though most of the time I misspell it as `parameterize`!

plasmapy/utils/checks.py
```diff
@@ -285,7 +287,7 @@ def _check_quantity(arg, argname, funcname, units, can_be_negative=True,
 
 
 def _check_relativistic(V, funcname, betafrac=0.1):
-    r"""Raise UserWarnings if a velocity is relativistic or superrelativistic
+    r"""Warn if a velocity is relativistic or superrelativistic
```
namurphy: Thank you for fixing this!

samurai688:
Hi all,

Thank you for the review comments!  I will try to make the requested changes shortly.

I was reading and thinking a bit more about exceptions and I've come across a few concepts that I think we might want to implement.

- In the Python 3.6 documentation, it mentions, "When creating a module that can raise several distinct errors, a common practice is to create a base class for exceptions defined by that module, and subclass that to create specific exception classes for different error conditions."  Something similar is mentioned in [this fairly recent blog post](https://julien.danjou.info/blog/2016/python-exceptions-guide) about python exceptions.  The idea is, this would make it easy for user/application code to catch "any exception raised by plasmapy" by simply catching PlasmaPyException, or whatever we would call it. All our more specific exceptions would subclass this.

- There's the philosophical question of when to raise an exception versus when to raise a milder warning from the warnings module. My current thinking is, we should raise an exception when the outside user/application code has given plasmapy a value that it can't handle, i.e., they're asking plasmapy to tell them the stopping length of an ion traveling faster than the speed of light. I think there could be some gray area though, for instance, I think I would favor a policy of being lenient to the outside world if plasmapy has an internal issue. For example, our current Alfven speed function will happily calculate Alfven speeds greater than the speed of light. If the user/application code asks us for an Alfven speed of some plasma, and our non-relativistic function thinks it's 2c, it might be jarring to return an exception. Rather we could return 2c along with a warning that the plasmapy function is being used out of its range of expected validity. Then if the world persists and sends the superluminal speed back into another plasmapy function, we could raise the RelativityError.  

Alternatively, we could be more strict, and say, world, you are venturing into bad places where the current plasmapy is not valid, we are going to raise an exception. Don't say we didn't do everything we could to try to save you.

- Oh right, and I think we should have a RelativityError for input speeds > c, which I haven't yet included.

What do you guys think?

namurphy:
The first point you made about having a base exception for a module and then subclassing it sounds like a good idea.  I support us going ahead and doing that.  A possibility would be `PlasmaPyError` being the base exception, and then `PhysicsError` being a subclass of `PlasmaPyError`, and then `RelativityError` being a subclass of `PhysicsError` (and maybe the same with warnings).

Right now I think I'm inclined for a `RelativityError` to be raised for any superluminal velocities, and a `RelativityWarning` to occur when a non-relativistic formula is used in a situation where relativistic effects are important.  I'm not sure what the best practice would be, though.

Thank you again!

namurphy:
This is getting really close!  Is there much more that needs to be done before we accept the pull request?

samurai688:
As it stands now, it's got the warnings, the physics and relativity exceptions, and a module for holding more custom exceptions, all of which I think are good things.

It has not implemented the idea of "all exceptions raised by plasmapy will subclass PlasmaPyError", that is, the code still raises lots of vanilla ValueErrors and such. 

I've come to second-guess whether the above is a good idea. We could define "PlasmaPyValueError" and so on, and then teach ourselves to always raise our custom exceptions instead of the default, but is the payoff worth it? Would we or anyone ever really use it, or just make ourselves learn a bunch of new names?

I guess on the whole I'm leaning towards not creating new exceptions where they don't add any descriptive value over the defaults, and giving up on the idea of having a plasmapy-specific subclass for truly all exceptions. We could keep the PlasmaPyError base class that we've got and use it as a subclass just for our own custom exceptions. This is basically how it is now.

If folks agree with that and/or pending any other comments, I think it's ready to go.

namurphy:
I agree with what you said.  The two reasons why I've been hoping for these new warnings and exceptions are that (1) it provides more descriptive and useful information, and (2) it makes it easier for one of our methods to raise different exceptions for different problems that can then be handled differently, instead of having each problem raise a `ValueError`.  I don't think there's a need for a global PlasmaPy exception, especially given the points you just made.  

If nobody else has any feedback, then we can pull this in soon.  Again, thank you!

lemmatum:
So in this new scheme where are vanilla ValueErrors raised and where are PhysicsErrors or PlasmaPyErrors raised?

I'm not sure what you mean by the idea "all exceptions raised by plasmapy will subclass PlasmaPyError". Is this referring to making our custom physics errors subclass PlasmaPyError or PhysicsError, or does this refer to actually swapping out all errors in the PlasmaPy package with PlasmaPyError so that we know that the error came from our package as opposed to some other package?

samurai688:
@lemmatum, as of right now, the former.  Such that, the PlasmaPyError class would be a base class for our custom exceptions only.  

We had talked about doing the latter, where indeed we would have every exception raised be a custom exception that subclasses PlasmaPyError, but I was thinking that it would give too little benefit, and have the downside that everyone who works on the project in the future would have to learn to raise all these PPValueError instead of ValueError, etc. So currently I'm not advocating that.

lemmatum:
Congrats on your first merged pull request @samurai688 ! 🎆 🎉

samurai688:
Woo!! Thanks all! Thanks all for welcoming me into this process, I'm super excited about this project! :zap: :zap:

PR #171, opened 2017-12-02
Kappa function

lemmatum:
Adding 1D and 3D kappa velocity distribution functions.
Addresses #101 

There are still some issues with the 1D distribution in particular not integrating to 1. Need to figure out the correct normalization constant for that function. The other 1D tests are failing because I haven't added the expected values yet. I don't want to do this until we figure out the correct normalization constant as the expected values will change.
There is another issue that for large values of kappa both the 1D and 3D functions return NaNs. This makes it difficult to test if the kappa distribution converges to a Maxwellian as kappa approaches infinity.

Help on these would be appreciated!

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -429,7 +429,7 @@ def thermal_speed(T, particle="e", method="most_probable"):
 @check_quantity({
     'T': {'units': units.K, 'can_be_negative': False}
 })
-def kappa_thermal_speed(T, kappa, particle="e"):
+def kappa_thermal_speed(T, kappa, particle="e", method="most_probable"):
     r"""Returns the most probable speed for a particle within a Kappa
     distribution.
 
@@ -500,12 +500,22 @@ def kappa_thermal_speed(T, kappa, particle="e"):
     except Exception:
         raise ValueError("Unable to find {} mass in thermal_speed"
                          .format(particle))
-    # thermal velocity of Kappa distribution function
-    coeff = (kappa - 3/2) / kappa
-    vTh = np.sqrt(coeff) * thermal_speed(T=T, 
-                                         particle=particle,
-                                         method="most_probable")
-    return vTh
+    # thermal velocity of Kappa distribution function is just Maxwellian
+    # thermal speed modulated by the following factor
+    coeff = np.sqrt((kappa - 3/2) / kappa)
+    
+    # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
+    if method == "most_probable":
+        vTh = (np.sqrt(2*k_B*T/m)).to(units.m/units.s)
+    elif method == "rms":
+        vTh = (np.sqrt(3*k_B*T/m)).to(units.m/units.s)
+    elif method == "mean_magnitude":
+        vTh = (np.sqrt(8*k_B*T/(m/np.pi))).to(units.m/units.s)
+    else:
+        raise(ValueError("Method {} not supported in thermal_speed"
+                         .format(method)))
+    
+    return coeff*vTh
 
 @check_quantity({
     'B': {'units': units.T}
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -158,7 +158,7 @@ def Alfven_speed(B, density, ion="p"):
 
     _check_quantity(B, 'B', 'Alfven_speed', units.T)
     _check_quantity(density, 'density', 'Alfven_speed',
-                    [units.m**-3, units.kg/units.m**3], can_be_negative=False)
+                    [units.m**-3, units.kg / units.m**3], can_be_negative=False)
 
     B = B.to(units.T)
     density = density.si
@@ -172,13 +172,13 @@ def Alfven_speed(B, density, ion="p"):
                 Z = 1
         except Exception:
             raise ValueError("Invalid ion in Alfven_speed.")
-        rho = density*m_i + Z*density*m_e
+        rho = density * m_i + Z * density * m_e
 
-    elif density.unit == units.kg/units.m**3:
+    elif density.unit == units.kg / units.m**3:
         rho = density
 
     try:
-        V_A = (np.abs(B)/np.sqrt(mu0*rho)).to(units.m/units.s)
+        V_A = (np.abs(B) / np.sqrt(mu0 * rho)).to(units.m / units.s)
     except Exception:
         raise ValueError("Unable to find Alfven speed")
 
@@ -190,7 +190,7 @@ def Alfven_speed(B, density, ion="p"):
     'T_i': {'units': units.K, 'can_be_negative': False},
     'T_e': {'units': units.K, 'can_be_negative': False}
 })
-def ion_sound_speed(*ignore, T_e=0*units.K, T_i=0*units.K,
+def ion_sound_speed(*ignore, T_e=0 * units.K, T_i=0 * units.K,
                     gamma_e=1, gamma_i=3, ion='p'):
     r"""Returns the ion sound speed for an electron-ion plasma.
 
@@ -318,8 +318,8 @@ def ion_sound_speed(*ignore, T_e=0*units.K, T_i=0*units.K,
     T_e = T_e.to(units.K, equivalencies=units.temperature_energy())
 
     try:
-        V_S_squared = (gamma_e*Z*k_B*T_e + gamma_i*k_B*T_i)/m_i
-        V_S = np.sqrt(V_S_squared).to(units.m/units.s)
+        V_S_squared = (gamma_e * Z * k_B * T_e + gamma_i * k_B * T_i) / m_i
+        V_S = np.sqrt(V_S_squared).to(units.m / units.s)
     except Exception:
         raise ValueError("Unable to find ion sound speed.")
 
@@ -414,16 +414,17 @@ def thermal_speed(T, particle="e", method="most_probable"):
 
     # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
     if method == "most_probable":
-        V = (np.sqrt(2*k_B*T/m)).to(units.m/units.s)
+        V = (np.sqrt(2 * k_B * T / m)).to(units.m / units.s)
     elif method == "rms":
-        V = (np.sqrt(3*k_B*T/m)).to(units.m/units.s)
+        V = (np.sqrt(3 * k_B * T / m)).to(units.m / units.s)
     elif method == "mean_magnitude":
-        V = (np.sqrt(8*k_B*T/(m/np.pi))).to(units.m/units.s)
+        V = (np.sqrt(8 * k_B * T / (m / np.pi))).to(units.m / units.s)
     else:
         raise(ValueError("Method {} not supported in thermal_speed"
                          .format(method)))
 
     return V
+
 
 @check_relativistic
 @check_quantity({
@@ -437,7 +438,7 @@ def kappa_thermal_speed(T, kappa, particle="e", method="most_probable"):
     ----------
     T : Quantity
         The particle temperature in either kelvin or energy per particle
-        
+
     kappa: Quantity
         The kappa parameter is a dimensionless number which sets the slope
         of the energy spectrum of suprathermal particles forming the tail
@@ -492,7 +493,7 @@ def kappa_thermal_speed(T, kappa, particle="e", method="most_probable"):
     # Checking thermal units
     T = T.to(units.K, equivalencies=units.temperature_energy())
     # must have kappa > 3/2 for distribution function to be valid
-    if kappa <= 3/2:
+    if kappa <= 3 / 2:
         raise ValueError(f"Must have kappa > 3/2, instead of {kappa}.")
     # obtaining particle mass
     try:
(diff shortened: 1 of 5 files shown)


lemmatum:
Found a paper with a 1D version of the kappa velocity distribution function. All the tests should now be passing.

plasmapy/physics/parameters.py
```diff
@@ -399,18 +399,112 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> kappa_thermal_speed(5*u.eV, 4, 'p')
+    <Quantity 24467.878463594967 m / s>
+    """
+    # Checking thermal units
+    T = T.to(units.K, equivalencies=units.temperature_energy())
+    # must have kappa > 3/2 for distribution function to be valid
```
namurphy: Lately I've been trying to follow the advice laid out in the paper on [_Best Practices for Scientific Computing_  by Wilson et al. (2014)](http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1001745).  One of their suggestions is to: "Refactor code in preference to explaining how it works."  A possibility here would be to include the information in this comment in the ValueError message, and then delete the comment.  That way the code would be more self-documenting.

plasmapy/physics/parameters.py
```diff
@@ -399,18 +399,112 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
+    if kappa <= 3 / 2:
+        raise ValueError(f"Must have kappa > 3/2, instead of {kappa}.")
+    # obtaining particle mass
+    try:
+        m = ion_mass(particle)
+    except Exception:
+        raise ValueError("Unable to find {} mass in thermal_speed"
```
namurphy: Another place for an f-string!  Something like...
```Python
        raise ValueError(f"Unable to find {particle} mass in thermal_speed")
```

plasmapy/physics/parameters.py
```diff
@@ -399,18 +399,112 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
-        V = (np.sqrt(2*k_B*T/m)).to(units.m/units.s)
+        V = (np.sqrt(2 * k_B * T / m)).to(units.m / units.s)
     elif method == "rms":
-        V = (np.sqrt(3*k_B*T/m)).to(units.m/units.s)
+        V = (np.sqrt(3 * k_B * T / m)).to(units.m / units.s)
     elif method == "mean_magnitude":
-        V = (np.sqrt(8*k_B*T/(m/np.pi))).to(units.m/units.s)
+        V = (np.sqrt(8 * k_B * T / (m / np.pi))).to(units.m / units.s)
     else:
         raise(ValueError("Method {} not supported in thermal_speed"
```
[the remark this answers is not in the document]
StanczakDominik: Hahaha, sure we can, but it's probably not strictly necessary. I do hype them up a bunch :laughing:

plasmapy/physics/parameters.py
```diff
@@ -399,18 +399,112 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
                          .format(method)))
 
     return V
 
 
+@check_relativistic
+@check_quantity({
+    'T': {'units': units.K, 'can_be_negative': False}
+})
+def kappa_thermal_speed(T, kappa, particle="e", method="most_probable"):
```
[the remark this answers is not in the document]
StanczakDominik: I like the option with the warning, depending on how often the kappa distribution (heard nothing about it) is needed for electrons.

StanczakDominik:
I'll look at this ASAP, which means Soon(TM).

plasmapy/physics/parameters.py
```diff
@@ -162,13 +162,13 @@ def Alfven_speed(B, density, ion="p"):
                 Z = 1
         except Exception:
             raise ValueError("Invalid ion in Alfven_speed.")
-        rho = density*m_i + Z*density*m_e
+        rho = density * m_i + Z * density * m_e
```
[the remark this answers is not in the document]
lemmatum: Yes, please! Spaces are way more readable! I've been trying to slowly introduce this change 👿

plasmapy/physics/parameters.py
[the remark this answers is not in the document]
lemmatum: I like f-string, will change it.

plasmapy/physics/parameters.py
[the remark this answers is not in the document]
lemmatum: I second the warning method. Is PhysicsWarning ready yet?

plasmapy/physics/parameters.py
lemmatum: I can see that being cleaner here, though I don't know if I would agree with that sentiment in general.

plasmapy/physics/parameters.py
lemmatum: changed.

plasmapy/physics/parameters.py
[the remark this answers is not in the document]
StanczakDominik: That it does. Actually, shouldn't AutoPEP8 mention lines not having spaces between operators?

plasmapy/physics/parameters.py
```diff
@@ -394,23 +394,113 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
+    UserWarning
+        If the particle thermal speed exceeds 10% of the speed of light, or
+        if units are not provided and SI units are assumed.
+
+    Notes
+    -----
+    The particle thermal speed is given by:
+
+    .. math::
+        V_{th,i} = \sqrt{(2 \kappa - 3)\frac{2 k_B T_i}{\kappa m_i}}
```
StanczakDominik: This part definitely needs to show how the formula changes depending on the method chosen. I'm just reading the code for now, but I'll try to dig into reading up on that distribution soonish.

plasmapy/physics/parameters.py
```diff
@@ -394,23 +394,113 @@ def thermal_speed(T, particle="e", method="most_probable"):
...
+    # obtaining particle mass
+    try:
+        m = ion_mass(particle)
+    except Exception:
+        raise ValueError(f"Unable to find {particle} mass in thermal_speed")
+    # thermal velocity of Kappa distribution function is just Maxwellian
+    # thermal speed modulated by the following factor.
+    # This is true for the "most probable" velocity, though it may change
+    # for the other two methods. Must be checked!
+    coeff = np.sqrt((kappa - 3 / 2) / kappa)
```
StanczakDominik: Note to self to remember to check this!

plasmapy/physics/tests/test_distribution.py
```diff
@@ -635,7 +692,467 @@ def test_value_drift_units(self):
...
                           rtol=1e-8,
-                          atol=0.0), errStr
\ No newline at end of file
+                          atol=0.0), errStr
+
+
+#%% kappa
+
+# test class for kappa_velocity_1D function:
+class Test_kappa_velocity_1D(object):
```
StanczakDominik: As an idea, perhaps one day we could just have a `Plasma` or `PlasmaBlob` instance for this kind of test?

lemmatum: I'm not sure how that would work, especially as tests should thoroughly check edge cases. What do you envision?

plasmapy/physics/parameters.py
lemmatum: yes, please 😅

plasmapy/physics/parameters.py
lemmatum: I think this was mentioned in a different pull request, but do multi-line latex equations get properly rendered in the documentation?

plasmapy/physics/parameters.py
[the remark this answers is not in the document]
lemmatum: My current thinking is we should make this a new issue, default to electrons, and then implement the warning using a physics specific warning based on #168  and #170  once that pull request is completed. Then make sure such defaults are implemented throughout plasmapy (as we assumes particles in many functions). Thoughts?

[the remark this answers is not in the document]
namurphy: Creating another issue for this does sound like the best idea since it is much broader than this pull request.  A possibility for the moment would be to put in a `UserWarning` that can be changed to a `PhysicsWarning` or something more specific later on, if you'd like.  But, it'd be fine with me to wait to resolve this with the later issue.

plasmapy/physics/parameters.py
```diff
@@ -394,8 +394,7 @@ def thermal_speed(T, particle="e", method="most_probable"):
     try:
         m = ion_mass(particle)
     except Exception:
-        raise ValueError("Unable to find {} mass in thermal_speed"
-                         .format(particle))
+        raise ValueError("Unable to find {particle} mass in thermal_speed")
```
namurphy: A quick suggestion before my cat sits on my laptop.  There should be an `f` before the first quotation mark to note that it's an f-string here, ten lines below, and then around line ~500.  (`f"Unable to find {particle}..."`)

StanczakDominik: Oh shit, yeah, the only way this is passing right now is because we're probably not testing ValueError here.

Aren't we?

lemmatum: @namurphy good catch!

@StanczakDominik 
Nah, it probably just doesn't render the value and instead literally prints `{particle}`

StanczakDominik: That it does, I just checked.

plasmapy/physics/parameters.py
lemmatum: So, I'm not sure how to check this. I've tried analytically integrating the 1D distribution function multiplied by v^2 to get the RMS speed, but I get an odd answer. Seems to result in RMS speed = sqrt(k_B * T / m), which is precisely the RMS speed for a 1D Maxwell-Boltzmann distribution.

Then again, this might make sense since the most probable thermal velocity of the kappa distribution is in the distribution itself and ends up canceling all the kappas from the integration.

This is messing with my intuition as I would think that the high energy tails would shift the mean by some factor of kappa to high velocities compared to the Maxwellian distribution, and that it would only converge to the Maxwellian RMS speed for large values of kappa.

lemmatum: There seems to be a similar sort of cancellation in 3D as well (as far as the constants and thermal velocities are concerned) though I haven't managed to do the triple integral as it is quite complex and not easily separable into x,y,z components like the Maxwellian.

lemmatum:
I just tried to rebase with upstream master. Hopefully no major errors come up! 🤞

plasmapy/physics/parameters.py
lemmatum: @namurphy thoughts on how to check RMS and mean magnitude speeds for the kappa distribution function?

Perhaps this can be numerically checked as we have 2 hypotheses as to what the speed should be? (i.e., how it is currently defined in kappa thermal speed vs. the analytical approach which seems to yield the Maxwellian thermal speeds for RMS and mean magnitude)

lemmatum:
What is the new way of writing `H-1`? I am getting a warning from the rebase.
`ValueError: Invalid particle H-1 in inertial_length.`

StanczakDominik:
Paging @namurphy for the latest comment.

namurphy:
The quick fix would be to use `'p'` or `'p+'` for a proton.

A difficulty and ambiguity that I've been running into is that isotopes do not have charge information.    For hydrogen, there's less ambiguity if we know the particle has to be charged.  For heavier elements, we wouldn't be able to tell if `'He-4'` corresponds to singly ionized or doubly ionized helium.  For the moment we've been assuming singly ionized species, but "explicit is better than implicit" so I'm thinking that we should require charge information be given.

namurphy:
Also, my apologies for changing the application programming interface (API)!  Unfortunately, the API for the atomic package is still somewhat unstable...

StanczakDominik:
Don't worry about changing the API, that's completely to be expected in this early age. It should worry us if it wasn't changing enough :smile: 

This is probably not the best place for this, so I apologize, but: maybe we're actually approaching this from the completely wrong angle. There's a finite amount of space in a string you add information into  to keep it coherent as an information about a type of particle. Sooner or later, if we want to cover all possible angles, we're going to either run out or end up with a complete mess.

Maybe the way to go here is to have a particle class. Have it keep an `atomic_number` (instanced from our current tables), an `atomic_number` (so this covers isotopes), and an `ionization_state` (this covers charge).

This may be completely missing the point on the issues of the `atomic` subpackage, but I think it's probably better to raise ideas and have them judged as irrelevant than to miss opportunities by staying silent.

lemmatum:
I agree with @StanczakDominik , I am fine with API changes at this stage. Just want to get the right particle in so that Travis and coveralls check my code 😉 

Btw, have either of you taken a stab at the RMS and mean magnitude speed for the kappa distribution?

plasmapy/physics/parameters.py
lemmatum: Similar test for mean magnitude
```
Numerically obtained vmean: 14850.474638693995
Functionally obtained vmean: 12346.067837955574 m / s
Maxwellian vmean: 15616.677805956093 m / s
```

It's clear that the way the kappa_thermal_speed is currently defined is wrong. The Maxwellian mean magnitude and the integration method are fairly close. I think the integral is off because there is a lot of weight in the tails of a kappa distribution and those parts become hard to integrate due to numerical errors.

So yea looks like the RMS and mean speeds for kappa distribution are equivalent to that of a Maxwellian.

side note: found an error in how mean magnitude velocity was defined in thermal_speed(), so I'll commit the correction.

lemmatum:
@StanczakDominik @namurphy if y'all are happy with my numerical investigations of the RMS and mean kappa thermal speeds then I believe this is ready for a merge.

StanczakDominik:
Hey, I'm sorry I still haven't been able to get into this. I tried googling, but it turns out Wikipedia is unaware of this distribution and basically the only thing that seems relevant is [this article on Arxiv](https://arxiv.org/abs/1003.3532). I'll be reading this ASAP, for what it's worth.

@namurphy has a colleague who's apparently familiar with kappa distributions, but they're at a conference this week. We thus have two options - merge this now and possibly make an issue saying "Verify physical sense of Kappa function implementation", merge this later once we ourselves dig into the docs on the distribution, or merge this in a week or so once @namurphy can ask if this is reasonable. Three options. Whatcha think?

By the way, since comments are hard to find and you've put a very reasonable discussion on testing this in there, I'll be linking [that discussion](https://github.com/PlasmaPy/PlasmaPy/pull/171#discussion_r156555339) from here. :)

lemmatum:
@StanczakDominik Here are a few more articles which I found to be helpful while writing these functions up:
* [NASA article](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040161173.pdf) on kappa distributions and dispersion 
* [Effects of a kappa distribution function of electrons on incoherent scatter spectra (2000)](https://link.springer.com/article/10.1007%2Fs00585-000-1216-2?LI=true) - This is where I got the 1D velocity distribution function, eq(13)
* And then there a few articles referenced by the Arxiv article you mentioned, which talk about how to extend definitions of thermodynamics (things like temperature) for suprathermal distributions. These seem to mostly be based off of Tsallis's work.

I'm leaning towards the merge now, fix later approach. Tests are giving reasonable results and I pulled equations directly from scientific articles for the distributions. I think the only uncertainty is the mean magnitude kappa thermal speed. Though it makes sense to me that if we didn't have these numerical errors at the tails of the distribution then the mean magnitude should be the same as the Maxwellian mean magnitude (especially since the RMS speeds are the same).

Good point, thanks for linking those tests!

StanczakDominik:
Awesome, then - I'll merge this PR. Do you think you could write up a small issue (mostly a bunch of links) on verifying this? You're the one best acquainted with this.

I'll be reading that article I linked first, and then I'll try to move into the ones you used (thanks for sharing those!). Something something cross-entropy decreases best with multiple data sources etc etc. :smile:

lemmatum:
Done, see #186 for new issue.

StanczakDominik:
Thank you!

PR #172, opened 2017-12-05
Test against astropy dev

SolarDrew:
Added a parameter in .travis.yml so that Travis will build and test against the development version of astropy as well as the released version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: python
 python:
         - "3.6"
 
+env:
+        - ASTROPY_VERSION='dev'
+
 git:
     submodules: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ language: python
 python:
         - "3.6"
 
-env:
-        - ASTROPY_VERSION='dev'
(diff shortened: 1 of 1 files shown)


StanczakDominik:
As it stands we're downloading conda 4.3.30, downgrading it to 4.3.27, installing three separate bunches of packages via pip only for Astropy to scream about `ImportError: Numpy version 1.10.0 or later must be installed to use Astropy`. I have no idea what's happening here and why, but it doesn't quite look good.

.travis.yml
```diff
@@ -3,6 +3,10 @@ language: python
 python:
         - "3.6"
 
+matrix:
+    include:
+        - env: ASTROPY_VERSION='dev'
```
dstansby: I'm guessing you need another matrix entry for non-dev astropy?

dstansby: Might be worth explicitly selecting python 3.6 here too

namurphy:
I'm going to tentatively close this PR since we haven't worked on it in a long time, but we can re-open it later if we want to finish figuring out how to test against astropy dev.

Issue #178, opened 2017-12-11
Add new syntax to represent isotopes?


namurphy:
Current, isotopes are represented with strings like `'H-1'`, `'D'`, and `'helium-4'`.  In #169 it was brought up that isotopes can also be represented as something like `'6Li'` in addition to `'Li-6'`.  This can be mostly incorporated by letting `isotope_symbol` in `atomic.py` take in something like `'6Li'` and returning `'Li-6'`.  

One place where this will end up being a bit more complicated is in `nuclear_reaction_energy` since that function currently allows that format where the `'6'` in `'6Li'` would be treated as a multiplicative factor (e.g., six lithium atoms).  The way to resolve this would be by changing the multiplier format to require an asterisk: `'2*Li-6'` or `'2*6Li'`.  There will probably be a few other places that I'm forgetting too.

Thank you!

lemmatum:
`6Li` and `6*Li` look very similar to me and it is not immediately obvious that one refers to an isotope while the other refers to 6 Lithium ions. It is also quite fragile to typos.
Is there are clearer way to distinguish the two, perhaps by forcing `element-number` for isotopes and `number*element` for reaction quantities?

PR #180, opened 2017-12-11
Require an asterisk for multiplicative factors in nuclear_reaction_energy

namurphy:
Up until now, `nuclear_reaction_energy` allowed inputs such as:

```Python
nuclear_reaction_energy('Be-8 --> 2He-4')
```

where the `2` in the above code snippet has been treated as a multiplicative factor.  However, an idea has been brought up in #169 and #178 to allow a different format for isotopes where the mass number precedes the atomic symbol (e.g., `'4He'` would be the same as `'He-4'`).  If we do implement that idea, then there would have been ambiguity between mass numbers and multiplicative factors (e.g., would `'4He'` be one `'He-4'` isotope or four `'He'` atoms?).  This pull request changes the required format to

```Python
nuclear_reaction_energy('Be-8 -> 2*He-4')
```
to reduce the chance for ambiguity.  We're still discussing whether or not to allow isotopes to be represented like `'4He'`.

### DIFF
--- a/plasmapy/atomic/nuclear.py
+++ b/plasmapy/atomic/nuclear.py
@@ -2,9 +2,10 @@
 
 from astropy import units, constants
 import re
+from itertools import repeat
 from .atomic import (isotope_symbol, mass_number, isotope_mass, ion_mass,
                      atomic_number, charge_state, _is_neutron, _is_electron,
-                     _is_positron, _is_antiproton)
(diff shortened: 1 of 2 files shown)


namurphy:
> Coverage decreased (-0.001%) to 99.582%

Well, at least I have an excuse to use GitHub emoji...  🙀 📉 😔 🚳

PR #192, opened 2017-12-15
attempt at shortening a bunch of tests in Species

StanczakDominik:
As of now, PlasmaPy tests alone (I'm not counting the setup here) take 126.58 seconds on Travis. I'm going to see if I can trim some fat from tests in the Species submodule, which is currently pretty bad on performance.

This is heavily work in progress right now, I just don't feel like setting benchmarks up locally.

### DIFF
--- a/plasmapy/classes/species.py
+++ b/plasmapy/classes/species.py
@@ -59,7 +59,7 @@ class Species:
     def __init__(self, plasma, particle_type='p', n_particles=1, scaling=1,
                  dt=np.inf * u.s, nt=np.inf):
 
-        if np.isinf(dt) and np.isinf(nt):
+        if np.isinf(dt) and np.isinf(nt):  # coveralls: ignore
             raise ValueError("Both dt and nt are infinite.")
 
         self.q = atomic.charge_state(particle_type) * constants.e.si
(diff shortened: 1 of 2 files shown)


StanczakDominik:
Down to 42.96s on the current version.

StanczakDominik:
So yeah, I'm down to around 42.5 seconds on all tests. With `test_species` completely removed, this goes down to 27.14s. If sacrificing 15 seconds of additional testing time for regression testing on a submodule intended to undergo a complete rewrite once I regain some faith in the PIC method after burning out on it for my thesis sounds reasonable (this is, note, a speedup by 84s over current master), feel free to merge this!

plasmapy/classes/species.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, plasma, particle_type='p', n_particles=1, scaling=1,
                  dt=np.inf * u.s, nt=np.inf):
 
         if np.isinf(dt) and np.isinf(nt):
-            raise ValueError("Both dt and nt are infinite.")
+            raise ValueError("Both dt and nt are infinite.") # coveralls: ignore
```
namurphy: A minor, 100% optional stylistic point...   I usually find the code to look a bit cleaner if the `  # coveralls: ignore` is after the conditional statement.  The coverage reports then ignore the entire clause.
```Python
        if np.isinf(dt) and np.isinf(nt):  # coveralls: ignore
-            raise ValueError("Both dt and nt are infinite.")
```

namurphy:
PlasmaPy's motto should be: `Coverage decreased (-0.0009%) to 99.607%`

plasmapy/classes/species.py
StanczakDominik: Yeah, I literally missed and hit the wrong line here :D

PR #196, opened 2017-12-17
Use intersphinx to link to other python projects

dstansby:
This makes it possible to link to documentation in other Python projects - for example, I've changed the parameter documentation so now "Quantity" is a clickable link, that takes the reader to the Astropy website documentation for the `Quantity` class.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,11 +32,16 @@ sys.path.insert(0, os.path.abspath('..'))
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-# Warning: I installed numpydoc on my local machine to get this to work
-#          but I am not sure if there is supposed to be a requirements
-#          page for the docs section. So you might need to install this
-#          to use sphinx here. ~Colby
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -32,11 +32,16 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-# Warning: I installed numpydoc on my local machine to get this to work
-#          but I am not sure if there is supposed to be a requirements
-#          page for the docs section. So you might need to install this
-#          to use sphinx here. ~Colby
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.mathjax', 'numpydoc']
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.mathjax',
+              'sphinx.ext.napoleon', 'sphinx.ext.intersphinx']
```
dstansby: The mapping didn't place nicely with `numpydoc`, so I changed it to `sphinx.ext.napoleon` which also does `numpydoc` docstrings, and is included with `sphinx` anyway.

StanczakDominik: That sounds reasonable. I haven't heard of Napoleon before (thanks!), but after a quick search it seems like a solid option.

Issue #184, opened 2017-12-13
Add average ionization functionality


lemmatum:
We should allow for average ionization as an argument to functions like plasma_frequency, instead of just assuming integer values of ionization based on what particle is passed.

Related to #182

lemmatum:
We should decide whether we want this by 0.1 or not.

namurphy:
In addition to allowing for average ionization functionality, we should also have a way to represent the full charge state distribution (e.g., what fraction of iron atoms is in each of of the Fe 0+, Fe 1+, Fe 2+, ... states).

A colleague and I have done before this for a [prototype non-equilibrium ionization modeling package](https://github.com/namurphy/SunNEI) for solar eruptions.  [The applications we had in mind were low density astrophysical plasmas that rapidly heated or cooled such that ionization and recombination were not able to keep up with the fast temperature changes.]  I ended up using a dictionary of dictionaries of dictionaries, which makes me cringe a little bit when I think about it, so I think we can represent this in a way that is more reasonable and intuitive and less cumbersome.  We'd also need the charge state distribution to predict spectra.  I've discussed this a bit with @wtbarnes who is developing [fiasco](https://github.com/wtbarnes/fiasco).  This extra issue is beyond what we should plan on including for 0.1 though!

namurphy:
With respect to how to modify the atomic subpackage to account for this, would it be enough to allow for syntax like: `"Fe-56 2.7+"`?

wtbarnes:
Ionization equilibrium is one of the few things I've actually implemented in fiasco thus far. fiasco actually provides two ways of getting at the ionization equilibria,

1. CHIANTI ships ionization fractions (as a function of temperature) as part of the database so this can be accessed as a property of an `Ion`. This just pulls data straight out of the database and interpolates it to a given temperature.
```python
>>> import fiasco
>>> import astropy.units as u
>>> import numpy as np
>>> ion = fiasco.Ion('fe_15', np.logspace(5,8,100)*u.K) #note by fe_15 I mean Fe XV or Fe +14
>>> ion.ioneq
```

2. Calculate the ionization fraction of every ion of an element by calculating the ionization recombination rates as a function of temperature and solving the ionization balance equations,
```python
>>> el = fiasco.Element('iron', np.logspace(5,8,100)*u.K)
>>> ioneq = el.equilibrium_ionization()
>>> ioneq[:,14] # gives ionization fraction of Fe 15 (or +14)
```

[...]

PR #201, opened 2017-12-20
Average ionization

lemmatum:
Added optional argument `z_mean=None` to a bunch of functions in `parameters.py`. When value is `None` it defaults to using atomic charge state and signals a `PhysicsWarning`. These warnings may get annoying, so either tests need to be changed to use some value, like `z_mean=1` to avoid the warnings during pytest or can remove the warnings entirely.

I also added tests for `z_mean` under `test_parameters.py`.

Addresses #184

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -650,7 +650,6 @@ def gyrofrequency(B, particle='e', signed=False, z_mean=None):
                 Z = atomic.charge_state(particle)
             except ValueError:
                 Z = 1
-            Z = abs(Z)
         else:
             # using user provided average ionization
             Z = z_mean
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -866,6 +866,7 @@ def plasma_frequency(n, particle='e', z_mean=None):
         else:
             # using user provided average ionization
             Z = z_mean
+        Z = np.abs(Z)
     except Exception:
         raise ValueError(f"Invalid particle, {particle}, in "
                          "plasma_frequency.")
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -9,12 +9,13 @@ import plasmapy.atomic as atomic
 # from plasmapy.atomic import ion_mass, charge_state
 
 import numpy as np
+import warnings
 
 # For future: change these into decorators.  _check_quantity does a
 # bit more than @quantity_input as it can allow
 import plasmapy.utils as utils
 from plasmapy.utils.checks import _check_quantity
-from plasmapy.utils.exceptions import PhysicsError
+from plasmapy.utils.exceptions import PhysicsError, PhysicsWarning
 
 
 r"""
@@ -176,6 +177,8 @@ def Alfven_speed(B, density, ion="p", z_mean=None):
         try:
             m_i = atomic.ion_mass(ion)
             if z_mean == None:
+                warnings.warn("No z_mean given, defaulting to atomic charge",
+                              PhysicsWarning)
                 try:
                     Z = atomic.charge_state(ion)
                 except ValueError:
@@ -183,6 +186,8 @@ def Alfven_speed(B, density, ion="p", z_mean=None):
             else:
                 # using average ionization provided by user
                 Z = z_mean
+            # ensuring positive value of Z
+            Z = np.abs(Z)
         except Exception:
             raise ValueError("Invalid ion in Alfven_speed.")
         rho = density * m_i + Z * density * m_e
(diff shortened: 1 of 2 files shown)


lemmatum:
Do you guys think that the places I've added z_mean are appropriate/macroscopic enough? Is the docstring sufficiently descriptive?

plasmapy/physics/parameters.py
```diff
@@ -167,10 +170,18 @@ def Alfven_speed(B, density, ion="p"):
...
+            if z_mean == None:
+                warnings.warn("No z_mean given, defaulting to atomic charge",
+                              PhysicsWarning)
+                try:
+                    Z = charge_state(ion)
+                except ValueError:
+                    Z = 1
+            else:
+                # using average ionization provided by user
+                Z = z_mean
```
StanczakDominik: I think this whole part from line 173 to here could be reasonably dumped into a _private function, to avoid duplication.

plasmapy/physics/tests/test_parameters.py
```diff
@@ -137,6 +137,22 @@ def test_Alfven_speed():
...
 
+    # tests for z_mean functionality
+    # testing if warning is signaled for default value
+    with pytest.warns(PhysicsWarning):
+        Alfven_speed(1 * u.T, 5e19 * u.m**-3, ion='p')
+    # testing for user input z_mean
+    testMeth1 = Alfven_speed(1 * u.T,
+                             5e19 * u.m**-3,
+                             ion='p',
+                             z_mean=0.8).si.value
```
StanczakDominik: The name may not be the most fortunate here, as testing meth could probably be delegated to PyChemistry. Or Pysenberg, or something. ;)

The line itself is completely fine though and I'm fine with keeping this as it is!

samurai688:
I like the code, but I think maybe it should be spelled out in the documentation what the z_mean inputs are for. Like, the user might ask, if I use a Z_mean of 1.5 in the ion_acoustic speed for example, what does it mean? Would that indeed be the unique ion acoustic speed of such a plasma? Would you see multiple speeds, each corresponding to different species, and it's useful for other purposes? Is it simply an approximation? Does it all depend? etc. It'd be nice to be able to point them in the right direction. Also I don't know the answers, so now I'd be curious to know.

plasmapy/physics/tests/test_parameters.py
lemmatum: Well I am in Albuquerque so Pysenberg would be quite appropriate 😷 💰 💵 🔬

plasmapy/physics/parameters.py
lemmatum: That sounds like a terribly reasonable idea. Perhaps we can even call it something like `_boilerPlate()` 😛

StanczakDominik: Try to be more descriptive, we might need multiple boiler plates :D

lemmatum: How about `_ZBoilerPlate()`? With an emphasis on 'Zee', to be read like a mad German scientist.

lemmatum:
So shall I just remove the z_mean warnings?

lemmatum:
That reset was nasty... hopefully things are correctly merged.

StanczakDominik:
It looks like some of those got duplicated... I think we forgot to `git fetch --all` before resetting, and your 'origin/averageIonization` didn't update. I'm sorry, I should have mentioned that.

The pipeline here should be

* fetch all to update your local view of the remote version
* reset to state at origin, maybe we should have gone with a soft one, so if you made any changes to files they remain in the directory and can be added in another commit
* upon adding any more commits, a normal git push should be able to push them to your remote branch

But this won't deal with the duplicates. Would you mind me removing them and trying again, with the git fetch this time?

The z_mean warnings I'm not entirely sure about. I think we should have reasonable defaults, and if z_mean is optional and "off" by default, we shouldn't scare people with warnings for using the default arguments.

At the same time, z_mean is important but it seems (to me at least) to be on another level of complexity. You probably wouldn't need to use it in a simple analysis.

This would logically lead to a removal of the warnings... but perhaps the documentation for stuff that uses z_mean should stress its importance more? Documentation doesn't really scream at people as much as warnings can be perceived to do.

@namurphy, what's your take on this?

StanczakDominik:
@lemmatum I'm gonna assume I can try to fix it, as I'm pretty sure you didn't mean to add `.orig` files to the branch!

StanczakDominik:
I commented out the warnings (I felt it'd be easier to possibly return to them this way) and removed some unused imports as well. If you guys decide it's better with warnings, feel free to revert the change :)

lemmatum:
I've added some more info on the meaning of using `z_mean` for each parameter. Let me know if these make sense.
Also is there another place in the docs where such descriptions should be added?

samurai688:
Looks good to me!

PR #202, opened 2017-12-20
Automatically build module documentation

dstansby:
Hello, apologies that this is a bit of a monster PR, but in the long run it should make documentation more automated.

Currently there is just a list of functions that are documented; when a new function is added to PlasmaPy an entry needs to manually put in the `.rst` files. This PR uses https://sphinx-automodapi.readthedocs.io/en/latest/#user-guide to automatically document the modules. This means if a new function is added, it automatically gets documented. It also means that each and every function doesn't have to be manually imported to the 'parent' module.

Also

- Had to tidy up some docstrings that had warnings (before this PR they weren't even showing up the docs!)
- Had to remove some manual external imports, and import the module instead


plasmapy/atomic/__init__.py
```diff
@@ -1,12 +1,5 @@
...
-                     known_isotopes, common_isotopes, stable_isotopes,
-                     isotopic_abundance, charge_state, electric_charge)
-
-from .nuclear import (nuclear_binding_energy, nuclear_reaction_energy)
-
-from .elements import Elements
-
-from .isotopes import Isotopes
+"""Physical and mathematical constants for use within PlasmaPy."""
+from .atomic import *
```
StanczakDominik: ... we should probably have thought of that. :laughing:

plasmapy/atomic/atomic.py
```diff
@@ -3,16 +3,24 @@
...
 
-
+"""
+__all__ = ['atomic_symbol', 'isotope_symbol', 'atomic_number',
+           'is_isotope_stable', 'half_life', 'mass_number',
+           'element_name', 'standard_atomic_weight', 'isotope_mass',
+           'ion_mass', 'known_isotopes', 'common_isotopes',
+           'stable_isotopes', 'isotopic_abundance', 'charge_state',
+           'electric_charge']
+"""
```
StanczakDominik: I suppose this is here simply for debugging purposes? Am I wrong? Is this somehow necessary for automodapi?

plasmapy/physics/__init__.py
```diff
@@ -1,34 +1,8 @@
...
-                           Maxwellian_speed_1D,
-                           Maxwellian_velocity_3D,
-                           Maxwellian_speed_3D,
-                           kappa_velocity_1D,
-                           kappa_velocity_3D)
+from .parameters import *
+from .quantum import *
+from .relativity import *
+from .transport import *
+from .distribution import *
```
StanczakDominik: We've had problems with forgetting to put stuff into __init__ before, so I love this solution!

plasmapy/atomic/atomic.py
dstansby: Woops, this shouldn't be here!

PR #205, opened 2017-12-21
Add util docs

dstansby:
I think this finishes off adding everything to the docs!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -49,12 +49,20 @@ Classes
 
     plasma/index
 
+Utilities
+=========
+
+.. toctree::
+   :maxdepth: 1
+
+   utils/index
+
 Development
 ===========
 
 .. toctree::
     :maxdepth: 1
-       
+
     development/code_guide
     development/testing_guide
     development/doc_guide
--- a/docs/utils/index.rst
+++ b/docs/utils/index.rst
@@ -1,0 +1,4 @@
+.. currentmodule:: plasmapy.utils
+
+.. automodapi:: plasmapy.utils
+   :no-inheritance-diagram:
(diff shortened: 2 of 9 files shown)


StanczakDominik:
I think you might have missed git adding docs/utils, so that's why the docs aren't building right now! Why the tests are taking failing I won't really be able to tell until tomorrow.

StanczakDominik:
I think I was able to fix this, though I'm not 100% certain on how.

This is our `utils/__init__.py`:

```python
from .checks import *
from .exceptions import *
```

But if you try to import `utils`, private variables from `checks` aren't visible in that namespace:

```python
In [1]: from plasmapy import utils

In [2]: utils
Out[2]: <module 'plasmapy.utils' from '/home/dominik/Code/PlasmaPy/plasmapy/utils/__init__.py'>

In [3]: utils._check_relativistic
AttributeError                            Traceback (most recent call last)
<ipython-input-3-67acd506541e> in <module>()
----> 1 utils._check_relativistic

AttributeError: module 'plasmapy.utils' has no attribute '_check_relativistic'
```
So, I guess that's a thing. Maybe we should make `_check_relativistic` a public function... have a functional version of that in `checks` and then a decorator in, I dunno, `decorators`?

-------------
I also removed an `else` clause as I don't see why we shouldn't check thermal velocities for relativistic effects. That may come into play for high temperatures, right? We should still warn about this. @namurphy, does that make sense?


-------------
The decrease in coverage is negligible as usual and we can probably merge this if nobody raises objections.

I should really look into having an accepted decrease or something in `.coveragerc`...

StanczakDominik:
And that's it for this one! Thanks @dstansby for the stellar work, thanks @namurphy for input! :D

dstansby:
Thanks for tidying that up!

The problem was that `from blah import *` doesn't import methods that start with an underscore (are they called private methods? I'm not really sure), so the only way to use `_check_quantity` is to call it directly from the file it lives in.

PR #212, opened 2017-12-22
Build pdf docs on circleci

dstansby:


StanczakDominik:
YES

Thanks a lot @dstansby, this is going to make the bloody latex bug hunting a whole lot easier!

dstansby:
I've just realised that the circleci machines don't have latex installed. I'm guessing installing latex for every build will take ages, so depending on how much you want to do this I could look into a custom docker image: https://circleci.com/docs/2.0/custom-images/ (suggested here: https://discuss.circleci.com/t/texlive-via-docker/11495)

StanczakDominik:
Oh... Well, that sucks. Still, this would be quite handy - we're going to have a whole lot of math equations and is it'd be really nice to be able to check for errors in those without building the docs manually and looking through the formulas themselves.

I've been meaning to figure out Docker for a while now, though. Perhaps I could take over on this one? I really don't want to overburden you.

dstansby:
🎉 looks like this works, and only takes a few minutes. I might look into docker images in the future out of interest, since they seem to be the way to go: https://circleci.com/docs/2.0/custom-images/

StanczakDominik:
Awesome! Let's get this merged, then!

PR #217, opened 2017-12-28
Create exceptions and warnings for atomic subpackage

### LINKED ISSUE #168: Create physics exceptions and warnings

namurphy:
The atomic ⚛️ subpackage currently raises `ValueError` and `TypeError` a whole bunch of times.  These exceptions are appropriate but are not specific enough to distinguish between different types of problems and make use of exception subclassing.  For example, a particle such as `'Fe-56'` is an invalid ion but a valid isotope.  In this case, this particle should raise something like an `IonError` but not an `IsotopeError`.

I started out by adding `AtomicError`, `IonError`, `IsotopeError`, `ElementError`, and `ParticleError` as exceptions and `AtomicWarning` as a warning.  We will need to think more carefully about how to handle subclassing.

I also added "pass" to each of the exceptions and warnings.  This does not appear to be needed, but I was thinking that this line would show up in coverage tests for exceptions and warnings that are not being tested.  _(Update: This line does not actually show up in coverage tests.)_

This builds upon the work completed in #170 that closed #168.

### DIFF
--- a/plasmapy/utils/exceptions.py
+++ b/plasmapy/utils/exceptions.py
@@ -36,7 +36,16 @@ class AtomicError(PlasmaPyError):
     pass
 
 
-class IonError(AtomicError):
+class MissingAtomicDataError(AtomicError):
+    """Error for use when atomic data is missing."""
+    pass
+
+
+class NoChargeInfoError(AtomicError):
+    """Error for use when charge information is needed but missing."""
+
+
+class IonError(NoChargeInfoError):
     """Error for use when an ion is invalid."""
     pass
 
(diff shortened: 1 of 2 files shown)


namurphy:
Thank you @StanczakDominik for the speedy review!  This is a pretty straightforward change so I'll merge it now.

PR #218, opened 2017-12-28
Fix example in betafrac doc

dstansby:
Just needed an extra line before the 2nd example to parse it as code. Hopefully with the shiny new CircleCi doc viewing you should be able to see this change in all it's glory!


StanczakDominik:
To be precise, https://21-114539428-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/api/plasmapy.utils.check_relativistic.html?highlight=betafrac is what we should be looking at. And it's pretty cool, indeed!

Issue #195, opened 2017-12-17
Change sequential tests to fixture and class->function based ones


lemmatum:
Our tests are a terrible mess.... Look at this!

```
def test_gyroradius():
    r"""Test the gyroradius function in parameters.py."""

    assert gyroradius(B, T_e).unit == u.m

    assert gyroradius(B, 25 * u.m / u.s).unit == u.m

    assert gyroradius(T_e, B) == gyroradius(B, T_e)

    assert gyroradius(V, B) == gyroradius(B, V)

    assert gyroradius(B, V) == gyroradius(B, -V)

(and so on...)
```
How are you supposed to differentiate which assert failed?
What are these asserts doing? How do we know the tests are meaningful without any comments describing what is being tested?

Heavy refactoring is required. I think a class based approach, like what I've done with `Test_kappa_thermal_speed`, makes sense. It allows us to organize multiple tests under the header of testing a single function. One assert or one failure in a with context per method.

StanczakDominik:
Okay, I agree that a class based approach to this would be nice.

I might be able to take a shot on this over the next few days.

lemmatum:
I've been gradually tackling this for functions where I've had to write new tests. So it should already be partially done. `test_parameters.py` will probably be the worst off.

namurphy:
I'm going to procrastinate Fortran and make `test_dispersion.py` less horrible. 💾 🖥 🖱 ⌨️

lemmatum:
I'm assuming that we are shooting for class based tests for v0.1 with possibility of switching over to fixtures in the future. Is that a safe assumption?

StanczakDominik:
I'd say we're shooting for whatever tests work for now, with cleanup intended in the future : laughing:

StanczakDominik:
I think what we have is good enough for now. We could really use a cleanup in the future.

PR #219, opened 2017-12-28
Make atomic subpackage tests less horrible

namurphy:
The purpose of this pull request is to continue to improve the tests for the atomic subpackage in response to #195.  The atomic subpackage tests that I originally wrote in #24 and #32 were a bunch of different assert statements that did not produce clear messages when tests failed.  This shortcoming made it harder to determine what went wrong when things went wrong.  The tests were already greatly improved in #119 with the use of `pytest.mark.parametrize` to split the tests into much smaller ones.  This pull request makes some more incremental improvements by adding short docstrings to the test functions and significantly more helpful error messages when things go awry.  I also used a lot of f-strings because f-strings are awesome.


plasmapy/atomic/tests/test_atomic.py
```diff
@@ -328,21 +373,23 @@ def test_element_name(argument, expected):
...
 
-def test_standard_atomic_weight_not_none():
-    assert standard_atomic_weight('N') is not None
-
-
 def test_standard_atomic_weight_value_between():
-    assert 30.973 < standard_atomic_weight('P').value < 30.974
+    """Test that standard_atomic_weight returns approximately the correct
+    value for phosphorus."""
+    assert 30.973 < standard_atomic_weight('P').value < 30.974, \
```
StanczakDominik: Wait a minute, you can do three way inequalities now?! All that code I wrote, so ugly T.T

namurphy: This reminds me of the moment that I really began to appreciate the awesomeness of Python, which was when I saw the following code:
```Python
a, b = b, a
```

PR #220, opened 2017-12-29
Add type hints to atomic subpackage

namurphy:
I was just reading about function annotations, and began to realize how awesome type hints are.  As an example, let's start with a simple function:
```Python
def add(a, b):
    return a + b
```
Adding type hints changes this to:
```Python
def add(a: float, b: float) -> float:
    return a + b
```
More complex things can be done with the [`typing`](https://docs.python.org/3/library/typing.html) module (including `Union`, `Any`, `Optional`, `List`, etc.), which is provisionally included in the standard library so backward compatibility is not guaranteed.  

While Python is a dynamically typed language and the type hints are never enforced, this is helpful because:
 - Type hints make the code more readable
 - Type hints allow optional static type checking with packages such as `mypy` (which is currently in development) which can help us find bugs.  🐝 🐞 🐛 🐜 

This pull request adds type hints to the atomic subpackage.  

References:
 - [PEP 483: The Theory of Type Hints](https://www.python.org/dev/peps/pep-0483/)
 - [PEP 484: Type Hints](https://www.python.org/dev/peps/pep-0484/)

### DIFF
--- a/plasmapy/atomic/atomic.py
+++ b/plasmapy/atomic/atomic.py
@@ -3,6 +3,7 @@
 import numpy as np
 import re
 from astropy import units as u, constants as const
+from astropy.units import Quantity
 import warnings
 from .elements import atomic_symbols, atomic_symbols_dict, Elements
 from .isotopes import Isotopes
@@ -386,7 +387,7 @@ def is_isotope_stable(argument: Union[str, int],
     return is_stable
 
 
-def half_life(argument: Union[int, str], mass_numb: int = None) -> u.Quantity:
+def half_life(argument: Union[int, str], mass_numb: int = None) -> Quantity:
     r"""Returns the half-life in seconds for unstable isotopes, and
     numpy.inf for stable isotopes.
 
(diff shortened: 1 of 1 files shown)


plasmapy/atomic/atomic.py
```diff
@@ -1473,7 +1486,7 @@ def _extract_charge_state(argument):
     return argument, charge_state
 
 
-def _is_neutron(argument, mass_numb=None):
+def _is_neutron(argument: Any, mass_numb: int = None) -> bool:
```
StanczakDominik: I haven't really gotten into the typing documentation yet, but here's a question: is there a point in hinting that this `argument` can be `Any`? Is there any added value over leaving it empty, besides being explicit, I guess?

namurphy: If we used a static type checker in PlasmaPy, I doubt it would make a difference whether we had `Any` or didn't have anything.  I think the benefit would be that it makes what's going on more clear.

This particular annotation may end up being short-lived because I'm thinking about refactoring the atomic subpackage so that we don't have to deal with all of these special private functions like `_is_neutron`.  We can probably take care of all of the annoying aliases in a much cleaner way.

Issue #179, opened 2017-12-11
Improve handling of aliases and special cases in atomic subpackage


namurphy:
Right now there is some rather messy code in `atomic.py` to account for the pesky special cases that arise.  In particular are the functions at the bottom such as `_is_electron` and `_is_positron` that go through all of the different possible aliases for each of those particles and return `True` if there's a match and `False` otherwise.  There is definitely room for improvement.  This is very related to #174.

One possibility would be to create a dictionary or dictionaries with the aliases that account for both case-sensitive and case-insensitive instances (e.g., `'H'` is case-sensitive whereas `'hydrogen'` is case-insensitive).  This would hopefully let us simplify the code a great deal.

Another possibility would be to create a function called `particle_symbol` along the lines of `atomic_symbol` and `isotope_symbol`.  This function would then be responsible for these special cases, and would also take care of elements, isotopes, and ions by putting them into our standard format.

Issue #174, opened 2017-12-08
Exotic particles


lemmatum:
It might be worthwhile to add some exotic particles to the particle library, such as muons, tauons, and neutrinos.

Opinions?

namurphy:
I suspect that cosmic ray scientists (including many plasma astrophysicists) would use this functionality.  This sounds like a good idea to me, especially since the implementation should be rather light.

namurphy:
To implement this, I think we could create a file called `particles.py` which would have electrons, neutrinos, muons, tau particles and their anti-particles.  This would also be really helpful to account for all of the irksome special cases that keep showing up.  I've also been wondering how to cleanly account for all of the different aliases that are out there, some of which are cases sensitive and others of which are case-insensitive.  I tentatively assigned this to myself since there's more work that I want to do to clean up and refactor the atomic subpackage.

StanczakDominik:
Does #221 solve this, or is there more to be done here?

namurphy:
#221 partially addressed this issue by adding in dictionaries for the different particles, but there is still work to do to make sure the methods in the atomic subpackage accesses the data in these dictionaries.  Right now none of the existing public methods can get the mass of a muon, for example.  I was planning on adding `particle_symbol` and `particle_mass` functions as well (plus maybe some more like `baryon_number` and `lepton_number`) that would allow access to this information.  I'd keep this issue open for now.

namurphy:
Closed by #239.

PR #221, opened 2017-12-29
Storing particle information and resolving particle aliases

namurphy:
This pull request creates `Particles` which is a dictionary in `atomic/particles.py` that contains information on particles such as electrons, muons, taus, neutrinos, protons, neutrons, and their antiparticles.

This pull request is preparation work for eventually providing a resolution to #174, and is related to #179.


namurphy:
The methods used to create the dictionary for the particles first started out as tests, and then I came to the conclusion that it would be more reliable to construct the dictionary using those methods rather than inputting them in by hand.  Plus, if the spin of fermions ever changes from 1/2, it's now much easier to change!  :atom:  😸 

> Looks complex, but reasonable.

The dictionary containing all of the alias information is unavoidably going to be complex (sigh...).  My hope has been to concentrate the complex parts of all of this into a few methods so that the rest of the atomic subpackage can be relatively simple.  

> Perhaps it'd be a good idea to add an example showing these two dictionaries? Literally just pretty printing them out. Looks like it'd be just about 5 lines of code.

That would be helpful, yes.  The stuff contained in this dictionary is probably not going to be part of the public API, so maybe it would go in the docstrings for the methods to create the dictionaries.

plasmapy/atomic/particles.py
```diff
@@ -0,0 +1,219 @@
...
+
+    if alias in _case_sensitive_aliases.keys():
+        return _case_sensitive_aliases[alias]
+    elif alias.lower() in _case_insensitive_aliases.keys():
+        return _case_insensitive_aliases[alias.lower()]
+    else:
+        return alias
+
+
+print(_case_insensitive_aliases)
```
StanczakDominik: Put this in to an `if __name__=="__main__":` block, or it's going to print the aliases on every import of this module!

PR #191, opened 2017-12-15
Transport eqs

lemmatum:
Addresses #35 
Adds coupling parameter for #49 
Heavily expands Coulomb logarithm for #154 

Adding a bunch of transport functionality and re-arranging some functions within transport.py. Much of the additions are based off of [Gericke, Murillo, and Schlanges (2002)](https://doi.org/10.1103/PhysRevE.65.036418) for extending Coulomb logarithm to the dense plasma case where quantum effects become important. To that end, I am also adding a coupling parameter function, which is handy for determining which regime a plasma is in. The coupling parameter itself is quite involve and requires things like: ideal chemical potential, Fermi integral, polylogarithms.

I am also adding functions for collision frequency and mean free path (which I believe I wrote based off of F. F. Chen's book).

PR #223, opened 2018-01-02
Adding cython integration

### LINKED ISSUE #222: Cython integration
Astropy's template has a way to [integrate Cython and C extensions.](http://docs.astropy.org/en/stable/development/ccython.html) We should get this working so we can have super optimized code for simulations!

lemmatum:
Closes #222 

Initial tests with cython integration. `parameters_cython.pyx` does indeed get recognized and compiles when pip installing the package. Next need to figure out how to ship the compiled `.c` files from `.pyx`, skip compilation, and import the `.c` functions to `.py` modules.


lemmatum:
We need some clever way to include Cython as a dependency for the development version of plasmapy but not for the distribution version. Perhaps some hints can be obtained from astropy?

lemmatum:
CircleCI seems to be failing at the dependency step... perhaps because cython is currently not declared as a dependency?

StanczakDominik:
> We need some clever way to include Cython as a dependency for the development version of plasmapy but not for the distribution version. Perhaps some hints can be obtained from astropy?

For now the distribution version is basically the same as the dev version, so let's keep them the same for now until we figure this out.

I fixed CircleCI's install progress step being muted (there's a --no-progress-bar option in pip master, but it's not in releases yet - a shame). I don't yet understand why the build's failing, especially given that Travis is building. nicely.

Could you add an actual test for the cython fib functionality? So we can make sure Travis is building and executing it right?

lemmatum:
Currently I get a `parameters_cython.cpython-36m-x86_64-linux-gnu.so` at the top most directory of the package rather than inplace when building.

lemmatum:
circleCI is getting an import error in `astropy_helpers` where it doesn't find the numpy module. Perhaps we are missing the numpy module that ships with astropy template for dealing with cython? I think the cython compatible numpy module is different from the standard library one.

BTW I added a test of the `fib()` function.

lemmatum:
Travis is failing because it doesn't compile the `.c` file. I probably need to add compilation of `.c` files under `setup_package.py`.

lemmatum:
So I've figured out how to modify `setup_package.py` such that compiled `.so` libraries get put "inplace", but I haven't figured out how to import for testing in `test_parameters_cython.py` ... I think there is still something wrong with how I am compiling things.

lemmatum:
[This](https://stackoverflow.com/questions/8024805/cython-compiled-c-extension-importerror-dynamic-module-does-not-define-init-fu) may be related to the problem I am currently having.

lemmatum:
So for travis to work it needs to build the binaries from the cython file, which means adding a bunch of dependencies like gcc. Thoughts on how to do this @StanczakDominik ?

StanczakDominik:
I can say with complete certainty that *huh*.

Well, it'd probably be best to look at how [Astropy does it](https://github.com/astropy/astropy/blob/master/.travis.yml)... but I probably won't be able to do that tonight :zzz: :disappointed:

.circleci/config.yml
```diff
@@ -28,6 +32,7 @@ jobs:
     steps:
       - checkout
       - run: *deps-install
+      - run: *plsma-install
```
StanczakDominik: I seriously have no idea why this line is necessary. AFAIK `steps` are not done in parallel.

plasmapy/setup_package.py
```diff
@@ -0,0 +1,35 @@
...
+# getting C and Cython extensions
+cython_testFile = 'plasmapy/physics/parameters_cython.pyx'
+c_testFile = 'plasmapy/physics/parameters_cython.c'
+
+# names for compiling the .c files to
+compileName = 'parameters_cython'
+#compileNamePath = 'plasmapy/physics/parameters_cython'
+compileNamePath = 'plasmapy.physics.parameters_cython'
+compileName_cython = 'plasmapy/physics/parameters_cython_pyx'
+compileName_c = 'plasmapy/physics/parameters_cython_c'
```
StanczakDominik: Most of these names are unused. Is that intended?

StanczakDominik:
Oh yeah, definitely add setup_packages.py to .coveragerc excluded files. It doesn't seem worth testing.

.circleci/config.yml
lemmatum: I think that's precisely why it needs to be run like this. Steps are run sequentially so the dependency installation step gets to complete and then numpy exists at setup time. Whereas this:
```
|
     python3 -m venv venv
     . venv/bin/activate
-    pip install -q --user -r requirements/automated-documentation-tests.txt
-    pip install -q --user .
```
doesn't run as separate commands but rather as one large command. (The entire thing only has one unix dollar sign associated with it when looking at the circleci output rather than a separate dollar sign for each line.)

plasmapy/setup_package.py
lemmatum: This was for testing how to get inplace binaries (in the correct directories), as setuptools was fickle when the binary name and the .pyx file name didn't match exactly.
To fix this I just resorted to explicitly cythonizing the extensions in `setup.py` rather than letting setuptools handle it implicitly.

All of this will eventually have to be cleaned up to allow for building from the distributed `.c` files rather than cythonizing for every build, which require a `USE_CYTHON` flag in `setup.py` [like this](http://docs.cython.org/en/latest/src/reference/compilation.html#distributing-cython-modules) I think.

lemmatum:
@namurphy are the coverage reductions due to my code or due to your changes in `atomic`?

lemmatum:
Last commit was due to rebasing/git magic

lemmatum:
Things that need to be done:
* Add `USE_CYTHON` flag to switch between building from `.pyx` and building from `.c` in `setup.py`
* Add a method for dealing with empty list of cython extensions in `setup.py`, otherwise `cythonize` will fail horribly.
* Create a more generalized method for fetching cython modules in `setup_package.py`. Right now need to do it explicitly.
* Cleanup `setup.py` and `setup_package.py`
* Add some real functions to `parameters_cython.pyx` and corresponding tests. Remove `fib()`
* Test numpy integration with cython files
* add `parameters_cython` to `__init__.py` files? Not sure if necessary...

Some of this could maybe be fixed by figuring out what in the world `astropy_helpers` is trying to do by passing a non-cythonized list of extensions to `ext_modules` in `setup()`.

lemmatum:
@StanczakDominik I'm liking these speedier tests 👍

StanczakDominik:
I'm not entirely sure the rebase has gone as it should have here, we probably shouldn't have Nick's commits here.

I didn't remember to remind you about git fetch on Riot... >_<

No matter though, I can still probably clean this up later.

lemmatum:
I'm 99% sure I git fetched before merging/rebasing. First I merged your changes to Travis from this branch and then I rebased on top of upstream/master.

StanczakDominik:
I am definitely being insufficiently poked and prodded and thus keep forgetting PR's. :( All the duplicate commits are cleared and tests are passing.

StanczakDominik:
Coverage is decreasing a little bit, but I would say this is just about ready for a merge. I would like to wait with that for about 24h, as this PR too will prove instructional. :smile: 

[A relevant article](http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html) claiming that if we want to coverage test our Cython files, we're going to be taking performance hits on them. And that's kind of a conundrum.

lemmatum:
Nice! Shall we include a cythonized version of some real physics function that is frequently used instead of fibonacci? Perhaps an optimized version of the 3D Maxwellian distribution function?

StanczakDominik:
Now that is a pretty good idea! :)

lemmatum:
I replaced fibonacci function with a cythonized thermal velocity function.
Had to do some nasty rebasing in the process, so let me now if you catch any errors! 😅

lemmatum:
A few points on the cythonized thermal velocity function:
* It doesn't handle units (yet)
* it is still a normal python function instead of a cdef or cpdef function, which could optimize it further to avoid re-wrapping the returned value
* we should test how much faster it is than the pure python version 😼

.circleci/config.yml
```diff
@@ -4,6 +4,7 @@ deps-run: &deps-install
     python3 -m venv venv
     . venv/bin/activate
     pip install --user -r requirements/automated-documentation-tests.txt | cat
+    pip install --user . | cat
```
lemmatum: Big mistake!

lemmatum:
@StanczakDominik Dude, why did you push these quick commits and then merge this branch? CircleCI got messed up when you added `pip install --user . | cat` back into `&deps-install`. There was a reason I separated the plasmapy installation into `&plsma-install`. If you run both under the same command then numpy isn't available for plasmapy, which is necessary for Cython integration. This error has now propagated to #191  and you can see it in the circleCI logs.

StanczakDominik:
Ah crud. Terribly sorry, I thought I had checked them working on my fork and wanted to get this done quick, then go to sleep. My bad; won't happen again.

PR #225, opened 2018-01-04
Remove PIC example for now

StanczakDominik:
The PIC example in docs is slow, it slows down building the docs, it should be moved into a subpackage so it doesn't slow down plasmapy's general builds. For now, let's remove it.


lemmatum:
Sounds reasonable. Have you saved the file elsewhere or can we just comment this out?

plasmapy/examples/plot_particle_stepper.py
```diff
@@ -1,72 +0,0 @@
-
```
lemmatum: Is it possible to just comment this out instead?

StanczakDominik:
Or that ... Didn't think about it. I ended up just removing the plot part of the name. That way it does not get executed. We can add it back some other time.

PR #226, opened 2018-01-04
Run two CircleCI jobs in parallel

StanczakDominik:
So I think it's a tragedy that we're currently waiting for Latex to slowly build at the end of the job after all the HTML is ready. I thus went ahead and split that in two.

Ideally I'd like to be able to install (PlasmaPy and its dependencies) and (latex from apt-get) simultaneously, but [I wasn't yet able to do that](https://github.com/StanczakDominik/PlasmaPy/commits/faster-circle).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,3 @@
-
 deps-run: &deps-install
   name: Install Python dependencies
   command: |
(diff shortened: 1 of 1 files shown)


lemmatum:
Hooray for parallelization! 🎆

Issue #135, opened 2017-10-03
Check whether we need to update our doctests on every Astropy.constants update


StanczakDominik:
With #129 in the works and given that

* we have a bunch of functions using those values and numerically they do change
* our doctests check for exact results without going through anything like `np.isclose`

we might be setting ourselves up for some future work and bugs. I'm not sure what the best solution would be here.

RoberTnf:
Maybe hardcode the constants that are susceptible of changing in the tests?

StanczakDominik:
Very likely fixed by #230

PR #228, opened 2018-01-07
Modified plasma_dispersion_func to use wofz() in scipy. Fixed tests

### LINKED ISSUE #208: Use Dawson function for dispersion
`plasma_dispersion_func` under `mathematics.py` currently uses `erf()` along with some other terms. This can be simplified to Dawson function, [dawsn](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.dawsn.html), and may even offer some minor speedups if scipy implements it in C code.

siddharth185:
Modified plasma_dispersion_func to use wofz() in scipy as Faddeeva function in mathematic.py . Modified rtol values so the tests pass. Modified the values in Docstring so the Doctests pass.

Closes #208

### DIFF
--- a/plasmapy/mathematics/mathematics.py
+++ b/plasmapy/mathematics/mathematics.py
@@ -3,6 +3,7 @@
 import numpy as np
 from scipy import special
 from astropy import units as u
+from scipy.special import wofz as Faddeeva_function
 
 
 def plasma_dispersion_func(zeta):
@@ -57,9 +58,9 @@ def plasma_dispersion_func(zeta):
     >>> plasma_dispersion_func(0)
     1.7724538509055159j
     >>> plasma_dispersion_func(1j)
-    0.7578721561413119j
+    0.757872156141312j
     >>> plasma_dispersion_func(-1.52+0.47j)
-    (0.6088888957234255+0.3349458388287403j)
+    (0.6088888957234254+0.33494583882874024j)
 
     """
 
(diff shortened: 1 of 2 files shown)


plasmapy/mathematics/mathematics.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 from scipy import special
 from astropy import units as u
+from scipy.special import wofz as Faddeeva_function
```
StanczakDominik: ... huh, that was there all along? :laughing:

plasmapy/mathematics/tests/test_dispersion.py
```diff
@@ -35,7 +35,7 @@ def test_plasma_dispersion_func(w, expected):
 
     Z_of_w = plasma_dispersion_func(w)
 
-    assert np.isclose(Z_of_w, expected, atol=1e-12 * (1 + 1j), rtol=1e-12), \
+    assert np.isclose(Z_of_w, expected, atol=1e-12 * (1 + 1j), rtol=1e-6), \
```
StanczakDominik: I was pretty much guessing blindly with the 1e-6 value. Could you try something lower, closer to the old 1e-12? Go as low as you can - ideally we'd like to both be accurate in testing the old tabularized values and not so accurate that machine precision will break these tests.

StanczakDominik:
As for the test failures... suddenly all the quantity docstrings seem to be failing, but @siddharth185 didn't change anything related to that. I checked on master at my own fork and well hello there, they're failing on master. This is looking like an @astropy change... I'm looking into it, because I think a potential change on their end may have fixed #135.

plasmapy/mathematics/tests/test_dispersion.py
siddharth185: Okay! I'll work on this.

StanczakDominik:
The newest changes on master should solve the docstring problem.

StanczakDominik:
I'm pretty sure this can be merged after it's rebased on top of master! I'll make sure and merge it tonight.

StanczakDominik:
Merging it. Thanks, @siddharth185 !

PR #235, opened 2018-01-13
Fix init py

### LINKED ISSUE #227: Fix __init__.py files
I've come to dislike the star imports in our `__init__.py`. I don't think dumping everything into the same namespace is good Python, while it may make it easier to find stuff.

I think we should figure out ways to break that up.

I dunno, maybe we should replace the star imports with named imports again... I'll look through other libraries, see how they do it.

siddharth185:
Closes #227 

Doctests are failing for me because I am using Conda env which won't let me install the newest numpy due to dependency issues.


StanczakDominik:
Hey, I'm adding some more fixes. Once I fix what I broke with the fixes, I'll merge it. Thanks @siddharth185 :)

PR #239, opened 2018-01-19
Create and implement Particle class

### LINKED ISSUE #182: Create an object-oriented Particle interface
This has been brewing in a bunch of places, including discussion on #171 and [this chat on Matrix](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/$15130263015090890CUoWY:matrix.org).

Right now, we're calling all our particle-dependent functionality via strings. Those tend to be problematic for fitting in the complete information on those (there's different elements, isotopes, ionization states, @namurphy I'm likely missing a bunch so please correct me here) while keeping them readable.

The idea here is that we could create a class for Particles, keep that information (element type, charge state, isotope) there as fields and do all the calculations and operations on that. This would let us avoid having a bunch of duplicate code in every single place a particle type is necessary.

We'd still like to be able to pass strings in (as before, `particle='e'`), so what @namurphy proposed is a decorator to go through the particle type argument and if a string was passed, turn that string into a Particle and pass that in. This seems reasonable AND it involves using a decorator, so it's totally baller to me.

The whole class setup would have to be figured out - do we want to have the same kind of object for atoms/nuclei and for electrons/other charged leptons (no isotopes for the latter)? What kind of inheritance scheme is best for those? I'm not sure yet, but this is part of figuring this out.

namurphy:
The purpose of this pull request is to create a `Particle` class to contain basic information about a particular particle.  An example:
```Python
>>> from plasmapy.atomic import Particle
>>> H = Particle('H')
>>> H
<Particle "H">
>>> H.m
<Quantity 1.67382335e-27 kg>
```
I still need to complete the code for this class, document it, and make the tests more thorough and less horrible.  

The next step (which might be worth including in this pull request) would be to create a decorator that takes in standard atomic inputs (e.g., a particle string, the integer charge `Z`, and `mass_numb`) and pass through `Particle(particle, Z=Z, mass_numb=mass_numb`).

Closes #182.


StanczakDominik:
Good timing, I literally came home as you posted it :laughing:

plasmapy/atomic/classes.py
```diff
@@ -0,0 +1,250 @@
...
+            self._is_antimatter = _Particles[particle]['antimatter']
+
+            if particle in _leptons + _antileptons:
+                self._is_lepton = True
+                self._generation = _Particles[particle]['generation']
+            else:
+                self._is_lepton = False
+
+            if particle in _fermions:
+                self._is_fermion = True
```
StanczakDominik: `self._is_fermion = particle in _fermions`, without the if here? The previous is understandable as you also need to allocate `self._generation`.

plasmapy/atomic/classes.py
```diff
@@ -0,0 +1,250 @@
...
+        if self._half_life is not None:
+            return self._half_life == np.inf * u.s
+        else:
+            raise MissingAtomicDataError(
+                f"Stability data for '{self._particle_symbol}' is not "
+                "available.")
+
+    @property
+    def is_antimatter(self):
+        return self._is_antimatter
```
StanczakDominik: I take it this is sort of in preparation for possible future upgrades? Otherwise, you'd just set `self.is_antimatter = self._is_antimatter` in `init`, or even get rid of the latter altogether, right?

plasmapy/atomic/parsing.py
```diff
@@ -0,0 +1,395 @@
...
+    return symbol
+
+
+def _is_special_particle(alias: Union[str, int]) -> bool:
+    r"""Returns true if a particle is a special particle, and False
+    otherwise."""
+
+    symbol = _dealias_particle_aliases(alias)
+
+    return symbol in _special_particles
```
StanczakDominik: Nice!

StanczakDominik:
I'll handle the CircleCI html failure if you don't mind. It's probably a case of having to install graphviz to build those docs...

plasmapy/atomic/parsing.py
namurphy: I may have learned this from you!

plasmapy/atomic/classes.py
namurphy: Good catch!  The previous can still be simplified/shortened using this strategy.

I'm still unlearning Fortran, it would seem!

StanczakDominik:
The latex builder needed the same thing, I think. Gonna push in two minutes if that works.

namurphy:
> The latex builder needed the same thing, I think. Gonna push in two minutes if that works.

Sure, thank you!

plasmapy/atomic/classes.py
namurphy: I'm trying to figure out a way to set it up so that if someone tries to do
```Python
self.is_antimatter = 'maybe?'
```
then they will end up causing an `AttributeError`.   I think we can do that with the [`@property` decorator and setters](http://samwize.com/2012/09/19/how-you-should-write-getter-slash-setter-for-python/).

StanczakDominik:
Okay, that should do it for the documentation builders! Go forth and write tests! :smiley:

namurphy:
Here's a prototype of what I was trying to say:
```Python
@property
def is_antimatter(self):
    return self._is_antimatter

@is_antimatter.setter
def is_antimatter(self, value):
    raise AttributeError
```
This then leads to:
```
>>> H = Particle('H')
>>> H.is_antimatter
False
>>> H.is_antimatter = 'maybe?'
-----------------------------------------------------------------------
AttributeError                        Traceback (most recent call last)
<ipython-input-20-17c3338acb8e> in <module>()
----> 1 H.is_antimatter = 'maybe?'

AttributeError: can't set attribute
```

Probably should try to find a way to automate setting this up for all the attributes.

An alternative to using a class would be a `NamedTuple` since those are immutable and can lead to similar behavior, but these don't allow for inheritance and all those other nice class properties that we might want eventually.

StanczakDominik:
I have no idea how one would optimally go about automating this :disappointed:

plasmapy/physics/tests/test_transport.py
```diff
@@ -37,7 +37,7 @@
 
 
 def count_decimal_places(digits):
-    '''Return the number of decimal places of the input digit string'''
+    r'''Return the number of decimal places of the input digit string'''
```
StanczakDominik: If I've learned anything from rebasing and merging the other heavy pull requests we've got going now... PLEASE, PLEASE PLEASE keep PRs self-contained and don't start refactoring functions in other files (I'm totally guilty of this there!) unless you really need to change something meaningful. This kind of change can go with the next pull request that touches `test_transport.py`.

plasmapy/physics/transport.py
```diff
@@ -548,7 +548,7 @@ def __init__(self, T_e, n_e, T_i, n_i, ion_particle, m_i=None, Z=None,
             self.theta = self.T_e / self.T_i
 
     def resistivity(self):
-        """
+        r"""
```
StanczakDominik: oh no

plasmapy/physics/tests/test_transport.py
namurphy: You are right, of course.  I should have listened to the voice inside my head that was telling me this was a bad idea while I was doing it!  I'll revert it.

namurphy:
Okay, I rebased to remove the evil commits like `4bdc3aa` that changed docstrings to raw docstrings in a bunch of different places.  I'll keep this pull request focused on the atomic subpackage from now on.  

When rebasing, I always feel like I have to be careful so that I do not accidentally destroy the universe.

StanczakDominik:
Well, the universe lives, but I think what just broke is that you rebased on top of your local master, and that wasn't synced to upstream (repository) master and commits from your previous pull request got added in here. My apologies, I should have said `git fetch --all`, then `git rebase -i upstream/master`. Please allow me to fix this in a few minutes.

StanczakDominik:
Okay. I think I've got this.

Please do
```bash
git fetch --all
git reset --soft origin/particle-class
```
to apply the history changes to your local repo.

I'm assuming you've got `git@github.com:namurphy/PlasmaPy.git` set up as your `upstream` remote.

How about I get my thesis stuff wrapped up, and then we talk rebases and git stuff on this/next telecon?

plasmapy/atomic/classes.py
```diff
@@ -40,6 +39,12 @@
 class Particle():
     r"""A class for individual particles or antiparticles."""
 
+    def __repr__(self):
+        return f'<Particle "{self._particle_symbol}">'
+
+    def __str__(self):
+        return f"{self._particle_symbol}"
```
StanczakDominik: Good idea on the dunder methods. I tend to like my class to have `__init__` at the top, as that's what gets executed first and so on, but I guess this is fine as well.

Makes me wonder... we could also overload addition or other operators for these particles. Something something reactions?

plasmapy/atomic/classes.py
```diff
@@ -90,20 +96,21 @@ def __init__(self,
             self._lepton_number = _Particles[particle]['lepton number']
             self._baryon_number = _Particles[particle]['baryon number']
             self._integer_charge = _Particles[particle]['charge']
-            self._mass = _Particles[particle]['mass']
-
-            self._is_stable = _Particles[particle]['is_stable']
-
+            self._mass = _Particles[particle].get('mass', None)
```
StanczakDominik: I'm not sure, but if the only thing you expect to get in these cases is an exact numerical value or `None` if that's not in there... make sure you're aware of `collections.defaultdict`!

plasmapy/atomic/classes.py
```diff
@@ -144,109 +151,145 @@ def __init__(self,
...
+    def isotope(self) -> str:
+        r"""Returns the isotope symbol, or raises an InvalidIsotopeError
+        if the particle does not correspond to an isotope."""
+        if not self._isotope_symbol:
+            raise InvalidIsotopeError(self._isotope_errmsg)
         return self._isotope_symbol
 
     @property
-    def ion(self):
+    def ion(self) -> str:
```
StanczakDominik: All hail the mighty type hints!

plasmapy/atomic/classes.py
```diff
@@ -144,109 +151,145 @@ def __init__(self,
...
+        r"""Returns the baryon number, or raises a MissingAtomicDataError if
+        the baryon number is unavailable."""
+        if self._baryon_number is None:
+            raise AtomicError(
+                f"The baryon number for '{self.particle}' is not "
+                f"available.")
+        return self._baryon_number
+
+    @property
+    def half_life(self) -> u.Quantity:
```
StanczakDominik: I kind of wonder whether we couldn't do `-> 1/u.s` here. You know, to signalize that this returns an inverse of time.

I think @units.quantity_input also plays well with that in most cases (*grumble grumble electronvolts*), but I'll have to check back. I think I should have some time to play around with unit checks and annotations next week, I feel like it's sort of finishing gestating in my head.

plasmapy/atomic/names.py
```diff
@@ -454,6 +454,8 @@ def _extract_integer_charge(argument: str) -> Tuple[str, int]:
 
     """
 
+    warnings.warn("_extract_integer_charge is deprecated.", DeprecationWarning)
+
```
StanczakDominik: Well... this is a good reason to make this PR less self-contained, I guess. What do you think is the best way to do this after these changes?

plasmapy/atomic/parsing.py
```diff
@@ -97,19 +92,31 @@ def _dealias_particle_aliases(alias: Union[str, int]) -> str:
...
 
 
+def _invalid_particle_errmsg(argument, mass_numb=None, Z=None):
+    r"""Returns a string with an appropriate error message for an
+    InvalidParticleError."""
+    errmsg = f"The argument {argument} "
+    if mass_numb is not None:
+        errmsg += f"with mass_numb = {mass_numb} "
+    if mass_numb is not None and Z is not None:
+        errmsg += "and "
```
StanczakDominik: :laughing: beautiful!

plasmapy/atomic/particles.py
```diff
@@ -119,10 +119,13 @@ def _create_Particles_dict() -> typing.Dict[str, dict]:
                 Particles[thing]['mass'] = 3.167_47e-27 * u.kg
                 Particles[thing]['half-life'] = 2.906e-13 * u.s
 
-    for thing in ['p', 'p-']:
+    for thing in _neutrinos + _antineutrinos:
+        Particles[thing]['mass'] = None
```
StanczakDominik: I'm not sure why we're doing this.

plasmapy/atomic/tests/test_classes.py
```diff
@@ -1,8 +1,143 @@
...
+]
+
+
+@pytest.mark.parametrize("arg, kwargs, expected_dict", test_Particle_table)
+def test_Particle_class(arg, kwargs, expected_dict):
+    r"""Test required properties of items in _Particles dictionary."""
+
+    # To allow tests of a Particle class to continue after the first
+    # exception, error messages will be appended as new lines to errmsg.
+    errmsg = ""
```
StanczakDominik: Well, that certainly looks like a novel approach!

plasmapy/atomic/tests/test_classes.py
```diff
@@ -1,8 +1,143 @@
...
+
+
+@pytest.mark.parametrize("symbol", _special_particles)
+def test_Particle_special(symbol):
+    r"""Test the properties of special particles that do not
+    correspond to elements."""
+
+    particle = Particle(symbol)
+
+    assert particle._atomic_symbol is None, \
```
StanczakDominik: I have a sick and twisted idea on how to stick this into `parametrize` using `lambda`s.

But perhaps that's overkill. I could try implementing it on my copy of this branch if you'd like, it shouldn't take long.

StanczakDominik:
Also, thanks for teaching me to pay closer attention to commit messages via increasing their comedic value above the norm!

> Does the failure of a PR's rebase on GitHub set off a government shutdown in the US?

plasmapy/atomic/classes.py
namurphy: @Cadair had brought this idea up a while ago too, to have something like:
```Python
>>> Be-8 > 2 * He-4
<whatever the energy output is>
```
I think he said that the thing to be careful about was the order of operations, in that the `-4` would signify assigning a mass number to an element.  This might be worth creating an issue for, since it won't happen before the v0.1.0 feature freeze.  This would be neat functionality, though.

namurphy: As an aside, Be-8 really doesn't like being Be-8.  It's half-life is 6.7e-17 seconds before it turns into two alpha particles.  Despite that incredibly short half-life, Be-8 is an important part of the triple-alpha process for nucleosynthesis in stars.

plasmapy/atomic/classes.py
namurphy: I'll look into that!

plasmapy/atomic/classes.py
namurphy: Interesting point.  In terms of physics, it does make a lot of sense to use inverse seconds as a more natural unit.  The phrase "half-life" implies a timescale and so from that name I think most people would expect seconds.  I tried searching for "inverse of half-life" since I didn't know the name of it, and the first term I found was "radioactive decay constant", though that's less widely used than "half-life".  Do we think it would be worth having another attribute called `decay_constant` with attributes of inverse seconds?

plasmapy/atomic/names.py
namurphy: My plan is to remove these private functions as part of this pull request since `parsing.py` and the `Particle` class should make these private functions no longer needed, so the DeprecationWarnings are primarily for my short-term benefit.  PyCharm also uses a strikethrough font to denote deprecated functions which makes them easier to find!

plasmapy/atomic/parsing.py
namurphy: Useful error messages are useful!

plasmapy/atomic/tests/test_classes.py
namurphy: I'm planning on deleting this test since `test_Particle_class` should do a much better job of testing this.

namurphy:
Question for the universe...  For `ion = Particle(symbol)`, would we prefer an API like:
```Python
>>> p.m
>>> p.q
>>> p.Z
```
or an API like:
```Python
>>> p.mass
>>> p.charge  # or p.electric_charge
>>> p.integer_charge
```
❓  I'm inclined for the second option because I personally find it more readable.  Another possibility would be to allow both.

plasmapy/atomic/classes.py
StanczakDominik: Oh crudddd yeah I really wasn't thinking here. Obviously it returns in seconds, haha! ^^'

What I meant was that we could change the return type hint from `-> u.Quantity`, as that's really generic, to `-> u.s`, as that's a `Quantity` and it's also more specific and informative.

plasmapy/atomic/names.py
StanczakDominik: It does? TIL, thanks! That's a good tactic, then. #stealingthisbuild

plasmapy/atomic/classes.py
```diff
@@ -163,6 +171,66 @@ def __init__(self,
             f"The particle '{self.particle}' is not an ion, so this"
             f"attribute is not available.")
 
+    def __repr__(self):
+        r"""Returns a string of the call that would recreate this class."""
```
StanczakDominik: Actually, `recreate this object`. We're not doing metaclasses here yet! :)

plasmapy/atomic/classes.py
```diff
@@ -39,6 +39,11 @@
 class Particle():
     r"""A class for individual particles or antiparticles."""
 
+    # TODO: Write an actual docstring of wonder and amazement
+    # TODO: Add a method to get the reduced mass from this and another particle
+    # TODO: Write a decorator to turn atomic inputs (a particle string, Z,
+    #       and mass_numb) into a Particle.
```
StanczakDominik: And then you could even put the decorator on the reduced mass method!

plasmapy/atomic/classes.py
```diff
@@ -369,3 +369,31 @@ def spin(self) -> Union[int, float]:
...
+                f"{self.particle} is not available.")
+
+        try:
+            mass_that = other.mass
+        except MissingAtomicDataError:
+            raise MissingAtomicDataError(
+                f"Unable to find the reduced mass because the mass of "
+                f"{other.particle} is not available.")
+        except AttributeError:
+            raise
```
StanczakDominik: I don't know too too much about exception handling, but is there really that much point to this sort of "catch exception, do nothing in particular with it"?

plasmapy/atomic/classes.py
```diff
@@ -7,64 +7,19 @@
...
-    '_spin',
-    '_generation',
-    '_periodic_table_group',
-    '_periodic_table_period',
-    '_charge',
-    '_electric_charge',
-]
+from .elements import _Elements
+from .isotopes import _Isotopes
+from .particles import _Particles, ParticleZoo
```
StanczakDominik: Import what now? :laughing:

plasmapy/atomic/classes.py
```diff
@@ -154,21 +109,54 @@ def __init__(self,
...
+
+        if not isinstance(argument, (int, str)):
+            raise TypeError(
+                "The first positional argument when creating a Particle "
+                "object must be either an integer or string.")
+
+        # If the data is missing, then the private attribute should still
+        # exist but just be set to None.  This initialization had previously
+        # been done in a loop using exec on a string, but this does not play
+        # well with static type checkers such as PyCharm.
```
StanczakDominik: Good choice going back on that, I think. It was probably too fancy for its own good, though kudos on the idea.

plasmapy/atomic/particles.py
```diff
@@ -1,34 +1,110 @@
...
+    def antiparticles(self) -> Set[str]:
+        r"""Returns a set of strings representing antiparticles."""
+        return self._taxonomy_dict['antimatter']
+
+    @property
+    def everything(self) -> Set[str]:
+        r"""Returns a set of strings representing all particles and
+        antiparticles"""
+        return \
+            self._taxonomy_dict['matter'] | self._taxonomy_dict['antimatter']
```
StanczakDominik: This is... interesting. And a clever way to do this, too! I think I like it.

plasmapy/atomic/particles.py
StanczakDominik: I'm still not really sure why we're doing this. Neutrinos aren't massless, right? I remember hearing about that on A Capella Science. Is `mass` the mass number?

plasmapy/atomic/tests/test_parsing.py
```diff
@@ -255,7 +255,7 @@ def test_parse_InvalidParticleErrors(arg, kwargs):
         _parse_and_check_atomic_input(arg, **kwargs)
 
 
-@pytest.mark.parametrize('arg', _special_particles)
+@pytest.mark.parametrize('arg', ParticleZoo.everything - {'p+'})
```
StanczakDominik: :clap: 

:clap: 

:clap: 

Beautiful.

Did you seriously just go to all that effort to be able to write "everything but protons" in elegant Python?

That's amazing. Well done! This IS really elegant Python. :laughing:

namurphy: I also did this because I think users studying cosmic rays would end up finding it useful to be able to easily access the sets of leptons, fermions, etc.  This also cleaned up the code in a bunch of different places since we no longer have to worry about a zoo of separate private variables like `_leptons`, `_fermions`, etc..  I should have used a class for this from the beginning!

StanczakDominik: That is, of course, very reasonable. :+1:

namurphy:
Also, this closes #174.

namurphy:
Apart from a few small things here and there, this pull request is almost ready to be merged.  I'm going to check a few things, like making sure that the docstrings are sufficient, and probably also make sure that all of the important parts of the code are being tested.  Any final suggestions would be appreciated!

It would be fine to ignore most of the code in `names.py` and `atomic.py` since that functionality got put into the `Particle` class and is going to be replaced soon once there is a decorator that can take atomic input and pass through the corresponding `Particle` to the inner function.

plasmapy/atomic/classes.py
```diff
@@ -603,34 +778,35 @@ def _make_into_set(arg: Union[str, Set, Tuple, List]) -> Set[str]:
...
             return False
 
-        if any:
-            return not categories.isdisjoint(self._categories)
-        else:
-            return categories.issubset(self._categories)
+        if any and categories & self._categories:
+            return True
+
+        return categories <= self._categories
```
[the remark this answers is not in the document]
namurphy: I hadn't really used sets in Python before, and I was happy with how straightforward the operations ended up being!

namurphy:
My understanding is that it's now more or less accepted that neutrinos do have mass since they are able to oscillate between the different flavors, but only upper limits on the mass are known.  I put the masses of the neutrinos as `None` since they are not known.  This means that we may need to do something special for `nuclear_reaction_energy` when the `Particle` class gets implemented in that.

PR #242, opened 2018-01-26
Create @particle_input decorator for Particle class

namurphy:
The purpose of this pull request will be to create a decorator that can take atomic input parameters and pass through the appropriate instance of the Particle class (which was created in #239).  An example of how this may end up working is:
```Python
@particle_input(must_be_element=True)
def atomic_number(particle):
    """The decorator would return particle as the corresponding particle object,
    and would handle all the appropriate exception handling."""
    return particle.atomic_number
```
In this example, the decorator would return `particle` as the `Particle` object based on the input, and then be able to retrieve the atomic number as an attribute of `particle`.  The exception handling would also be taken care of by `@particle_input.`  In principle, this could be used for all sorts of functions and classes that depend on different particles, and would greatly standardize things.


Cadair:
so, as you have proposed / implemented this it only does things with the first argument. I have a suggestion, use annotations to tell the decorator which arguments it should be converting. i.e.

```
@particle_input
def myfunc(arg1, arg2, particle: Particle):
    ...
```

the decorator can then find all the arguments that are annotated with `Particle` and ensure they are of type `Particle`.

StanczakDominik:
I like @Cadair's suggestion. Especially since we can do both that and roll `u.quantity_input` in the future, once we have more annotations in place.

By the way, please *don't* solve the merge conflicts just yet, I see a use for them in about 21 hours. ^^

namurphy:
Yes, @Cadair's suggestion to use annotations sounds great.  I don't have that much experience with decorators or annotations, so this will be a fun learning experience. 😸

plasmapy/atomic/particle_input.py
```diff
@@ -0,0 +1,169 @@
...
+                        f"with @particle_input.")
+
+            # If the number of arguments and keywords annotated with
+            # Particle is exactly one, then the Z and mass_numb keywords
+            # can be used without potential for ambiguity.
+
+            Z = arguments.get('Z', None)
+            mass_numb = arguments.get('mass_numb', None)
+
+            if Z is not None and not isinstance(Z, int):
```
Cadair: Do you need to do this here? Won't the constructor of the `Particle` class do it?

namurphy: I see why you're taking exception to this exception.  The constructor of the `Particle` class does take care of all of this, so I can delete this code.

namurphy:
I also just wanted to check if AppVeyer is set to run tests.  The configuration file in #249 hasn't been merged yet, which is the reason why the AppVeyer tests are failing so far.  The reason why Travis CI tests are failing is because I broke the code. 😹

namurphy:
Awesome!  I broke all the tests!  💯 🥇 🏆 🎆 🎓 🌋

StanczakDominik:
Let's try merging this into master for once. I'm curious to see how it'd look like.

namurphy:
Thanks, @StanczakDominik! There was a file that I renamed that showed up again after the merge, but on first glance everything else looks fine.

namurphy:
I'm running into some errors related to Cython:

> ```
> =================================== ERRORS ====================================
> ______ ERROR collecting plasmapy/physics/tests/test_parameters_cython.py ______
> plasmapy\physics\tests\test_parameters_cython.py:12: in <module>
>     from ..parameters_cython import (thermal_speed,
> E   ModuleNotFoundError: No module named 'plasmapy.physics.parameters_cython'
> ______ ERROR collecting plasmapy/physics/tests/test_parameters_cython.py ______
> ImportError while importing test module 'C:\projects\plasmapy\plasmapy\physics\tests\test_parameters_cython.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> plasmapy\physics\tests\test_parameters_cython.py:12: in <module>
>     from ..parameters_cython import (thermal_speed,
> E   ModuleNotFoundError: No module named 'plasmapy.physics.parameters_cython'
> !!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!
> =========================== 2 error in 3.10 seconds ===========================
> ```

namurphy:
To help me keep track of this, the things I'd like to do are: update the `any` boolean keyword in `Particle.is_category` and `particle_input` to `any_of` which would take a set or list instead; add a small subclass for the periodic table group, etc; update the exceptions in `nuclear_reaction_energy`; and make sure that the test coverage is sufficient.

namurphy:
Oh, and docstrings, as always.

StanczakDominik:
Looks like I won't be able to fit helping out here between work and packing after all. My apologies :(

Seems like a reasonable checklist though.

namurphy:
No worries whatsoever, @StanczakDominik!  Work-life balance is important, so please feel no obligation to work on this in the meantime.  I hope you have a great vacation!

plasmapy/atomic/particle_class.py
```diff
@@ -0,0 +1,701 @@
+r"""The Particle class."""
+
+import numpy as np
+import warnings
+from typing import (Union, Set, Tuple, List, Optional)
+from astropy import units as u, constants as const
```
Cadair: These should be seperate lines (pep8) also `import astropy.units as u` 😉

plasmapy/atomic/particle_class.py
```diff
@@ -0,0 +1,701 @@
...
+        r"""Returns the integer charge of the particle, or raises a ChargeError
+        if the charge has not been specified."""
+        if self._attributes['integer charge'] is None:
+            raise ChargeError(
+                f"The charge of particle {self.particle} has not been "
+                f"specified.")
+        return self._attributes['integer charge']
+
+    @property
+    def charge(self) -> u.C:
```
Cadair: You are mixing units and types as return annotations here, I suggest you decorate the function with `u.quantity_input`  when you annotate with units rather than types.

plasmapy/atomic/particle_input.py
```diff
@@ -0,0 +1,241 @@
...
+
+    Notes
+    -----
+
+    This version of particle_input only works with functions, but should
+    be extended to work with classes and methods.
+
+    """
+
+    def _particle_errmsg(argname: str, argval: str, Z: int = None,
```
Cadair: Do these helper functions need to be defined inside the decorator? I feel like it would be neater if they were outside.

plasmapy/atomic/particle_class.py
namurphy: I had been wondering about the best way to annotate methods that return `Quantity` objects while still keeping track of units, so it's great to know that `u.quantity_input` can handle this.  Thanks!

plasmapy/atomic/particle_input.py
namurphy: I had been debating this too, so I'll define these outside.

plasmapy/atomic/atomic.py
```diff
@@ -1,60 +1,37 @@
-"""Functions that retrieve or are related to elemental or isotopic data."""
+r"""Functions that retrieve or are related to elemental or isotopic data."""
```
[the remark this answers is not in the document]
namurphy: I ran into some problems a while back with doc builds that I had trouble pinpointing, and it turned out that the issue was that there was LaTeX in there but I didn't mark it as a raw docstring.  There were a few other occasions where I added math to a docstring and forgot to put in the `r`, which led to errors that caused problems.  I have been using raw docstrings pretty diligently because if an inexperienced contributor adds an equation to a docstring where there wasn't one before, then they won't run into the same problem.  Plus, if inexperienced contributors see `r"""` at the beginning of docstrings, they may get into that habit without really knowing why, and thus avoid this problem automatically should they write a new docstring with math in it.  Admittedly, doing this for one-line docstrings with no chance of math being added might be too much!

plasmapy/atomic/atomic.py
```diff
@@ -469,109 +365,100 @@ def ion_mass(argument: Union[str, int, Quantity], Z: int = None,
...
-    if is_isotope:
 
-        if isotope == 'H-1' and Z == 1:
-            return const.m_p
-        elif isotope == 'D' and Z == 1:
-            return 3.343583719e-27 * u.kg
-        elif isotope == 'T' and Z == 1:
-            return 5.007356665e-27 * u.kg
+@particle_input(exclude={'neutrino', 'antineutrino'})
+def particle_mass(particle: Particle, *, Z: int = None,
```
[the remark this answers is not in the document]
namurphy: For this I was intending to not allow `Z` and `mass_numb` to be positional arguments, since they're both integers and could get confused with each other.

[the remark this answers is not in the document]
Cadair: oh yeah of course!

namurphy:
Another thought about the `@particle_input` decorator...  It would not be that hard to add in a check against the return type annotation. Suppose we have functions declared like this with the following return annotations:
```Python
def func() -> int:
def func() -> u.kg:
def func() -> u.Quantity:
```
The decorator would then be able to easily check to make sure that the returned value is an instance of the type/class in the return annotation; or to make sure that the returned `Quantity` has the right units. This seems potentially very helpful, but a worry is that it is beyond the scope of what this decorator is intended to do and thus could cause unintended behavior.  On the other hand, it would provide a free check to make sure that the decorated function or method is doing what it is intended to do.  Do we think this is worthwhile to put in?

Some of this functionality is in `@quantity_input`, but there are some functions that do not have any `Quantity` objects as inputs and thus don't work that well with `@quantity_input`.

namurphy:
The functionality within `isotope_mass` and `ion_mass` is a subset of the functionality of `particle_mass`.  Should we remove `isotope_mass` and `ion_mass` and just use `particle_mass` instead?  

This would reduce namespace pollution.  However, the advantage of having `isotope_mass` and `ion_mass` is that it is much more explicit what is being returned, since these will have checks to make sure that the particle is an isotope or ion.

I also thought about renaming `particle_mass` to just `mass`, but this could also lead to some confusion since `mass` could also refer to the total mass in a plasma, for example.  Keeping it as `particle_mass` makes it more clear what the mass is of.

StanczakDominik:
Have we tested how quantity input works on functions which don't have annotations for their arguments? This seems weird to me. It should work, and I don't think it's a good idea to make particle input do everything. After all, stacking decorators is viable, isn't it?

Besides... The more type checks and so on we throw onto our functions, the more their performance will suffer. We'll have to keep that in mind.

I like `particle_mass`. I'm not too sure about the others, but I guess I don't have strong opinions on that either way... Maybe except for, again, programming really defensively and performing a lot of checks at the cost of code complexity and performance.

namurphy:
Here's functionality where I get an error with `u.quantity_input` using the branch from this pull request and Astropy 2.0:
```Python
import astropy.units as u
from plasmapy.atomic import Particle

proton = Particle('p+')

@u.quantity_input
def func(x: Particle) -> u.kg:
    return x.mass

func(proton)
```
This results in
```Python
TypeError: <class 'plasmapy.atomic.particle_class.Particle'> can not be converted to a Unit
```

StanczakDominik:
Ah, I see. Well then... Looks like we could contribute a `@quantity_output` to astropy!

namurphy:
I'm going to merge this shortly so that we can start implementing `@particle_input` in other places in the code.  We'll still need to do a code review of this and the atomic subpackage prior to the v0.1.0 release, so suggestions will continue to be helpful and very much appreciated.  

Let the feature freeze begin! :snowflake: :snowman_with_snow: :snowflake: :snowman_with_snow:  :snowflake:

PR #243, opened 2018-01-29
Standardize import alias for astropy.units as u

### LINKED ISSUE #241: Standardize import alias for astropy.units as u
There are some places in PlasmaPy where we use:
```Python
```
and other places where we use
```Python
from astropy import units
```
As @StanczakDominik and others have pointed out on Matrix, it would be helpful to have one single alias for `astropy.units` that used uniformly throughout the code so that it's more consistent and readable.  In the discussion on Matrix, the preference from a few people was to use `u` for brevity.  This issue is to go through the code and replace occurrences of `from astropy import units` and `units.` with `from astropy import units as u` and `u.`, including in examples in docstrings.

Thank you!
-Nick

nurbu5:
Closes #241

### DIFF
--- a/docs/development/code_guide.rst
+++ b/docs/development/code_guide.rst
@@ -399,7 +399,7 @@ Equations and Physical Formulae
   the physical constants.  For example, the following line of code
   obscures information about the physics being represented:
 
->>> omega_ce = 1.76e7*(B/units.G)*units.rad/units.s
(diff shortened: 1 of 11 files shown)


StanczakDominik:
We should probably also add a note to the coding guidelines on this...

StanczakDominik:
[Huh.](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/development/code_guide.rst#coding-style)

PR #244, opened 2018-01-29
Add periodic table data, get methods

### LINKED ISSUE #173: Add period, group, block, and category for each element
Each element in the periodic table has a [_period_](https://en.wikipedia.org/wiki/Period_(periodic_table)) and a [_group_](https://en.wikipedia.org/wiki/Group_(periodic_table)) related to its chemical properties.  Each row in the periodic table corresponds to a period, whereas each column corresponds to a group.  For example: lithium is in period 2, group 1 and helium is in period 2, group 18.  Elements of the same group have similar chemical properties.  The periodic table contains four [_blocks_](https://en.wikipedia.org/wiki/Block_(periodic_table)) ('s', 'p', 'd', and 'f') named after characteristic orbitals.  Additionally, there are different [categories](https://en.wikipedia.org/wiki/Names_for_sets_of_chemical_elements).  For example, the [periodic table from LANL](http://periodic.lanl.gov/index.shtml) uses the following categories: alkali metal, alkaline earth metal, transition metal, post-transition metal, metalloid, lanthanide, actinide, nonmetal, halogen, noble gas (see also the [IUPAC recommendations for nomenclature of inorganic chemistry from 2005](http://old.iupac.org/publications/books/rbook/Red_Book_2005.pdf)).  

I propose that we add functionality to get period, group, block, and category information for each element as an enhancement for the atomic subpackage.  The steps to do this are:

 - [ ] Add period information to `Elements` dictionary
 - [ ] Add group information to `Elements` dictionary
 - [ ] Add block information to `Elements` dictionary
 - [ ] Add category information to `Elements` dictionary
 - [ ] Create a function to retrieve an element's period (perhaps `periodic_table_period`)
 - [ ] Create a function to retrieve an element's group (perhaps `periodic_table_group`)
 - [ ] Create a function to retrieve an element's block (perhaps `periodic_table_block`)
 - [ ] Create a function to retrieve an element's category (perhaps `periodic_table_category`)

The `Elements` dictionary is in [`plasmapy/atomic/elements.py`](https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/atomic/elements.py).  Each function should have a docstring and tests.  Different function names would be fine if you can think of something more readable.  I did notice some differences in categories from different references, so it'd probably be best to cite what reference we use to get the category information.

Thank you! 🔬 🔭 

Cross-references because I know too many librarians: #52, #104, #105

jams2:
Closes #173 

Added data (period, group, block, category) to _Elements dictionary in elements.py.
Added functions to retrieve, and helper functions to keep it readable/avoid redundancy.
Added corresponding unit tests for all functions


jams2:
Resolving some errors in the tests presently.

StanczakDominik:
Hey, please revert the commit removing all the newlines in docstrings, that's going to break a ton of stuff. :(

jams2:
Sorry @StanczakDominik ! Was following pep8 script suggestions. Obviously overzealous. Will do.

StanczakDominik:
Hey, don't worry about it! Pep8 has messed up plenty of commits before. :D feel free to reopen this whenever you feel like it. If you'd like help with solving those problems, we'll be happy to provide it :D

jams2:
Ok, that was a crash course in git. Sorry about multiple tiny commits, I've figured out how to properly run the testing and coverage locally now. Anyway, looks good on my end now, happy to hear feedback!

StanczakDominik:
We can squash the more *experimental* commits into a single one while merging if you'd like :D

plasmapy/atomic/atomic.py
```diff
@@ -1257,3 +1257,438 @@ def stable_isotopes_for_element(argument: Union[str, int],
...
+        raise TypeError("The argument to periodic_table_category must be " +
+                        "either a string representing the element or its " +
+                        "symbol, or an integer representing its atomic " +
+                        "number.")
+    atomic_symbol = _get_atomic_symbol(argument)
+    category = _Elements[atomic_symbol]["category"]
+    return category
+
+
+def _get_atomic_symbol(argument: Union[str, int]) -> str:
```
namurphy: It looks like there's some duplication here with `atomic_symbol()`, which is in `atomic/names.py` and has most of the same functionality as `_symbol_from_symbol()`, `_symbol_from_name()`, and `_symbol_from_number()`.  One difference is that the atomic symbols are case-sensitive in `atomic_symbol`, which helps reduce ambiguity from things like `'p'` being a proton and `'P'` being phosphorus.  We could probably use `atomic_symbol()` in the `periodic_table_*` functions to be consistent with the other atomic routines.

plasmapy/atomic/tests/test_atomic.py
```diff
@@ -1286,3 +1290,382 @@ def test_extract_integer_charge_warnings(test_input, expected_warning):
     """Test that _extract_integer_charge issues the expected warnings."""
     with pytest.warns(expected_warning):
         _extract_integer_charge(test_input)
+
+
+# argument, expected
+period_table = [
+    ("ca", 4),
```
namurphy: If we use `atomic_symbol`, then these tests would need to be changed to be case-sensitive.  The element names are not case-sensitive, though.

jams2:
@namurphy Sure thing, I'll get onto that.

@StanczakDominik I'll figure out how to squash a few commits on my end.

StanczakDominik:
@jams2 don't worry about squashing, we can do that automatically at merge time!

jams2:
I handed off most of the error handling to names.atomic_symbol() . We could check that the return from _Elements is ok, but it seems redundant as atomic_symbol will return a valid symbol, what do you think?

namurphy:
This looks great, and the AppVeyor test failure is unrelated, so I'll merge it.  Thank you!  -Nick

jams2:
No worries @namurphy . Feel free to tag me in any issues that come up that you think I might be able to help with.

PR #249, opened 2018-02-02
First attempt at an appveyor file

### LINKED ISSUE #246: Add AppVeyor continuous integration testing
We presently use Travis CI and CircleCI for continuous integration testing.  These tests are performed largely in Linux environments, but we should also add tests for Windows environments so that we can fix any compatibility problems sooner rather than later.  In particular, [AppVeyor](https://www.appveyor.com/) tests builds on Windows virtual machines and thus would be able to mostly fill this gap in our testing.  The goal of this issue is to add AppVeyer to the suite of continuous integration tests that are automatically performed with each pull request.

dstansby:
Hello, I hear you want to run tests on windows! Here is a first attempt at that using Appveyor. I'm guessing someone with organisation access will need to enable the build on https://ci.appveyor.com - once that's done and this PR is building I'm happy to play around with it until it works.

Fixes #246

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,12 +23,12 @@ install:
   - "powershell ci-helpers/appveyor/install-miniconda.ps1"
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
   - "activate test"
-  - "pip install -r requirements/automated-code-tests.txt"
+  - "pip install --quiet -r requirements/automated-code-tests.txt"
 
 
 # Not a .NET project, we build plasmapy in the install step instead
 build: false
 
 test_script:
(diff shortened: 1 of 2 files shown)


namurphy:
@dstansby - Thank you once again!  I granted AppVeyer the permissions from the PlasmaPy organization that I think it needs, and also put an AppVeyer badge on our main README.  Did you want to refine the settings any more before we merge this?

lemmatum:
Looks like something in the `_nondim_tc_e_ji_held` and `_nondim_tec_ji_held` functions isn't playing nicely?

lemmatum:
One (possibly crazy) idea for where the error is coming from: Could it be the backslashes for continuation of long equations?

I recently cleaned those up in #191

PR #257, opened 2018-02-08
Cleanup Test_classical_transport

### LINKED ISSUE #250: Cleanup test_transport.py
`class Test_classical_transport` is in dire need of some error strings which return the compared values when a test fails. Best to do this with f-strings.

Nismirno:
Closes #250 
Added error messages to `class Test_classical_transport` to compare values when test fails.


plasmapy/physics/tests/test_transport.py
```diff
@@ -1388,7 +1460,10 @@ def test_ion_thermal_conductivity_by_model(self, model, expected):
                           5.08966116e-06]}.items())
     def test_dictionary(self, key, expected):
         calculated = self.all_variables[key]
-        assert np.allclose(expected, calculated.si.value)
+        testTrue = np.allclose(expected, calculated.si.value)
+        errStr = (f"Expected values of {key} are {expected} and not"
+                  f"{calculated.si.values}.")
```
StanczakDominik: If you take a look at the [test results](https://travis-ci.org/PlasmaPy/PlasmaPy/builds/339017903#L1256), it seems this line is killing all the failing tests due to a typo:
```
.values -> .value
```
If you fix that, we should be good to go with merging this PR!

StanczakDominik:
All right! This is looking neat. Test coverage decreased by a bit, but we could probably ignore that. I'd like to merge this, but if you don't mind, I'd like to squash these commits together as these changes are relatively minor and I'd like to keep the commit history a bit cleaner.

Nismirno:
Sure @StanczakDominik ! 
That would be better to squash them.

StanczakDominik:
All right! Thanks for this one!

Issue #105, opened 2017-09-09
Include half-life data for all isotopes with a known half-life


namurphy:
Half-life data is presently included for only a subset of isotopes, since I wasn't able to find an easily usable database of nuclide stability.  We still need to include half-life data for all nuclides with known half-lives.  It would probably be helpful to include information on the uncertainty of each half-life, and information on the source of each half-life.  

This is related to #104 since the half-life data would go in the HDF5 file containing isotope data.  I'll assign myself to this, but would be happy to work with someone else.

If you have a good, recent table on the stability of nuclides, please post it here!

namurphy:
These could be good sources for half-life and other atomic/nuclear data:
- [NUBASE 2016 reference](https://www-nds.iaea.org/amdc/ame2016/NUBASE2016.pdf) 
- [NUBASE 2016 ASCII formatted table](https://www-nds.iaea.org/amdc/ame2016/nubase2016.txt)
- [Nuclear Wallet Cards](http://www.nndc.bnl.gov/wallet/)

namurphy:
A good first contribution would be to add the half-lives for some isotopes (especially important ones).  It would be a major effort to add all of them.

misupova:
I searched for issues marked as "Good first contribution", and this one attracted my attention. As I am new to contributing, I want to clarify some things.

So as I understood, the first step is to add half-life information in Isotopes dictionary, and then create corresponding function (e.g. isotope_half_life) to retrieve this information, am I getting it correctly?

misupova:
I searched for issues marked as "Good first contribution", and this one got my attention, but I want to clarify some things.

How could I include half-lives for isotopes, for which exact half-life is not known? For example, in one of the the sources we can find >260ns for Al-40, what could be sensible way to include such information in dictionary?

StanczakDominik:
Hey @misupova, that's a pretty good question. I suppose a reasonable solution would be to put in None in that case and return an error saying there's no precisely known half life.

 Or we could put in a "<240ns" string and if the function gets a string instead of a float or quantity, raise an error saying something along the lines of "all we know is it's <240ns, use that as a ballpark estimate if you really need to.".

@namurphy, other suggestions?

PR #260, opened 2018-02-10
Added half-lives for all isotopes with a known half-life

misupova:
Related to issue #105.

Added half-lives, which were taken from this document [NUBASE 2016 ASCII formatted table](https://www-nds.iaea.org/amdc/ame2016/nubase2016.txt). For elements which exact half-lives were not known, but were given in this table in form like '>910 ys', I have added a string with that value. 

atomic.py was modified, so if the function half_life gets a string instead of a number, the warning is raised, saying that there is no exact data available, just as @StanczakDominik suggested.


misupova:
Need help dealing with failing tests.

As I understand, there is assertion error, half_life function should return None for Cn-276. But now there is a value, where there has to be None.

StanczakDominik:
The tests are doing what they were expected to do! They're just testing results as they were expected to behave in the previous version. We'll have to update the tests as well for cases where they're failing. Travis should be printing out their locations for each failure.

namurphy:
Thank you, @misupova!  The tests that are failing were ones that I put in to test cases where the half-life is not known, but these tests are no longer necessary.  I just went in and deleted or commented out these tests from `test_atomic.py`.  I restarted the tests, so if they pass then I think this will be ready to merge.  Thank you!

-Nick

PR #264, opened 2018-02-14
Test against python 3.7-dev

namurphy:
Python 3.7 is now in beta, and is likely to be released around this summer.  We should make sure our 0.1.0 release works with the most recent development version of Python 3.7 since our 0.2.0 release will probably not occur until a few months after the Python 3.7 release.  This pull request adds a line in `.travis.yml` to test against Python 3.7-dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ install:
         - chmod u+x requirements/dev_version_hack.sh
         - source requirements/dev_version_hack.sh
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
+        - export DEBUG=True
         - source ci-helpers/travis/setup_conda.sh
         - "pip install -r requirements/automated-code-tests.txt"
         # building so that we have cython files compiled for tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,14 @@
 language: python
 python:
         - "3.6"
-	- "nightly"
+	- "3.7-dev"
 
 git:
     submodules: false
 
 # Install dependencies
 install:
+        - ./requirements/dev_version_hack.sh
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
         - source ci-helpers/travis/setup_conda.sh
         - "pip install -r requirements/automated-code-tests.txt"
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ git:
 # Install dependencies
 install:
         - chmod u+x requirements/dev_version_hack.sh
-        - ./requirements/dev_version_hack.sh
+        - source requirements/dev_version_hack.sh
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
         - source ci-helpers/travis/setup_conda.sh
         - "pip install -r requirements/automated-code-tests.txt"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ git:
 
 # Install dependencies
 install:
+        - chmod u+x /requirements/dev_version_hack.sh
         - ./requirements/dev_version_hack.sh
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
         - source ci-helpers/travis/setup_conda.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: python
 python:
         - "3.6"
+        - "3.7-dev"
 
 git:
     submodules: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ git:
 
 # Install dependencies
 install:
-        - chmod u+x /requirements/dev_version_hack.sh
+        - chmod u+x requirements/dev_version_hack.sh
         - ./requirements/dev_version_hack.sh
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
         - source ci-helpers/travis/setup_conda.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: python
 python:
         - "3.6"
-        - "3.7-dev"
+        - "3.7.0b1"
 
 git:
     submodules: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,7 @@
 language: python
 python:
         - "3.6"
-	- "3.7-dev"
-        - "3.7.0b1"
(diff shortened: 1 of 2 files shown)


namurphy:
I'm getting an error for Travis CI that says:

> We couldn't find the repository PlasmaPy/PlasmaPy

so there may be a problem with Travis CI at the moment.  In a while we should trigger another Travis CI build so that this issue will hopefully get resolved.

lemmatum:
I think your first try was correct (Here's the [Travis](https://travis-ci.org/PlasmaPy/PlasmaPy/jobs/341531000) from that one) as Travis does find Python 3.7-dev:

```
$ python --version
Python 3.7.0a4+
```

But then astropy's ci-helpers fails because it tries to fetch python3.7-dev from conda and ends up installing Python3.6.3:
```
$ source ci-helpers/travis/setup_conda.sh
==================== Starting executing ci-helpers scripts =====================
--2018-02-14 17:15:31--  https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
Resolving repo.continuum.io (repo.continuum.io)... 104.16.19.10, 104.16.18.10, 2400:cb00:2048:1::6810:120a, ...
Connecting to repo.continuum.io (repo.continuum.io)|104.16.19.10|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 57669415 (55M) [application/x-sh]
Saving to: ‘miniconda.sh’
100%[======================================>] 57,669,415   213MB/s   in 0.3s   
2018-02-14 17:15:31 (213 MB/s) - ‘miniconda.sh’ saved [57669415/57669415]
PREFIX=/home/travis/miniconda
installing: python-3.6.3-h6c0c0dc_5 ...
Python 3.6.3 :: Anaconda, Inc.
```

Further down

```
installation finished.
Package plan for installation in environment /home/travis/miniconda:
The following packages will be DOWNGRADED:
    conda: 4.3.31-py36_0 --> 4.3.27-py36h2866c0b_0
PackageNotFoundError: Packages missing in current channels:
            
  - python 3.7-dev*
We have searched for the packages in the following channels:
            
  - https://repo.continuum.io/pkgs/main/linux-64
  - https://repo.continuum.io/pkgs/main/noarch
  - https://repo.continuum.io/pkgs/free/linux-64
  - https://repo.continuum.io/pkgs/free/noarch
  - https://repo.continuum.io/pkgs/r/linux-64
  - https://repo.continuum.io/pkgs/r/noarch
  - https://repo.continuum.io/pkgs/pro/linux-64
  - https://repo.continuum.io/pkgs/pro/noarch
```

So I would revert this back to `3.7-dev` and then start playing with this line: `- source ci-helpers/travis/setup_conda.sh`.

There may some [flag or something](https://github.com/astropy/ci-helpers) you can throw at `ci-helpers` to force it to use the already installed python3.7 rather than fetching it via conda. I think conda only has stable builds of cpython.

lemmatum:
One hacky idea might be to add a line before `source ci-helpers/travis/setup_conda.sh` which makes `python 3.7-dev` an alias of whatever gets returned from `$ python --version`

lemmatum:
Digging into ci-helpers, I think this line is the culprit
```
if [[ -z $PYTHON_VERSION ]]; then
    PYTHON_VERSION=$TRAVIS_PYTHON_VERSION
fi
```

So if no `$PYTHON_VERSION` is declared it then defaults to whatever we passed to Travis, which is `3.7-dev` and then anaconda can't find that because that is not the actual name of the python version we just installed, it's just an alias that Travis uses.

I'm going to add some lines to `travis.yml` to test for this. If this is indeed the problem then we can add a line which does something like
```
if [[ -z $PYTHON_VERSION ]]; then
    # fetches python --version output
    version="$(python --version 2>&1)"
    echo "Current version is $version"
    # binds output of python --version to PYTHON_VERSION
    PYTHON_VERSION="$version"
    echo "PYTHON_VERSION is now $PYTHON_VERSION"
fi
```

This would be placed before `source ci-helpers/travis/setup_conda.sh`

lemmatum:
Any idea what this rake business is all about?
`The command "rake" exited with 1.`

lemmatum:
Looks like the last commit which didn't have this rake error was fb857af, I think...
https://travis-ci.org/PlasmaPy/PlasmaPy/builds/341539086

lemmatum:
Looks like my hacky bash script isn't working... it doesn't even echo the python version back to Travis. Or is that just suppressed because it is stdout instead of stderr?

lemmatum:
Progress! The script seems to sort of work, although it needs to be parsed out to just give the version numbers with no spaces

```
PackageNotFoundError: Packages missing in current channels:
            
  - python Python*
```

lemmatum:
So version number now gets correctly passed using the hack script, and 3.6 tests are building again, but conda doesn't like 3.7. Does anyone know of a way to force conda to fall back on the system installed version of python?

lemmatum:
We are now indeed failing [here](https://github.com/astropy/ci-helpers/blob/e80308d5a142b293f71bb0c0a0f5de34157dddc8/travis/setup_dependencies_common.sh#L74) where conda tries to `conda create -n test python=3.7.0a4+` and then throws `CondaValueError: invalid package specification: python=3.7.0a4+`.

Python 3.7-dev is definitely in Travis's PATH, but it is behind miniconda's version of python
`++++PATH=/home/travis/miniconda/bin:/home/travis/virtualenv/python3.7-dev/bin:`

Maybe we can force conda to use the 3.7-dev version by rearranging PATH and symlinks, [like so?](https://dev.to/bgalvao/conda--dealing-with-conflicting-pythons-in-your-system-62n)

lemmatum:
TODO:
- [ ] Check why we are using ci-helpers in the first place
- [ ] Setup a .sh script or a line in travis.yml which checks whether `$TRAVIS_PYTHON_VERSION` has `dev` in its name. If it does then avoid ci-helpers when setting up the virtualenv, if it doesn't then run ci-helpers as usual.

namurphy:
I wonder if we can figure out how to test against astropy-dev as well, which we tried to figure out how to do in #172.

Cadair:
for astropy-dev with ci-helpers you can just do `ASTROPY_VERSION='dev'`?!

namurphy:
I'm closing this because figuring out a way to test against the development version of Python 3.7 is probably not worth the effort since it's going to be released in ~2-3 months anyway.

PR #265, opened 2018-02-14
Particle.reduced_mass() as a module level function

### LINKED ISSUE #262: Create an `atomic.reduced_mass` function from existing `Particle` method
The issue is simple. Take https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/atomic/classes.py#L714 this method and turn it into a module level `reduced_mass` function that can be called outside of the object oriented `Particle` interface, like `ion_mass`. Simplify the method by calling the function. Simple and helps out #255 !

ritiek:
Closes #262.

Should we the name module level function as `reduced_mass()` or something else (like `get_reduced_mass()`, the way it is in this PR)?

I think naming both `Particle.reduced_mass()` object method and module level function as `reduced_mass()` could confuse beginners, just my opinion however.

Let me know if there is anything that could be better!

### DIFF
--- a/plasmapy/atomic/classes.py
+++ b/plasmapy/atomic/classes.py
@@ -27,6 +27,7 @@ from .particles import _Particles, ParticleZoo
 
 # TODO: Write a decorator to turn atomic inputs into a Particle.
 
+
 def get_reduced_mass(mass_this, other, Z=None, mass_numb=None) -> u.kg:
     r"""Finds the reduced mass between two particles, or will raise a
     MissingAtomicDataError if either particle's mass is unavailable or
@@ -775,7 +776,6 @@ class Particle:
                 f"{self.particle} is not available.") from None
 
         return get_reduced_mass(mass_this, other, Z=Z, mass_numb=mass_numb)
-
 
     def is_category(self, *categories, any=False,
                     exclude: Union[Set, Tuple, List] = set()) -> bool:
--- a/plasmapy/atomic/classes.py
+++ b/plasmapy/atomic/classes.py
@@ -28,7 +28,7 @@ from .particles import _Particles, ParticleZoo
 # TODO: Write a decorator to turn atomic inputs into a Particle.
 
 
-def get_reduced_mass(mass_this, other, Z=None, mass_numb=None) -> u.kg:
+def reduced_mass(mass_this, other, Z=None, mass_numb=None) -> u.kg:
     r"""Finds the reduced mass between two particles, or will raise a
     MissingAtomicDataError if either particle's mass is unavailable or
     an AtomicError for any other errors.  The other particle may be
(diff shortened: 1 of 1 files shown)


StanczakDominik:
I think we're going to have some overlap anyway. I wouldn't mind having them under the same name. I mean, numpy does that, so does matplotlib...

ritiek:
Okay, I'll update them with `reduced_mass()`.

plasmapy/atomic/classes.py
```diff
@@ -736,26 +775,7 @@ def reduced_mass(self, other, Z=None, mass_numb=None) -> u.kg:
...
-            try:
-                mass_that = other.to(u.kg)
-            except Exception as exc:  # coveralls: ignore
-                raise AtomicError(
-                    f"{other} must be either a Particle or a Quantity or "
-                    f"Constant with units of mass in order to calculate "
-                    f"reduced mass.") from exc
-
-        return (mass_this * mass_that) / (mass_this + mass_that)
+        return get_reduced_mass(mass_this, other, Z=Z, mass_numb=mass_numb)
```
StanczakDominik: First argument should probably be self.mass here

StanczakDominik:
Seems like Travis is having issues right now so it's hard to judge this without running it!

plasmapy/atomic/classes.py
ritiek: Would it also be a good idea to pass `self.particle` since it will be used when raising `MissingAtomicDataError` when relevant code moves to module level function from [here](https://github.com/PlasmaPy/PlasmaPy/pull/265/files#diff-cc7af528697aead5e7d3524336ce958aR776)?

StanczakDominik: Actually, just self, probably, and then make the checks in the module level functions accept particles and strings both! The particle input decorator should handle both eventually, but in the meantime, let's do it that way!

namurphy:
Thank you for working on this, @ritiek!  I would vote for calling the module level function `reduced_mass` since that's the shortest name that indicates what the function is doing.  My other quick suggestions are that it would be helpful to put this function at the end of `atomic.py` instead of `classes.py`, and that there probably needs to be an import statement in `atomic/__init__.py`.  Having a separate function for this will be really helpful!

plasmapy/atomic/classes.py
ritiek: If we directly pass `self` from `Particle.reduced_mass()` to module level `reduced_mass()`, it works when calling `Particle.reduced_mass()` method but we'll get an error when passing `ion_mass` through the same module level `reduced_mass()`:

For example:
```python
>>> from plasmapy.atomic import classes, atomic
>>> ion = atomic.ion_mass(10)
>>> classes.reduced_mass(ion, 2)
```
```
AttributeError: 'Quantity' object has no 'mass' member
```

```python
def reduced_mass(particle_type, other, Z=None, mass_numb=None) -> u.kg
    ...
     try:
         mass_this = particle_type.mass.to(u.kg)  # <-- here
     except MissingAtomicDataError:
         raise MissingAtomicDataError(
            f"Unable to find the reduced mass because the mass of "
            f"{particle_type.particle} is not available.") from None
    ...
```

How should we go about it?

namurphy:
Thank you again, @ritiek!  I'm going to merge this now, and then resolve some probable conflicts with #242 so that #242 can be merged soon too.

PR #272, opened 2018-02-22
Customize pycodestyle, pydocstyle, and pep8speaks

namurphy:
The [pycodestyle](http://pycodestyle.readthedocs.io/en/latest/) and [pydocstyle](http://www.pydocstyle.org/en/2.1.1/) packages check code and docstrings against style guidelines.  This PR updates `setup.cfg` to include some configuration options for these two packages.  The code and docstring style customizations that I chose are not the ones that we want to go with, so other suggestions on ones to include or exclude would be great.  Alternatively, we could try these out and see how much we like or dislike them.  The most important things are consistency and readability.

The extra [pycodestyle error codes](http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) are:

>  - E226: missing whitespace around arithmetic operator
>  - E241: multiple spaces after ‘,’
>  - E242: tab after ','
>  - E704: multiple statements on one line (def)
>  - W504: line break after binary operator

The extra [pydocstyle error codes](http://www.pydocstyle.org/en/2.1.1/error_codes.html) are:
    
> Include:
>  - D212: Multi-line docstring summary should start at the first line
>  - D402: First line should not be the function's "signature"
>  - D413: Missing blank line after last section
>    
> Ignore:
>  - D202: No blank lines allowed after function docstring
>  - D302: Use u""" for Unicode docstrings  (not necessary for Python 3.6+)
>  - D412: No blank lines allowed between a section header and its content

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,14 @@
+message:
+    opened:
+        header: "Hello @{name}! Thanks for submitting your pull request."
+        footer: "PlasmaPy follows the [PEP8 style guide](https://www.python.org/dev/peps/pep-0008/), which may be checked locally using [pycodestyle](http://pycodestyle.pycqa.org/en/latest/intro.html)"
+    updated:
+        header: "Hello @{name}! Thanks for updating your pull request."
+        footer: ""
+    no_errors: "Congratulations! There are no PEP8 issues in this pull request. :smile_cat:"
+
+scanner:
+    diff_only: True
+
+only_mention_files_with_errors: True
+descending_issues_order: False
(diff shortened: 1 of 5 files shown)


namurphy:
The extra pycodestyle options that I like most are the missing whitespace around operators (which I think was brought up by @lemmatum a long time ago as a way to make math code much more readable), and warning if the line break is after a binary operator (which is very helpful for subtraction especially).

The pydocstyle option that I like most is letting there be a blank line after a docstring (which I sometimes do to separate different paragraphs of code).

These are just my own preferences.  It should definitely be a community decision. :smile_cat:

PR #278, opened 2018-02-24
Create a test helper function

namurphy:
Many of the tests for the atomic and physics subpackages involve calling a function with different positional and keyword arguments, and then making sure that the actual results meet the expected results.  Sometimes the comparisons are between `Quantity` objects.  Sometimes the tests are to check that certain inputs lead to certain exceptions or warnings.  When warnings are issued, sometimes there are simultaneous checks that the returned value is what is expected also.  Occasionally these tests are run while a cat is sitting on half of a laptop's keyboard.  As a consequence, there is a lot of duplication of code, including error messages.

The purpose of this pull request is to create a general purpose test helper.  Ideally, the inputs for the tests will be a list of tuples, where the tuples contain the function to be tested, the positional arguments, the keyword arguments, and the expected result (which could be an exception, a warning, or a tuple containing a warning and the expected result).  This function should take care of comparisons between different types and create useful error messages for the different tests.  This is still a work in progress (as are we all), but hopefully this will make a bunch of the tests more straightforward.


namurphy:
Questions I've been thinking about:
 - Does the functionality I'm thinking of exist somewhere else?
 - Is what I'm thinking of the best way to organize these tests?
 - Would an object oriented approach work better than a functional approach?

namurphy:
Right now I'm trying to figure out how to print only part of the traceback, since the line where `run_test` is called just adds clutter and extraneous information.  I'm going to start digging into traceback objects to try and learn how to do this.

namurphy:
To be more specific, I'd like to trim the traceback to not include the parts associated directly with the test helper function since that adds a lot of clutter and makes it more difficult to find the root cause of the error.  The parts that show up (hopefully) in bold are what I'd like to exclude.

```Python
In [10]: def f():
    ...:     raise ValueError
    ...: 

In [11]: run_test(f, (), {}, TypeError)
/media/Scratch/namurphy/Projects/PlasmaPy/plasmapy/utils/pytest_helpers.py in run_test(func, args, kwargs, expected_outcome, rtol, atol)
    188         try:
--> 189             result = func(*args, **kwargs)
    190         except expected_exception:

<ipython-input-10-1e378d1a95b8> in f()
      1 def f():
----> 2     raise ValueError

ValueError: 

**The above exception was the direct cause of the following exception:

UnexpectedExceptionError                  Traceback (most recent call last)
<ipython-input-11-f73da0a01b25> in <module>()
----> 1 run_test(f, (), {}, TypeError)

/media/Scratch/namurphy/Projects/PlasmaPy/plasmapy/utils/pytest_helpers.py in run_test(func, args, kwargs, expected_outcome, rtol, atol)
    197                 f"did not raise {exc_str(expected_exception)} as expected, "
    198                 f"but instead raised {exc_str(unexpected_exception)}."
--> 199             ) from exc_unexpected_exception
    200         else:
    201             raise MissingExceptionError(**

UnexpectedExceptionError: Running the command:

  f()

did not raise a TypeError as expected, but instead raised a ValueError.

```

StanczakDominik:
Bloody hell... I missed that you added the ready for review label, and now this merge request has broken docs on master :( before I start looking into it... If you need something reviewed, just say so, or you can use the review request button and assign me, I won't mind - ever. On my part, I'm sorry I missed the label.

StanczakDominik:
Okay, I apologize for jumping to conclusions again. I think it's a bug upstream:
https://github.com/rtfd/readthedocs.org/issues/3842
We can only hope they fix it soon, I think.
Still, you're getting your code review. :D

plasmapy/atomic/tests/test_atomic.py
```diff
@@ -48,352 +48,489 @@
...
-    ('H 2+', InvalidParticleError),
+metatable = [
+    (atomic_TypeError_funcs_table, atomic_TypeError_badargs, TypeError),
+    (atomic_ParticleErrors_funcs_table, atomic_ParticleError_badargs, InvalidParticleError),
 ]
 
+for funcs, badargs, error in metatable:
+    for func in funcs:
+        for badarg in badargs:
+            all_tests += [[func, badarg, error]]
```
StanczakDominik: Looking at this, I'm afraid I might not be able to find my way through any failing tests using this framework anymore...

plasmapy/utils/pytest_helpers.py
```diff
@@ -93,35 +207,134 @@ def run_test(func: Callable, args: Any = (), kwargs: Dict = {}, expected_outcome
...
+            (func, 3, {'raise_exception': True}, ValueError),
+            (func, 4, {'issue_warning': True}, UserWarning),
+            (func, 5, {'issue_warning': True}, (5, UserWarning)),
+        ]
+
+        @pytest.mark.parametrize('inputs', inputs_table)
+        def test_func(inputs):
+            run_test(inputs)
+
+    """
```
StanczakDominik: This example is pretty darn long. Do you think it would be helpful to let people play with it via jupyter notebook?

plasmapy/utils/pytest_helpers.py
```diff
@@ -165,43 +374,332 @@ def exc_str(ex: Exception) -> str:
...
+    The inputs may also be passed in as a whole as a `tuple` or `list`.
+
+    >>> run_test_equivalent_calls(f, -1, 1)
+    >>> run_test_equivalent_calls([f, -1, 1])
+
+    If `require_same_type` is `False`, then an exception will not be
+    raised if the results are of different types.
+
+    >>> run_test_equivalent_calls(f, -1, 1.0, require_same_type=False)
+
```
StanczakDominik: Likewise here... I think we need a whole example page on the pytest_helpers module.

plasmapy/utils/tests/test_pytest_helpers.py
```diff
@@ -38,47 +72,53 @@ def test_call_string(f, args, kwargs, expected):
...
+    [return_arg, u.kg, {}, u.g, u.UnitsError],
+    [return_arg, u.C, {'should_warn': True}, (u.C, UserWarning), None],
+    [adams_number, 1, {'x': 1}, u.pc, u.UnitsError],
 
-    (issue_warning, 12, {'y': 1}, PlasmaPyWarning, None),
-    (issue_warning, (13,), {'y': 1}, PlasmaPyWarning, None),
-    (issue_warning, (14, 1), {}, PlasmaPyWarning, None),
+    [return_arg, Particle('p+'), {}, Particle('proton'), None],
+    [return_arg, Particle('e+'), {}, Particle('e-'), UnexpectedResultError],
+    [return_arg, Particle('mu+'), {}, type, InconsistentTypeError],
```
StanczakDominik: Is this what you changed `__name__` for?

plasmapy/utils/tests/test_pytest_helpers.py
```diff
@@ -88,13 +128,235 @@ def adams_number(*args, **kwargs):
...
 )
 def test_run_test(f, args, kwargs, expected, whaterror):
-    if whaterror is None:
-        run_test(f, args, kwargs, expected)
-    else:
-        with pytest.raises(whaterror, message = (
-                f"run_test did not raise an exception for:\n\n"
-                f"  {call_string(f, args, kwargs)}\n\n"
-                f"with expected = {expected}")):
+    """
```
StanczakDominik: Looks like this could use `None` as the default `whaterror`!

PR #282, opened 2018-02-27
Increase pycodestyle maximum line length to 99

namurphy:
The topic of the maximum line length came up in a recent code review session.  The PEP 8 limit is 79 characters because (1) really long lines adversely impact readability, (2) wrapping of long lines by editors disrupts the flow, and (3) this line length makes it easier to display two windows of code side by side.  The 79 character limit resulted in some awkward line breaks using backslashes in the Particle
class, whereas the code would be more readable were it just be on the same line.  Those of us who were present came to a consensus that we should increase the line length so that we can avoid awkward line breaks.

This commit increases the maximum line length by ten characters so that we can avoid these awkward line breaks while still avoiding really long lines.  Comments and docstrings should continue to be wrapped at 72 characters to make things more readable, and the 79 character limit is still suggested.

### DIFF
--- a/docs/development/code_guide.rst
+++ b/docs/development/code_guide.rst
@@ -214,12 +214,12 @@ Coding Style
   * PEP 8 compliance may be checked locally using
     [pycodestyle](http://pycodestyle.pycqa.org/en/latest/).
 
-  * PEP 8 limits lines of code to 79 characters in order to improve
-    readability and allow side-by-side comparisons between code
-    displayed in different windows.  The maximum line length for Python
-    code in PlasmaPy is 89 characters, though lines longer than 79
-    characters should be used sparingly.  Docstrings and comments
-    should generally be limited to 72 characters.
+  * Line lengths should be chosen to maximize the readability and
+    elegance of the code.  The maximum line length for Python code in
+    PlasmaPy is 99 characters.
+
+  * Docstrings and comments should generally be limited to
+    72 characters.
 
 * Follow the existing coding style within a subpackage.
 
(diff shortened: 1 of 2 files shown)


Cadair:
89?!

> For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters)

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

StanczakDominik:
I wasn't aware that pep8 has guidelines on possible choices for line lengths. Let's go along with that and change this to 100 characters.

Also, if anyone would like to stay with the good old '79, we can do that. Just say so.

namurphy:
My initial choice of 89 might or might not have been because I like prime numbers. :3rd_place_medal: 

I personally find lines longer than ~80-90 characters to be less readable, but having it at 79 has made me use lots of awkward and clunky line breaks so if it were a choice between 79 and 99 then I'd vote 99.

namurphy:
But as I mentioned during the code review, it would be great if a line ending with an `=` implicitly implied a continuation to the next line if the next line was indented:
```Python
    very_long_variable_name_which_is_very_long =
        very_long_other_variable_name_which_is_very_long
```
as opposed to using a backslash:
```Python
    very_long_variable_name_which_is_very_long = \
        very_long_other_variable_name_which_is_very_long
```

namurphy:
There were no objections to this when I asked about it, so I'm merging this.

PR #281, opened 2018-02-27
Particle class code review fixes

StanczakDominik:
A few minor improvements to `particle_class.py` as discussed today during the code review session.

PR #284, opened 2018-02-28
Update and refactor Particle class

namurphy:
This PR continues to implement the changes from our first code review session on the Particle class, with many helpful suggestions from @lemmatum and @StanczakDominik (see also #281).  The changes I made were to:
  - Create `Particle.electron_number` attribute
  - Remove `Particle.reduced_mass` in favor of `reduced_mass` function that was recently added to atomic.py (this changes the API)
 - Instead of defining lists of keys to be used, just get them directly from the `dict` objects
 - Move proton attributes to `special_particles.py`
 - Move `_PeriodicTable` `namedtuple` to `elements.py`
 - Update `neutron_number` attribute so that it returns `1` for a neutron
 - Remove defining an alpha particle as a boson since we should do that for all nuclides/ions instead of doing it for just one


plasmapy/atomic/particle_class.py
```diff
@@ -632,11 +587,29 @@ def mass_number(self) -> int:
...
         """
-        Return the number of neutrons of the isotope corresponding
-        to this particle, or raises an `~plasmapy.utils.InvalidIsotopeError` if
-        the particle does not correspond to an isotope.
+        Return the number of neutrons in the isotope corresponding
+        to this particle, or raise an
+        `~plasmapy.utils.InvalidIsotopeError` if the particle does not
+        correspond to an isotope or neutron.
+        """
+        if self.particle == 'n':
```
StanczakDominik: Oh right, I forgot about that one. Bloody special cases.

plasmapy/atomic/particle_class.py
```diff
@@ -878,6 +811,7 @@ def become_set(arg: Union[str, Set, Tuple, List]) -> Set[str]:
 
     def is_electron(self) -> bool:
         """
-        Returns True if the particle is an electron.
+        Return `True` if the particle is an electron, and `False`
+        otherwise.
```
StanczakDominik: :open_mouth:

namurphy:
I'll probably merge this relatively soon, in case anyone would like to do a final code review before then.  Sooner or later I'd like to update more atomic tests too, but that might be best for after #278.  Thanks!

PR #290, opened 2018-03-05
Accept Roman numerals when specifying charge states

### LINKED ISSUE #286: Allow Roman numeral spectroscopic notation to represent ions
In astronomy it is common to use Roman numerals to represent the ion species responsible for the emission from a particular spectral line.  The Roman numeral is the charge state of the ion plus one (e.g., Fe I corresponds to Fe<sup>0+</sup> and Fe X corresponds to Fe<sup>9+</sup>).  

Some examples:
```Python
>>> from plasmapy.atomic import *
>>> integer_charge('Fe X')
9
>>> Particle('O VI').ion
'O 5+'
```
It would be worth looking for already existing packages that parse Roman numerals so that we do not need to reinvent this functionality.

I personally dislike this notation because the off-by-one difference is confusing and using Roman numerals greatly decreases readability for high charge states, but it is used commonly enough that we should allow it.

ritiek:
Closes #286.

This PR uses the [roman](https://pypi.python.org/pypi/roman) package to parse Roman numerals.

Some examples (same as in https://github.com/PlasmaPy/PlasmaPy/issues/286#issue-301633568):
```python
>>> from plasmapy.atomic import *
>>> integer_charge('Fe X')
9
>>> Particle('O VI').ion
'O 5+'
```

It may raise a `ValueError` with strings like `O -VI` or `O +VI` as `-VI` and `+VI` won't be recognized as roman numerals. So, I've added a try-except block to supress `ValueError` and raise an `InvalidParticleError` instead.

### DIFF
--- a/plasmapy/atomic/parsing.py
+++ b/plasmapy/atomic/parsing.py
@@ -1,5 +1,6 @@
 """Functionality to parse representations of particles into standard form."""
 
+import roman
 import numpy as np
 import re
 import warnings
@@ -205,13 +206,19 @@ def _parse_and_check_atomic_input(argument: Union[str, int], mass_numb: int = No
             if not sign_indicator_only_on_one_end and just_one_sign_indicator:
                 raise InvalidParticleError(invalid_charge_errmsg) from None
 
-            if '-' in charge_info:
-                sign = -1
-            elif '+' in charge_info:
-                sign = 1
-
             charge_str = charge_info.strip('+-')
-            Z_from_arg = sign * int(charge_str)
+
+            try:
+                if roman.romanNumeralPattern.match(charge_info):
+                    Z_from_arg = roman.fromRoman(charge_info) - 1
+                elif '-' in charge_info:
+                    Z_from_arg = - int(charge_str)
+                elif '+' in charge_info:
+                    Z_from_arg = int(charge_str)
+                else:
+                    raise InvalidParticleError(invalid_charge_errmsg) from None
+            except ValueError:
+                raise InvalidParticleError(invalid_charge_errmsg) from None
 
         elif arg.endswith(('-', '+')):  # Cases like 'H-' and 'Pb-209+++'
             char = arg[-1]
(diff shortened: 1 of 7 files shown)


namurphy:
My best guess for why the CircleCI test didn't pass is that the dependency information wasn't updated in the documentation test requirements file.  If that's the problem then you'll need to append the newest dependency to the list in [`requirements/automated-documentation-tests.txt`](https://github.com/PlasmaPy/PlasmaPy/blob/master/requirements/automated-documentation-tests.txt)

ritiek:
I already did [this](https://github.com/PlasmaPy/PlasmaPy/pull/290/files#diff-4dd1610cc69335320b5559602c99dc82R4). Maybe CircleCI is using Python 2 because I get the same error [as the one in CircleCI](https://circleci.com/gh/PlasmaPy/PlasmaPy/931?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link):
```
$ python setup.py install

  File "setup.py", line 159
    **package_info,
                  ^
SyntaxError: invalid syntax
```
when running setup.py with Python 2.

namurphy:
I think you're right about the problem.  I was surprised that this error happened on this pull request, but didn't on other recent pull requests.  I'm re-running the CircleCI tests for #282 to see what happens or if something changed.

namurphy:
Oh...could it be a problem with dependencies for the roman subpackage?  It's a few years old, so it might not be supported for Python 3.6, since it's PyPI page says it is supported for Python 3.3 at the latest.  If that's the case, maybe it's best to copy the necessary functions from roman and put them in the utils directory.  We would just need to put the copyright notice and the license in the `licenses/` directory and/or the file itself.

ritiek:
Can't say if that is the case because I took a look and the only dependency roman package uses is `re` (regex). And it works great on my Python 3.6. Not sure what's up with CircleCI.

Anyway, I think copying necessary functions and putting a copyright license should be a good workaround.

StanczakDominik:
If it's compatible with 3.3, it's compatible with 3.6. I don't see any way that could be an issue.

Before we commit to copying code an including licenses and so on, let's do some more digging into what is going on in here.

StanczakDominik:
Okay, I think I have a lead. 

Here's a recent build for another PR that worked:


image


Here's this one:

image

Note that the interface is quite different. The working builds are running on CircleCI 2.0 - for some reason doc builds for this PR cannot find our configuration file, so they're moving to inference and doing commands we don't necessarily want.

PR #299, opened 2018-03-08
Add type hints and update tests/docstrings in mathematics

namurphy:


plasmapy/mathematics/mathematics.py
```diff
@@ -1,17 +1,16 @@
...
 from mpmath import polylog
 from scipy.special import wofz as Faddeeva_function
+from typing import Union
 
-# TODO: Add type hint annotations to this module.
 
-
-def plasma_dispersion_func(zeta):
+def plasma_dispersion_func(zeta: Union[complex, int, float, np.ndarray, u.Quantity]
+                           ) -> Union[complex, float, u.Quantity]:
```
StanczakDominik: I think we're missing `np.ndarray` on the return end here. Seems to me like if you pass an array in, an array should come out.

plasmapy/mathematics/mathematics.py
```diff
@@ -92,7 +91,8 @@ def plasma_dispersion_func(zeta):
     return Z
 
 
-def plasma_dispersion_func_deriv(zeta):
+def plasma_dispersion_func_deriv(zeta: Union[complex, int, float, np.ndarray, u.Quantity]
+                                 ) -> Union[complex, float, u.Quantity]:
```
StanczakDominik: Likewise here.

plasmapy/mathematics/mathematics.py
```diff
@@ -194,32 +194,30 @@ def Fermi_integral(x, j):
...
-    See also
+    See Also
     --------
     plasma_dispersion_func
 
     Notes
     -----
-    The complete Fermi-Dirac integral is defined as: [1]_
+    The `complete Fermi-Dirac integral
+    <https://en.wikipedia.org/wiki/Complete_Fermi-Dirac_integral>`_ is
```
StanczakDominik: Why the change in links here?

namurphy: I was hoping to have this appear as an in-line link, which makes it easier to get to compared to the references.

plasmapy/mathematics/mathematics.py
namurphy: Good catch!  It's turning out to be pretty easy to miss type hints...

Issue #234, opened 2018-01-12
Code review for v0.1


lemmatum:
Prior to completing the [v0.1 milestone](https://github.com/PlasmaPy/PlasmaPy/milestone/1) we should do a code review, as per [Best Practices for Scientific Computing](http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1001745). We already do this to some extent via code reviews during pull requests, but it would be nice to have a look at the codebase as a whole.

This would entail sifting through all the code (including getting fresh eyes on code since many of us develop our own niches) looking for:
* bugs
* stylistic inconsistency
* bad practices
* lack of documentation or places where it could be improved
* non-exhaustive tests
* performance issues and potential optimizations

One way we could go about doing this is to open a code review issue for each module and assign 1 or 2 people (preferably people who haven't worked with and who are unfamiliar with the module) to that issue. Those persons will then sift through the code and open up new, minor issues. We can then have either reviewers or people more familiar with the module correct those issues, whichever is more appropriate.

Thoughts?

siddharth185:
Looks like a good learning experience. I can do a module.

lemmatum:
@namurphy When are we planning the freeze and would this be best done before or during the freeze? (e.g., freeze code, review, then unfreeze for fixes raised in review)

StanczakDominik:
@lemmatum I think the term we used was "feature freeze", and that was supposed to be just about now. In other words, we'll probably try to squeeze the current PRs in before mid february. Perhaps the start of February could be a good time to do the code review?

lemmatum:
@StanczakDominik sounds good to me. I'll try and get my PRs done in time for that.

I'd like to explicitly add to the list of things to check: consistency in vTh definitions. Sometimes plasma functions are defined with vTh that has a factor of 2 in the square root and sometimes without it. We should explicitly check that the equations we use are compatible with the sqrt(2) version as this is what we use for "most probable" thermal velocity.

namurphy:
We probably will have time for about one or two ~60-90 minute code review session per week, so we should prioritize.  What modules are the most important ones for us to review?

In the atomic subpackage, we just reviewed the `Particle` class in `particle_class.py` this past week.  I'd also like for us to review the `@particle_input` decorator, and possibly `parsing.py`.  We are planning on going over `physics/parameters.py` next Tuesday during our usual biweekly meeting time.  After that, I would say that the transport and Braginskii coefficient functionality is the highest priority.

[...]

PR #300, opened 2018-03-11
Correct annotation of atomic_number and isotopic_abundance

hzxusx:
Hi, I'm looking through the codebase and find two small bugs.

The type annotations are inconsistent with their doc string, and I think the doc is more reasonable. 

I see the review issue #234 is still open, so you will probably notice anyhow. 😃


StanczakDominik:
Yeah, those were obvious mistakes. Thanks a ton for pointing them out, @hzxusx ! :smiley:

Issue #315, opened 2018-03-28
Expand coverage for `PlasmaBlob` tests


lemmatum:
Most of the work here is just adding multiple plasma configurations to hit the various possible regimes for coupling and degeneracy.

ritiek:
I'll give this a shot. :)

not7cd:
#320 resolved this issue

StanczakDominik:
That it did. Thanks, @not7cd!

PR #320, opened 2018-03-31
Add tests for PlasmaBlob.regimes()

ritiek:
Addresses #315. I am not sure if it is possible to have a set of input values that would result in combination of such as strongly coupled regime with thermal kinetic energy dominant, etc.

### DIFF
--- a/plasmapy/classes/tests/test_Plasma.py
+++ b/plasmapy/classes/tests/test_Plasma.py
@@ -98,12 +98,12 @@ def test_Plasma3D_derived_vars():
 
 
 class Test_PlasmaBlobRegimes():
-    def test_intermediate_and_both_energies(self):
+    def test_intermediate_coupling(self):
         r"""
-        Method to test Gamma and Theta values on PlasmaBlob.regimes()
+        Method to test for coupling parameter for a plasma.
 
-        Tests that input values return an intermediate coupling regime
-        where both Fermi and thermal energy are important.
+        Tests against expected value for coupling parameter for a
+        plasma in the intermediate coupling regime.
 
         The input values in this case have no special significance
         and are just to get the desired output.
@@ -118,21 +118,19 @@ class Test_PlasmaBlobRegimes():
                                  Z=Z,
                                  particle=particle)
 
-        regimes = ['Intermediate coupling regime: Gamma = 10.585076050938532.',
-                   'Both Fermi and thermal energy important: Theta = 0.04113394941084522']
+        expect_regime = 'Intermediate coupling regime: Gamma = 10.585076050938532.'
+        regime, _ = blob.regimes()
+        testTrue = regime == expect_regime
 
-        testTrue = blob.regimes() == regimes
-
-        errStr = (f"Regimes \n{regimes}\n should be equal to that of class\n"
-                  f"{blob.regimes()}.")
+        errStr = f"Regime should be {expect_regime}, but got {regime} instead."
         assert testTrue, errStr
 
-    def test_strong_and_fermi(self):
+    def test_strongly_coupled(self):
         r"""
-        Method to test Gamma and Theta values on PlasmaBlob.regimes()
+        Method to test for coupling parameter for a plasma.
 
-        Tests that input values return a strongly coupled regime
-        where Fermi energy is dominant.
+        Tests against expected value for coupling parameter for a
+        plasma in the strongly coupled regime.
 
         The input values in this case have no special significance
         and are just to get the desired output.
@@ -147,21 +145,20 @@ class Test_PlasmaBlobRegimes():
                                  Z=Z,
                                  particle=particle)
 
-        regimes = ['Strongly coupled regime: Gamma = 104.02780112828943.',
-                   'Fermi quantum energy dominant: Theta = 0.008226789882169044']
+        expect_regime = 'Strongly coupled regime: Gamma = 104.02780112828943.'
 
-        testTrue = blob.regimes() == regimes
+        regime, _ = blob.regimes()
+        testTrue = regime == expect_regime
 
-        errStr = (f"Regimes \n{regimes}\n should be equal to that of class\n"
-                  f"{blob.regimes()}.")
+        errStr = f"Regime should be {expect_regime}, but got {regime} instead."
         assert testTrue, errStr
 
-    def test_weak_and_thermal(self):
+    def test_weakly_coupled(self):
         r"""
-        Method to test Gamma and Theta values on PlasmaBlob.regimes()
+        Method to test for coupling parameter for a plasma.
 
-        Tests that input values return a weakly coupled regime
-        where thermal kinetic energy is dominant.
+        Tests against expected value for coupling parameter for a
+        plasma in the weakly coupled regime.
 
         The input values in this case have no special significance
         and are just to get the desired output.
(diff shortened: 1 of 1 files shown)


lemmatum:
@ritiek here is a decent diagram showing various plasma regimes based on coupling and quantum effects. I'm pretty sure in the "spherical compression" regions (which refers to inertial confinement fusion) you can have strong coupling and thermally dominated plasmas. The "dense plasma" region is sometimes referred to as "hot dense plasma" for this reason. Note the dashed mu = 0 line which represents quantum effects based on when chemical potential becomes important. In other such diagrams I've also seen the degeneracy parameter, Theta (which compares Fermi energy to thermal energy),mapped across density and temperature space. This parameter has a slightly different curve to mu, though they broadly describe similar regions.

wdm_statediagram

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
...
+        particle = 'p'
+        blob = plasma.PlasmaBlob(T_e=T_e,
+                                 n_e=n_e,
+                                 Z=Z,
+                                 particle=particle)
+
+        regimes = ['Intermediate coupling regime: Gamma = 10.585076050938532.',
+                   'Both Fermi and thermal energy important: Theta = 0.04113394941084522']
+
+        assert blob.regimes() == regimes
```
lemmatum: Can you add an errStr here so that a clear message shows up for when this test fails? It should be in f-string format and should include the expected output along with output from `blob.regimes()`

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
...
+        particle = 'p'
+        blob = plasma.PlasmaBlob(T_e=T_e,
+                                 n_e=n_e,
+                                 Z=Z,
+                                 particle=particle)
+
+        regimes = ['Strongly coupled regime: Gamma = 104.02780112828943.',
+                   'Fermi quantum energy dominant: Theta = 0.008226789882169044']
+
+        assert blob.regimes() == regimes
```
lemmatum: same as above

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
...
+        particle = 'p'
+        blob = plasma.PlasmaBlob(T_e=T_e,
+                                 n_e=n_e,
+                                 Z=Z,
+                                 particle=particle)
+
+        regimes = ['Weakly coupled regime: Gamma = 0.0075178096952688445.',
+                   'Thermal kinetic energy dominant: Theta = 389929.7932144623']
+
+        assert blob.regimes() == regimes
```
lemmatum: same as above

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
...
+        particle = 'p'
+        blob = plasma.PlasmaBlob(T_e=T_e,
+                                 n_e=n_e,
+                                 Z=Z,
+                                 particle=particle)
+
+        regimes = ['Intermediate coupling regime: Gamma = 0.5234187230217864.',
+                   'Thermal kinetic energy dominant: Theta = 120.65958493847927']
+
+        assert blob.regimes() == regimes
```
lemmatum: same as above

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
...
+        particle = 'p'
+        blob = plasma.PlasmaBlob(T_e=T_e,
+                                 n_e=n_e,
+                                 Z=Z,
+                                 particle=particle)
+
+        regimes = ['Intermediate coupling regime: Gamma = 86.68983427357452.',
+                   'Fermi quantum energy dominant: Theta = 0.009872147858602853']
+
+        assert blob.regimes() == regimes
```
lemmatum: same as above

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -97,6 +97,84 @@ def test_Plasma3D_derived_vars():
     assert np.allclose(test_plasma.alfven_speed.value, 10.92548431)
 
 
+class Test_PlasmaBlobRegimes():
+    def test_intermediate_and_both_energies(self):
```
lemmatum: Add docstrings for all of these tests to be clear on what is being tested. The names themselves are not so clear... what does "intermediate" and "both energies" refer to? This is important because eventually `PlasmaBlob` will have all sorts of parameters and not just coupling parameter and Fermi energy.

ritiek:
@lemmatum Thanks for the review! I've update my PR. I am still not sure how should we name our methods.
> I'm pretty sure in the "spherical compression" regions (which refers to inertial confinement fusion) you can have strong coupling and thermally dominated plasmas.

That's nice. I'll take a look if I can reproduce this case for our tests.

lemmatum:
@ritiek no problem, and thanks for working on this!
Perhaps it would be better to just test each regime individually. In other words, separate testing of coupling parameter and Fermi energy. Then you can name the methods like `test_strongly_coupled()`, `test_weakly_coupled`, `test_intermediate_coupling`.

I think this would require grabbing just one of the elements from the `blob.regimes()` list via indexing.

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -99,6 +99,16 @@ def test_Plasma3D_derived_vars():
 
 class Test_PlasmaBlobRegimes():
     def test_intermediate_and_both_energies(self):
+        r"""
+        Method to test Gamma and Theta values on PlasmaBlob.regimes()
+
+        Tests that input values return an intermediate coupling regime
```
lemmatum: I think this sentence is mixing two different regime parameters. Ask I discussed on the comment about the state space diagram, coupling and degeneracy are two different concepts (even though they both tend to appear at higher densities).

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -111,10 +121,23 @@ def test_intermediate_and_both_energies(self):
         regimes = ['Intermediate coupling regime: Gamma = 10.585076050938532.',
                    'Both Fermi and thermal energy important: Theta = 0.04113394941084522']
 
-        assert blob.regimes() == regimes
+        testTrue = blob.regimes() == regimes
 
+        errStr = (f"Regimes \n{regimes}\n should be equal to that of class\n"
```
lemmatum: I would change this to "Regime should be {regimes}, but got {blog.regimes()} instead.". That way it is clearer which is the expected answer and which is the answer from the method.

plasmapy/classes/tests/test_Plasma.py
```diff
@@ -111,10 +121,23 @@ def test_intermediate_and_both_energies(self):
...
+                  f"{blob.regimes()}.")
+        assert testTrue, errStr
 
     def test_strong_and_fermi(self):
+        r"""
+        Method to test Gamma and Theta values on PlasmaBlob.regimes()
+
+        Tests that input values return a strongly coupled regime
```
lemmatum: Same problem here. This should be something like "Tests against expected parameters for a plasma which is simultaneously in the strongly coupled and (Fermi) degenerate regimes."

lemmatum:
I don't think you need to go that far. The regimes are [calculated separately](https://github.com/PlasmaPy/PlasmaPy/blob/127a6f305dbce43fb514628847c6fd12fedf5cc0/plasmapy/classes/plasma.py#L183) (note that it is two separate if blocks) so checking each parameter individually makes the most sense.

I would start with rewriting it very simply (one test for one parameter regime per method) and then we can discuss whether we want a list of tests or some other more extensive framework for each method/regime.

ritiek:
I see. I've modified to test only for a single regime at a time. Do let me know if there is else that needs to be changed!

StanczakDominik:
Okay, for some reason seeing this many open, reviewed, minor, non-feature PRs really bugs me. These changes are neat and I'm going to merge them - good job @ritiek, if you come up with anything else you'd like to add here, feel free to open up a new one!

PR #323, opened 2018-04-02
Pre-create commonly used Particle instances

namurphy:
I added `proton`, `electron`, `neutron`, `positron`, `deuteron`, `triton`, and `alpha` as pre-created instances of the `Particle` class in `atomic/__init__.py`.  This change allows us to import these `Particle` instances directly from the atomic subpackage, e.g.:
```Python
from plasmapy.atomic import proton, electron
```
These are the most frequently used particles in plasma physics, so I suspect that it will be a useful shortcut for many of us.

### DIFF
--- a/docs/atomic/particle_class.rst
+++ b/docs/atomic/particle_class.rst
@@ -36,6 +36,15 @@ keyword and the integer charge may be represented with the `Z` keyword.
 
 >>> proton = Particle(1, mass_numb=1, Z=1)
 
+The most frequently used `~plasmapy.atomic.Particle` instances may be
+imported directly from the atomic subpackage.  
+
+>>> from plasmapy.atomic import proton, electron
+
+The `~plasmapy.atomic.Particle` instances that may be imported
+directly are: `proton`, `electron`, `neutron`, `positron`, `deuteron`,
+`triton`, and `alpha`.
+
 Accessing Particle Properties
 =============================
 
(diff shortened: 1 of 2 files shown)


plasmapy/atomic/__init__.py
```diff
@@ -38,3 +38,11 @@
...
     nuclear_reaction_energy,
 )
+
+proton = Particle("p+")
+electron = Particle("e-")
+neutron = Particle("n")
+positron = Particle("e+")
+deuteron = Particle("D 1+")
+triton = Particle("T 1+")
+alpha = Particle("He-4 2+")
```
lemmatum: I'm wondering if there is a more efficient way of doing this. Maybe using a dictionary?
Might not be important if we are doing this just for a few particles.

docs/atomic/particle_class.rst
```diff
@@ -36,6 +36,15 @@ keyword and the integer charge may be represented with the `Z` keyword.
...
 
+The most frequently used `~plasmapy.atomic.Particle` instances may be
+imported directly from the atomic subpackage.  
+
+>>> from plasmapy.atomic import proton, electron
+
+The `~plasmapy.atomic.Particle` instances that may be imported
+directly are: `proton`, `electron`, `neutron`, `positron`, `deuteron`,
+`triton`, and `alpha`.
+
```
lemmatum: Does what it says on the tin :smile:

plasmapy/atomic/__init__.py
StanczakDominik: I like it how it is, given this is a shorthand for the most commonly used Pericles.

Oh, the wonders of autocorrect.

Issue #312, opened 2018-03-25
Merge functionalities of collision_frequency, collision_rate_{ion,electron}_ion


StanczakDominik:
I came across this working on #307 and #303. As discussed during the last code review session, we have three functions that sort of do the same thing, but not quite. In transport.py, there's @lemmatum's `collision_frequency` function. In [Callen's notes](http://homepages.cae.wisc.edu/~callen/chap2.pdf), it's equation 2.14.

https://github.com/PlasmaPy/PlasmaPy/blob/6d7942dc3ca473d7e9bc999d1e8a6e71ec10b8ca/plasmapy/physics/transport.py#L526-L684
And then in parameters.py there's @samurai688's version - two functions named `collision_rate_ion_ion` and `collision_rate_electron_ion`. However, Callen calls them the "fundamental ion (and electron, respectively) collison frequencies":
https://github.com/PlasmaPy/PlasmaPy/blob/6d7942dc3ca473d7e9bc999d1e8a6e71ec10b8ca/plasmapy/physics/parameters.py#L562-L688
From what I was able to gather, they're related by (in pseudocode):
```python
from conveniently_named_variables import *
collision_rate_electron_ion = 4 / (3 * np.sqrt(np.pi)) collision_frequency(thermal_speed(..., 'e' ...)
collision_rate_ion_ion = (m_e/m_i)**2 *(T_e/T_i)**1.5 * Z_i**2 * v_e / np.sqrt(2)
```
where that last square root of 2 is somehow related to reduced masses and interparticle motion in the ion-ion analysis. So... it looks like it's not going to be too trivial to simply replace these! I'm not really sure how to proceed here. I'm thinking about replacing some of the "fundamental" calculations in `rate` functions (they start from the Coloumb logarithm) in terms of our `collision_frequency` as seen in my pseudocode above.

StanczakDominik:
Also, I'm happy I got this github syntax to work!

samurai688:
Nice, looks right to me!  I think you've cracked the code.

It sounds like what we need is to understand and document all these different flavors of collision frequencies, which you've begun here. Then we could make the collision_frequency function able to produce all the different flavors on request, with some sort of sensible defaults. Then in each assorted function that needs collision frequency, be it hall parameter or transport or whatnot, they can call the function in the proper way for their need.

What do you think?

StanczakDominik:
[...]

Issue #298, opened 2018-03-07
Add narrative documentation for transport.py


namurphy:
There is an incredibly extensive amount of functionality in `transport.py`.  We will need to have narrative documentation in our `docs` directory that goes over how to use this module.  Given the difficulty of the subject, the transport documentation will need to be written by people who have a good understanding of plasma transport theory.  The documentation should be useful and understandable to someone who is taking their first class in plasma physics.  We will need this for our v0.1.0 release.

Thanks!
-Nick

lemmatum:
I'll get on the `collisions.py` documentation once #303 is done.

lemmatum:
Look at [atomic narrative doc](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/atomic) for example. Place transport narrative docs [here](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/physics/transport)

lemmatum:
- [ ] Check if `collision_rate_electron_ion` docstring in `collisions.py` compiles correctly. Especially `* nu_e\i_S from ref [2]_ eqn (1) pp. #,`

PR #324, opened 2018-04-03
Rewriting collision_rate_ion_ion in parameters.py

lemmatum:
Rewriting collision_rate_ion_ion in parameters.py in terms of collision_frequency from collision.py.
This partially addresses #312


plasmapy/physics/parameters.py
```diff
@@ -713,32 +716,42 @@ def collision_rate_ion_ion(T_i, n_i, ion_particle,
...
+    from plasmapy.physics.transport.collisions import (Coulomb_logarithm,
+                                                       collision_frequency)
     T_i = T_i.to(u.K, equivalencies=u.temperature_energy())
     m_i = atomic.ion_mass(ion_particle)
-    if V is not None:
-        V = V
+    particles = [ion_particle, ion_particle]
+    if V:
+        pass
     else:
```
StanczakDominik: You could just do `if not V: V = np.sqrt(...)`, no?

StanczakDominik:
I'll finish this up and merge before telecon today :)

plasmapy/physics/parameters.py
StanczakDominik: Actually no, because we have to handle zero V via  `if V is not None`.

StanczakDominik:
Anddddd I'm a bit stuck over at lemmatum/Plasmapy#4.

plasmapy/physics/parameters.py
lemmatum: Shouldn't we be raising an error when `V=0`? This basically means the user messed up somewhere and is passing us a static system where collisions are not possible.

StanczakDominik: Okay, you're right. I completely missed that.

plasmapy/physics/parameters.py
```diff
@@ -713,32 +716,42 @@ def collision_rate_ion_ion(T_i, n_i, ion_particle,
     <Quantity 95918.76240877 1 / s>
 
     """
-    from plasmapy.physics.transport.collisions import Coulomb_logarithm
+    from plasmapy.physics.transport.collisions import (Coulomb_logarithm,
+                                                       collision_frequency)
     T_i = T_i.to(u.K, equivalencies=u.temperature_energy())
     m_i = atomic.ion_mass(ion_particle)
```
namurphy: There may be a conflict here with #340, but I think that can be resolved by switching `ion_mass` to `particle_mass` here.  It would probably make sense for this PR to be merged first, and then I can fix the conflicts in #340 after that.

plasmapy/physics/parameters.py
namurphy: On second thought, I'm not sure if we should raise an exception if `V=0`, since there may be instances where people are plotting collision_rate stuff as a function of `V`, and would use `V=0` as an endpoint in the scan.  I imagine this use would be pretty rare though.

lemmatum:
TODO:
- [ ] Fix `collision_frequency` so that it can accept `'e-'` as a particle and correctly go to electron-electron or electron-ion collisions case instead of ion-ion collisions.
- [ ] Add a check to `collision_frequency` to see if a velocity has been passed, and if it hasn't, to calculate the appropriate thermal velocity. This will fix the test failure when passing a velocity for electron-electron and electron-ion collisions.
- [ ] Check if we need to pass this velocity to the `Coulomb_logarithm` in each case, since currently `np.nan` is being passed.

plasmapy/physics/parameters.py
lemmatum: @namurphy What would you return though? I don't think you can return a sensible collision frequency for a static system.

StanczakDominik: Zero collisions per unit time (any time, in fact) seems reasonable enough!

StanczakDominik: In fact, barring errors and just looking at the math of it, that seems to be exactly what we're returning right now, given that frequency is ~V^3.

StanczakDominik: Please disregard what I said. Mistakes were made.

lemmatum:
So I believe these tests should pass and we should be ready for another review.

plasmapy/physics/transport/collisions.py
```diff
@@ -633,27 +633,34 @@ def collision_frequency(T,
...
-                                                       V=V)
-    if particles[0] == 'e' and particles[1] == 'e':
+    T, masses, charges, reduced_mass, V_red = _boilerPlate(T=T,
+                                                           particles=particles,
+                                                           V=V)
+    if particles[0] in ('e','e-') and particles[1] in ('e','e-'):
+        # if a velocity was passed, we use that instead of the reduced
+        # thermal velocity
+        if np.isnan(V):
+            V = V_red
```
namurphy: I got slightly confused by the abbreviated variable name `V_red`.  I think it would more readable if it's changed to `V_reduced`.

plasmapy/physics/transport/collisions.py
```diff
@@ -689,6 +700,235 @@ def collision_frequency(T,
...
+    .. [1] Braginskii
+
+    .. [2] Formulary
+
+    .. [3] Callen Chapter 2, http://homepages.cae.wisc.edu/~callen/chap2.pdf
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> collision_rate_electron_ion(0.1*u.eV, 1e6/u.m**3, 'p')
```
namurphy: Perhaps add spaces around each of the operators in the examples sections?

plasmapy/physics/transport/collisions.py
```diff
@@ -689,6 +700,235 @@ def collision_frequency(T,
...
+        The relative velocity between particles.  If not provided,
+        thermal velocity is assumed: :math:`\mu V^2 \sim 2 k_B T`
+        where `mu` is the reduced mass.
+
+    coulomb_log : float or dimensionless ~astropy.units.Quantity, optional
+        Option to specify a Coulomb logarithm of the electrons on the ions.
+        If not specified, the Coulomb log will is calculated using the
+        ~plasmapy.physics.transport.Coulomb_logarithm function.
+
+    coulomb_log_method : string, optional
```
namurphy: Can change `string` to `str`.

plasmapy/physics/transport/collisions.py
```diff
@@ -689,6 +700,235 @@ def collision_frequency(T,
...
+    From [3]_, equations (2.17) and (2.120)
+
+    Considering a Maxwellian distribution of "test" electrons colliding with
+    a Maxwellian distribution of "field" ions.
+
+    This result is an electron momentum relaxation rate, and is used in many
+    classical transport expressions. It is equivalent to:
+    * 1/tau_e from ref [1]_ eqn (1) pp. #,
+    * 1/tau_e from ref [2]_ eqn (1) pp. #,
+    * nu_e\i_S from ref [2]_ eqn (1) pp. #,
```
namurphy: The formatting for `nu_e\i_S` looks a little strange to me, so it would be worth checking the doc build to make sure it comes out as intended.

plasmapy/physics/transport/collisions.py
```diff
@@ -689,6 +700,235 @@ def collision_frequency(T,
...
+
+    coulomb_log_method : string, optional
+        Method used for Coulomb logarithm calculation (see that function
+        for more documentation). Choose from "classical" or "GMS-1" to "GMS-6".
+
+    References
+    ----------
+    .. [1] Braginskii
+
+    .. [2] Formulary
```
namurphy: If you have the time, it would be good to add in the full references here including a link to the Formulary.  Alternatively, if this comes up in a bunch of places, we could create an issue for it.

plasmapy/physics/transport/collisions.py
```diff
@@ -689,6 +700,235 @@ def collision_frequency(T,
...
+
+    .. [3] Callen Chapter 2, http://homepages.cae.wisc.edu/~callen/chap2.pdf
```
namurphy: I do like this reference, and I wonder if he'll end up finishing the book!

lemmatum: No idea, haven't looked into this book in depth. What have you used it for?

plasmapy/physics/transport/collisions.py
lemmatum: It'll make line 636 too long... but I have a cheeky compromise...

plasmapy/physics/transport/collisions.py
lemmatum: I've made a note of this under #298

PR #340, opened 2018-04-05
Remove ion_mass and isotope_mass and replace with particle_mass

namurphy:
The functionality of `ion_mass` and `isotope_mass` is a subset of the functionality of `particle_mass`, and it doesn't make much sense to have three separate functions to find a particle's mass.  This PR deletes `ion_mass` and `isotope_mass` and replaces all instances of their use with `particle_mass` instead, which is the most general of these functions.  There still remains `standard_atomic_weight`, which I think is sufficiently different from `particle_mass` to justify keeping.

> There should be one-- and preferably only one --obvious way to do it.

I also had to change `"H"` to `"H+"` in some tests because `ion_mass` had assumed that the particle was singly charged if no charge is given.  Both `ion_mass` and `isotope_mass` did checks to make sure that the particle was of the correct classification, and would raise an `InvalidIonError` or `InvalidIsotopeError` if necessary.  I recommend using `@particle_input` to do these checks instead.

In any case, the deleted functions will live on in our memories.  We'll miss you, `ion_mass` and `isotope_mass`! :crying_cat_face:


plasmapy/atomic/atomic.py
```diff
@@ -180,186 +176,13 @@ def standard_atomic_weight(element: Particle) -> u.Quantity:
...
-    >>> ion_mass('T+')
-    <Quantity 5.00735666e-27 kg>
-    >>> ion_mass(26, Z=1, mass_numb=56)
-    <Quantity 9.28812345e-26 kg>
-    >>> ion_mass('Fe-56 1+')
-    <Quantity 9.28812345e-26 kg>
-
-    """
-
-    # TODO: Remove deprecated functionality elsewhere in the code
```
StanczakDominik: So this is the PR that removes all the `DeprecationWarning`s? Awesome!

namurphy: Some of them, at least!  There are still a bunch that are popping up for other tests.

PR #346, opened 2018-04-06
Improve treatment of neutral atoms in Particle class and add __invert__ attribute

### LINKED ISSUE #343: Separate how Particle class treats ions and neutral atoms
Right now the `Particle` class has an attribute `Particle.ion` which returns the ionic symbol if the particle is either an ion or a neutral charged particle (see also #338 which changes `ion_symbol` to `ionic_symbol`).  This can lead to confusion since if we treat `Particle.ion` as a `bool` to check whether or not something is an ion, then it will return `True` for neutral atoms where the charge information is given.  

One possibility would be to create an `is_ion` attribute for `Particle` that returns a `bool` of whether or not the particle is a charged ion.  In that case, we could rename `Particle.ion` as `Particle.ionic_symbol` (and `Particle.isotope` as `Particle.isotope_symbol`, etc.) and also create a `Particle.is_isotope` attribute.  We should keep in mind that plasma physicists sometimes worry about electron-positron pair plasmas as we are doing this.

The `categories` and `is_category` attributes of `Particle` need to updated to reflect that `'H 0+'`, for example, is not classified as an `'ion'`.  This would also affect how the `@particle_input` decorator treats things (e.g., the `any_of`, `require`, and `exclude` keyword arguments).

I'm also wondering if the `atomic` subpackage should keep separating elements (e.g., `'H'`, `'He'`, etc.) from neutrally charged atoms (e.g., `'H 0+'`, `'He 0+'`, etc.).  I'm still inclined to keep them separate, since explicitly putting in the charge is better than implicitly assuming it, but this is worth thinking more about too.

namurphy:
This PR addresses improves the separation between neutral atoms with charge information and charged ions.  A new and exciting feature is the `is_ion` attribute for the `Particle` class, and I think I'm renaming the `Particle.ion` attribute to `Particle.ionic_symbol`.  I'm putting more detail in the commit log.  This mostly closes #343.


namurphy:
So I think I'm addicted to magic methods.  On a whim I made it so that `Particle` instances that are elementary particles or antiparticles can be inverted with the unary operator in order to get their antiparticles.
```Python
>>> electron = Particle('e-')
>>> positron = ~electron
```

docs/atomic/particle_class.rst
```diff
@@ -30,7 +30,7 @@ but element names and many aliases are not.
 >>> hydride = Particle('H-')
 
 An `int` may be used as the first positional argument to
-`plasmapy.atomic.Particle` to represent an atomic number.  For isotopes
+`~plasmapy.atomic.Particle` to represent an atomic number.  For isotopes
```
StanczakDominik: Just a question. What does the `~` do, again? Why do some of these links have the `~` character and some don't need to? Does it signify an external (non-standard lib) import?

docs/atomic/particle_class.rst
```diff
@@ -162,5 +165,17 @@ a particle is in one of these categories.
...
-            print(f"{particle} corresponds to ion {particle.ion}")
+
+.. _particle-class-antiparticles
+
+Returning Antiparticles
+=======================
+
+The antiparticle of an elementary particle or antiparticle may be found
+by using the unary operator `~` to invert a `~plasmapy.atomic.Particle`
+instance.
```
StanczakDominik: That's a pretty cool idea! :smile:

plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -470,7 +510,8 @@ def test_particle_class_mass_nuclide_mass(isotope: str, ion: str):
     Isotope = Particle(isotope)
     Ion = Particle(ion)
 
-    if Isotope.particle == Ion.particle and Isotope.particle in ('n', 'p+'):
+    if Isotope.categories & {'isotope', 'baryon'} and Ion.categories & {'ion', 'baryon'}:
```
StanczakDominik: Clever use of notation!

plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -524,3 +565,32 @@ def test_particle_is_electron(p, is_one):
...
+        assert particle.integer_charge == -opposite.integer_charge, \
+            (f"The charges of {particle} and {opposite} are not "
+             f"opposites, which indicates that there is a problem "
+             f"using the unary operator to find the antiparticle of "
+             f"{particle}.")
+
+        assert particle._attributes['mass'] == opposite._attributes['mass'], \
+            (f"The masses of {particle} and {opposite} are not equal, "
+             f"which indicates that there is a problem using the unary "
+             f"operator to find the antiparticle of {particle}.")
```
StanczakDominik: Well documented testing right here. However, could we try a non-sequential approach? I'll PR a proposal into your branch!

StanczakDominik: Aaaand namurphy/PlasmaPy#1 is live!

docs/atomic/particle_class.rst
namurphy: Without putting in the `~`, the doc build here wrote out the whole "plasmapy.atomic.Particle".  With the `~`, it just writes out `Particle`.  I think it lets ReST or Sphinx know that it's a link to a particular object.  I'm by no means a ReST expert though!

plasmapy/atomic/tests/test_particle_class.py
namurphy: My original intention with this test was only to check whether or not the `~` returned the antiparticle, so with that in mind, I used sequential tests on purpose to avoid getting redundant error messages.  I thought about this more while cooking just now because turnips are boring, and came to the conclusion that I was thinking too narrowly and that it is worthwhile to rigorously check all of the properties too, so your PR will be a good basis for that.

plasmapy/atomic/particle_class.py
```diff
@@ -480,6 +501,35 @@ def __bool__(self):
...
+        elif 'anti_' in old_particle:
+            return Particle(old_particle.replace('anti_', ''))
+        elif 'anti' in old_particle:
+            return Particle(old_particle.replace('anti', ''))
+        else:
+            try:
+                return Particle('anti' + old_particle)
+            except InvalidParticleError:
+                return Particle('anti_' + old_particle)
+
```
namurphy: I think I made this more complicated than I needed it to be.  I am going create a dict (or maybe a defaultdict) that has particles as keys and their antiparticles as the corresponding values.

StanczakDominik:
I think one thing to consider with a dictionary of this sort is that you're generating more objects that have to be kept in memory from import time onward... But it's before 6 AM, I'm still a tad sleepy, so I was thinking about a Particle: Particle dict which is a really bad idea, and you went with the completely reasonable string to string mapping. I completely agree with this decision. :+1:

namurphy:
Thanks for the code review!  Is there anything else we should do before merging this?

StanczakDominik:
Nope, green across the board so I'm going to go ahead and hit merge!

PR #335, opened 2018-04-05
Physics documentation improvements

StanczakDominik:
Closes #333 and #297.

Writing docs is *fun*, it turns out! Still, I may be having too much of it and taking too many liberties with style, so please point stuff out.

PR #351, opened 2018-04-09
WIP: Various documentation improvements

StanczakDominik:
Things I'd like to get done with this:

- [x] Overhaul the documentation layout and content a bit for mathematics
- [x] Utils
- [ ] Figure out the structure for `atomic` docs (per #335 )
- [ ] Improve the development guide a bit (move stuff around), make things more granular

If anyone has anything to add, I'm listening!


lemmatum:
Transport fixes are under issue #298 ?

StanczakDominik:
Oh right... They are. Well, in that case I'll deal with the other subpages first. Thanks for the reminder!

StanczakDominik:
Writing 840d71f has let me realize that I don't really actually like our current `__init__.py` setup with squashing all the functionally different stuff (the different `checks`, `exceptions`, `pytest` and `import_helpers` into a single namespace. I mean, look at [this list](https://389-90793234-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/utils/index.html#functions)! That's horrible. This stuff mixes everything up.

I'll have to look at some other friendly Python packages and pay closer attention to how they organize their submodules and docs.

StanczakDominik:
On the other hand, the currently pushed state of the docs for Utils is how I envision these looking. Sort of. Up to actually chaning the stuff I copy pasted.

namurphy:
Yeah, that's a good point that we have packed a bunch of stuff into `utils` so it does seem like it would make sense to have a few namespaces within `utils`.  I do like the class inheritance diagram.

StanczakDominik:
Okay then, I'll remove a bunch of stuff from that init, then. I'm really itching to do the same thing to physics, honestly, but let's take things slowly.

namurphy:
That's a good point, we do need to think more about namespaces for what we have in the physics subpackage, don't we?  I've been starting to think that transport should be its own subpackage rather than a subsubpackage within the physics subpackage.  In any case, starting with utils first is probably best! :zzz:

lemmatum:
Yea physics is way too broad (everything is physics!). One way to separate it would be according to subfields, e.g. transport, quantum, relativistic, magnetized, classical, etc. Another way would be to arrange by shared functionality. For example, transport is currently quite self-contained whereas parameters is a bit all over the place.

StanczakDominik:
As is probably pretty visible, I'm currently wondering about conventions for doc chapter names. I'm trying a few ways to approach this.

StanczakDominik:
Notes from reading astropy docs:
* We'll need to put the logo in here somewhere!
* Instead of organizing the docs by code package, they have a toctree for a each of different functionalities on the index page and they link to documents from all over the place. There's separation between the code structure and documentation structure.

StanczakDominik:
* [ ] fix the module index at http://docs.plasmapy.org/en/latest/py-modindex.html

docs/physics/transport/braginskii.rst
```diff
@@ -21,3 +21,4 @@ Reference/API
 .. automodapi:: plasmapy.physics.transport.braginskii
    :no-heading:
    :no-main-docstr:
+   :skip: Coulomb_logarithm grab_charge
```
namurphy: Should we make `grab_charge` a private function by renaming it as `_grab_charge` if it doesn't need to be part of the public API?

StanczakDominik: Test!

StanczakDominik: Darn autocorrect to heck! I meant to say yes. :D

StanczakDominik:
Yeah, Giles is a bit intermittent like that. We should probably investigate that and make a bug report.

For now, could you take a look here?

http://plasmapy-stanczakdominik-test.readthedocs.io

StanczakDominik:
Wrong link, sorry! http://plasmapy-stanczakdominik-test.readthedocs.io/en/mathematics-documentation/index.html

namurphy:
Thanks!  Now that I can see it, it really does look like a major improvement.  It all still looks good to me.

lemmatum:
@StanczakDominik I noticed that `dielectric.py` is titled "Dielectric Tensors", I think limiting it to tensors may be too specific. Although at the moment the functions involve tensors, I'm planning on adding some functions that don't necessarily involve tensors (e.g. the standard longitudinal susceptibility used for Thomson scattering).

StanczakDominik:
I kinda went in another direction here. I like how this looks on my end! Ready for final review.

PR #353, opened 2018-04-10
Make @particle_input work with tuples of particles

### LINKED ISSUE #336: Have @particle_input work with tuples of particles
Presently the @particle_input decorator works by converting an argument annotated with the `Particle` class into a corresponding `Particle` instance.  An example:
```Python
from plasmapy.atomic import Particle, particle_input

@particle_input
def some_function(particle: Particle):
    print(particle.integer_charge)

>>> some_function("p+")
1
```
However, there are functions that take in tuples or lists of particles.  For example, `Coulomb_logarithm` has a `particles` argument that is for a tuple or list of the test particle and the target particle.  This usage is not supported by `@particle_input`, and needs to be implemented.  An example use case would be:

```Python
def another_function(particles: (Particle, Particle)):
    print(particle[0].integer_charge)
    print(particle[1].integer_charge)

>>> another_function(particles=('e-', 'p+'))
-1, 1
```
It should work for arguments that are tuples or lists.  Once this capability is added to `@particle_input`, then it will need to be implemented in `Coulomb_logarithm` and other functions in `collisions.py`.

namurphy:
The goal of this pull request is to allow the `particle_input` decorator to work for arguments with annotations such as `(Particle, Particle)`.  More details and sample code is provided in #336.  I am starting out by trying test-driven development.  Closes #336.

### DIFF
--- a/plasmapy/atomic/tests/test_particle_input.py
+++ b/plasmapy/atomic/tests/test_particle_input.py
@@ -1,5 +1,5 @@
 import pytest
-from typing import Optional
+from typing import Optional, Union
 
 from ...utils import (
     AtomicError,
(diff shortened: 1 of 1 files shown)


namurphy:
@pep8speaks pep8ify

StanczakDominik:
@namurphy need a hand on this?

ritiek:
Oh, I didn't notice this was already open before. Actually, since this PR contains only tests and I haven't written any in #483. It might be a good idea if this were to merge with my [`particle-input`](https://github.com/ritiek/PlasmaPy/tree/particle-input) branch and then we could work continue working on #483. @namurphy What say?

namurphy:
Sounds like a good idea!  I had been stalled on this for so long that I had forgotten that I had written tests.

StanczakDominik:
And I had completely forgotten that I saw tests here.

This is actually a wonderful way to collaborate! Tests let you design an interface in a way you most often can't simply describe in words to another person, without actually writing the code you'd like to pass the tests. I had the chance to try this at work recently (a colleague was using a package I wrote and wanted another feature), it was awesome.

namurphy:
Yeah, definitely.  That's a great point.  I think when I raise issues, I'll start writing and including some basic tests to make it clearer what I was thinking (at least when it makes sense to).

Issue #331, opened 2018-04-04
Turn astropy quantities into strings with Python syntax for __repr__ fields


StanczakDominik:
As per #303 and the __repr__ of our PlasmaBlob method,

> I don't 100% like how the __repr__ came out, as ideally you'd like to be able to copy the output, paste it into the console and have the same object initialized, but it turns out printing out astropy units like 10*u.eV instead of 10 eV is going to be non trivial unless
a) I'm missing something from their docs
b) we contribute upstream and figure out a way to add this string format.

>  annoying thing here would be that the Quantities won't know what the astropy.units subpackage is imported as (u here).

I'm thus noting this down so as to remember to figure this out. It seems nontrivial for now, but if we figure it out we might be able to contribute something back to astropy.

StanczakDominik:
Closed by #358!

PR #358, opened 2018-04-14
Handle `Quantity` objects in `call_string()`

### LINKED ISSUE #337: Make call_string work with Quantity objects
The [`call_string`](https://github.com/PlasmaPy/PlasmaPy/blob/28156768b242564265e33b3baf8ae231f1a7c166/plasmapy/utils/pytest_helpers.py#L82) function in `utils` is designed to return a string of a function call, which is often really useful to put into error messages.  An example is:
```Python
>>> from plasmapy.utils import call_string
>>> def f(x, y=2): return x + y
>>> call_string(f, 1, {'y': 2})
'f(1, y=2)'
```
However, `call_string` does not work so well with Quantity objects:
```Python
>>> import astropy.units as u
>>> call_string(f, 5 * u.m)
'f(<Quantity 5. m>)'
```
Since we generally do `import astropy.units as u`, we probably should make `call_string` behave like this:
```Python
>>> call_string(f, 5 * u.m)
'f(5 * u.m)'
```
This is very related to #331.  Thanks!

ritiek:
Closes #337 and #331.

For example, now passing a `Quantity` object to `call_string` results in following output:
```python
>>> from plasmapy.utils import call_string
>>> import astropy.units as u
>>> def f(x, y, z):
...        pass
>>> call_string(f, 5 * u.m, {'y': 8*u.m, 'z': 5*u.kg})
'f(5.0 * u.m, y=8.0 * u.m, z=5.0 * u.kg)'
```

However, I personally find this a bit more readable :smile::
```python
'f(5.0*u.m, y=8.0*u.m, z=5.0*u.kg)'
```

What do you guys think?

### DIFF
--- a/plasmapy/classes/plasma.py
+++ b/plasmapy/classes/plasma.py
@@ -155,10 +155,10 @@ class PlasmaBlob():
         >>> PlasmaBlob(1e4*u.K, 1e20/u.m**3, particle='p')
         PlasmaBlob(T_e=10000.0 K, n_e=1e+20 1 / m3, particle=p, Z=1)
         """
-        argument_dict = { 'T_e'      : self.T_e,
-                          'n_e'      : self.n_e,
-                          'particle' : self.particle,
-                          'Z'        : self.Z }
+        argument_dict = {'T_e': self.T_e,
+                         'n_e': self.n_e,
+                         'particle': self.particle,
+                         'Z': self.Z}
 
         return call_string(PlasmaBlob, (), argument_dict)
 
--- a/plasmapy/classes/plasma.py
+++ b/plasmapy/classes/plasma.py
@@ -23,6 +23,8 @@ from plasmapy.atomic import (particle_mass,
                              integer_charge,
                              )
 
+from plasmapy.utils import call_string
+
 class Plasma3D():
     """
     Core class for describing and calculating plasma parameters with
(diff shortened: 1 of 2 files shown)


StanczakDominik:
In arguments, I think it's reasonable to go for no spaces between operators. I think pycharm was shouting at me for that, at least.

Do you think you could also implement this in PlasmaBlob.__repr__? That would close #331 as well!

namurphy:
Yeah, since `call_string` is mostly used in error messages, using the more compact notation that you suggested sounds good to me!

StanczakDominik:
That "?". :D

ritiek:
Okay, `call_string` now works with `PlasmaBlob.__repr__` too:

```python
>>> from plasmapy.classes.plasma import PlasmaBlob
>>> import astropy.units as u

>>> T_e = 5 * 11e3 * u.K
>>> n_e = 1e23 * u.cm ** -3
>>> Z = 2.5
>>> particle = 'p'
>>> blob = PlasmaBlob(T_e=T_e,
...                   n_e=n_e,
...                   Z=Z,
...                   particle=particle)
>>> blob
PlasmaBlob(T_e=55000.0*u.K, n_e=1e+23*u.cm**-3, particle='p', Z=2.5)
# before
# PlasmaBlob(T_e=55000.0 K, n_e=1e+23 1 / cm3, particle=p, Z=2.5)
```

I'll fix the tests in a moment.

PR #381, opened 2018-04-18
Create subpackage stability table in docs

namurphy:
Astropy's docs have a wonderfully useful page describing the [current status of subpackages](http://docs.astropy.org/en/stable/stability.html).  This page contains information on how stable Astropy's different subpackages are (e.g., if they're stable or undergoing active development).  It would be extremely useful for us to post that information for PlasmaPy's subpackages as well.  

Please let me know if you think we should change any of the subpackage statuses that I tentatively put in, since I wasn't quite sure what to put for `classes` since we'll be making major changes as will be described in PLEP 7 and for `diagnostics` since we're just now beginning to develop it and it's mostly in the planning stage.  I was thinking that we should err on the side of saying that a subpackage is less stable rather than suggesting that it is more stable than it actually is.

I imagine it's going to be a nice feeling in the coming years when we get to change subpackages from "planned" to "dev" to "mature" and finally to "stable".  :smile_cat:

### DIFF
--- a/docs/stability.rst
+++ b/docs/stability.rst
@@ -91,7 +91,7 @@ PlasmaPy's planned and existing subpackages are:
             <td>
                 plasmapy.atomic
             </td>
-            <td align='dev'>
+            <td align='center'>
                 <span class="dev"></span>
             </td>
             <td>
@@ -103,7 +103,7 @@ PlasmaPy's planned and existing subpackages are:
             <td>
                 plasmapy.classes
             </td>
-            <td align='planned'>
+            <td align='center'>
                 <span class="dev"></span>
             </td>
             <td>
@@ -116,7 +116,7 @@ PlasmaPy's planned and existing subpackages are:
             <td>
                 plasmapy.constants
             </td>
-            <td align='mature'>
+            <td align='center'>
                 <span class="stable"></span>
             </td>
             <td>
@@ -131,8 +131,8 @@ PlasmaPy's planned and existing subpackages are:
             <td>
                 plasmapy.diagnostics
             </td>
-            <td align='dev'>
-                <span class="planned"></span>
+            <td align='center'>
+                <span class="dev"></span>
             </td>
             <td>
                 This subpackage is in the early stages of development.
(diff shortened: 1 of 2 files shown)


docs/stability.rst
```diff
@@ -0,0 +1,197 @@
...
+            </td>
+            <td>
+                The
+                <tt class="docutils literal"><span class="pre">mathematics</span></tt>
+                subpackage is in the early stages of development.
+            </td>
+        </tr>
+        <tr>
+            <td>
+                plasmapy.parameters
```
lemmatum: Why `plasmapy.parameters` instead of `plasmapy.physics`?

docs/stability.rst
```diff
@@ -0,0 +1,197 @@
...
+                The
+                <tt class="docutils literal"><span class="pre">parameters</span></tt>
+                subpackage may undergo significant reorganization with potentially
+                major changes to the API.  We anticipate that there will be major
+                additions to this package.
+            </td>
+        </tr>
+        <tr>
+            <td>
+                plasmapy.parameters.transport
```
lemmatum: Same here, I'm pretty sure transport is a submodule of physics and not parameters :smile:

docs/stability.rst
```diff
@@ -0,0 +1,197 @@
...
+    </style>
+
+    <table align='center'>
+      <tr>
+        <td align='center'><span class="planned"></span></td>
+        <td>Planned</td>
+      </tr>
+      <tr>
+        <td align='center'><span class="dev"></span></td>
+        <td>Actively developed, be prepared for possible significant changes.</td>
```
lemmatum: I would change the comma to a full stop, like in the last couple of statuses.

docs/stability.rst
```diff
@@ -0,0 +1,197 @@
...
+      </tr>
+      <tr>
+        <td align='center'><span class="stable"></span></td>
+        <td>Reasonably stable, any significant changes/additions will generally include backwards-compatiblity.</td>
```
lemmatum: comma to full stop

docs/stability.rst
namurphy: I'm pretty sure this is just what happens when I git too little sleep!  Thanks for catching this!

StanczakDominik: > git too little sleep

OOOOOOOOOOH

PR #382, opened 2018-04-18
Equations and references in Langmuir analysis documentation

### LINKED ISSUE #369: Add Druyvesteyn method equation to Langmuir docstring
The function for obtaining the EEPF needs an equation in the docstring. It'll look something like second derivative of probe current with respect to bias equals blah.

### LINKED ISSUE #374: Add references in plasmapy.langmuir docstrings
At the moment the `plasmapy.langmuir` docstrings lack references altogether. These need to be added ASAP. A small part of the functions (namely initial fitting parameters and electron temperature-based offset of the exponential region extraction) are empirical. Is that fine?

jasperbeckers:
I'll add more equations and references as I come across them. Suggestions welcome. This also fixes a computational error in the EEDF calculation.

Closes #369 
Closes #374


plasmapy/diagnostics/langmuir.py
```diff
@@ -200,13 +200,13 @@ def swept_probe_analysis(probe_characteristic, probe_area, gas,
 
     Parameters
     ----------
-    probe_characteristic : Characteristic
+    probe_characteristic : ~plasmapy.diagnostics.langmuir.Characteristic
         The swept probe characteristic that is to be analyzed.
```
StanczakDominik: Actually if this is in the current namespace, and it is, do we need the full specification? Could you check if `Characteristic`suffices?

plasmapy/diagnostics/langmuir.py
```diff
@@ -565,6 +565,11 @@ def get_ion_density_LM(ion_saturation_current, T_e,
     probe is equal to the Bohm loss. The acoustic Bohm velocity is obtained
     from the electron temperature and the ion mass.
 
+    The ion saturation current is given by
+
+    .. math::
+        I_{is}=0.6eA_pn_i\sqrt{\frac{T_e}{m_i}}.
```
StanczakDominik: Spaces, like `A_p n_i` would make this a lot more readable outside of latex.

plasmapy/diagnostics/langmuir.py
```diff
@@ -37,7 +37,7 @@ def _fit_func_double_lin_inverse(x, x0, y0, T0, Delta_T):
 
 
 class Characteristic:
-    r"""Class representing a single I-V probe characteristic for convenient
+    r"""Class representing a single :math:`I-V` probe characteristic for convenient
```
StanczakDominik: Don't go too hard on the latex! The docstrings need to be readable primarily from the likes of ipython, pydoc and Jupyter. Sphinx is nice, but if tex doesn't add much value (here it doesn't) and make stuff less readable from plain text (here it does), skip it.

plasmapy/diagnostics/langmuir.py
```diff
@@ -209,50 +209,51 @@ def swept_probe_analysis(probe_characteristic, probe_area, gas,
...
-    plot_EEDF : bool, Optional
+    plot_EEDF : bool, optional
         If True, the EEDF is computed and shown. Default is False.
 
     Returns
     -------
-    A dictionary with the following entries:
 
-    'T_e' : ~astropy.units.Quantity
+    Results are returned as Dictionary
```
StanczakDominik: This is absolutely not something we should change or do right now, just a loose idea I just had.

Instead of a `dict`, we could return a `PlasmaBlob`.

plasmapy/diagnostics/langmuir.py
```diff
@@ -568,7 +569,7 @@ def get_ion_density_LM(ion_saturation_current, T_e,
     The ion saturation current is given by
 
     .. math::
-        I_{is}=0.6eA_pn_i\sqrt{\frac{T_e}{m_i}}.
+        I_{is} = 0.6 e A_p n_i \sqrt{\frac{T_e}{m_i}}.
```
StanczakDominik: Much better!

plasmapy/diagnostics/langmuir.py
```diff
@@ -1151,15 +1152,16 @@ def get_EEDF(probe_characteristic, visualize=False):
...
 
     The Druyvesteyn analysis results in the following equation
     [druyvesteyn-1930]_:
 
     .. math::
-        N_e(\epsilon)=\frac{2}{A_pe^2}\sqrt{\frac{2m\epsilon}{e}}
-        \frac{\textrm{d}^2I}{\textrm{d}V^2}
+        N_e \left( \epsilon \right) =
+        \frac{2}{A_pe^2} \sqrt{\frac{2 m \epsilon}{e}}
+        \frac{\textrm{d}^2 I}{\textrm{d} V^2}
```
StanczakDominik: Well formatted!

plasmapy/diagnostics/langmuir.py
jasperbeckers: Yeah, strange quirk I guess.

plasmapy/diagnostics/langmuir.py
jasperbeckers: Absolutely, but I'd prefer the user to have the option which format is returned.

StanczakDominik:
I've been seeing a lot of these `Merge branch 'master' into langmuir-update` merge commits lately. I guess this is what happens when you go too hard on required tests... I'll disable the requirement.

plasmapy/diagnostics/langmuir.py
StanczakDominik: Fair enough. I was thinking this isn't such a big deal as in essence you either get dictionary variable access
```python
result['T_e']
```
or namedtuple variable access
```python
result.T_e
```
The latter seems to me to be cleaner.

PR #383, opened 2018-04-18
Put logo into documentation index

### LINKED ISSUE #397: Fix existing Automodapi documentation, enable doc build failures again
Continue work started in #394 and fix existing documentation to not raise errors. To do this properly, this will require us to modify every single non-test .py file in the project and probably write a PLEP about it. Given I'm waiting for the build to merge that, here goes a draft... @lemmatum I wanted to "get back to you on this" a few days ago.

We currently use the wonderful sphinx automodapi plugin to automatically gather visible items in every module, use their docstrings and generate documentation from them without having to write it separately in the docs.

Here's an example from `docs/physics/index.rst:
```rst
Reference/API
=============

.. automodapi:: plasmapy.physics.parameters
```
That's literally it. However, the way this works (as I understand is) is that automodapi opens a module, inspects its namespace, takes everything that's not a variable or an imported module (remember this fact!) and for every one of those items, takes the `__doc__`string and builds pretty docs out of those.

It's neat.

But we have to consider one thing! Here's a typical line at the beginning of a PlasmaPy module:

```python
from plasmapy.atomic import particle_mass, integer_charge
```

And congratulations, you've now put two functions straight from `atomic` into your `physics.parameters` namespace, losing all information of where those functions actually came from. At least that's what seems to be happening.

There are a few ways of dealing with this:
* don't use `from` imports, import modules and refer to functions via modules. This is obviously bad as this will make our lines' length grow by a lot.
* use `skip` directives in the documentation, like this:
```rst
.. automodapi:: plasmapy.physics.quantum
   :skip: Lorentz_factor
   :skip: Fermi_integral
   :skip: check_quantity
```
But I think that's also bad, as you'll forget to add a `skip` for some import you add one day and boom, we've got the wrong items appearing.

And then there's the third approach, seemingly taken by SunPy and Astropy:
* set the `__all__`  special variable at the beginning of the module you want to document to a list of relevant items, like this:

https://github.com/sunpy/sunpy/blob/15d646f8172bcb6f239e07ebd825d1884046983e/sunpy/physics/solar_rotation.py#L15

This seems like the best option out of all of these because then you're setting explicitly what you want to document.

So, a PR to solve this issue will have to accomplish three things:
* [x] put functions and classes we want to be documenting in `__all__` for each module
* [x] add `-w` back in to `.circleci/config` so that our tests start actually failing when there's something wrong with the docs, try to make them pass on every PR
* [x] fix any current issues with HTML formatting, etc. That's not pleasant, but it'll have to be done, most likely.

StanczakDominik:
Sooo @jasperbeckers, I think something might be off  with the .svg file. At the top is a prerendered png, at the bottom is the svg:

image

Could you take a look? If the .svg is fine, do you have any other ideas?

Closes #397


Cadair:
you have to have the font to correctly render the svg, which you could add as a web font.

jasperbeckers:
Like @Cadair says, you need to have to fonts installed. I had to change the type of one of them to get it to show up in inkscape.

jasperbeckers:
Good! By the way, the "py" part of the text uses Kaffeesatz Thin, not Light.

StanczakDominik:
Which of these do you all think looks better?
* In the nav bar:

* In the main page:

namurphy:
I'd vote for putting it in the main page to avoid making people have to scroll down in the index.  Thank you for working on this!

StanczakDominik:
Okay, we have doc tests back online.

I think this will be all, so I'll head to bed and merge this after waking up unless anyone has any further suggestions.

StanczakDominik:
I remembered that @dstansby mentioned the physics docs would be more readable being split into subpages, so I went ahead and did that.

Also, it turns out automodapi does the py:module stuff on its own! I think that's pretty neat.

PR #391, opened 2018-04-23
Clean up 1D maxwellian example

dstansby:
This gives a bit of a clean to the example.

### DIFF
--- a/plasmapy/examples/plot_distribution.py
+++ b/plasmapy/examples/plot_distribution.py
@@ -38,7 +38,7 @@ print(p_dens)
 # probability. Let's test that for a bunch of particles:
 
 T = 3e5 * u.K
-dv = 10
+dv = 10 * u.m / u.s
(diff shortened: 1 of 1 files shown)


plasmapy/examples/plot_distribution.py
```diff
@@ -11,73 +12,54 @@
...
-start = -5000
-stop = - start
-dv = 10000 * u.m / u.s
-
-v_vect = np.arange(start, stop, dtype='float64') * dv
-
-print(v_vect[Maxwellian_1D(v_vect, T=T_e,
-                           particle='e', V_drift=V_drift).argmax()])
+print('T from standard deviation:', T_theo)
+print('Initial T:', T)
```
StanczakDominik: This is weird... previously, via `print(T_theo / T)` we got `0.9999999999930749`. Now, from your code, we get 

> T from standard deviation: 258377.0318828946 K
> Initial T: 300000.0 K

That's a ratio of ~0.861...

PR #395, opened 2018-04-23
Rename collision rates

StanczakDominik:


lemmatum:
Closes #312

StanczakDominik:
Okay, I think this should be doing better now.

StanczakDominik:
Okay, latex doesn't blow up anymore! I think this is mergable.

plasmapy/physics/transport/collisions.py
```diff
@@ -934,6 +935,31 @@ def collision_rate_ion_ion(T_i,
...
+    is the ion thermal velocity (the average, for a Maxwellian distribution).
+
+    Note that in the derivation, it is assumed that electrons are present
+    in such numbers as to establish quasineutrality, but the effects of the
+    test ions colliding with them are not considered here. This is a very
+    typical approximation in transport theory.
+
+    This result is an ion momentum relaxation rate, and is used in many
+    classical transport expressions. It is equivalent to:
+    * 1/tau_i from ref [1]_ eqn (1) pp. #,
```
namurphy: For the Braginskii article (ref [1]), I believe this should be equation (2.5i) on page 215.  That equation defines tau_i.  It would probably be worth checking the next two equations and page numbers to double check those too.

plasmapy/physics/transport/collisions.py
```diff
@@ -943,24 +969,29 @@ def collision_rate_ion_ion(T_i,
        revised." Naval Research Lab. Report NRL/PU/6790-16-614 (2016).
        https://www.nrl.navy.mil/ppd/content/nrl-plasma-formulary
 
-    .. [3] Callen Chapter 2, http://homepages.cae.wisc.edu/~callen/chap2.pdf
+    .. [3] J.D. Callen, Fundamentals of Plasma Physics draft material,
+       Chapter 2, http://homepages.cae.wisc.edu/~callen/chap2.pdf
 
     Examples
     --------
```
namurphy: It would be good to have one of these examples using a temperature in kelvin. :joy_cat: That way, the users would know that the temperature does not have to be in eV, and it would also provide a test.

plasmapy/physics/transport/tests/test_collisions.py
```diff
@@ -183,13 +183,13 @@ def test_GMS1_negative(self):
...
         a negative (invalid) Coulomb logarithm is returned.
         """
-        with pytest.raises(exceptions.PhysicsError, match="< 0"):
-            methodVal = Coulomb_logarithm(self.temperature2,
-                                          self.density2,
-                                          self.particles,
-                                          z_mean=np.nan * u.dimensionless_unscaled,
-                                          V=np.nan * u.m / u.s,
-                                          method="GMS-1")
+        with pytest.warns(exceptions.CouplingWarning, match="relies on weak coupling"):
```
namurphy: I have not used the `match` keyword in `pytest.warns`, but I should start doing so!

plasmapy/utils/checks.py
```diff
@@ -103,6 +103,7 @@ def wrapper(*args, **kwargs):
                 can_be_complex = validation_settings.get('can_be_complex', False)
                 can_be_inf = validation_settings.get('can_be_inf', True)
                 can_be_nan = validation_settings.get('can_be_nan', False)
+                can_be_none = validation_settings.get('can_be_none', False)
```
namurphy: Somewhere else in the code, I used a keyword with a name like `none_shall_pass`, which I think is used similarly to `can_be_none` here.  It might be helpful to check for consistency between keywords, though I made a note for myself so I'll end up checking it at some point if you don't want to do it here.

The only reason why I prefer `none_shall_pass` is that whenever I read it, I think of Gandalf (as portrayed by Ian McKellen) saying it (though admittedly it's not an exact quote).   :dragon_face: :mountain: :ring:

PR #400, opened 2018-04-24
Implement plot defaults

### LINKED ISSUE #177: Sensible plot defaults
It would be useful to have some sensible plot defaults based on matplotlib to visually unify any examples we generate in the documentation.

These could also be used for generating publication quality plots (like for #159 ) and sort through some of the hiccups in matplotlib (e.g. unifying LaTeX rendered fonts with regularly rendered fonts, line plots with shaded regions representing errors, etc.).

lemmatum:
Closes #177


StanczakDominik:
Aaaand 400! :tada:

lemmatum:
@tulasinandan here is the PR for creating plot defaults.

Things to do:
- [ ] Possibly get this to automatically import with PlasmaPy and therefore dynamically set plot defaults? Though I'm not sure we should do this - don't want to overwrite someone's matplotlibrc settings.
- [ ] Add plot functions to `__init__.py`
- [ ] Add tests for plot functions
- [ ] Make an example notebook to demonstrate plotting capabilities

tulasinandan:
Sorry for the delay in my response guys! Travel within travel, so am having some connectivity issues. @lemmatum I like the fonts that you set. I got the following error when I tried this locally:

`In [9]: matplotlib.rcParams['mathtext.fontset'] = 'dejavuserif'
<ipython-input-9-855486ca520c> in <module>()
----> 1 matplotlib.rcParams['mathtext.fontset'] = 'dejavuserif'

/usr/lib/python3/dist-packages/matplotlib/__init__.py in __setitem__(self, key, val)
    882                 cval = self.validate[key](val)
    883             except ValueError as ve:
--> 884                 raise ValueError("Key %s: %s" % (key, str(ve)))
    885             dict.__setitem__(self, key, cval)
    886         except KeyError:

ValueError: Key mathtext.fontset: Unrecognized fontset string "dejavuserif": valid strings are ['cm', 'custom', 'stix', 'stixsans']`

Other fonts worked fine. I have a few helper routines that I modified from @colbych 's Py3D library for own customizations. I will add those to the visualization directory as well.

tulasinandan:
I think we should not load the PlasmaPy visualization default globally. Only when the PlasmaPy plotting helpers are called. I had this problem with seaborn. It automatically loaded its defaults globally and it was irritating. I like their plot styles but obviously want simpler defaults for most of my work. I had to turn it off.

I guess many people are like this, so I vote for loading the defaults only when PlasmaPy plot helpers are called. If someone wants, they could load our defaults globally but we should not force it.

lemmatum:
Ah yes, we should either do a try-except block or check whether the fonts exists beforehand. 

I agree with your sentiment on global loading. We'll just keep it as an available module and show it off in examples and such.

StanczakDominik:
I think we could also do this in the docs conf.py file if we just want to apply these settings to examples. In other words, let this wait for Sunday!

lemmatum:
@tulasinandan give it a try now. I add try-except blocks and added the stix font as a fallback. I think this should cover most platforms...
You can test it out by doing something like:
```Python
# comparing fonts
plt.title(r"cm-3 $\rm cm^{-3}$")
plt.show()
```
which by default would render in two different fonts. If the the pltDefaults module works then the fonts should be the same now.

Issue #352, opened 2018-04-09
Create a class to represent ionization state distributions


namurphy:
There are a variety of situations for which plasma physicists need to keep track of ionization state distributions (e.g., what fraction of iron is neutral, singly charged, doubly charged, etc.).  We should create a base class that is able to store and access this information.  This can probably be related to the `Particle` class.  

This class should be able to be extended to have functions that can, e.g., predict the resulting spectrum.  There should probably also be a standard way to include information such as the temperature (or a more general particle distribution function) and number density, as well as relative abundances of elements.

We might want to create a way to more easily iterate across charge states and elements.

@lemmatum and @wtbarnes - This seems like a great thing to collaborate on among [SpectroscoPy](https://github.com/PlasmaPy/SpectroscoPy), [fiasco](https://github.com/wtbarnes/fiasco), PlasmaPy, and perhaps other existing spectroscopy packages.  I've been meaning to look more deeply into how [fiasco](https://github.com/wtbarnes/fiasco) and [specutils](https://github.com/astropy/specutils) store these data.

Cross-reference: https://github.com/NEI-modeling/NEI/issues/18

wtbarnes:
When using the CHIANTI database, one can obtain the ionization fractions 1 of 2 ways,

* Access pre-computed ionization fractions stored in CHIANTI tabulated at different temperatures (there are multiple datasets available for the same elements/ions)
* Compute the ionization fractions yourself using the ionization and recombination rates 

Note that in both cases, it is assumed that the rates (and by extension the fractions) are **not** density-dependent and that the ionization and recombination rates are **in equilibrium** (though of course the rates can be used to compute the time-dependent ionization states).

In fiasco, you can compute the the ionization fractions as a function of temperature using either method, e.g. for iron,

```python
import numpy as np
import astropy.units as u
import fiasco
import matplotlib.pyplot as plt
iron = fiasco.Element('iron', np.logspace(4,9,100)*u.K)
# Compute ionization fractions using ionization and recombination rates
ioneq = iron.equilibrium_ionization()
for i,ion in enumerate(iron):
    # Method 1--precomputed data from database
    plt.plot(ion.temperature, ion.ioneq, f'C{i%10}', ls='--')
    # Method 2--computed from rates
    plt.plot(ion.temperature, ioneq[:,i], f'C{i%10}', ls='-', label=ion.ion_name)
plt.xscale('log')
plt.show()
```
[...]

PR #404, opened 2018-04-24
Create classes to represent ionization state distributions

namurphy:
My plan for this PR is to create classes to represent the ionization state distributions of one or more elements.  I am going to add in a bunch of dunder methods like `__getitem__` and maybe `__call__` to help making access to the ionization states more straightfoward and intuitive.  Any suggestions on the naming convention will be helpful so that we can maximize readability. 

Eventually we'll need a way to calculate ionization state distributions assuming collisional ionization equilibrium, but that will be for a different PR.  The purpose of this PR is to set up how to store and access the ionization distributions.  This will be discussed in #352.

This will address some of #352.  It will probably be best to wait until after the `0.1.0` release to merge this, since this PR is only for a partial implementation anyway.


plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,87 @@
...
+            raise NotImplementedError(
+                "IonizationState instances are not yet able to "
+                "calculate collisional ionization equilibrium.")
+
+        if isinstance(distribution, (tuple, list)):
+            distribution = np.array(distribution)
+        elif not isinstance(distribution, np.ndarray):
+            raise TypeError
+
+        if len(distribution) != element.atomic_number + 1:
```
lemmatum: We should probably be a bit more strict. Either here or in another spot check for shape of the array (I assume it should be 1D).

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,87 @@
...
+            raise AtomicError("Incorrect number of ionization states.")
+
+        ionsum = np.sum(distribution)
+        if not np.isclose(ionsum, 1.0):
```
lemmatum: add absolute and relative error tolerances and present them as default arguments for the class/method so that the user can adjust them?

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,87 @@
...
+    def __getitem__(self, value):
+        if isinstance(value, int) and 0 <= value <= self.atomic_number:
+            return self.ionization_state[value]
+        elif isinstance(value, Particle):
+            if value.element == self.element:
+                return self.ionization_state[value.integer_charge]
+
+        # TODO: Add slicing
+
+    def __str__(self):
```
lemmatum: might need to add `__repr__` as well.

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,87 @@
...
+        ...
+
+
+
+    @property
+    def atomic_number(self):
+        return self._element.atomic_number
+
+    @property
+    def nstates(self):
```
lemmatum: this name is too generic (states refers to so many things!) maybe rename to `nChargeStates`? Or something similar?

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,87 @@
...
+
+        ionsum = np.sum(distribution)
+        if not np.isclose(ionsum, 1.0):
+            raise ValueError(
+                f"The sum of the ionization states of {self.element} "
+                f"is {ionsum}, which does not approximately equal one.")
+
+        self._ionization_state = distribution
+
+    def __getitem__(self, value):
```
lemmatum: add `__iter__` so we can iterate through charge states.

StanczakDominik: Probably not going to be trivial, especially on the design front. But it's an idea.

namurphy: I've never implemented `__iter__` before, so that sounds really fun!  It will probably also be really useful.

It's also kind of funny to have a magic method in a plasma package called `__iter__`!  ...though I suppose it would be funnier if it were `__ITER__`.

lemmatum: also worth noting, you'll want to implement `__next__` along with `__iter__`

namurphy: Thanks, I didn't know that!  I would have guessed `__DEMO__`.

namurphy: I'm wondering how best to do this iteration, since we might want to get any combination of the integer charge, the `Particle` instance, the ionic symbol, and the ionization fraction for that state.  Maybe a `collections.namedtuple`?  Something like...

```Python
for state in ionization_states:  # using maximally verbose attribute nomenclature
    print(state.integer_charge, state.ionization_fraction, state.particle_instance)
```

lemmatum: I think a tuple is reasonable, especially since then you could do:
```Python
for charge, fraction, particle in ionizationInfo:
    # important physics here
```

plasmapy/atomic/tests/test_ionization_state.py
```diff
@@ -0,0 +1,297 @@
...
+        instance = self.instances[test_name]
+
+        inputted_fractions = test_cases[test_name]['ionic_fractions']
+        if isinstance(inputted_fractions, u.Quantity):
+            inputted_fractions = inputted_fractions.to(u.m ** -3)
+            inputted_fractions = (inputted_fractions / inputted_fractions.sum()).value
+
+        try:
+            assert np.allclose(instance.ionic_fractions, inputted_fractions)
+        except:
```
ritiek: I am guessing, we're going to catch (only) `AssertionError` here. If so, it would be a good idea if we specifically mention that (`except AssertionError: ...`). We could also display values for `instance.ionic_fractions` and `inputted_fractions` when re-throwing the error. :blush:

plasmapy/atomic/tests/test_ionization_states.py
```diff
@@ -0,0 +1,304 @@
...
+            try:
+                expected = instance.ionic_fractions[key]
+            except Exception as exc:
+                raise AtomicError(
+                    f"Unable to get ionic_fractions for '{key}' in "
+                    f"test='{test}'.") from exc
+
+            try:
+                actual = instance[key].ionic_fractions
+            except Exception as exc:
```
ritiek: I don't think we need `as exc` if its not being used when re-throwing the error.

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+    Only one of `n_e` and `n_elem` may be set.
+
+    """
+
+    # TODO: Allow this class to (optionally?) handle negatively charged
+    # ions.  There are instances where singly negatively charged ions
+    # are important in astrophysical plasmas, such as H- in the
+    # atmospheres of relatively cool stars.  There may be some rare
+    # situations where doubly negatively charged ions show up too,
+    # though triply negatively charged ions are very unlikely.
```
StanczakDominik: What would it take to allow this? A list like `positive` = [True, False]`? Or would you need to provide, instead of `atomic_number +1`, `2atomic_number + 1` floats to the second argument?

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+
+    # TODO: Add in functionality to find equilibrium ionization states.
+    # How much data will this require?
```
StanczakDominik: That sounded distinctly machine-learning-ish.

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+            self.tol = tol
+            self.T_e = T_e
+            self.n_elem = n_elem
+            self.n_e = n_e
+            self.ionic_fractions = ionic_fractions
+
+            # This functionality has not yet been implemented:
+
+            # if self._ionic_fractions is None and self.T_e is not None:
+            #     self.equilibrate()
```
StanczakDominik: `raise NotImplementedError` exists for this very purpose, and it's going to show us an uncovered line!

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+
+        except Exception as exc:
+            raise AtomicError(
+                f"Unable to create IonizationState instance for "
+                f"{particle.particle}.") from exc
+
+    def __getitem__(self, value) -> State:
+        """Return the ionic fraction(s)."""
```
StanczakDominik: It'd be nice to have a few examples here of how you'd like this to work.

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+                if not same_element or not same_isotope:
+                    raise AtomicError("Inconsistent element or isotope.")
+                elif not has_charge_info:
+                    raise ChargeError("No integer charge provided.")
+        return result
+
+    def __iter__(self):
+        """Initialize the instance prior to an iteration."""
+        self._charge_index = 0
+        return self
```
StanczakDominik: Oh dear, you've actually done it! :smile:

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+            result = State(
+                self._charge_index,
+                self._ionic_fractions[self._charge_index],
+                self.ionic_symbols[self._charge_index],
+            )
+            self._charge_index += 1
+            return result
+        else:
+            del self._charge_index
+            raise StopIteration
```
StanczakDominik: I'm just going through this file before going to bed today, so I'll finish the review later, but: this definitely needs a test. If it already has one, awesome.

... how many lines, again?!

plasmapy/atomic/ionization_states.py
```diff
@@ -0,0 +1,965 @@
...
+        Examples
+        --------
+        >>> hydrogen_states = IonizationState('H', [0.9, 0.1])
+        >>> hydrogen_states.ionic_fractions
+        array([0.9, 0.1])
+
+        """
+        return self._ionic_fractions
+
+    @ionic_fractions.setter
```
StanczakDominik: Okay, that's probably enough Jython for today. Note to self, continue later!

namurphy:
Thanks for the code review, @StanczakDominik!  We've been using this and related functionality in a separate but related [project for non-equilibrium ionization (NEI) modeling](https://github.com/NEI-modeling/NEI) that we're still developing and refining, and that I'm hoping will get implemented into a different project.  

There are a few things I've been wondering about on larger scales.  In which project should the `IonizationStates` class (or something similar) go?  And also...where should NEI modeling functionality go?  The ionization state of a plasma is a really basic property, so I've been thinking that it should go in PlasmaPy, even though fiasco and SpectroscoPyx are also really good possibilities.  NEI modeling is a bit more specialized, but the data on recombination and ionization rates is pretty much the same data that's needed to calculate equilibrium ionization fractions, which is a really fundamental thing too so PlasmaPy might be the best place.  

I should probably create a separate issue to ask this question, but I'm going on vacation in three minutes so it'll have to wait! :bullettrain_front: :bridge_at_night:

plasmapy/atomic/ionization_state.py
```diff
@@ -0,0 +1,501 @@
+from typing import Dict, List, Optional, Tuple, Union
```
StanczakDominik: I don't quite understand the split between these two similarly named files. If there really is a lot to do here, maybe a subdirectory within `atomic` would be a better idea?

plasmapy/atomic/ionization_state.py
```diff
@@ -0,0 +1,501 @@
...
+from plasmapy.atomic.particle_class import Particle
+from plasmapy.atomic.particle_input import particle_input
+from plasmapy.utils import (AtomicError, ChargeError, InvalidParticleError, check_quantity)
+
+State = collections.namedtuple(
+    'State', [
+        'integer_charge',
+        'ionic_fraction',
+        'ionic_symbol',
+    ])
```
StanczakDominik: Ooh, we could probably use a 3.7 dataclass here! It's not necessary by any means, but I'd just like to note down the possibility. I've been itching to try those.

AND I HAD TO WRITE 3.5 COMPATIBLE PYTHON AT WORK, SO NO F-STRINGS :sob: 

So yeah, I think I'm gonna write something therapeutic in 3.7. :laughing:

plasmapy/atomic/tests/test_ionization_state.py
```diff
@@ -294,3 +340,72 @@ def test_execeptions(test):
...
+    as expected.
+    """
+
+    expected = expected_properties[key]
+    actual = eval(f'instance.{key}')
+
+    if isinstance(expected, u.Quantity):
+        assert expected.unit == actual.unit, \
+            f"Unit mismatch for IonizationState.{key}"
+        assert np.allclose(expected, actual, atol=1e-15 * expected.unit), \
```
StanczakDominik: I can and will keep linking [assert_quantity_allclose](http://docs.astropy.org/en/stable/api/astropy.tests.helper.assert_quantity_allclose.html) for as long as necessary. :laughing:

namurphy: Ah!  I forgot about that.  Good catch!

plasmapy/atomic/tests/test_ionization_states.py
```diff
@@ -335,12 +327,13 @@ def test_execeptions(test):
 
 
 def test_setitem():
+    # TODO: parametrize this
```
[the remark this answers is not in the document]
StanczakDominik: The test seems to fail currently on the line below. I guess the API changed and the test wasn't updated?

StanczakDominik:
Also, to make sure this doesn't get lost in GitHub's review interface, [this comment discusses the test failure](https://github.com/PlasmaPy/PlasmaPy/pull/404#discussion_r224181200).

namurphy:
> **Transport PR Incident**

:joy_cat: 

I still have more tests and docstrings to put in, but I don't have to finish all of them in this PR.  I was planning on working on this for one more day when I can take some time away from Fortran, and then start working on #521.  This is the last of my backlog of PRs to finish up, so I'm eager to get it done soon!

plasmapy/atomic/ionization_state.py
```diff
@@ -136,15 +136,60 @@ def __str__(self) -> str:
...
+        if np.isfinite(self.T_e):
+            T_e = "{:.2e}".format(self.T_e.value) + " K"
+            output += f'  T_e    = {T_e}\n'
+        if np.isfinite(self.n_elem):
+            n_elem = "{:.2e}".format(self.n_elem.value) + " m**-3"
+            n_e = "{:.2e}".format(self.n_e.value) + " m**-3"
+            output += f'  n_elem = {n_elem}\n'
+            output += f'  n_e    = {n_e}'
+
+        return output
```
StanczakDominik: Oh wow, that's a pretty dense `__repr__`. It's probably no issue here, but it got me thinking... dense enough `__repr__`s (probably in some limit...  that might fall under High Energy Python or something more relative) may actually slow down PyCharm and other "smarter" editors (vim doesn't care) which render `__repr__`s dynamically and often.

plasmapy/atomic/ionization_state.py
```diff
@@ -327,18 +378,19 @@ def n_elem(self, value: u.m ** -3):
...
     def number_densities(self) -> u.m ** -3:
         """Return the number densities for each state."""
-        if not np.isnan(self._n_elem):
+        try:
             return (self.n_elem * self.ionic_fractions).to(u.m ** -3)
-        else:
-            raise AtomicError(
-                "Insufficient information to return number densities.")
+        except Exception:
+            return np.full(self.atomic_number + 1, np.nan) * u.m ** -3
```
StanczakDominik: Is there any way we could potentially make this more selective than the broad `except Exception` clause?

plasmapy/atomic/ionization_state.py
namurphy: My motivation here was so that people using this class could quickly diagnose what the ionization state is for this particular element when working on the command line.  
```Python
>>> IonizationState('He', [0.1, 0.9, 0.0], n_e=4e19*u.m**-3, T_e=4e4*u.K)
IonizationState instance of He with Z_mean = 0.90

  He  0+: 0.100     n_i = 4.00e+18 m**-3
  He  1+: 0.900     n_i = 3.60e+19 m**-3

  T_e    = 4.00e+04 K
  n_elem = 4.00e+19 m**-3
  n_e    = 3.60e+19 m**-3
```

StanczakDominik: Oh wait, I just noticed the newlines. That might well crash stuff, though I'm not certain it's actually going to happen. I'll look into it later.

StanczakDominik: Okay, I just checked and this is completely fine. Turns out pycharm and the like use `str`, not `repr`.

image

And here's the result with `str` and `repr` swapped (the gist of it is: nothing blew up):
image

StanczakDominik:
The Travis builds are only failing because I messed up a few git remote magic tricks; they're passing [over here](https://travis-ci.org/PlasmaPy/PlasmaPy/builds/441512524).

plasmapy/atomic/particle_class.py
```diff
@@ -540,6 +540,9 @@ def __ne__(self, other) -> bool:
         """
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return hash(self.__repr__())
```
StanczakDominik: You got any more of them cat emojis? I could also use one right now.

To the best of my current and badly limited knowledge, I think this is okay.

plasmapy/atomic/tests/test_ionization_state.py
```diff
@@ -244,34 +251,92 @@ def test_particles(self, test_name):
...
+        when using equivalent keys (integer charge, particle symbol, and
+        Particle instance).
+
+        For example, if we create
+
+        >>> He_states = IonizationState('He', [0.2, 0.3, 0.5])
+
+        then this checks to make sure that `He_states[2]`,
+        `He_states['He 2+']`, and `He_states[Particle('He 2+')]` all
+        return the same result.
```
StanczakDominik: Huh. That looks like a funny interface, but... sure, I guess.

plasmapy/atomic/ionization_state.py
```diff
@@ -233,6 +235,17 @@ def __getitem__(self, value) -> State:
...
+    def __setitem__(self, key: Any, value: Any):
+        """
+        Raise a `NotImplementedError` because the ionic fractions for
+        different ionization levels should be set simultaneously due to
+        the normalization constraint.
+        """
+        raise NotImplementedError(
+            "Dictionary assignment of an IonizationState instance is "
+            "not allowed. because the ionic fractions of for different "
+            "ionization levels should be set simultaneously.")
```
StanczakDominik: Looks like a good future issue!

plasmapy/atomic/ionization_state.py
```diff
@@ -455,43 +497,64 @@ def equilibrate(self, T_e: u.K = None):
...
+        sum becomes equal to one.
+        """
+        self._ionic_fractions = self._ionic_fractions / np.sum(self._ionic_fractions)
 
     @property
-    def element(self) -> str:
+    def _element(self) -> str:
         """Return the atomic symbol of the element."""
-        return self._particle.element
+        return self._particle_instance.element
```
StanczakDominik: I'm not quite sure why you would make this a private-ish property.

plasmapy/atomic/ionization_state.py
namurphy: For this, I'm thinking that dictionary assignment should not be allowed because of the normalization constraint that the sum of ionic fractions must equal one.  If we changed the ionic fraction of 'Fe 2+', then we would have to change other levels too at the same time.  It's probably better to do something like:
```Python
>>> helium_states = IonizationState('He', [0.5, 0.3, 0.2])
>>> helium_states[2] = 0.4  # raises exception because of normalization constraint
>>> helium_states.ionic_fractions = [0.6, 0.3, 0.1]  # this should work fine
```

plasmapy/atomic/ionization_state.py
namurphy: I created the `base_particle` attribute to return the element or isotope that an `IonizationState` instance is keeping track of, and to be more general.  On the other hand, there may be times when the element in particular is needed to be known, like to look up stuff in an atomic database which is broken down by element.  I'll go ahead and de-privatize this.

namurphy:
The downside of having `__repr__` return multiple lines of diagnostic data is that we can end up with messy error messages with `'\n'` showing up a bunch of times.
```
___________ Test_IonizationStates.test_inequality[number densities -> inputs] ___________
plasmapy/atomic/tests/test_ionization_states.py:159: in test_inequality
    assert original_instance != altered_instance
E   assert IonizationStates instance for: H, He\n\n  H  0+: 0.400\n  H  1+: 0.600\n\n  He  0+: 0.217\n  He  1+: 0.304\n  He  2+: 0.478\n\n != IonizationStates instance for: H, He\n\n  H  0+: 0.400\n  H  1+: 0.600\n\n  He  0+: 0.217\n  He  1+: 0.304\n  He  2+: 0.478\n\n
```
I'm still wondering if being able to easily print out really useful diagnostic data like this is worth it.  Hm...
```Python
In [1]: IonizationStates({'H': [0.1, 0.9], 'He': [0.9, 0.099, 0.001]})
Out[1]: 
IonizationStates instance for: H, He

  H  0+: 0.100
  H  1+: 0.900

  He  0+: 0.900
  He  1+: 0.099
```

namurphy:
> Re: repr, could that perhaps be put into another custom method?

Yeah, that's what I was starting to think too, so let's go with this approach.  Thanks!

plasmapy/atomic/ionization_states.py
```diff
@@ -241,20 +239,45 @@ def __eq__(self, other):
...
+            this = eval(f"self.{attribute}")
+            that = eval(f"other.{attribute}")
+
+            this_equals_that = np.any([
+                this == that,
+                this is that,
+                np.isnan(this) and np.isnan(that),
+                np.isinf(this) and np.isinf(that),
+                u.quantity.allclose(this, that, rtol=min_tol),
+            ])
```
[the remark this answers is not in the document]
namurphy: Yeah, it's a pain to arbitrarily compare things that might be `nan`, `inf`, a `Quantity`, and perhaps even things that don't have `__eq__` implemented.  Might be worth creating a function called something like `same_enough` in `plasmapy.utils.pytest_helpers.py` that holds all of this, especially due to code duplication between `IonizationState` and `IonizationStates`.  Other suggestions on how to do this better would be appreciated!

I'm kind of proud of the names `this` and `that`, since it makes the code really readable (to me at least).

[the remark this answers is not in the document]
StanczakDominik: That it does, it gave me a wide smile when I saw that :D

plasmapy/atomic/ionization_states.py
```diff
@@ -276,10 +361,16 @@ def __eq__(self, other):
                 ])
 
                 if not this_equals_that:
+
+                    print('this = ', this)
+                    print('that = ', that)
```
StanczakDominik: Leftover debug prints, possibly?

plasmapy/atomic/ionization_states.py
```diff
@@ -592,6 +688,30 @@ def ionic_fractions(self, inputs: Union[Dict, List, Tuple]):
                     if not np.isclose(np.sum(fractions), 1, atol=self.tol, rtol=0):
                         raise AtomicError(f"Ionic fractions for {key} are not normalized to 1.")
 
+            # When the inputs provide the densities, the abundances must
+            # cannot have been provided because that would be redundant
```
StanczakDominik:
```suggestion
            # not have been provided because that would be redundant

Comment (StanczakDominik) [reply plasmapy/atomic/ionization_states.py]: Ooooh, that works! Shiny!

File: plasmapy/atomic/tests/test_ionization_states.py [hunk 28]
```diff
@@ -735,6 +767,6 @@ def test_equality(self, this, that):
             print(f"{that} kwargs:\n {self.dict_of_kwargs[that]}\n")
             self.instances[that].info()
             descriptor = "equal" if expect_equality else "unequal"
-            raise AtomicError(
+            raise pytest.fail.Exception(
```
StanczakDominik: So I had to check the docs for this...

```
In [2]: pytest.fail.Exception?
Init signature: pytest.fail.Exception(msg=None, pytrace=True)
Docstring:      raised from an explicit call to pytest.fail() 
Type:           type

In [3]: pytest.fail?
Signature: pytest.fail(msg='', pytrace=True)
Docstring:
Explicitly fail an executing test with the given message.

:param str msg: the message to show the user as reason for the failure.
:param bool pytrace: if false the msg represents the full failure information and no
    python traceback will be reported.
File:      ~/.local/lib/python3.7/site-packages/_pytest/outcomes.py
Type:      function

```

So... maybe just this?
```suggestion
            pytest.fail(
```

plasmapy/atomic/ionization_states.py
```diff
@@ -234,37 +234,78 @@ def __getitem__(self, *values) -> IonizationState:
...
+            try:
+                new_number_densities = value.to(u.m ** -3)
+            except u.UnitConversionError:
+                raise ValueError(
+                    f"{errmsg} because the units of value do not "
+                    f"correspond to a number density.") from None
+
+            old_n_elem = np.sum(self.number_densities[particle])
+            new_n_elem = np.sum(new_number_densities)
+            if not u.quantity.allclose(old_n_elem, new_n_elem, rtol=self.tol):
```
StanczakDominik: Yay! :1st_place_medal:

plasmapy/atomic/ionization_states.py
```diff
@@ -112,7 +128,43 @@ def __str__(self) -> str:
...
+        ... )
+        >>> states.info()
+        IonizationStates instance for: H, He
+        <BLANKLINE>
+          H  0+: 0.100     n_i = 3.00e+14 m**-3
+          H  1+: 0.900     n_i = 2.70e+15 m**-3
+        <BLANKLINE>
+          He  0+: 0.950     n_i = 2.85e+14 m**-3
+          He  1+: 0.050     n_i = 1.50e+13 m**-3
+        <BLANKLINE>
```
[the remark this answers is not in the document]
namurphy: Decided to use your suggestion and went with `64 * '-'`.

namurphy:
The docstrings still need to describe the consequences of dunder methods and it'd be good to fill in a few more gaps in text coverage, but I am ready for this to be merged.  After then, I can create separate issues for the remaining tasks.

StanczakDominik:
I'd just do one final round of review and likely merge this today - but what of the commit messages? I could try to trim some of it down to something more manageable. Is that okay?

namurphy:
I was thinking that it'd be easiest if I did the squash merge after you approve it or I make any final changes.  That way I can remove commits that are now extraneous and maybe add some context to the remaining commits.  Does that sound good?

namurphy:
Thank you, @StanczakDominik!  I am very ready for this to be merged. :joy_cat:  I am working on the summary commit message now.  

A final thought is that some of the tests could likely be refactored into fixtures.  I did not do this because I do not yet know how to write fixtures, but that is fine to leave for the future.

StanczakDominik:
That sounds like a possible future low priority issue. :laughing:

StanczakDominik:
:tada:

PR #406, opened 2018-04-25
Replace pickle with numpy.load

### LINKED ISSUE #379: Use numpy.load and save instead of pickle
As @StanczakDominik put it:

> By the way, apparently pickle is unsafe due to allowing arbitrary code execution, and we're now including those in Langmuir samples. @jasperbeckers do you think we could transition to numpy.save and numpy.load .npz files? We're just storing two arrays in each of those anyway, right?

jasperbeckers:
Closes #379


plasmapy/examples/plot_langmuir_analysis.py
```diff
@@ -22,8 +21,8 @@
 # a Electron Energy Distribution Function (EEDF) as well.
 
 # Load the bias and current values stored in the .p pickle file.
-path = os.path.join("langmuir_samples", "Beckers2017.p")
-bias, current = pickle.load(open(path, 'rb'))
+path = os.path.join("langmuir_samples", "Beckers2017.npy")
+bias, current = np.load(open(path, 'rb'))
```
StanczakDominik: Numpy just takes the path here. This way you wouldn't close the file. Actually, you didn't close it in the first implementation anyway, come to think about it :D

Issue #342, opened 2018-04-06
Check that correct velocity is used in collision_frequency


lemmatum:
`collision_frequency` uses the Coulomb logarithm, but in [most places](https://github.com/PlasmaPy/PlasmaPy/blob/28156768b242564265e33b3baf8ae231f1a7c166/plasmapy/physics/transport/collisions.py#L664) the argument for velocity to `Coulomb_logarithm` is `np.nan`. I'm not sure that this is correct.

``` Python
        cou_log = Coulomb_logarithm(T,
                                    n,
                                    particles,
                                    z_mean,
                                    V=np.nan * u.m / u.s,
                                    method=method)
```

We should check if just passing the velocity would be more efficient and correct.

lemmatum:
Related to #348

StanczakDominik:
Let's do this for 0.2

PR #407, opened 2018-04-25
Verify custom velocities not being passed in

### LINKED ISSUE #348: Check if we should pass user-defined velocity to coulomb_logarithm in collision_frequency
`collision_frequency` in `collisions.py` uses the `coulomb_logarithm` function to account for small angle collisions, but in all cases `V=np.nan * u.m / u.s` is passed, which means that `coulomb_logarithm` defaults to using the reduced mass thermal velocity instead of whichever velocity the user has passed (if they passed one). I think this needs to be fixed, and it'll break the tests where a velocity is indeed passed.

StanczakDominik:
I started doing a little bit of tinkering with this to hopefully close #348. Just putting this in here for now so the tests show up.

### DIFF
--- a/plasmapy/transport/collisions.py
+++ b/plasmapy/transport/collisions.py
@@ -763,7 +763,7 @@ def collision_frequency(T,
                                     n,
                                     particles,
                                     z_mean,
-                                    V=np.nan * u.m / u.s,
+                                    V=V,
                                     method=method)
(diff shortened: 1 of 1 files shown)


lemmatum:
Might also close #342

StanczakDominik:
And after 12 months, this is ready for review! :D

PR #408, opened 2018-04-25
[WIP] Add custom mass arguments to functions in Transport

### LINKED ISSUE #328: Add custom mass argument for functions in `collisions.py`
The reduced mass that is currently used throughout `collisions.py` is currently in conflict with different definitions for these transport functions. One way to easily implement other definitions of transport functions is to just have the particle mass as an optional argument. This would allow us to avoid calculating strange coefficients between different definitions of things like collisional frequency by hand.  Once this is implemented, refactor `collision_rate_ion_ion` and similar functions.

StanczakDominik:
Basically, I did the mechanical part, which is putting custom mass overrides everywhere. I think I went overboard with it a bit. @lemmatum, you probably know better when these should actually happen!

Closes #328

### DIFF
--- a/plasmapy/physics/transport/collisions.py
+++ b/plasmapy/physics/transport/collisions.py
@@ -20,14 +20,16 @@ from plasmapy.mathematics import Fermi_integral
 
 
 @utils.check_quantity({"T": {"units": u.K, "can_be_negative": False},
-                       "n_e": {"units": u.m ** -3}
+                       "n_e": {"units": u.m ** -3},
+                       "mass_override": {"units": u.kg, "can_be_nan": True},
                        })
 def Coulomb_logarithm(T,
                       n_e,
                       particles,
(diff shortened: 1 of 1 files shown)


plasmapy/physics/transport/collisions.py
```diff
@@ -253,7 +256,10 @@ def _boilerPlate(T, particles, V):
     charges = [np.abs(p.charge) for p in particles]
 
     # obtaining reduced mass of 2 particle collision system
-    reduced_mass = atomic.reduced_mass(*particles)
+    if not np.isinf(mass_override):
+        reduced_mass = mass_override
```
lemmatum: I would rename these to `mass = mass_overide` and `mass = atomic.reduced_mass(*particles)`

StanczakDominik:
I will get to this when refactoring transport some time this year. But I'll do it better, now.

PR #409, opened 2018-04-27
Merge doc updates and other changes from master to stable prior to 0.1.0

namurphy:


StanczakDominik:
This doesn't seem like the way to go for this, as now we'll have merge commits on stable that don't happen on master! The way I'd do this would be to finish doing whatever we want on master, then checkout stable, reset --hard master and git push to let the remote catch up. We'll now need to do a single push -f to remove the merge commit.

namurphy:
Thank you, I was wondering about how to go about this!  When you have a chance, would you be able to add this to the [release guide](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/development/release_guide.rst)?

StanczakDominik:
Sure, will do. Thanks for the reminder!

PR #435, opened 2018-05-04
Subclassing Quantity to implement uncertainty

jasperbeckers:
This is a first (unsolicited) foray into the development of an intuitive uncertainty system built atop AstroPy's `Quantity` class. The class `UncertaintyQuantity` extends `Quantity` by adding the property `uncertainty`. Uncertainty then propagates through operations whenever an implemented `ufunc` is called. Highly experimental of nature but mostly functional for basic operations.

The purpose of this PR is not necessarily a merge into PlasmaPy but rather to share the development so far and to open it up to contributions from others, as well as to foster discussion on what the best place for something like this would be.

### DIFF
--- a/plasmapy/examples/plot_uncertainty.py
+++ b/plasmapy/examples/plot_uncertainty.py
@@ -3,15 +3,17 @@
 Uncertainty propagation
 =======================
 
-A short demonstration of how `UncertaintyQuantity` can be used to compute uncertainties.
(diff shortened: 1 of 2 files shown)


dstansby:
I would seriously recommend working on this upstream in astropy; this is a really nifty feature that is definitely not plasma physics specific, and eventually it should go alongside the current units implementation in astropy anyway.

StanczakDominik:
I think I agree with @dstansby.

Issue #428, opened 2018-05-01
add axis labels to ITER parameters plot


lemmatum:
The ITER parameters plot [here](http://docs.plasmapy.org/en/stable/auto_examples/plot_physics.html#sphx-glr-auto-examples-plot-physics-py) could use some labels and units in said labels.

manasbedmutha98:
I would like to take this.

lemmatum:
go for it @manasbedmutha98

manasbedmutha98:
I have added the labels. Should I also label the data point in the scatter plot differently?
Another thing I didn't understand is where this is being used.
https://github.com/PlasmaPy/PlasmaPy/blob/a452dadd51d612120588c6551dbca7a958b45ea5/plasmapy/examples/plot_physics.py#L12

manasbedmutha98:
@lemmatum Should the data point in the scatter plot be labelled separately?

lemmatum:
I think that's a question for @StanczakDominik - what are you trying to show with the scatter point in relation to the plotted line?

StanczakDominik:
Nah, no need! We're not comparing our results to any particular ITER data, though that could be, I suppose, arranged. Also, the `Axes3D` import is indeed useless for now. We could get rid of it for now.

PR #438, opened 2018-05-06
Adding online_help; Closes #277

### LINKED ISSUE #277: Create an online_help function?
Astropy has an [`online_help`](https://github.com/astropy/astropy/blob/v3.0/astropy/__init__.py#L293) function that lets users query their online documentation from the command line.  This then opens the results in their default web browser.  We may wish to create such a function for PlasmaPy, which could go in our main `__init__.py` or perhaps somewhere in `plasmapy/utils`.

We could use much of the machinery that has already been developed for Astropy's function, and just change it to suit our needs.  We've already put Astropy's BSD license in our `licenses` directory so we have fulfilled the requirements to use any of their code (yay open source!), though we should put a link to the license in any files that include code directly from Astropy.  

One special case worth thinking more about is when users query about units.  In those cases, it may end up being most worthwhile to bring up the documentation for astropy.units within Astropy's documentation.

Thanks!
-Nick

manasbedmutha98:
This adds the online_help through command line as discussed in #277. Units will redirect to Astropy.units while rest will redirect the search query to the stable version of PlasmaPy documentation from plasmapy.org


manasbedmutha98:
I went through the Plasmapy Testing Documentation [here](http://docs.plasmapy.org/en/latest/development/testing_guide.html#code-coverage) but couldn't understand what is to be done for the coverage checks. I request you to please help.

ritiek:
Looks like astropy does not include a test for this either. Quoting from here astropy/astropy#3642:

> not sure how to test it without going to the lengths of installing a puppet web browser on Travis-CI.

plasmapy/__init__.py
```diff
@@ -33,3 +33,26 @@ class UnsupportedPythonError(Exception):
...
+    query : str
+        The search query.
+    """
+    from urllib.parse import urlencode
+    import webbrowser
+
+    url = 'http://docs.plasmapy.org/en/stable/search.html?\
+    {0}&check_keywords=yes&area=default'.format(urlencode({'q': query}))
+
+    if((query.lower() == 'units') or (query.lower() == 'unit')):
```
ritiek: This looks nice to me but @namurphy and others would probably give the final call if this check is needed. You could write it something like this:
```python
...
query_lower = query.lower()
if query_lower == 'units' or query_lower == 'unit'
    ...
```
to avoid compute `query.lower()` twice. :smile:

lemmatum: Another way to do it (if you don't need to use `query_lower` ever again)
```Python
if query.lower() in ['units', 'unit']:
  ...
```

ritiek: That's even better! We most likely want to use a tuple instead of a list though (just a bit faster) `('units', 'unit')`.

manasbedmutha98:
I wanted to close since I have to generate a new PR for #428 
I hope it's fine to close before not merging? Will reopen once new PR created. Sorry for the trouble.

SolarDrew:
No trouble @manasbedmutha98 :) But you can have more than one PR open and we can always hold off on merging  a PR if you have more to add to it. Anyway, thanks for the contribution :D

PR #451, opened 2018-05-09
Remove maxwellian_speed_3D

dstansby:
As I've mentioned a couple of times in the past, I'm not really sure what `Maxwellian_speed_3D` computes. This is a bit of a provocative PR, but unless someone can provide a mathematical explanation of the number it produces at end, I think it should be deleted.


lemmatum:
And here I was under the impression that I had explained it in Riot... I guess I'll give it another go.

dstansby:
Sorry, I'm still not convinced! At best it needs a correct docstring equation.

lemmatum:
I'm 90% the docstring equation is correct, and I thought we had agreed that the confusion was that you didn't see how the code translates into the docstring equation, which I then subsequently explained... 

Hold on, I'll make it clearer by writing it up in LaTeX.

lemmatum:
I think I've found the source of the confusion... `Maxwellian_speed_1D` is wrong, it is actually what `Maxwellian_speed_3D` should be.

plasmapy/physics/distribution.py
```diff
@@ -410,168 +410,6 @@ def Maxwellian_speed_1D(v,
         return distFunc
 
 
-def Maxwellian_speed_3D(vx,
```
lemmatum: This can be removed, but `Maxwellian_speed_1D` needs to be renamed to `Maxwellian_speed_3D` and moved here. Docstrings and tests should be changed accordingly. We then need to create a correct implementation for  `Maxwellian_speed_1D`.

lemmatum:
I'm closing this PR, it'll be handled by #452

Issue #387, opened 2018-04-20
Reorganize distribution.py


StanczakDominik:
The code in distribution.py has multiple functions for the same physical distribution. This could probably use an object oriented interface!

In fact, it could be a good idea to look at how scipy.stats does their probability distribution functions and basically subclass that interface. That, however, is worth a lot more of bonus points, so it's more of an optional thing for later. Or it could be that this saves a lot of planning the user interface, so it might be the easy way to do this. I'm not sure myself.

lemmatum:
To be clear, we have:
* 1D Maxwellian speed distribution
* 1D Maxwellian velocity distribution
* 3D Maxwellian speed distribution
* 3D Maxwellian velocity distribution
and then the same for kappa distributions. How do you want to organize this? 

One way would be to combine functions with the same dimension and then add a method flag to select between speed or velocity distribution. I think combining function of different dimensionality will be a bit trickier. Might need a tuple or array interface for the velocity arguments.

StanczakDominik:
I think it'd make sense to have something like 

```python
class Distribution: # possibly an abstract base class
# define some basic methods like Sample1DSpeed, Sample1DVelocity, Sample3DSpeed, Sample3DVelocity

class MaxwellianDistribution(Distribution)
# define init, redefine methods if necessary
class KappaDistribution(Distribution):
# same thing
```
I'm up for discussing the dimensional issue, there's probably a better solution in there.

lemmatum:
Is there a way to use classes such that you don't have to go through the "constructor"? Otherwise this just seems like extra hoops for a user to jump through.

Function interfaces are nice because if you have a decent IDE then it'll pop-up and tell you what are the positional and optional arguments. It also means the arguments are explicit.

That being said, maybe we can make a function called `maxwellian_distribution()` that does something like this:
```Python
def maxwellian_distribution(temperature, velocities, drift_velocities, method="velocity"):
    if method == "velocity":
        if len(velocities) == 1:
            _maxwellian_distribution_1D(temperature, velocity[0], drift_velocities[0])
        elif len(velocities) == 3:
            _maxwellian_distribution_3D(temperature, velocity[0], velocity[1], velocity[2], etc.)
        else:
            raise Exception(f"No method for {len(velocities)} dimensional distribution. Try 1D or 3D.")
    elif method == "speed":
            "Do same stuff as above"
    else:
            raise Exception(f"No such method: {method} found.")
```

[...]

PR #452, opened 2018-05-09
Maxwellian 3D speed fixes

### LINKED ISSUE #392: Add correct 3D Maxwellian formulae to docs
Currently the 3D Maxwellian docstrings have the formulae for 1D distributions; these should be replaced by the correct ones for 3D distributions:

http://plasmapy.readthedocs.io/en/master/api/plasmapy.physics.distribution.Maxwellian_velocity_3D.html#plasmapy.physics.distribution.Maxwellian_velocity_3D
http://plasmapy.readthedocs.io/en/master/api/plasmapy.physics.distribution.Maxwellian_speed_3D.html#plasmapy.physics.distribution.Maxwellian_speed_3D

lemmatum:
Supersedes #451 
Closes #392 

Our speed distributions are all wrong. The 1D Maxwellian speed distribution is actually the 3D distribution, so that is being renamed. I'm also adding the correct form for the 1D distribution and adding 2D distributions.

### DIFF
--- a/plasmapy/physics/distribution.py
+++ b/plasmapy/physics/distribution.py
@@ -9,14 +9,14 @@ from plasmapy.physics import parameters
 import numpy as np
 from scipy.special import gamma
 
-__all__ = [
-    "Maxwellian_1D",
-    "Maxwellian_velocity_3D",
-    "Maxwellian_speed_1D",
-    "Maxwellian_speed_3D",
-    "kappa_velocity_1D",
-    "kappa_velocity_3D",
-]
+__all__ = ["Maxwellian_1D",
+           "Maxwellian_velocity_2D",
+           "Maxwellian_velocity_3D",
+           "Maxwellian_speed_1D",
+           "Maxwellian_speed_2D",
+           "Maxwellian_speed_3D",
+           "kappa_velocity_1D",
+           "kappa_velocity_3D"]
 
 def _v_drift_units(v_drift):
     # Helper method to assign units to  v_drift if it takes a default value
@@ -30,7 +30,7 @@ def _v_drift_units(v_drift):
 def Maxwellian_1D(v,
                   T,
                   particle="e",
-                  V_drift=0,
+                  v_drift=0,
                   vTh=np.nan,
                   units="units"):
     r"""
@@ -53,7 +53,7 @@ def Maxwellian_1D(v,
         ``'D+'`` for deuterium, or ``'He-4 +1'`` for :math:`He_4^{+1}`
         (singly ionized helium-4)), which defaults to electrons.
 
-    V_drift: ~astropy.units.Quantity, optional
+    v_drift: ~astropy.units.Quantity, optional
         The drift velocity in units convertible to m/s.
 
     vTh: ~astropy.units.Quantity, optional
@@ -103,7 +103,7 @@ def Maxwellian_1D(v,
     >>> from plasmapy.physics import Maxwellian_1D
     >>> from astropy import units as u
     >>> v=1*u.m/u.s
-    >>> Maxwellian_1D(v=v, T= 30000*u.K, particle='e',V_drift=0*u.m/u.s)
+    >>> Maxwellian_1D(v=v, T= 30000*u.K, particle='e',v_drift=0*u.m/u.s)
     <Quantity 5.91632969e-07 s / m>
     """
 
@@ -113,7 +113,7 @@ def Maxwellian_1D(v,
         v = v.to(u.m / u.s)
         # Catching case where drift velocities have default values, they
         # need to be assigned units
-        V_drift = _v_drift_units(V_drift)
+        v_drift = _v_drift_units(v_drift)
         # convert temperature to Kelvins
         T = T.to(u.K, equivalencies=u.temperature_energy())
         if np.isnan(vTh):
@@ -132,7 +132,7 @@ def Maxwellian_1D(v,
     # Get thermal velocity squared
     vThSq = vTh ** 2
     # Get square of relative particle velocity
-    vSq = (v - V_drift) ** 2
+    vSq = (v - v_drift) ** 2
     # calculating distribution function
     coeff = (vThSq * np.pi) ** (-1 / 2)
     expTerm = np.exp(-vSq / vThSq)
@@ -147,8 +147,8 @@ def Maxwellian_velocity_2D(vx,
                            vy,
                            T,
                            particle="e",
-                           Vx_drift=0,
-                           Vy_drift=0,
+                           vx_drift=0,
+                           vy_drift=0,
                            vTh=np.nan,
                            units="units"):
     r"""
@@ -175,10 +175,10 @@ def Maxwellian_velocity_2D(vx,
         ``'D+'`` for deuterium, or ``'He-4 +1'`` for :math:`He_4^{+1}`
         (singly ionized helium-4)), which defaults to electrons.
 
-    Vx_drift: ~astropy.units.Quantity, optional
+    vx_drift: ~astropy.units.Quantity, optional
         The drift velocity in x-direction units convertible to m/s.
 
-    Vy_drift: ~astropy.units.Quantity, optional
+    vy_drift: ~astropy.units.Quantity, optional
         The drift velocity in y-direction units convertible to m/s.
 
     vTh: ~astropy.units.Quantity, optional
@@ -237,8 +237,8 @@ def Maxwellian_velocity_2D(vx,
     ... vy=v,
     ... T=30000*u.K,
     ... particle='e',
-    ... Vx_drift=0*u.m/u.s,
-    ... Vy_drift=0*u.m/u.s)
+    ... vx_drift=0*u.m/u.s,
+    ... vy_drift=0*u.m/u.s)
     <Quantity 3.5002957e-13 s2 / m2>
 
 
@@ -250,8 +250,8 @@ def Maxwellian_velocity_2D(vx,
         vy = vy.to(u.m / u.s)
         # catching case where drift velocities have default values, they
         # need to be assigned units
-        Vx_drift = _v_drift_units(Vx_drift)
-        Vy_drift = _v_drift_units(Vy_drift)
+        vx_drift = _v_drift_units(vx_drift)
+        vy_drift = _v_drift_units(vy_drift)
         # convert temperature to Kelvins
         T = T.to(u.K, equivalencies=u.temperature_energy())
         if np.isnan(vTh):
@@ -270,7 +270,7 @@ def Maxwellian_velocity_2D(vx,
     # accounting for thermal velocity in 3D
     vThSq = vTh ** 2
     # Get square of relative particle velocity
-    vSq = ((vx - Vx_drift) ** 2 + (vy - Vy_drift) ** 2)
+    vSq = ((vx - vx_drift) ** 2 + (vy - vy_drift) ** 2)
     # calculating distribution function
     coeff = (vThSq * np.pi) ** (-1)
     expTerm = np.exp(-vSq / vThSq)
@@ -286,9 +286,9 @@ def Maxwellian_velocity_3D(vx,
                            vz,
                            T,
                            particle="e",
-                           Vx_drift=0,
-                           Vy_drift=0,
-                           Vz_drift=0,
+                           vx_drift=0,
+                           vy_drift=0,
+                           vz_drift=0,
                            vTh=np.nan,
                            units="units"):
     r"""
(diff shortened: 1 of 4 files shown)


lemmatum:
Here is an outline of what I'm doing
[maxwellian_speed_3D.pdf](https://github.com/PlasmaPy/PlasmaPy/files/1989773/maxwellian_speed_3D.pdf)

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
...
+def Maxwellian_velocity_2D(vx,
+                           vy,
+                           T,
+                           particle="e",
+                           Vx_drift=0,
+                           Vy_drift=0,
+                           vTh=np.nan,
+                           units="units"):
+    r"""
+    pdf of velocity for a 2D Maxwellian distribution.
```
namurphy: Perhaps spell out pdf as probability function, since not everyone will know the acronym.

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
...
+
+    Return the probability of finding a particle with velocity components
```
dstansby: Technically this returns the probability density function, not the probability (to get a probability you have to multiply by `dv_x . dv_y`).

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
...
+
+    T: ~astropy.units.Quantity
+        The temperature, preferably in Kelvin.
+
+    particle: str, optional
+        Representation of the particle species (e.g., ``'p'`` for protons,
+        ``'D+'`` for deuterium, or ``'He-4 +1'`` for :math:`He_4^{+1}`
+        (singly ionized helium-4)), which defaults to electrons.
+
+    Vx_drift: ~astropy.units.Quantity, optional
```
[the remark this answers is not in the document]
dstansby: Having lower case parameter names is a python convention, so I'd vote for `vx_drift` (but sometimes it's best to ignore that for e.g. magnetic field as `B`)

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
         return distFunc
 
 
+def Maxwellian_velocity_2D(vx,
```
dstansby: This doesn't seem to be appearing in the built docs: https://1504-46810954-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/physics/distribution.html

plasmapy/physics/distribution.py
dstansby: Re-phrasing as "pdf of velocity for a Maxwellian distribution in 2D" would make it clearer that this is specifically for things in two dimensions.

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
...
+    units: str, optional
+        Selects whether to run function with units and unit checks (when
+        equal to "units") or to run as unitless (when equal to "unitless").
+        The unitless version is substantially faster for intensive
+        computations.
+
+    Returns
+    -------
+    f : ~astropy.units.Quantity
+        Probability in Velocity^-1, normalized so that
```
dstansby: This should be "Probability density"

plasmapy/physics/distribution.py
```diff
@@ -143,6 +143,144 @@ def Maxwellian_1D(v,
...
+                                            method="most_probable"))
+        elif not np.isnan(vTh):
+            # check units of thermal velocity
+            vTh = vTh.to(u.m / u.s)
+    elif np.isnan(vTh) and units == "unitless":
+        # assuming unitless temperature is in Kelvins
+        vTh = parameters.thermal_speed(T * u.K,
+                                       particle=particle,
+                                       method="most_probable").si.value
+    # accounting for thermal velocity in 3D
```
dstansby: "3D" > "2D"

plasmapy/physics/distribution.py
```diff
@@ -370,7 +508,129 @@ def Maxwellian_speed_1D(v,
...
+    Notes
+    -----
+    In 2D, the Maxwellian speed distribution function describing
+    the distribution of particles with speed :math:`v` in a plasma with
+    temperature :math:`T` is given by:
+
+    .. math::
+
+       f = 2 \pi \vec{v} (\pi v_{Th}^2)^{-1} \exp(-(\vec{v} -
+       \vec{V}_{drift})^2 / v_{Th}^2)
```
dstansby: I don't think this is correct; when the drift speed is not zero, it is non-trivial to integrate over angles to get a probability density function for speed; see my working here: https://www.dropbox.com/s/rsjb65nvl2v9kbh/max_2d.pdf?dl=0

namurphy:
A longer term note...should we perhaps create a `ParticleDistributionFunction` class, with `MaxwellianDistribution`, `KappaDistribution`, and the relativistic equivalent as subclasses?  We could specify the dimension, temperature, particle, and other parameters and then be able to calculate a bunch of different values from each of them.  I think it would be best to merge this pull request before attempting that, so if we think this is a worthwhile idea, then we could create an issue on it.

StanczakDominik:
@namurphy that's basically the point of #387 ;)

namurphy:
Oh!  Cool!  You're way ahead of me then. :smiley_cat:

plasmapy/physics/distribution.py
lemmatum: Doesn't the drift velocity just shift the position of distribution in velocity space? (thinking in Cartesian coordinates). Just thinking about it spatially/geometrically, this shouldn't affect the normalization coefficient.

Thinking about it numerically, if you want to get the normalization coefficient you can just make the substitution `v_new = v_old - v_drift`. Since `v_drift` is a constant we have `dv_new = dv_old` and we can just do the integration as we would in the case of `v_drift = 0`.

The one caveat here is that I think it modifies the velocity terms in the normalization coefficient. So for a 3D distribution we have `4 * pi * v ** 2` representing the spherical shell over which we are integrating. I think this shell gets shifted to be centered around the drifting Maxwellian, so it would be something like  `4 * pi * (v - v_drift) ** 2`.

plasmapy/physics/distribution.py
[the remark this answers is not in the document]
lemmatum: Made them lowercase.

plasmapy/physics/distribution.py
lemmatum: done

plasmapy/physics/distribution.py
lemmatum: fixed

plasmapy/physics/distribution.py
lemmatum: done throughout

plasmapy/physics/distribution.py
lemmatum: done

plasmapy/physics/distribution.py
lemmatum: Do I have to manually add stuff in the docs?

lemmatum: Looks like it's showing up now
https://1508-46810954-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/physics/distribution.html?highlight=distribution#module-plasmapy.physics.distribution

lemmatum:
So the only thing left is to check whether the normalization coefficient for speed distributions is correct when considering non-zero drift.

plasmapy/physics/distribution.py
```diff
@@ -9,14 +9,14 @@
...
-    "kappa_velocity_3D",
-]
+__all__ = ["Maxwellian_1D",
+           "Maxwellian_velocity_2D",
+           "Maxwellian_velocity_3D",
+           "Maxwellian_speed_1D",
+           "Maxwellian_speed_2D",
+           "Maxwellian_speed_3D",
+           "kappa_velocity_1D",
+           "kappa_velocity_3D"]
```
StanczakDominik: The reason I like having `]` in its own line and having every entry in a list like this ending with a comma is that when you add an entry, that shows in git as modifying one line instead of modifying 2-3.

StanczakDominik:
That last part sounds like a good unit test!

lemmatum:
@StanczakDominik I'm not sure how to unit test that. I think we first need to work through the theory math before designing a unit test. Thoughts?

plasmapy/physics/distribution.py
lemmatum: Ah, thanks for the tip. I'll keep that in mind for future modifications.

StanczakDominik:
Oh... Yeah, sorry, I skimmed the earlier discussion and thought you had the generic drift velocity case already worked out, just not tested. Yeah, the math comes first ofc.

plasmapy/physics/distribution.py
dstansby: It doesn't affect the normalisation coefficient, but it does effect the probability distribution function (pdf) of the radial velocity coordinate (ie speed) in a polar coordinate system - see my above derivation. Let me know if anything doesn't make sense, but I'm pretty sure that with a drift velocity the pdf of speed currently in the code is wrong.

lemmatum: Yea I've taken a look at the derivation and I'm not convinced. What you argue in the derivation would imply an effect on the normalization coefficient (since the exponentials cannot be affected because they're exponentials), otherwise nothing is affected and the current implementation is correct.

I'll try to work this out on paper using change of variables.

dstansby: My point is that doing the integral in my equation (6) isn't possible (or at least doesn't come out with the equation currently in the docstring) for v_drift != 0.

If v_drift = 0 the integral is possible, and results in the equation in the docstring (with v_drift = 0).

dstansby: I've added a bit at the end with v_drift = 0, to show how it reduces to the expected equation which might help?

lemmatum: So here is my derivation for the current implementation of the velocity and speed distribution including drift. This is using a change of variables to shift the velocity distribution to the origin, do the integration, and then shift back the normalized distribution(s).

img_20180511_024812

dstansby: I don't think the final equation can be correct, as it goes negative for v > v_drift (and probability densities can't be negative!). I think the problem is that the final "shifting back" stage is not valid - shifting in velocity space is not the same as shifting in speed (radial coordinate) space.

dstansby: Yeah, in that last step, you have to take into account the differential bit when changing coordinates, ie. 
image
 and I think you have assumed that dv/dv' = 1, which isn't true.

lemmatum: hmmm, yea it goes negative for `v < v_drift`... perhaps the normalization constant is just `v` then, since the size of the area of integration doesn't change due to drift offsetting the distribution in velocity space.

Good point about shifting in velocity vs in speed space. There must be a way to convert the shift in velocity space to one in speed space. I think that will involve an absolute value since `v = |\vec{v}|`... so maybe the `v - v_drift` part of the normalization coefficient is wrapped in an absolute value. 

I'm also digging into your approach and seeing if I can crack it with some trig identities. If you don't assume drift velocity along one axis, and instead keep it general, you end up with a fairly compact form for the velocity terms:
`v_r ** 2 - 2 * v_r * v_drift_r * cos(v_phi - v_drift_phi) + v_drift_r **2`

You could then integrate each of the 3 terms separately.

lemmatum: I would like to note, it's astounding how I can't find a single page on the internet that discusses this problem :sob:

dstansby: Yep, I agree that in theory you might be able to integrate those three terms (you only actually need to integrate the middle one, since that's the only one that depends on v_phi. It's possible, and you get a modified Bessel function: https://www.wolframalpha.com/input/?i=integrate+e%5E(-a*cos(x))+between+0+and+2pi

dstansby: So my conclusion is that the function and docstring equation should have a modified Bessel function in it somewhere.

lemmatum: So I'm working this out and am having trouble getting the normalization constant in the case of non-zero drift. That damn Bessel function has a radial velocity and so it shows up in the radial integral from 0 to Inf. [Wolfram doesn't know what to do with it](https://www.wolframalpha.com/input/?i=integrate+x+*+e%5E(-x%5E2+%2F+a%5E2)+*+I_0(b+*+x)+between+0+and+inf)

lemmatum: It looks like the 0th modified Bessel function of the first kind has a [nice series expansion](https://reference.wolfram.com/language/ref/BesselI.html), this would then yield a series of integrals which are just different moments of `v`. Then perhaps there is a way to return the function to some sensible form, maybe another Bessel function? Just some ponderings...

lemmatum: Ooh, I found the first mention of Bessel functions and probability distribution functions! Apparently the p.d.f. [follows a Ricean probability density function](http://elec424.wikia.com/wiki/Modified_Bessel_Functions). Funny how once you find the right jargon you start finding all sorts of [links](https://en.wikipedia.org/wiki/Rice_distribution).

lemmatum: Might be able to use [SciPy's Rice p.d.f. function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rice.html).

lemmatum:
I'm currently thinking of adding a line to the 2D and 3D speed distributions which raises an exception for non-zero drift velocity and mentions that this feature is a work in progress. Then we can merge this PR and deal with Rician (and whatever the 3D equivalent is) distributions in a separate PR.

namurphy:
@lemmatum - Yeah, raising a `NotImplementedError` for the time being and addressing this issue in a separate PR sounds like a good idea.  Again, my thanks to everyone who is working on this!

lemmatum:
@namurphy I've added `NotImplementedError` where relevant and modified the pytests for non-zero drift speed. I've also opened issue #462 for follow-up on this. 
Going to merge this PR now.

Issue #167, opened 2017-11-13
Use OpenPMD standard for output data


jlvay:
There is a new standard for storing particle and field data that is being adopted by a growing number of Particle-In-Cell codes in the U.S. and Europe. The name of the standard is OpenPMD and its github repo is at OpenPMD standard (https://github.com/openPMD). It can be used with HDF5 or other self-describing data formats and general purpose writer/reader are in development in Python, C++ and FORTRAN. It may be a good standard to adopt for PlasmaPy for I/O. Also, the standard can be extended to accommodates needs that may not be covered yet. It also enables extensions for more specialized subtopics.

namurphy:
This sounds like a great idea to me!  I am strongly in favor of adopting open standards as much as we can.  Many thanks to @jlvay for proposing this.  

A few of the things I am wondering about are:

- How compatible would this standard be with our adoption of [astropy.units](http://docs.astropy.org/en/stable/units/)?
- Are there any other standards for representing PIC simulation results?  As far as I know, there aren't any other detailed specifications like the [OpenPMD standard](https://github.com/openPMD/openPMD-standard) that are in wide use, though I don't do PIC simulations.  We definitely want functionality for the open [PICKSC](https://picksc.idre.ucla.edu/) codes, and to coordinate with them too.
- How can we best avoid duplication of functionality?  (e.g., common methods for atomic data, common classes for representing different species, etc.)
- Is there a similar standard for representing simulations of plasmas using fluid models?  We would probably need a way to compare fluid and PIC models directly, since that is often done by several of us who study magnetic reconnection.
- Can we also work with [yt](http://yt-project.org/) after their [demeshening](https://www.youtube.com/watch?v=pkZgQIGac6I)?  I did notice that there is an [openPMD-yt](https://github.com/openPMD/openPMD-yt/tree/dev-1.0.0) repository.

@PlasmaPy/plasmapy-community 

Possibly related to: https://github.com/PlasmaPy/PlasmaPy/issues/87, https://github.com/PlasmaPy/PlasmaPy/pull/12, https://github.com/PlasmaPy/PlasmaPy-PLEPs/issues/6, and possibly https://github.com/PlasmaPy/PlasmaPy/issues/52

RemiLehe:
Hi @namurphy,

Thanks for your interest in openPMD. Here are a few answers to your questions:

[...]

PR #459, opened 2018-05-16
[WIP] Basic plasma factory implementation

ritiek:
We can now instantiate the expected plasma subclass based on input keyword arguments.

```python
>>> import astropy.units as u
>>> import numpy as np
>>> import plasmapy.classes

>>> T_e = 25 * 15e3 * u.K
>>> n_e = 1e26 * u.cm ** -3
>>> Z = 2.0
>>> particle = 'p'
>>> blob = plasmapy.classes.Plasma(T_e=T_e,
...                                n_e=n_e,
...                                Z=Z,
...                                particle=particle)

>>> type(blob)
plasmapy.classes.sources.plasmablob.PlasmaBlob

>>> three_dims = plasmapy.classes.Plasma(domain_x=np.linspace(0, 1, 3) * u.m,
...                                      domain_y=np.linspace(0, 1, 3) * u.m,
...                                      domain_z=np.linspace(0, 1, 3) * u.m)

>>> type(three_dims)
plasmapy.classes.sources.plasma3d.Plasma3D
```

I am not sure if relying solely on keyword arguments is good enough to assume the plasma subclass but I don't have a better idea either (SunPy uses different parameters for `data` and `header`, I don't know how well will it work in our case).

We also need to add common plasma data to `GenericPlasma` and as of now `PlasmaFactory` does nothing but just inherits from `BasicRegistrationFactory`.

### DIFF
--- a/plasmapy/classes/plasma_base.py
+++ b/plasmapy/classes/plasma_base.py
@@ -1,8 +1,5 @@
-from collections import OrderedDict
 from abc import ABC
 
-# GenericPlasma subclass registry
-PLASMA_CLASSES = OrderedDict()
 
 class GenericPlasmaRegistrar(ABC):
     """
@@ -12,7 +9,8 @@ class GenericPlasmaRegistrar(ABC):
     when a subclass of `GenericPlasma` is defined. If it exists it will add that
     class to the registry.
     """
-    _registry = PLASMA_CLASSES
+    # GenericPlasma subclass registry
+    _registry = dict()
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
@@ -22,5 +20,4 @@ class GenericPlasmaRegistrar(ABC):
 
 class GenericPlasma(GenericPlasmaRegistrar):
     def __init__(self, **kwargs):
-        for key, value in kwargs.items():
-            setattr(self, key, value)
+        pass
--- a/plasmapy/classes/plasma_base.py
+++ b/plasmapy/classes/plasma_base.py
@@ -1,9 +1,10 @@
-from abc import ABC
+from abc import ABC, abstractmethod, abstractproperty
 
 
-class GenericPlasmaRegistrar(ABC):
+class BasePlasma(ABC):
     """
-    Registration class for `~plasmapy.classes.GenericPlasma`.
+    Registration class for `~plasmapy.classes.GenericPlasma` and declares
+    some abstract methods for data common in different kinds of plasmas.
 
     This class checks for the existance of a method named ``is_datasource_for``
     when a subclass of `GenericPlasma` is defined. If it exists it will add that
@@ -17,7 +18,55 @@ class GenericPlasmaRegistrar(ABC):
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
 
+    # This class is supposed to declare abstract methods (@abstractmethod or
+    # @abstractproperty as appropriate) that are common in most plasmas
+    # (like `electronTemperature`, `ionTemperature`,
+    # `electronDensitry`, `ionDensity`, `averageIonization`, etc.)
+    # where as `GenericPlasma` class will hold the definitions for these
+    # abstract methods.
 
-class GenericPlasma(GenericPlasmaRegistrar):
+    # For reference, see
+    # https://github.com/sunpy/ndcube/blob/master/ndcube/ndcube.py#L26
+
+    @abstractproperty
+    def electron_temperature(self):
+        pass
+
+    @abstractproperty
+    def ion_temperature(self):
+        pass
+
+    @abstractproperty
+    def electron_density(self):
+        pass
+
+    @abstractproperty
+    def ion_density(self):
+        pass
+
+    @abstractproperty
+    def average_ionization(self):
+        pass
+
+
+class GenericPlasma(BasePlasma):
     def __init__(self, **kwargs):
+        pass
+
+    # The definitions for the abstract methods declared in `BasePlasma`
+    # goes here.
+
+    def electron_temperature(self):
+        pass
+
+    def ion_temperature(self):
+        pass
+
+    def electron_density(self):
+        pass
+
+    def ion_density(self):
+        pass
+
+    def average_ionization(self):
         pass
--- a/plasmapy/classes/plasma_base.py
+++ b/plasmapy/classes/plasma_base.py
@@ -1,0 +1,26 @@
+from collections import OrderedDict
+from abc import ABC
+
+# GenericPlasma subclass registry
+PLASMA_CLASSES = OrderedDict()
+
+class GenericPlasmaRegistrar(ABC):
+    """
+    Registration class for `~plasmapy.classes.GenericPlasma`.
+
+    This class checks for the existance of a method named ``is_datasource_for``
+    when a subclass of `GenericPlasma` is defined. If it exists it will add that
+    class to the registry.
+    """
+    _registry = PLASMA_CLASSES
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if hasattr(cls, 'is_datasource_for'):
+            cls._registry[cls] = cls.is_datasource_for
+
+
+class GenericPlasma(GenericPlasmaRegistrar):
+    def __init__(self, **kwargs):
+        for key, value in kwargs.items():
+            setattr(self, key, value)
--- a/plasmapy/classes/plasma_base.py
+++ b/plasmapy/classes/plasma_base.py
@@ -50,6 +50,10 @@ class BasePlasma(ABC):
 
 
 class GenericPlasma(BasePlasma):
+    """
+    A Generic Plasma class. This class contains definitions for abstract
+    methods declared in the `~plasmapy.classes.plasma_base.BaseClass`.
+    """
     def __init__(self, **kwargs):
         pass
 
--- a/plasmapy/classes/plasma_base.py
+++ b/plasmapy/classes/plasma_base.py
@@ -1,5 +1,9 @@
 from abc import ABC, abstractmethod, abstractproperty
 
+__all__ = [
+    "BasePlasma",
+    "GenericPlasma",
+]
 
 class BasePlasma(ABC):
     """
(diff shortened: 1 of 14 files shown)


ritiek:
So, how should we go about implementing `GenericPlasma`? What common plasma data would it deal with?

lemmatum:
If `GenericPlasma` is supposed to be a particular instance of a plasma I would rename it to `ClassicalPlasma` and make it deal with the high temperature, low density regime (weak coupling, sufficient particles in the Debye sphere, etc.). If instead `GenericPlasma` is going to be used to create various subclasses and it is just a container for shared properties between different kinds of plasmas then I would include things like `electronTemperature`, `ionTemperature`, `electronDensity`, `ionDensity`, `averageIonization`, etc.

Thoughts @namurphy @StanczakDominik ?

namurphy:
The output from an magnetohydrodynamic simulation will generally include the magnetic field vector **B**, the velocity vector **V**, the current density vector **J**, the number density of particles per unit volume *n* (or sometimes the mass density *ρ*), the plasma pressure *p*, and the temperature *T*.  Probably each individual data file will have information on a single time, and there will usually be multiple files.  The simulations will sometimes be 2D and sometimes be 3D, and the grid positions will need to be recorded too.  Sometimes the data will be in rectangular coordinates and other times it'll be in cylindrical coordinates, or even another coordinate system.  Occasionally there will be other quantities that need to be included, like the magnetic vector potential **A** (which will reduce to a scalar in 2D).  Simulation output will sometimes have units (and thus will need to be `astropy.units.Quantity` instances) and other times will be dimensionless (and thus will end up being `numpy.ndarray` instances).   

Perhaps a good thing to do for developing the general framework would be to create your own data to try to read in and set up the appropriate subclass.  It would be possible to create some mock HDF5 files using `h5py` that contain plasma data on a 2D or 3D grid, where the names of the variables in the HDF5 files correspond to the above variables.  The `Plasma` class could then select the appropriate subclass from maybe `Plasma2D` and `Plasma3D` based on the information contained within the files.  

Another possibility would be to create a class to represent data sets that use the [OpenPMD](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md) standard (see also #167 and https://github.com/PlasmaPy/PlasmaPy-PLEPs/issues/13), and make it so that `Plasma` is able to select that appropriate subclass.

ritiek:
I guess I'll mess around with some HDF5 files for a bit then.

plasmapy/classes/plasma_base.py
```diff
@@ -0,0 +1,26 @@
+from collections import OrderedDict
+from abc import ABC
+
+# GenericPlasma subclass registry
+PLASMA_CLASSES = OrderedDict()
```
ritiek: Do we really need to use an `OrderedDict` here or a simple `dict` object would suffice? SunPy seems to use `OrderedDict` but I am not sure about its usefulness (plus it is slower).

Cadair: Given you are Python 3.6+ you do not need `OrderedDict` here. Also I don't think it really matters at all.

plasmapy/classes/plasma_base.py
```diff
@@ -0,0 +1,26 @@
...
+
+class GenericPlasmaRegistrar(ABC):
```
Cadair: I am assuming that you are planning on using this class for API definition and not just to be the keeper of the registry, so I would suggest you rename it to `BasePlasma` or something similar. As this is the class you will want to do `isinstance` with to test if it implements your plasma API

plasmapy/classes/plasma_base.py
Cadair: I know this is how it's done in SunPy, but actually I would not have this as a module level constant. I would just have it as a class attribute on `GenericPlasmaRegistrar`

plasmapy/classes/plasma_factory.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Plasma factory class. Used to create a variety of Map objects. Valid plasma
+    structures are specified by registering them with the factory.
+    """
+    pass
+
+
+Plasma = PlasmaFactory(default_widget_type=GenericPlasma,
+                       additional_validation_functions=['is_datasource_for'])
+Plasma.registry = PLASMA_CLASSES
```
Cadair: Given my comment above, this would become `GenericPlasma._registry` (as it would be inherited from the registrar)

plasmapy/classes/plasma_base.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if hasattr(cls, 'is_datasource_for'):
+            cls._registry[cls] = cls.is_datasource_for
+
+
+class GenericPlasma(GenericPlasmaRegistrar):
+    def __init__(self, **kwargs):
+        for key, value in kwargs.items():
+            setattr(self, key, value)
```
Cadair: I don't know if you have just implemented this as a stop-gap testing thing or not, but I would recommend not doing this in the long term.

plasmapy/utils/datatype_factory_base.py
```diff
@@ -0,0 +1,191 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Project     : https://github.com/sunpy/sunpy
+# File        : sunpy/util/datatype_factory_base.py
+# Commit hash : f6330eea602ea796b5b004dee283b8877b24da23
+
+
+import inspect
+
+class BasicRegistrationFactory(object):
```
Cadair: oh the copy paste horror of it all :stuck_out_tongue_closed_eyes: 

you don't need `(object)` here

plasmapy/classes/plasma_factory.py
Cadair: Actually pass `registry=GenericPlasma._registry` to the `PlasmaFactory` constructor.

plasmapy/classes/plasma_base.py
ritiek: Ok, I guess then we should probably stick with `dict` atleast for now.

plasmapy/classes/plasma_base.py
ritiek: I don't know for sure if we are going to use this class for API definition or not.

Based on my understanding, `GenericPlasma` is supposed to deal with data that is common in all plasmas. If we are going to have `GenericPlasmaRegistrar` do this task as well then we could remove `GenericPlasma` class and have this class renamed to `BasePlasma` and deal with both common plasma data and do the registration thing.

Thoughts @SolarDrew, @namurphy?

plasmapy/classes/plasma_base.py
ritiek: Yep, good idea.

plasmapy/classes/plasma_factory.py
ritiek: That's nicer!

plasmapy/classes/plasma_base.py
Cadair: It's a good idea to have an Abstract base class that dosen't implement any (much) API and then a base class which implements the abstract class in a way which is generic for plasmapy. This means that if anyone wants to do something *very* different down the line they can have all the benifits of the ABC and the registration without having to worry about any of the implementation in the generic class.

For an example of this see [ndcube](https://github.com/sunpy/ndcube/blob/master/ndcube/ndcube.py#L26)

ritiek: I think I understand now, that ndcube example is very interesting. So, we do both the registration thing and *declare* some abstract methods on the `BasePlasma` but define them in `GenericPlasma`. I'm good with that.

SolarDrew:
> Simulation output will sometimes have units (and thus will need to be astropy.units.Quantity instances) and other times will be dimensionless (and thus will end up being numpy.ndarray instances).

I disagree with the last point - I think if values are dimensionless they should use the dimensionless unit, both for consistency and to avoid ambiguity.

ritiek:
Ah, yes. Astropy does have a dimensionless unit.
```python
>>> u.m/u.m == u.dimensionless_unscaled
True
```

SolarDrew:
Looks good to me so far :+1: Does anyone know what's up with the circleci failure? Is that the sphinx problem we've been having for a while?

ritiek:
> what's up with the circleci failure?

It probably has something to do with [docs/plasma/index.rst](https://github.com/ritiek/PlasmaPy/blob/0cb66386295c51375790a7c486abe837497259bd/docs/plasma/index.rst). I'll see if I can fix it.

ritiek:
The coverage drop is because [plasma_base.py](https://github.com/ritiek/PlasmaPy/blob/485d32e71996194fd5bf7bddb2535b5753b23572/plasmapy/classes/plasma_base.py) has [some uncovered methods](https://coveralls.io/builds/17143819/source?filename=plasmapy%2Fclasses%2Fplasma_base.py). Everything else coverage related to this PR seems fine to me.

SolarDrew:
Sounds reasonable to me. Any final opinions before I merge this then, @namurphy @StanczakDominik @lemmatum?

StanczakDominik:
Actually... Could you please hold off for now so we can finish 0.1.1 up on master? That'd also give me an opportunity to run a review here. I should be able to get this reviewed by today evening, barring unforeseen Linux errors.

SolarDrew:
No problem @StanczakDominik , just let us know when you're ready. I think there is a workflow for keeping bugpatch changes separate from changes going into new releases though, but I can't remember what that is. I'll make an issue, since it's something that will probably come up again.

StanczakDominik:
Sure, that workflow is just not forgetting to merge bug fixes into stable, which I forgot to do :D

SolarDrew:
Lol, sure, but I did think there was more logistical wizardry to it than that :P Let's move further conversation here: #474 :)

StanczakDominik:
Okay, reviewing this now because Linux decided to crap out on me again... let's just merge into master and I'll sort out the stable/unstable thing later.

ritiek:
By the way, do we need an `__all__` list for [plasma_base.py](https://github.com/ritiek/PlasmaPy/blob/9536cdb50f3eede1a7ae50841d333f72a7a58db4/plasmapy/classes/plasma_base.py) and [plasma_factory.py](https://github.com/ritiek/PlasmaPy/blob/9536cdb50f3eede1a7ae50841d333f72a7a58db4/plasmapy/classes/plasma_factory.py)?

StanczakDominik:
If we intend to document this via automodapi, they can't hurt! I'd be in
favor of including them as a sane default.

By the way if you wanna merge this into master go right ahead, I'll do some
git magic on stable to finish 0.1.1 today.

StanczakDominik:
All right, the master branch definitely needs more metaclass now. :laughing:

PR #461, opened 2018-05-18
Corrected an import statement

jhillairet:
Correcting a single import statement which was failing.

### DIFF
--- a/plasmapy/physics/dielectric.py
+++ b/plasmapy/physics/dielectric.py
@@ -1,7 +1,7 @@
 """Functions to calculate plasma dielectric parameters"""
 
 from astropy import units as u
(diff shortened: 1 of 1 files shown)


ritiek:
I am curious to what error it causes for you! I've always thought they meant the same. However, I don't see a problem with this PR being merged in if this change takes care of it.

lemmatum:
I agree with @ritiek , no problems with merging this.
@jhillairet  Can you post a comment including the error message you got with the old import statement, along with python version/distribution?

jhillairet:
Well, I don't remember the error. That was an import error I had under Jupyter notebook, which I discovered after pulling the last revision of PlasmaPy (since few months...)

PR #466, opened 2018-05-19
tried "try .to" approach in _check_relativistic to avoid .si

### LINKED ISSUE #465: .si in check_relativistic is extra slow
Hi all,

I was making some plots with PlasmaPy today.  So my basic plan was, I have some arrays of densities and temperatures and magnetic fields, and I want to make some contour plots of various plasma parameters.  I didn't think plasmapy would be able to do vectorized operation on the numpy arrays, so I said, screw it, I'll just loop through each array and do each one individually.

Well. I knew it would be slow, but it was even slower than I expected. I ran it through the profiler and I think some of it is from Python itself, some of it is from the astropy units stuff, but then there was a massive amount of time in check_relativistic due to the use of the .si function on the astropy units, somehow.  Apparently that function is really slow?  I'm going to see if we can avoid the use of it.

samurai688:
Can we get away with this?  It seemed to work much faster, cut the run time of my whole script by a factor of 3.

Hmm, using timeit shows a speedup from 4.019467848003842 to 1.8043721090070903 .... which I guess is kind of close to a factor of 3.

Closes #465

### DIFF
--- a/plasmapy/utils/checks.py
+++ b/plasmapy/utils/checks.py
@@ -410,14 +410,14 @@ def _check_relativistic(V, funcname, betafrac=0.05):
         raise TypeError(errmsg)
 
     try:
-        V = V.to(u.m / u.s)
+        V_over_c = (V / c).to_value(u.dimensionless_unscaled)
     except Exception:
         raise u.UnitConversionError(errmsg)
 
     if np.any(np.isnan(V.value)):
         raise ValueError(f"V includes NaNs in {funcname}")
 
-    beta = np.max(np.abs((V / c).value))
+    beta = np.max(np.abs((V_over_c)))
 
     if beta == np.inf:
         raise RelativityError(f"{funcname} is yielding an infinite velocity.")
--- a/plasmapy/utils/checks.py
+++ b/plasmapy/utils/checks.py
@@ -409,13 +409,15 @@ def _check_relativistic(V, funcname, betafrac=0.05):
     if not isinstance(V, u.Quantity):
         raise TypeError(errmsg)
 
-    if V.si.unit != u.m / u.s:
+    try:
+        V = V.to(u.m / u.s)
+    except Exception:
         raise u.UnitConversionError(errmsg)
 
     if np.any(np.isnan(V.value)):
         raise ValueError(f"V includes NaNs in {funcname}")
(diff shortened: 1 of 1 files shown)


SolarDrew:
Just left a comment on #465 because I didn't realise you'd already opened a PR :man_facepalming: ...

samurai688:
Whoops, yeah, I think your way is better.  My timeit on that is down to 1.22. :)

Oddly my script doesn't seem to get faster... maybe my computer just has some sort of special philosophical objection to .si. This speed stuff is black magic.

SolarDrew:
Curious. Different numpy or astropy versions, maybe? Either way, this looks good to me so if you're happy with it I'll go ahead and merge.

samurai688:
Could be astropy, I haven't updated that in a while.  Hmm, trying with astropy 3.0.2, timeit check_relativistic keeps the same ~3x speed boost... but script got 30% slower. Thanks astropy!  Haha, I suppose the real solution is to figure how to handle these things with the fast numpy array ufunc deals, or something else, but eh, that's an issue for another day.  If you're fine with this I'm happy with it, it seems to be a nice speed boost for my loop script for whatever reason.

samurai688:
Oh wait no, it's actually a tiny bit faster with the new astropy, probably just some loading of stuff on the first run. I take it back.

SolarDrew:
Weird. Well it's already faster so I call that a win. Thanks for the help @samurai688 :D

plasmapy/utils/checks.py
```diff
@@ -409,13 +409,15 @@ def _check_relativistic(V, funcname, betafrac=0.05):
     if not isinstance(V, u.Quantity):
         raise TypeError(errmsg)
 
-    if V.si.unit != u.m / u.s:
+    try:
+        V_over_c = (V / c).to_value(u.dimensionless_unscaled)
+    except Exception:
```
Cadair: Should this not be `u.UnitsError`?

StanczakDominik: Definitely. I'll get a follow-up fix here.

SolarDrew: Whoops, missed that... Thanks @Cadair

PR #468, opened 2018-05-21
Added Particle class method to return ion symbols in Roman notation.

### LINKED ISSUE #464: Create Particle class method to return Roman numeral spectroscopic notation
Right now the `Particle` class accepts atoms and ions in Roman numeral notation (see #286 and #290 for details), where the Roman numeral after the atomic symbol represents the ionization level plus one:
```Python
>>> Particle("Fe XIII")
Particle("Fe 12+")
```
It would be helpful to have a `Particle` class attribute that would return the Roman numeral notation of `Particle` instances that are ions or neutral atoms.  
```Python
>>> Particle("Fe 12+").ionic_symbol_roman
'Fe XIII'
```
I'm not sure if `ionic_symbol_roman` would be the best method name; just using `roman` would be an alternative.  

This should only work for `Particle` instances where the integer charge is specified in order to be consistent with other parts of the class.  

Thanks!

Elfhelm:
Adds a property to the Particle class to return ion symbols in Roman notation.  The Roman numeral is the charge of the ion plus one.  For example,

```
>>> from plasmapy.atomic import *
>>> proton = Particle('proton')
>>> proton.roman_symbol
'H II'
```

Closes #464.

### DIFF
--- a/plasmapy/atomic/particle_class.py
+++ b/plasmapy/atomic/particle_class.py
@@ -4,6 +4,7 @@ import numpy as np
 import warnings
 from typing import (Union, Set, Tuple, List, Optional)
 import collections
+import roman
 
 import astropy.units as u
 import astropy.constants as const
@@ -607,6 +608,35 @@ class Particle:
 
         """
         return self._attributes['ion']
+
+    @property
+    def roman_symbol(self) -> Optional[str]:
+        """
+        Return the spectral name of the particle (i.e. the ionic symbol in
+        Roman numeral notation).  If the particle is not an ion or neutral
+        atom, return `None`.
+
+        Examples
+        --------
+        >>> proton = Particle('proton')
+        >>> proton.roman_symbol
+        'H II'
+        >>> hydrogen_atom = Particle('H', Z=0)
+        >>> hydrogen_atom.roman_symbol
+        'H I'
+
+        """
+        if not self._attributes['element']:
+            return None
+        if self._attributes['integer charge'] is None:
+            raise ChargeError(f"The charge of particle {self} has not been specified.")
+        if self._attributes['integer charge'] < 0:
+            raise roman.OutOfRangeError('Cannot convert negative charges to Roman.')
+
+        element = self._attributes['element']
+        integer_charge = self._attributes['integer charge']
+        roman_charge = roman.toRoman(integer_charge + 1)
+        return element + ' ' + roman_charge
 
     @property
     def element_name(self) -> str:


plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -192,6 +200,7 @@
       'element_name': 'hydrogen',
       'isotope': 'D',
       'ionic_symbol': 'D 1+',
+      'roman_symbol': 'H II',
```
lemmatum: I'm not sure this is correct... I would've thought this should be `D II` so that it is unambiguous.
@namurphy  thoughts?

plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -192,6 +200,7 @@
       'element_name': 'hydrogen',
```
lemmatum: @namurphy Why is deuterium listed as hydrogen? Is this for convenience? Should we add an isotope name property and spell out deuterium, tritium, etc?

plasmapy/atomic/particle_class.py
```diff
@@ -608,6 +609,35 @@ def ionic_symbol(self) -> Optional[str]:
         """
         return self._attributes['ion']
 
+    @property
+    def roman_symbol(self) -> Optional[str]:
+        """
+        Return the spectral name of the particle (i.e. the ionic symbol in
```
lemmatum: I think the docstring can be expanded a bit. If you look in other parts of the code we explicitly include argument names, acceptable types, a short description of the argument, and likewise for the returned value(s).

plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 from astropy import units as u
 import inspect
+import roman
 
 from ...constants import m_p, m_e, m_n, e
```
lemmatum: Add some tests for the new `roman_symbol` method?

plasmapy/atomic/particle_class.py
namurphy: This docstring is pretty similar in length and in detail to many of the other docstrings for attributes in the `Particle` class, since as an attribute it doesn't take in any arguments.  The only thing I can think of would be to add that the Roman numeral represents one plus the integer charge, and that it'll raise a `ChargeError` if no charge has been specified and a `roman.OutOfRangeError` if the charge is negative.

plasmapy/atomic/tests/test_particle_class.py
namurphy: Yep, I forgot to mention isotopes in the description.  I brought this up above.

plasmapy/atomic/tests/test_particle_class.py
namurphy: According to Coveralls it looks like the tests that @Elfhelm added below cover everything except the `roman.OutOfRangeError` line, which I'm not too worried about.  Might not be a bad idea to sooner or later add a `Particle` class test of a negatively charged ion, like "H 1-" (which is important for opacity in stellar atmospheres for relatively cool stars, surprisingly enough).

plasmapy/atomic/tests/test_particle_class.py
lemmatum: sounds good

plasmapy/atomic/particle_class.py
lemmatum: Thanks for clarifying about the lack of arguments, I totally missed the fact that it was an attribute.

Do we currently have an error check for if the user inputs a charge which is beyond the atomic number for the atom/isotope? If not, it should be included under `ChargeError`.

plasmapy/atomic/tests/test_particle_class.py
lemmatum: Fair enough!

plasmapy/atomic/particle_class.py
namurphy: Yep, if we do something like `Particle('H++')` it will yield an `InvalidParticleError`.  `Particle` also accepts negatively charged ions, but issues an `AtomicWarning` for integer charges less than -3 or so.  Maybe we should do an `InvalidParticleError` for impossibly negatively charged ions as well.

StanczakDominik: Perhaps we should include an `InvalidLanguageError` for `Particle("C++")`.

StanczakDominik:
Hey all, I haven't really been following what's been going on here lately, but, at a glance looks like there's a few more minor fixes and we could merge this soon-ish! Need a hand, @Elfhelm ? :)

ritiek:
@Elfhelm If you're still here any chance you could make those minor fixes? :)

Issue #469, opened 2018-05-22
working with unitized numpy arrays in plasma parameters functions


samurai688:
Life seems pretty good when we can work with unitized numpy arrays in plasma parameter functions. I guess these are technically Quantity objects, i.e. as created by `a = np.array([1,2,3,4]) * u.m` A good current example is the Debye length function.  You can pass it big unitized numpy arrays of densities and temperatures, and it will give you the elementwise result of debye length.  It seems fast for big arrays just like you'd hope from numpy. I haven't tested it too rigorously but it seems to work really well out of the box.

Some of our current functions will fail when given numpy arrays. I'm looking at you, gyroradius, collision_frequency, impact_parameter, and some others.  It seems like most of the problems show up when we have an if statement like 

```
if bPerp > lambdaBroglie:
    bmin = bPerp
else:
    bmin = lambdaBroglie
```

If we try to give it inputs such that `bPerp` arrives as a unitized numpy array, it will give an exception "the truth value of an array is ambiguous, use .any() or .all()" or some such.  A wrinkle of this one in particular, we want to get the maxes into our `bmin` result, which would then be a numpy array. I would typically do `np.max((bPerp, lambdaBroglie), axis=0)` but that seems to ignore the units, and drop them, for good measure.  The comparison `bPerp > lambdaBroglie` seems to work in a nice unit-aware way, so I guess we can work with the boolean array we get from that...

So anyway,
1) do you think we should support these objects throughout our functions?
2) any clever ideas to handle the issues of the type above?
3) general thoughts?

namurphy:
Yes, this is an important issue and we should allow all of our plasma parameters functions to accept and return `astropy.units.Quantity` objects.  We should also make sure that there are tests for each of the plasma parameters functions that take in `Quantity` objects.

Cross-reference for another issue about `Quantity` objects that we should keep in mind: #59

lemmatum:
I'll add that sometimes arrays with units are clunky for things like list comprehensions. Perhaps we should add some convenience functions for handling these cases?

StanczakDominik:
Closed by #477, #480, #485 and #492 and thanks @samurai688 once again! :D

PR #477, opened 2018-05-26
Modify gyroradius function to work with numpy array Quantities

samurai688:
First baby step of progress towards issue #469, this one is focused solely on the gyroradius function.

Gyroradius seemed to have trouble due to its unique feature, in which it will accept a T_i input, or a Vperp input, but not both.  I put in some conditional statements to hopefully extend this behavior to numpy array Quantities.  

The main new behavior that I have tried to test for and achieve is:
- If only Ti is input and is a no-nan-Qarray, function should work (result is a Qarray calculated with Ti)
- If only Vperp is input and is a no-nan-Qarray, function should work (result is a Qarray calculated with Vperp)
- If both Vperp and Ti are input and are both Qarray, but with nans in each array such that for each element there is exactly 1 nan, function should work (result is a Qarray calculated with the non-nans)
- if, for some reason, Vperp is a Qarray of all nan and Ti is valid, function should work. The result I went with was to return a Qarray of len(Vperp) all calculated with the valid scalar Ti, but I'm open to other returns here. Hopefully this case won't get activated much.
- same as above but vice versa with Vperp, Ti

And hopefully raise ValueError on all the various invalid combinations. Of note is
- if Ti is a valid scalar, and Vperp is a Qarray that has some nan but not all, raise ValueError
- same as above but vice versa with Vperp, Ti

This one is debateable, I think doing it this way of raising the ValueError keeps to the spirit of the original "you must provide either Ti or Vperp but not both" and is hopefully more consistent and least confusing, but I could also imagine a way in which the scalar would be used only in the cases where there are the nans in the array.

---

Actually, it looks like gyroradius and hall parameter are the only two plasma parameter functions that currently have issues when given numpy array Quantities.  So, we'd be 50% done!...  Except that hall parameter calls collision_frequency, which opens up the collisions module, which looks to be in a world of pain.

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -881,6 +881,7 @@ def gyroradius(B, particle='e-', *, Vperp=np.nan * u.m / u.s, T_i=np.nan * u.K):
     # the underlying values of the astropy quantity may be numpy arrays
     if np.any(np.logical_not(np.logical_xor(isfinite_Vperp, isfinite_Ti))):
         raise ValueError("Must give Vperp or T_i, but not both, as arguments to gyroradius")
+    
 
     # check 2: get Vperp as the thermal speed if is not already a valid input
     if np.isscalar(Vperp.value) and np.isscalar(T_i.value):  # both T_i and Vperp are scalars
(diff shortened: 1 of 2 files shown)


StanczakDominik:
All right, I think we're good to go here, so I'll merge it! Thanks, @samurai688 !

samurai688:
Haha, yeah, I'll be the first to admit the situation here is not great... hey, if the abstract plasma class helps I'll be the first to sing its praises.  The only stylistic issue on which I'll never yield is electronvolts :)

PR #480, opened 2018-05-30
Handling nparray Quantities part deux: Coulomb_logarithm

samurai688:
Chipping away at issue #469 , this time working on Coulomb_logarithm and the functions it relies on, namely impact_parameter and _boilerplate. 

As before, the goal this time is to make it so that Coulomb_logarithm works with len>1 array Quantities, such as `T_e = np.linspace(1,10,10) * u.eV`.  Let's see how this works.

One thought I had while working on this is the following.  So far I've found two ways to handle both scalar and vector Quantity inputs that need incompatible syntax:
- way 1: 
```
if np.isscalar(T_e.value):
    # handle scalar case
else: 
    # handle array case
```

vs. 
- way 2:
```
try:
    # handle scalar case
except ValueError: 
    # handle array case
```

I've heard a proverb that the "ask forgiveness not permission" is more pythonic, so maybe way 2 is better?  It makes me a little uneasy... maybe just takes some getting used to.  On way 1, it does feels a little clunky to have to get the `.value` for `np.isscalar`, but it doesn't seem to work correctly if you give it the full Quantity object.  Anybody have a preference or know a better / more preferred way?

Let's see how the tests are doing.


StanczakDominik:
I remember reading something about how one of these ways can be slower than the other... we might be able to figure something out here via `%%timeit`.

plasmapy/physics/transport/collisions.py
```diff
@@ -306,11 +309,22 @@ def _boilerPlate(T, particles, V):
     # obtaining reduced mass of 2 particle collision system
     reduced_mass = atomic.reduced_mass(*particles)
 
-    if V == 0:
+    if np.any(V == 0):
         raise utils.exceptions.PhysicsError("You cannot have a collision for zero velocity!")
     # getting thermal velocity of system if no velocity is given
-    if np.isnan(V):
+    if V is None:
```
StanczakDominik: Won't this need a change in the function's signature? We might also then have to add `V: {'none_shall_pass': True}` to the `check_quantity` dictionary.

plasmapy/physics/transport/collisions.py
```diff
@@ -306,11 +309,22 @@ def _boilerPlate(T, particles, V):
...
         V = parameters.thermal_speed(T, mass=reduced_mass)
+    elif np.any(np.isnan(V)):
+        if np.isscalar(V.value) and np.isscalar(T.value):
+            V = parameters.thermal_speed(T, mass=reduced_mass)
+        elif np.isscalar(V.value):
+            V = parameters.thermal_speed(T, mass=reduced_mass)
+        elif np.isscalar(T.value):
+            V = V.copy()
```
StanczakDominik: :+1:

plasmapy/physics/transport/collisions.py
```diff
@@ -592,6 +615,14 @@ def impact_parameter(T,
         bmin = (lambdaBroglie ** 2 + bPerp ** 2) ** (1 / 2)
     else:
         raise ValueError(f"Method {method} not found!")
+    # ARRAY NOTES
+    # it could be that bmin and bmax have different sizes. If Te is a scalar,
+    # T and V will be scalar from _boilerplate, so bmin will scalar.  However
+    # if n_e is an array, than bmax will be an array. if this is the case,
+    # do we want to extend the scalar bmin to equal the length of bmax?
+    # Eh, maybe we should just leave it as a scalar?
```
StanczakDominik: This is a good question that I think deserves some further thought... @namurphy @lemmatum what do you think?

plasmapy/physics/transport/tests/test_collisions.py
```diff
@@ -38,12 +41,77 @@ def setup_class(self):
         self.gms6 = 3.635342040477818
         self.gms6_negative = 0.030720859361047514
 
+    def test_handle_invalid_V(self):
```
StanczakDominik: Yay for class-grouped tests! :smiley:

plasmapy/physics/transport/collisions.py
samurai688: Hmm, good point, I'm not sure.  The current decorator situation seems to allow me pass a None in as the V input, I think I did that in one of the tests. 

My rationale for including this check was to allow for either None or np.nan * unit to indicate "no input specified, use the default behavior". I think there are parts of the code right now where we've used either one or the other as the argument default.

StanczakDominik: Yeah, the `np.nan` thing was a hack to pass in a `None`-ish `Quantity. I'll look at it a bit later.

plasmapy/physics/transport/collisions.py
samurai688: My thought on leaving it scalar at the moment was that, well firstly I guess they're likely to be numpy-operator'ed at each other anyway as log( bmax / bmin) in the coulomb log, but philosophically it would be a hands-off approach, letting the user feel like they're just using numpy and if they get scalars or vectors, it's true to their own inputs.

plasmapy/physics/transport/collisions.py
lemmatum: Weird, previously passing `None` would always fail unit checks from the decorator...

plasmapy/physics/transport/collisions.py
lemmatum: I think there is an implicit assumption here that arrays of temperatures and velocities would be equal in length and produce an array of Coulomb logarithms of equal length. Am I understanding this part correctly?

If so, this should be explicitly documented in the docstring and perhaps add some if statements or try-except blocks to catch when this dimensions don't match.

plasmapy/physics/transport/collisions.py
lemmatum: This is a tricky one... having the same dimensions would make it easier to maintain the code, but there may be valid reasons for wanting multiple `bmin`s and `bmax`s.
If you were to leave it as is, how would you handle the use of `impact_parameter` within `Coulomb_logarithm`?

plasmapy/physics/transport/collisions.py
samurai688: I don't think it should be assumed that T and V are the same length here, this is inside the `np.any(np.isnan(V))`, so V could e.g. be the default scalar nan, while we would still want T to be allowed to be a vector.

samurai688:
Alright, I think this addresses comments so far.  It didn't seem to make any difference in Coulomb_logarithm performance to extend bmin to be always the same length as bmax when needed, so I put that in, I agree as @lemmatum said that's good behavior from the function.

@StanczakDominik  .. btw our rapidly shifting network of test coverage plugins seems to be behaving oddly with this PR.  Coveralls is suddenly mad at me about langmuir.py, which I haven't touched. Is it a rebase-needed thing? Though I guess it's better than Codecov, which gives me a cubist representation of unhappiness, but shuts down attempts to learn more due to lack of a "base report".  :)

StanczakDominik:
Oh yeah, that's definitely a rebase-needed issue! I'll rebase later today and I think it should clear itself up. Don't worry about it! Thanks for making my day with your vivid description of that plot, though :D

StanczakDominik:
Okay, that was funny. I think I messed this up by mistakenly pushing the rebased commits onto a "pr/480" branch on the main repo instead of onto your fork's branch. I think that messed up coverage even further.

I'm pretty sure this is fine, now, though. Going to merge it!

PR #481, opened 2018-05-31
Improve documentation and simplify syntax of check_quantity

StanczakDominik:
This PR kind of grew on me. While reworking the examples, I realized this syntax was dumb:
```python
@utils.check_quantity({
    'T': {'units': u.K, 'can_be_negative': False},
    'n_e': {'units': u.m**-3, 'can_be_negative': False},
})
```

And this syntax looked better:
```python
@utils.check_quantity(
    T={'units': u.K, 'can_be_negative': False},
    n_e={'units': u.m**-3, 'can_be_negative': False},
)
```

So I went ahead and changed everything up. This of course required changes to wherever `check_quantity` was used.

In fact, I think it would make sense to change this up further in a couple of different ways... but that's for another issue, I think.


StanczakDominik:
@pep8speaks pep8ify

plasmapy/utils/checks.py
```diff
@@ -11,20 +11,72 @@
...
+
+    This is probably best illustrated with an example:
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> @check_quantity(x={"units": u.m,
+    ...       "can_be_negative": False,
+    ...       "can_be_complex": True,
+    ...       "can_be_inf": True})
```
SolarDrew: This should change to reflect the new syntax.

SolarDrew: Disregard that, it's already right and I'm a moron.

PR #483, opened 2018-06-02
[WIP] Accept tuple or list of corresponding particles or atomic numbers in `particle_input`

ritiek:
Addresses #336. Now `particle_input` can take in a tuple or a list of particles or atomic numbers.

```python
>>> from plasmapy.atomic import Particle, particle_input

>>> @particle_input
>>> def happy_function(particles: (Particle, Particle)):
...    print(particles[0].integer_charge)
...    print(particles[1].integer_charge)

>>> happy_function(('e-', 'p+'))
-1
1

>>> @particle_input
>>> def cute_function(particles: [Particle]):
...    for particle in particles:
...        print(particle.integer_charge)

>>> cute_function(('e-', 'p+', 'alpha'))
-1
1
2
```

It raises a `TypeError` when input parameters do not match expected parameters, for example:
```python
>>> @particle_input
>>> def sad_function(particles: (Particle, Particle, Particle)):
...    print(particles[0].integer_charge)
...    print(particles[1].integer_charge)
...    print(particles[2].integer_charge)

>>> sad_function(('e-', 'p+'))
TypeError: Number of parameters allowed in the tuple (3 parameters) are not equal to number of parameters passed in the tuple (2 parameters).

>>> @particle_input
>>> def unlucky_function(particles: [Particle, Particle]):
...    print(particles[0].integer_charge)
...    print(particles[1].integer_charge)

>>> unlucky_function(('e-', 'p+'))
TypeError: A list can take any number of parameters without explictly mentioning the number of parameters in declaration. Use a tuple instead or '[Particle]' should be enough.
```

If this looks good enough, I can start writing tests for them!


ritiek:
I guess this is ready to merge. I'd be happy if someone can review it.

plasmapy/atomic/particle_input.py
```diff
@@ -127,9 +127,12 @@ def particle_input(wrapped_function: Callable = None,
     Raises
     ------
     `TypeError`
-        If the annotated argument is not a `str`, `int`, or
-        `~plasmapy.atomic.Particle`; or if `Z` or `mass_numb` is
-        not an `int`.
+        If the annotated argument is not a `str`, `int`, `tuple`, `list`
+        or `~plasmapy.atomic.Particle`; or if `Z` or `mass_numb` is
+        not an `int`; or if number of elements in a tuple do not match
+        the expected parameters; or if invalid list annotations in
```
StanczakDominik: Invalid list annotations sounds pretty unclear. However, I'm wondering whether we might be getting too verbose here... I think we can cut some of the description, if someone gets an exception they'll be looking at the code anyway, right?

plasmapy/atomic/particle_input.py
```diff
@@ -218,10 +238,33 @@ def wrapper(*args, **kwargs):
...
+                            f"Number of parameters allowed in the tuple "
+                            f"({expected_params} parameters) are "
+                            f"not equal to number of parameters passed in "
+                            f"the tuple ({received_params} parameters).")
+                elif isinstance(annotations[argname], list):
+                    annotated_argnames = annotations[argname]
+                    expected_params = len(annotated_argnames)
+                    if expected_params > 1:
+                        raise TypeError(
+                            f"A list can take any number of parameters "
```
StanczakDominik: Maybe "Put in [Particle] as the annotation to accept an arbitrary number of Particle arguments."

I love the idea though

plasmapy/atomic/tests/test_particle_input.py
```diff
@@ -173,6 +173,132 @@ def test_function_with_ambiguity():
...
+
+def test_invalid_number_of_tuple_elements():
+    with pytest.raises(TypeError):
+        # Passed 3 elements when function only takes 2 in tuple
+        function_with_tuple_annotation(('e+', 'e-', 'alpha'), q='test')
+
+
+def test_invalid_list_type():
+    @particle_input
+    # This should be just [Particle] instead of [Particle, Particle]
```
StanczakDominik: Good testing

plasmapy/atomic/particle_input.py
ritiek: I thought so too. If somebody gets an exception, they are told about the reason they are getting the exception. Maybe we should just have these
> If the annotated argument is not a `str`, `int`, `tuple`, `list` or `~plasmapy.atomic.Particle`; or if `Z` or `mass_numb` is not an `int`.

Not sure though.

plasmapy/atomic/particle_input.py
```diff
@@ -173,6 +176,25 @@ def particle_input(wrapped_function: Callable = None,
...
+        @particle_input
+        def decorated_tuple_function(particles: (Particle, Particle)):
+            return particles
+        sample_particles = decorated_tuple_function(('He', 'Li'))
+
+        @particle_input
+        def decorated_list_function(particles: [Particle]):
+            return particles
+        sample_particles = decorated_list_function(('Al 3+', 'C'))
+        sample_particles = decorated_list_function(['He', 'Ne', 'Ar'])
```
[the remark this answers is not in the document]
ritiek: Yep, `(Particle, Particle)` accepts only 2 elements and `(Particle, Particle, Particle)` accepts 3 elements and so on.. (either as tuple or list).

while `[Particle]` accepts any number of `Particle` elements (either as tuple or list).

Rust programming language has a similar concept of passing vectors (think: lists in Python) and tuples of specific type to a function. I imagined it would be nice here as well. :smile:

plasmapy/atomic/particle_input.py
ritiek: Nice. That makes it easier to understand. I think maybe "arbitrary number of Particle elements" would be a better choice than "arbitrary number of Particle arguments". What do you think?

plasmapy/atomic/tests/test_particle_input.py
ritiek: Thanks!

plasmapy/atomic/particle_input.py
StanczakDominik: Nah, I think we should rather keep the 'elements' key word for iron, hydrogen etc here :smiley:

plasmapy/atomic/particle_input.py
StanczakDominik: That's beautiful. Put that in! :)

plasmapy/atomic/particle_input.py
ritiek: Haha, ok got it.

plasmapy/atomic/particle_input.py
ritiek: Okie!

PR #485, opened 2018-06-05
Numpy array operations in collision_frequency 's

samurai688:
Similar to before, the goal this time is to make it so that collision_frequency and friends work with len>1 array Quantities, such as `T_e = np.array([1, 2, 3]) * u.eV`.

Working on issue #469, hopefully this is the last PR needed that will touch the functions themselves.  Muahaha, can calculate so many numbers now! 

Still will probably need another PR to close the issue, just to add tests for array-capability to all the various functions that were already passing.  Aaaand probably Braginskii should be figured out to handle arrays, but... let's make that a separate issue. I'm not sure if it will be vectorize-able in all cases,  I think it will be.


StanczakDominik:
As discussed yesterday, looks nice and mergeable, thanks @samurai688 !

PR #489, opened 2018-06-06
Plasma subclass tests as separate test files

ritiek:
This PR just modifies the directory structure for plasma subclass tests. We used to have tests for both (`plasmapy/classes/sources/`) `PlasmaBlob` and `Plasma3D` in `plasmapy/classes/tests/test_Plasma.py`. Now these tests have a separate file in `plasmapy/classes/sources/tests/`.

This is the way [SunPy does it](https://github.com/sunpy/sunpy/tree/aed106fabec889de4eea831500f7a6230cb1520d/sunpy/map/sources) and personally looks neater to me, we test for main Plasma classes in `plasmapy/classes/tests/` and our Plasma subclasses in `plasmapy/classes/sources/tests/`.


ritiek:
This is weird. I have no idea why this decreases coverage.

StanczakDominik:
I'll take a look in the evening. Probably nothing too serious.

StanczakDominik:
The weird thing is, I took a look at this locally, comparing the changes between master and this branch and it does decrease coverage! Those few lines are indeed getting uncovered for some reason. I'll have to do some more digging.

StanczakDominik:
This is so weird. It looks like the same tests are running. I think there are two options here. Either

* to figure out why this is happening, make another branch with these same changes, only commit them more gradually, pushing every time to see which particular change kills coverage
* decide it's not worth it to worry about 0.4%, write a few more tests on top of what we have now to cover the missing lines aaaaand merge away.

ritiek:
Aww yiss, we did it! So, it was just nasty `__init__.py` missing from `classes/sources/tests/`. In fact, `classes/sources/tests/*` tests weren't even running (and lol it still gave us so good coverage %).

One thing I noticed that those tests did run with missing `__init__.py` when running pytest directly `pytest plasmapy/classes/` but not with `python setup.py test`. Gotta remember this now.

This should be ready to merge if everything looks good.

StanczakDominik:
Darn. We might be missing a bunch of other tests if they don't run with `setup.py test` without `__init__.py`! I'll have to take a look. Thank you!

Also, that's one elegant change :D

PR #491, opened 2018-06-07
Setup description

### LINKED ISSUE #422: Improve project description on PyPI
[Project description](https://pypi.org/project/plasmapy/#description) is currently bland. We should detail some functionality, problems that the project is trying to address, mention openAstronomy affiliation, etc.

StanczakDominik:
Closes #422. I tried to stay pretty close to Astropy's PyPI page, but if there's something else we'd like to mention, suggestions are very welcome.

A related note: @namurphy, if you don't mind, I'd like to ask you to take a look at the LICENSE field in the classifiers part of setup.py. One thought I've just had is that maybe we're being too fancy with ours - there's a long list of OSI approved licenses on that PyPI classifier list page I linked as a TODO in there, and off the top of my head I have no idea which one would our fall under. I could take the top line and put "PlasmaPy License", but that's not going to tell anyone anything.

Basically, as it stands, for people like me whose mind shuts off at the sight of legalese, it might take hiring a lawyer to figure out what our license is, given we don't mention what it's based on! I seem to remember we were on BSD-3 Clause modified for patents...? But that's not listed in PyPI. I could put just "BSD License".

So yeah, this is a bit of a tangent and may be better in a separate issue, but I think it's better to start the discussion now while we're at it.


namurphy:
The license that we use is the Open Source Initiative approved [BSD+Patent license](https://opensource.org/licenses/BSDplusPatent).  

One possibility would be to separate the license into one file (`LICENSE.rst`), and the patent grant in another file (`PATENT.rst`).  One interesting thing about the BSD+Patent license (either a bug or a feature) is that it does not specifically require that the patent language has to be shared in the list of conditions.  It requires the copyright notice, the list of conditions (which refers to the bullet point list), and the disclaimer to be shared, but it doesn't explicitly say that the patent clause has to be shared.  Thus we would not need to ask for all contributors to grant permission to separate this into two files, and we could clarify that PlasmaPy can be shared simply under the BSD 3-clause license.

The patent grant is particularly important for PlasmaPy because there are lots of potential commercial applications, and there is a non-zero chance that a business could try to patent a contribution someone made and prevent us from using it in PlasmaPy.

StanczakDominik:
To be clear, I have nothing against the BSD+Patent license. I just think it might make sense to replace the relative uninformative `# PlasmaPy License` header with `# BSD+Patent License ` so that people know at a glance what license we're running with!

PR #492, opened 2018-06-10
Numpy array tests

samurai688:
Working on issue #469 .  This adds tests to the functions in collisions.py and parameters.py to verify that they can work with nparray Quantities.

The main thing with this one is the addition of a helper function in the utils/pytest_helpers module, assert_can_handle_nparray, which I then used all over test_collision and test_parameters. This new test does test with multi-dimensional nparray Quantities. Comments and improvements welcome!

I was hoping this was going to close the issue, but under the cold light of all the new testing, I think there are a handful more issues, although they are mostly in non-default-argument nooks and crannies. For the time being I just added todo comments. I can address them as part of this pull request if you guys would like, or do a new one.

### DIFF
--- a/plasmapy/utils/pytest_helpers.py
+++ b/plasmapy/utils/pytest_helpers.py
@@ -802,11 +802,11 @@ def assert_can_handle_nparray(function_to_test, insert_some_nans=[], insert_all_
         elif param_name == "characteristic_length":
             unit = u.m
             magnitude = 1.0
-            
+
         # else, last resort, if it has a default argument, go with that:
         elif not (param_default is _empty):
             return param_default, param_default, param_default
-        
+
         else:
             raise ValueError("Unrecognized function input")
 
--- a/plasmapy/utils/pytest_helpers.py
+++ b/plasmapy/utils/pytest_helpers.py
@@ -726,15 +726,18 @@ from inspect import signature, _empty
 from astropy.tests.helper import assert_quantity_allclose
 
 
-def assert_can_handle_nparray(function_to_test, force_args, override_V, override_cl, include_nans=[]):
+def assert_can_handle_nparray(function_to_test, can_be_vector="auto", insert_some_nans=[], 
+                              insert_all_nans=[], kwargs={}):
     """
-    Specifiied test for ability to handle numpy array quantities.
+    Specified test for ability to handle numpy array quantities.
     """
-    def prepare_input(param_name, param_default, force_args, include_nans, override_V, override_cl):
-        for key in force_args:
-            if param_name == key:
-                return force_args[key], force_args[key], force_args[key]
-        if param_name == "particle" or param_name == "ion_particle":
+    def prepare_input(param_name, param_default, can_be_vector, insert_some_nans,
+                      insert_all_nans, kwargs):
+        if can_be_vector != "auto":
+            raise NotImplementedError("Haven't implemeneted non-auto can_be_vector yet.")
+        if param_name in kwargs.keys():
+            return kwargs[param_name], kwargs[param_name], kwargs[param_name]
+        elif param_name == "particle" or param_name == "ion_particle":
             if not (param_default is _empty):
                 return param_default, param_default, param_default
             else:
@@ -753,10 +756,10 @@ def assert_can_handle_nparray(function_to_test, force_args, override_V, override
         elif param_name == "B":
             unit = u.G
             mag = 1e3
-        elif override_V and (param_name == "V"):
+        elif param_name == "V":
             unit = u.m / u.s
             mag = 1e5
-        elif override_cl and (param_name == "coulomb_log"):
+        elif param_name == "coulomb_log":
             unit = 1.0
             mag = 1e1
         elif not (param_default is _empty):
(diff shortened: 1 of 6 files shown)


plasmapy/utils/pytest_helpers.py
```diff
@@ -719,3 +720,145 @@ def run_test_equivalent_calls(*test_inputs, require_same_type: bool = True):
...
+                return ("e", "p"), ("e", "p"), ("e", "p")
+        elif param_name in ["T", "T_i", "T_e", "temperature"]:
+            unit = u.eV
+            magnitude = 1.0
+        elif param_name in ["n", "n_i", "n_e", "density"]:
+            unit = u.m ** -3
+            magnitude = 1e20
+        elif param_name == "B":
+            unit = u.G
+            magnitude = 1e3
```
StanczakDominik: Sure, but I thought we were going with reasonable units? :laughing: I kid. It's the first time I'm seing good old Gauss in `astropy.units`, though.

plasmapy/utils/pytest_helpers.py
```diff
@@ -719,3 +720,145 @@ def run_test_equivalent_calls(*test_inputs, require_same_type: bool = True):
...
+            input_data_1d[1] = np.nan
+        elif param_name in insert_all_nans:
+            input_data_2d = np.ones((2, 2)) * np.nan
+            input_data_1d = np.ones(4) * np.nan
+        input_data_2d *= magnitude
+        input_data_2d *= unit
+        input_data_1d *= magnitude
+        input_data_1d *= unit
+        input_data_0d = input_data_1d[3]
+        return input_data_0d, input_data_1d, input_data_2d
```
StanczakDominik: This is neat, I've got two thoughts here.

Firstly, I was thinking that this would be a nice candidate for a pytest fixture (finally!), but then again... the neat thing about this is that it's generated dynamically, and fixtures - at least, I can't think of a way they could handle that.

Secondly, this could be for a future improvement, but: do you think we could also test 3d arrays here? Those will probably be a common case here.

plasmapy/utils/pytest_helpers.py
```diff
@@ -719,3 +720,145 @@ def run_test_equivalent_calls(*test_inputs, require_same_type: bool = True):
...
+        scalar_testable = result_0d.value
+    except AttributeError:
+        scalar_testable = result_0d
+    if np.isscalar(scalar_testable):
+        astrohelper.assert_quantity_allclose(result_0d, result_1d[3])
+        astrohelper.assert_quantity_allclose(result_0d, result_2d[1, 1])
+    else:
+        for idx, res_0d in enumerate(result_0d):
+            astrohelper.assert_quantity_allclose(res_0d, result_1d[idx][3])
+            astrohelper.assert_quantity_allclose(res_0d, result_2d[idx][1, 1])
```
StanczakDominik: I would love to have an idea for how to make this have shared setup, be as easy to call as you've made it (:clap: ) and still get us independent result for cleaner pytest results. I don't, right now.

plasmapy/utils/pytest_helpers.py
samurai688: Haha, oh you don't like Gauss?  You should try it out, way better than Tesla. :smiling_imp: Especially kilogauss and megagauss.

plasmapy/utils/pytest_helpers.py
samurai688: We could do 3d arrays, I suppose, I was thinking that 0d, 1d, N-d would be good enough, but it might be good to explicitly test 0, 1, 2, and 3 since those correspond to euclidean space and will likely be used.  We could even go 4 to get time in there.  I don't know... on one hand, testing each might be overkill, but sometimes overkill is underrated.

plasmapy/utils/pytest_helpers.py
samurai688: Thanks!  Yeah... I was hoping to find a nice out-of-the-box pytest way to have it do what I wanted, but, I couldn't quite figure it out, so I have this thing at the moment. I'm 100% open to suggestions.

StanczakDominik:
For anyone curious, this seems to make our tests take ~160s (up from ~120s). Reasonable, to me.

samurai688:
Whoa really, 120 seconds to 160?  That seems longer than I was seeing locally, sorry if I dropped that on you.  I ran some local tests before I kill the branch and my timings were

branch:
2031 passed, 196 warnings in 121.04 seconds
2031 passed, 196 warnings in 127.37 seconds
2031 passed, 196 warnings in 130.32 seconds

(old) master:
1944 passed, 155 warnings in 110.97 seconds
1944 passed, 155 warnings in 137.01 seconds
1944 passed, 155 warnings in 116.46 seconds

Hmm, so a lot of variability, but overall I wasn't seeing quite as big of an impact as that. Probably a good 10 s in there though, which isn't great. I agree it's quite wasteful with all the different array setups, I can look at optimizing it in the next PR.  Next PR in which we will cloooooose this mega issue! :)

StanczakDominik:
They might take longer, but I might also have misread something. Don't worry too much about it!

PR #493, opened 2018-06-10
Use @particle_input decorator on functions in collisions.py

ritiek:
Addresses #336.

I've used `@particle_input` decorator wherever possible on functions in `collisions.py`. Also, replaced `TypeError` with a `ValueError` in `particle_input` when number of input elements do not match expected elements in the tuple passed, since it seems more fitting. :)

### DIFF
--- a/plasmapy/atomic/particle_input.py
+++ b/plasmapy/atomic/particle_input.py
@@ -242,7 +242,7 @@ def particle_input(wrapped_function: Callable = None,
                     expected_params = len(annotated_argnames)
                     received_params = len(arguments[argname])
                     if not expected_params == received_params:
-                        raise TypeError(
+                        raise ValueError(
                             f"Number of parameters allowed in the tuple "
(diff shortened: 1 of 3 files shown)


plasmapy/atomic/particle_input.py
```diff
@@ -242,7 +242,7 @@ def wrapper(*args, **kwargs):
                     expected_params = len(annotated_argnames)
                     received_params = len(arguments[argname])
                     if not expected_params == received_params:
-                        raise TypeError(
+                        raise ValueError(
```
StanczakDominik: This will probably take an update in the docstring

ritiek: Okay. I pushed some changes. Let me know if we need changes in the wording.

ritiek:
Yep, I don't think so. This should be ready to merge.

StanczakDominik:
All right, then, thanks!

PR #494, opened 2018-06-11
Add missing __init__.py in diagnostics/tests/

ritiek:
As suggested in https://github.com/PlasmaPy/PlasmaPy/pull/489#issuecomment-395860369 we're missing an `__init__.py` in `plasmapy/diagnostics/tests/` (and tests aren't probably running either). This PR just adds an empty `__init__.py` to that directory.


ritiek:
Woah! The tests weren't running indeed!

I skimmed through other tests directories, looks like they all have `__init__.py`, so we should be good to go.

StanczakDominik:
Whoa! You know, I might have removed those init files earlier, as I thought we wouldn't import stuff from there so they were unnecessary. Pretty crazy! Thanks for figuring this out!

PR #499, opened 2018-06-16
Initial tests for 2D OpenPMD sample

StanczakDominik:
All right, @ritiek, this is what I came up with for the initial test. I just went in and handled the 2D example dataset to begin with; we can move ahead to the 3D and the theta one later on, once we have something we're happy with.

I updated [the gist](https://gist.github.com/StanczakDominik/d7a6468eb41b92433676bcb08744af88) with an exploratory notebook via h5py (I haven't checked out the API yet).

I'm not sure about a few issues here: I think you'll find the notes in tests.

This was weird to write as I couldn't quite test it as I except the whole thing to fail instantly as we aren't handling hdf5's yet.


ritiek:
Thanks and that gist is something I'll certainly have to refer to :) As I see from the tests, we're going to be integrate our subclass with `numpy` and  `astropy.units`, seems a very good idea.

By the way, should we use `openPMD` bindings for python or just stick with `h5py`? I could get `openPMD` working for python using [conda installation](https://github.com/openPMD/openPMD-api#conda) method but [spack](https://github.com/openPMD/openPMD-api#spack) and [building from source](https://github.com/openPMD/openPMD-api#from-source) were a pain in vain for me. If you're going to try it, you might have better luck with conda like me.

ritiek:
I made a basic skeleton [here](https://github.com/ritiek/PlasmaPy/tree/openpmd-hdf5reader). Should I merge your changes on my branch and open a separate PR to PlasmaPy? I can't push stuff to your branch directly. :D

PR #500, opened 2018-06-17
[WIP] OpenPMD HDF5Reader Plasma subclass

ritiek:
Also see #499. This PR adds the ability to read HDF5 files based on OpenPMD standard when passing a keyword argument `hdf5`.

```python
>>> plasmapy.classes.Plasma(hdf5='plasmapy/classes/tests/data/data00000255.h5')
```

The initial idea is to set appropriate properties for dealing with 2D datasets on the subclass. Once it works out nicely, we can go ahead having a similar implementation for working with 3D datasets.


ritiek:
And happy #500!

ritiek:
I am going to compare [h5py](https://github.com/h5py/h5py) and [openPMD-api](https://github.com/openPMD/openPMD-api) soon, so we can choose whichever one seems better for our purpose.

ritiek:
OpenPMD API looks something like this:

```python
>>> import openPMD
>>> series = openPMD.Series('plasmapy/classes/tests/data/data00000255.h5',
                            openPMD.Access_Type.read_only)

>>> items = list(series.iterations.items())
>>> items
[(255, <openPMD.Iteration of at t = '0.000000 s'>)]

>>> iteration, data = items[0]
>>> meshes = data.meshes.items()
>>> particles = data.particles.items()

>>> list(meshes)
[('E', <openPMD.Mesh>), ('rho', <openPMD.Mesh>)]
>>> list(particles)
[('Hydrogen1+', <openPMD.ParticleSpecies>),
 ('electrons', <openPMD.ParticleSpecies>)]
```

I personally think we should just stick with `h5py`. It is [packaged on PyPi](https://pypi.org/project/h5py/) which makes it easy to distribute unlike `OpenPMD-api` which is available [only through conda (and spack)](https://github.com/openPMD/openPMD-api#installation). Also, it seems a bit weird to me working with `iter` objects with OpenPMD API.

I found we could also do this with h5py to navigate through multiple levels easily:
```python
>>> h5 = h5py.File('plasmapy/classes/tests/data/data00000255.h5')
>>> h5.get('data/255/fields/E/x')
<HDF5 dataset "x": shape (51, 201), type "<f8">
```

What do you guys think?

StanczakDominik:
Well, I guess we could start out with h5py for now. As long as we're using just hdf5 files, we don't really get advantages from the API right now besides potentially being nice to the guys at OpenPMD and testing their API out for them. But that's obviously a lower priority than getting this working! I'm fine with skipping that for now.

StanczakDominik:
By the way, I'm absolutely not sure we should default to OpenPMD on a "hdf5" keyword argument. I think there are two solutions:

* `openPMD = True` forces initialization via OpenPMD
* I think there's an `attr` saying this is an OpenPMD file in `h5 = h5py.File('plasmapy/classes/tests/data/data00000255.h5'); h5.attrs`. We could read that and if it's found, awesome, we know how to handle that. If it isn't, `ValueError` or something.

StanczakDominik:
Sounds reasonable :+1:

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -0,0 +1,46 @@
...
+    assert openPMD2DPlasma.charge_density.to(u.C/u.m**3)  # unless it's some
+    # 2D charge density
+
+def test_correct_shape_electric_field(openPMD2DPlasma):
+    assert openPMD2DPlasma.electric_field.shape == (3, 51, 201)
+    # IIRC this is how we defined it in the old Plasma class but I can be
+    # convinced otherwise if another way makes more sense
+
+def test_x(openPMD2DPlasma):
+    assert openPMD2DPlasma.x.shape == (51, 201)
```
ritiek: This is whose position coordinates we are talking about? If this is for particles like electrons, etc., it would make more sense to have something like `openPMD2DPlasma.electrons.x`.

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -0,0 +1,46 @@
...
+            "data",
+            "data00000255.h5")
+        )
+
+def test_has_electric_field_with_units(openPMD2DPlasma):
+    "It would be nice to let the fields have the correct units from the get go"
+    assert openPMD2DPlasma.electric_field.to(u.V / u.m)
+
+def test_has_charge_density_with_units(openPMD2DPlasma):
+    assert openPMD2DPlasma.charge_density.to(u.C/u.m**3)  # unless it's some
```
ritiek: We might be able to check units this way:
```python
>>> import h5py
>>> h5 = h5py.File('data00000255.h5')
>>> path = 'data/255/fields/rho'
>>> grid_size = len(h5[path].attrs['axisLabels'])
>>> units = u.C / u.m**grid_size
```

Would this work?

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -0,0 +1,46 @@
...
+    # 2D charge density
+
+def test_correct_shape_electric_field(openPMD2DPlasma):
+    assert openPMD2DPlasma.electric_field.shape == (3, 51, 201)
+    # IIRC this is how we defined it in the old Plasma class but I can be
+    # convinced otherwise if another way makes more sense
```
ritiek: How about something like this?
```python
assert openPMD2DPlasma.electric_field.x.shape == (51, 201)
assert openPMD2DPlasma.electric_field.y.shape == (51, 201)
assert openPMD2DPlasma.electric_field.z.shape == (51, 201)
```

ritiek: I'm fine with either of them though.

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
StanczakDominik: It definitely would! I thought I mentioned it, but guess not. OpenPMD lets you attach discrete particle data to your mesh data, but I think it would be fine to skip the `particle`s for now and just handle the mesh data well. Adding particles would mean we'd have to rethink our `species` class... which we could, of course, do.

Also, I now have doubts about the fact that OpenPMD lets you define separate coordinates for each field (e.g. you can have electric and magnetic fields defined at a spatial offset to each other, like for [Yee grids](https://upload.wikimedia.org/wikipedia/commons/thumb/1/1b/FDTD_Yee_grid_2d-3d.svg/800px-FDTD_Yee_grid_2d-3d.svg.png) )... we'll have to think more about how to handle this...

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -0,0 +1,46 @@
...
+            "data00000255.h5")
+        )
+
+def test_has_electric_field_with_units(openPMD2DPlasma):
+    "It would be nice to let the fields have the correct units from the get go"
+    assert openPMD2DPlasma.electric_field.to(u.V / u.m)
+
+def test_has_charge_density_with_units(openPMD2DPlasma):
+    assert openPMD2DPlasma.charge_density.to(u.C/u.m**3)  # unless it's some
+    # 2D charge density
```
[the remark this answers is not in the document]
StanczakDominik: Actually, there's a way that synergizes with what OpenPMD provides for us way more nicely:

```python
In [4]: dict(h5[path].attrs)['unitDimension']
Out[4]: array([-3.,  0.,  1.,  1.,  0.,  0.,  0.])
```

And then we can interpret that based on [documentation for the standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#required-for-each-record) as 

> powers of the 7 base measures characterizing the record's unit in SI (length L, mass M, time T, electric current I, thermodynamic temperature theta, amount of substance N, luminous intensity J)

so we'll have length^-3 (1/volume, reasonable), no mass, time ^1 and current ^1 gets you electric charge. We can then multiply each of these by its appropriate `astropy.unit` and we'll get the complete Quantity for that field! I really liked that system while reading the docs. I guess I was too hyped up to note that down.

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
StanczakDominik: That might make sense, but it also might make using the electric field harder in simulations if you can't just access it as one big 3D array... but there may be ways around that. I'm not sure what the optimal choice is here. I think this is a good point to talk about at tomorrow's telecon, if we're all up for that.

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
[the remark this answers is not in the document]
ritiek: Wow, this is really neat!

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
ax3l: > I think it would be fine to skip the particles for now

yes, fully optional. mesh-only works as well.

> [o]penPMD lets you define separate coordinates for each field

Just in case this is relevant for you. Otherwise just assign them to zero for now if it's irrelevant for your use case (e.g. node-centered).

ax3l:
> User could force HDF5 to read as [o]penPMD by passing `openPMD=True`. If this keyword argument is not supplied, we could automatically decide checking `openPMDextension` attribute.

For *reading* a `.h5` file I would do the following detection (as we implemented in [yt](https://github.com/yt-project/yt/blob/yt-3.4.1/yt/frontends/open_pmd/data_structures.py#L495-L519)):
- is a hdf5 file?
  - [yes] has the attribute [`openPMD` in `/`](https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#hierarchy-of-the-data-file)?
    - [yes]attribute `openPMD` contains a supported version (e.g. >= 1.1.0 < 2.0.0) [yes]
      - [yes] is use openPMD reader
      - [no] warn user about unsupported openPMD version, if version in file is *too old*, [offer an update](https://github.com/openPMD/openPMD-updater#openpmd-updater) of it
  - [no] test for a different HDF5 reader

[`openPMDextension`](https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#hierarchy-of-the-data-file) is not the main identifier, but this attribute will also be present. It allows to add further, *domain-specific meaning* on top of the *base standard* data description, e.g. for [*electro-magnetic PIC codes*](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_ED-PIC.md) (1.0+), [*physical particle species*](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_SpeciesType.md) (2.0+) or [*accelerator codes*](https://github.com/openPMD/openPMD-standard/pull/183) (2.0+). Maybe you even want to propose your own as well later on?

ritiek:
Ok, as of now we can handle both 2D and 3D HDF5 datasets for mesh (electric field and electric density) data but not particle data. What do we want to do with particle data? Should we just implement them the way we did with mesh data?

ritiek:
Also, should we ship the example OpenPMD HDF5 datasets that we are using for our tests with our main codebase? Maybe we could have them downloaded the moment a user runs those tests but I don't know how hard that would be practically.

StanczakDominik:
Oh yeah, downloading the test data is a sick idea that I completely forgot about! Maybe we could use that --online flag to setup.py test and emulate Astro/sunpy functionality.

By the way, we'll have to keep in mind to rebase and remove the files from these commits before we merge this PR into master. I can do that without a problem.

ritiek:
Good idea. `--remote-data` flag seems good to me.

We need to upload `data00000100.h5` and `data00000255.h5` somewhere. Do you have any preference? I'd put them on AWS but they ask for scary details during registration. :/

> By the way, we'll have to keep in mind to rebase and remove the files from these commits before we merge this PR into master. I can do that without a problem.

Yep, once we are ready to merge this, we're going to have to take care of this.

ritiek:
We need to replace the URLs in `test_openpmd_hdf5.py` once we have those datasets uploaded somewhere and then we should be ready to merge this as a prototype. :D

ritiek:
There is one problem at the moment that test data would need be to downloaded every single time the tests are run. We could workaround this by creating a module similar to https://github.com/sunpy/sunpy/blob/master/sunpy/data/sample.py and then data would need downloading only once?

For example, SunPy downloads all the data at once and stores it somewhere for future use when a code involving external datasets is first run.
```python
>>> import sunpy.map
>>> import sunpy.data.sample  
>>> mymap = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE) 
```

StanczakDominik:
Having thought about this for a bit, I think we're fine putting this sample data directly in the repository. These datasets are not really large, and [SunPy appears to include a bit of its sample data](https://github.com/sunpy/sunpy/tree/master/sunpy/data/test) right in the repo. I thus think we're fine not having to set up AWS or other storage for now!

We definitely need to reference the source for it (which I probably forgot to do...), and I think it may be good to store it in `plasmapy/data/` - this would be a better place for Langmuir data as well, most likely.

ax3l:
Just a note: I am not 100% sure we already compressed the checked-in files since we use them under git lfs. Please try to use `h5repack`, e.g. with `gzip`, on them and see if their size changes before checking them in to save space. Here is an example ("h5compress"): https://github.com/ax3l/cluster-scripts

ritiek:
@ax3l Looks like they were indeed not compressed. Total size for (those two) HDF5 files reduced from about 9 MB to 6 MB. Thanks!

@StanczakDominik @ax3l We already mention from where we downloaded the datasets, is there anything else we need to do?

ax3l:
That looks perfect, you also mention the license CC-0 so I think that's it :-)

ax3l:
Sorry for the hijack of the thread: I would like to mention PlasmaPy with the PlasmaPy logo in upcoming conference talks/posters as part of the "openPMD community". Is that fine with you? I won't go into details in those but I think it's cool to mention that you are in the process of adding openPMD support.

We also should not forget to add you here so people find you: https://github.com/openPMD/openPMD-projects

StanczakDominik:
@ritiek awesome, looking into the changes and I hope I can merge this in today!

@ax3l by all means, go right ahead. Pick a version you like from [the logo source repo](https://github.com/PlasmaPy/PlasmaPy-logo)! I'd love to see the poster once you're done there.

Also, if you want to not hijack issues in the future, check out [our Matrix chatroom](https://riot.im/app/#/room/#plasmapy:openastronomy.org) or our [mailing list](https://groups.google.com/forum/#!forum/plasmapy)! :D

plasmapy/classes/sources/openpmd_hdf5.py
```diff
@@ -0,0 +1,136 @@
...
+        ----------
+        electric_field : `astropy.units.Quantity`
+            An (x, y, z) array containing electric field data.
+        charge_density : `astropy.units.Quantity`
+            An array containing charge density data.
+
+        Parameters
+        ----------
+        hdf5 : `str`
+            Path to HDF5 file.
```
StanczakDominik: Looks like this could use a `References` link to the site: http://openpmd.org/

plasmapy/classes/sources/openpmd_hdf5.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Core class for accessing various attributes on HDF5 files that
+        are based on OpenPMD standards.
+
+        Attributes
+        ----------
+        electric_field : `astropy.units.Quantity`
+            An (x, y, z) array containing electric field data.
+        charge_density : `astropy.units.Quantity`
+            An array containing charge density data.
```
StanczakDominik: So... my bad here, but I guess I kind of forgot that the 2D  and 3D examples were showing off electrostatic + particle simulations. If you look at the thetaMode example, you will find two more mesh quantities:

```
(base) 20:04:58 dominik@dell: ~/Code/openPMD/openPMD-example-datasets/example-thetaMode/hdf5 $ h5ls -r data00000100.h5 
/                        Group
/data                    Group
/data/100                Group
/data/100/fields         Group
/data/100/fields/B       Group
/data/100/fields/B/r     Dataset {3, 51, 201}
/data/100/fields/B/t     Dataset {3, 51, 201}
/data/100/fields/B/z     Dataset {3, 51, 201}
/data/100/fields/E       Group
/data/100/fields/E/r     Dataset {3, 51, 201}
/data/100/fields/E/t     Dataset {3, 51, 201}
/data/100/fields/E/z     Dataset {3, 51, 201}
/data/100/fields/J       Group
/data/100/fields/J/r     Dataset {3, 51, 201}
/data/100/fields/J/t     Dataset {3, 51, 201}
/data/100/fields/J/z     Dataset {3, 51, 201}
/data/100/fields/rho     Dataset {3, 51, 201}
```
B being the magnetic field and J the electric current, which can be neglected in electrostatic simulations. I think we should probably handle those as well - electrostatics are a fraction of all possible simulations. Do you think it would be much of a hassle to add them?

plasmapy/classes/sources/openpmd_hdf5.py
```diff
@@ -0,0 +1,136 @@
...
+                               "standards defined by OpenPMD")
+
+    @property
+    def electric_field(self):
+        path = 'data/' + self.subname + '/fields/E'
+        if path in self.h5:
+            units = _fetch_units(self.h5[path].attrs["unitDimension"])
+            return np.array((self.h5[path]['x'],
+                             self.h5[path]['y'],
+                             self.h5[path]['z'])) * units
```
StanczakDominik: I'll definitely have to look at how `np.array` here is going to work... but it's nice enough for the prototype!

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -0,0 +1,76 @@
...
+    # Downloaded from
+    # https://github.com/openPMD/openPMD-example-datasets/blob/draft/example-3d.tar.gz
+    # per the Creative Commons Zero v1.0 Universal license
+    h5 = openpmd_hdf5.HDF5Reader(hdf5=os.path.join(rootdir, "data00000100.h5"))
+
+    def test_has_electric_field_with_units(self):
+        assert self.h5.electric_field.to(u.V / u.m)
+
+    def test_correct_shape_electric_field(self):
+        assert self.h5.electric_field.shape == (3, 26, 26, 201)
```
StanczakDominik: I might mess around with this one so that the interface handles better in a soon-to-be-made PR.

plasmapy/classes/sources/openpmd_hdf5.py
StanczakDominik: A few tests for any of those datasets would also be very nice, of course.

plasmapy/utils/exceptions.py
```diff
@@ -78,6 +78,11 @@ class InvalidParticleError(AtomicError):
     pass
 
 
+class OpenPMDError(PlasmaPyError):
```
StanczakDominik: Maybe `DataStandardError`? `OpenPMDError` feels quite specific.

plasmapy/classes/sources/openpmd_hdf5.py
ritiek: Sure! I'll work on them today.

ritiek:
I've pushed the suggested changes. For the failing coverage, its because we don't test for exceptions. We would have to include more data which in my opinion isn't worth it. Otherwise, this should be ready to merge.

PR #508, opened 2018-07-03
Python 3.7 tests

### LINKED ISSUE #433: Test against Python 3.7
Python 3.7 is due to be released around 2018 June 15.  Once that happens, we'll need to include both Python 3.6 and Python 3.7 in our test suite on Travis CI and AppVeyer.  I'm not sure what we'll need to do for CircleCI.

We should check out [what's new in Python 3.7](https://docs.python.org/3.7/whatsnew/3.7.html) when we do this too, since any removed features that cause errors should be described there.  If our most recent `0.1.*` release doesn't work on Python 3.7, then we'll need to do a patch fix release quickly.

Cross-reference: #264

StanczakDominik:
Closes #433. Hopefully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
         - PYTHON_VERSION=3.6
         - SETUP_CMD='test'
         - MAIN_CMD='python setup.py'
-        - CONDA_DEPENDENCIES='Cython'
+        - CONDA_DEPENDENCIES='Cython astropy scipy h5py matplotlib mpmath'
         - CONDA_ALL_DEPENDENCIES='Cython astropy scipy h5py matplotlib mpmath'
 
 # TODO include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,36 @@ python:
 git:
     submodules: false
 
+env:
+    global:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,22 +6,43 @@ python:
...
+        - env: PYTHON_VERSION=3.7 SETUP_CMD='test'
+
 # Install dependencies
 install:
         - git clone git://github.com/astropy/ci-helpers.git --depth 1
         - source ci-helpers/travis/setup_conda.sh
-        - pip install -r requirements/requirements.txt
-        - pip install codecov coveralls flake8 pytest-cov pytest-astropy
+#        - pip install -r requirements/requirements.txt
+#        - pip install codecov coveralls flake8 pytest-cov pytest-astropy
```
namurphy: Do we still need `coveralls` here since we switched primarily to `codecov`?  I'm not sure if `codecov` uses `coveralls` or something like that.

StanczakDominik: Oh yeah that's a mistake right there. Thanks, I'll get rid of it!

I also saw a neat way to simplify our requirements files on sunpy so I guess I'll do that as well while I'm at it.

PR #510, opened 2018-07-10
Add `isotope_name` attribute on `Particle` instance

### LINKED ISSUE #503: Add an `isotope_name` attribute on `Particle` class
> We could definitely add an `isotope_name` attribute as well, which for isotopes in general would return something like `'iodine-131'` plus `'deuterium'` and `'tritium'` for those special cases. The `element` and `element_name` attributes just return the name of the element without respect to the isotope. I'll try to remember to create a separate issue for this.

[as mentioned here](https://github.com/PlasmaPy/PlasmaPy/pull/468/files#r189747292) by @namurphy.

ritiek:
Closes #503.

If we want to return `deuterium` and `tritium` for those special cases, maybe we should make appropriate changes to `element_name` property since `isotope_name` just returns formatted output of `element_name` (and `mass_number`)? Or do we want to add these checks specifically to `isotope_name` so that `element_name` stays as is?

I haven't added any checks for these special cases yet, so at the moment it looks like this:
```python
>>> deuterium = Particle("D+")
>>> deuterium.element_name
'hydrogen'
>>> deuterium.isotope_name
'hydrogen-2'
```


ritiek:
Okay, got it.

plasmapy/atomic/tests/test_particle_class.py
```diff
@@ -196,6 +203,7 @@
         'particle': 'H-1 0+',
         'element': 'H',
         'isotope': 'H-1',
+        'isotope_name': 'hydrogen-1',
```
namurphy: There are two names that we could adopt for [H-1](https://en.wikipedia.org/wiki/Isotopes_of_hydrogen#Hydrogen-1_(protium)): `'hydrogen-1'` and `'protium'`.  Do we have a preference between the two?  Using `'protium'` would be more consistent with the way we use `'deuterium'` and `'tritium'`, but `'protium'` is less well known.

SolarDrew:
LGTM :+1:

plasmapy/atomic/tests/test_particle_class.py
StanczakDominik: I like hydrogen-1 here, I don't think I've ever heard "protium" used. That's not to say I've heard much.

PR #446, opened 2018-05-08
Added axis labels to ITER parameters plots; Fixing #428

manasbedmutha98:
Adds axis labels to the plot of Langmuir Plasma Frequency vs Electron Concentration. Will fix #428

PR #511, opened 2018-07-13
Fix URL formatting in `online_help()`

ritiek:
It seems like this issue went unnoticed while merging #438 (later superseded by #446). It was causing spaces to accumulate in the URL just before the query part, so it returned no search results in browser:

```python
>>> from plasmapy import online_help
# This used to open
# http://docs.plasmapy.org/en/stable/search.html?%20%20%20%20q=particle&check_keywords=yes&area=default
# instead of actual url
# http://docs.plasmapy.org/en/stable/search.html?q=particle&check_keywords=yes&area=default
>>> online_help("particle")
```

This PR makes a small change to URL formatting to prevent spaces from accumulating so docs search can work properly.


SolarDrew:
:+1:

Issue #93, opened 2017-08-24
Get onto Conda


StanczakDominik:
This is somewhat related to #36. We need to be accessible and most of the cutting edge scientific Python is on [Conda](https://conda.io/docs/), probably due to ease of downloading stuff like C extensions, Numba etc.

Anyway, this gets closed once we can do `conda install plasmapy` in a new environment and have it work out of the box.

I haven't been able to find a how-to on this yet. I'll update this issue if I find anything.

https://conda-forge.org/ this is at least 70% relevant.

SolarDrew:
I'll sort this out if nobody objects. I've been meaning to learn how to do conda packaging anyway and unless anyone has a pressing desire to take on dealing with the pip installing it seems reasonable for me to deal with both.

wtbarnes:
Relevant PR: conda-forge/staged-recipes#4793

What is the status of this? The problem (at least right now) with the staged recipe is that the `roman` package is not included, presumably because it was added as a dependency after this PR was opened

StanczakDominik:
The necessary functionality of `roman` got absorbed into `plasmapy/utils/roman` in an earlier PR, so that seems to have been solved. I suppose it would be best to continue this discussion at conda-forge/staged-recipes#4793 and get this done.

StanczakDominik:
I finally got the builds to pass - the waiting game begins now.

StanczakDominik:
We are now on Conda-forge. :)

https://github.com/conda-forge/plasmapy-feedstock

PR #514, opened 2018-07-16
Incorporate roman package into PlasmaPy

namurphy:
Earlier this year, we added the roman package as a dependency so that we can deal with spectroscopic notation for ions (#286, #290, #464, #468, #504).  While roman is available on PyPI, it is not available on conda-forge.  Because of that, we ran into problems while trying to make PlasmaPy available on conda-forge as well (related to #93).  Our options were to put roman on conda-forge to fix this, or to incorporate roman directly into PlasmaPy.  Because all of the needed functionality in roman is contained in its `src/roman.py` file, it makes a lot of sense to just incorporate that file into PlasmaPy under the terms of its license.  

This PR incorporates roman into PlasmaPy as `plasmapy.utils.roman` and removes roman as one of PlasmaPy's dependencies.  I put this into `plasmapy.utils` instead of something like `plasmapy.extern` because I thought this was where it would make the most sense to put the functionality.  I added detailed docstrings with examples, and did some refactoring.  I also added some detailed tests in order to help with reliability and maintainability (in particular by testing the full range from 1 to ~120 because those are the numbers we would need for spectroscopic notation).

The roman package is copyright 2001 by Mark Pilgrim and is available under the terms of the [Python 2.1.1 license](https://www.python.org/download/releases/2.1.1/license/).


ritiek:
I don't know whether it would be a good idea but we could also add the roman package as a submodule to PlasmaPy (similar to the way we have astropy_helpers). One good thing would be that people would immediately realize it as an external dependency but would increase the repo size a bit by including files unnecessary to us like [CHANGES.txt](https://github.com/zopefoundation/roman/blob/8b2654f58bb411bda6a7e02796b6ed218dd45a42/CHANGES.txt), etc.

I have never worked with submodules myself before, so take everything I said with a pinch of salt.

Cadair:
I would advocate for `plasmapy.extern` and having a:
```
try:
    import roman
except ImportError:
    import _roman as roman
```

so that if people install it via pip, or for system package managers like debian etc (who really don't like vendored things) can easily just drop back to the externally manged version.

plasmapy/utils/roman/roman.py
```diff
@@ -0,0 +1,213 @@
...
+    >>> is_roman_numeral("CXVII")
+    True
+    >>> is_roman_numeral("42")
+    False
+
+    """
+    if not isinstance(s, str):
+        raise TypeError("Only strings may be tested ")
+    result = bool(_romanNumeralPattern.match(s))
+    return result
```
StanczakDominik: Oh wow, you weren't joking when you said this was a small module. There's literally more *glorious numpydoc* in here than code, though if I remember correctly... well procrastinated, @namurphy ? :smile:

PR #517, opened 2018-07-20
Make Particle(particle) == Particle(Particle(particle))

### LINKED ISSUE #486: Have Particle class return corresponding Particle instance for Particle arguments
I've run into a bunch of situations where I have some objects that are either `Particle` class instances or `str`/`int` representations of particles, and I want to make sure that they are all `Particle` class instances.  I have usually written code like this:
```Python
>>> orig_particles = ['H', Particle('He')]
>>> particles = [
        particle if isinstance(particle, Particle) else Particle(particle) 
        for particle in orig_particles
    ]
```
This code is rather obfuscating since it checks on whether or not `particle` is a `Particle` class instance.  It would be simpler to have `Particle` return the corresponding `Particle` instance which would allow code like this:
```Python
>>> particles = [Particle(particle) for particle in particles]
```
@StanczakDominik suggested that this be a copy of the `Particle` instance rather than returning self, which is like how `numpy.array` behaves.

A good test would be to check that `Particle(something) == Particle(Particle(something))`, e.g., something like:
```Python
from plasmapy.atomic import Particle
def test_particle_of_particle(particle_representation):
    particle = Particle(particle_representation)
    particle_of_particle = Particle(particle)
    assert particle == particle_of_particle
    assert particle is not particle_of_particle  # to make sure it is a copy rather than self
```
Thanks! 
-Nick

namurphy:
This makes `Particle(particle) == Particle(Particle(particle))` while making sure that `Particle(particle) is not Particle(Particle(particle))`. 

Closes #486, using the suggestion made by @colbych.  See #486 and the commit log for details.

### DIFF
--- a/plasmapy/atomic/particle_class.py
+++ b/plasmapy/atomic/particle_class.py
@@ -261,10 +261,18 @@ class Particle:
         attributes.
         """
 
-        if not isinstance(argument, (int, str)):
+        if not isinstance(argument, (int, np.integer, str, Particle)):
             raise TypeError(
                 "The first positional argument when creating a Particle "
-                "object must be either an integer or string.")
+                "object must be either an integer, string, or another"
+                "Particle object.")
+
+        # If argument is a Particle instance, then we will construct a
+        # new Particle instance for the same Particle (essentially a
+        # copy).
+
+        if isinstance(argument, Particle):
+            argument = argument.particle
 
         if mass_numb is not None and not isinstance(mass_numb, int):
             raise TypeError("mass_numb is not an integer")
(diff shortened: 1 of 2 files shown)


namurphy:
I am wondering if I set the record for the number of times a particular word is used in a pull request title.

plasmapy/atomic/particle_class.py
```diff
@@ -238,6 +239,25 @@ class Particle:
...
+
+    If the previously constructed `~plasmapy.atomic.Particle` instance
+    represents an element, then the `Z` and `mass_numb` arguments may be
+    used to specify an ion or isotope.
+
+    >>> iron = Particle('Fe')
+    >>> Particle(iron, Z=1)
+    Particle("Fe 1+")
+    >>> Particle(iron, mass_numb=56)
+    Particle("Fe-56")
```
StanczakDominik: I really like how this interface came out!

PR #519, opened 2018-07-21
Implement ionize and recombine methods in Particle class

### LINKED ISSUE #287: Enable ionization and recombination of Particle objects
There may be occasions where it will be useful to be able to change a `Particle` object representing an ion into a higher or lower charge state.  For example, someone might want to loop through the different charge states of an ion to return something like the ionization energy for each one.

One possibility for the notation would be to create `Particle` class methods called `ionize` and `recombine`:
```Python
>>> from plasmapy.atomic import Particle
>>> deuteron = Particle('D 0+').ionize()
>>> protium = Particle('p+').recombine()
```
These methods would probably be straightforward to implement.

Another possibility would be to use magic methods in the `Particle` class, and enable electrons to be added to or subtracted from ions:
```Python
>>> electron = Particle('e-')
>>> deuteron + electron
Particle('D 0+')
>>> deuteron + 'e-'
Particle('D 0+') 
```
This would involve defining the `Particle.__add__` and `Particle.__sub__` magic methods.

I'm not sure how commonly this would be used, so this might be lower priority.  Alternatively, we could use these magic methods for getting nuclear reaction energies (e.g., `Be-8 > 2 * He-4`) also or instead.

namurphy:
These methods take a `Particle` object and run `__init__` again for the same element/isotope but for a different charge.  The `Particle.ionize` method will raise an `InvalidIonError` if a bare nucleus is attempted to be ionized.  The `recombine` method will issue a warning if the charge becomes negative; however, no exception will be raised in that case.

One behavior that I'm not certain about is both running `__init__` again and returning the new Particle instance.  We may wish to have just one behavior or the other.  If we have just the in-place re-initialization, then a disadvantage is that we would not be able to do comparisons like:
```Python
Particle("Fe 5+").recombine() == Particle("Fe 4+")
```
which require the new `Particle` instance to be returned.  I would appreciate more opinions on whether we should do both `__init__` and returning `self`, or just one of them.  Or, if there is even a better approach than I am taking here.  

I'd also like to check to make sure that there are no side effects from `__init__` again (e.g., if I have a list and add items to it such that every item in the list shows up twice).

Closes #287.

### DIFF
--- a/plasmapy/atomic/particle_class.py
+++ b/plasmapy/atomic/particle_class.py
@@ -5,6 +5,7 @@ import warnings
 from typing import (Union, Set, Tuple, List, Optional)
 import collections
 import plasmapy.utils.roman as roman
+import numbers
 
 import astropy.units as u
 import astropy.constants as const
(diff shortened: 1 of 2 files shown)


plasmapy/atomic/particle_class.py
```diff
@@ -1328,3 +1329,136 @@ def is_ion(self) -> bool:
...
+
+        base_particle = self.isotope if self.isotope else self.element
+        new_integer_charge = self.integer_charge + n
+
+        self.__init__(base_particle, Z=new_integer_charge)
+
+        # Returning self here allows comparisons like
+        # Particle("Fe 5+").ionize() == Particle("Fe 6+")
+
+        return self
```
StanczakDominik: One could probably argue for an `inplace=True` default  argument, like Pandas does for a lot of its methods. It would be pretty simple  to include, simply wrap the two previous lines of code in an `if` statement.

namurphy: Interesting!  Using the same notation as :panda_face: would be a really good idea.

namurphy:
I implemented @StanczakDominik's suggestion, so I'll plan on merging this soon unless there are any more suggestions.

I also rebased this to remove extraneous formatting commits and get a cleaner commit history.  I'm starting to appreciate how helpful rebasing can be in keeping the commit log clean while still being able to keep track of the commits that show the important parts of the development history and/or contain information that we may wish to remember for the future.

PR #520, opened 2018-07-23
Reorganize documentation on installing PlasmaPy

namurphy:
Right now we have information on how to install PlasmaPy in a couple of places.  The instructions for using `pip` to install PlasmaPy from PyPI are in `INSTALL.md` in our top-level directory, so the installation link in our docs points back to our GitHub repository which means that information won't show up in searches.  We also have installation information in our code guide on how to set up and install a development version of PlasmaPy in our development guide in the same file as our coding style guidelines, which I've been thinking about splitting off into a different file to separate installation information and code style guidelines.

In this PR I am planning to:
 - [x] Move the information in `INSTALL.md` into `docs/install.rst`.
 - [x] Put a link to the installation information in our docs from `INSTALL.md`.
 - [x] Extend the information in `docs/install.rst` to include how to install PlasmaPy in a conda environment (recommended) or a virtual environment.
 - [x] Recommend against installing PlasmaPy with a system installed Python and link to this [xkcd on Python installations](https://xkcd.com/1987/).
 - [x] Put our development guide's instructions on how to set up a development version of PlasmaPy into a separate file.
 - [x] Link to the development guide's setup instructions from `docs/install.rst`.

Thoughts are welcome!  I also might need some help with virtual environments since I've never used them.


namurphy:
I think I made all of the changes that I was hoping to make for the main installation instructions (`docs/install.rst`).  There probably are additional updates to go into the installation instructions in the development guide (`docs/development/install-dev.rst`) but I'd prefer to create an issue and leave these to another pull request (since otherwise it'd probably take me another couple of weeks to get to it).  Comments and suggestions are welcome!

namurphy:
> This monster of a command could probably be replaced by the much simpler conda env create -f requirements/environment.yml, assuming we make sure that file's updated... which I'm doing right now.

The downside of  creating a conda environment from `environment.yml` in these instructions is that it assumes that new users will have downloaded the repository, whereas users doing `pip install plasmapy` will have not done so.  I had included the long `conda create` command because I thought it would be easier for users to copy and paste that command than do the additional steps of downloading the file, making sure it's in the right location, and then copying the command.  For the moment, perhaps we could list both techniques for depending on whether or not the user is going to do `pip install plasmapy` or has downloaded the source.

This problem should be temporary since we should eventually be able to do something like
```
conda create -n plasmapy python=3.7 plasmapy -c conda-forge
```
once we have conda-forge set up.

StanczakDominik:
That's a very good point which I did not consider at all. I'll be happy to revert my change once I'm back home from work today, and then we can merge this one!

StanczakDominik:
And here we go! Thanks, @namurphy !

PR #525, opened 2018-07-29
Allow nans by default when checking unit input

dstansby:
I think this was the intended default (the documentation says the default should be `True`), and I also think in general most functions should be able to handle nans. Also added a test to check that `beta` passes through nans fine.

### DIFF
--- a/plasmapy/physics/tests/test_dimensionless.py
+++ b/plasmapy/physics/tests/test_dimensionless.py
@@ -24,4 +24,5 @@ def test_beta_nan():
     n = np.array([1, 1]) * u.cm**-3
     T = np.array([1, 1]) * u.K
     out = beta(T, n, B)
-    assert out[1] == np.nan * u.dimensionless_unscaled
+    assert np.isnan(out[1])
+    assert out[1].unit == u.dimensionless_unscaled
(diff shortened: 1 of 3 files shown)


plasmapy/physics/tests/test_dimensionless.py
```diff
@@ -1,12 +1,27 @@
...
+    float(quantum_theta(T, n))
+
+
+def test_beta_nan():
+    # Check that nans are passed through properly
+    B = np.array([1, np.nan]) * u.T
+    n = np.array([1, 1]) * u.cm**-3
+    T = np.array([1, 1]) * u.K
+    out = beta(T, n, B)
+    assert out[1] == np.nan * u.dimensionless_unscaled
```
StanczakDominik: Apparently NaN has unusual comparison rules:
```python
import numpy as np
np.nan == np.nan
Out[3]: False
```

This is what I got working here, which I think is close to what you intended:
```python
    assert np.isnan(out[1])
    assert out[1].unit == u.dimensionless_unscaled
```

dstansby: Looks good to me 👍

PR #531, opened 2018-08-08
Updated typechecks in function definitions

### LINKED ISSUE #518: Use ABCs from number package to type check integers & real numbers
Up until recently, I've been writing code like this:
```Python
def return_integer(n):
    if not isinstance(n, int):
        raise TypeError
    return n
```
This is all well and good...until we try to pass in a `numpy.int` object and get a `TypeError`.  Grumblemuffins!

However, abstract base classes to the rescue!  We can instead do things like:
```Python
import numbers
def return_integer(n):
    if not isinstance(n, numbers.Integral):
        raise TypeError
    return n
```
This function now works for `numpy.int` objects as if by magic!  Similarly, we can use `numbers.Number`, `numbers.Complex`, and `numbers.Real` instead of checking for, e.g., a `float` or `complex` object.

However, I've included code like the first example above in quite a few places, including in `atomic` and `mathematics`, and this is likely to have come up in a bunch of other places as well.  We should update these type checks so that we use `isinstance(n, numbers.Integral)` and `isinstance(r, numbers.Real)` instead of `isinstance(n, int)` and `isinstance(r, float)`.

I've also used `int` instead of `numbers.Integral` in type hint annotations as well.  However, this isn't necessarily as bad since type hints are not automatically used for type checks.  That said, if we want to use [mypy](http://mypy-lang.org/), we would probably want to update the type hint annotations too.

mj-fischer:
Used the Integral, Complex, and Real abstract base classes from the
'numbers' module to update the 'isinstance' typechecks in function
definitions. Also updated the function hint annotations and return types
as well for integration with mypy and other such features.

Closes #518.


StanczakDominik:
Thanks for doing this! According to CircleCI:
```python
Exception occurred:
  File "/home/circleci/project/plasmapy/atomic/symbols.py", line 95, in <module>
    def isotope_symbol(isotope: Particle, mass_numb: numbers.Integral = None) -> str:
NameError: name 'numbers' is not defined
```
So we're going to have to import the `numbers` module wherever it isn't visible (like `symbols.py`).

mj-fischer:
Just saw this, sorry about that. Fixing now. 


Michael

StanczakDominik:
I figured out the bug with importing `atomic`! I'll push to your branch, please pul once my change is there...

StanczakDominik:
Oh wait, your fix (lowercase integer_charge) would probably do it just as well... That's fine, then. Whatever's happening on travis-ci seems pretty arcane. Going to run tests locally and see what's up.

PR #532, opened 2018-08-10
Clean up ReST formatting and links in docs

namurphy:
I'm going through some parts of the docs and finding ReST formatting and links that need to be updated.  For example, I'm using single back ticks for when linking is necessary (e.g., to functions, classes, class methods, etc.) and double back ticks for code samples.  I'm starting with the atomic subpackage.  This should be a fairly minor PR.


StanczakDominik:
What's 532e41817947f3d3b35a84cfa6d786128a38b441 doing in the commit history here? Am I right guessing that it's 6b4682e85fac61b7a61a248178b7d5a3e051acbd reborn in the fires of a rebase? :smile:

namurphy:
> What's 532e418 doing in the commit history here? Am I right guessing that it's 6b4682e reborn in the fires of a rebase? :smile: 

Oops, yeah, it would seem so!

PR #533, opened 2018-08-15
Turn h5py into an optional dependency

### LINKED ISSUE #522: Importing a plasmapy module fails if h5py isn't installed
```python
    from plasmapy.physics import dimensionless
  File "/Users/dstansby/github/plasmapy/plasmapy/__init__.py", line 29, in <module>
    from . import classes
  File "/Users/dstansby/github/plasmapy/plasmapy/classes/__init__.py", line 4, in <module>
    from . import sources
  File "/Users/dstansby/github/plasmapy/plasmapy/classes/sources/__init__.py", line 3, in <module>
    from .openpmd_hdf5 import HDF5Reader
  File "/Users/dstansby/github/plasmapy/plasmapy/classes/sources/openpmd_hdf5.py", line 1, in <module>
    import h5py
ModuleNotFoundError: No module named 'h5py'
```
The problem is that `__init__.py` in the `plasmapy` directory tries to import *everything*, which means e.g. `h5py` is required for plasmapy to work. I'm guessing this is an optional dependency, and shouldn't be required?

StanczakDominik:
Closes #522 .

This is a pretty minimal solution - I'm not sure about the best practice for this kind of thing.
I could also, I suppose, put `import h5py` into the two functions that actually use it. That way, they'd get imported at first function runtime, if I'm correct.

### DIFF
--- a/plasmapy/classes/sources/openpmd_hdf5.py
+++ b/plasmapy/classes/sources/openpmd_hdf5.py
@@ -1,4 +1,3 @@
-import h5py
 import numpy as np
 import astropy.units as u
 
@@ -8,6 +7,11 @@ from plasmapy.utils import DataStandardError
 import os
 from distutils.version import StrictVersion
 
+# make h5py an optional dependency
+try:
+    import h5py
+except ImportError:
+    h5py = None
 
 _OUTDATED_VERSION = "1.1.0"
 _NEWER_VERSION = "2.0.0"


plasmapy/classes/sources/openpmd_hdf5.py
```diff
@@ -8,6 +7,11 @@
 import os
 from distutils.version import StrictVersion
 
+# make h5py an optional dependency
+try:
+    import h5py
+except ImportError:
+    h5py = None
```
dstansby: I would raise another `ImportError` here, but with a message that is PlasmaPy specific (something like 'Using the openpmd_hdf5 module requires an installation of the hdf5 library').

ritiek: @dstansby Don't think that would solve the original problem. This will still give an error when importing `plasmapy` module.
```python
>>> import plasmapy
  File "/home/ritiek/PlasmaPy/plasmapy/classes/sources/openpmd_hdf5.py", line 12, in <module>
    import h5py
ModuleNotFoundError: No module named 'h5py'

During handling of the above exception, another exception occurred:

  File "/home/ritiek/PlasmaPy/plasmapy/__init__.py", line 29, in <module>
    from . import classes
  File "/home/ritiek/PlasmaPy/plasmapy/classes/__init__.py", line 4, in <module>
    from . import sources
  File "/home/ritiek/PlasmaPy/plasmapy/classes/sources/__init__.py", line 3, in <module>
    from .openpmd_hdf5 import HDF5Reader
  File "/home/ritiek/PlasmaPy/plasmapy/classes/sources/openpmd_hdf5.py", line 15, in <module>
    raise ImportError("Using the openpmd_hdf5 module requires "
ImportError: Using the openpmd_hdf5 module requires an installation of the hdf5 library
```

ritiek:
If we're indeed going to make h5py an optional dependency, it might also be useful with the current approach to check if `h5py` is `None` when creating an `HDF5Reader` instance and raise a custom error message about missing h5py package if it is. At the moment, this isn't helpful:
```python
AttributeError: 'NoneType' object has no attribute 'File'
```

StanczakDominik:
Yeah, that was a terrible first solution. I think it'd be best to discuss this at the community meeting tomorrow. I might even write a PLEP to summarize this and [neat discussion we had on Matrix](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/ 5343287502750mRBUJ:matrix.org).

plasmapy/classes/sources/openpmd_hdf5.py
dstansby: Ahhh, that's because importing `plasmapy` imports *everything* - is this deliberate/is there a good reason for doing this?

StanczakDominik: It was sort of deliberate, but I wouldn't be sure about a good reason existing for this...

StanczakDominik:
This PR is a bad first attempt and must be destroyed. Let me try this again.

PR #538, opened 2018-09-18
Move transport from physics to its own subpackage

namurphy:
I was using PlasmaPy to do some calculations, and decided that it would have been easier and more intuitive if `transport` was its own subpackage rather than being a subsubpackage.  This PR thus moves `plasmapy.physics.transport` to `plasmapy.transport`.  Plasma transport theory is really broad, and I think warrants its own subpackage.

At @StanczakDominik's suggestion, I used PyCharm to make this change which made things a lot easier and I only had to fix one error manually.  I still need to update the narrative documentation.


namurphy:
So we meet again, CircleCI, my old arch-nemesis...

namurphy:
@StanczakDominik - thanks for the review!  Also this PR is probably of interest to @lemmatum and @samurai688.

PR #540, opened 2018-09-22
Namedtuples for dielectric 

StanczakDominik:
Closes #539. Names can of course probably be improved.

PR #544, opened 2018-09-28
Return namedtuples in dielectric.py functions

### LINKED ISSUE #539: Return namedtuples in dielectric.py functions
There are a few functions in [`plasmapy/physics/dielectric.py`](https://github.com/PlasmaPy/PlasmaPy/blob/1511167bfb7631b0b54d1e4919d0adaf001c59aa/plasmapy/physics/dielectric.py) that return three values, e.g.
```Python
def cold_plasma_permittivity_SDP(B, species, n, omega):
    ...
    return S, D, P
```
It would probably be helpful if we could return these as `collections.namedtuple` objects (or perhaps a new class) since that would group the values together in a convenient object.  If it is returned as a `namedtuple`, then it should preserve existing behavior when doing: 
```Python
>>> S, D, P = cold_plasma_permittivity_SDP(B, species, n, omega)
```
Thanks!
-Nick

HaraldNordgren:
Closes #539.

### DIFF
--- a/plasmapy/physics/dielectric.py
+++ b/plasmapy/physics/dielectric.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 from astropy import units as u
+from collections import namedtuple
 from plasmapy import utils
 from plasmapy.physics import parameters
 from plasmapy.constants import (pi, m_e, c, mu0, e, eps0)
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Hey @HaraldNordgren, sorry, but this was already mostly done + has tests in #540 :( so I guess I'll close this one.

PR #548, opened 2018-10-02
Add Mypy Support To Travis

### LINKED ISSUE #268: Use mypy for static type checking in continuous integration

```Python
def func() -> u.eV:
```

BTaskaya:
Add Mypy Support To Travis.

Fixes #268 

(mypy will throw alot errors when you merge this, be careful)


BTaskaya:
+1

PR #554, opened 2018-10-05
Issue #515: ignore CouplingWarning, RelativityWarning when running tests

### LINKED ISSUE #515: Reduce number of warnings issued during tests
Right now our tests are getting a lot of warnings that we can safely ignore.  These warnings are often a `RelativityWarning` or a `CouplingWarning`, and they don't need to show up when running our test suite because we are testing values for regimes where those warnings are issued.  Example warnings are:

```
  plasmapy/physics/transport/tests/test_collisions.py::Test_Knudsen_number::()::test_symmetry
  /tmp/plasmapy-test-72pekk_1/lib/python3.6/site-packages/plasmapy/physics/transport/collisions.py:284: CouplingWarning: Coulomb logarithm is 3.4858100325145664, you might have strong coupling effects
    utils.CouplingWarning)
  /tmp/plasmapy-test-72pekk_1/lib/python3.6/site-packages/plasmapy/physics/transport/collisions.py:284: CouplingWarning: Coulomb logarithm is 3.4858100325145664, you might have strong coupling effects
    utils.CouplingWarning)

plasmapy/physics/transport/tests/test_transport.py::Test_classical_transport::()::test_resistivity_units
  /tmp/plasmapy-test-72pekk_1/lib/python3.6/site-packages/plasmapy/utils/checks.py:433: RelativityWarning: V is yielding a velocity that is 6.256% of the speed of light. Relativistic effects may be important.
    RelativityWarning)
```
It would be helpful to [suppress the warnings using a context manager](https://docs.python.org/3.7/library/warnings.html#temporarily-suppressing-warnings), for example like:

```Python
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    ...
```

The `DeprecationWarnings` are covered by a previous issue (#345).  Thanks!

schambers:
Closes #515 

You can specifically ignore CouplingWarning, RelativityWarning in setup.cfg. After looking at a couple of different options, this seemed to be the easiest and best approach. This doesn't ignore all warnings as mentioned as an undesirable approach in #515 comments, it ONLY ignores the two mentioned warning types.

This brings the number of warnings when running tests down to 85. The remaining warnings are likely valid and should be investigated separately.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,8 @@ norecursedirs = "astropy_helpers" "docs" "build" "docs/_build" "examples" "auto_
 filterwarnings =
     once::DeprecationWarning
     once::PendingDeprecationWarning
+    ignore::plasmapy.utils.exceptions.RelativityWarning
+    ignore::plasmapy.utils.exceptions.CouplingWarning
 
 [ah_bootstrap]
 auto_use = True


schambers:
haha. ty!

It wasn't a very deep change, just more of research. Worth noting that there's quite a few ways to tweak the warnings. This seemed to be the easiest approach without smattering changes around a lot of tests.

PR #556, opened 2018-10-06
Add Optional type hint for arguments with a default value of None

### LINKED ISSUE #553: Use Optional explicitly for type hints when arguments default to None
I just looked up [PEP 484 -- Type Hints](https://www.python.org/dev/peps/pep-0484) again in response to a comment from @StanczakDominik in #385.  PEP 484 now says:

> A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
> ```Python
> def handle_employee(e: Employee = None): ...
> ```
> This would have been treated as equivalent to:
> ```Python
> def handle_employee(e: Optional[Employee] = None) -> None: ...
> ```
> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

There are a lot of places in PlasmaPy where we use the first format above (from when that format was recommended), but explicit is better than implicit, so we should change everything to the second format.

It would probably also be good to consistently use `from` imports for the `typing` package, like `from typing import Optional, Dict, List`. Using `typing.Optional` rather than just `Optional` in annotations really reduces readability, especially if we have multiple type hint annotations in the same function/method call.  All of this would be good to add to the code guide in our development guide too.

Thanks!
-Nick

lgoenner:
Closes #553

### DIFF
--- a/docs/development/code_guide.rst
+++ b/docs/development/code_guide.rst
@@ -49,6 +49,9 @@ Coding Style
 * For multiline imports, instead of using ``\`` to break lines, wrap the
   imported names in ``()`` parentheses and use consistent whitespace.
 
+* Use ``Optional[type]`` for type hinted keyword arguments with a default value of
+  ``None``.
+
 * There should be at most one pun per 1284 lines of code.
 
 Branches, commits, and pull requests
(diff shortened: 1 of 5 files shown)


StanczakDominik:
Looks good, I'll merge it! Thanks, @lgoenner !

Issue #345, opened 2018-04-06
Resolve DeprecationWarnings associated with Quantity comparisons


namurphy:
When running our tests on Travis CI, we are getting a bunch of warnings associated with something like `astropy.units.Quantity` objects.  We should track down the reason why we are getting this `DeprecationWarning` so frequently, and resolve it if possible.  If it's something in Astropy itself that hasn't been addressed in the development version, then we could contribute an upstream fix or just ignore it until something breaks.
```
plasmapy/atomic/atomic.py::plasmapy.atomic.atomic.common_isotopes
 /home/travis/miniconda/envs/test/lib/python3.6/site-packages/astropy/units/quantity.py:1016:
DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
    return super().__eq__(other)
```

To get each `DeprecationWarning` to show up in pytest, we can put the following lines under the `[tool:pytest]` in `setup.cfg` (though we don't want these enabled by default):
```
filterwarnings =
    once::DeprecationWarning
    once::PendingDeprecationWarning
```

StanczakDominik:
After much deliberation, I have come to the conclusion that I dunno.

It's something between `common_isotopes`, `isotopic_abundance` and `known_isotopes`, but at the same I haven't been able to pinpoint exactly where they happen. They also pop up in multiple places in `test_atomic`.

lgoenner:
I took a look at this and the biggest issue is [here](https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/atomic/particle_class.py#L428): 
`if attributes['half-life'] == np.inf * u.s`
Sometimes `attributes['half-life']` is a string (see [here](https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/atomic/isotopes.py#L974) or [here](https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/atomic/isotopes.py#L407)), which causes the comparison with a Quantity to raise a warning.

Most of the `AstropyDeprecationWarning `come from implicit boolean casts such as [this one](https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/classes/sources/tests/test_openpmd_hdf5.py#L53).

lgoenner:
Currently creating a PR with some example fixes

PR #557, opened 2018-10-07
Reduce pytest warnings

lgoenner:
See #345 

- Added an isinstance check that catches strings before their comparison with a `Quantity`.
Since the comparison should always return false anyway, this should be correct
- Changed the asserts of quantities to an isinstance check since the implicit boolean cast always returns true and raises a warning

### DIFF
--- a/plasmapy/atomic/particle_class.py
+++ b/plasmapy/atomic/particle_class.py
@@ -399,7 +399,6 @@ class Particle:
                 else:
                     attributes['half-life'] = Isotope.get('half-life', None)
 
-
             if element and not isotope:
                 attributes['standard atomic weight'] = Element.get('atomic mass', None)
 


lgoenner:
[10 warnings](https://travis-ci.org/PlasmaPy/PlasmaPy/jobs/438204604#L1346) compared to [65 warnings](https://travis-ci.org/PlasmaPy/PlasmaPy/jobs/438067961#L1566) previously

plasmapy/classes/sources/tests/test_openpmd_hdf5.py
```diff
@@ -16,7 +16,7 @@ class TestOpenPMD2D:
     h5 = openpmd_hdf5.HDF5Reader(hdf5=os.path.join(rootdir, "data00000255.h5"))
 
     def test_has_electric_field_with_units(self):
-        assert self.h5.electric_field.to(u.V / u.m)
+        assert isinstance(self.h5.electric_field.to(u.V / u.m), u.Quantity)
```
StanczakDominik: So the inner statement checks whether `electric_field` converts to V/m, and then `isinstance` returns a boolean on the type check. Neat solution!

plasmapy/atomic/particle_class.py
```diff
@@ -425,7 +425,9 @@ def __init__(
             categories.add('uncharged')
 
         if attributes['half-life'] is not None:
-            if attributes['half-life'] == np.inf * u.s:
+            if isinstance(attributes['half-life'], str):
+                categories.add('unstable')
+            elif attributes['half-life'] == np.inf * u.s:
```
StanczakDominik: Looks okayish to me. @namurphy ?

Issue #547, opened 2018-10-02
Use absolute rather than relative imports


namurphy:
**Note: This issue is reserved for a pull request tutorial.**

The [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008) for Python code recommends [using absolute imports rather than relative imports](https://www.python.org/dev/peps/pep-0008/?#imports).  It is preferable to write
```Python
from plasmapy import utils  # more readable, leads to better error messages
```
rather than
```Python
from .. import utils  # unclear where ".." refers to
```

For our tutorial, we will each pick one of these files and change the relative imports to absolute imports and then go through the mechanics of a pull request.
 - [x] `plasmapy/atomic/atomic.py`
 - [x] `plasmapy/atomic/nuclear.py`
 - [x] `plasmapy/atomic/parsing.py`
 - [ ] `plasmapy/atomic/particle_class.py`
 - [ ] `plasmapy/atomic/particle_input.py`
 - [x] `plasmapy/atomic/special_particles.py`
 - [ ] `plasmapy/atomic/symbols.py`
 - [x] `plasmapy/classes/species.py`
 - [ ] `plasmapy/transport/braginskii.py`

Before the tutorial, it would be helpful to [create a (free) GitHub account](https://github.com/join).

The steps we will take during the tutorial will be:
 1. Fork PlasmaPy's repository (by clicking the button in the upper right corner that says "Fork")
 2. Clone your fork of PlasmaPy to your laptop with a command like: 
     `git clone https://github.com/your_username/PlasmaPy.git`
 3. Enter PlasmaPy's repository
    `cd PlasmaPy`
 4. Create a branch with a descriptive name: 
    `git branch new-branch-name`
 5. Check out the branch with git:
    `git checkout new-branch-name`
 6. Tell git which branch to link to on your fork on GitHub
    `git push --set-upstream your_username new-branch-name`
 7. Enter the file of your choice and change relative imports to absolute imports
 8. Commit your changes
    `git add path/to/edited-file.py`
    `git commit -m "Change relative to absolute imports in PlasmaPy"`
 9. Push your local changes to the branch you just created on GitHub
    `git push`
 10. [Create a pull request](https://help.github.com/articles/creating-a-pull-request/), and reference this issue (#547)

Looking forward to it!
-Nick

namurphy:
Partially addressed by #550, #551, #552.  Thank you to our new contributors! :smiley_cat:   

[...]

PR #558, opened 2018-10-09
Change relative to absolute imports in PlasmaPy

fsilvapavon:
Edited lines using relative imports to absolute standard


namurphy:
Addresses #547.

PR #560, opened 2018-10-11
[WIP] Dreicer electric field

### LINKED ISSUE #529: Calculate Dreicer electric field
We should have a function that calculates the Dreicer electric field, and include a good physical description of what the Dreicer electric field means and it's significance.

My own preference would be to have a function name that represents the physics of what's happening rather than including someone's name, though I don't have a good idea for what that name might be.

The docstring should also cite the [original paper](https://journals.aps.org/pr/abstract/10.1103/PhysRev.115.238).

Thanks!

StanczakDominik:
Closes #529. 

I'm not quite sure if all the variables are correct (oh, transport theory, you magnificent mess). I found a description in Callen's draft book, and I think I kind of sort of understand the process - but I'll look through the other positions linked in the initial issue and see whether I couldn't perhaps find some numerical comparisons to ensure I'm right here.


plasmapy/transport/collisions.py
```diff
@@ -1791,3 +1796,99 @@ def coupling_parameter(T,
...
+
+    T : ~astropy.units.Quantity
+        Temperature in units of temperature or energy per particle,
+        which is assumed to be equal for both the test particle and
+        the target particle.
+
+    n_i : ~astropy.units.Quantity
+        The ions density in units convertible to per cubic meter.
+
+    ion_particle : Particle
```
namurphy: Perhaps shorten this to just `ion`?

plasmapy/transport/collisions.py
```diff
@@ -1791,3 +1796,99 @@ def coupling_parameter(T,
...
+
```
namurphy: Should add `coulomb_log` to the list of parameters.

plasmapy/transport/collisions.py
```diff
@@ -1791,3 +1796,99 @@ def coupling_parameter(T,
...
+    References
+    ----------
+    TODO
+    Callen chapter 2
+    """
+    if coulomb_log is None:
+        coulomb_log = Coulomb_logarithm(T,
+                                        n_i,
+                                        (Particle('e'), ion_particle), method=coulomb_log_method
+                                        )
```
namurphy: The part that I was confused about earlier was which two particles to calculate the Coulomb logarithm for.  I think you have this right, though I don't fully understand the derivation.

StanczakDominik: I'll check this in a few more sources.

lemmatum: pg 242 of the Dreicer paper says "Let us consider electron-ion encounters", so I think this is right...?

plasmapy/transport/collisions.py
```diff
@@ -1794,3 +1799,105 @@ def coupling_parameter(T,
...
+@check_quantity()
+def Dreicer_electric_field(T: u.K,
+                           n_i: 1/u.m**3,
+                           ion: Particle,
+                           coulomb_log: numbers.Real=None,
+                           vTh: u.m / u.s = np.nan * u.m / u.s,
+                           coulomb_log_method: str = "classical",
+                           ) -> u.V / u.m:
+    r"""
+    Calculates the minimum electric field for runaway electrons to occur
```
namurphy: Perhaps change `Calculates` to `Calculate` for consistency with other functions

StanczakDominik: Done!

plasmapy/transport/collisions.py
```diff
@@ -1794,3 +1799,104 @@ def coupling_parameter(T,
...
+@particle_input
+@check_quantity()
+def Dreicer_electric_field(T: u.K,
+                           n_i: 1/u.m**3,
+                           ion: Particle,
+                           coulomb_log: numbers.Real = None,
+                           vTh: u.m / u.s = np.nan * u.m / u.s,
+                           coulomb_log_method: str = "classical",
+                           ) -> u.V / u.m:
+    r"""
```
SolarDrew: Why's this a raw string?

plasmapy/transport/collisions.py
```diff
@@ -1794,3 +1799,104 @@ def coupling_parameter(T,
...
+    ~astropy.units.UnitsWarning
+        If units are not provided, SI units are assumed
+
+    ~plasmapy.utils.RelativityWarning
+        If the input velocity is greater than 5% of the speed of
+        light.
+
+    Notes
+    -----
+    TODO
```
StanczakDominik: Right, I guess this shouldn't get merged yet...

plasmapy/transport/collisions.py
StanczakDominik: For latex formulas in the currently empty notes section - thanks for making me check those!

PR #561, opened 2018-10-12
Second attempt at optional dependencies

### LINKED ISSUE #522: Importing a plasmapy module fails if h5py isn't installed

StanczakDominik:
Closes #522 later.

This implements what I started writing in PLEP8 (`pip install plasmapy[all]`) and solves the issue of what @dstansby [was referring to over here](https://github.com/PlasmaPy/PlasmaPy/pull/533#discussion_r220880549). This is going to be a breaking change (I'm not yet 100% sure I've got the implementation right), but I believe it's one we need to make (will explain later).

Checklist:
* [x] extras_require
* [x] remove main namespace imports
* [x] update documentation

### DIFF
--- a/plasmapy/optional_deps.py
+++ b/plasmapy/optional_deps.py
@@ -1,0 +1,50 @@
+""" Useful error messages for optional dependencies that aren't found. """
+from typing import Optional
+
+def _template(pkgname: str,
+              url: str,
+              library: Optional[str] = None) -> ImportError:
+    """
+    Simple template to prevent text duplication.
+
+    Parameters
+    ----------
+    pkgname : str
+        Package name on pip or conda
+    url : str
+        Link to install page for given package
+    library : str (optional)
+        Full package name
+
+    Returns
+    -------
+    ImportError
+
+    """
+    library = pkgname if library is None else library
+    template = f"""
+    {library} could not be found. Try either
+
+        conda install {pkgname}
+        
+    on Anaconda environments or
+        
+        pip install {pkgname}
+
+    in general. In case of trouble refer to
+    {url}"""
+
+    return ImportError(template)
+
+h5py_import_error = _template("h5py",
+                              "http://docs.h5py.org/en/latest/build.html")
+
+mpl_import_error = _template("matplotlib",
+                             "https://matplotlib.org/users/installing.html")
+
+mpmath_import_error = _template("mpmath",
+                                "http://mpmath.org/doc/current/setup.html#download-and-installation")
+
+lmfit_import_error = _template("lmfit",
+                               "https://lmfit.github.io/lmfit-py/installation.html")
+
--- a/plasmapy/optional_deps.py
+++ b/plasmapy/optional_deps.py
@@ -1,9 +1,12 @@
 """ Useful error messages for optional dependencies that aren't found. """
 from typing import Optional
 
-def _template(pkgname: str,
-              url: str,
-              library: Optional[str] = None) -> ImportError:
+
+def _optional_import_error_template(pkgname: str,
+                                    url: str,
+                                    library: Optional[str] = None,
+                                    conda_channel: Optional[str] = None,
+                                    ) -> ImportError:
     """
     Simple template to prevent text duplication.
 
@@ -15,6 +18,8 @@ def _template(pkgname: str,
         Link to install page for given package
     library : str (optional)
         Full package name
+    conda_channel: str (optional)
+        set to "conda-forge" to add -c conda-forge to the conda install line
 
     Returns
     -------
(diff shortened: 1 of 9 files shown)


dstansby:
I think it might be most useful to explicitly specify optional dependencies for each sub-module of plasmapy, instead of for certain functionality. That way `pip install plasmapy[submodule]` is a nice way of doing "I want to use this sub-module, this command will install all the dependencies I need".

StanczakDominik:
Ooooh. That's a really neat idea!

StanczakDominik:
Okay, now this seems rather silly. Apparently, some of our .rst documentation files were not getting checked for doctests (I have no idea what caused them to get tested now). The first of these is not a big deal - a nuclear reaction syntax example.

But the second is `plasmapy.test()`.

In a doctest.

And this is getting launched now.

StanczakDominik:
I did forget to update the docs to account for @dstansby 's idea, so that's one thing!

StanczakDominik:
All in all, I'm not sure the split of optional dependencies that we have right now (as per David's suggestion) is optimal. I have no idea how to describe this in docs. What we might need to do is reorganize our subpackages again:
* merge `math`, `physics` and `transport` into, maybe, `theory` (this would then involve lmfit and mpmath)
* keep diagnostics depending on `matplotlib` (maybe even throw in an optional pandas interface because if you're working with experimental data you most likely want something compatible with dataframes)
* rename `classes` somehow because that name is completely out  of whack right now (there's a bunch of classes in each module now). I'm not sure what a good name would be.
* remove unused `io` and `visualization` subpackages, maybe even get rid of `plasmapy.constants` because they're wrappers to astropy anyway.

But that's probably out of scope for this PR. For optional dependency name sets, I'm kind of pondering going back to what we had before 414c and leaving the above changes for a future PR.

Thoughts?

StanczakDominik:
I should probably still add tests to the generated ImportErrors. I'll try to figure out a way to do that.

plasmapy/optional_deps.py
```diff
@@ -0,0 +1,50 @@
...
+                              "http://docs.h5py.org/en/latest/build.html")
+
+mpl_import_error = _template("matplotlib",
+                             "https://matplotlib.org/users/installing.html")
+
+mpmath_import_error = _template("mpmath",
+                                "http://mpmath.org/doc/current/setup.html#download-and-installation")
+
+lmfit_import_error = _template("lmfit",
+                               "https://lmfit.github.io/lmfit-py/installation.html")
```
namurphy: The hyperlinks for the installation pages may change over time because of [link rot](https://en.wikipedia.org/wiki/Link_rot).  I wonder if we should include tests to make sure that links that we have in our package continue to work.  We would probably want to have a general test helper function since this issue may come up elsewhere in the package.  This seems out of scope for this particular PR, so we could wait to do this to the future if we decide to do this at all.

plasmapy/optional_deps.py
```diff
@@ -0,0 +1,50 @@
+""" Useful error messages for optional dependencies that aren't found. """
+from typing import Optional
+
+def _template(pkgname: str,
```
namurphy: Perhaps rename `_template` to something more specific, like `_import_error_template`?

plasmapy/optional_deps.py
```diff
@@ -0,0 +1,50 @@
...
+    {library} could not be found. Try either
+
+        conda install {pkgname}
+        
+    on Anaconda environments or
+        
+        pip install {pkgname}
+
+    in general. In case of trouble refer to
+    {url}"""
```
namurphy: This error message would indeed be helpful!

namurphy:
> I should probably still add tests to the generated ImportErrors. I'll try to figure out a way to do that.

The only way I can think of for testing if an `ImportError` or `ModuleNotFoundError` is raised appropriately would be to create a separate Travis CI test with the minimal installation, which seems like it'd be too much.  I'd be okay with foregoing tests on these.

PR #562, opened 2018-10-12
Upkeep: use Particle class

### LINKED ISSUE #373: Integrate Particle into arguments in langmuir.py
Currently, the `langmuir.py` module uses the `gas` argument in a number of functions to define the mass of the plasma species. This should be refactored to use the Particle class instead.

carolyz:
Closes #373 

### Done:

- `gas` argument refactored to use Particle class
- replaced `gas` input for `get_ion_density_OML` and `get_ion_density_LM` with `gas.mass` to 
1. avoid touching those functions too much, and 
2. keep the docstrings the same
- `gas` removed from function decorators because Particle class has the necessary constraints and validation

### Not done:
- not sure what to rename `gas` to as I'm still familiarizing myself with the codebase

### DIFF
--- a/plasmapy/examples/plot_langmuir_analysis.py
+++ b/plasmapy/examples/plot_langmuir_analysis.py
@@ -39,14 +39,14 @@ probe_area = (probe_length * np.pi * probe_diameter +
 # ion mass number of 4 is entered. The results are visualized and the obtained
 # EEDF is also shown.
 print(swept_probe_analysis(characteristic,
-                           probe_area, 4 * u.u,
+                           probe_area, 'He-4+',
                            visualize=True,
                            plot_EEDF=True))
 
 ######################################################
 # The cyan and yellow lines indicate the fitted electron and ion currents,
 # respectively. The green line is the sum of these and agrees nicely with the
-# data. This indicates a succesfull analysis.
+# data. This indicates a successful analysis.
 
 ######################################################
 # The next sample probe data is provided by David Pace. It is also obtained
@@ -63,7 +63,7 @@ characteristic = Characteristic(np.array(bias) * u.V,
 # of the electron growth region will be plotted.
 swept_probe_analysis(characteristic,
                      0.738 * u.cm**2,
-                     40 * u.u,
+                     'Ar-40 1+',
                      bimaxwellian=False,
                      plot_electron_fit=True)
 
@@ -73,7 +73,7 @@ swept_probe_analysis(characteristic,
 # with bimaxwellian set to True, which yields improved results.
 print(swept_probe_analysis(characteristic,
                            0.738 * u.cm**2,
-                           40 * u.u,
+                           'Ar-40 1+',
                            bimaxwellian=True,
                            visualize=True,
                            plot_electron_fit=True))
@@ -105,7 +105,7 @@ probe_area = (probe_length * np.pi * probe_diameter +
 # The fit converges nicely to the two slopes of the electron growth region.
 print(swept_probe_analysis(characteristic,
                            probe_area,
-                           4 * u.u,
+                           'He-4+',
                            bimaxwellian=True,
                            plot_electron_fit=True,
                            visualize=True))


carolyz:
The error is `TypeError: The first positional argument when creating a Particle object must be either an integer, string, or anotherParticle object.` so it looks it's because `test_ion_mass_unit()` is passing a param that is now invalid, assuming it is still the old gas param. 

Should the test be updated to fix this? Or are there other options?

namurphy:
@carolyz - Thank you for working on this!  

You're right, the tests will need to be updated.  In the tests for `swept_probe_analysis`, the arguments like `40 * u.u` would need to be replaced with strings that represent different ions, like "Ar-40 1+" which has an isotopic mass of about 40 u.  The numbers that get compared in the tests will need to be updated too, I think.

namurphy:
I think what I'd change in `plasmapy/diagnostics/tests/test_langmuir.py` would be:
 - Change `40 * u.u` to `'Ar-40 1+'` in each of the methods in `Test__swept_probe_analysis`
 - Delete `test_ion_mass_unit` in `Test__swept_probe_analysis` (since I think it'll become obsolete after these changes)
 - Update the number in the `assert` statement in `test_get_ion_density_OML_without_return_fit`, and use `'p+'` instead of `1 * u.u` as the argument.

@lemmatum and @jasperbeckers, do either of you have any thoughts on how the tests would need to be updated?  This is one of the parts of the code that I'm least familiar with, unfortunately!

carolyz:
Thanks for the quick and helpful reply @namurphy! Travis is now complaining about `'p+'` not being a unit. What can `'p+'` be replaced with?

namurphy:
Hm...looks like the problem is this line:
```Python
n_i_OML = np.sqrt(-slope * u.mA ** 2 / u.V * np.pi ** 2 * gas / 
    (probe_area ** 2 * const.e ** 3 * 2))
```
I think what we need to do here is add `ion = Particle(gas)` before that line, and then change `gas` in the above line to `ion.mass`.

plasmapy/diagnostics/langmuir.py
```diff
@@ -302,7 +303,7 @@ def swept_probe_analysis(probe_characteristic, probe_area, gas,
     # The OML method is used to obtain an ion density without knowing the
     # electron temperature. This can then be used to obtain the ion current
     # and subsequently a better electron current fit.
-    n_i_OML, fit = get_ion_density_OML(probe_characteristic, probe_area, gas,
+    n_i_OML, fit = get_ion_density_OML(probe_characteristic, probe_area, gas.mass,
                                        return_fit=True)
```
StanczakDominik: I got the recent tests to pass by removing `.mass` from this line - `get_ion_density_OML` uses the `Particle`'s attributes inside.

plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -330,7 +313,7 @@ def test_get_ion_density_OML_without_return_fit():
 
     char = characteristic()
     density = langmuir.get_ion_density_OML(char, 5000000*u.m**2,
-                                           1*u.u, return_fit=False)
+                                           'p+', return_fit=False)
     assert np.isclose(density.value, 383949768.0764418)
```
StanczakDominik: I did also have to update this test value to `385344135.12064785`, which is a bit strange, but given that this was a regression test... no, actually, I'm not quite sure why this would change.

namurphy: The mass of a proton is about 1.007 u, so we're using a different mass.  The definition of an atomic mass unit is based on 1/12 of the mass of a neutral atom of carbon-12.

StanczakDominik: Okay, I guess I wasn't thinking clearly then. Too much multitasking is not good for your code reviews!

carolyz: Ah makes sense, going to do that as soon as I get home! I'm still new to tests so thanks for all the guidance so far!

StanczakDominik:
Getting close! Travis CI passed, now it's just CircleCI (and we do need to improve how that outputs its errors >_< ) which seems to be crashing on docstrings:
```python

/home/circleci/project/plasmapy/examples/plot_langmuir_analysis.py failed leaving traceback:
  File "/home/circleci/project/plasmapy/examples/plot_langmuir_analysis.py", line 44, in <module>
    plot_EEDF=True))
  File "/home/circleci/project/plasmapy/utils/checks.py", line 160, in wrapper
    return f(**given_params_values)
  File "/home/circleci/project/plasmapy/diagnostics/langmuir.py", line 290, in swept_probe_analysis
    gas = Particle(argument=gas_argument)
  File "/home/circleci/project/plasmapy/atomic/particle_class.py", line 293, in __init__
    "The first positional argument when creating a Particle "
TypeError: The first positional argument when creating a Particle object must be either an integer, string, or anotherParticle object.
```

So I'm guessing a doctest has to be updated for the new particle syntax instead of for mass input in terms of u.

carolyz:
I'm having trouble locating the doctest :( Any pointers? 
I noticed `swept_probe_analysis` is called in plot_langmuir_analysis.py with `40 * u.u`, does that need an update as well?

namurphy:
> I noticed swept_probe_analysis is called in plot_langmuir_analysis.py with 40 * u.u, does that need an update as well?

I think that the `40 * u.u` should be changed to `'Ar-40 1+'`, and that `1 * u.u` should be changed to `'p+'` for the functions that you changed.  

The Travis CI build was successful, so the doctests in docstrings should all be fine, I think.  The only errors I found were the ones in `plot_langmuir_analysis.py`.  

I'll be around for the next hour or so and will try to check GitHub in case any more questions come up.  Thank you again! :slightly_smiling_face:

namurphy:
> Thanks for all your help! Just to confirm, `4 u*u` means it's helium so I would replace that with `'He-4+'`?

Correct!  And thank you for making this contribution!

StanczakDominik:
And then you'll also need to update the value for that test, like before
(because 4u is close to the mass of helium but not quite)! Just a reminder
:)

namurphy:
It looks like the new test failure for AppVeyor is the same as in #570, and unrelated to this pull request.  Perhaps we can merge #570 and re-run the tests here, since I think this pull request is just about done.

carolyz:
Thanks so much for your help and guidance @namurphy @StanczakDominik

Issue #453, opened 2018-05-10
Remove implicit default particle assumptions


namurphy:
We currently specify a default particle in keyword arguments for most of our functions in the physics subpackage (including in a bunch of code that I originally wrote).  An sample:
```Python
from plasmapy.atomic import particle_input, Particle

@particle_input  # turn a particle string into a Particle class instance
def particle_mass(particle: Particle = 'e-'):
    return particle.mass
```
The disadvantage of setting a default particle is that users could end up making incorrect assumptions or interpretations of the results.  We do state the assumptions in the documentation, but I know I personally often just skim through documentation without reading it in detail and could end up missing something like this.  Different fields also assume different default ions (e.g., deuterons in fusion vs. protons in astronomy).  In some cases, users may not immediately realize that a particle needs to be specified.  These implicit assumptions can also make the resulting code less readable for someone who is unfamiliar with the package:
```Python
>>> particle_mass()  # unclear which particle we are finding the mass of
<Quantity 9.10938356e-31 kg>
```
Since explicit is better than implicit, I propose that we do not specify default particles in keyword arguments for our functions so that an exception is raised when particle information is not provided, and instead either not set a default particle (for positional arguments) or set the default particle to `None` (for keyword arguments):
```Python
@particle_input
def mass_of_particle(particle: Particle):
    return particle.mass

@particle_input
def particle_massiveness(particle: Particle = None):
    return particle.mass
```
Related to: #34, #252, #309, #341

StanczakDominik:
I think one point worth raising is that for cases like stuff in `parameters`, default arguments for particles do one important job well - they let you very quickly see roughly what format the argument should be in, without looking at the documentation. Thus, I would be against removing those defaults.

For stuff like in your example, `particle_mass` etc, that should absolutely not have default arguments.

As a relative outsider to the field, I think it's okay to keep electrons/protons for default particles: they feel like the most "basic" ones. I know we're getting into weird areas here here, but can you really say a deuteron is more "basic" than a proton?

namurphy:
I have been starting to think that we should not set a default particle for many of the same reasons why we should not specify a default magnetic field strength or density.  The composition of a plasma is fundamental to that plasma's properties, and thus really should be something that users specify.

[...]

Issue #341, opened 2018-04-06
Implement @particle_input and Particle class in physics subpackage


namurphy:
The [`@particle_input`](https://github.com/PlasmaPy/PlasmaPy/blob/28156768b242564265e33b3baf8ae231f1a7c166/plasmapy/atomic/particle_input.py#L66) decorator and [`Particle`](https://github.com/PlasmaPy/PlasmaPy/blob/28156768b242564265e33b3baf8ae231f1a7c166/plasmapy/atomic/particle_class.py#L100) class were recently created in order to take an object-oriented approach to atomic data, and simplify how particles as arguments are passed into functions.  The [`@particle_input` decorator documentation](http://docs.plasmapy.org/en/latest/atomic/decorators.html) and [`Particle` class documentation](http://docs.plasmapy.org/en/latest/atomic/particle_class.html) discuss how to use these.

The `@particle_input` decorator should be implemented in functions that take a particle as an input, and the resulting instance of the `Particle` class should be used to get things like a particle's mass and charge.  In situations where the particle is required to be, e.g., charged, then the `require`, `any_of`, and `exclude` keywords in `@particle_input` should be used.  

The files to be updated are:
 - [ ] parameters.py
 - [ ] parameters_cython.pyx
 - [ ] distribution.py
 - [ ] dielectric.py
 - [ ] transport/collisions.py
 - [ ] transport/braginskii.py

We'll need to think some more about how to handle situations where the mean integer charge is a possible argument.  If any problems arise with the decorator or class in `parameters_cython.pyx`, please raise an issue since these are not tested in Cython.  This issue should also wait until after #340 is merged.  Working with `collisions.py` should also wait until after #336 is resolved, so that `@particle_input` can be used for arguments that accept tuples or lists of particles.  The `Particle` class was created in #239 and the `@particle_input` decorator was created in #242.

Thanks!
-Nick

namurphy:
This is also related to #309.

ritiek:
@ionwyn Would you like to give this a try? Feel free to ask about anything. :)

ionwyn:
@ritiek Sure!  Thanks for offering the task and help!

ionwyn:
[...]

PR #140, opened 2017-10-03
Consolidate separate functions for ions and electrons in plasma parameter calculations

RoberTnf:
Closing #106

The checklist of things to do are:

- [X]  Merge electron_thermal_speed and ion_thermal_speed into thermal_speed
- [x]  Merge electron_gyrofrequency and ion_gyrofrequency into gyrofrequency
- [x]  Merge electron_gyroradius and ion_gyroradius into gyroradius
- [x]  Merge electron_plasma_frequency and ion_plasma_frequency into plasma_frequency
- [x]  Merge electron_inertial_length and ion_inertial_length into inertial_length
- [x] add a parameter method to thermal_speed 
I will be updating as I progress with the tasklist.

Issue #106, opened 2017-09-09
Consolidate separate functions for ions and electrons in plasma parameter calculations


namurphy:
When I was working on the plasma parameter calculations, I put in separate functions for ions and electrons.  In retrospect, it would probably work better to have a single function that has the particle species as the input instead of separate ion/electron functions. 

The checklist of things to do are:
- [ ] Merge `electron_thermal_speed` and `ion_thermal_speed` into `thermal_speed`
- [ ] Merge `electron_gyrofrequency` and `ion_gyrofrequency` into `gyrofrequency`
- [ ] Merge `electron_gyroradius` and `ion_gyroradius` into `gyroradius`
- [ ] Merge `electron_plasma_frequency` and `ion_plasma_frequency` into `plasma_frequency`
- [ ] Merge `electron_inertial_length` and `ion_inertial_length` into `inertial_length`

A sample call would be:
```Python
>>> from astropy import units 
>>> thermal_speed(5*units.K, 'e')
```

With respect to `thermal_speed`, there is some ambiguity about whether it is the root mean square (rms) or most probable thermal speed, and between 1D and 3D (see [Wikipedia article on thermal speeds](https://en.wikipedia.org/wiki/Thermal_velocity)).  I'm wondering if the following subsequent change would be helpful:
- [ ] Change `thermal_speed` to `thermal_speed_rms` and `thermal_speed_mostprob` 
Probably we want to give it in 3D.  In any case, the documentation should specify what is being calculated.

jhillairet:
Shouldn't the gyrofrequency be signed ?  In the current implementation, ion and electron gyrofrequency has the same sign. Is it designed?

StanczakDominik:
That's a convention thing, I believe. Could do with an optional Boolean flag in the function's signature!

jhillairet:
I believe it's more than a convention thing. For example, the sign of the cyclotron frequency should affect the calculation of the cold plasma dielectric tensor. Here is a capture from the Stix book :
![image](https://user-images.githubusercontent.com/4642848/30514551-689384e8-9b17-11e7-85b4-024b03a7631a.png)

By the way, I was trying to make a cold tensor calculator using plasmapy, but I discovered than the Numpy vectorize decorator is scraping the unit of the quantities... So for instance, it remove the unit of the electron_gyrofrequency function output... Is anyone already ran into this problem?

namurphy:
[...]

PR #568, opened 2018-10-15
Update inertial_length

namurphy:
I fixed an error in the docstring of inertial_length which still used the definition from when it was split into ion_inertial_length and electron_inertial_length.  I took this chance to use `particle_input` and `Particle`, remove the default particle assumption, update tests, and update the description of the physics.

Cross-references: #106, #140, #341, #453

### DIFF
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -1166,8 +1166,7 @@ def inertial_length(n: u.m**-3, particle: atomic.Particle):
     Parameters
     ----------
     n : ~astropy.units.Quantity
-        Particle number density in units convertible to
-        m\ :superscript:`-3`\ .
+        Particle number density in units convertible to m ** -3.
 
     particle : str, optional
         Representation of the particle species (e.g., 'p+' for protons,
--- a/plasmapy/physics/parameters.py
+++ b/plasmapy/physics/parameters.py
@@ -1154,28 +1154,29 @@ def Debye_number(T_e: u.K, n_e: u.m**-3):
     return N_D.to(u.dimensionless_unscaled)
 
 
+
 @utils.check_quantity(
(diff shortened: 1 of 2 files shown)


plasmapy/physics/parameters.py
```diff
@@ -1154,28 +1154,29 @@ def Debye_number(T_e: u.K, n_e: u.m**-3):
...
+def inertial_length(n: u.m**-3, particle: atomic.Particle):
+    r"""
+    Calculate a charged particle's inertial length.
 
     Parameters
     ----------
     n : ~astropy.units.Quantity
-        Particle number density in units convertible to m**-3.
+        Particle number density in units convertible to
+        m\ :superscript:`-3`\ .
```
StanczakDominik: There's a trade off here between having our docs pretty and readable from sphinx (which this accomplishes beautifully) and from the embedded docs, Python help function, Ipython/Jupyter's beautiful `inertial_length?` magic... which this superscript change is going to make worse :( Do you think `m**-3` is not a reasonable compromise?

plasmapy/physics/parameters.py
```diff
@@ -1191,22 +1192,27 @@ def inertial_length(n: u.m**-3, particle='e-'):
...
     -----
     ~astropy.units.UnitsWarning
-        If units are not provided, SI units are assumed
+        If units are not provided and SI units are assumed.
 
     Notes
     -----
-    The particle inertial length is also known as an particle skin depth and is
-    given by:
+    The inertial length of a particle of species :math:`s` is given by
```
StanczakDominik: Likewise here, the `:math:` part doesn't add much to a single letter besides making it the single most beautiful *s* to ever get sent to Mathjax from a browser. I love it for complex expressions, though.

PR #569, opened 2018-10-16
Initial draft for AnalyticTwoFluidPlasma

### LINKED ISSUE #565: Create Plasma subclass for plasma described analytically in Cartesian coordinates with fluid approximation
Often plasma physicists needs to describe plasmas through analytical or user-provided functions.  We should create a [`Plasma`](https://github.com/PlasmaPy/PlasmaPy/blob/6d78d8c31e8e93c27b1d47a8fb0c29a91eaadad5/plasmapy/classes/plasma_factory.py#L22) subclass that allows users to provide functions that describe magnetic field, density, temperature, etc.  

In the interest of finite keeping pull requests, the first step will be to do this in Cartesian/rectangular coordinates and use the fluid approximation for a fully ionized, two component (ion-electron) plasma.  This functionality will also provide a chance to work with a `Plasma` subclass and make refinements to the abstract base class.

The next steps should be to expand this to cylindrical/toroidal coordinates (especially important for the fusion community) and spherical coordinates.

StanczakDominik:
Closes #565 .

This is much an initial draft based on what I could understand from @namurphy 's description (it's a neat way of handling it - I often find myself at a loss for how to handle analytic stuff and just do numerics to compensate).

It's my first time working with the two fluid description (apologies, but I have very little experience with anything that doesn't deal resolve particles!), so I'll probably have a bunch of question.

I was able to find a summary of the relevant equations [here (page 10)](https://ocw.mit.edu/courses/nuclear-engineering/22-611j-introduction-to-plasma-physics-i-fall-2003/lecture-notes/chap4.pdf).
As I understand it, we need a spatial dependence for our functions, as the velocity dependence from kinetic approaches is integrated out.

At this very early stage I have electron and ion densities and temperatures and a magnetic field - based on the PDF I also expect we want to be able to input external electric fields and mean velocities as functions of position. Is that correct?

I also had to change `abstractproperty` in the `BasePlasma` to `abstractmethod`, as you can't supply a position value to a property.

@ritiek, if you get some spare time, I'd really appreciate any input and comments you could potentially provide - I don't want to butcher this ABC too badly :laughing:

### DIFF
--- a/plasmapy/classes/plasma.py
+++ b/plasmapy/classes/plasma.py
@@ -1,0 +1,26 @@
+from plasmapy.classes.plasma_base import BasePlasma
+
+class Plasma3DSpatialOnly(BasePlasma):
+    def electron_temperature(self, r):
+        raise NotImplementedError
+    def ion_temperature(self, r):
+        raise NotImplementedError
+    def electron_density(self, r):
+        raise NotImplementedError
+    def ion_density(self, r):
+        raise NotImplementedError
+
+
+class AnalyticTwoFluidPlasma(Plasma3DSpatialOnly):
+    def __init__(self, electron_density,
+                     electron_temperature,
+                     ion_density,
+                     ion_temperature,
+                     magnetic_field,
+                     ):
+        self.electron_density = electron_density
+        self.electron_temperature = electron_temperature
+        self.ion_density = ion_density
+        self.ion_temperature = ion_temperature
+        self.magnetic_field = magnetic_field
+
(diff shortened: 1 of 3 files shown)


StanczakDominik:
Notes from telecon:
* Combine __call__ and __getitem__, use __call__ for interpolation/analytic values, __getitem__ for indexing on grids
* Create a class for plasma properties (not just an array) that would handle both indexing (via regular array stuff) and interpolation

StanczakDominik:
I'm going to do this in #675 instead.

PR #570, opened 2018-10-18
[WIP] Cleaner warning display in CircleCI

StanczakDominik:
I have applied BASH MAGIC to the doc builder, it'll save stdout and stderr to files - now I can:
* [x] hopefully display them separately as needed
* [x] check if warnings are empty, if they are, return 1 and fail the build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,17 +1,3 @@
-deps-run: &deps-install
-  name: Install Python dependencies and PlasmaPy
-  command: |
-    pip install --user -r requirements/automated-documentation-tests.txt | cat
-    pip install --user . | cat
-
-html-run: &doc-html
-  name: Build HTML documentation
-  command: python setup.py build_docs -w
-
-latex-run: &doc-latex
-  name: Build LaTeX documentation
-  command: python setup.py build_docs -b latex -w
-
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Phew! I got it working. It should now print stdout and stderr in separate tabs, so stderr can be scanned for errors way more easily.

I also updated astropy-helpers to 3.0.2 to remove a bunch of Sphinx DeprecationWarnings and the like.

.circleci/config.yml
```diff
@@ -1,32 +1,34 @@
...
+      - run: 
+          name: Install Python dependencies and PlasmaPy
+          command: |
+            pip install --user -r requirements/automated-documentation-tests.txt | cat
+            pip install --user . | cat
+      - run:
+          name: Build HTML documentation
+          shell: /bin/bash
+          command: |
+            python setup.py build_docs -W 2> >(tee /tmp/stderr) 1> >(tee /tmp/stdout) || true; echo $? | tee /tmp/EXITCODE
```
namurphy: Nice!

StanczakDominik: Thanks, I'm rather proud of this little random-walk-spawned monstrosity ^^

PR #571, opened 2018-10-20
Fix check_quantity docs

StanczakDominik:
Issue from my personal backlog list. That doc is [pretty borked right now](http://docs.plasmapy.org/en/latest/api/plasmapy.utils.checks.check_quantity.html?highlight=quantity).

Includes commits from #570 because this is a good test bed for those.


StanczakDominik:
This is failing right now because of reasons... and I'm not quite sure what those are. But I think if we merge #561 first, it's gonna be fine. I'll try to figure out what's causing this Travis failure in the meantime.

plasmapy/utils/checks.py
```diff
@@ -102,6 +104,13 @@ def check_quantity(**validations):
         are provided, a `UnitsWarning` will be raised and the inputted
         units will be assumed.
 
+    Notes
+    -----
+    This functionality may end up becoming deprecated due to 
+    noncompliance with the `IEEE 754 standard
+    <https://en.wikipedia.org/wiki/IEEE_754#Exception_handling>`_
+    and in favor of `~astropy.units.quantity_input`.
```
namurphy: It'd be helpful to say the reason why this is in noncompliance with IEEE 754, either in the docstring or in a comment in the function itself.

StanczakDominik:
Gonna raise a separate issue for that discussion as time permits!

namurphy:
In that case, merge away! :+1:

PR #574, opened 2018-10-25
Debug PR for CircleCI

StanczakDominik:
I may have accidentally broken CircleCI. Debugging right now.


StanczakDominik:
Okay, I no longer think this is broken.

I [migrated a bunch of our services](https://github.com/PlasmaPy/PlasmaPy/settings/installations) to apps as [services are being deprecated](https://developer.github.com/changes/2018-04-25-github-services-deprecation/). We will still need to swap out RTD once that has transitioned to the new style of integrations.

I also swapped the required-to-pass CircleCI test to the new workflow layout instead of the older one, which wasn't being launched and thus never delivered results. It was, however, set to required, so GitHub kept waiting for those results. That's no longer the case.

PR #275, opened 2018-02-23
Improve docstrings and readability of the atomic subpackage

namurphy:
I'm going through and updating atomic docstrings so that they are consistent with the defaults for [pydocstyle](http://www.pydocstyle.org/en/latest/) in `setup.cfg` created in #272 (see also #273).  There should be relatively few substantive changes, besides editing the content of various docstrings.  

Closes #104 since it improves the readability of `_Isotopes` in `isotopes.py`, and makes it so that it is not really necessary to convert it to json or HDF5.  See the commit log for more information.

All in all, this is a pretty boring pull request. :smile_cat:

Issue #104, opened 2017-09-08
Improve readability of elements.py and isotopes.py


namurphy:
At present, the `atomic` subpackage stores data for elements and isotopes as dictionaries that are coded into `elements.py` and `isotopes.py`.  I did this as a way to get the functionality done quickly, but this does not strike me as the best way to go about this.  I am wondering if it would be better to store the atomic data in HDF5 form and then to use `pytables` or `h5py` to access the information.  It would also be helpful to include more information on the provenance of the data (i.e., pointing out which databases the information came from).  During this process, the functions within `atomic.py` will need to be revised as well.

I'll assign myself to this since I still have the methods to create `elements.py` and `isotopes.py` somewhere (I hope!).

antoinelpp:
Hi,
Putting data outside of the code seams good to me, however, `HDF5` seems a bit too much to me... I would consider `JSon` for informations like mass, charge, Inonization energy, and so on.
The only data that may require `HDF5` would be the cross sections, but I'm not sure it is worth it.

Json is very similar to a dictionary, but the library [JsonPickle](https://jsonpickle.github.io/) allow to load and dump python objects directly.

StanczakDominik:
Antoine does have a point. It might be overkill, given that you're not
really doing any writing during usage and the data you're reading is sparse
(by which I mean, not arrays but maybe a few numbers with different
conceptual meaning per element or isotope). You don't really need lighting
fast data access for this one. I kind of like the json idea - but in all
honesty, given we already kind of have a dictionary for this one... I'm not
one hundred percent sure this needs changing.

12.09.2017 10:21 "antoinelpp" <notifications@github.com> napisał(a):

namurphy:
All good points!  JSON seems like a good choice if we decide to do this, so I updated the title.  Since it's in an adequate format now, we can wait to do this until later.

Issue #18, opened 2017-04-17
Need a way to store atomic data


namurphy:
We are frequently going to need to reference basic atomic data such as the atomic number and atomic mass, including for different ionization states and different isotopes.  We will need this information for numerous applications, including as basic as getting the Alfven speed from the number density and the ion species.  There will also be situations that require significantly more atomic data (e.g., calculating radiative cooling rates or ionization/recombination rates as a function of temperature/energy) so our approach should be able to be extended.  

There do exist atomic databases with Python interfaces including [ChiantiPy](http://chiantipy.sourceforge.net/) and [pyatomdb](http://atomdb.readthedocs.io/en/master/), though the latter is in a pre-alpha state of development.  We could look into how these databases store both basic and more detailed information, and perhaps using one of them as an optional dependency.

My questions are:
- How should we store atomic data?
- What atomic data should we store?
- How should we access stored atomic data?  (e.g., what API should we use?)

namurphy:
See pull request: https://github.com/PlasmaPy/PlasmaPy/pull/24

PR #591, opened 2018-12-29
Put isotope data into JSON

StanczakDominik:
Continues [a discussion we've had on Matrix](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/$1546027415314043ntuti:matrix.org?via=matrix.org&via=openastronomy.org&via=cadair.com&via=t2bot.io). I dumped everything into json like this:
```python
In [41]: def convert(x): 
    ...:     if x.unit == u.u: 
    ...:         return x.value 
    ...:     else: 
    ...:         return x.si.value 
    ...:           
In [42]: with open("isotopes.json", "w") as f: 
    ...:     json.dump(isotopes._Isotopes, f, default=convert, indent=2) 
```

And, given that the diff is large, this is how I was able to recreate the dict with units:

```python
"""
Create a dictionary containing basic information for isotopes and
neutrons.
"""

import astropy.units as u
import json
import os

with open(os.path.join(os.path.dirname(__file__), "isotopes.json")) as f:
    _Isotopes = json.load(f)

for isotope in _Isotopes:         # I'm kind of suspecting that something smarter than a for loop could be done here
    _Isotopes[isotope]["mass"] *= u.u
    if "half-life" in _Isotopes[isotope] and not isinstance(
            _Isotopes[isotope]["half-life"], str):
        _Isotopes[isotope]["half-life"] *= u.s
````

As a check, with the old version imported and the new version in local scope:
```python
In [40]: isotopes._Isotopes == _Isotopes                                                                                                                           
Out[40]: True
```


namurphy:
Putting the isotopes information into json format is a worthwhile idea.  It would probably make sense to do the same for `elements.py` for consistency.  Another possibility would be to create a json file containing information for particles and antiparticles, though this would be a little more involved since `special_particles.py` contains code that sets some of the particle attributes, so it could be relegated to another issue.  

Related issues: #18, #104, #275

StanczakDominik:
By the way, before I forget - I realized that this will break pip installs due to `wheels`, and we'll probably need to update MANIFEST or some such black packaging magic and change up the `__file__` call for this to work.

StanczakDominik:
* [ ] add note on adding manifest file to development guide

StanczakDominik:
Note to self that vn-ki on Riot has a better implementation of the JSON parser.

PR #597, opened 2019-01-04
Make README more beautiful by adding logo

StanczakDominik:
Self explanatory. Picked up the idea from [CuPy](https://github.com/cupy/cupy/blob/master/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<div align="center"><img src="https://raw.githubusercontent.com/PlasmaPy/PlasmaPy-logo/master/exports/with-text-dark.png" width="600"/></div>
+
 # PlasmaPy
 
 [![PyPI version](https://badge.fury.io/py/plasmapy.svg)](https://badge.fury.io/py/plasmapy)


StanczakDominik:
Bump because after what happened earlier I don't want to abuse my admin privileges... which I seem to have disabled, so I can't even abuse them now. *Nice!*

README.md
```diff
@@ -1,3 +1,5 @@
+<div align="center"><img src="https://raw.githubusercontent.com/PlasmaPy/PlasmaPy-logo/master/exports/with-text-dark.png" width="600"/></div>
```
namurphy: If I am reading this correctly, this sets a fixed width for the logo.  I'm wondering how this will appear on mobile devices if the image width is greater than the screen width.  This isn't a big deal and is easy to change, so I'm happy to approve this now.

StanczakDominik: screenshot_20190123-200118_firefox I think it's gonna work all right :)

PR #602, opened 2019-02-02
Remove message keyword argument in pytest.raises and pytest.warns

### LINKED ISSUE #601: Pytest deprecation of "message" parameter to raises
Followup to #599. https://github.com/pytest-dev/pytest/issues/3974 is a thing, thus we need to do something like https://github.com/PlasmaPy/PlasmaPy/pull/600/commits/e9cd7c593ed1e766993d3daaf7994e046a768242 except without blowing all of pytest-helpers up.

namurphy:
The message keyword argument is being deprecated from pytest.raises and may be removed instead of renamed.  This PR removes the use of message in our tests and checks.  Closes #601.

### DIFF
--- a/plasmapy/atomic/tests/test_atomic.py
+++ b/plasmapy/atomic/tests/test_atomic.py
@@ -634,12 +634,12 @@ def test_half_life_u_220():
 
     with pytest.raises(MissingAtomicDataError):
         half_life(isotope_without_half_life_data)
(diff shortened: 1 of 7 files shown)


namurphy:
Feels good to make a pull request again for the first time in a while!  I've been too busy writing papers for the plasma physics decadal review in the US.

PR #605, opened 2019-02-19
Test class methods and attributes

namurphy:
The purpose of this PR is to create a pytest helper function that can test class methods and attributes.  The arguments to this function will be the class to be tested, arguments and keyword arguments used for creating a particular instance of that class, and a dictionary with keys that are strings representing class methods and attributes and values that are the expected outcome (including exceptions and warnings).

The code will primarily come from [an existing test](https://github.com/PlasmaPy/PlasmaPy/blob/a709aa7227f074341e8be33da12110e3688d2144/plasmapy/atomic/tests/test_particle_class.py#L410) of the `Particle` class.  This test of the `Particle` class has the potential to be much more general, and as such is worth separating out as for `run_test` and `run_test_equivalent_calls`.


namurphy:
A few thoughts that I want to put down on pen & paper (or on silicon wafers & electrons) while I'm still thinking about how to implement this.

A test that checks class attribute references will have considerable overlap with the existing `run_test` function which checks that a function or callable will produce the expected result (including exceptions and warnings), so I am thinking that it would be best to turn parts of `run_test` into smaller functions.

One particularly helpful utility would be a general purpose function that checks whether or not two objects are either equal or close enough to equal given a tolerance.  Of particular importance will be special handling of arrays and `astropy.Quantity` objects with careful checks that units match, and we'd also need to make sure that `numpy.nan` is equal enough to `numpy.nan` and `numpy.inf` is equal enough to `numpy.inf`.  We might want to call this function something like `close_enough`, unless we can think of something more descriptive.  This may be more generally useful so `pytest_helpers.py` might not be the best place to put this.

Another helpful function would be something that takes in a call string and uses `eval` to check that the expected outcome results.  Again, this would require handling of exceptions and warnings, including cases where a warning is issued and we want to check the result too.  We could call this something like `check_call`, and it could perhaps return a class with attributes to store the appropriate exception to be raised and the appropriate error message to be issued.  The higher level function could then be what raises the exception, since there may be some cases where multiple calls are checked in a single test.  This class should evaluate to `False` if there are no exceptions and `True` if there are exceptions.

After moving this functionality from `run_test`, then it would be reasonably straightforward to create a different function to check class attribute references without having to copy and paste a bunch of code.

Another thing I'd like to avoid is having to instantiate a class with the same positional and keyword arguments more than once.  This could slow down tests noticeably if we have classes that take a while to instantiate.  This means we would have to be careful about how we parametrize things.  It would probably be helpful to skip tests that fail because a particular instance could not be instantiated, as long as there is one test failure that shows what happened.  We could use the `pytest_generate_tests` hook to customize our parametrization further too.

namurphy:
Here's a paragraph related to this PR that I wrote for a proposal but tentatively decided to cut for space.  But I had fun writing this paragraph so here it is:

> Many of PlasmaPy’s tests have a simple premise. We check that providing a function or method with a set of positional and keyword arguments leads to the expected outcome. The situation becomes more complex with the realization that the expected outcome may be a string, float, array, or quantity with units. We may expect a specific exception to be raised or a warning to be issued. We may be expecting an infinity or NaN that may or may not have units, and have to take into account that `numpy.inf == numpy.inf` and `numpy.nan == numpy.nan` both evaluate to `False`. We might want to check if values or elements of an array are approximately equal to within a specified tolerance. In all of these cases, we need descriptive error messages that can help new contributors quickly pinpoint what the problem is. Accounting for all of these situations leads to repeated boilerplate code that makes tests harder to write, read, and maintain. This in turn creates an unnecessary barrier to entry for new contributors, especially as unit testing and continuous integration testing are not widely used in much of the plasma physics community.

StanczakDominik:
While that last one is about right, sometimes I wonder - hypothetically - whether maybe it might actually be preferable to have simpler tests without the simplification of pytest_helpers and at the expense of duplicating test code - this would let people not feel like they have to learn another testing framework to contribute... Maybe we should make a note in CONTRIBUTING that it's nice to have sometimes but feel free to do whatever tests you like so long as they're clean and documented.

namurphy:
I think I'd like to take a different approach than this (mostly as described in the comments above).

PR #608, opened 2019-03-12
updated the cold plasma example to match latest API changes

jhillairet:
As the functions `cold_plasma_permittivity_SDP` and `cold_plasma_permittivity_LRP` moved from the `physics` to `physics.dielectric`

### DIFF
--- a/plasmapy/examples/plot_cold_plasma_tensor_elements.py
+++ b/plasmapy/examples/plot_cold_plasma_tensor_elements.py
@@ -9,7 +9,7 @@ elements for various electromagnetic wave frequencies.
 import numpy as np
 import matplotlib.pyplot as plt
 from astropy import units as u
-from plasmapy.physics.dielectric import (cold_plasma_permittivity_SDP, 
+from plasmapy.physics.dielectric import (cold_plasma_permittivity_SDP,
                                          cold_plasma_permittivity_LRP)
 
 #######################################################################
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Hmm, I'm confused. https://github.com/PlasmaPy/PlasmaPy/blob/master/plasmapy/physics/__init__.py#L16 seems to imply that these should be visible from  the physics namespace - and the example at http://docs.plasmapy.org/en/latest/auto_examples/plot_cold_plasma_tensor_elements.html seems to be building as well. So it should be working as stated in the example - it's just accessible from two separate namespaces (regardless of whether that's actually good design!). Did you have problems with calling them that way?

I do think the way of calling these functions is way cleaner, though. :+1:

jhillairet:
Well, I was confused either. With the latest version I could not even see the physics submodule from the root namespace (?), but I managed using a `from` import as in the proposed example...

StanczakDominik:
Oh darn, I think I get it - that you did not see `plasmapy.physics` in the root namespace is probably a side effect of #561. Then again, not loading those modules by default accomplishes the goal of having dependencies be optional, so in a way, it's intended behavior (though it could be up for discussion!).

The namespace (and especially `help(plasmapy)`) could use some sprucing up to make this clearer - but that's another issue, in which I'll reference this one in a minute).

I'll merge this for now, as it both cleans the example up and makes it compatible with current master! Thanks for reporting this as a bug :)

PR #615, opened 2019-04-09
Reshuffle subpackages (#576 followup)

### LINKED ISSUE #576: Reshuffle subpackages (plasmapy.theory)
From the telecon today and based on stuff that came up in #561, this is something that probably deserves wider discussion.

* [x] move `plasmapy.math`, `plasmapy.physics` and `plasmapy.transport` into a common `plasmapy.theory.*` subpackage (so we'd have `plasmapy.theory.math` etc)
* [x] possibly rename `physics` to `parameters` along the way
* [ ] fix all the import errors that this causes
* [ ] possibly fix docs
* [ ] possibly split up optional dependency pip install options again (see #561)

While an implementation is pretty easy and thus a good first contribution, whether we actually do it should wait until consensus is reached here.

StanczakDominik:
Closes #576


StanczakDominik:
@pep8speaks pep8ify

StanczakDominik:
@namurphy I forget if we decided we'd call the subpackage `theory` or `parameters` :( Probably the latter, right?

namurphy:
> @namurphy I forget if we decided we'd call the subpackage `theory` or `parameters` :( Probably the latter, right?

Hm...I don't think it was `plasmapy.the_latter`.  It was `plasmapy.parameters` :cat:

StanczakDominik:
Obsoleted once again by #692

PR #618, opened 2019-04-30
Update development roadmap

### LINKED ISSUE #426: Update development roadmap
ROADMAP.md is out of date. We should mentioned what has been included in v0.1 and update what we expect for the future based on our milestones.

namurphy:
This PR is to update `ROADMAP.md` (which hadn't been updated in over a year) in advance of our version 0.2.0 release.  I think it'll be best to avoid specifying which changes are slated to go into each release since that changes a lot and can be tracked better with the projects and milestone features on GitHub.  Plus, given how frequently we update the roadmap, we don't want to make changes that'll go out of date more obviously.  So far, I'm using some of what I wrote for our recent [NSF CSSI proposal for PlasmaPy](https://zenodo.org/record/2633287).

One possibility for the future would be to make a separate document for our roadmap that we can archive on Zenodo as a versioned document.  This would make the history of our roadmap more transparent.  Our roadmap would also be citable, which would mean that we could refer to it directly in things like grant proposals.

Closes #426.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -51,19 +51,19 @@ ABC.
 
 ### Experimental support
 
-
 ### Plasma simulation capabilities
 
+### Turbulence analysis tools
 
-### Plasma parameters
+### Dispersion relation solver
 
-- Dimensionless numbers functionality
+### Plasma parameters and transport coefficients
 
-- Code and documentation infrastructure
-- Plasma parameter calculations
-- Transport coefficient functions
-   - Braginskii theory
-- Base PlasmaBlob class
+PlasmaPy currently has numerous functions and classes that compute
+plasma parameters and transport coefficients.  Plasma physics is a
+very broad discipline, so there remain many plasma parameter functions
+that have not yet been implemented.  We will continue to expand and
+refactor this functionality.
 
 
 ### Features to be included in Version 0.2 release
@@ -87,8 +87,3 @@ ABC.
 - Analytical plasma physics tools (with SymPy)
 
 
-### Future events
-
-- Code development meeting
-- Software Carpentry workshops at plasma physics conferences
-- Python in Plasma Physics conference?
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,24 +1,72 @@
 # PlasmaPy Development Roadmap
 
-This document summarizes the overall development plans for PlasmaPy.
-This roadmap is very fluid and is subject to change due to community
-priorities.  For the most up-to-date information, see PlasmaPy's
-[issues](https://github.com/PlasmaPy/plasmapy/issues), [pull
-requests](https://github.com/PlasmaPy/plasmapy/pulls), and
-[projects](https://github.com/PlasmaPy/PlasmaPy/projects).
+This document summarizes the long-term development plans for PlasmaPy.
+This roadmap is very fluid and will continue to be updated and changed
+throughout the development process based on community input.  
 
-## Features to be included in Version 0.1 release
+If you have feature requests, we highly encourage you to [raise an
+issue](https://github.com/PlasmaPy/PlasmaPy/issues/new) in [PlasmaPy's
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -1,24 +1,72 @@
 # PlasmaPy Development Roadmap
 
-This document summarizes the overall development plans for PlasmaPy.
-This roadmap is very fluid and is subject to change due to community
-priorities.  For the most up-to-date information, see PlasmaPy's
-[issues](https://github.com/PlasmaPy/plasmapy/issues), [pull
-requests](https://github.com/PlasmaPy/plasmapy/pulls), and
-[projects](https://github.com/PlasmaPy/PlasmaPy/projects).
+This document summarizes the long-term development plans for PlasmaPy.
+This roadmap is very fluid and will continue to be updated and changed
```
StanczakDominik: So it's a rather collisional roadmap, then?

ROADMAP.md
```diff
@@ -1,24 +1,72 @@
...
+abstract methods that have to be defined for each of the subclasses
+for different plasma representations (thus ensuring a more consistent
+API).  Users can then invoke the `Plasma` class factory to select and
+instantiate the appropriate subclass based on inputs provided by the
+user.  The framework for base data structures was implemented by
+Google Summer of Code student Ritiek Malhotra in 2018, along with the
+[OpenPMD](https://github.com/openPMD/openPMD-standard) subclass.  The
+next major development tasks are to create subclasses for different
+plasma representations and to expand/refine the API defined by the
+ABC.
```
StanczakDominik: Reading this now, an idea comes to mind: perhaps we could say that for 0.3 we want to have an API design document for the `Plasma`. I don't think the ABC we have now really defines what we want to have there - it's the start of an implementation, sure, but since it's missing a lot of moving parts and access points, it doesn't explicitly say how we want that to behave.

Perhaps we could make discussing this the main point of a telecon sometime?

ROADMAP.md
```diff
@@ -1,24 +1,72 @@
...
 - Code and documentation infrastructure
 - Plasma parameter calculations
 - Transport coefficient functions
    - Braginskii theory
 - Base PlasmaBlob class
-- Dimensionless numbers functionality
 
-## Features to be included in Version 0.2 release
+
+### Features to be included in Version 0.2 release
```
StanczakDominik: I appreciate the formatting change, but none of these seem likely to go in by 0.2 :laughing:

namurphy: Nope!  That's old text that I was planning on deleting anyway.

ROADMAP.md
namurphy: I was very tempted to say `This roadmap is very plasma` instead but decided against it.

StanczakDominik:
Ping! How can I help with this one? :)

namurphy:
Hm...probably the next things to do would be to keep incorporating text from the [proposal](http://doi.org/10.5281/zenodo.2633287) into the empty sections.  Since we're doing that we should put a link to the proposal and CC BY 4.0 license at the bottom.  Feel free to keep working on that!  I'm pretty swamped these days, unfortunately... 😞   And thanks!

StanczakDominik:
Gah, I'm finding it really hard to motivate myself to work on this one... 0.3.0 or followup soon after 0.2.0?

namurphy:
My vote would be to remove `ROADMAP.md` from the main repository and put it somewhere else so that we can version it separately and make it citable.  Would that work for you?

I have a good idea of what to put in it, but I'm pretty swamped right now and will need to prioritize writing a proposal and DPP-CPP white papers over the next month.  Sigh...

StanczakDominik:
Sure, that's reasonable. Do you think you could take a minute to do that so we could close this PR? I'd love to have all three PRs ready for tomorrow and keep to a deadline for once! :smile:

PR #619, opened 2019-05-08
Presevering function hints when decorated by `check_quantity` and `check_relativistic`

rocco8773:
Decorators in Python 3.4+ do preserve signatures of their wrapped functions, but they still loose the ability to give parameter hints in IDEs.

For example, in PyCharm, the current `master` branch would yield parameter hints for `plasmapy.physics.parameters.Alfven_speed()` as:

Screen Shot 2019-05-07 at 7 44 17 PM

But after defining the `__signature__` method for the wrapper function, the parameter hint now shows the arguments as:

Screen Shot 2019-05-07 at 7 47 11 PM

The `__signature__` dunder has existed since Python 3.3 and is well documented in [PEP 362](https://www.python.org/dev/peps/pep-0362/).

### DIFF
--- a/plasmapy/atomic/particle_input.py
+++ b/plasmapy/atomic/particle_input.py
@@ -233,7 +233,11 @@ def particle_input(wrapped_function: Callable = None,
     def decorator(wrapped_function: Callable):
         wrapped_signature = inspect.signature(wrapped_function)
 
-        @functools.wraps(wrapped_function)
+        # add '__signature__' to methods that are copied from
+        # wrapped_function onto wrapper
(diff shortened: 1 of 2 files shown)


rocco8773:
With all my research, I could not find how PyCharm or IPython actually produced parameters hints.  In the PyCharm editor, I think the whole scipt is being inspected and, thus, the hints are still present.  Whereas, in the console its only looking at the top level of `__wrapped__`, if `__signature__` does not exist.  Similar to performing `inspect.signature(func, follow_wrapped=False)`. ¯\\_(ツ)_/¯

In any case, it is helpful! @StanczakDominik , thanks for the merge.

PR #625, opened 2019-05-20
Improve mathematics narrative docs

### LINKED ISSUE #296: Add narrative documentation for mathematics subpackage
The `mathematics` subpackage is relatively small in comparison to our other subpackages, but it would still be helpful to have some narrative documentation in our `docs` directory to describe what it is for and what functionality it includes.  

We do not necessarily need to complete this for our v0.1.0 release since the functions have helpful docstrings already, but it would be helpful to have for our v0.2.0 release.

Thanks!
-Nick

StanczakDominik:
Closes #296 

When I say improves... There's little I could find to say about a subpackage with three functions that isn't said in their docstring.  This is probably one more reason for #576.

I'm out of ideas for mathematics, but this can be a start.

I cut the wofz rename to the Faddeeva function from the namespace to highlight the fact that it's not our function.

### DIFF
--- a/docs/mathematics/index.rst
+++ b/docs/mathematics/index.rst
@@ -4,8 +4,6 @@
 Mathematics (`plasmapy.mathematics`)
 ************************************
 
-`plasmapy.mathematics` gathers general mathematical formulae used elsewhere in the code.
-
 .. note::
    As PlasmaPy is currently in very early stages of development, APIs for most
    functionality should be considered unstable. Specifically, `plasmapy.mathematics`
--- a/docs/mathematics/index.rst
+++ b/docs/mathematics/index.rst
@@ -1,8 +1,17 @@
+.. py:module:: plasmapy.mathematics
+
 .. _mathematics:
 
 ************************************
 Mathematics (`plasmapy.mathematics`)
 ************************************
+
+`plasmapy.mathematics` gathers general mathematical formulae used elsewhere in the code.
+
+.. NOTE
+   As PlasmaPy is currently in very early stages of development, APIs for most
+   functionality should be considered unstable. Specifically, `plasmapy.mathematics`
+   may move to a subset of `plasmapy.theory`.
 
 .. automodapi:: plasmapy.mathematics
    :no-heading:
--- a/docs/mathematics/index.rst
+++ b/docs/mathematics/index.rst
@@ -1,5 +1,3 @@
-.. py:module:: plasmapy.mathematics
-
 .. _mathematics:
 
 ************************************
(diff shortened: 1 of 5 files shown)


StanczakDominik:
Welp, at this point this does literally nothing but add a note.

Still, I think if we're going to go with #576 (and given that this subpackage has a single file with three functions in it, we probably should), it might be fine to merge-as-in-git this for now and merge-as-in-common-namespace these in the future.

plasmapy/mathematics/mathematics.py
```diff
@@ -91,7 +91,8 @@ def plasma_dispersion_func(zeta: Union[complex, int, float, np.ndarray, u.Quanti
         raise ValueError("The argument to plasma_dispersion_function is "
                          "not finite.")
 
-    Z = 1j * np.sqrt(np.pi) * Faddeeva_function(zeta)
+    # wofz is the so-called Faddeeva function
+    Z = 1j * np.sqrt(np.pi) * scipy.special.wofz(zeta)
```
namurphy: I'm a little hesitant about this change.  The root of the problem is that `wofz` is a bad name for a function since it gives very little indication of what it does except to people who are already familiar with the notation.  It took me a while to even realize that it means "w of z".  One of the best practices that I've found in a few places is to write the code so that it says how it work rather than relying on comments.  Maybe a compromise would be to change `Faddeeva_function` in the original to `scipy_fadeeva_function` so that it shows that it is from SciPy and also provides the name?

StanczakDominik: I just realized http://docs.plasmapy.org/en/master/mathematics/index.html doesn't get `Fadeeva` in the namespace because we're importing from `mathematics/__init__.py`, not `mathematics.py`, so this change is completely useless. Best to skip it.

docs/mathematics/index.rst
```diff
@@ -4,6 +4,11 @@
 Mathematics (`plasmapy.mathematics`)
 ************************************
 
+.. note::
+   As PlasmaPy is currently in very early stages of development, APIs for most
+   functionality should be considered unstable. Specifically, `plasmapy.mathematics`
+   may move to a subset of `plasmapy.theory`.
```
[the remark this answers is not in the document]
StanczakDominik: That's a fair point. Might as well skip the change, then - warnings about API stability are global at this point.

StanczakDominik:
Given that both of the I-thought-reasonable changes turn out to not be so, it's probably best to abandon ship for this PR!

StanczakDominik:
Still, not to let this go unspoken - thanks for the review @namurphy :smile:

Issue #622, opened 2019-05-14
Create a CodeMeta instance file


namurphy:
The [CodeMeta Project](https://codemeta.github.io/) is an effort to standardize metadata for different software projects that I learned about recently from the head librarian where I work.  As I understand it, we create a JSON file in our top-level directory using the tools provided on the website.  This file will contain the metadata describing PlasmaPy.  The CodeMeta standard is changing and our metadata will keep changing too, so we'll probably need to include updates to this in our release guide.

I'll tentatively assign myself to this since I can ask the librarian here for help, but the resulting file should be double checked by several of us.  We can schedule this for 0.2 since it shouldn't be too much extra effort and metadata is important.

owlice:
This example might help get you started: https://ascl.net/1503.001/codemeta.json

You can add "/codemeta.json" to any ASCL entry URL. This is still in beta but has already been through several rounds of vetting; we'll roll it out later this year.

StanczakDominik:
Given the existence and proximity of deadlines, I think this can be safely bumped up to 0.3.0.

PR #628, opened 2019-05-30
Add file with metadata for Zenodo upload

namurphy:
The purpose of the `.zenodo.json` file is to store the metadata needed by Zenodo for when we upload releases.  The metadata fields for a Zenodo upload should be automatically populated with the metadata spelled out in this file.  

Unfortunately, if we create a codemeta file (#622), a lot of the information will probably be duplicated so we may end up having inconsistencies.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,177 @@
+{
+  "creators": [
+    {
+      "name": "PlasmaPy Community"
+    },
+    {
+      "name": "Murphy, Nicholas A.",
+      "affiliation": "Center for Astrophysics | Harvard & Smithsonian",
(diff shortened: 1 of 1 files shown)


namurphy:
Looks like there are some open issues about how `codemeta.json` could supersede `.zenodo.json` (https://github.com/zenodo/zenodo/issues/1504, https://github.com/codemeta/codemeta/issues/213).  That would be great, especially given the shortage of documentation for `.zenodo.json` files.

namurphy:
Ah...it appears that [the `.zenodo.json` feature is not officially supported](https://gitter.im/zenodo/zenodo/archives/2016/04/21), so we may end up deleting this in favor of a CodeMeta file.

PR #629, opened 2019-05-31
Metadata updates prior to version 0.2.0

namurphy:
I updated some of the information about PlasmaPy in `setup.cfg` and the information on how to cite PlasmaPy (in particular by saying that users should cite both the specific version of PlasmaPy, as well as our standard informational reference from last year).  

I also changed `plasmapy.__citation__` to be a list of DOI links, partially because the metadata for a Zenodo upload has the potential to change, and DOIs are the most robust identifiers.  Another possibility would be to include a link to the directions in our documentation, though this approach might be prone to bit rot when releases become old in case we ever reorganize our docs.

### DIFF
--- a/CITATION.rst
+++ b/CITATION.rst
@@ -2,41 +2,40 @@ Acknowledging and Citing PlasmaPy
 =================================
 
 If you use PlasmaPy for a project resulting in a publication, we ask
-that you cite the following reference (`BibTeX
-<https://zenodo.org/record/1238132/export/hx#.WvMkQK0cChc>`_):
+that you cite both the specific version of PlasmaPy used in your
+project and an informational reference.  Including these references
+provides credit to developers of PlasmaPy and enables greater
+scientific reproducibility.
+
+Version 0.2.0 of PlasmaPy may be cited using the following reference.
+
+* PlasmaPy Community et al. (2019). *PlasmaPy version 0.2.0*, Zenodo,
+  https://doi.org/10.5281/zenodo.3235817
+
+The current standard informational reference for PlasmaPy is
 
 * PlasmaPy Community, Nicholas A. Murphy, Andrew J. Leonard, Dominik
   Stańczak, Pawel M. Kozlowski, Samuel J. Langendorf, Colby C. Haggerty,
   Jasper P. Beckers, Stuart J. Mumford, Tulasi N. Parashar, and Yi-Min
-  Huang. (2018, April). *PlasmaPy: an open source community-developed
-  Python package for plasma physics.* Zenodo.
-  http://doi.org/10.5281/zenodo.1238132
+  Huang. (2018). *PlasmaPy: an open source community-developed
+  Python package for plasma physics*, Zenodo,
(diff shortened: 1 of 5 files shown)


namurphy:
If I were to access the personnel record of a certain Star Fleet android, would I be looking up the...wait for it...metaData?

namurphy:
Incidentally, it feels a little strange to have just "force-pushed" a rebase after referencing Star Trek, not Star Wars.

CITATION.rst
```diff
@@ -2,41 +2,40 @@ Acknowledging and Citing PlasmaPy
...
-<https://zenodo.org/record/1238132/export/hx#.WvMkQK0cChc>`_):
+that you cite both the specific version of PlasmaPy used in your
+project and an informational reference.  Including these references
+provides credit to developers of PlasmaPy and enables greater
+scientific reproducibility.
+
+Version 0.2.0 of PlasmaPy may be cited using the following reference.
+
+* PlasmaPy Community et al. (2019). *PlasmaPy version 0.2.0*, Zenodo,
+  https://doi.org/10.5281/zenodo.3235817
```
StanczakDominik: I guess thi-

> I reserved the DOI for version 0.2.0 on Zenodo, so the links to that
will not work yet.

this answers my question! Good thinking.

plasmapy/__init__.py
```diff
@@ -77,22 +78,8 @@ def online_help(query):
...
-  month        = apr,
-  year         = 2018,
-  doi          = {10.5281/zenodo.1238132},
-  url          = {https://doi.org/10.5281/zenodo.1238132}
-}"""
+
+__citation__ = [
+    "https://doi.org/10.5281/zenodo.1238132",
+    "https://doi.org/10.5281/zenodo.3235817",
+]
```
StanczakDominik: I'm of two minds about this one - on the one hand, this is the first usage of `__citation__` I've seen that isn't basically a BibTeX entry. On the other, this provides easy (if multi-step) access to other citation methods here.

I guess it's fine, but I'll just note how much I'd love it to be able to do something like the following

```python
for imported_package in ag("~/scientific_software", "import (.*)"):
   try:
       print(imported_package.__citation__)
   except AttributeError:
       pass
```

for future publications. But that's way out of scope of this issue :smiley_cat:

setup.cfg
```diff
@@ -19,7 +19,7 @@ math_requires = mpmath
 physics_requires = lmfit
 
 # version should be PEP386 compatible (http://www.python.org/dev/peps/pep-0386)
-version = 0.1.1.dev
+version = 0.2.0.dev
```
StanczakDominik: Thanks for remembering about this one! Wouldn't want to miss it later today :)

plasmapy/__init__.py
namurphy: I was looking this up recently if there were any standards and came across https://github.com/adrn/CitationPEP/issues/2 which suggests creating a type for citations that can handle the different citation possibilities.  The loop you suggested could also probably be done if CodeMeta files become standardized and widely used (#622).

PR #647, opened 2019-06-15
Move "check" decorators to the plasmapy.utils.decorators package

rocco8773:
* Move the `check.py` module, and associated tests, from `plasmapy.utils` to `plasmapy.utils.decorators`.
* clean imports of modules that import "check" decorators


rocco8773:
I hit a lot of files here while checking and cleaning imports.  If this is not desired, at all or in this PR, then I can easily revert the commits.

plasmapy/utils/decorators/tests/test_checks.py
```diff
@@ -217,7 +217,7 @@ def func(x, y, z=10 * u.eV):
         func(1 * u.m, 1 * u.s, z=np.inf * u.eV)
 
 
-class Test_check_quantity_none_shall_pass:
+class TestCheckQuantityNoneShallPass:
```
StanczakDominik: Yeah, I guess that's fair. I didn't really see that aspect of pep8 as relevant to test grouping classes - but sure, why not. :)

Issue #620, opened 2019-05-09
Have decorator `check_quantity` enforce specified unit


rocco8773:
Currently [`check_quantity`](https://github.com/PlasmaPy/PlasmaPy/blob/efd7bff5bf34c64e70259886c9b4111b046f4d79/plasmapy/utils/checks.py#L20) does not enforce the specified unit. For example,
```
>>> import astropy.units as u
>>> from plasmapy.utils import check_quantity
>>>
>>> @check_quantity(**{'var': {'units': u.cm}}
... def show_unit(var: u.Quantity):
...     print(var)
>>> show_unit(5. * u.km)
5.0 km
```

I propose that `check_quantity` should enforce the units, such that...
```
>>> show_unit(5. * u.km)
500000.0 cm
```
I'm assuming the unit(s) specified to `check_quantity` are the one(s) you'll want to use in the decorated function.  If `check_quantity` is already examining the units, I don't see why it should not return the units you'll want to used.

For the case that multiple units are specified, then said units must not be equivalent.  That is, `var` would only be convertible to one of the specified units.

If this is a desired/accepted functionality, then I can initiate a PR.  I've already done this for a [branch on `bapsflib`](https://github.com/BaPSF/bapsflib/blob/add_astropy_units_to_plasma/bapsflib/utils/helpers.py), along with including an `equivalencies` key in the validations dictionary.  It does need a little cleaning and additional tests before establishing a PR.

StanczakDominik:
Huh. If I understand it correctly, as it stands, `check_quantity` will 

1. error for your `show_unit` function if I were to input something unrelated to distance such as `5 * u.kg`
2. pass silently when putting in `5 * u.m` or any unit of distance as the argument, assuming that `astropy.units` can handle any necessary conversions
3. do nothing when putting in `5 * u.cm` as specified (which is kind of a subset of 2)

However, I'm not sure this is all that wrong if we are to trust `astropy.units`. Whenever we use the quantity values directly in functions, we tend to use `var.si.value` anyway. Consider a slight rewrite:

```
>>> import astropy.units as u
>>> from plasmapy.utils import check_quantity
>>>
>>> @check_quantity(**{'var': {'units': u.cm}}
... def show_unit(var: u.Quantity) -> u.cm:     # annotate return type here
...     return var                              # return instead of print
>>> show_unit(5. * u.km)
500000.0 cm
```

This is either something we already have, something `astropy.units.quantity_input` has or something we wanted to have.

[...]

PR #648, opened 2019-06-18
Create validate decorators ("validators")

rocco8773:
The motivation for a set of "validating" decorators came from the discussion on the desired behavior for the `check_quantity` decorator in issue #620.  A "check" decorator is suppose to review input/output arguments of a function and raise warnings/errors when appropriate.  When an argument is an `astropy.unit.Quantity` object, then it could be useful to convert an input/output argument to the desired units, but this goes beyond the function of a "check."  So, the "validator" decorators will encompass the same check behavior of "check" decorators and perform input/argument conversions when appropriate.

In creating this module 3 decorators will be defined:

1. - [x] `check_values`
    * Decorator functionality will be fully defined by the `CheckValues` class
    * **Purpose:** To limit/control values of input arguments, such as allowing or not allowing `NaN`'s, `Inf`'s, non-negative numbers, etc.
1. - [x] `check_units`
    * Decorator functionality will be fully defined by the `CheckUnits` class
    * **Purpose:** To limit/control the units of input/output arguments.  By default the decorator will enforce a strict matching to the expected/desired unit, unless a `pass_equivalencies` is specified.
    * This decorator will not perform any unit conversions.
1. - [x] `validate_quantity`
   * Decorator functionality will be fully defined by the `ValidateUnit` class
   * Will leverage `check_values` and `check_units` to validate an `astropy.units.Quantity` object.
   * **Purpose:** To control/convert input/output arguments such that the arguments match the expected/desired values and units.  By default, unit conversion will take place when possible and raise an error if not.

**Notes:**
* `validate_quantity` is intended to replace `check_quantity`


StanczakDominik:
Looks neat for now! Definitely an improvement in terms of readability and having this part be less monolithic.

plasmapy/utils/decorators/checks.py
```diff
@@ -14,7 +14,257 @@
...
+        # adding a new validation key and default value here will automatically
+        # update the checks of the `checks` passed in
+        _defaults = {
+            'can_be_negative': True,
+            'can_be_complex': False,
+            'can_be_inf': True,
+            'can_be_nan': True,
+            'none_shall_pass': False,
+        }
+        return _defaults
```
StanczakDominik: Hm, since these are static, wouldn't it be enough to have this as a class field, as in the following?
```python
class CheckValues:
    _defaults = {
        'can_be_negative': True,
        'can_be_complex': False,
        'can_be_inf': True,
        'can_be_nan': True,
        'none_shall_pass': False,
    }
```

StanczakDominik: Unless the whole point is having the docstring attached to this dictionary, if that's the case - I'm not sure how else to do this.

rocco8773: @StanczakDominik If defined as a class attribute, then a user could potentially alter the default values. That is,

```python
>>> from plasmapy.utils.decorators import CheckValues
>>> cv = CheckValues()
>>> cv._defaults
{'can_be_negative': True,
 'can_be_complex': False,
 'can_be_inf': True,
 'can_be_nan': True,
 'none_shall_pass': False}
>>>
>>> cv._defaults['none_shall_pass'] = True
>>> cv._defaults
{'can_be_negative': True,
 'can_be_complex': False,
 'can_be_inf': True,
 'can_be_nan': True,
 'none_shall_pass': True}
```

By declaring it as a property, I can then prevent any unwanted changes to the default values.  Not that this situation is likely to come up, if ever.

StanczakDominik: Oh, I wouldn't have thought of that. Since defaults are marked as implementation details by `_` prefixed, I think any user who bothers modifying them either knows what they're doing or is unlikely to exist in the first place, as you said.

rocco8773: You're right, this way is not a very pythonic approach, and probably does not deserve this level of rigidity.  I'm continuously trying to break my old-school rigid programming ways and rely more on python's conventions.

I'll move the defaults into a "protected" class attribute.

rocco8773:
It's not ready yet.  I've only completed the `CheckValues` decorator class and `check_values` decorator.  Right now, I'm translating some old code to make `CheckUnits` and `ValidateQuantity`.  Then I'll test and ensure the documentation is complete.  For completeness, I'll also update the original PR comment with further details (under an EDIT heading).

I'll also replace all the `check_quantity` decorator usages with `validate_quantity` and, then, I'll mark the PR as ready for review.

StanczakDominik:
Just going to merge the latest master into this, see how it affects CI.

rocco8773:
Ok, I think this is getting close to completion.  I've gotten everything coded up and tested, but I still need to...

1. update docstrings/documentation
2. drop `check_quantity` for `validate_quantity`

I am having issues with 2 tests failing on TravisCI that do not fail on my local machine.  I'm bit at a loss for why at the moment though.

@StanczakDominik and @namurphy when I replace `check_quantity` with `validate_quantity`, do you want me to setup the decoration such that equivalent units pass for input arguments and the function return, or have them converted?

plasmapy/utils/decorators/tests/test_checks.py
```diff
@@ -10,10 +11,939 @@
...
+    def test_cu_preserves_signature(self):
+        """Test `CheckValues` preserves signature of wrapped function."""
+        # I'd like to directly dest the @preserve_signature is used (??)
+
+        wfoo = CheckUnits()(self.foo_no_anno)
+        assert hasattr(wfoo, '__signature__')
+        assert wfoo.__signature__ == inspect.signature(self.foo_no_anno)
+
+    @mock.patch(CheckUnits.__module__ + '.' + CheckUnits.__qualname__,
+                side_effect=CheckUnits, autospec=True)
```
StanczakDominik: What's the point of patching this?

plasmapy/utils/decorators/tests/test_checks.py
```diff
@@ -10,10 +11,939 @@
...
+    def test_decorator_func_def(self, mock_cu_class):
+        """
+        Test that :func:`~plasmapy.utils.decorators.checks.check_units` is
+        properly defined.
+        """
+        # create mock function (mock_foo) from function to mock (self.foo_no_anno)
+        mock_foo = mock.Mock(side_effect=self.foo_no_anno, name='mock_foo',
+                             autospec=True)
+        mock_foo.__name__ = 'mock_foo'
+        mock_foo.__signature__ = inspect.signature(self.foo_no_anno)
```
StanczakDominik: Does this mock serve any purpose besides  checking if it's been called?

rocco8773: I could be open to an argument to remove this mock.

This mock is setup to ensure the decorator `check_units` actually calls the wrapped function. (see the `assert mock_foo.called` [below](https://github.com/rocco8773/PlasmaPy/blob/576e49bb3ef2d909833449cc9be918a28dbbb276/plasmapy/utils/decorators/tests/test_checks.py#L542))

plasmapy/utils/decorators/tests/test_checks.py
rocco8773: This patch creates the `mock_cu_class` Mock object, which is used to ensure the `check_units` decorator calls the `CheckUnits` class with the correct arguments (see lines [541-549](https://github.com/rocco8773/PlasmaPy/blob/576e49bb3ef2d909833449cc9be918a28dbbb276/plasmapy/utils/decorators/tests/test_checks.py#L541-L549) below).  This then allows me to do 99% of my tests on `CheckUnits` and not have to duplicate them on `check_units`.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1083 @@
...
+            valueerror_msg = f"The return value "
+        else:
+            valueerror_msg = f"The argument '{arg_name}' "
+        valueerror_msg += f"to function {self.f.__name__}() can not contain"
+
+        # check values
+        for ckey in self.__check_defaults:
+            if ckey == 'none_shall_pass':
+                if arg is None and arg_checks[ckey]:
+                    return
```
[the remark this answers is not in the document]
rocco8773: This should probably be a `break` instead of a `return`.  If the argument is `None` and `none_shall_pass=True`, then there is no point in doing the remaining checks.  `return` and `break` essentially do the same thing here, but, when `return` is used, we'd run into issues of unaccessible code if there was code after the for-loop.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1083 @@
...
+        'units': None,
+        'equivalencies': None,
+        'pass_equivalent_units': False,
+        'none_shall_pass': False,
+    }
+
+    def __init__(
+            self,
+            checks_on_return: Union[u.Unit,
+                                    List[Union[u.Unit]],
```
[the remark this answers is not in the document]
rocco8773: Yeah, not sure how this happened. It'll be fixed.

plasmapy/utils/decorators/validators.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Parameters
+    ----------
+    func
+    validations_on_return
+    validations
+
+    Returns
+    -------
+
```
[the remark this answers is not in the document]
rocco8773: Yep, I'm working my way to filling out the validator docstrings.  This will come shortly.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1083 @@
...
+        if checks_on_return is not None:
+            self._checks['checks_on_return'] = checks_on_return
+
+    @property
+    def checks(self):
+        """
+        Requested checks on the decorated function's input arguments
+        and/or return.
+        """
+        return self._checks
```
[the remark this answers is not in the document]
rocco8773: Yeah, I've always liked properties for key variables, since (1) it's easier to produce clean documentation, (2) can make code more readable IMHO, and (3) it gives me some setter and getter controls down the line if needed.

StanczakDominik:
The `missing_file` failure means we'll need to add `changelog/648.feature.rst` and fill it out.

rocco8773:
@StanczakDominik Yes...yes...I have not merged in the updated `master` or setup the `changelog` yet.  I'll do that in the near future.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1083 @@
...
+        ================ ======= ================================================
+        Key              Type    Description
+        ================ ======= ================================================
+        can_be_negative  `bool`  [DEFAULT `True`] values can be negative
+        can_be_complex   `bool`  [DEFAULT `False`] values can be complex numbers
+        can_be_inf       `bool`  [DEFAULT `True`] values can be :data:`~numpy.inf`
+        can_be_nan       `bool`  [DEFAULT `True`] values can be :data:`~numpy.nan`
+        none_shall_pass  `bool`  [DEFAULT `False`] values can be a python `None`
+        ================ ======= ================================================
+
```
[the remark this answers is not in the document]
rocco8773: I doubt `CheckValues` will have anything *checks* that are not boolean values.  I kept this format since `CheckUnits` does have non-boolean *checks* and I was keeping the docstrings between the two classes consistent.

rocco8773:
I think this is finally good for a full review.

Check out my changelog entry.  I used sphinx roles in hopes they will successfully link to the respective class and function documentation, but I don't know how to fully compile the changelog to see if this works.

I'll also pick one formulary module to apply the decorators to.  Probably `plasmapy.formulary.relativity` since it only has one function.

rocco8773:
@StanczakDominik I'd like to save this merge until PR #722 is approved.  Then I have a chance of discovering any unforeseen errors as I implement `validate_quantities()`.

plasmapy/formulary/relativity.py
```diff
@@ -9,13 +9,12 @@
 import numpy as np
 
 from astropy import units as u
-from plasmapy import utils
 from astropy.constants import c
+from plasmapy import utils
+from plasmapy.utils.decorators import validate_quantities
```
namurphy: I'm wondering if the name `decorators` (which I think we've used for a while) is a bit general.  Perhaps change `plasmapy.utils.decorators` to `plasmapy.utils.validators`?

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+                                       RelativityError)
 from textwrap import dedent
-from typing import (Dict, Union)
+from typing import (Any, Dict, List, Tuple, Union)
+
+try:
+    from astropy.units.equivalencies import Equivalency
+except ImportError:
+    # TODO: remove once we have dependency Astropy >= 3.2.1
+    # astropy defined the Equivalency class in v3.2.1
```
namurphy: Oh, I didn't know that Astropy did this!  I like it.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+                if arg_name == 'checks_on_return':
+                    continue
+
+                # check argument
+                self._check_value(bound_args.arguments[arg_name],
+                                  arg_name,
+                                  checks[arg_name])
+
+            # call function
+            _return = f(**bound_args.arguments)
```
namurphy: I'm wondering if there would be an incompatibility here with the [positional only parameters](https://www.python.org/dev/peps/pep-0570/) that were implemented in Python 3.8.  I don't know if it would be worth fixing that now, but we might want to add a comment or note in the docstring if this wouldn't work with positional only parameters, or perhaps create an issue.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+            out_checks[param.name] = {}
+            for v_name, v_default in self.__check_defaults.items():
+                try:
+                    out_checks[param.name][v_name] = \
+                        param_in_checks.get(v_name, v_default)
+                except AttributeError:
+                    # for the case that checks are defined for an argument,
+                    # but is NOT a dictionary
+                    # (e.g. CheckValues(x=u.cm) ... this scenario could happen
+                    # curing subclassing)
```
namurphy: Possible typo: perhaps `during` instead of `curing`

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+                    out_checks[param.name][v_name] = v_default
+
+        # Does `self.checks` indicate arguments not used by f?
+        missing_params = [
+            param
+            for param in set(self.checks.keys()) - set(out_checks.keys())
```
namurphy: I like set operations in Python!  :+1:

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+            @check_values(arg1={'can_be_negative': False, 'can_be_nan': False},
+                          arg2={'can_be_inf': False},
+                          checks_on_return={'none_shall_pass': True)
+            def bar(self, arg1, arg2):
+                return None
+    """
+    if checks_on_return is not None:
+        checks['checks_on_return'] = checks_on_return
+
+    if func is not None:
```
namurphy: I'm a little confused about what `func is not None` represents.  Maybe an intermediate variable name?  (This could have been a useful place for a walrus operator!)

Is this related to perhaps allowing this to be used as a decorator for both when the decorator itself has arguments and when it doesn't?

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+        return CheckValues(**checks)(func)
+    else:
+        return CheckValues(**checks)
```
namurphy: An alternative (which you can decide if it would be cleaner or not):
```Python
return CheckValues(**checks) if func is None else CheckValues(**checks)(func)
```

plasmapy/formulary/relativity.py
rocco8773: Well, I see `plasmapy.utils.decorators` as a general grab-bag of decorators for PlasmaPy.  Within the `decorators` sub-package I've split the **"check"** decorators into `checks.py` and the **"validate"** decorators into `validators.py`.  I just also make the key classes and functions accessible in the `plasmapy.utils.decorators` name space.

I could import from `plasmapy.utils.decorators.validators`, but I think `validate_quantities` is fairly self explaining.

plasmapy/utils/decorators/checks.py
rocco8773: Yeah, I forget how I stumbled across this.  I'm glad they have, but they haven't quite propagated the class throughout AstroPy.  The old-styple equivalencies were a list of 2-4 element tuples and the new-style class sub-classes `UserList`.  Thus, the new-style plays nicely with the old foundation and the old foundation hasn't been specifically modified for the `Equivalencies` class.

plasmapy/formulary/relativity.py
namurphy: Understood.  Makes sense!

plasmapy/utils/decorators/checks.py
rocco8773: Ummm....I'll need to think about this more.  I do think this would be good as a separate issue, assuming a quick alteration does not come to mind.

plasmapy/utils/decorators/checks.py
rocco8773: they're set-tacular!

plasmapy/utils/decorators/checks.py
rocco8773: Exactly!  The `func is not None` corresponds to use `check_values(foo, **checks)`.  The `func is None` corresponds to the "sugar" syntax use and `check_values(**checks)(foo)`.

plasmapy/utils/decorators/checks.py
rocco8773: When the expression gets long enough, I prefer to break the if-statement over multiple lines.  It's just more readable to me.  Of course, this is subjective.

plasmapy/formulary/tests/test_transport.py
```diff
@@ -55,6 +55,8 @@ class Test_classical_transport:
     @classmethod
     def setup_class(self):
         """set up some initial values for tests"""
+        # TODO: when implementing validate_quantities remove this set equivalencies...
+        #       it affects the entire test environment
```
StanczakDominik: Oh. Yeah, it probably would... It would probably be best to apply equivalencies to tests via context managers, [as per Astropy docs](https://docs.astropy.org/en/stable/units/equivalencies.html#using-equivalencies-in-larger-pieces-of-code), but I'd expect a lot of code to break if we removed this here. Definitely material for another issue...

StanczakDominik: This is now #725 !

plasmapy/formulary/relativity.py
StanczakDominik: I agree that this makes sense as it stands. :+1:

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+                     checks_on_return={'none_shall_pass': True)
+        def foo(arg1, arg2):
+            return None
+
+        # on a method
+        class Foo:
+            arg1={'can_be_negative': False, 'can_be_nan': False},
+                         arg2={'can_be_inf': False},
+                         checks_on_return={'none_shall_pass': True)
+            def bar(self, arg1, arg2):
```
StanczakDominik: Doesn't look like we're actually calling `@CheckValues` here!

plasmapy/utils/decorators/checks.py
StanczakDominik: I would be really suprised if the addition of positional-only params broke inspect usages... I think it's probable to be fine.

plasmapy/formulary/tests/test_transport.py
rocco8773: A context manager could be a useful thing for the future. I’ve done a with decorator using `h5py.File` that worked beautifully. 

Luckily, this bit of code is not necessary when `validate_quantities` is properly applied. I’ve already deleted this TODO and the referenced code in PR #722 .

plasmapy/utils/decorators/checks.py
rocco8773: Bizarre! How did that happen? Ok, I’ll probably get to that fix tomorrow.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+        list
+            a flattened list of astropy :mod:`~astropy.units.equivalencies`
+
+        """
+        new_list = []
+        for el in elist:
+            if not isinstance(el, list):
+                new_list.append(el)
+            else:
+                new_list.extend(self._flatten_equivalencies_list(el))
```
StanczakDominik: Ooooh, my recursion sense is tingling! This would deal with any level of nesting, if I'm correct, right?

StanczakDominik: Ooooh, my recursion sense is tingling! This would deal with any level of nesting, if I'm correct, right?

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+    .. _astropy equivalencies:
+        https://docs.astropy.org/en/stable/units/equivalencies.html
+    """
+    if checks_on_return is not None:
+        checks['checks_on_return'] = checks_on_return
+
+    if func is not None:
+        return CheckUnits(**checks)(func)
+    else:
+        return CheckUnits(**checks)
```
StanczakDominik: E L E G A N T

I also like how the docstring for this is 20x longer. :laughing: 

Seriously, though, well refactored :)

plasmapy/utils/decorators/checks.py
StanczakDominik: In that case...
```suggestion
    if func is not None:  # direct check_values(foo, **checks) usage
```

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+                          checks_on_return={'none_shall_pass': True)
+            def bar(self, arg1, arg2):
+                return None
+    """
+    if checks_on_return is not None:
+        checks['checks_on_return'] = checks_on_return
+
+    if func is not None:
+        return CheckValues(**checks)(func)
+    else:
```
StanczakDominik: and here:
```suggestion
    else:  # decorator @check_values usage
```

plasmapy/utils/decorators/checks.py
StanczakDominik: Shorter, but a bit more confusing due to the two variants differing only in calling `(func)` at the end. I'd prefer keeping it as it is.

plasmapy/utils/decorators/tests/test_validators.py
```diff
@@ -0,0 +1,490 @@
...
+            vq = ValidateQuantities(**validations)
+            vq.f = self.foo
+
+            assert vq._validate_quantity(*args, validations) == case['output']
+            assert mock_cu_checks.called
+            assert mock_cv_checks.called
+
+    def test_vq_preserves_signature(self):
+        """Test `ValidateQuantities` preserves signature of wrapped function."""
+        # I'd like to directly dest the @preserve_signature is used (??)
```
StanczakDominik: Inconsequential, I know, but, typo fix:
```suggestion
        # I'd like to directly test the @preserve_signature is used (??)
```

I'm not sure how this goal could be achieved without profiling which functions are being called... and I think that might be more trouble than it's worth.

plasmapy/utils/decorators/tests/test_validators.py
```diff
@@ -0,0 +1,490 @@
...
+            @ValidateQuantities(validations_on_return={'can_be_negative': False})
+            def bar(self, x: u.cm) -> u.m:
+                return x + self.y
+
+        foo = Foo(-10 * u.cm)
+        assert foo.bar(20.*u.cm) == 0.1*u.m
+        with pytest.raises(ValueError):
+            foo.bar(5*u.cm)
+
+    @mock.patch(ValidateQuantities.__module__ + '.' + ValidateQuantities.__qualname__,
```
StanczakDominik: Thanks, I didn't know about `__qualname__`!

plasmapy/utils/decorators/tests/test_checks.py
```diff
@@ -10,10 +11,1066 @@
...
+    @staticmethod
+    def foo_stars(x: u.Quantity, *args, y=3 * u.cm, **kwargs):
+        return x.value + y.value
+
+    @staticmethod
+    def foo_with_none(x: u.Quantity, y: u.cm = None):
+        return x.value + y.value
+
+    def test_inheritance(self):
+        assert issubclass(CheckUnits, CheckBase)
```
StanczakDominik: That is way more meticulous than I would have thought about if writing it on my own!

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+        Dict[str, Dict[str, Any]]
+            A complete 'checks' dictionary for checking function input arguments
+            and return.
+        """
+        # initialize validation dictionary
+        out_checks = {}
+
+        # Iterate through function bound arguments + return and build `out_checks`:
+        #
+        # artificially add "return" to parameters
```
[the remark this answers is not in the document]
rocco8773: I do it this way because we scan through all `signature(foo).parameters` to review annotations and compare them to **checks**.  However, `signature(foo)` does not include the return annotations in the `parameters` dictionary.   I don't want to write separate code to handle return checks, so I artificially create a `'checks_on_return'` parameter (`inspect.Parameter()`) with the return annotations.  This way, the input parameters and return parameters run through the same "checks" code.

plasmapy/utils/decorators/checks.py
```diff
@@ -1,20 +1,1219 @@
...
+            if param.annotation is not inspect.Parameter.empty:
+                # unit annotations defined
+                _units_anno = param.annotation
+
+            if _units is None and _units_anno is None and param_checks is None:
+                # no checks specified and no unit annotations defined
+                continue
+            elif _units is None and _units_anno is None:
+                # checks specified, but NO unit checks
+                msg = f"No astropy.units specified for "
```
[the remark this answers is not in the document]
rocco8773: I discovered f-string when review PlasmaPy code...they are so nice to use!

plasmapy/utils/decorators/checks.py
rocco8773: Yes, up to the first instance of another list.  It will not flatten lists within lists.

plasmapy/utils/decorators/checks.py
rocco8773: Thank you! :)

PR #651, opened 2019-06-25
Nuke plasmapy.constants

StanczakDominik:
I thought we had done that a while ago.

`plasmapy.constants` is purely a wrapper for `astropy.constants`. It provides no useful functionality of its own except introducing confusion.

Off with its HEAD, as they say.


StanczakDominik:
Good point on the docs!

If there are any *constants* that are irrelevant to astronomy (as per documentation to `astropy.constants`) and relevant to plasma physics, we can always keep them in `physics` if need be.

PR #663, opened 2019-07-11
Remove astropy_helpers submodule

### LINKED ISSUE #660: Remove astropy_helpers
We've come to the conclusion that we've outgrown the functionality the submodule provides and we never really used its facilities for Cython that much.

StanczakDominik:
Closes #660


StanczakDominik:
- [ ] `plasmapy.test` is broken right now

StanczakDominik:
It seems that `import plasmapy; plasmapy.test()` works, while `python setup.py test` does not. But I'd honestly much rather just deprecate the latter and use tox instead.

ah_bootstrap.py
```diff
@@ -1,1010 +0,0 @@
-"""
```
namurphy: I'm going to miss this file, even though I had no idea what it did.

StanczakDominik:
Thanks!

PR #664, opened 2019-07-11
Reenable parallel testing

### LINKED ISSUE #661: Bring back parallel testing in tox.ini
I forgot a flag in there while debugging.

StanczakDominik:
Closes #661

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,12 @@ matrix:
     fast_finish: true
     include:
         - name: "3.7"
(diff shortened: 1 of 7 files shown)


StanczakDominik:
I'm getting some rather weird test collection failures in `plasmapy.atomic.tests.test_ionization_states.py` - I'm starting to think rewriting those using fixtures, so that `pytest-xdist` can pick and choose which tests to send to which worker earlier, may solve it. I'm going to try doing that and see where that leads.

PR #666, opened 2019-08-22
Fix failing Pytests (deprecation of `raises(message=)`)

StanczakDominik:
This PR fixes currently failing tests caused by the final deprecation in Pytest 5.1 of the `message` argument to `pytest.raises`, which we had unwisely muted along with all deprecation warnings in setup.cfg.

It also removes the same argument used in pytest.warns, which is slated for deprecation
in a future release.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,8 +77,6 @@ doctest_plus = enabled
 doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
 # TODO we probably need to get rid of the following...
 filterwarnings =
-    once::DeprecationWarning
-    once::PendingDeprecationWarning
     ignore::plasmapy.utils.exceptions.RelativityWarning
     ignore::plasmapy.utils.exceptions.CouplingWarning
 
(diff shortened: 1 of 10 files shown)


StanczakDominik:
Circleci failures related to miniconda docker image PATH issues: https://github.com/ContinuumIO/docker-images/pull/141

setup.cfg
```diff
@@ -77,8 +77,6 @@ doctest_plus = enabled
 doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
 # TODO we probably need to get rid of the following...
 filterwarnings =
-    once::DeprecationWarning
-    once::PendingDeprecationWarning
```
namurphy: I wholeheartedly support this change, now that I think about it.  ...as well as after doing an `import this`.

> Errors should never pass silently.
> Unless explicitly silenced.

StanczakDominik: Yeah, I'll try to tackle the removal of the other warning filters next. Come to think about it, we're probably exposing any potential users to a whole lot of Coupling or RelativityWarnings we have no idea about.

StanczakDominik:
Thanks for the review @namurphy!

PR #668, opened 2019-08-27
Enable particle_stepper to run under CI

StanczakDominik:
As pointed out by @diego7319, this example wasn't running in CI and wasn't being actively tested. I simplified it a bit to run faster along the way.


StanczakDominik:
This should look better now!

StanczakDominik:
I went back to `trajectory`, because the lines were getting really verbose.

Issue #645, opened 2019-06-13
Implement decorator adding to_radians


namurphy:
Presently many of the functions in `plasmapy.physics` return angular frequencies with units of radians per second.  
```Python
>>> from plasmapy.physics import plasma_frequency
>>> from astropy import units as u
>>> plasma_frequency(1e19 * u.m ** -3, 'e-')
<Quantity 1.78398636e+11 rad / s>
```
Radians are dimensionless (#29); radians are units like meters per meter.  By default, `astropy.units` does not treat radians as dimensionless (though you can use the `u.dimensionless_angles()` equivalency).  Many plasma physics formulae treat angular frequencies as having units of inverse seconds, such as the electron inertial length which is dₑ = c / ωₚₑ.  If we return frequencies in units of radians per second then the problem is it's often necessary to include a new unit conversion with that equivalency enabled.
```Python
>>> (c / omega_pe).to(u.s ** -1, equivalencies = u.dimensionless_angles()).
```
After using `plasmapy.physics` to calculate various things, having to deal with the extra conversion factor has been a bit cumbersome.  I propose that we change functions that calculate angular frequencies to be in units of `1 / s` instead of `rad / s`.  The downside is that there may be confusion between regular frequencies and angular frequencies since there's a factor of 2π difference.  However, angular frequencies come up much more often than regular frequencies in plasma physics, in my experience.  

If we make this change, we should note this convention clearly in the documentation and update our development guide.

#29

rocco8773:
@namurphy I ran into this issue not too long ago.  My two cents...`1/s` and `rad/s` are too easily confused, especially when something is returned as `1/s`.  You're relying on the user to search for the correct answer, and that's assuming they are aware that this might be an issue.  The solution I came up with, something like `plasma_frequency()` would, by default, return `rad/s` second.   Then it is clear that the value is an angular frequency.  However, I provide a keyword, like `to_Hz=True`, that would apply the `1/2(pi)` and unit conversion to `1/s`.  Then the ambiguity between the two is mitigated.

StanczakDominik:
I agree with @rocco8773, I had the same kind of solution in mind. With the ecosystem of unit packages engulfed bigger turmoil than the Galactic Republic, I don't think we have the right to introduce more confusion, as in, `plasmapy-flavored astropy.units`.

Do you think a decorator would be a good choice here? ;)

rocco8773:
[...]

PR #670, opened 2019-09-04
from radians to hz decorator 

diego7319:
Closes https://github.com/PlasmaPy/PlasmaPy/issues/645

New decorator to address the new feature


plasmapy/physics/parameters.py
```diff
@@ -34,22 +34,18 @@
 
 def _grab_charge(ion, z_mean=None):
     """Utility function to merge two possible inputs for particle charge.
-
```
StanczakDominik: Do you think you could revert the whitespace changes here?

StanczakDominik:
Oh yeah, a bunch of the doctest failures in tests actually look like they're being caused by the whitespace removals! See e.g. this:

```

Expected:
    <Quantity 34922.9856304 m / s>
    References
    ----------
    .. [1] PlasmaPy Issue #186, ***/issues/186
    See Also
    --------
    plasmapy.physics.kappa_thermal_speed
    plasmapy.physics.kappa_velocity_1D
Got:
    <Quantity 34922.9856304 m / s>
```

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,38 @@
...
+Decorator to convert units of functions in /physics methods
+"""
+from astropy import units as u
+import functools
+import inspect
+from plasmapy.utils.decorators import preserve_signature
+__all__ = ["from_radians_to_hz"]
+
+
+def from_radians_to_hz(fn):
```
rocco8773: This decorator name is a little misleading. You're not converting radians to Hz, you're converting angular frequency to Hz.  I'd rename the decorator more along the lines of `rads_per_sec_to_hz` or `angular_freq_to_hz`.

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,38 @@
+"""
+Decorator to convert units of functions in /physics methods
+"""
+from astropy import units as u
+import functools
+import inspect
+from plasmapy.utils.decorators import preserve_signature
+__all__ = ["from_radians_to_hz"]
```
rocco8773: PEP8 standards have the `__all__` above the imports

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,38 @@
+"""
+Decorator to convert units of functions in /physics methods
+"""
+from astropy import units as u
```
rocco8773: For ease of reading, I generally like to group my imports by `import ...` and `from ... import ...`.  Then I alphabetize them.  This is not a PEP8 thing, just a me thing I think improve readability. :)

rocco8773:
Another thing we should consider, how do we document this functionality?  Basically, how do we show in the documentation that a certain decorated function has the `to_hz` keyword?

If you decorate a function like

``` python
@from_radians_to_hz
gyrofrequency():
    ...
   return value
```
how do we get the documentation to show `gyrofrequency()` has the `to_hz` keyword?  Is there a way for the decorator to inject this into the docstrings?  This seems like it could be messy.

plasmapy/utils/decorators/converter.py
StanczakDominik: They do? Cool, TIL, thanks!

rocco8773: Yeah, this rule applies to any module level dunder.  The only import statement that precedes dunders is `from __future__ ...`,

https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

plasmapy/physics/parameters.py
diego7319: yes of course, I will change it and add some tests on the next update

plasmapy/utils/decorators/converter.py
diego7319: changing that on the next push

plasmapy/utils/decorators/converter.py
diego7319: changing it on the next push

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,47 @@
...
+    -------
+    callable
+        Wrapped version of the function.
+    -------
+    """
+    wrapped_sign = inspect.signature(fn)
+    fname = fn.__name__
+    @preserve_signature
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
```
rocco8773: As `wrapper()` is written, this will cause an `Exception` to be raised if the wrapped function does not defined keyword `to_hz` and/or `**kwargs`.  In my opinion, requiring the wrapped function to define these keywords defeats the purpose of creating a decorator. @namurphy and @StanczakDominik , you can correct me on this if you think differently.  This can easily be corrected by one of two ways...

``` python
def wrapper(*args, to_hz=False, **kwargs):
    
    _result = fn(*args, **kwargs)

    if to_hz:
        return _result.to(u.Hz, equivalencies=[(u.cy/u.s, u.Hz)])
    return _result
```
or
``` python
def wrapper(*args, **kwargs):

    to_hz = kwargs.pop('to_hz', False)

    _result = fn(*args, **kwargs)

    if to_hz:
        return _result.to(u.Hz, equivalencies=[(u.cy/u.s, u.Hz)])
    return _result
```

There's no fundamental difference between the two, but the former is a little more readable and concise.

* You can delete all the `bound_args`, `given_params`, etc.  They were only needed in the decorators you saw because those decorators actually examined the function parameters to determine what it was suppose to do. You don't need that here.

* I'll add another comment to the main PR thread about adding documentation to the wrapped function.  This goes along with the statement I made at the beginning of this comment in reference to not requiring the wrapped function to be code knowing it's decorated.

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,47 @@
...
+        The function being wrapped.
+
+    Returns
+    -------
+    callable
+        Wrapped version of the function.
+    -------
+    """
+    wrapped_sign = inspect.signature(fn)
+    fname = fn.__name__
```
rocco8773: delete...this is never used

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,47 @@
...
+    fn: callable
+        The function being wrapped.
+
+    Returns
+    -------
+    callable
+        Wrapped version of the function.
+    -------
+    """
+    wrapped_sign = inspect.signature(fn)
```
rocco8773: delete...see previous comment about defining `wrapper()`

plasmapy/utils/decorators/converter.py
StanczakDominik: That's a very good point! In fact, now that I think about it, I think something along these lines  was what we had in mind a while discussing - apologies for not recalling that sooner.

I think I like the second option more: the first one is going to fail for functions without `**kwargs` in the signature.

rocco8773: The first option would not cause an error since `to_hz` is not in the `kwargs` dictionary, so it's never passed to the wrapped function.  The second option would have that error if `'to_hz'` was not explicitly popped form the `kwargs` dictionary.

rocco8773: I'll add, we should probably raise an exception if the wrapped function has defined the `to_hz` keyword.

StanczakDominik: You are, of course, absolutely right, while I am definitely way past my optimal bedtime right now :zzz:. I like the first one now.

rocco8773:
> Another thing we should consider, how do we document this functionality? Basically, how do we show in the documentation that a certain decorated function has the `to_hz` keyword?
> 
> If you decorate a function like
> 
> ```python
> @from_radians_to_hz
> gyrofrequency():
>     ...
>    return value
> ```
> 
> how do we get the documentation to show `gyrofrequency()` has the `to_hz` keyword? Is there a way for the decorator to inject this into the docstrings? This seems like it could be messy.

This is going to be a little tricky to figure out.  My first thought on achieving this would be appending the function docstring like...

``` python
def angular_freq_to_hz(fn):
    
    fn.__doc__ += """
    Parameters
    ------------
    to_hz: bool
        Set `True` to to convert function output from angular frequency to Hz
    """

    @preserve_signature
    @functools.wraps(fn)
    def wrapper(*args, to_hz=False, **kwargs):
        ...
    return wrapper
```

There are two potential problems with this...
1. How does Sphinx handle a docstring with two parameters sections!?...eek!
1. Will this alter the function signature sphinx constructs for documentation?
    * I'm talking about the blue section heading for the function ([like this heading](http://docs.plasmapy.org/en/stable/api/plasmapy.physics.parameters.mass_density.html#plasmapy.physics.parameters.mass_density))
    * If not, we might have to touch the `fn.__signature__` to achieve this.

This will have to be played with to find out the solution.

StanczakDominik:
The docstring issue is a big one, going with this approach - it would probably be easiest to split on docstring sections and paste a pre-written to_hz flag description as the last one for Parameters.

But it might honestly be the best use of resources to reconsider the whole decorator approach and just add the argument manually to the few functions that need it, and do the conversion inside of their body.

I know at least I have a bad case of overengineering solutions. This may have been one of them.

rocco8773:
Yeah, I tend to over engineer things too. The plus side would be, you discover nuances you wouldn’t have seen before.

This might be simple if we can leverage the [Other Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) docstring section.

diego7319:
what if we dont use a decorator and just replaced it with an if in the last return

rocco8773:
@diego7319 that is what @StanczakDominik was getting at with his over engineering statement.

We've gotten this far, so I'm curious how the documentation would build if you did what I said in the above comment.  Just replace the 
```
Parameters
----------
```
with
```
Other Parameters
----------------
```

If the documentation come out clean, then I think we have a clever decorator; otherwise, we'll be over engineering and should just resort to coding per function.

diego7319:
> @diego7319 that is what @StanczakDominik was getting at with his over engineering statement.
> 
> We've gotten this far, so I'm curious how the documentation would build if you did what I said in the above comment. Just replace the
> 
> ```
> Parameters
> ----------
> ```
> 
> with
> 
> ```
> Other Parameters
> ----------------
> ```
> 
> If the documentation come out clean, then I think we have a clever decorator; otherwise, we'll be over engineering and should just resort to coding per function.

I just made a commit to test that

diego7319:
I think we also need to delete this part from the docs (in gyrofrequency)
`The recommended way to convert from angular frequency to frequency is to use an equivalency between cycles per second and Hertz, as Astropy’s dimensionles_angles equivalency does not account for the factor of 2*pi needed during this conversion. The dimensionless_angles equivalency is appropriate when dividing a velocity by an angular frequency to get a length scale.`

plasmapy/physics/tests/test_parameters.py
```diff
@@ -443,17 +443,24 @@ def test_gyrofrequency():
...
+                      27992490076.528206)
+
     assert np.isclose(gyrofrequency(2.4 * u.T, signed=True).value,
                       -422116821083.3284)
 
     assert np.isclose(gyrofrequency(1 * u.G).cgs.value,
                       1.76e7, rtol=1e-3)
 
+    assert np.isclose(gyrofrequency())
+
```
StanczakDominik: [This test is faling](https://dev.azure.com/plasmapy/PlasmaPy/_build/results?buildId=122), as it probably should be without any arguments being passed in!

plasmapy/physics/parameters.py
```diff
@@ -1424,6 +1456,12 @@ def lower_hybrid_frequency(B, n_i, ion='p+'):
         which defaults to protons.  If no charge state information is
         provided, then the ions are assumed to be singly charged.
 
+    to_hz : bool, optional
+        By default the output of this function will return the gyrofrequency in
+        radians per second. However, it can also be expressed in units of Hertz.
+        Default is 'False', if it is changed to True decorator angular_freq_to_hz
+        will convert it to Hertz.
+
```
StanczakDominik: If the decorator with the `+=` parameters works as intended, we can get rid of this added docstring note wherever it's used - otherwise we're gonna see `to_hz` twice.

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,28 @@
...
+import functools
+import inspect
+from astropy import units as u
+from plasmapy.utils.decorators import preserve_signature
+
+
+def angular_freq_to_hz(fn):
+
+    fn.__doc__ += """
+    Parameters
```
StanczakDominik: Oh wow, this works! Awesome!

https://2687-46810954-gh.circle-artifacts.com/0/root/project/docs/_build/html/api/plasmapy.physics.parameters.gyrofrequency.html?highlight=gyrofrequency

Maybe, so as to not duplicate the Parameters heading:

```suggestion
    Additional parameters
```

Of course it would be amazing to put it along with `Parameters`, but let's skip the parsing.

StanczakDominik:
> I think we also need to delete this part from the docs (in gyrofrequency)
> `The recommended way to convert from angular frequency to frequency is to use an equivalency between cycles per second and Hertz, as Astropy’s dimensionles_angles equivalency does not account for the factor of 2*pi needed during this conversion. The dimensionless_angles equivalency is appropriate when dividing a velocity by an angular frequency to get a length scale.`

I think it would be fine to leave it in - our functions will now be able to handle it, but people will still need to convert units later, and this is a good heads up for them.

plasmapy/physics/tests/test_parameters.py
diego7319: my bad, I forgot to enter values in it. fixing it now

plasmapy/utils/decorators/converter.py
rocco8773: Yeah, that's the same result I got when testing this locally.  The `to_hz` parameter definition is inject at the very bottom of the documentation.   This still happens identically when I used the ["Other Parameters" heading suggested by the numpydoc documentation](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard).  The one downside, the signature/function heading (the blue heading) does not get updated with `to_hz=False`.

I've been searching though the sphinx documentation to see how this heading is generated but its still not clear to me.  If this is okay with you, then we could make this an issue for future improvement. ¯\\_(ツ)_/¯

rocco8773:
@diego7319 and @StanczakDominik , I still think we should add a bit of code to the decorator that raises an exception if the wrapped function's signature uses the `to_hz` keyword.  Something like,

``` python
sig = inspect.signature(fn)
if 'to_hz' in sig.parameters:
    raise ValueError(f"Wrapped function '{fn.__name__}' can not use keyword 'to_hz'.  Keyword reserved for decorator functionality.")
```

This can be placed between the decorator definition and the wrapper definition.

StanczakDominik:
Sure, thanks for the reminder! That one thing, a quick test and that's a wrap. :)

plasmapy/utils/decorators/converter.py
rocco8773: @diego7319 and @StanczakDominik, I think I figured out how to update the signature as well...

``` python
def angular_freq_to_hz(fn):
    # raise exception if fn uses the 'to_hz' kwarg
    sig = inspect.signature(fn)
    if 'to_hz' in sig.parameters:
        raise ValueError

    # make new signature for fn
    new_params = sig.parameters.copy()
    new_params['to_hz'] = inspect.Parameter('to_hz', inspect.Parameter.POSITIONAL_OR_KEYWORD, default=False)
    new_sig = inspect.Signature(parameters=new_params.values(), return_annotation=sig.return_annotation)
    fn.__signature__ = new_sig

    # add 'to_hz' to fn docstring
    fn.__doc__ = """
    Other Parameters
    ----------------
    to_hz: bool
        Set `True` to to convert function output from angular frequency to Hz
    """

    def wrapper(*args, to_hz=False, **kwargs):
        ...
```

This accomplished everything we wanted, except placing `'to_hz'` into the original parameters section, on my local documentation build.

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,46 @@
...
+import inspect
+from astropy import units as u
+from plasmapy.utils.decorators import preserve_signature
+
+
+def angular_freq_to_hz(fn):
+    # raise exception if fn uses the 'to_hz' kwarg
+    sig = inspect.signature(fn)
+    if 'to_hz' in sig.parameters:
+        raise raise ValueError(f"Wrapped function '{fn.__name__}'
```
rocco8773: double `raise` delete one

plasmapy/utils/decorators/converter.py
```diff
@@ -0,0 +1,46 @@
...
+    fn.__signature__ = new_sig
+
+    # add 'to_hz' to fn docstring
+    fn.__doc__ = """
+    Other Parameters
+    ----------------
+    to_hz: bool
+        Set `True` to to convert function output from angular frequency to Hz
+    """
+    fn.__doc__ += """
```
rocco8773: you've added to `fn.__doc__` twice...delete one instance...note for section heading that the "-" line must match the length of the heading, that is...
```
Other Parameters
----------------
```

not 

```
Other Parameters
-------------
```
nor
```
Other Parameters
---------------------
```

diego7319: thank you for that, im pushing code again

diego7319: so, the idea is to put the to_hz = True/False as an argument on the decorator?. Because the line
`if 'to_hz' in sig.parameters:`
will crash in the current way 
`def lower_hybrid_frequency(B, n_i, ion='p+', to_hz=False):`

rocco8773: This is the point.  We're enforcing that the `to_hz` keyword is reserved for the decorator and, thus, the function being wrapped should NOT use it.  None of the function definitions should declare the `to_hz` keyword.

rocco8773:
@StanczakDominik , What are you thoughts on testing the `to_hz` keyword for each decorated function? I'm torn on it.  I feel the decorator should be directly tested to make sure it's converting correctly and, then, we assume it works correctly for all decorated functions.

plasmapy/utils/decorators/converter.py
diego7319: So how are we going to do the conversion? if we use it inside another function like `lower_hybrid_frequency` which is using this 3 functions 
`    omega_ci = gyrofrequency(B, particle=ion)
    omega_pi = plasma_frequency(n_i, particle=ion)
    omega_ce = gyrofrequency(B)`
It should return with rad / s, but if we called from 
`gyrofrequency(0.1*u.T)` 
in this case should it return in hz by itself?

rocco8773: I'm not sure what issue you're trying to point out.  The definition for all characteristic plasma frequencies (lower-hybrid, plasma frequency, gyro-frequencies, etc.) are in terms of `rads/s`.  Thus, any defined "frequency" function should, by default, return a quantity in units of `rads/s`.  The decorator just takes that result and converts it to `Hz` when requested.  So...

`gyrofrequency(0.1*u.T)` will be in `rads/s`
`gyrofrequency(0.1*u.T, to_hz=True)` will be in `Hz`

In the `to_hz` instance, the calculation inside the `gyrofrequency()` function is still done in `rads/s` and then converted by the decorator.  The function never even sees the `to_hz` keyword.

diego7319:
> > @StanczakDominik , What are you thoughts on testing the `to_hz` keyword for each decorated function? I'm torn on it. I feel the decorator should be directly tested to make sure it's converting correctly and, then, we assume it works correctly for all decorated functions.
> 
> I would indeed love a quick test that defines a decorated function returning 2 pi rad/s and checks that it outputs 1 Hz with the flag.
> 
> I like how the code looks now, but I think it broke with some clunky test code we're using elsewhere (see: tests on azure). I will take a bite out of breakfast and then handle that. I'll add the test along the way.

Test are failing in assert_can_handle_nparray(lower_hybrid_frequency) including plasma_frequency, upper_hybrid_frequency and gyrofrequency, seems like a conflict with the decorator

StanczakDominik:
Yeah, and interestingly the code didn't seem to mind it before messing with the signature.

This is not a before breakfast kind of problem though, and my dog is getting pretty insistent about a proposed walk, so... A bit later :D

StanczakDominik:
Dropping `@functools.wraps(fn)` seems to have discarded most of the information sphinx needed to build its docs. This might help the other, Azure issue as well.

StanczakDominik:
There, that should do it! `wraps` took care of the issue with handles_nparray, but the test I added had no actual docstring, so it was failing on the "add new docstring bit to None" part.

If this passes, it's a speedy merge, otherwise I'm taking a break anyway. :laughing:

StanczakDominik:
And done, thanks @diego7319 and @rocco8773 ! :)

PR #671, opened 2019-09-05
Remove lingering astropy-helpers references

namurphy:
Following up on #663.  Pretty insignificant changes.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,8 +2,6 @@
 omit =
     ci-helpers/*
     */tests/*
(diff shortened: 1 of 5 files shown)


namurphy:
My first pull request since May!  :tada: :birthday: :1st_place_medal:   That's what you get from having  five work trips, proposal deadlines, [DPP-CPP](https://sites.google.com/pppl.gov/dpp-cpp/home) papers, a bunch of talks, ongoing Fortran obligations, and (thankfully) some vacation...

PR #674, opened 2019-09-05
Move pytest helpers to subdirectory of utils

namurphy:
Our pytest helpers functionality has been in a single file in plasmapy.utils, which means that the functionality in there is jumbled together with a bunch of other unrelated things.  Putting it in its own subdirectory will help with the organization.  I also split the existing functionality in pytest.
    
I updated the namespaces for run_test, etc. in a bunch of test files throughout the package.  I also made a few non-consequential changes in the test files that I noticed, mostly removing unused import statements.
    
I am planning on doing more refactoring to the pytest helpers functionality, which will enable me to do some more refactoring of tests (discussed in comments in #605).  The API for utils.pytest_helpers should be considered unstable at the moment.


plasmapy/atomic/parsing.py
```diff
@@ -433,13 +432,14 @@ def _reconstruct_ion_symbol(element: str, isotope: Integral = None, Z: Integral
...
-        keyword_string = ", " + str(kwargs).strip(r"}{'").replace("'", "").replace(":", " =")
-    else:
-        keyword_string = ""
-    return f"Particle({repr(arg)}{keyword_string})"
+    return pytest_helpers.call_string(Particle, arg, kwargs)
+#    if kwargs != {}:
+#        keyword_string = ", " + str(kwargs).strip(r"}{'").replace("'", "").replace(":", " =")
+#    else:
+#        keyword_string = ""
+#    return f"Particle({repr(arg)}{keyword_string})"
```
StanczakDominik: Dump the commented out code, maybe?

plasmapy/atomic/tests/test_parsing.py
```diff
@@ -9,7 +9,7 @@
     _case_insensitive_aliases,
     _case_sensitive_aliases,
     _parse_and_check_atomic_input,
-    _call_string,
+    _particle_call_string,  # duplicate with utils.pytest_helpers.error_messages.call_string?
```
StanczakDominik: What's that about?

plasmapy/physics/__init__.py
```diff
@@ -36,6 +36,4 @@
 
 from .relativity import Lorentz_factor
 
-from plasmapy import transport
-
```
StanczakDominik: ... wow, I have no idea why this was here. I'm surprised things didn't break by removing this!

plasmapy/atomic/parsing.py
namurphy: Oops!  I forgot about this.  Thanks!

plasmapy/atomic/tests/test_parsing.py
namurphy: I made a more general `plasampy.utils.pytest_helpers.call_string` function which creates a string that replicates the call from a supplied callable, args, and kwargs to help with error messages, so this private function for a specific case is no longer necessary.  I might as well go back and finish this change that I started but did not finish.

namurphy:
> I saw pytest_helpers messing up when plugged into the parallel testing of `pytest_xdist` somewhere along the last few of my PRs

Huh, strange!  If you have an idea for a failing test, that would be great.  Otherwise I can try to play around with it to see what problems I might run into.

> as this code base grows, we're going to be finding more edge cases...  I guess my point is that we should really make sure we aren't reinventing Pytest's own functionalities in here and stick with the "canonical" solutions to problems. 

My thought for the time being is that `pytest_helpers` should be for the testing patterns that come up over and over again for us, but are more specialized than what's in pytest.  I don't think that `pytest_helpers` should deal with edge cases too much, as those cases won't correspond to a common testing pattern.  I spent some time yesterday looking through pytest's documentation to get a better idea of what functionality already exists.  I remember you bringing this up before, so I'm trying to keep this in mind as I'm working on this.  There are also tools in Astropy too that we should make sure not to reinvent.
 
I want something like
```Python
>>> run_test(callable_object, args_list, kwargs_dict, expected_outcome)
```
that works if `expected_outcome` is a warning, exception, `Quantity`, `Unit`, number, `ndarray`, or whatever, and generates really useful error messages for all of these different types of comparisons.  The canonical `pytest` patterns would be under the hood, so this would essentially be an interface so that we wouldn't have to write the same code over and over.  I think this would simplify writing and reading for a lot of our tests.  (Plus, having an interface like `run_test` would make it easier to change to a different testing framework besides pytest if we end up having strong reasons to, countless eons into the future.)

Issue #673, opened 2019-09-05
Enable skipping of slow tests when using pytest from the command line


namurphy:
Some of our tests take noticeably longer than others, and often it's not necessary to run the full test suite.  The [pytest documentation](https://docs.pytest.org) describes how to [mark tests as slow and control whether or not they get run using command line flags](https://docs.pytest.org/en/latest/example/simple.html?highlight=slow#control-skipping-of-tests-according-to-command-line-option).  It would be great to implement this for our tests.  Our longest tests are maybe half a minute so this is not urgent, but it would be helpful to do this since our test suite is probably only going to grow.

My preference would be to skip the slow tests by default when running on the command line because most of the time we want the tests to run quickly so that we can find test failures quickly.  We do want to run the full test suite for pull requests, but I'd suggest that we have at least one run of our test suite only attempt the quick tests so that we can detect test failures quickly.

I'd also find it helpful if we could set it up so that the slow tests are run last, and have an option that the slow tests are only run if all of the quick tests pass.  That's a bit more work, so maybe we can leave that as a stretch goal.  Thanks!

StanczakDominik:
In other words, this would require us to change:

1. Find which tests are running the slowest
2. Tag them with `@pytest.mark.slow`, which is one time we probably won't have to argue about naming

I would be wary of skipping any tests as a default, though, as that would likely be more confusing to newcomers! Using `pytest -m 'not slow'` is simple enough, whereas I don't see any way easy way to tell Pytest "okay, run all the tests" with `not slow` being the default.

The ordering part can then be solved by simply using two commands: `pytest -m 'not slow'`, `pytest -m slow`. Not as flashy as a custom solution, but I briefly looked and don't think anything exists for ordering tests based on mark as of now.

For any active development you should probably be running `pytest plasmapy/transport` or whichever module you're working on at the time any way. That's going to limit the issue of having slow tests elsewhere.

StanczakDominik:
Oh ho ho, look what I've found:

https://stackoverflow.com/a/27899853/4417567

namurphy:
> I would be wary of skipping any tests as a default, though, as that would likely be more confusing to newcomers!

True enough!  In that case, let's go with running all tests as the default.

StanczakDominik:
Closed by #677!

PR #677, opened 2019-09-11
Enable skipping of slow tests when using pytest from the command line

goodab:
Adds the functionality requested in #673

Adds a marker and two options to pytest to speed up the test suite for routine tests. Slow tests can be marked with ```@pytest.mark.slow```. These tests can then be skipped with ```pytest --not-slow``` or exclusively executed with ```pytest --slow```.

### DIFF
--- a/plasmapy/conftest.py
+++ b/plasmapy/conftest.py
@@ -17,11 +17,14 @@ def pytest_addoption(parser):
     )
 
 def pytest_configure(config):
-    config.addinivalue_line("markers", ("slow: mark test as slow to run. Test can be skipped with"
-        " --not-slow or exclusively executed with --slow."))
+    config.addinivalue_line(
+        "markers",
+        ("slow: mark test as slow to run. Test can be skipped with --not-slow or exclusively "
+        "executed with --slow.")
+    )
 
 def pytest_collection_modifyitems(config, items):
-    skip_mark = None
+    skip_condtion = None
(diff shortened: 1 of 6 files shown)


plasmapy/conftest.py
```diff
@@ -1,7 +1,42 @@
...
+        # Skip slow tests
+        skip_mark = pytest.mark.skip(reason="Test is marked slow.")
+        skip_condition = lambda x: "slow" in x.keywords
+    elif config.getoption("--slow"):
+        skip_mark = pytest.mark.skip(reason="Test isn't marked slow.")
+        skip_condition = lambda x: "slow" not in x.keywords
+    if skip_condition is not None:
+        for item in items:
+            if skip_condition(item):
+                item.add_marker(skip_mark)
```
StanczakDominik: Huh! This looks like it'd probably work, but doesn't it duplicate the behavior of the following default calls to `pytest` without modifying this file?
```bash
pytest -m 'slow'
pytest -m 'not slow'
```
Since that's built in, it'd be probably better to take advantage of that and have less code to maintain, wouldn't it?

StanczakDominik:
Looks good! Do you think you could still update the description in conftest and put a note describing usage of the mark with the -m flag in the testing part of the documentation?

A shame about parametrize, but taking down two thirds of the runtime is a massive improvement nonetheless!

goodab:
@StanczakDominik Is it alright if I add ```# coverage : ignore``` to the additions to conftest?

PR #680, opened 2019-09-13
First go at a DistributionFuntion object

dstansby:
Here is a first go at making a `DistributionFunction` object, as discussed on riot with @namurphy and @blalterman. Please fire away with any and all suggestions at how I have implemented this (/tell me why my code is awful!). Some notes

- I think that an analytical and sampled distribution function will have such different ways of doing stuff that I've made two different classes. In the future these should definitely have a similar API though.
- I have left `AnalyticDistributionFunction` to be filled in since this isn't my area of speciality
- This needs tests/better docs, but I'm putting it out there for initial review before I work on those bits

### DIFF
--- a/plasmapy/distributions/distributions.py
+++ b/plasmapy/distributions/distributions.py
@@ -12,7 +12,7 @@ class DistributionFunction(metaclass=abc.ABCMeta):
     @n_spatial_dims.setter
     def n_spatial_dims(self, n_spatial_dims):
         if not 0 <= n_spatial_dims <= 3:
-            raise ValueError('Number of spatial dimensions must be bewteen '
+            raise ValueError('Number of spatial dimensions must be between '
                              f'0 and 3 (got {n_spatial_dims}).')
         self._n_spatial_dims = n_spatial_dims
 
@@ -23,7 +23,7 @@ class DistributionFunction(metaclass=abc.ABCMeta):
     @n_velocity_dims.setter
     def n_velocity_dims(self, n_velocity_dims):
         if not 1 <= n_velocity_dims <= 3:
-            raise ValueError('Number of velocity dimensions must be bewteen '
+            raise ValueError('Number of velocity dimensions must be between '
                              f'1 and 3 (got {n_velocity_dims}).')
         self._n_velocity_dims = n_velocity_dims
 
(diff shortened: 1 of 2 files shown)


blalterman:
Ok. I know this is silly and pendantic, but line 1 should probably be

class DistributionFunction(object):


if for no other reason than explicit is better than implicit.

---------------------
B. L. Alterman
Pronouns: He/Him/His
<https://spectrumcenter.umich.edu/article/designated-pronouns>
Candidate, Applied Physics
Solar and Heliospheric Research Group
Climate and Space Sciences and Engineering
University of Michigan
balterma@umich.edu

dstansby:
That's not required any more in python 3 (and it's customary not to include it in all the other projects I've contributed to)

blalterman:
All I was suggesting was explicit over implicit. Same reason nested
generators aren't the greatest idea.

---------------------
B. L. Alterman
Pronouns: He/Him/His
<https://spectrumcenter.umich.edu/article/designated-pronouns>
Candidate, Applied Physics
Solar and Heliospheric Research Group
Climate and Space Sciences and Engineering
University of Michigan
balterma@umich.edu

StanczakDominik:
> as discussed on riot with @namurphy and @blalterman

Could you link that discussion here, for context?

I'll take a look at the code a bit later today :)

dstansby:
https://matrix.to/#/!mThWGgbzeWtFrtMzCa:matrix.org/$1568310829519351FWmFu:matrix.org?via=matrix.org&via=openastronomy.org&via=cadair.com

plasmapy/distributions/distributions.py
```diff
@@ -0,0 +1,118 @@
...
+            self._spatial_untis = []
+            spatial_idxs = index_types.index('space')
+            for i in range(vel_idxs):
+                self._spatial_indexes.append(indexes[i])
+                self._spatial_units.append(index_units[i])
+
+        self.data = data
+        self.index_units = index_units
+
+        @property
```
namurphy: The indentation here appears to be one step too many

blalterman:
I'm in favor of the ABC. I think that's an ideal way to go about it and
help ensure a consistent API.

dstansby:
I've added some `abc` to the mix, although I haven't added any enforced methods yet. Things like calculating number density, velocity, and temperature moments will be good candidates for this though.

@StanczakDominik , nope, at the moment the input is 'processing agnostic' in the sense that it just takes some numpy arrays (as document). Will add an example somewhere!

PR #683, opened 2019-09-19
Guidelines for supporting old versions of Python, NumPy, and Astropy

namurphy:
[NumPy Enhancement Proposal (NEP) 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) includes suggested community guidelines for the support of old versions of Python and NumPy.  The idea behind NEP 29 is to provide predictability to when old packages stop being supported throughout the scientific Python ecosystem.  This pull request adapts these guidelines for PlasmaPy, in particular by including requirements for older versions of Astropy.  

The guidelines were adjusted to allow us some flexibility, since there may be situations where it would provide a much greater benefit to update sooner.  While we are still early in the development phase, we can get away with being less strict with supporting older versions of packages.  As we build up a user base and improve package stability, we should depart from these guidelines more sparingly.

### DIFF
--- a/docs/development/release_guide.rst
+++ b/docs/development/release_guide.rst
@@ -30,7 +30,11 @@ Pre-release
   * Use ``git shortlog -n -s -e`` for ``.mailmap``
   * Use ``astropy-tools/author_lists.py`` for ``credits.rst``
 
-* Edit ``setup.cfg`` to remove ``.dev`` from ``version = x.y.z.dev``
+* Update ``setup.cfg``
+
+  * Remove ``.dev`` from ``version = x.y.z.dev``
+  * Update minimum versions of required packages, including
+    ``python_requires``, ``install_requires``, and other variables
 
 * Reserve a digital object identifier on Zenodo, and update citation
   information (e.g., in ``plasmapy.__citation__`` and ``README.md``)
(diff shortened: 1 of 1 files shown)


docs/development/release_guide.rst
```diff
@@ -105,3 +109,31 @@ Post-release
...
+  Python.
+* PlasmaPy should support minor versions of NumPy initially released
+  in the 24 months prior to a planned project release date or the
+  oldest version that supports the minimum Python version (whichever is
+  higher).
+* PlasmaPy should support at least the 3 latest minor versions of
+  NumPy and Astropy.
+
+Upstream package requirements should only change during major or minor
+releases, and never during a patch release.  All supported minor
```
StanczakDominik: A question: suppose a NumPy patch release (say, 1.17.2) fixes a bug occuring in PlasmaPy 0.3.0. Do we:

- tell people to update their numpy version, as allowed by this note, or
- release 0.3.1 as a bugfix release which basically bumps the required NumPy version to 1.17.2? This seems to be actively discouraged according to this sentence.

docs/development/release_guide.rst
```diff
@@ -105,3 +109,31 @@ Post-release
...
+versions of Python should be in the test matrix and have binary
+artifacts built for releases.
```
StanczakDominik: As of now, we don't actually have binary artifacts, since we got rid of Cython, we're running on conda in so-called "noarch" mode as a pure Python dependency.

Issue #634, opened 2019-05-31
Create a class to represent arbitrary particle distribution functions


namurphy:
Plasmas often have particle distribution functions that differ from the standard analytic distribution functions (e.g., Maxwell-Boltzmann, kappa, etc.).  PlasmaPy should have a way to represent arbitrary particle distribution functions.  This would enable PlasmaPy to be used for the analysis of in situ spacecraft data.  It would be helpful to have this done before #11.

namurphy:
Hm...one possibility would be to create a `ParticleDistributionFunction` base class (maybe an abstract base class?) and then have subclasses for the different distribution types.  Might need to distinguish between 1D and 3D.

StanczakDominik:
What API would you imagine this with?

namurphy:
I'm not really sure yet.  It'd probably be helpful to come up with some use cases and design it from there.  I personally use a fluid approximation for most of my research, so I may not be the best person to ask!

diego7319:
Maybe using a design pattern should be helpful in this case, particularly https://refactoring.guru/design-patterns/factory-method  "Factory method". One Abstract mother class and a child distribution class.
A second option could be using the "adapter method" (https://refactoring.guru/design-patterns/adapter/python/example#lang-features, I think this is clearer https://www.djangospin.com/design-patterns-python/adapter/ )

Issue #447, opened 2018-05-08
Create custom particle subclass


lemmatum:
As per discussion on #439 it'd be useful to have a method for generating custom particles as part of the `Particle` class. In particular, we want to be able to pass a mass and a charge to create custom dust grain particles.

namurphy:
Now that #404 has been merged, one possibility for this issue would be to create a composite `Particle` from an `IonizationState` instance.  From that we would be able to get the mean and rms integer charge.  I don't think this should be the only way to create a custom `Particle`, but it would provide at least one way.

PR #555, opened 2018-10-05
Start of a dispersion package

dstansby:
A first go an starting a dispersion package. At the moment it is *very* bare bones, but includes an Input class and an Output class, that all dispersion solvers can use. I thought I'd open this up to start discussion.

PR #686, opened 2019-09-26
Dispersion api

StanczakDominik:
This PR takes the notes we wrote [during the Tuesday meeting](https://github.com/PlasmaPy/plasmapy-meeting-notes/blob/master/2019-09-24-design-meeting.md) and implements them as a failing test. Ideally, once we implement this, we'll remove the xfail and just have a passing, though basic, test.

As pointed out by @namurphy on (in the?) Matrix, #634, #447 are relevant here.

@dstansby, seeing that you started work on #680, what do you think about this approach?

### DIFF
--- a/plasmapy/dispersion/tests/test_api.py
+++ b/plasmapy/dispersion/tests/test_api.py
@@ -1,0 +1,85 @@
+import astropy.units as u
+from astropy import constants
+import pytest
+
+@pytest.mark.xfail(strict=True)
+def test_dispersion_api():
+    """
+    This is a broad overview of the initial ideas we had on the sample API for the dispersion solver
+    as per https://github.com/PlasmaPy/plasmapy-meeting-notes/blob/master/2019-09-24-design-meeting.md
+
+    It is marked as xfail - ideally a later PR will cause this 
+    """
+
+    p = Particle('p')
+
+    # class Species(Particle):
+    #     def __init__(self, particle_name, temperature, anisotropy, density, drift):
+    #         ...
+        
+    from plasmapy.atomic import CustomParticle # probably the best place for it but negotiable
+    custom_ion = CustomParticle(mass=25*electron_mass, charge=2 * constants.e.si)
+
+    from plasmapy.dispersion import Species # place negotiable? better name?
+
+    species0 = Species(custom_ion, 
+        anisotropy = 1/3.1415,
+        # or solve for anisotropy from T_perpendicular, T_parallel, but for now - it's fine to supply it
+        density = 1e23 * u.m**-3,
+        # in general, example: drift = [0, 0, 2] * (u.m/u.s), but for now don't worry about direction
+        drift = 2 * (u.m/u.s),
+    )
+
+    # physics initialization 
+    from plasmapy.dispersion import PhysicalInputs #, DispersionSolver
+    dispersion_input = PhysicalInputs(species = [species0],
+        magnetic_field = u.Quantity([0, 0, 5], u.T),
+        # kwargs:
+        wavenumber = np.linspace(0, 10, 1000) / u.m,
+        # wavenumber = np.array([0.1, 10, 3,]) / u.m,
+        angle = np.linspace(0, np.pi, 1000),
+        # v_Alfven/c needs to be controlled - magnetic_field, density give us v
+        # TODO what do we need for the relativistic dispersion?
+        # TODO Ask Daniel Verscharen and other astrophysicists
+        )    
+
+    # class DispersionSolver:
+    #     solvers = {'kinetic': KineticDispersionSolver,
+    #                'two_fluid': tfst}
+        
+    #     def __init__(self, solver_type, **solver_kwargs):
+    #         # doc issues but easy to extend
+        
+    from plasmapy.dispersion import KineticDispersionSolver
+    kds = KineticDispersionSolver(dispersion_input,
+                                  number_of_iterations = 10000,         
+                                  #Threshold for the determinant of the dispersion tensor:
+                                  # If det D <= det_D_threshold, the Newton iteration will 
+                                  # be stopped
+                                  det_D_threshold = 0.0001,  # TODO put reasonable float value here
+                                  # Maximum of sum in Bessel functions (both regular 
+                                  # and modified)
+                                  # can be very low (e.g., 3) for quasi-parallel propagation
+                                  nmax=1000,
+                                  # If I_n is less than this value, higher n are neglected:
+                                  Bessel_zero=1, # was 1.d-50, I'm unsure of what this syntax represents
+                                  # Initial guess for the frequency (complex number):
+                                  initial_guess=0.01+0.0001j,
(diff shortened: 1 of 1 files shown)


StanczakDominik:
I'm sorry for the massive review request spam, but I thought it'd be good to gather any second thoughts that you guys might have had afterwards!

dstansby:
Writing a test seems like an odd way to define an API to me - personally I think the way to go is to create some actual classes with explanatory documentation, that might be Abstract-Base-Classes (like I did in #680). Something more like https://github.com/PlasmaPy/PlasmaPy/pull/555 with appropriate modification?

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+    #     def __init__(self, particle_name, temperature, anisotropy, density, drift):
+    #         ...
+        
+    from plasmapy.atomic import CustomParticle # probably the best place for it but negotiable
+    custom_ion = CustomParticle(mass=25*electron_mass, charge=2 * constants.e.si)
+
+    from plasmapy.dispersion import Species # place negotiable? better name?
+
+    species0 = Species(custom_ion, 
+        anisotropy = 1/3.1415,
```
tulasinandan: We should either compute from distribution function or if we want to supply the parameters, then we need to supply one of the temperatures (preferably parallel) AND anisotropy. We will need the parallel temperature to compute parallel beta.

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+    # class DispersionSolver:
+    #     solvers = {'kinetic': KineticDispersionSolver,
+    #                'two_fluid': tfst}
+        
+    #     def __init__(self, solver_type, **solver_kwargs):
+    #         # doc issues but easy to extend
+        
+    from plasmapy.dispersion import KineticDispersionSolver
+    kds = KineticDispersionSolver(dispersion_input,
+                                  number_of_iterations = 10000,
```
tulasinandan: These numerical parameters were directly copied from Daniel Verscharen's NHDS code. Depending on whether we write our own or if we use a compiled version of Daniel's code, we might need to change these.

namurphy:
> Writing a test seems like an odd way to define an API to me - personally I think the way to go is to create some actual classes with explanatory documentation, that might be Abstract-Base-Classes (like I did in #680).

We've been trying to do more [test-driven development](https://en.wikipedia.org/wiki/Test-driven_development), where the test is written first and then the coding to get the test to pass happens after.  I personally have found this strategy to be quite helpful a lot of the time, and often regret it when I don't use it.  With that said, I think the abstract base class approach would probably go along nicely with the test-driven development approach.

StanczakDominik:
> Writing a test seems like an odd way to define an API to me - personally I think the way to go is to create some actual classes with explanatory documentation, that might be Abstract-Base-Classes (like I did in #680). Something more like #555 with appropriate modification?

Yeah, I may have gone overboard on the TDD approach :laughing: ABCs sound good!

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    This is a broad overview of the initial ideas we had on the sample API for the dispersion solver
+    as per https://github.com/PlasmaPy/plasmapy-meeting-notes/blob/master/2019-09-24-design-meeting.md
+
+    It is marked as xfail - ideally a later PR will cause this 
+    """
+
+    p = Particle('p')
+
+    # class Species(Particle):
```
dstansby: I don't think `Species` should subclass `Particle`, but should take a `Particle` object as one of it's `__init__` parameters. (ie. a single `Particle` can't have a temperature, but a `Species` that is comprised of lots of the same type of particle can)

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+    #     def __init__(self, particle_name, temperature, anisotropy, density, drift):
+    #         ...
+        
+    from plasmapy.atomic import CustomParticle # probably the best place for it but negotiable
+    custom_ion = CustomParticle(mass=25*electron_mass, charge=2 * constants.e.si)
```
dstansby: I think it would be nice if the `Particle` class could also take these kwargs (without needing another `CustomParticle` class).

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+
+    from plasmapy.dispersion import Species # place negotiable? better name?
+
+    species0 = Species(custom_ion,
```
dstansby: This test code is fine, but it should be noted for the future that the exact kwargs that `Species` accepts will vary, ie. in standard MHD a species just has a single temperature and no temperature anisotropy.

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+    species0 = Species(custom_ion, 
+        anisotropy = 1/3.1415,
+        # or solve for anisotropy from T_perpendicular, T_parallel, but for now - it's fine to supply it
+        density = 1e23 * u.m**-3,
+        # in general, example: drift = [0, 0, 2] * (u.m/u.s), but for now don't worry about direction
+        drift = 2 * (u.m/u.s),
+    )
+
+    # physics initialization 
+    from plasmapy.dispersion import PhysicalInputs #, DispersionSolver
```
dstansby: I think `PhysicalInputs` should end up being an `ABC`, which individual solvers have to subclass to define what their own exact physical inputs are.

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+                                  # Amplitude mode: If 1, then ampl = delta B_x / B0. 
+                                  # If 2, then ampl = delta B_y / B0.
+                                  # If 3, then ampl = delta B_z / B0
+                                  # ampl_mode=3,
+                                  # Amplitude for the calculation of polarization
+                                  # properties:
+                                  # ampl=1.d0
+                                  amplBETTERNAME = (0, 0, 1)
+                                  )
+
```
dstansby: There should be something like `kds.solve()` here, and then some lines that access the computed solution.

plasmapy/dispersion/tests/test_api.py
```diff
@@ -0,0 +1,85 @@
...
+                                  # If 3, then ampl = delta B_z / B0
+                                  # ampl_mode=3,
+                                  # Amplitude for the calculation of polarization
+                                  # properties:
+                                  # ampl=1.d0
+                                  amplBETTERNAME = (0, 0, 1)
+                                  )
+                     
+    from plasmapy.dispersion import FluidDispersionSolver
+    fds = FluidDispersionSolver(dispersion_input)
```
dstansby: As above, a fluid solver probably won't have the same input as a kinetic solver, so I think a new `dispersion_input` should be defined for this.

PR #689, opened 2019-09-29
bump sphinx version (>2) to generate binder link

jdkent:
I just read and went through @StanczakDominik's [helpful guide](https://stanczakdominik.github.io/posts/simple-binder-usage-with-sphinx-gallery-through-jupytext/) and applied it to my own repository, but I noticed the binder badges are not showing at the [end of the tutorials](http://docs.plasmapy.org/en/stable/auto_examples/particle_stepper.html#sphx-glr-auto-examples-particle-stepper-py).

For my repository the solution was [bumping the version of sphinx that needed to be installed](https://github.com/HBClab/NiBetaSeries/blob/76f9779c17ef8069410eb6774ac0bb95b917cefd/setup.cfg#L66) since readthedocs by default installs `sphinx==1.8.5`.

I'm hoping this solution works for this repository as a way of saying thank you for providing that guide.

P.S., I do not know if it is necessary, but [I wiped the build environment](https://docs.readthedocs.io/en/stable/guides/wipe-environment.html) before trying to change the sphinx version readthedocs uses

### DIFF
--- a/requirements/environment.yml
+++ b/requirements/environment.yml
@@ -18,7 +18,7 @@ dependencies:
     - numpydoc
     - pytest
     - lmfit
-    - sphinx
+    - sphinx>2
     - sphinx-gallery
     - sphinx-automodapi
     - sphinx_rtd_theme


StanczakDominik:
Hey, thanks for this PR, and I'm happy that my guide proved helpful :)

However, this issue seems to result from the fact that our RTD setup points to the `stable` (0.2 release) branch, rather than to the current master branch. [This is how it looks on master](http://docs.plasmapy.org/en/latest/auto_examples/plot_particle_stepper.html#sphx-glr-auto-examples-particle-stepper-py).

I'll update PlasmaPy's RTD to point to the current master, and, with your permission, I'll add a note to the guide along the lines of "@jdkent on GitHub [suggests pinning Sphinx version to >2](https://github.com/PlasmaPy/PlasmaPy/pull/689) helps if your binder badges aren't showing up).

StanczakDominik:
Actually you're probably right that sphinx needs to be modern-ish, if you take a look at our build logs, we're pulling Sphinx 2.2 from conda: https://readthedocs.org/projects/plasmapy/builds/9702951/ (hopefully you can access that...)

I didn't realize that! Thanks, TIL! :grinning:

PR #692, opened 2019-10-09
plasmapy.formulary - reshuffle

StanczakDominik:
This is hopefully the final attempt!


namurphy:
I just realized, we're going to have to update a lot of things in the docs, aren't we?  

Another thought is that it would probably be more straightforward for `plasmapy.formulary` to be imported automatically when doing `import plasmapy`, rather than having to specifically do `import plasmapy.formulary` or `from plasmapy import formulary`.  This will probably be our most commonly used subpackage, and the first one that many people would go to.  

Thanks for doing this!

StanczakDominik:
Sure, can do!

I also wanted to discuss the API here. I rather dislike having to do `from plasmapy.formulary.physics.parameters import a_thing`; I'm thinking of getting rid of the pretty arbitrary separation between `physics`, `mathematics` and `transport`, and just putting all of that into the main `formulary` namespace. *Flat is better than nested* etc etc.

namurphy:
Yeah, allowing users to do something like `from plasmapy.formulary import *` and just have everything right there would be a real convenience for when they're interactively analyzing some data set.  I like it!

namurphy:
Another thought...probably `plasmapy.formulary` is going to be one of the first packages to need a stable API (probably along with `plasmapy.atomic`).  That is to say, it should soon become open for extension but not for modification.  Probably version 0.3 will be too soon to promise that, but maybe 0.4 or 0.5.

StanczakDominik:
Stable-ish - I think once we move stuff to `formulary` and do a cross-check with The Formulary itself to see what's missing, it'll be fine.

I'll try to finish the docs (again...) tomorrow.

StanczakDominik:
Well, I may have gone a little overboard on the common namespace thing. See https://3007-46810954-gh.circle-artifacts.com/0/root/project/docs/_build/html/formulary/index.html. I think:

* [ ] I gave the page an auto-generated summary, but it's trying to look for definitions in the `formulary/__init__.py` file (due to star imports) and failing, so links aren't generating. I guess we could keep the per-topic summaries and say it's all available via `plasmapy.formulary.alfven_speed`.
- [ ] magnetostatics elements either need a separate place besides the formulary or shouldn't be *-imported into the
* [ ] I guess the only reasonable approach here is nuking the central autosummary - I'll do that.

By the way, I'm doing the star imports because we're defining `__all__` in the modules, so it should be fine.

StanczakDominik:
If this goes as expected, it should be ready for a round of review!

plasmapy/classes/sources/plasma3d.py
```diff
@@ -6,15 +6,9 @@
...
-                               m_e,
-                               c,
-                               mu0,
-                               k_B,
-                               e,
-                               eps0,
+from plasmapy.formulary.magnetostatics import MagnetoStatics
+
+from astropy.constants import (mu0,
                                )
```
SolarDrew: Auto-formatting shenanigans? Newline seems a bit excessive with a single import...

plasmapy/classes/sources/plasmablob.py
```diff
@@ -5,11 +5,10 @@
...
 import astropy.units as u
 
-from plasmapy.physics.parameters import _grab_charge
-from plasmapy.physics.dimensionless import (quantum_theta,
-                                            )
-from plasmapy.transport import (coupling_parameter,
-                                        )
+from plasmapy.formulary.parameters import _grab_charge
+from plasmapy.formulary.dimensionless import (quantum_theta,
+                                                      )
```
SolarDrew: See above.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -4,20 +4,22 @@
...
-                                collision_frequency,
-                                mean_free_path,
-                                mobility,
-                                Knudsen_number,
-                                coupling_parameter)
-from plasmapy.transport import (Spitzer_resistivity,
-                                fundamental_electron_collision_freq,
-                                fundamental_ion_collision_freq)
+from plasmapy.formulary.braginskii import (Coulomb_logarithm,
+                                           )
```
SolarDrew: And here.

plasmapy/classes/sources/plasma3d.py
StanczakDominik: Curse you, PyCharm! Thanks, Drew! :)

PR #697, opened 2019-10-15
Feature/improve particle comparison

raymonshansen:
**Comparing defaultdicts**
As per discussion in Matrix chat, this pull request adds the following:
- A 'collections_helpers.py' file containing a function which compares defaultdicts without mutating them.
- A 'collections_helpers_tests.py' with unittests for above helper function.
- Slight modifications to the Particle class, using the new helper function.

**NOTE**
There are two skipped tests that concerns how dicts with differing default types are compared. See also chat for a slightly more confusing overview of the matter. :)

This particular part of the Particle class is ignored by codecoverage?? :(

All code should be PEP8 compliant. I have tried to conform to the going style(for docstrings) to the best of my ability. Please let me know about all the things I have missed and I'll update the PR as soon as I can.

Thanks for feedback and help with requirements and questions.

Raymon Skjørten Hansen
Tromsø, Norway

### DIFF
--- a/plasmapy/utils/collections_helpers.py
+++ b/plasmapy/utils/collections_helpers.py
@@ -1,0 +1,46 @@
+from collections import defaultdict
+
+
+def compare_default_dicts(a: defaultdict, b: defaultdict) -> bool:
+    """Compare two defaultdicts, return True if equal, else False.
+    Does a benign or soft compare. If the defaultdicts COULD become
+    equal, they are considered equal.
+
+    * Does NOT change the memory imprint of any of the dictionaries.
+    * Any overlapping keys, must have same value.
+    * Keys unique to one, must have the default value of the other.
+    * Order of input does NOT matter.
+
+    Example:
+    a = defaultdict(lambda: "", a=42, b=42, c="")
+    b = defaultdict(lambda: 42, c="", d="")
+    compare_defaultdicts(a, b) -> True
+
+    Parameters
+    ----------
+    a : defaultdict
+        Default dictionary from collections
+    b : defaultdict:
+        Default dictionary from collections
+
+    Returns
+    -------
+    bool : True if equal, else False
+
+    """
+    a_keys = set(a)
+    b_keys = set(b)
+    a_unique_keys = (a_keys | b_keys) - b_keys
+    b_unique_keys = (a_keys | b_keys) - a_keys
+
+    # The intersecting keys must have the same value
+    if not all(a[key] == b[key] for key in (a_keys & b_keys)):
+        return False
+
+    # Keys unique to one, must have default value of other.
+    if not all(b.default_factory() == a[key] for key in a_unique_keys):
+        return False
+    if not all(a.default_factory() == b[key] for key in b_unique_keys):
+        return False
+
+    return True
--- a/plasmapy/utils/tests/collections_helpers_tests.py
+++ b/plasmapy/utils/tests/collections_helpers_tests.py
@@ -1,0 +1,100 @@
+import unittest
+from collections import defaultdict
+from plasmapy.utils import compare_default_dicts
+
+
+class TestDictCompare(unittest.TestCase):
+    """
+    Tests custom comparing of pythons defaultdicts.
+    Note that default value of int = 0 and default of
+    str = ''
+    """
+
+    def test_same_keys(self):
+        msg = "Two equal defaultdicts should be same."
+        a = defaultdict(int, a=1, b=1, c=1)
+        b = defaultdict(int, a=1, b=1, c=1)
+        result_a_first = compare_default_dicts(a, b)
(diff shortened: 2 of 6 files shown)


StanczakDominik:
I'll take a look at it a bit later!

plasmapy/utils/collections_helpers.py
```diff
@@ -0,0 +1,46 @@
...
+
+    * Does NOT change the memory imprint of any of the dictionaries.
+    * Any overlapping keys, must have same value.
+    * Keys unique to one, must have the default value of the other.
+    * Order of input does NOT matter.
+
+    Example:
+    a = defaultdict(lambda: "", a=42, b=42, c="")
+    b = defaultdict(lambda: 42, c="", d="")
+    compare_defaultdicts(a, b) -> True
```
StanczakDominik: This could be put into docstring form to get tested on each doc build! Take our other functions for examples.

plasmapy/utils/collections_helpers.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Parameters
+    ----------
+    a : defaultdict
+        Default dictionary from collections
+    b : defaultdict:
```
StanczakDominik:
```suggestion
    b : defaultdict
```

plasmapy/utils/tests/collections_helpers_tests.py
```diff
@@ -0,0 +1,100 @@
+import unittest
+from collections import defaultdict
+from plasmapy.utils import compare_default_dicts
+
+
+class TestDictCompare(unittest.TestCase):
```
StanczakDominik: I was a bit confused as to why you'd use `unittest` here, but sure, if you want to write them that way :) Pytest handles unittest-style tests, so it's fine.

plasmapy/utils/tests/collections_helpers_tests.py
```diff
@@ -0,0 +1,100 @@
...
+    def test_comparing_should_not_mutate_any_dict(self):
+        msg = "Comparing should not change any dict."
+        a = defaultdict(int, a='Hello', b=0.5, c=[])
+        a_copy = a.copy()
+        b = defaultdict(str, a=1001, b={}, c='World', d=[])
+        b_copy = b.copy()
+
+        # We're not interested in the actual dictionaries,
+        # just whether comparing mutated any of them.
+        _ = compare_default_dicts(a, b)
```
StanczakDominik: We can probably just skip the assignment, then:

```suggestion
        compare_default_dicts(a, b) 
```

plasmapy/utils/tests/collections_helpers_tests.py
```diff
@@ -0,0 +1,100 @@
...
+        result_a_first = compare_default_dicts(a, b)
+        result_b_first = compare_default_dicts(b, a)
+        self.assertEqual(result_a_first, True, msg)
+        self.assertEqual(result_b_first, True, msg)
+
+    @unittest.skip("Awaiting specific requirements")
+    def test_two_empty_with_different_default_value(self):
+        msg = "???"
+        a = defaultdict(int)
+        b = defaultdict(str)
```
StanczakDominik: @namurphy, care to give your opinion on this one?

plasmapy/utils/tests/collections_helpers_tests.py
```diff
@@ -0,0 +1,100 @@
...
+        result_a_first = compare_default_dicts(a, b)
+        result_b_first = compare_default_dicts(b, a)
+        self.assertEqual(result_a_first, False, msg)
+        self.assertEqual(result_b_first, False, msg)
+
+    @unittest.skip("Awaiting specific requirements")
+    def test_disjoint_values_match_default_of_other(self):
+        msg = "???"
+        a = defaultdict(str, a=0, b=0, c="")
+        b = defaultdict(int, c="", d="", e="")
```
StanczakDominik: I would expect this to fail because `0 == ""` is `False`. @namurphy ?

plasmapy/utils/tests/collections_helpers_tests.py
raymonshansen: This was to explicitly note that we are ditching the result. This is also why I made the dicts VERY different. We're not actually interested in the result from compare. We just want to make sure that the act of comparing doesn't modify any of them. But I'm happy either way. :)

plasmapy/utils/collections_helpers.py
raymonshansen: Oops!

plasmapy/utils/tests/collections_helpers_tests.py
raymonshansen: Well, both of the above boils down to whether or not we care about the default value in itself or just the k,v pairs actually present in the dicts. But they do make for some interesting edge cases!

PR #699, opened 2019-10-15
Update test_roman.py

raymonshansen:
Fix copy-paste bug. 'test_from_roman' now tests the from_roman method. :)

Raymon Skjørten Hansen
Tromsø, Norway

### DIFF
--- a/plasmapy/utils/roman/tests/test_roman.py
+++ b/plasmapy/utils/roman/tests/test_roman.py
@@ -167,7 +167,7 @@ def test_from_roman(integer, roman_numeral):
     Test that `~plasmapy.utils.roman.from_roman` correctly converts
     Roman numerals to integers.
     """
-    run_test(func=roman.to_roman, args=integer, expected_outcome=roman_numeral)
+    run_test(func=roman.from_roman, args=integer, expected_outcome=roman_numeral)
 
 
 @pytest.mark.parametrize('input, expected_exception', toRoman_exceptions_table)
--- a/plasmapy/utils/roman/tests/test_roman.py
+++ b/plasmapy/utils/roman/tests/test_roman.py
@@ -167,7 +167,7 @@ def test_from_roman(integer, roman_numeral):
     Test that `~plasmapy.utils.roman.from_roman` correctly converts
     Roman numerals to integers.
     """
-    run_test(func=roman.from_roman, args=integer, expected_outcome=roman_numeral)
+    run_test(func=roman.from_roman, args=roman_numeral, expected_outcome=integer)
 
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Looks like tests failed, though, could you take a look at why?

raymonshansen:
On it...

raymonshansen:
Looking at the output from tests, it looks like they are failing when 'from_roman' outputs int and the expected from the test parameter list is an Int16, Int32 and Int64 respectively. These are types from numpy. If from_roman is intended to support optional integer output, it needs rewriting, otherwise remove them from the expected results perhaps?

plasmapy/utils/roman/tests/test_roman.py
```diff
@@ -167,7 +167,7 @@ def test_from_roman(integer, roman_numeral):
     Test that `~plasmapy.utils.roman.from_roman` correctly converts
     Roman numerals to integers.
     """
-    run_test(func=roman.to_roman, args=integer, expected_outcome=roman_numeral)
+    run_test(func=roman.from_roman, args=roman_numeral, expected_outcome=integer)
```
namurphy:
```suggestion
    run_test(func=roman.from_roman, args=roman_numeral, expected_outcome=int(integer))
```

namurphy: The output of `from_roman` is supposed to be an `int`, while `run_test` is being made to expect an `np.int32` or similar when it does its type checks.  Doing `expected_outcome=int(integer)` should hopefully resolve this, and is more in accordance with how `from_roman` is expected to behave.

raymonshansen: That should fix it. Although from a readability point of view, it isn't super obvious why this is done. A prettier way would perhaps be to test to_roman with np types in a separate test. It should be easy to see what a test is testing. I could do that if yous want? If not, I'm happy to commit the suggestion. :)

namurphy: That's a good point that the intention is not clear from my suggestion.  I had also wondered about renaming `integer` to something that conveys the intent better (maybe `input_for_to_roman`?  not sure...) or to have a conditional that shows when/why we need to make this conversion:
```Python
if isinstance(integer, np.integer): 
    integer = int(integer)
```
I like your suggestion too, so please feel free to take whatever approach you think is most readable.  Thanks again!

raymonshansen: I'll commit the quick-fix and then I can separate the tests in a different PR.

PR #700, opened 2019-10-16
Generalize ion sound speed

### LINKED ISSUE #316: Generalize `ion_sound_speed`
Currently `ion_sound_speed` under `parameters.py` only works in the limit `alpha = 1 / (k * lambda_De)` is large(?), such that the plasma is not dispersive. This should be generalized for all values of `alpha` and the docstring should be edited to reflect this generalization.

The general equation is on pg 2 of this [Google Doc](https://docs.google.com/document/d/1EcuxYoSzy-MpjjjqBE3E6zE8g6xYU2BH1MorU7_DQQA/edit#heading=h.88burrsgujyu). Please note that this equation is for `omega`, whereas what we want is `v = omega / k`, so just drop the `k` on the right hand side.

tvarnish:
Closes #316. Generalizes `ion_sound_speed` to work for any case, not just for the limit when `(k * lambda_D) ** 2` is small. Additional term from Chapter 5.1, Froula, Glenzer, Luhmann and Sheffield "Plasma Scattering of Electromagnetic Radiation".

Docstring updated to remove note about the non-dispersive limit. Updated docstring examples, and included a new example with a large value for `(k * lambda_D) ** 2`.

Updated tests for ion_sound_speed, and added additional tests for the (general) large limit. Updated the `assert_can_handle_nparray` pytest helper method to work for the additional parameter, `k`,
added to `ion_sound_speed`.


tvarnish:
(I haven't added a changelog entry to the PR yet as I wasn't sure what type to use.)

StanczakDominik:
> (I haven't added a changelog entry to the PR yet as I wasn't sure what type to use.)

It's definitely a `feature` :)

plasmapy/physics/parameters.py
```diff
@@ -210,10 +210,14 @@ def Alfven_speed(B, density, ion="p+", z_mean=None):
...
     T_i={'units': u.K, 'can_be_negative': False},
-    T_e={'units': u.K, 'can_be_negative': False}
+    T_e={'units': u.K, 'can_be_negative': False},
+    n_e={'units': u.m ** -3, 'can_be_negative': False},
+    k={'units': u.m ** -1, 'can_be_negative': False}
     )
 def ion_sound_speed(T_e,
                     T_i,
+                    n_e,
+                    k,
```
StanczakDominik: I wonder whether it might not be possible to supply default arguments for these two, so that the default behavior is the non-dispersive limit we had previously?

plasmapy/physics/parameters.py
```diff
@@ -288,36 +299,38 @@ def ion_sound_speed(T_e,
 
     Notes
     -----
-    The ion sound speed :math:`V_S` is approximately given by
+    The ion sound speed :math:`V_S` is given by
```
StanczakDominik: Always nice to be rid of `approximately`!

plasmapy/physics/parameters.py
tvarnish: That would make a lot more sense! I'll work on updating it...

plasmapy/physics/parameters.py
tvarnish: That said, I'll make sure to add a note about approximating the function in the non-dispersive limit if the values for `k` and/or `n_e` aren't given (in response to your other question).

StanczakDominik: Yeah, it's definitely a good idea to add that :)

StanczakDominik:
I just merged #692 in, so this PR will also need to merge those changes, since the location for `parameters` changed to `plasmapy/formulary/parameters.py`. Apologies for the trouble!

tvarnish:
@StanczakDominik I'm still getting the hang of using GitHub, so I just wanted to check (before I break anything): how should I go about merging that PR into this one? Should I pull the changes into my local master, then merge that with my local branch?

StanczakDominik:
Precisely! I would thus go:
```bash
git commit  # your current changes if any
git checkout master
git pull           # assuming you have `master` set up to track PlasmaPy/PlasmaPy#master, not tvarnish/PlasmaPy#master
git checkout generalize_ion_sound_speed
git merge master
```

tvarnish:
Great! I think that's worked.

StanczakDominik:
You'll still need to push the merge commit, though! :)

plasmapy/physics/parameters.py
StanczakDominik: That's a good question! It is getting late in my timezone, though, and I'm rather tired. Would you mind if I left the pondering of it and digging into the code for tomorrow? Alternatively, you could ask in [our chatroom](https://riot.im/app/#/room/#plasmapy:openastronomy.org) if anyone has any ideas!

tvarnish: I wouldn't mind at all! Feel free to look through it whenever's best for you. :)

plasmapy/formulary/parameters.py
```diff
@@ -209,10 +209,14 @@ def Alfven_speed(B, density, ion="p+", z_mean=None):
...
     T_i={'units': u.K, 'can_be_negative': False},
-    T_e={'units': u.K, 'can_be_negative': False}
+    T_e={'units': u.K, 'can_be_negative': False},
+    n_e={'units': u.m ** -3, 'can_be_negative': False},
+    k={'units': u.m ** -1, 'can_be_negative': False}
     )
 def ion_sound_speed(T_e,
                     T_i,
+                    n_e=0*u.m**-3,
+                    k=0*u.m**-1,
```
rocco8773: I'd use `None` as default values...something like

``` python
n_e = None,
k = None,
```

I think this would solve all your checking of default value issues.  At this point, if the user passes in a bad value (such as `n_e = 0.0 * u.m**-3`), then errors will be thrown accordingly, as they should.

plasmapy/formulary/parameters.py
```diff
@@ -333,9 +356,17 @@ def ion_sound_speed(T_e,
...
+    if n_e != 0:
+        lambda_D = Debye_length(T_e, n_e)
+    else:
+        lambda_D = 0 * u.m
 
     try:
-        V_S_squared = (gamma_e * Z * k_B * T_e + gamma_i * k_B * T_i) / m_i
+        V_S_squared = (gamma_e * Z * k_B * T_e + gamma_i * k_B * T_i) / (m_i \
+        * (1 + k ** 2 * lambda_D ** 2))
         V_S = np.sqrt(V_S_squared).to(u.m / u.s)
```
rocco8773:
``` python
if n_e is None or k is None:
    klD2 = 0.0
else:
    lambda_D = Debye_length(T_e, n_e)
    klD2 = (k * lambda_D) ** 2

try:
    V_S_squared = (gamma_e * Z * k_B * T_e + gamma_i * k_B * T_i) / (m_i * (1. + klD2))
    V_S = np.sqrt(V_S).to(u.m / u.s)
```

plasmapy/formulary/parameters.py
```diff
@@ -209,10 +209,14 @@ def Alfven_speed(B, density, ion="p+", z_mean=None):
 @utils.check_relativistic
 @utils.check_quantity(
     T_i={'units': u.K, 'can_be_negative': False},
-    T_e={'units': u.K, 'can_be_negative': False}
+    T_e={'units': u.K, 'can_be_negative': False},
+    n_e={'units': u.m ** -3, 'can_be_negative': False},
+    k={'units': u.m ** -1, 'can_be_negative': False}
```
rocco8773: Add `None` to allowed values

``` python
n_e={'units': u.m ** -3, 'can_be_negative': False, none_shall_pass: True},
k={'units': u.m ** -1, 'can_be_negative': False, none_shall_pass: True},
```

tvarnish:
Thanks for the feedback @rocco8773! Using `None` for the default values was a much simpler solution that anything I was considering. I've also put in a `UserWarning` if only one of `k` or `n_e` is specified, as you suggested. You (and/or @StanczakDominik) might want to take a look at that message to check it's not too long and gets the point across clearly.

StanczakDominik:
Thanks @rocco8773 for handling this! I wasn't able to look at it yesterday :(

plasmapy/formulary/parameters.py
```diff
@@ -342,6 +344,11 @@ def ion_sound_speed(T_e,
     <Quantity 229585.96150738 m / s>
 
     """
+    
+    if (n_e is None and k is not None) or (k is None and n_e is not None):
+        raise UserWarning("The non-dispersive limit has been assumed for this "
+                          "calculation. To prevent this, values must be "
+                          "specified for both n_e and k.")
```
StanczakDominik: This will interrupt execution (as with an exception), with warnings you need `warn` from the `warnings` module, rather than `raise`!

Also, we could probably throw a `plasmapy.utils.PhysicsWarning` - we've done that for other approximations elsewhere.

plasmapy/formulary/tests/test_parameters.py
```diff
@@ -215,6 +215,12 @@ def test_ion_sound_speed():
     with pytest.raises(RelativityError):
         ion_sound_speed(T_i=T_i, T_e=T_e, n_e=n_e, 
                         k=k_1, gamma_i=np.inf)
+        
+    with pytest.raises(UserWarning):
+        ion_sound_speed(T_i=T_i, T_e=T_e, n_e=n_e)
+        
+    with pytest.raises(UserWarning):
```
StanczakDominik: This, then, becomes `pytest.warns`!

StanczakDominik:
And the warning message is awesome! :)

plasmapy/formulary/parameters.py
tvarnish: Ah... ok! I've not used warnings before, so had assumed it would be the same as an exception. Thanks for spotting that! I'll go ahead and change that now.

tvarnish: I've fixed the warning (using `warnings.warn`), and changed it from `UserWarning` to `PhysicsWarning`.

plasmapy/formulary/parameters.py
```diff
@@ -284,41 +303,58 @@ def ion_sound_speed(T_e,
...
+    <Quantity 203155.03286794 m / s>
+    >>> ion_sound_speed(T_e=500*u.eV, T_i=200*u.eV, n_e=n, k=k_1, ion='D+')
+    <Quantity 229585.96150738 m / s>
 
     """
+    
+    if (n_e is None and k is not None) or (k is None and n_e is not None):
+        warnings.warn("The non-dispersive limit has been assumed for "
+                      "this calculation. To prevent this, values must "
+                      "be specified for both n_e and k.", PhysicsWarning)
```
rocco8773: Moves this down to be with the `if-else` that determines `klD2`.  This would make the code more readable and better grouped.

plasmapy/formulary/parameters.py
```diff
@@ -333,9 +369,16 @@ def ion_sound_speed(T_e,
 
     T_i = T_i.to(u.K, equivalencies=u.temperature_energy())
     T_e = T_e.to(u.K, equivalencies=u.temperature_energy())
+    
+    # Assume non-dispersive limit if values for n_e (or k) are not specified
+    if n_e is None or k is None:
+        klD2 = 0.0
+    else:
+        lambda_D = Debye_length(T_e, n_e)
+        klD2 = (k * lambda_D) ** 2
```
rocco8773: As the [above comment suggests](https://github.com/PlasmaPy/PlasmaPy/pull/700/files#r336746787), place warning in this `if-else` statement

```python
if n_e is None or k is None:
    klD2 = 0.0

    if n_e is None and k is None:
        warnings.warn("say stuff")
else:
    lambda_D = Debye_length(T_e, n_e)
    klD2 = (k * lambda_D) ** 2
```

tvarnish: Do you think the warning should just go within the first if statement (rather than being within the nested if statement)? That way, the function would warn the user whenever the non-dispersive limit is assumed.

tvarnish: Or... do you think it would make sense to only call the warning when one of `k` or `n_e` is given, so scripts that are dependant on PlasmaPy and were previously happy to use the non-dispersive limit don't cause warnings to be shown?

rocco8773: Good point.  I think the `n_e=None` and `k=None` case should not throw a warning since it is the default state.  @StanczakDominik correct me if you think otherwise.  In that case, I would restructure it like...

```python
klD2 = 0.0
if (n_e is None) ^ (k is None):
    warnings.warn("say stuff")
elif n_e is not None and k is not None:
    lambda_D = Deby_length(T_e, n_e)
    klD2 = (k * lambda_D) ** 2
```

The parenthesis are needed in this case since the xor operator `^` only acts on booleans.  The parenthesis force the conditionals into booleans before the xor operation is poerformed.

tvarnish: Neat! I never knew there was an XOR operator in python. I see what you’re saying about the default case. That makes sense.

StanczakDominik: Yeah, defaults should probably not throw warnings! I haven't (consciously) seen XOR in Python yet either. I like the proposed way of doing things :)

StanczakDominik:
And thanks @rocco8773 for the very helpful advice!

tvarnish:
Brilliant! I'm pleased I could help. :) Thank you @StanczakDominik and @rocco8773 for your really helpful code reviews and advice!

tvarnish:
@StanczakDominik I think the `docs/about/credits.rst` file might have broken... I saw there was a conflict so just went to check it had been properly resolved, and it seems the contents of someone's email has been inserted into the contributor list!

StanczakDominik:
Insert expletive here. Curse the middle click on Linux!

Thanks a ton for pointing that out; I've cleaned it up on the master branch. Phew.

PR #707, opened 2019-10-20
created a new function to calculate visualization sweap

techieashish:
I've refactored the code to calculate visualization sweap. Docstring shouldn't be required as function name is enough to tell what exactly is happening.

### DIFF
--- a/plasmapy/diagnostics/langmuir.py
+++ b/plasmapy/diagnostics/langmuir.py
@@ -389,7 +389,7 @@ def swept_probe_analysis(probe_characteristic, probe_area, gas_argument,
                                   hot_fraction), probe_area)
 
     if visualize:  # coverage: ignore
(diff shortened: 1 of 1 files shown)


StanczakDominik:
All right, there the docs are building!

If you take a look at https://3307-46810954-gh.circle-artifacts.com/0/root/project/docs/_build/html/diagnostics/index.html, you'll see a hole in the docs where the plotting function is - which is why, as a rule, we don't allow public-facing functionality without docstrings.

However, I've been thinking about it more and, while Langmuir probes are not my area of expertise, I've come to agree that it might indeed not make sense to move this functionality out of that function in this way. Perhaps @lemmatum had something in mind when creating that issue back then, but it seems to make little sense to me now. Perhaps we should close this PR and that issue.

namurphy:
Closing this since the Langmuir functionality is undergoing a major overhaul/refactoring.

Issue #653, opened 2019-06-28
All our numerical docstring outputs broke on Astropy.constants value updates


StanczakDominik:
https://travis-ci.org/PlasmaPy/PlasmaPy/jobs/551231362

To quote a single case that explains just about everything:

```
306     >>> electric_charge('p+')
Expected:
    <<class 'astropy.constants.codata2014.EMCODATA2014'> name='Electron charge' value=1.6021766208e-19 uncertainty=9.8e-28 unit='C' reference='CODATA 2014'>
Got:
    <<class 'astropy.constants.codata2018.EMCODATA2018'> name='Electron charge' value=1.602176634e-19 uncertainty=0.0 unit='C' reference='CODATA 2018'>
````

This propagates to cases like

```
AssertionError: Coulomb logarithm for GMS-6 should be 0.030720859361047514 and not 0.030720879053202647.
```

Clearly we will need a more clever way to handle this.

StanczakDominik:
It seems scipy solves this by simply... not doctesting; which I suppose does make a bunch of sense with inherently floating-point output.

We could also add `+SKIP` flags to docstrings where necessary; but I fear that messes up readability of those as docstrings and doesn't scale well (76 tests fail as of now - I didn't count how many of those are docstrings).

https://github.com/astropy/pytest-doctestplus/issues/55

StanczakDominik:
A good bunch of these are `isclose` errors with `atol=0`, which could simply be bumped up.

StanczakDominik:
I'm pretty sure #712 closed this!

Issue #229, opened 2018-01-07
Prevent numpy 0.14 from breaking all docstrings


StanczakDominik:
This relates to #135 and is the reason why tests are failing for #228 - and everything that's been updated in the last 21 hours.

[Numpy 1.14 has just been released](https://github.com/numpy/numpy/releases/tag/v1.14.0), and within the release notes there's this tidbit:

> "Many changes to array printing, disableable with the new "legacy" printing mode"

So yeah. All our docstrings are broken now, basically. Possible solutions, in order of "from most temporary":
* We could set our tests to run on 1.13 for now. [This is how to implement this](https://github.com/PlasmaPy/PlasmaPy/compare/master...StanczakDominik:doctestplus), and that branch could be merged.
* We could do the legacy printing mode, but that involves adding it to all of our files, I think.
* Eventually we'd probably want to rewrite our docstrings to numpy 1.14 standard.

samurai688:
Sounds like they had us in mind with the legacy mode, huh?  Maybe we could just do that where needed and then we can go back and re-write the doctests at leisure.

StanczakDominik:
Actually I just figured out how to semi-quickly rewrite this in pycharm. I guess I'll just do it up.

PR #712, opened 2019-10-25
Repository housekeeping varia, python 3.8 tests

### LINKED ISSUE #424: Test against minimum versions of required packages
We have defined minimum versions for many of our required packages in our `requirements` directory.  However, to the best of my knowledge we do not explicitly test against the minimum versions to make sure that PlasmaPy works with them.  It would be really helpful to add a test to Travis CI or AppVeyer that specifically tests the minimum supported versions as listed in those files.

One thing that we should be careful of is that the version on PyPI for the `roman` package differs from the `roman.__version__` attribute.

It would be most helpful to set things up so that the minimum versions are taken directly from `automated-code-tests.txt` rather than us having to type in and update the minimum versions somewhere else manually.

Thanks!

### LINKED ISSUE #635: Add python 3.8 to testing matrix

StanczakDominik:
This implements https://github.com/OpenAstronomy/azure-pipelines-templates for CI.

Closes #635, closes #424.


StanczakDominik:
The py37-astropydev error is a result of #653 and I'm still not sure how to handle it

I will also add #424 here

StanczakDominik:
- [ ] `py36-minimal-conda [linux])` is failing due to calls to `units.quantity.allclose`, which either wasn't there yet or was in another namespace back in Astropy 2.0. But I think simply bumping the minimum to 3.0, or even 3.1 as I seem to remember some breakage, would be more sensible.
- [ ] `py37-astropydev [linux]` is all #653, and I think it could be possible to fix most of it via `atol`s. I'll try it tomorrow or so.

I'm also liking how local access to these envs is more comfortable now.

StanczakDominik:
Also, @Cadair, consider this tested and I've got only positive feedback for you :) it seems to be working like a charm.

StanczakDominik:
I've decided to bump the minimum required versions of

- Astropy from 2.0 to 3.1 because Seriously, That's One Major Version Ago and Let's Respect Each Other
- Numpy from 1.13 to 1.14 because https://github.com/PlasmaPy/PlasmaPy/issues/229 And That Should Be Enough Said

namurphy:
As much as I'll miss broken doctests, bumping the minimum version of NumPy to 1.14 sounds like a good idea to me.  Same for bumping minimum Astropy to 3.1.  Thanks!

StanczakDominik:
Turns out the astropydev problem consists of a bunch of different problems and I'd much rather merge this now, then figure out fixes for each...

StanczakDominik:
ELLIPSIS seems broken within astropy/pytest-doctestplus, I'll try to report that there, but most definitely not tonight.

For now, I'm instead going to get rid of pytest-astropy, as I'm realizing we're not actually using any of its functionality except FLOAT_CMP, and that can be reproduced with `pytest >= 5.1` via the NUMBER flag.

I'm frustrated, so I shall now vent with a witty video:

https://www.youtube.com/watch?v=djZkTnJnLR0

StanczakDominik:
https://www.youtube.com/watch?v=AXwGVXD7qEQ

except for 3.8 failing on pip installing hdf5, and astropy-dev is still failing, but that's still PROGRESS!

Cadair:
I *think* you might be able to solve your hdf5 issue by installing the hdf5 c lib with apt.

StanczakDominik:
I think I'll just take the lazy route and wait for https://github.com/h5py/h5py/issues/1410 :laughing:

Cadair:
That looks like a troublesome bug in the Azure images...

docs/install.rst
```diff
@@ -19,9 +19,9 @@ PlasmaPy requires Python version 3.6 or newer, and is not compatible
 with Python 2.7.  PlasmaPy requires the following packages for
 installation:
 
-- `NumPy <http://www.numpy.org/>`_ 1.13 or newer
+- `NumPy <http://www.numpy.org/>`_ 1.14 or newer
```
namurphy: And there was much rejoicing

PR #716, opened 2019-10-30
Style improvements

flaixman:
This PR doesn't solve bugs, but improves the actual style guide of python3 in some modules using pylint3 as a standard score.

Great project!


StanczakDominik:
Looks good, I went ahead and added a changelog entry :)

Issue #606, opened 2019-02-21
Create package for plasmapy pytest helpers?


namurphy:
Astropy has a few pytest helper packages to help with performing their test suite.  A longer term possibility for us would be to create a new package that contains functionality that simplifies the process of setting up our most common test patterns and makes our tests easier to read and maintain.  Some of the advantages of doing this are that it would help keep our core package lean and shorten the time it takes our tests to run since we won't have to test our test helpers.

My position is that test helper functions should be higher priority because the more tests we write without using them, the more tests we'll have to refactor later on (i.e., we'll accumulate technical debt).

namurphy:
For reference later on in case we need it, here are some of our most common test patterns.  Our most common one is:
```Python
@pytest.mark.parametrize(...)
def run_test_simplified(func, args, kwargs, expected_outcome):
    assert func(*args, **kwargs) ≈≈ expected_outcome
```
The `≈≈` means that it's not just testing equality as with `==` but rather it could be approximately equal to, or the two things could both be `numpy.nan` (which painfully does not equal itself), or the `expected_outcome` could be an exception or warning.  The `≈≈` encompasses a lot of special handling.

Another pattern that comes up fairly often in `plasmapy.atomic` is a function to check that a `callable` with one set of positional and keyword arguments provides an equivalent result to another `callable` with a different set of positional and keyword arguments.  For `plasmapy.physics` there is another function that asserts that functions can handle `numpy.array` objects.  I'm working on another helper function in #605 that checks that class attribute references produce the expected results (which has some overlap in functionality with the first test pattern).

If any of us notice common test patterns, it would be really helpful to raise an issue and reference this issue).  In some cases, pytest fixtures will likely reduce code duplication too.

namurphy:
This may end up being useful in the long term, but for now keeping all of the testing functions in the main package will probably be the best approach.  I'll close this for the time being, though we may want to revisit this if our test helper functionality ends up becoming large.

PR #728, opened 2019-12-05
Refactor test helpers

namurphy:
Background from a [previous comment](https://github.com/PlasmaPy/PlasmaPy/pull/605#issuecomment-480386397):

> Many of PlasmaPy’s tests have a simple premise. We check that providing a function or method with a set of positional and keyword arguments leads to the expected outcome. The situation becomes more complex with the realization that the expected outcome may be a string, float, array, or quantity with units. We may expect a specific exception to be raised or a warning to be issued. We may be expecting an infinity or NaN that may or may not have units, and have to take into account that `numpy.inf == numpy.inf` and `numpy.nan == numpy.nan` both evaluate to `False`. We might want to check if values or elements of an array are approximately equal to within a specified tolerance. In all of these cases, we need descriptive error messages that can help new contributors quickly pinpoint what the problem is. Accounting for all of these situations leads to repeated boilerplate code that makes tests harder to write, read, and maintain. This in turn creates an unnecessary barrier to entry for new contributors, especially as unit testing and continuous integration testing are not widely used in much of the plasma physics community.

We currently have a test runner function, [`plasmapy.utils.run_test`](https://github.com/PlasmaPy/PlasmaPy/blob/9ba3f4eb5dbbb1a3d7b25527e0d5eb62c5086edf/plasmapy/utils/pytest_helpers/pytest_helpers.py#L83), that is designed to simplify the process of writing tests by consolidating the functionality needed to compare the actual outcome with the expected outcome.  This function is quite helpful, but it contains a whole bunch of functionality in one place which makes it hard to maintain or extend.  There's also repeated code between `run_test` and `run_test_equivalent_calls`.  

My goal here is to refactor our test runner capabilities to break up the responsibilities in `run_test` into separate classes, each with a single responsibility.  Making it modular like this would make it easier to reuse code, since some of the tests are similar but not quite the same.  As a first step, I am creating classes to represent the expected an actual test outcomes.

Cross-references: #278, #605, #606, #674


rocco8773:
I suggest that this PR should also try to satisfy issue #746

namurphy:
> I suggest that this PR should also try to satisfy issue #746

Agreed!  I was thinking exactly that when I saw #746.

namurphy:
After too long and too many distractions, this PR is finally ready for review.   Starting a review in the revised section of `docs/development/testing_guide.rst` should give an idea of the motivation and reasoning.  The highest-level functions are in `runners.py`.  

The changes I made are:

 - I created the `plasmapy.tests.helper` module to host the test helper functionality.  The pytest dependency will therefore be isolated to `plasmapy.tests`.  We therefore would not need to require pytest as a hard dependency if we don't automatically import `plasmapy.tests` (related to #737 and #738).
 - I moved functions to assist with formatting error messages to `plasmapy.utils.formatting` because they have more general uses and do not depend on pytest.  Some changes got propagated through to `plasmapy.atomic` and elsewhere.
 - I created classes in `plasmapy.tests.helper.inputs` to take care of storing/validating the function or class attribute/method and the positional and keyword arguments.  The test is run using the `call` method for each of these classes.
 - I created the `ExpectedTestOutcome` class to store the expected test outcome and allow access to its different properties (`expected.py`).
 - I created the `ActualTestOutcome` class to store the true outcome of the test and allow access to its different properties (`actual.py`).
 - I created the `CompareActualExpected` class to compare instances of `ActualTestOutcome` and `ExpectedTestOutcome`, which sometimes uses the `CompareValues` class to compare values (rather than warnings or exceptions).  
 - Exceptions for tests are stored in `exceptions.py` with names that are meant to be descriptive and give users/developers a good idea of what went wrong.  (Suggestions on better names are welcome.)
 - I removed mutable default arguments in the original pytest helper functions (#752) even though these are likely to be removed soon.  

The things left to do are adding examples to docstrings (especially `runners.py`), update the changelog, look through the code coverage results to see which tests are missing, and hopefully make sure that the tracebacks/error messages are not too cluttered.

I tried to follow [my own advice on writing clean code](https://zenodo.org/record/3491142) by using descriptive names, writing short functions/methods that do exactly one thing, having the code read like a top-down narrative, and writing classes that each have a single responsibility.  At the very least, it's cleaner than other code I've written!

In retrospect, I should have broken this up into multiple smaller PRs, like one for each of the main classes.  Unfortunately, it took me a while to figure out how to fit the different pieces together.

Closes #746

StanczakDominik:
Wow, you even added a reviewer's manual, that's awesome :D

I think it's fine to do huge PRs for complex multicomponent pieces of code that need to work together, when you really need to start tinkering with it before you see the optimal way to achieve a goal, as long as the resultant code is clean and well described. Seems like that's going nicely!

I should be able to find some time to get started on reviewing this after Thursday, and I think I should be able to finish by next Tuesday.

StanczakDominik:
I started looking through the new code and to be honest I'm having issues grasping the scope of it.

Do you think you could try using the new functionality to simplify some of our existing tests, say, somewhere in particles (not all of them!) and maybe somewhere in the formulary? I think that'd make this much more understandable.

plasmapy/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import helper
```
rocco8773: (*low hanging fruit* 🤣) I feel this sub-package would be more appropriately named as `plasmapy.tests.helpers`. There's more than one helper contained here.

namurphy: Yeah, I was following Astropy's lead since they have `astropy.test.helper`.  I personally would prefer `plasmapy.test.helpers` so we can go with that. :joy_cat:

rocco8773: Or, we could spend another 2 hrs in the next meeting debating the singular vs plural. 🤣

plasmapy/tests/helper/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+
+from plasmapy.tests.helper.exceptions import (
+    Failed,
+    UnexpectedResultError,
+    InvalidTestError,
+    MissingWarningError,
+    MissingExceptionError,
+    UnexpectedExceptionError,
+    InconsistentTypeError,
+)
```
rocco8773: Try running this through [isort](https://pypi.org/project/isort/) and see how well it organizes the imports.

rocco8773:
I think we should have some ``.. automodapi::`` documentation for the classes and functions defined here.  It would probably help contributors learn how to use the code.  I'm not not sure where to place it at the moment though.

namurphy:
This pull request got too large, so I split it up into a bunch of different smaller pull requests: #919, #920, #922, #923, #924, #925, #926, #927, #928, and #929.  The development history that went into those different pull requests is still contained within this one.

Assuming I'm not forgetting anything, the only thing in this pull request that hasn't yet been put in a new pull request is the refactoring of `test_atomic.py` so that it uses the new `plasmapy.tests.helpers.test_runner` instead of the old `plasmapy.utils.pytest_helpers.run_test`.

Issue #723, opened 2019-11-26
Add docstrings to decorator `angular_freq_to_hz`


rocco8773:
Decorator [`plasmapy.utils.decorators.converter.angular_freq_to_hz`](https://github.com/PlasmaPy/PlasmaPy/blob/9ba3f4eb5dbbb1a3d7b25527e0d5eb62c5086edf/plasmapy/utils/decorators/converter.py#L13) needs docstrings.

JvPy:
I would like to take that

rocco8773:
@JvPy Have at it! Let us know if you have any additional questions.

rocco8773:
closed by #729

PR #729, opened 2019-12-11
Add docstring to converter.py

JvPy:
Added docstrings to  plasmapy/utils/decorators/converter.py
angular_freq_to_hz method

#723

- [ ] If adding new functionality, I have added (passing) tests and
- [ ] I have fixed any new failing tests (if you're unsure why
  they're failing, ask!).

### DIFF
--- a/changelog/729.doc.rst
+++ b/changelog/729.doc.rst
@@ -1,1 +1,1 @@
-Add docstrings to decorator :func:plasmapy.utils.decorators.converter.angular_freq_to_hz
+Add docstrings to decorator :func:`plasmapy.utils.decorators.converter.angular_freq_to_hz`
--- a/changelog/729.doc.rst
+++ b/changelog/729.doc.rst
@@ -1,1 +1,1 @@
-Add docstrings to decorator :func:`plasmapy.utils.decorators.converter.angular_freq_to_hz`
+Add docstrings to decorator :func:`plasmapy.utils.decorators.converter.angular_freq_to_hz`.
--- a/changelog/729.doc.rst
+++ b/changelog/729.doc.rst
@@ -1,0 +1,1 @@
+Added docstrings to the file `PlasmaPy/plasmapy/utils/decorators/converter.py`.
--- a/changelog/729.doc.rst
+++ b/changelog/729.doc.rst
@@ -1,1 +1,1 @@
-Added docstrings to the file `PlasmaPy/plasmapy/utils/decorators/converter.py`.
+Add docstrings to decorator :func:plasmapy.utils.decorators.converter.angular_freq_to_hz
--- a/plasmapy/utils/decorators/converter.py
+++ b/plasmapy/utils/decorators/converter.py
@@ -26,7 +26,7 @@ def angular_freq_to_hz(fn):
     ValueError
         fn.__name__ can not be `to_hz`
 
-    Returns 
+    Returns
     -------
 
     _result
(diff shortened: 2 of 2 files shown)


JvPy:
Need help with the test, cant make it pass :(

plasmapy/utils/decorators/converter.py
```diff
@@ -11,6 +11,28 @@
 
 
 def angular_freq_to_hz(fn):
+    """
+    Converts angular frequency (ω) frenquency to Hertz(hz)
```
rocco8773: This needs to be more descriptive of what the decorator is doing.  Something along the lines off...

"A decorator that adds to a function the ability to convert the function's return from angular frequency (rad/s) to frequency (Hz).

A `kwarg` `to_hz` is added to the function's signature, with a default value of `False`.  The keyword is also added to the function's docstring under the **Other Parameters** heading."

plasmapy/utils/decorators/converter.py
```diff
@@ -11,6 +11,28 @@
...
+
+    Parameters
+    ----------
+
+    fn : obj
+        The object to be converted
```
rocco8773: Change to...

``` python
"""
fn:
    The function to be decorated
"""
```

plasmapy/utils/decorators/converter.py
```diff
@@ -11,6 +11,28 @@
...
+
+    Raises
+    ------
+
+    ValueError
+        fn.__name__ can not be `to_hz`
```
rocco8773: Change to...
```python
""" 
ValueError
    If `fn` has already defined a kwarg `to_hz`
"""
```

plasmapy/utils/decorators/converter.py
```diff
@@ -11,6 +11,28 @@
...
+
+    Returns
+    -------
+
+    _result
+        result in Hertz(hz)
```
rocco8773: Change to...
``` python
"""
callable
    The decorated function
"""
```

plasmapy/utils/decorators/converter.py
```diff
@@ -11,6 +11,28 @@
...
+
```
rocco8773: We need to add some examples here.  It would be nice if they would work with doctests too.  The following might work, but you'd have to play with it

```python
import astropy.units as u
from plasmapy.utils.decorators.converter import angular_freq_to_hz

@angular_freq_to_hz
def foo(x):
    return x

foo(5 * u.rad / u.s)
<Quantity 0.79577472 Hz>
```

changelog/729.doc.rst
```diff
@@ -0,0 +1 @@
+Added docstrings to the file `PlasmaPy/plasmapy/utils/decorators/converter.py`.
```
rocco8773: Change to...

Add docstrings to decorator :func:`plasmapy.utils.decorators.converter.angular_freq_to_hz`

plasmapy/utils/decorators/converter.py
JvPy: Added the doctest as requested, but im not sure if what i did was enough.

Should i add another cases to this tests?

rocco8773:
@JvPy I made a few changes including...

1. a few stylistic changes to be consistent with the rest of the package
1. adding an example for decorating a method
1. a note about using the decorator with decorator [`validate_quantities`](https://github.com/PlasmaPy/PlasmaPy/blob/9ba3f4eb5dbbb1a3d7b25527e0d5eb62c5086edf/plasmapy/utils/decorators/validators.py#L378)

Otherwise, everything looks good. I just want to confirm how the docs look before merging.

Good job!

rocco8773:
closes #723

PR #742, opened 2020-01-28
Rename plasmapy.atomic to plasmapy.particles

namurphy:
As part of the changes associated with the forthcoming PLEP 7 (https://github.com/PlasmaPy/PlasmaPy-PLEPs/pull/26), we decided to rename `plasmapy.atomic` to `plasmapy.particles`.  One of the primary reasons for this is that this subpackage includes not just atomic data but also data and representations for fundamental particles.  Needless to say, this PR is a huge API chance.   

 - We used `particles` as a variable in a bunch of different places.  I went through and changed these to avoid namespace conflicts.  
 - I renamed `ion_particle` to `ion` in a few places in `collisions.py`, `braginskii.py`, and related files.
 - I used PyCharm to refactor and rename `plasmapy.atomic` to `plasmapy.particles`.  This amazingly almost everything that was necessary in the code, and I only had to fix one test failure.
 - I went through the docstrings more manually, as we use the word "atomic" in a whole bunch of places and I wanted to make sure there weren't any unnecessary changes.
 - I went through the `docs` directory more carefully too, and changed a few section titles.  Hopefully the docs build okay...
 - I made good use of `grep` to find any lingering places where atomic was used.  I'm hugely grateful for Software Carpentry teaching me this command.

The rebase associated with #728 might be a pain since I had to update a bunch of files in `plasmapy.atomic` in small ways while rearranging some formatting things in `plasmapy.utils`. In order to (possibly? hopefully?) make rebasing easier, I'd suggest not doing a squash merge on this one so that git can decipher all of the little changes separately.   Wish me luck!  Or perhaps more specifically, wish me *good* luck....


StanczakDominik:
What you could do here is merge this directly into the tester helpers pr first, then squash merge this, then squash merge the latter. I think git will handle it.

Also grep is great! Ripgrep is a bit nicer on the interface and faster, but that's personal preference. Glad you're picking these tricks up!

namurphy:
> What I'd like to suggest is keeping plasmapy.atomic around for one release (less than a year this time!) as an alias to particles that also DeprecationWarns. We actually have users for this part (ping the illustrious @wtbarnes ) and it'd be rude not to, I think.

Yes, that is a real issue.  I raised an issue in the fiasco repository (https://github.com/wtbarnes/fiasco/issues/52), and offered to submit a PR there to account for this API change after it gets included in a release.  There might be one or two more repositories too.  There has not yet been an official release of fiasco so the PR could suffice.  @wtbarnes, if there's an approach to this that would work particularly well for you, please let us know!

Since `atomic` is not the only subpackage that's getting renamed in our reshuffle, perhaps we could release 0.2.1 with deprecation warnings for all of the things we're changing.  Alternatively, we could do something like the following in `__init__.py`:

```Python
import particles
import particles as atomic
import formulary
import formulary as physics
```
...though we would want to had DeprecationWarnings in there too.

namurphy:
> What you could do here is merge this directly into the tester helpers pr first, then squash merge this, then squash merge the latter. I think git will handle it.

🙀

StanczakDominik:
Admittedly I don't worry about formulary as much because I'm not aware of users for those yet - and if there are any, I'd imagine surface level data analysis to be easier to change than libraries e.g. fiasco.

wtbarnes:
I sincerely appreciate the effort to accommodate fiasco but I think breaking changes between minor versions pre 1.0 is perfectly acceptable. By depending on a pre-1.0 package like plasmapy, fiasco is accepting the risk that things might break and that is ok! And as you say, fiasco has not even had a formal release yet so changing things up to accommodate upstream issues is not a problem at all. I’m happy to do it. I would recommend just doing whatever makes your life as a developer easiest! Thanks for keeping me in the koop!

namurphy:
Thank you for your flexibility and understanding, @wtbarnes!  We are trying to get breaking changes done sooner rather than later.

I'll put the question about merging this on the agenda for next week's community meeting.

StanczakDominik:
Or, since @wtbarnes is being gloriously *agile* in mindset about this (and this is the first time I use the term non-sarcastically), we could simply do the brute force renaming :)

PR #751, opened 2020-02-02
Remove mutable arguments from `Particle.is_category` method

Ishinomori:
I noticed that there are mutable arguments in `Particle.is_category` method, that should never be done.
Nice article about its 'features' --> https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

So the mutable arguments are removed and replaced by `None` with handling by `become_set` similarly as it was before.

### DIFF
--- a/changelog/751.trivial.rst
+++ b/changelog/751.trivial.rst
@@ -1,0 +1,1 @@
+ Remove mutable arguments from `Particle.is_category` method.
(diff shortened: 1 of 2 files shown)


namurphy:
Thank you!  Looks good to me too, so I'll merge it.  I'll create a broader issue for us to check if there are mutable default arguments elsewhere in the code.

PR #754, opened 2020-02-04
Remove mutable default arguments

### LINKED ISSUE #752: Check for and remove any mutable arguments in functions & methods
Callable objects [should not have mutable default values](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) (like lists, sets, and dictionaries) for functions and methods because unexpected behavior like this can happen:
```
In [2]: def f(a=[]): 
   ...:     a.append(0) 
   ...:     print(a) 
   ...:

In [3]: f()
[0]

In [4]: f() 
[0, 0]
```
We should check for mutable arguments in functions and methods, and then remove any that we do find.  The most likely place for this is in what will soon be `plasmapy.particles` (`plasmapy.atomic` until #742).  It may show up a bit also in `plasmapy.formulary`, perhaps `plasmapy.utils`, and maybe elsewhere.

Thanks to @Ishinomori for bringing this up and making a fix to this in #751!

Ishinomori:
Closes #752.

Every occurrence of mutable default argument mentioned in #752 is now replaced with `None` and proper handling. I ignored `plasmapy/utils/pytest_helpers/pytest_helpers.py` and `plasmapy/utils/pytest_helpers/error_messages.py`


plasmapy/utils/datatype_factory_base.py
```diff
@@ -160,7 +163,7 @@ def register(self, WidgetType, validation_function=None, is_default=False):
                 if hasattr(WidgetType, vfunc_str):
                     vfunc = getattr(WidgetType, vfunc_str)
 
-                    # check if classmethod: stackoverflow #19227724
+                    # check if classmethod: stackoverflow #19227g724
```
namurphy:
```suggestion
                    # check if classmethod: stackoverflow #19227724
```
I think an extra character slipped in here.

StanczakDominik:
```suggestion
                    # check if classmethod: stackoverflow #19227724
```
Either typo or SABOTAGE, and the first one is much more likely ;)

namurphy:
There's one small typo but as soon as that's done then I will merge it.  Thank you again!

plasmapy/utils/datatype_factory_base.py
StanczakDominik: @namurphy beat me to it :laughing:

namurphy:
Update: I'll merge this now, and fix the typo in #742 when I do a rebase.

Issue #687, opened 2019-09-26
Create a class hierarchy for particles


namurphy:
Right now, ions, electrons, leptons, baryons, and so on are all represented by instances of the `plasmapy.atomic.Particle` class.  This class contains a method, `Particle.is_category`, and can be used to tell whether or not a `Particle` instance is a particular category, like:
```Python
>>> electron = Particle('e-')
>>> electron.is_category('lepton')
True
>>> electron.is_category('baryon')
```
My thinking is that it would be more Pythonic to be able to do something like:
```Python
>>> isinstance(electron, Lepton)
True
>>> isinstance(proton, Baryon)
False
```
We could keep `Particle` as the factory for particles, and define `GenericParticle` as the abstract base class for particles.  Then we could have `Lepton`, `Baryon`, `Ion`, `NeutralAtom`, `AntiParticle` and whatever else inherit from `GenericParticle` and each other, as necessary.  `CustomParticle` as described in #447 would inherit from `GenericParticle` too.

A potential downside is that we would probably end up using multiple inheritance, which can be messy to modify.  With that said, the particle hierarchy exists in nature, so it would be sensible to reflect this in the code.  This idea warrants further discussion before we possibly implement it, so we should think of this as a proposal for the time being.

namurphy:
Oh...another benefit from doing this is that we would be able to do things like this:

```Python
@particle_input
def ion_inertial_length(ion: Ion, ...):
    return d_i
```

We could then have `@particle_input` enforce that `ion` is in fact an `Ion`.  Right now I think we do that a little bit based on what the argument's name is (i.e., `@particle_input` will check that arguments named `ion` are actually ions, unless I'm misremembering).  

We probably would also want `ChargedParticle`, `NeutralParticle`, and `NeutralAtom` classes in the hierarchy too.

goodab:
I agree that the current implementation of `Particle` is pretty dense and hard to extend. 

A pretty simple class hierarchy with `Particle` as the base class that `SubatomicParticle` and `ElementParticle` both subclass and an additional subclass `Isotope` to `ElementParticle`, handles most of the shared functionality without much code duplication. It runs into an issue with defining the proton, since it is both a `SubatomicParticle` and an `Isotope`, but that can be fixed with a specific `Proton` subclass with multiple inheritance.

An alternative to the class structure is to use composition for the particle properties. This makes the proton edge case more attractive at the cost of making everything else clunkier.

[...]

PR #755, opened 2020-02-04
Create abstract and custom classes for particles

### LINKED ISSUE #447: Create custom particle subclass

namurphy:
My goals here are:

 - Create an abstract particle class
 - Have `Particle` inherit from that abstract class
 - Create a class for dimensionless particles that inherits from the abstract class
 - Create a class for custom (dimensional) particles that inherits from the abstract class

Thus far it looks like the classes will include `mass` and `charge` attributes.  I'm also wondering about creating a composite particle class, which would have things like the root mean square charge which comes up occasionally in `plasmapy.formulary`.

Closes #447.  Related to #687.  This pull request is brought to you by proudly procrastinating doing other things.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -3,7 +3,7 @@
 import warnings
 from typing import Union, Set, Tuple, List, Optional
 from collections import defaultdict, namedtuple
-from numbers import Integral, Real
+from numbers import Integral, Real, Complex
 
 import numpy as np
 import astropy.units as u
@@ -1661,7 +1661,6 @@ class DimensionlessParticle(AbstractParticle):
     """
 
     def __init__(self, *, mass: Real = None, charge: Real = None):
-
         try:
             self.mass = mass
             self.charge = charge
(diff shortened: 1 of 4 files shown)


StanczakDominik:
> This pull request is brought to you by proudly procrastinating doing other things.

For those about to postpone doing other work by writing some cool abstract classes, we'll salute you... later.

plasmapy/particles/particle_class.py
```diff
@@ -1585,3 +1605,185 @@ def recombine(self, n: Integral = 1, inplace=False):
...
+        if isinstance(q, Real):
+            self._charge = q
+        elif isinstance(q, u.Quantity) and q.unit is u.dimensionless_unscaled:
+            self._charge = q.to(u.dimensionless_unscaled).value
+        elif q is None or q is np.nan:
+            self._charge = np.nan
+        else:
+            raise InvalidParticleError(
+                "The charge of a dimensionless particle must be a real number."
+            )
```
StanczakDominik: I'd have thought dumping non-dimensionless Astropy Quantities via `.si.value` or as multiple of electron charge would have been in the cards. Maybe that's not explicit enough, though.

plasmapy/particles/particle_class.py
```diff
@@ -1585,3 +1605,185 @@ def recombine(self, n: Integral = 1, inplace=False):
...
+
+    # TODO: Add a root mean square charge attribute.
```
StanczakDominik: Okay, now I'm curious. What's that for?

plasmapy/particles/particle_class.py
```diff
@@ -105,7 +110,23 @@ def _category_errmsg(particle, category: str) -> str:
...
+
+    @property
+    @abstractmethod
+    def mass(self) -> Union[u.Quantity, Real]:
+        pass
+
+    @property
+    @abstractmethod
+    def charge(self) -> Union[u.Quantity, Real]:
+        pass
```
StanczakDominik: I think I used `particle.name` in `RefactorTracker` (hope you don't mind I call it by branch name, it's simpler for me that way - I've been checking it out a lot lately). Should I switch to `str(particle)`?

plasmapy/particles/particle_class.py
namurphy: Hm...I think that actually may be out of scope now.  The formulary includes some functions that have `z_mean` as a keyword argument in order to account for composite particles.  I had thought there were functions that allowed for `z_rms` to different from `z_mean` but I couldn't find them just now.  For the time being, I'd say we can leave `z_rms` out until we have a need for it.

namurphy:
I think this is just about ready to merge.  Two thoughts:
 - I am wondering if I could replace some of the validation steps with `validate_quantity`.  One potential barrier to this is that the current implementation for a custom particle accepts real numbers for charge, in which case it's treated as the integer charge.
 - We can merge this after #757 so that we can do the rebase in this PR, since I reformatted the file with black which will probably create conflicts.

namurphy:
- [ ] Add change log

plasmapy/particles/particle_class.py
```diff
@@ -478,15 +499,18 @@ def __eq__(self, other) -> bool:
                 return self.particle == other_particle.particle
             except InvalidParticleError as exc:
                 raise InvalidParticleError(
-                    f"{other} is not a particle and cannot be "
-                    f"compared to {self}.") from exc
+                    f"{other} is not a particle and cannot be " f"compared to {self}."
```
rocco8773: make a single f-string

namurphy:
As a side note, @StanczakDominik mentioned during code review that [hypothesis](https://hypothesis.readthedocs.io/en/latest/) may be a useful library for testing in `plasmapy.particles` and `plasmapy.formulary`

plasmapy/particles/particle_class.py
```diff
@@ -98,14 +88,38 @@ def _category_errmsg(particle, category: str) -> str:
...
-class Particle:
+class AbstractParticle(ABC):
+    """
+    An abstract base class that defines the interface for particles.
+    """
+
+    @property
+    @abstractmethod
+    def mass(self) -> Union[u.Quantity, Real]:
+        pass
```
rocco8773, line 107:
```suggestion
        raise NotImplementedError
```

plasmapy/particles/particle_class.py
```diff
@@ -98,14 +88,38 @@ def _category_errmsg(particle, category: str) -> str:
...
+
+    @property
+    @abstractmethod
+    def charge(self) -> Union[u.Quantity, Real]:
+        pass
```
rocco8773, line 112:
```suggestion
        raise NotImplementedError
```

PR #761, opened 2020-02-13
Reformat most subpackages with black code formatter

namurphy:
A few of us have been (at least occasionally) using [black](https://black.readthedocs.io/en/stable/) (the "uncompromising code formatter") to reformat our code.  If we use black while changing a file that was not pre-formatted with black, then our diffs can get kind of lengthy (though @StanczakDominik pointed out that there is an option to look at only a range of commits using GitHub).  I thought that reformatting with black now rather than later would simplify future diffs (which in itself is a goal of how black formats code).  I tried to avoid files that were being changed in active pull requests so as to minimize the need for rebasing.  I made no changes besides reformatting.

If we do accept this pull request, it would effectively make black our default code formatter.  In that case, we should change the pep8speaks message to refer to black and add black to our development guide too.


StanczakDominik:
I want that 0.02 percent coverage back, yo. : laughing:

In all seriousness, this is a good call; I'm pretty sure it'll need to be followed by infrastructural fixes: besides updating pep8speaks, this would mean adding styling to CI, updating pep8... In other words, there's stuff to pick up from SunPy once again!

PR #763, opened 2020-02-18
Handle ModuleNotFoundError when importing __version__ from version.py

rocco8773:
During development we code without `plasmapy` installed in our python environment.  Since `plasmapy/version.py` is generated by `setuptools_scm`, this `version.py` file is not generated and throws a `ModuleNotFoundError` when trying to import `__version__`.

Here, I just add a `try-except` statement with a warning to handle this.  If a version is not found, then `__version__` is set to `''`.  I chose this route so we do not have to manage a hard coded fallback version.


rocco8773:
@namurphy Do you think this needs a change-log?  I don't think it necessary.  If you agree, go ahead an merge.

StanczakDominik:
It is new (better!) behavior. Someone may eventually want to find where it originated from.

> In the face of ambiguity refuse the temptation to guess.

I'd be in favor of changelog entries for all PRs. This kind of thing is what the (IIRC) `trivial` category is there for.

plasmapy/__init__.py
```diff
@@ -38,7 +38,19 @@
...
 # ----------------------------------------------------------------------------
-from .version import __version__
+try:
+    # ./version.py is generated by setuptools_scm during install
+    from plasmapy.version import __version__
+except ModuleNotFoundError:
+    # package is not installed
+    from warnings import warn
+
+    warn('plasmapy.__version__ not generated, PlasmaPy is not an installed package.',
```
StanczakDominik: "... or `setuptools_scm` is not installed."

namurphy:
> @namurphy Do you think this needs a change-log?

> I'd be in favor of changelog entries for all PRs. This kind of thing is what the (IIRC) `trivial` category is there for.

I'm okay with either having a changelog entry for this or not, though this would indeed fit nicely in the `trivial` category.  With that said, I would be fine with skipping change log entries for PRs with solely formatting changes and typo fixes, especially in the development phase.  I don't have a strong opinion though.

plasmapy/__init__.py
rocco8773: Honestly, I think we should require `setuptools_scm` or have a fallback version so whenever `plasmapy` is installed it has a version tag.  How to do this is another question. I'll have to think about it.

rocco8773:
@namurphy The changelog is already added.

Now it's a matter of deciding if we need to require `setuptools_scm`. I fundamentally think there's a problem if `plasmapy` can be installed and a version `plasmapy.__version__` is not generated.  This could happen if `setuptools_scm` is not installed when `plasmapy` is installed.

StanczakDominik:
I'm not happy with the current setup either. For now let's just use setuptools scm with it; I'll try to look at workarounds next week.

plasmapy/__init__.py
rocco8773: Since we've defined `setup_requires=setuptools_scm`, `setuptools_scm` will be downloaded for the install and stored in the `./egg` as per `setuptools` documentation.

> setup_requires
>
>    A string or list of strings specifying what other distributions need to be present in order for the setup script to run. setuptools will attempt to obtain these (using pip if available) before processing the rest of the setup script or commands. This argument is needed if you are using distutils extensions as part of your build process; for example, extensions that process setup() arguments and turn them into EGG-INFO metadata files.
>
>    (Note: projects listed in setup_requires will NOT be automatically installed on the system where the setup script is being run. They are simply downloaded to the ./.eggs directory if they’re not locally available already. If you want them to be installed, as well as being available when the setup script is run, you should add them to install_requires and setup_requires.)

rocco8773:
I've changed up the try-except statement a little.  Now, if `plasmapy` is not installed, then it'll try to defined `__version__` using `setuptools_scm`. If `setuptools_scm` is not installed, then the version will default to `0.0.0` with a warning.

rocco8773:
> Now it's a matter of deciding if we need to require `setuptools_scm`. I fundamentally think there's a problem if `plasmapy` can be installed and a version `plasmapy.__version__` is not generated. This could happen if `setuptools_scm` is not installed when `plasmapy` is installed.

[see comment about `setuptools_scm` being a`setup_requries`](https://github.com/PlasmaPy/PlasmaPy/pull/763#discussion_r384118913)

rocco8773:
As part of that issue, we should also figure out how to handle the `"version"` key in `codemeta.json`.  It's already out of date.

Issue #49, opened 2017-06-09
Create functionality for calculating dimensionless plasma parameters


namurphy:
We will soon need functionality to calculate dimensionless plasma parameters.  Obvious ones to include are:

- [ ] `Lundquist_number` (my favorite!)
- [ ] `magnetic_Reynolds_number` (note: similar to Lundquist number)
- [x] `Reynolds_number`
- [ ] `Prandtl_number`
- [x] `plasma_beta` (went in as `beta`)
- [ ] `Mach_number`
- [x] `Coulomb_logarithm`
- [ ] `Debye_number` (this may be better here than with `Debye_length` in the plasma parameters file)

There are a bunch more in the NRL Plasma Formulary.  Please add suggestions, as I'm undoubtedly missing lots of dimensionless numbers!  I'm also not sure what the best names are for things like the ratio of the gyroradius to the system size (`relative_gyroradius`?).

This functionality will need to be added after pull request https://github.com/PlasmaPy/PlasmaPy/pull/34 on plasma parameters and issue https://github.com/PlasmaPy/PlasmaPy/issues/35 on transport coefficients.  A possible file name would be `dimensionless.py`.  The docstrings for these functions should include descriptions of the physics behind them as well as caveats and considerations in using them.

Side note: an annoyance I have with the names of most of these dimensionless parameters is that they provide little insight into the physics that is being represented by them.  That said, I guess that the standard names are the ones most people are familiar with, so it's probably still best to go with them rather than use `turbulentness_number` instead of `Reynolds_number`.  However, I still think we should change "Rayleigh-Taylor instability" to "swirly mushroom instability".

namurphy:
```Python
def Douglas_Adams_number():
    return 42
```

LukeRoantree4815162342:
Unless somebody has already started work on this, I'd like to contribute here. I'll update each time I add functionality and be happy to coordinate with anyone else working on this to avoid duplication.
I'll be starting with ```Mach_number``` unless someone is already working on it

StanczakDominik:
Hey, I believe Nick has already done some work on this in #34 , #60 and possibly some others. I don't think we have the Mach number just yet, though, so if you could look over there, that'd be amazing! Thanks for the heads up!

namurphy:
I haven't worked on `Mach_number` yet so it would be great if you could!  I'm in the middle of working on `Coulomb_logarithm` at the moment.

lemmatum:
Some other ones:
[...]

PR #769, opened 2020-02-27
4 New dimensionless parameters Reynolds and Prandtl numbers (including magnetic)

diego7319:
This pull request will partially solve 
#49 
-->

- [x] I have added a changelog entry for this pull request 
- [ ]  magnetic Prandtl number (missing tests, need help)
- [ ]  Prandtl number (nothing yet)
- [ ] Reynolds number (nothing yet)
- [ ] magnetic Reynolds number (nothing yet)


rocco8773:
I see the benefit of adding `Prandtl` numbers, but I have a problem of providing functions that are only doing simple divisions.  As they currently stand, they do not provide any increase in functionality to the user.  The only reason I can see these being added to the code is if they do 1 or 2 of the following (1) provide some improved functionality in calculating the number (and its variations) or (2) provide a robust documentation to explain to/educate users about these numbers.

diego7319:
> I see the benefit of adding `Prandtl` numbers, but I have a problem of providing functions that are only doing simple divisions. As they currently stand, they do not provide any increase in functionality to the user. The only reason I can see these being added to the code is if they do 1 or 2 of the following (1) provide some improved functionality in calculating the number (and its variations) or (2) provide a robust documentation to explain to/educate users about these numbers.

Thanks for your reply, I just added more documentation that I think will give a clear explanation about what is that function for. And following reason 1, what other functionality do you think would be helpful in this case?  I think implementing the two formulas  from wikipedia and compute according the input variables?

diego7319:
> Thank you for working on this! I'd suggest having the Reynolds numbers in a different pull request. Once in a while I've had a multi-part pull request get delayed because problems came up in one of the parts, so I've usually found it better to split things up.
> 
> The use case for these functions will usually be some having basic plasma parameters (electron temperature, ion temperature, electron number density, ion number density, etc.) and wanting to calculate the dimensionless parameters corresponding to them. The arguments should be things like `T_i`, `T_e`, `n_i`, `n_e`, etc.
> 
> One difficulty with using the regular Prandtl number for us is that there are multiple thermal diffusivities in a magnetized plasma. The thermal diffusivity along a magnetic field line is usually significantly greater than the thermal diffusivity orthogonal to a magnetic field line. I'm not certain which thermal conductivity we should use, but for now let's go with the [electron thermal conductivity](http://docs.plasmapy.org/en/latest/api/plasmapy.formulary.braginskii.ClassicalTransport.html#plasmapy.formulary.braginskii.ClassicalTransport.electron_thermal_conductivity) from `ClassicalTransport` in `plasmapy.formulary.braginskii` since that's probably going to be the largest. For the viscosity, it's probably best to use the ion viscosity which I believe is also in `ClassicalTransport`.
> 
> It would be worth checking the units, since there are different ways to represent them. For comparing diffusivities, conductivities, and viscosities, it's generally best to have them in units of m²/s. The output of `Spitzer_resistivity` will probably need to be divided by `astropy.constants.mu0` in order to get into these units. The other ones might need to be divided by mass density.

In that case it seems that we could use one of two options, add this dimensionless parameters as a function inside the ClassicalTransport class using the internal variables->OR<- pass the entire object as an input variable (the reference for ClassicalTransport) in dimenssionless which is probably not pythonic. It would be something like this 

```
def prandtl_number(ClasicalTransport):
    t_1,t_2 = ClasicalTransport.get_t_1, ClasicalTransport.get_t_2
    return t_1/t_2
```

PR #773, opened 2020-03-24
Set isort settings to be consistent with black

namurphy:
The default settings for [isort](https://readthedocs.org/projects/isort/) (which sorts imports) and [black](https://black.readthedocs.io/en/stable/) (which strictly formats code) are not consistent with each other.  This sets the default settings for isort to be consistent with black.  If you run black and isort once, then switching between them afterward will lead to no changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -113,16 +113,21 @@ ignore = D202,D205,D302,D400,D105,D107,F401
 exclude = extern,sphinx,*parsetab.py,conftest.py,docs/conf.py,setup.py
 max-line-length = 99
 
-
-
 [coverage:run]
 omit =
     ci-helpers/*
     */tests/*
     plasmapy/setup_package.py
     plasmapy/version.py
+
 [coverage:report]
 exclude_lines =
     coverage: ignore
     ImportError
     ModuleNotFoundError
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -113,16 +113,21 @@ ignore = D202,D205,D302,D400,D105,D107,F401
...
+
 [coverage:report]
 exclude_lines =
     coverage: ignore
     ImportError
     ModuleNotFoundError
+
+[isort]
+# Set sorting of imports to be consistent with black formatting
+line_length=88
```
rocco8773: We should probably make this consistent with the `max-line-length` for [`pycodestyle`](https://github.com/PlasmaPy/PlasmaPy/blob/10287a3aeaf82fa1b1df2036e0971f1acdbd8ee8/setup.cfg#L92) and [`flake8`](https://github.com/PlasmaPy/PlasmaPy/blob/10287a3aeaf82fa1b1df2036e0971f1acdbd8ee8/setup.cfg#L114), which are currently set to 99.

StanczakDominik:
```suggestiion
line_length=99 
```

PR #771, opened 2020-03-01
Setup files cleanup

StanczakDominik:
This is a followup to #763 that removes some lint from setup.py and setup.cfg
and uses pkg_resources, which is provided with setuptools and which basically
everyone is going to have, to do what #763 did, but in less code. I also
managed to do without the version.py file, which should simplify things.

- [ ] I have added a changelog entry for this pull request (please see
  `changelog/README.rst` for instructions, and if you need help with picking the PR type, ask!)
- [x] If adding new functionality, I have added (passing) tests and
      docstrings. (Tests pop up at the bottom, in the checks box).
- [ ] I have fixed any new failing tests (if you're unsure why
  they're failing, ask!).

PR #774, opened 2020-03-29
Fix definition of __version__

rocco8773:
1. fix use of `pkg_resources.get_distribution()`...was incorrectly used as `pkg.resources`
1. add `setuptools` to package dependencies since `pkg_resources` is now a runtime dependencey and distributed through `setuptools`
1. add a soft-dependency on `setuptools_scm` so `__version__` is generated for developers when working from source
1. if `__version__` can not be defined then it's set to `'unknown'`

This is a followup on PRs #771 and #763


StanczakDominik:
Geez, we really can't get this right, can we? I'll take a look after breakfast!

plasmapy/__init__.py
```diff
@@ -41,11 +41,37 @@
...
+        del get_version
+        warn_add = 'setuptools_scm failed to detect the version'
+    except ModuleNotFoundError:
+        # setuptools_scm is not installed
+        __version__ = fallback_version
+        warn_add = 'setuptools_scm is not installed'
+
+    if __version__ == fallback_version:
+        from warnings import warn
+        warn(f"plasmapy.__version__ not generated (0.0.0), PlasmaPy is not an "
```
StanczakDominik:
```suggestion
        warn(f"plasmapy.__version__ not generated, PlasmaPy is not an "
```

plasmapy/__init__.py
```diff
@@ -41,11 +41,37 @@
...
+        __version__ = get_version(root='..',
+                                  relative_to=__file__,
+                                  fallback_version=fallback_version)
+        del get_version
+        warn_add = 'setuptools_scm failed to detect the version'
+    except ModuleNotFoundError:
+        # setuptools_scm is not installed
+        __version__ = fallback_version
+        warn_add = 'setuptools_scm is not installed'
+
```
StanczakDominik:
```suggestion
   warn_add += ". Consider using `pip install -e {plasmapy_directory_root}`"
```

rocco8773: I don't think I want to encourage this because the version info is only generated during the install, so `__version__` won't get updated with each successive commit.

StanczakDominik: How would you setup a dev environment, then?

rocco8773: I use PyCharm, so I'm basically sitting in the root of the repository.  If we'd want a continuously updated `__version__` in the editable package case, then `setuptools_scm` would have to be used.  However, I don't know if there's a way to indicated if a package is installed as editable.

StanczakDominik: Do we really need a continuously up-to-date version attribute, though? Seriously. I struggle to think of practical use cases for this. Besides, if you really need the version tag to update, you can run pip install again and that'll trigger the version rebuild.

rocco8773: No, I don't think we need `__version__` to be continuously update, that's why I added the note in the comments about this.  If some unforeseen scenario arises, then the information is at least there.

The try-statement will basically cover any distribution installed with `pip`.  The `setuptools_scm` in the except-statement, will cover programmers like me who work in the repository's root directory instead of installing.  It's likely there will not very many use cases for people like me.  The only one I can think of off the top of my head, is checking the `__version__` is properly incremented when the commit tag is defined.

Of course, I did discover this root repository metadata thing when I was getting a defined `__version__` when I knew there shouldn't be one.  This was because I still had metadata in the root directory from a previous editable install and the uninstall didn't remove the metadata.

PR #783, opened 2020-04-11
Create plasmapy.addons sub-package

rocco8773:
Creating the `plasmapy.addons` sub-package to enable 3rd party extensions for `plasmapy` via the use of `entry_points`.  `entry_points` is a apparatus built into `setuptools` to provide a "plug in" mechanism for extensible applications.

## Example

*Write an example and link to the example repository.*

## Allternative Approaches

*Other approaches (like native namespace packages) and why those routes were not taken*

----
**Note:**  The write here will be expanded upon.  This is to get the PR started.


plasmapy/addons/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+__readme__ = _parse_readme()
+
+
+# import addon entry points
+for ep in iter_entry_points('plasmapy.addons'):
+    # __dict__[ep.name] = ep.load()
+    globals()[ep.name] = ep.load()
+    __all__.append(ep.name)
```
namurphy, line 21: A possibility would be to create and call a private function called something like `_import_addon_entry_points`.

setup.py
```diff
@@ -18,6 +17,8 @@
 # Concatenate all the values together for 'all'
 extras["all"] = list(chain.from_iterable(ex_extras.values()))
 
+# packages = find_packages() + ['plasmapy.addons', ]
```
namurphy, line 20: An extra commented out line?

plasmapy/addons/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+try:
+    del ep
+except NameError:
+    pass
```
namurphy, line 26:
```suggestion
if "ep" in locals():
    del ep
```
This could be a shorter way of doing it.  I think it should be `locals()` and not `globals()`.  The original way is perfectly fine too, just verbose.

plasmapy/addons/__init__.py
rocco8773: Not sure if that is necessary, since this isn’t functionality we’d use anywhere else and isn’t functionality we’d want exposed to users.

plasmapy/addons/__init__.py
rocco8773: Since we’re at the module level here, `locals()` and `globals()` are the same thing.

If I can, I usually code by [EAFP](https://docs.python.org/3/glossary.html#term-eafp) instead of [LBYL](https://docs.python.org/3/glossary.html#term-lbyl).  In this case I don’t think it really matters though.

plasmapy/__init__.py
```diff
@@ -33,12 +33,36 @@
...
+from . import (
+    addons,
+    classes,
+    data,
+    diagnostics,
+    formulary,
+    particles,
+    simulation,
+    utils,
+)
```
StanczakDominik, line 65: IIRC (and this really should have been documented in here...) this automatically-import-every-subpackage style of doing things (as opposed to, e.g., the scipy way) was done in order to ensure dependencies for some subpackages are optional if you're not willing to use them. We've been better about moving optional deps into functionality code recently, though, and I've been meaning to crack down on the optional ones anyway (because matplotlib is not really difficult to install when you've seen VTK!), so I'm all for this anyway.

plasmapy/addons/__init__.py
```diff
@@ -0,0 +1,28 @@
+__all__ = []
+
+from pkg_resources import (iter_entry_points, resource_filename)
+
+
+def _parse_readme():
+    fname = resource_filename('plasmapy', 'addons/README.md')
```
StanczakDominik, line 7: I take it this handles wheel installs, etc? Nice! I'll be coming back here often :laughing:

plasmapy/addons/__init__.py
StanczakDominik: I agree with @rocco8773, this is fine :)

plasmapy/addons/__init__.py
StanczakDominik: Cool, new acronyms!

setup.cfg
```diff
@@ -36,6 +36,14 @@ install_requires =
   scipy >= 0.19
   setuptools >= 41.2
 
+[options.packages.find]
+exclude =
+  plasmapy.examples
+  plasmapy.examples.*
```
StanczakDominik, line 42: So we don't want `examples` to be importable, and this ensures that? Cool.

setup.cfg
```diff
@@ -36,6 +36,14 @@ install_requires =
...
+
+[options.entry_points]
+plasmapy.addons =
```
StanczakDominik, line 45: I guess this would sort of initialize an empty list of `addons` entry point packages?

setup.cfg
rocco8773: This will exclude `examples` from the distribution all together.  I discovered this when doing the native namespace packages.  Since we plan to move the `examples` to the root directory, I figured I leave this here.

However, I discovered a little while back that if we exclude examples, then it will be a little more difficult for users to get their hands on the example data (e.g. the traces for langmuir probes). I prefer the idea of not distributing example data with the main `plasmapy` distribution, but we should consider some mechanism that makes it easy for a user to get his or her hands on the data. Maybe something like `plasmapy.utils.get_example_data()`. ¯\\_(ツ)_/¯  Or, we could place the data in a separate distribution that could be an extra dependency during install, like `pip install plasmapy[examples]`.

setup.cfg
rocco8773: I debated about leaving this here.  It doesn't "initialize" the `entry_point`, in fact, it does nothing.  I was originally trying to have a simple `entry_point` example here, but I haven't come up with a clean looking one, and why would you create an entry point into your own package from your package. **I kept this here because it advertises the entry point.**

setup.py
rocco8773: Ha! I originally thought you were suggesting this and I was beyond confused.  This was left over from my playing around, it should not be here.

plasmapy/__init__.py
rocco8773: This will automatically import everything.  However, I'm doing this to control the namespace of `plasmapy`.  I think properly exposing the namespace makes the package much more discoverable for new and old users.  Someone can actually find stuff using `dir()`.  Before this `dir(plasmapy)` would show a bunch of dunder objects, now it'll show all the available sub-packages.

This does lead to longer package loading times, but I'm okay with this for the increased user friendliness.  To counter act these loading times (especially as the package gets larger), I say we encourage specific module imports in our documentation (e.g. `import plasmapy.addons as pads` or `import plasmapy.formulary as pform`).

plasmapy/addons/__init__.py
rocco8773: Yes it does.  I actually forgot about this. 😆

From [`pkg_resources` documentation](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resource-extraction)...

> resource_filename(package_or_requirement, resource_name)
>
>     Sometimes, it is not sufficient to access a resource in string or stream form, and a true 
>     filesystem filename is needed. In such cases, you can use this method (or module-level 
>     function) to obtain a filename for a resource. If the resource is in an archive distribution 
>     (such as a zipped egg), it will be extracted to a cache directory, and the filename within 
>     the cache will be returned. If the named resource is a directory, then all resources within 
>     that directory (including subdirectories) are also extracted. If the named resource is a C 
>     extension or “eager resource” (see the setuptools documentation for details), then all C 
>     extensions and eager resources are extracted at the same time.

PR #675, opened 2019-09-08
Refactoring of ParticleTracker, Plasma class

StanczakDominik:
Closes #636 

This is going to be a pretty open-ended one. I've got a few ideas I'd like to tinker with:

- [x] Remove field interpolation code from ParticleTracker, as right now it's completely tied to Plasma3d on a uniform grid. That's completely idiotic and [those responsible should be sacked](https://github.com/PlasmaPy/PlasmaPy/blame/master/plasmapy/simulation/particletracker.py).
- [x] Put field interpolation in Plasma3d, because that's reasonable.
- [x] Replace `plasma3d` with draft AnalyticalPlasma in ParticleTracker, which probably shares a whole bunch of functionality with PlasmaBlob
- [x] reenable some of the tests
- [x] optimize ParticleTracker
- [ ] merge PlasmaBlob with AnalyticalPlasma as needed

Leftovers for future work:
   
- [x] Experiment with xarray for Plasma3D - working with the bare numpy arrays is proving to be a pain.
     - [ ] test field interpolation in there.
- [ ] Use notes from #569

PR #790, opened 2020-04-22
Various administrative cleanups

StanczakDominik:
This PR:

* cleans tox.ini up a bit
* decreases the amount of test envs that get spun up when running simply `tox` to two from way too many
* adds isort to pre-commit hooks
* prevents conda from nagging us about lack of `pip` as a conda dep in `environment.yml`
* adds a neat PLASMAPY_PLOT_TEST env variable that can be set when running tests; this makes plots appear
* removes a reference to a sunpy tox-env we don't use from .circleci/config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,15 +33,6 @@ tox-conda-install: &tox-conda-install
   command: |
     pip install tox-conda
 
-image-run: &image-tests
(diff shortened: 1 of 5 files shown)


StanczakDominik:
I'm using [no changelog entry needed] here because these are just various trivial changes that came up during #675.

PR #796, opened 2020-05-04
Update classes & docs for ionization states and fix docstring formatting in plasmapy.particles

namurphy:
This pull request adds narrative documentation for `IonizationState` and `IonizationStates` and fixes some typos.


namurphy:
I'm also wondering if it would be better to use the name `IonizationStates` or `IonizationStateCollection`, as I've noticed the latter style to be used occasionally.  The latter style also provides more of a distinction when compared to `IonizationState`.

StanczakDominik:
I agree with renaming them. Having it be one letter of a typo away seems like a bad idea, come to think of them.

namurphy:
I just renamed `IonizationStates` to `IonizationStateCollection`.  I'm also changing the `State` `namedtuple` to the `IonicFraction` class (which gets used when iterating over an ionization state).  The tests that are broken right now are because I'm in the middle of that change.

I'm using ionization state (singular) to describe the ionic fractions for _a single element_, whereas ionization states (plural) refers to ionic fractions _for multiple elements_.

namurphy:
Following https://github.com/namurphy/PlasmaPy-NEI/pull/1, I've started thinking that it would be worthwhile to rewrite the tests for `IonizationState` and `IonizationStateCollection` using pytest fixtures.  I think that would end up being a lot cleaner than the class-based approach currently in use.  So that I'm less likely to forget...
 - [ ] Make an issue about whatever it was that I just said

namurphy:
Again, following https://github.com/namurphy/PlasmaPy-NEI/pull/1:
 - [ ] Raise an issue on looking into using xarray under the hood in `IonizationState` and `IonizationStateCollection`

changelog/796.doc.rst
```diff
@@ -0,0 +1 @@
+Add narrative documentation on ionization state functionality.
```
StanczakDominik, line 1: Didn't know towncrier would handle two snippets per PR. Seems to work nicely, though. TIL, thanks! :+1:

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,141 @@
...
+
+>>> ionization_state.Z_rms
+0.9899...
+
+The ``Z_most_abundant`` attribute returns a `list` of the most abundant
+ion(s).  The `list` may contain more than one integer charge in case of
+a tie.
+
+>>> ionization_state.Z_most_abundant
+[1]
```
StanczakDominik, line 46: I would assume that for real processes, that's pretty unlikely given that real quantities are usually `float`s?

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,141 @@
...
+The ``info`` method prints out the ionic fraction for the ions with
+an abundance of at least 1%.
+
+>>> ionization_state.info()
+IonizationState instance for He with Z_mean = 0.96
+----------------------------------------------------------------
+He  0+: 0.050
+He  1+: 0.940
+He  2+: 0.010
+----------------------------------------------------------------
```
StanczakDominik, line 57: I wonder if we could get more descriptive than `info()`. `summarize()`? `summary` as a property? `describe()`, following pandas?

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,141 @@
...
+
+>>> states.n
+<Quantity 5.e+19 1 / m3>
+>>> states.abundances
+{'H': 1.0, 'He': 0.08}
+>>> states.number_densities["H"]
+<Quantity [5.00e+17, 4.95e+19] 1 / m3>
+
+The ``info`` method may also be used to get a summary of the ionization
+states.
```
StanczakDominik, line 130: We should of course keep the name of this method consistent between the two classes if we decide to change one of them.

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,100 @@
...
+    try:
+        instance = IonicFraction(
+            ion=ion, ionic_fraction=ionic_fraction, number_density=number_density
+        )
+    except Exception:
+        pytest.fail(
+            f"Unable to instantiate IonicFraction for ion = {ion}, "
+            f"ionic_fraction = {ionic_fraction}, and number_density = "
+            f"{number_density}"
+        )
```
StanczakDominik, line 37: This could just be the `IonicFraction` call - pytest will definitely return test pytest on a failure. There's no need to write these exceptions.

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,100 @@
...
+    """
+    Test that IonicFraction raises exceptions when the ionic fraction
+    is out of the interval [0,1] or otherwise invalid.
+    """
+    with pytest.raises(expected_exception):
+        IonicFraction(ion="Fe 6+", ionic_fraction=invalid_fraction)
+        pytest.fail(
+            f"IonicFraction did not raise the appropriate exception "
+            f"for ionic_fraction = {repr(invalid_fraction)}, which is "
+            f"invalid."
```
StanczakDominik, line 65: Likewise here:

```
============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/dominik/pytest_test
plugins: mock-2.0.0, forked-1.1.3, xdist-1.32.0, pudb-0.7.0, cov-2.8.1
collected 1 item                                                                                                                                                                                                 

test.py F                                                                                                                                                                                                  [100%]

==================================================================================================== FAILURES ====================================================================================================
__________________________________________________________________________________________________ test_raises ___________________________________________________________________________________________________

    def test_raises():
        with pytest.raises(ValueError):
>           print("RECOGNITION CODE 432 I AM A POTATO")
E           Failed: DID NOT RAISE <class 'ValueError'>

test.py:5: Failed
---------------------------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------------------------
RECOGNITION CODE 432 I AM A POTATO
============================================================================================ short test summary info =============================================================================================
FAILED test.py::test_raises - Failed: DID NOT RAISE <class 'ValueError'>
=============================================================================================== 1 failed in 0.09s ================================================================================================
```

so maybe don't duplicate pytest code, please? :smiley:

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,100 @@
...
+
+@pytest.mark.parametrize("invalid_particle", ["H", "e-", "Fe-56"])
+def test_ionic_fraction_invalid_particles(invalid_particle):
+    """
+    Test that `~plasmapy.particles.IonicFraction` raises the appropriate
+    exception when passed a particle that isn't a neutral or ion.
+    """
+    with pytest.raises(ParticleError):
+        IonicFraction(invalid_particle, ionic_fraction=0)
+        pytest.fail()
```
StanczakDominik, line 77: um, what's this `pytest.fail` doing here?

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,100 @@
...
+    ion = "He 1+"
+    unnecessary_ionic_fractions = [0.0, 0.0, 1.0]
+
+    with pytest.raises(ParticleError):
+        IonizationState(ion, ionic_fractions=unnecessary_ionic_fractions)
+        pytest.fail(
+            f"When IonizationState is supplied an ion input ({repr(ion)}) "
+            f"along with ionic fractions {unnecessary_ionic_fractions}, "
+            f"the expected exception is not raised."
+        )
```
StanczakDominik, line 96: Cut this as well, maybe...?

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -52,6 +133,9 @@
 test_names = test_cases.keys()
 
 
+# TODO: Refactor these tests using fixtures
```
StanczakDominik, line 136: ~~There's not actually much of a reason to do this here, because the computational cost to initialize the `IonizationState` objects is low and there's no need to store instances between tests. You also have a separate set of parameters for each test, so you wouldn't reuse much~~ Oh, you mean the tests below, right? In that case, sure!

plasmapy/particles/ionization_state.py
```diff
@@ -2,30 +2,169 @@
...
+            "can_be_negative": False,
+            "can_be_inf": False,
+            "none_shall_pass": True,
+        },
+    )
+    def number_density(self, n: u.m ** -3):
+        if n is None:
+            self._number_density = np.nan * u.m ** -3
+        else:
+            self._number_density = n.to(u.m ** -3)
```
StanczakDominik, line 167: I think I'm going to keep a bookmark on this to remember how to use `validate` on properties :smile:

docs/particles/ionization_states.rst
namurphy: Agreed...it would be extremely unlikely for reals.  For test cases, though, we might set something like 50% H⁰⁺ and 50% p⁺.  My motivation for returning a `list` is to avoid the situation where people have to write special code for cases where there's a unique most abundant ion and where there are multiple.  An alternative would be to issue a warning or raise an exception if there's more than one most abundant ion.  I'm not sure which way would work best.

docs/particles/ionization_states.rst
namurphy: > `summarize()`?

I was thinking the same thing!  I like both `summarize()` and `describe()`, with a slight preference for `summarize()`.

plasmapy/particles/ionization_state.py
namurphy: Turns out, trying to figure this out was the inspiration for #797!

plasmapy/particles/tests/test_ionization_state.py
namurphy: Yeah...the corresponding error message should go in `IonicFraction.__init__`.  That would clean up these and other tests.  I think I put the extra error messages in because the OS on my home computer had just given me a "Oops... Something has gone wrong. Please contact your system administrator." error while booting.

plasmapy/particles/tests/test_ionization_state.py
namurphy: - [ ] Create issue on using fixtures for these tests.  Be sure to include directional information.

plasmapy/particles/tests/test_ionization_state.py
namurphy: Removed.

docs/particles/ionization_states.rst
namurphy: Given the chance of a tie, I'm going to leave this as a `list`, but am open to suggestions for the future.

namurphy:
Okay, I think I addressed all of the comments.  The test coverage for the diff is low, so I still need to look for tests that may be missing.

namurphy:
I'm content with the coverage for tests.  Most if not all of the remaining lines are associated with exceptions.  Plus there are a few lines for equilibrium ionization fractions that are not implemented here, and I'd rather those lines not show up as having been tested.  Ready for another round of review!

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,153 @@
...
+>>> ionization_state.summarize()
+an abundance of at least 1%.
+
+>>> ionization_state.summarize()
+an abundance of at least 1%.
+
+>>> ionization_state.summarize()
+an abundance of at least 1%.
+
+>>> ionization_state.info()
```
StanczakDominik, line 63: Whoa there.
```suggestion
The ``summarize`` method prints out the ionic fraction for the ions with
an abundance of at least 1%.

 >>> ionization_state.summarize()
```

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,153 @@
...
+>>> ionization_state.number_densities
+<Quantity [5.0e+17, 9.4e+18, 1.0e+17] 1 / m3>
+
+Ionization states for multiple elements
+=======================================
+
+The `~plasmapy.particles.IonizationStateCollection` class may be used to
+represent the ionization state for multiple elements.
+
+>>> from plasmapy.atomic import IonizationStateCollection
```
StanczakDominik, line 97: Not anymore!
```suggestion
>>> from plasmapy.particles import IonizationStateCollection 
```

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,153 @@
...
+<Quantity 5.e+19 1 / m3>
+>>> states.abundances
+{'H': 1.0, 'He': 0.08}
+>>> states.number_densities["H"]
+<Quantity [5.00e+17, 4.95e+19] 1 / m3>
+
+The ``info`` method may also be used to get a summary of the ionization
+states.
+
+>>> states.info()
```
StanczakDominik, line 144:
```suggestion
The ``summarize`` method may also be used to get a summary of the ionization
states.

>>> states.summarize()
```

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,98 @@
...
+    # Prepare to check for the default values when they are not set
+
+    if ionic_fraction is None:
+        ionic_fraction = np.nan
+    if number_density is None:
+        number_density = np.nan * u.m ** -3
+
+    assert Particle(ion) == Particle(instance.ionic_symbol)
+    assert u.isclose(instance.ionic_fraction, ionic_fraction, equal_nan=True)
+    assert u.isclose(instance.number_density, number_density, equal_nan=True)
```
StanczakDominik, line 41: Yay for `units.isclose`! :smiley:

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -1,19 +1,98 @@
...
+    with pytest.raises(ParticleError):
+        IonicFraction(invalid_particle, ionic_fraction=0)
+
+
+
+@pytest.mark.parametrize("ion1, ion2", [("Fe-56 6+", "Fe-56 5+"), ("H 1+", "D 1+")])
+def test_ionic_fraction_comparison_with_different_ions(ion1, ion2):
+    """
+    Test that a `TypeError` is raised when an `IonicFraction` object
+    is compared to an `IonicFraction` object of a different ion.
```
StanczakDominik, line 72: I kinda wonder whether that should be a basic Python `TypeError`, or some custom `ParticleError` instance... `TypeError` is pretty broad. Feels a bit like `except Exception`.

plasmapy/particles/ionization_state.py
```diff
@@ -2,30 +2,167 @@
...
+            ) from exc
+
+    @particle_input
+    def __init__(self, ion: Particle, ionic_fraction=None, number_density=None):
+        try:
+            self._particle = ion
+            self.ionic_fraction = ionic_fraction
+            self.number_density = number_density
+        except Exception as exc:
+            raise ParticleError("Unable to create IonicFraction object") from exc
```
StanczakDominik, line 99: Fun fact! I've been tinkering with dataclasses when scripting recently, and except for the `particle_input` decorator, I think this could have been one - if not for the Py3.7 requirement. Then you wouldn't really have to write a bunch of these utility methods. But of course, that's not a change suggestion! I really wish GitHub would let me differentiate review comments between "change this" and "oh I'm just fawning over f-strings, don't mind me."

plasmapy/particles/ionization_state.py
```diff
@@ -374,15 +531,18 @@ def _is_normalized(self, tol: Optional[Real] = None) -> bool:
...
-        sum becomes equal to one.
+        sum of the ionic fractions becomes equal to one.
+
+        This method may be used, for example, to correct for rounding
+        errors.
         """
         self._ionic_fractions = self._ionic_fractions / np.sum(self._ionic_fractions)
 
     @property
     def equil_ionic_fractions(self, T_e: u.K = None):
```
rocco8773, line 542: ?? properties do not have arguments

rocco8773, line 542: I'd argue `equil_ionic_fractions` and `equilibrate` should be removed altogether since they do nothing and would be misleading to users.  I can see the use in the future, but, since the functionality does not exist, there shouldn't be placeholders and an issue should be made to create the functionality.

plasmapy/particles/ionization_state.py
```diff
@@ -393,8 +553,7 @@ def equilibrate(self, T_e: u.K = np.nan * u.K):
         Set the ionic fractions to collisional ionization equilibrium
         for temperature ``T_e``.  Not implemented.
         """
-        # self.ionic_fractions = self.equil_ionic_fractions
-        raise NotImplementedError
+        self.ionic_fractions = self.equil_ionic_fractions
```
rocco8773, line 556: See above comment.

rocco8773 (changes requested):
This is just a partial review of the included `.py` files.

plasmapy/particles/ionization_state_collection.py
```diff
@@ -79,7 +83,7 @@ class IonizationStates:
     {'H': array([0.5, 0.5]), 'He': array([0.95, 0.05, 0.  ])}
 
     The number densities are given by the ionic fractions multiplied by
-    the abundance and the
+    the abundance and the number density scaling factor ``n``.
```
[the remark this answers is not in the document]
namurphy: This parameter was originally the number density of hydrogen, but this would have required that hydrogen was one of the elements included in the collection.  The number density of an ionization state is: (abundance of element) × (ionic fraction) × (number density scaling factor).  It's not the overall density but rather a multiplicative factor with units of number density.  I'll try to clarify that in the docstring.  I'm wondering if it'd be better to rename ``n`` as ``n0`` or something.

plasmapy/particles/ionization_state.py
```diff
@@ -113,7 +261,20 @@ def __init__(
...
-        self._particle_instance = particle
+        if particle.is_ion or particle.is_category(require=("uncharged", "element")):
+            if ionic_fractions is None:
+                ionic_fractions = np.zeros(particle.atomic_number + 1)
+                ionic_fractions[particle.integer_charge] = 1.0
+                particle = Particle(
+                    particle.isotope if particle.isotope else particle.element
+                )
+            else:
+                raise ParticleError(
```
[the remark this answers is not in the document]
namurphy: Good idea!

plasmapy/particles/ionization_state.py
```diff
@@ -393,8 +553,7 @@ def equilibrate(self, T_e: u.K = np.nan * u.K):
         Set the ionic fractions to collisional ionization equilibrium
         for temperature ``T_e``.  Not implemented.
         """
-        # self.ionic_fractions = self.equil_ionic_fractions
-        raise NotImplementedError
+        self.ionic_fractions = self.equil_ionic_fractions
 
     @property
     @u.quantity_input
```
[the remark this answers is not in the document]
namurphy: :scream_cat:

plasmapy/particles/ionization_state.py
namurphy: Good catch!

namurphy: Yeah...I think you're right.

plasmapy/particles/ionization_state_collection.py
```diff
@@ -457,7 +458,7 @@ def ionic_fractions(self, inputs: Union[Dict, List, Tuple]):
 
         Raises
         ------
-        AtomicError
+        ParticleError
```
rocco8773, line 461:
```suggestion
        `ParticleError`
```

Missed this one.

plasmapy/particles/ionization_state.py
```diff
@@ -2,32 +2,168 @@
...
 
-#: Named tuple class for representing an ionization state (`collections.namedtuple`).
-State = collections.namedtuple(
-    "State", ["integer_charge", "ionic_fraction", "ionic_symbol", "number_density"]
-)
+class IonicFraction:
+    """
+    Representation of the ionic fraction for a single ion.
+
+    This class is created when indexing
```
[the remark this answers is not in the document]
namurphy: This is what is returned when using `__getitem__` for an `IonizationState` instance.  I removed the line since that sort of explanation goes better in `IonizationState`'s docstring.

plasmapy/particles/ionization_state.py
namurphy: Removed that line anyway.

plasmapy/particles/ionization_state.py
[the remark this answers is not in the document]
namurphy: I created these exceptions, but applying them would increase the scope of this PR more than I'd like since it would involve digging into `Particle` and its methods in more detail.  

- [ ] When this is merged, create an issue on implementing `ParticleTypeError` and `ParticleValueError` in `plasmapy.particles`

plasmapy/particles/ionization_state_collection.py
```diff
@@ -630,7 +633,9 @@ def ionic_fractions(self, inputs: Union[Dict, List, Tuple]):
             try:
                 _particle_instances = [Particle(particle) for particle in inputs]
             except (InvalidParticleError, TypeError) as exc:
-                raise AtomicError("Invalid inputs to IonizationStates.") from exc
+                raise ParticleError(
```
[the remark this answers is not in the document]
namurphy: I'd rather hold off on adding these tests until the test helper refactoring in #728 is implemented, since one of the purposes of that refactoring is to simplify writing tests like these.  

 - [ ] After merging, create issue on improving test coverage in `ionization_state.py` and `ionization_state_collection.py`, and state that this should be done after the test helper refactoring is complete.

plasmapy/particles/particle_class.py
```diff
@@ -2061,3 +2064,9 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
                 "a real number that represents the ratio of the charge to "
                 "the elementary charge."
             )
+
+
+# TODO: Describe valid particle representations in docstring of particle_like
+
+particle_like = Union[str, Integral, Particle]
+"""A typing construct for valid representations of a particle."""
```
namurphy, line 2072: @rocco8773 — With respect to [your comment](https://github.com/PlasmaPy/PlasmaPy/issues/912#issuecomment-706764253) in #912, I tentatively added a stub for `particle_like` here.  I started using it in this PR, so it'd be easiest for me if we could finish this PR, and then work on a separate PR that implements your ideas and puts `particle_like` in a more complete form, possibly in a different location.

Also:
 - `numbers.Integral` corresponds to the base class for integers representing atomic numbers.  
 - I used all lower case in `particle_like` since types like `int` and `float` are typically lower case, and because I think it's an instance rather than an actual class.  (With that said, I don't really have a good idea of the inner workings of `typing`.)

rocco8773: That's fine.  We can discuss `particle_like` vs `ParticleLike` in #912 and the subsequent PR.  I did choose `ParticleLike` because that's the nomenclature used in [Python's documentation for typing aliases](https://docs.python.org/3/library/typing.html#type-aliases).  (Even though I personally prefer `particle_like`).

With regards to `number.Integral`, it is not a traditional base class for for `int` (i.e. `int` does not inherit from it...`int.__mro__ == (int, object)`).  Instead it is an [abc "registered" subclass](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register), or a "virtual subclass".

changelog/796.breaking.rst
```diff
@@ -0,0 +1,7 @@
+In ``plasmapy.particles``, rename the ``IonizationStates`` class to
+``IonizationStateCollection`` and replace the ``State`` namedtuple with
+the ``IonicFraction`` class.  When the ``IonizationState`` class is now
+provided with an ion, the ionic fraction for that ion is set to 100% for
+the corresponding element or isotope.  Renamed ``AtomicError`` to
+``ParticleError`` and ``MissingAtomicDataError`` to
+``MissingParticleDataError``.
```
rocco8773, line 7: Use single back ticks `` ` `` so sphinx will auto-link the objects.

plasmapy/particles/decorators.py
```diff
@@ -98,17 +98,17 @@ def particle_input(
 
     require : `str`, `set`, `list`, or `tuple`, optional
         Categories that a particle must be in.  If a particle is not in
-        all of these categories, then an `~plasmapy.utils.AtomicError`
+        all of these categories, then an `~plasmapy.utils.ParticleError`
```
rocco8773, line 101:
```suggestion
        all of these categories, then an `~plasmapy.particles.exceptions.ParticleError`
```

plasmapy/particles/decorators.py
```diff
@@ -98,17 +98,17 @@ def particle_input(
...
         will be raised.
 
     any_of : `str`, `set`, `list`, or `tuple`, optional
         Categories that a particle may be in.  If a particle is not in
-        any of these categories, then an `~plasmapy.utils.AtomicError`
+        any of these categories, then an `~plasmapy.utils.ParticleError`
```
rocco8773, line 106:
```suggestion
        any of these categories, then an `~plasmapy.particles.exceptions.ParticleError`
```

plasmapy/particles/decorators.py
```diff
@@ -98,17 +98,17 @@ def particle_input(
...
         will be raised.
 
     exclude : `str`, `set`, `list`, or `tuple`, optional
         Categories that a particle cannot be in.  If a particle is in
-        any of these categories, then an `~plasmapy.utils.AtomicError`
+        any of these categories, then an `~plasmapy.utils.ParticleError`
```
rocco8773, line 111:
```suggestion
        any of these categories, then an `~plasmapy.particles.exceptions.ParticleError`
```

plasmapy/particles/decorators.py
```diff
@@ -165,7 +165,7 @@ def particle_input(
         'uncharged'}`` and the particle does not have charge information
         associated with it.
 
-    `~plasmapy/utils/AtomicError`
+    `~plasmapy/utils/ParticleError`
```
rocco8773, line 168:
```suggestion
    `~plasmapy.particles.excpetions.ParticleError`
```

Actually this whole "Raises" section uses slashes `/` instead of dots `.` and thus links are not properly created.  Update all these instances.

plasmapy/particles/atomic.py
```diff
@@ -229,7 +230,7 @@ def particle_mass(
     `~plasmapy.utils.InvalidParticleError`
         If the argument does not correspond to a valid particle.
 
-    `~plasmapy.utils.MissingAtomicDataError`
+    `~plasmapy.utils.MissingParticleDataError`
```
rocco8773, line 233:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`
```

plasmapy/particles/atomic.py
```diff
@@ -445,7 +446,7 @@ def is_stable(particle: Particle, mass_numb: Optional[Integral] = None) -> bool:
     `TypeError`
         If the argument is not a `str` or `int`.
 
-    `~plasmapy.utils.MissingAtomicDataError`
+    `~plasmapy.utils.MissingParticleDataError`
```
rocco8773, line 449:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`
```

plasmapy/particles/atomic.py
```diff
@@ -494,7 +495,7 @@ def half_life(particle: Particle, mass_numb: Optional[Integral] = None) -> u.Qua
         If the argument does not correspond to a valid particle
         or contradictory information is provided.
 
-    `~plasmapy.utils.MissingAtomicDataError`
+    `~plasmapy.utils.MissingParticleDataError`
```
rocco8773, line 498:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`
```

plasmapy/particles/atomic.py
```diff
@@ -506,9 +507,9 @@ def half_life(particle: Particle, mass_numb: Optional[Integral] = None) -> u.Qua
     Accurate half-life data is not known for all isotopes. Some isotopes
     may have upper or lower limits on the half-life, in which case this
     function will return a string with that information and issue a
-    `~plasmapy.utils.MissingAtomicDataWarning`.  When no isotope
+    `~plasmapy.utils.MissingParticleDataWarning`.  When no isotope
```
rocco8773, line 510:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`.  When no isotope
```

plasmapy/particles/atomic.py
```diff
@@ -506,9 +507,9 @@ def half_life(particle: Particle, mass_numb: Optional[Integral] = None) -> u.Qua
...
     information is available, then this function raises a
-    `~plasmapy.utils.MissingAtomicDataError`.
+    `~plasmapy.utils.MissingParticleDataError`.
```
rocco8773, line 512:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`.
```

plasmapy/particles/atomic.py
```diff
@@ -870,7 +871,7 @@ def reduced_mass(test_particle, target_particle) -> u.Quantity:
         If an argument is a `~astropy.units.Quantity` or
         `~astropy.units.Constant` but does not have units of mass.
 
-    `~plasmapy.utils.MissingAtomicDataError`
+    `~plasmapy.utils.MissingParticleDataError`
```
rocco8773, line 874:
```suggestion
    `~plasmapy.particles.exceptions.MissingParticleDataError`
```

rocco8773 (changes requested):
I started a review, but I'm going to stop here.  Almost all of the docstring and rst references to the new exceptions are improperly pointing to `plasmapy.utils` when they should be pointing to `plasmapy.particles.exceptions`.  Go through an do these and then I'll start another review.

changelog/796.breaking.rst
namurphy: I've been wondering about how we should handle links in the change log.  For example, past change logs may have references to `plasmapy.atomic`, though that subpackage doesn't exist anymore so the auto-links don't work.  I was putting them in double back ticks so that it would be future proof.  With that said...maybe it would still be better to do the auto-links regardless...though we'd need to remember to update the obsolete links in old changelogs.  I'll go with your suggestion.

namurphy: Update: I'm still using double back ticks for names for objects that no longer exist.

rocco8773:
In `atomic.py` there are still references to exceptions in `plasmapy.utils` that aren't there anymore.  I can not comment directly on them since they were not initially touched by this PR.  Not sure when things changed to break these links.  Do you want to fix this in this PR or do a separate one?

Additionally, the list of parameters in the `isotopic_abundance` docstring does not match the functions signature.

namurphy:
> Almost all of the docstring and rst references to the new exceptions are improperly pointing to `plasmapy.utils` when they should be pointing to `plasmapy.particles.exceptions`.

Yikes!  That's what I get for updating names with a "replace all".  I think I fixed all of those things in `plasmapy.particles`.  I'm making similar fixes in #942 for the `particles` docs.  I think it's ready for another review whenever you're ready.

changelog/796.breaking.rst
rocco8773: Note: The single ticks `` ` `` are "future proof".  They will render just fine.  They just won't link to anything.

I don't think it's worth our effort to go through and manually change any of this for our old changelogs.  Being consistent about it will be a real pain.  People fully expect links to no longer exist for ancient changes.  If users really want the original references, they can go to the version specific documentation where the links are still valid.

namurphy:
> In `atomic.py` there are still references to exceptions in `plasmapy.utils` that aren't there anymore. I can not comment directly on them since they were not initially touched by this PR. Not sure when things changed to break these links. Do you want to fix this in this PR or do a separate one?
> 
> Additionally, the list of parameters in the `isotopic_abundance` docstring does not match the functions signature.

I made the fixes to `atomic.py` as you were typing this comment, and I just updated the `isotopic_abundance` docstring too.

Also, I need to stop watching multiple episodes of Star Trek during the day so that I don't have to work at night to get my eight hours in.  Oof!

rocco8773:
```
@particle_input
def isotopic_abundance(isotope: Particle, mass_numb: Optional[Integral] = None) -> Real:
    ...
```

This is one of my sticking points with how `@particle_input` works.  `isotopic_abundance` does not use argument `mass_numb` anywhere in its definition; thus, it appears as an unused parameter.  If I didn't know that `@particle_input` wrangles this specific parameter, I would just remove it all together.  There is just too much implicitness here, versus explicitness.

plasmapy/particles/atomic.py
```diff
@@ -663,14 +655,14 @@ def common_isotopes(
 
     See Also
     --------
-    ~plasmapy.utils.known_isotopes : returns a list of isotopes that
-        have been discovered.
+    `~plasmapy.particles.known_isotopes` : returns a list of isotopes
```
rocco8773, line 658:
```suggestion
    ~plasmapy.particles.known_isotopes : returns a list of isotopes
```

The "See Also" section is one of the weird situations where you do NOT use the back ticks `` ` ``.  If you do, `Sphinx` does not know how to interpret it and it shows up as `None()`.

plasmapy/particles/atomic.py
```diff
@@ -663,14 +655,14 @@ def common_isotopes(
...
+        that have been discovered.
 
-    ~plasmapy.utils.stable_isotopes : returns isotopes that are stable
-        against radioactive decay.
+    `~plasmapy.particles.stable_isotopes` : returns isotopes that are
```
rocco8773, line 661:
```suggestion
    ~plasmapy.particles.stable_isotopes : returns isotopes that are
```

plasmapy/particles/atomic.py
```diff
@@ -663,14 +655,14 @@ def common_isotopes(
...
+        stable against radioactive decay.
 
-    ~plasmapy.utils.isotopic_abundance : returns the relative isotopic
-         abundance.
+    `~plasmapy.particles.isotopic_abundance` : returns the relative
```
rocco8773, line 664:
```suggestion
    ~plasmapy.particles.isotopic_abundance : returns the relative
```

plasmapy/particles/exceptions.py
```diff
@@ -1,41 +1,61 @@
...
 
 
-class MissingAtomicDataError(AtomicError):
+class ParticleTypeError(ParticleError, TypeError):
+    """
+    An exception for errors in the `~plasmapy.particles` subpackage
+    caused because an incorrect type was provided.
+    """
+
+    pass
```
rocco8773, line 34: Is this going to be for when a wrong particle type is given or when a wrong argument type is given for some functionality in `plasmapy.particles`?  To me it should be used for the former and the later still raises `TypeError`.

plasmapy/particles/exceptions.py
```diff
@@ -1,41 +1,61 @@
...
+
+
+class ParticleValueError(ParticleError, ValueError):
+    """
+    An exception for errors in the `~plasmapy.particles` subpackage
+    caused because an incorrect value was provided.
+    """
+
+    pass
```
rocco8773, line 43: Similar question to `ParticleTypeError`.  Is this going to be for when a wrong particle value is given (e.g. a lepton when expecting an ion, an atom when expecting an ion, etc.) or when a wrong value is given for some functionality in `plasmapy.particles`.  Again, to me it should be used for the former case and the later case still raises `ValueError`.

plasmapy/particles/exceptions.py
```diff
@@ -1,41 +1,61 @@
...
+
+
+class MissingParticleDataError(ParticleError):
     """An exception for missing atomic or particle data."""
 
     pass
```
rocco8773, line 49: Could this also inherit from `AttributeError`?  I'm guessing this for the case when you're trying to access `particle.charge` and the attribute does not exist.  Or, is it for when the particle data is missing from the `json` files?

plasmapy/particles/exceptions.py
```diff
@@ -1,41 +1,61 @@
...
 
 
-class ChargeError(AtomicError):
+class ChargeError(ParticleError):
```
rocco8773, line 52:
```suggestion
class ParticleChargeError(ParticleError):
```

To make error reporting more explicit.  Without context, it might not be immediately apparent the `ChargeError` is really referring to a `ParticleChargeError`.

plasmapy/particles/exceptions.py
namurphy: It's mostly the latter case.  In some cases where this is used, the attribute in question still exists.

plasmapy/particles/exceptions.py
namurphy: Hm...I think I'm going to drop `ParticleTypeError` and `ParticleValueError` from this pull request because I'm not actually implementing them anywhere and it's a bit outside the scope of this PR.  For the cases where the wrong particle category is given, there's already `InvalidIonError`, etc.  We could still create these later on if we end up needing something that isn't covered by the existing exceptions.

plasmapy/particles/atomic.py
namurphy: I just discovered that it's hard to find documentation for "See Also" in the Sphinx docs because all the "See Also" sections show up in the search results...

plasmapy/particles/exceptions.py
namurphy: `ChargeError` is a bit broader than just for particles, since it could refer to missing or invalid charge information elsewhere.  I'm wondering if it'd be better to put this in `plasmapy.utils.exceptions`...but that's not for this pull request.

plasmapy/particles/atomic.py
namurphy: Found [something about see also in the LSST docs](https://developer.lsst.io/python/numpydoc.html#py-docstring-see-also), and there was much rejoicing.

plasmapy/particles/exceptions.py
rocco8773: Ok, I think the docstring should be updated to reflect this error is raised when the requested data as not found in the "databases."

plasmapy/particles/exceptions.py
rocco8773: I'm in two minds about this.  I'm afraid we're heading down the path of having things like `MassError`, `VelocityError`, `DensityError`, etc.  Something like that would end up being a pain to maintain, especially since most cases would be better suited using `ValueError` or `TypeError`.  Now, I don't mind it being used in the context of `plasmapy.particles` to indicate this is an error is coming from the particle framework.  In that case, it should only be used within the particle framework.  If it would be used in a more general sense, then I take issue according to my initial point but it should also not inherit for `ParticleError`.  As stated in the `ParticleError` docstring, `ParticleError` is for errors specific to the particle framework.

namurphy:
> This is one of my sticking points with how `@particle_input` works. `isotopic_abundance` does not use argument `mass_numb` anywhere in its definition; thus, it appears as an unused parameter. If I didn't know that `@particle_input` wrangles this specific parameter, I would just remove it all together. There is just too much implicitness here, versus explicitness.

Yeah...I'm getting reticent about having `@particle_input` interpret `mass_numb` and `Z`...primarily because `mass_numb` and `Z` don't show up in the lists of parameters, etc.  I don't want to deal with that in this pull request though, since I've let in a lot of scope creep already, as I have a tendency to do...

rocco8773:
> 
> 
> > This is one of my sticking points with how `@particle_input` works. `isotopic_abundance` does not use argument `mass_numb` anywhere in its definition; thus, it appears as an unused parameter. If I didn't know that `@particle_input` wrangles this specific parameter, I would just remove it all together. There is just too much implicitness here, versus explicitness.
> 
> Yeah...I'm getting reticent about having `@particle_input` interpret `mass_numb` and `Z`...primarily because `mass_numb` and `Z` don't show up in the lists of parameters, etc. I don't want to deal with that in this pull request though, since I've let in a lot of scope creep already, as I have a tendency to do...

I agree with that!  I was thinking we'd have the discussion when I address issue #912 (and/or #918).

plasmapy/particles/exceptions.py
rocco8773: How about something like...

```suggestion
class MissingParticleDataError(ParticleError):
    """
    An exception for missing atomic or particle data in `plasmapy`'s 
    particle databases.
    """

    pass
```

namurphy: I made a change that was similar to that.

plasmapy/particles/exceptions.py
namurphy: `ChargeError` is a bit different than `MassError`, etc.  A `ChargeError` can come up when a function requires a charged particle, but the particle is not charged or lacks charge information, or when an invalid charge is supplied (i.e. H²⁺).  I've found it useful in the `particles` subpackage.  I'd rather leave the status quo for `ChargeError` at least for the time being, in part to avoid scope creep for this PR.

namurphy:
This is the issue I mentioned during the community meeting today, which would be really helpful to have merged by our 0.5.0 release.  There are some breaking changes here, and having this merged will make it easier for me when I get back to working on the non-equilibrium ionization modeling affiliated package...hopefully in the second half of December.  Thank you in advance!

plasmapy/particles/ionization_state_collection.py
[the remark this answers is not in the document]
StanczakDominik: Is the scaling factor more easily available at input time than the number density? It might be better to calculate it ourselves from number density, given that we already grab the abundances.

docs/particles/ionization_states.rst
```diff
@@ -0,0 +1,141 @@
...
+ion or neutral.
+
+>>> ionization_state.number_densities
+<Quantity [5.0e+17, 9.4e+18, 1.0e+17] 1 / m3>
+
+Ionization states for multiple elements
+=======================================
+
+The `~plasmapy.particles.IonizationStateCollection` class may be used to
+represent the ionization state for multiple elements.
```
StanczakDominik, line 83:
```suggestion
The `~plasmapy.particles.IonizationStateCollection` class may be used to
represent the ionization state for multiple elements. This can be used,
for example, to describe the various impurities in a fusion plasma.
```

Just something I just realized! :)

plasmapy/particles/ionization_state.py
```diff
@@ -2,32 +2,168 @@
...
+            ionic_fraction_within_tolerance = np.isclose(
+                self.ionic_fraction, other.ionic_fraction, rtol=1e-15,
+            )
+
+            number_density_within_tolerance = u.isclose(
+                self.number_density, other.number_density, rtol=1e-15,
+            )
+
+            return all(
+                [ionic_fraction_within_tolerance, number_density_within_tolerance,]
```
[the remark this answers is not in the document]
StanczakDominik: `black` doesn't complain and I'm not going to, either!

plasmapy/particles/ionization_state.py
```diff
@@ -2,32 +2,168 @@
...
+            )
+
+        except Exception as exc:
+            raise TypeError(
+                "Unable to ascertain equality between the following objects:\n"
+                f"  {self}\n"
+                f"  {other}"
+            ) from exc
```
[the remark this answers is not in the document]
StanczakDominik: Agreed!

plasmapy/particles/ionization_state.py
```diff
@@ -2,32 +2,168 @@
...
+
+    @property
+    def number_density(self) -> u.m ** -3:
+        """The number density of the ion."""
+        return self._number_density.to(u.m ** -3)
+
+    @number_density.setter
+    @validate_quantities(
+        n={
+            "units": u.m ** -3,
```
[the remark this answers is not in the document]
StanczakDominik: Agreed!

plasmapy/particles/ionization_state.py
```diff
@@ -2,32 +2,168 @@
...
+            "can_be_negative": False,
+            "can_be_inf": False,
+            "none_shall_pass": True,
+        },
+    )
+    def number_density(self, n: u.m ** -3):
+        if n is None:
+            self._number_density = np.nan * u.m ** -3
+        else:
+            self._number_density = n.to(u.m ** -3)
```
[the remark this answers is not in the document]
StanczakDominik: Yep. I'll just commit these once I'm done reviewing (don't want to keep refreshing the page...)

StanczakDominik (changes requested):
LGTM, except for a few comments added below and https://github.com/PlasmaPy/PlasmaPy/pull/796/files?file-filters%5B%5D=.py#r541546319 (number density scaling factor argument name "n" is confusing)!

plasmapy/particles/ionization_state_collection.py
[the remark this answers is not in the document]
namurphy: Thanks!  I'll change `n` to `n0`.

namurphy:
Ok!  I think I finished implementing the changes from code review, so if it looks good to you, feel free to merge.  Thanks!

StanczakDominik (approved):

PR #816, opened 2020-05-26
Enforcing and validating units in when creating distributions

FilipBolt:
  they're failing, ask!).

Step 1 to take care of https://github.com/PlasmaPy/PlasmaPy/issues/387 
Step 2 will refactor distributions, which should produce more reusable code, which should finish up https://github.com/PlasmaPy/PlasmaPy/issues/387 

I have made 3D norm tests slow tests as wrapping them up slowed them down considerably (it took 15 minutes for me to execute Maxwellian 3D after the wrapping).

### DIFF
--- a/plasmapy/formulary/distribution.py
+++ b/plasmapy/formulary/distribution.py
@@ -21,19 +21,16 @@ __all__ = [
     "kappa_velocity_3D",
 ]
 
+_velocity_unit = u.m / u.s
 
-@validate_quantities(
-    v={'units': u.m / u.s, 'pass_equivalent_units': True},
-    T={
-        'units': u.K, 'equivalencies': u.temperature_energy(),
-        'pass_equivalent_units': True
-      },
-    v_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vTh={'units': u.m / u.s, 'pass_equivalent_units': True},
-    validations_on_return=u.s / u.m
-)
+
+@validate_quantities(T={'equivalencies': u.temperature_energy()})
 def Maxwellian_1D(
-    v: u.m / u.s, T, particle="e", v_drift=0, vTh=np.nan
+    v: _velocity_unit,
+    T: u.K,
+    particle="e",
+    v_drift: _velocity_unit = 0 * _velocity_unit,
+    vTh: _velocity_unit = np.nan * _velocity_unit,
 ) -> u.s / u.m:
     r"""
     Probability distribution function of velocity for a Maxwellian
@@ -49,7 +46,7 @@ def Maxwellian_1D(
         The velocity in units convertible to m/s.
 
     T: ~astropy.units.Quantity
-        The temperature in Kelvin.
+        The temperature in units covertible to Kelvin.
 
     particle: str, optional
         Representation of the particle species(e.g., ``'p'`` for protons,
@@ -99,15 +96,12 @@ def Maxwellian_1D(
     --------
     >>> from astropy import units as u
     >>> v=1*u.m/u.s
-    >>> Maxwellian_1D(v=v, T=30000 * u.K, particle='e', v_drift=0 * u.m / u.s)
+    >>> Maxwellian_1D(v=v, T=30000 * u.K, particle='e', v_drift=0 * _velocity_unit)
     <Quantity 5.9163...e-07 s / m>
     """
     if np.isnan(vTh):
         # get thermal velocity and thermal velocity squared
         vTh = parameters.thermal_speed(T, particle=particle, method="most_probable")
-    elif not np.isnan(vTh):
-        # check units of thermal velocity
-        vTh = vTh.to(u.m / u.s)
 
     # Get thermal velocity squared
     vThSq = vTh ** 2
@@ -121,26 +115,15 @@ def Maxwellian_1D(
     return distFunc
 
 
-@validate_quantities(
-    vx={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vy={'units': u.m / u.s, 'pass_equivalent_units': True},
-    T={
-        'units': u.K, 'equivalencies': u.temperature_energy(),
-        'pass_equivalent_units': True
-      },
-    vx_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vy_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vTh={'units': u.m / u.s, 'pass_equivalent_units': True},
-    validations_on_return=(u.s / u.m) ** 2
-)
+@validate_quantities(T={'equivalencies': u.temperature_energy()})
 def Maxwellian_velocity_2D(
-    vx: u.m / u.s,
-    vy: u.m / u.s,
+    vx: _velocity_unit,
+    vy: _velocity_unit,
     T: u.K,
     particle="e",
-    vx_drift=0,
-    vy_drift=0,
-    vTh=np.nan
+    vx_drift: _velocity_unit = 0. * _velocity_unit,
+    vy_drift: _velocity_unit = 0. * _velocity_unit,
+    vTh: _velocity_unit = np.nan * _velocity_unit,
 ) -> (u.s / u.m) ** 2:
     r"""
     Probability distribution function of velocity for a Maxwellian
@@ -217,13 +200,13 @@ def Maxwellian_velocity_2D(
     Example
     -------
     >>> from astropy import units as u
-    >>> v=1 * u.m / u.s
+    >>> v=1 * _velocity_unit
     >>> Maxwellian_velocity_2D(vx=v,
     ... vy=v,
     ... T=30000*u.K,
     ... particle='e',
-    ... vx_drift=0 * u.m / u.s,
-    ... vy_drift=0 * u.m / u.s)
+    ... vx_drift=0 * _velocity_unit,
+    ... vy_drift=0 * _velocity_unit)
     <Quantity 3.5002...e-13 s2 / m2>
 
 
@@ -231,9 +214,6 @@ def Maxwellian_velocity_2D(
     if np.isnan(vTh):
         # get thermal velocity and thermal velocity squared
         vTh = parameters.thermal_speed(T, particle=particle, method="most_probable")
-    elif not np.isnan(vTh):
-        # check units of thermal velocity
-        vTh = vTh.to(u.m / u.s)
 
     # accounting for thermal velocity in 2D
     vThSq = vTh ** 2
@@ -246,30 +226,17 @@ def Maxwellian_velocity_2D(
     return distFunc
 
 
-@validate_quantities(
-    vx={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vy={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vz={'units': u.m / u.s, 'pass_equivalent_units': True},
-    T={
-        'units': u.K, 'equivalencies': u.temperature_energy(),
-        'pass_equivalent_units': True
-      },
-    vx_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vy_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vz_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vTh={'units': u.m / u.s, 'pass_equivalent_units': True},
-    validations_on_return=(u.s / u.m) ** 3
-)
+@validate_quantities(T={'equivalencies': u.temperature_energy()})
 def Maxwellian_velocity_3D(
-    vx: u.m / u.s,
-    vy: u.m / u.s,
-    vz: u.m / u.s,
+    vx: _velocity_unit,
+    vy: _velocity_unit,
+    vz: _velocity_unit,
     T: u.K,
     particle="e",
-    vx_drift=0,
-    vy_drift=0,
-    vz_drift=0,
-    vTh=np.nan
+    vx_drift: _velocity_unit = 0. * _velocity_unit,
+    vy_drift: _velocity_unit = 0. * _velocity_unit,
+    vz_drift: _velocity_unit = 0. * _velocity_unit,
+    vTh: _velocity_unit = np.nan * _velocity_unit,
 ) -> (u.s / u.m) ** 3:
     r"""
     Probability distribution function of velocity for a Maxwellian
@@ -358,15 +325,15 @@ def Maxwellian_velocity_3D(
     Example
     -------
     >>> from astropy import units as u
-    >>> v=1 * u.m / u.s
+    >>> v=1 * _velocity_unit
     >>> Maxwellian_velocity_3D(vx=v,
     ... vy=v,
     ... vz=v,
     ... T=30000 * u.K,
     ... particle='e',
-    ... vx_drift=0 * u.m / u.s,
-    ... vy_drift=0 * u.m / u.s,
-    ... vz_drift=0 * u.m / u.s)
+    ... vx_drift=0 * _velocity_unit,
+    ... vy_drift=0 * _velocity_unit,
+    ... vz_drift=0 * _velocity_unit)
     <Quantity 2.0708...e-19 s3 / m3>
 
 
@@ -374,9 +341,6 @@ def Maxwellian_velocity_3D(
     if np.isnan(vTh):
         # get thermal velocity and thermal velocity squared
         vTh = parameters.thermal_speed(T, particle=particle, method="most_probable")
-    elif not np.isnan(vTh):
-        # check units of thermal velocity
-        vTh = vTh.to(u.m / u.s)
 
     # accounting for thermal velocity in 3D
     vThSq = vTh ** 2
@@ -389,19 +353,13 @@ def Maxwellian_velocity_3D(
     return distFunc
 
 
-@validate_quantities(
-    v={'units': u.m / u.s, 'pass_equivalent_units': True},
-    T={
-        'units': u.K, 'equivalencies': u.temperature_energy(),
-        'pass_equivalent_units': True
-      },
-    v_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
-    vTh={'units': u.m / u.s, 'pass_equivalent_units': True},
-    validations_on_return=u.s / u.m
-)
+@validate_quantities(T={'equivalencies': u.temperature_energy()})
 def Maxwellian_speed_1D(
-    v: u.m / u.s, T: u.K,
-    particle="e", v_drift=0, vTh=np.nan
+    v: _velocity_unit,
+    T: u.K,
+    particle="e",
+    v_drift: _velocity_unit = 0. * _velocity_unit,
+    vTh: _velocity_unit = np.nan * _velocity_unit,
 ) -> (u.s / u.m):
     r"""
     Probability distribution function of speed for a Maxwellian distribution
@@ -472,17 +430,14 @@ def Maxwellian_speed_1D(
     Example
     -------
     >>> from astropy import units as u
-    >>> v=1 * u.m / u.s
-    >>> Maxwellian_speed_1D(v=v, T=30000 * u.K, particle='e', v_drift=0 * u.m / u.s)
+    >>> v=1 * _velocity_unit
+    >>> Maxwellian_speed_1D(v=v, T=30000 * u.K, particle='e', v_drift=0 * _velocity_unit)
     <Quantity 1.1832...e-06 s / m>
 
     """
     if np.isnan(vTh):
         # get thermal velocity and thermal velocity squared
         vTh = parameters.thermal_speed(T, particle=particle, method="most_probable")
-    elif not np.isnan(vTh):
-        # check units of thermal velocity
-        vTh = vTh.to(u.m / u.s)
 
     # Get thermal velocity squared
     vThSq = vTh ** 2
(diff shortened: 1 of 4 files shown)


plasmapy/formulary/distribution.py
```diff
@@ -104,27 +100,22 @@ def Maxwellian_1D(v, T, particle="e", v_drift=0, vTh=np.nan, units="units"):
...
     """
+    # unit checks and conversions
+    # checking velocity units
+    # v = v * (u.m / u.s)
+    v = v.to(u.m / u.s)
+    # Catching case where drift velocities have default values, they
+    # need to be assigned units
+    v_drift = v_drift.to(u.m / u.s)
+    # convert temperature to Kelvins
+    # T = T.to(u.K, equivalencies=u.temperature_energy())
```
StanczakDominik, line 111: The `validate_quantities` decorator you're using above should mostly handle this block, I think :)

plasmapy/formulary/distribution.py
```diff
@@ -133,12 +124,22 @@ def Maxwellian_1D(v, T, particle="e", v_drift=0, vTh=np.nan, units="units"):
...
     expTerm = np.exp(-vSq / vThSq)
     distFunc = coeff * expTerm
-    if units == "units":
-        return distFunc.to(u.s / u.m)
-    elif units == "unitless":
-        return distFunc
-
-
+    # return output in SI unit
+    return distFunc.to(u.s / u.m)
```
StanczakDominik, line 128: And here as well!

StanczakDominik:
https://dev.azure.com/plasmapy/PlasmaPy/_build/results?buildId=912&view=logs&j=8955e93f-e95e-5ac2-4406-a311ed91aa3e&t=adda2fec-11c4-56e4-c47d-199954e67231&l=5495 seems to point to some exceptions' behavior being changed; could you look into that?

Also, `breaking` will be fine for the changelog here :) It's not much of an API break, but it is a bit of one.

FilipBolt:
Thanks again @StanczakDominik , I've added some tests to verify the conversions between different units work and ensured all tests pass now.

plasmapy/formulary/distribution.py
```diff
@@ -264,24 +243,34 @@ def Maxwellian_velocity_2D(
...
+    vy={'units': u.m / u.s, 'pass_equivalent_units': True},
+    vz={'units': u.m / u.s, 'pass_equivalent_units': True},
+    T={
+        'units': u.K, 'equivalencies': u.temperature_energy(),
+        'pass_equivalent_units': True
+      },
+    vx_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
+    vy_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
+    vz_drift={'units': u.m / u.s, 'pass_equivalent_units': True},
+    vTh={'units': u.m / u.s, 'pass_equivalent_units': True},
```
StanczakDominik, line 260: Unrelated comment, but: hey @rocco8773, this makes me wonder... what was the rationale on having `pass_equivalent_units` default to `False`?

plasmapy/formulary/distribution.py
```diff
@@ -234,28 +228,13 @@ def Maxwellian_velocity_2D(
...
-        vTh = parameters.thermal_speed(
-            T * u.K, particle=particle, method="most_probable"
-        ).si.value
+    if np.isnan(vTh):
+        # get thermal velocity and thermal velocity squared
+        vTh = parameters.thermal_speed(T, particle=particle, method="most_probable")
+    elif not np.isnan(vTh):
+        # check units of thermal velocity
+        vTh = vTh.to(u.m / u.s)
+
```
StanczakDominik, line 237: The execution speed might get faster if, after this initial handling step, we were to drop the units like this:

```python
vTh = vTh.si.value
```

And on output, stick them back on:
```python
return u.Quantity(distFunc, u.s/u.m)
```

Another trick to speed up astropy.units is to precompute the composite units elsewhere in the code:

```python
_m_s = _velocity_unit = u.m / u.s  # name is stylistic preference
```

then refer to that precomputed unit everywhere.

plasmapy/formulary/distribution.py
rocco8773: `pass_equivalent_units=True` will allow equivalent units to pass (I know 🙄) and will NOT convert them to be the expected units.  We wanted the default behavior of `validate_quantities` to actually convert the units when possible.  That's the point of the decorator, validate the quantity to what you expect.

StanczakDominik: Ooooooh. So we don't need these here at all, right?

rocco8773: Correct! In fact, you don't even need to include `vx`, `vy`, and `vz` in the validator since the validator is automatically triggered by the typing annotation.

rocco8773: You can do something more along the lines of...

```python
@validate_quantities(T={'units': u.K, 'equivalencies': u.temperature_energy()})
def Maxwellian_velocity_3D(
    vx: u.m / u.s,
    vy: u.m / u.s,
    vz: u.m / u.s,
    T: u.K,
    particle="e",
    particle="e",
    vx_drift: u.m / u.s = 0. * u.m / u.s,
    vy_drift: u.m / u.s = 0. * u.m / u.s,
    vz_drift: u.m / u.s = 0. * u.m / u.s,
    vTh: u.m / u.s = np.nan * u.m / u.s,,
) -> (u.s / u.m) ** 3:
```

Here's the [primer on `validate_quantities`](http://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.validate_quantities.html#plasmapy.utils.decorators.validate_quantities)

StanczakDominik: I somehow completely forgot we do it by annotations now! Thanks!

plasmapy/formulary/distribution.py
FilipBolt: Precomputing as the velocity unit definitely helps. I've done that throughout now. 

However, adding
```
    vTh = vTh.si.value
    vx = vx.si.value
    vx_drift = vx_drift.si.value
    vy = vy.si.value
    vy_drift = vy_drift.si.value
```

after the method invocation does not really help. I believe it is the unit conversions that make things slow (that happen in the ```validate_quantities```, not so much the computations between units. 

Just to give you a number on how much slower this is: executing **all** distribution.py tests in unitless mode takes around ```4.10 seconds```. When enforcing units, executing **all but the two 3D norm tests with units** takes around ```12.55 seconds```. 

Any other ideas?

rocco8773: A few ideas that may give some improvement.

1. Don't re-define the values.  I think doing `vTh = vTh.si.value` actually causes another unit check/conversion.  The `validate_quantities` has already done this, so just rely on it and use `vTh.value`.
1. This is a good case for using `pass_equivalent_units=True` in the `validate_quantities`.  This will make sure the passed unit can be converted to the expected unit but will never actually perform the conversion.  Then, just before you do your math, normalize the velocity units and cast them to dimensionless units (i.e `(vx / vTh).to(u.dimensionless_unscaled)`).  As long as the units are equivalent, there's no need to convert them ahead of time.  Doing it like this will limit the number of conversions, decrease any floating point errors due to conversions, and reduce floating point errors from doing the square [`(vx / vTh) ** 2` will be more accurate than `(vx ** 2) / (vTh ** 2)`].

Side note, I'm not going to quibble over the testing time for unit-less functions vs functions handling units.  Unit-less functions are always going to run quicker than functions handling units, that's the trade-off for having the convenience.  [*Of course I say that where the execution time is still within "reason," and we do our part to mitigate the increased execution time.*]  Additionally, if the correct units are pass then there should not be a significant difference between execution times.  Additionally, there are more test cases for functions handling units, so testing is going to take longer by default.  We really need to look at the statistics of the function executing, not the tests.

plasmapy/formulary/distribution.py
```diff
@@ -640,13 +547,17 @@ def Maxwellian_speed_2D(v, T, particle="e", v_drift=0, vTh=np.nan, units="units"
...
 
-def Maxwellian_speed_3D(v, T, particle="e", v_drift=0, vTh=np.nan, units="units"):
+@validate_quantities(T={'equivalencies': u.temperature_energy()})
+def Maxwellian_speed_3D(
+    v: _velocity_unit,
+    T: u.K,
+    particle="e",
+    v_drift: _velocity_unit = 0. * _velocity_unit,
+    vTh: _velocity_unit = np.nan * _velocity_unit,
+) -> (u.s / u.m):
```
rocco8773, line 560: I think we need to reconcile the use of `T`, `particle`, and `vTh` as arguments here, and all the distribution functions.  First off, the default value for `vTh` should be `None` (i.e. `vTh: _velocity_unit = None`) and then add `none_shall_pass=True` to the decorator.  This is a bit ugly though, since the function still requires you to pass-in `T` and `particle`. ¯\\_(ツ)_/¯

Currently I can only see two possible solutions to this...

1. Make `particle`, `T`, and `vTh` keywords and come up with an elaborate checking mechanism inside the function.  *Still feels ugly to me.*
2. Drop `particle` and `T`, and only require `vTh`.  Then in the docstrings have an explicit example using `parameters.thermal_speed()` to calculate `vTh` .

plasmapy/formulary/distribution.py
```diff
@@ -6,6 +6,7 @@
 import astropy as astropy
 from astropy import units as u
 from plasmapy.formulary import parameters
```
rocco8773, line 8: As a practice we should start importing only the functions we need for the module.  This is already a practice we want to start putting forward in our documentation and it will mitigate the loading of unnecessary objects for users.

```suggestion
from plasmapy.formulary.parameters import (
    kappa_thermal_speed,
    thermal_speed,
}
```

plasmapy/formulary/distribution.py
FilipBolt: Option 2 definitely sounds cleaner to me, but how many more clients do you think this will break? Your thoughts @StanczakDominik ?

plasmapy/formulary/distribution.py
FilipBolt: For point 1., you are absolutely right, it does more conversions when doing ```vTh.si.value``` and simply using ```vTh.value``` speeds things up. 

I don't quite get your second point. I believe you're trying to point out to convert to dimensionless if possible. I believe this already happens, as ```np.exp``` works only with dimensionless units. I do agree that doing ```(vx / vTh) ** 2``` is better than ```(vx ** 2) / (vTh ** 2)```, but the slow cases really manifest themselves in higher dimensions where I don't think you can do the same (I might be mistaken):

```
    expTerm = np.exp(
        ((-1) * ((vx- vx_drift) ** 2 + (vy- vy_drift) ** 2) / vTh ** 2)
        .to(u.dimensionless_unscaled)
    )
    coeff = (vThSq * np.pi) ** (-1)
    distFunc = coeff * expTerm
```

I could apply this to the one dimensional cases of course and run some stats on performance.

plasmapy/formulary/distribution.py
StanczakDominik: Does it actually load just the two functions in the latter case? Might be tough to benchmark this...

plasmapy/formulary/distribution.py
StanczakDominik: Perhaps there's a 3rd option we're missing here. @FilipBolt, in the issue you raised this possible design (I removed the `units` argument):
```python
dist = DistributionFactory.createDistribution(type='maxwellian', dimension=1, property='speed')
dist(v, T, particle="e", v_drift=0, vTh=np.nan)
```
But suppose we did it like this:

```python
dist = DistributionFactory.createDistribution(type='maxwellian', dimension=1, property='speed')
dist.from_temperature(v, T=T, particle="e", v_drift=0)
dist.from_thermal_velocity(v, vTh=vTh, v_drift=0)
```

what do you guys think about this one?

plasmapy/formulary/distribution.py
rocco8773: I'd approach it more like this...

```python
# calc exponential term
exp_xvar = ((vx - vx_drift) / vTh).to(u.dimensionless_unscaled).value
exp_yvar = ((vy - vy_drift) / vTh).to(u.dimensionless_unscaled).value
exp_term = np.exp((-1) * ((exp_xvar ** 2) + (exp_yvar ** 2)))

# calc normalization term
vTh = vTh.to(_velocity_units)
coeff = ((vTh ** 2) * n.pi) ** (-1)

return coeff * expm_term
```

plasmapy/formulary/distribution.py
rocco8773: From my recent research, yes.  Look at the examples section for [python's documentation on the `import` statement.](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement).  I've checked this behavior by examining the module level namespace, and only `kappa_thermal_speed` and `thermal_speed` would be present.  Now, does Python import the whole module and only binds the function to the namespace?  That might happen, but would require digging into the `import` code to pindown (I'm not volunteering for that!!!) and the Python documentation suggests that is not the case.

plasmapy/formulary/distribution.py
rocco8773: Another idea, queuing off of units...

```python
@validate_quantities(
    v={"pass_equivalent_units": True},
    sigma={"pass_equivalent_units": True},
    v_drift={"pass_equivalent_units": True},
)
def Maxwellian_1D(
    v: _velocity_units,
    sigma: [u.K, _velocity_units],
    particle="e",
    v_drift: _velocity_units = 0. * _velocity_units,
):

    if sigma.unit.is_equivalend(_velocity_units):
        # we got a velocity, assume sigma is vTh
        vTh = sigma
    else:
        # we got temperature
        vTh= thermal_speed(sigma, particle=particle, method="most_probable")

    # do calculations

    return distribution
```

I think this is my favorite so far.  Obviously, `sigma` would have to be explicitly explained in the docstrings.

FilipBolt: @StanczakDominik That looks great. That way you could have different interfaces to specify the distribution from and it is more explicit. 

@rocco8773 I think your solution is fine, but I prefer the one by Dominik, since it is more explicit and requires less documentation reading IMHO. 

Thanks very much for the suggestions. I would like to do this change as part of a subsequent  CR, which will involve refactoring the entire distribution.py to behave like a builder design pattern.

plasmapy/formulary/distribution.py
FilipBolt: I think there is no performance difference between the two ways: https://softwareengineering.stackexchange.com/questions/187403/import-module-vs-from-module-import-function

Based on very little research, python loads the entire module, but only binds the listed methods to the namespace.

plasmapy/formulary/distribution.py
StanczakDominik: Sounds good!

plasmapy/formulary/distribution.py
StanczakDominik: Either way, I'm willing to completely neglect the time spent importing stuff until benchmarks show it's becoming a problem :laughing: and I wouldn't bother enforcing a particular import style, either.

StanczakDominik:
All right, then, if I'm correct this looks mergeable. Am I right, @FilipBolt? :)

plasmapy/formulary/distribution.py
FilipBolt: Got it, thanks. I tried it out, and no significant improvement. 

Btw., this is essentially what I'm running to test this out: https://pastebin.com/5HhST7aV

```
In [31]: %timeit run_unitless()
24.5 ms ± 1.35 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
vs. 
```
In [32]: %timeit run_units()
11.1 s ± 559 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

StanczakDominik: Whoa, that's 3 orders of magnitude of a difference...

plasmapy/formulary/distribution.py
rocco8773: The thing is @StanczakDominik's solution does not solve the functional use case.  Yes, it makes sense if working from the factory class, but not working directly with `Maxwellian_1D`.  Directly using the function still requires the user to pass a `T` argument when he or she is trying to use the `vTh` argument.  This will even be a problem for designing the factory class.

I don't like my option 1 solution because I feel it's poor design to make `T` and `Vth` optional arguments, but require the use of one for the function to work.  By definition, they're optional arguments so the function should work without them.

This is why I come back around to my option 3.  It's explicit in the arguments it requires and should only add a few lines to the documentation.

If you want to be more explicit with the naming of the function, then a 4th (no 5th) option would be...

```python
def Maxwellian_1D(v, vTh):
    # do your stuff
    return distribution

def Maxwellian_1D_from_temperature(v, T, particle='e'):
    vTh = thermal_speed(t, particle=particle, method="most_probable")
    return Maxwellian_1D(v, vTh)
```

StanczakDominik: I'd argue that there's probably not much value in devotedly sticking to a functional approach... Only a C programmer deals in this kind of absolutes! ;)

rocco8773: That's find, but as a public facing API it shouldn't have bugs like this.  Give it a leading underscore and make the factory class the public facing API, then I'll turn around and walk away grumbling under my breath. 🤣

plasmapy/formulary/distribution.py
rocco8773: Looking over your tests I understand your timing issues better.  I didn't realize you were running an integration test, I thought it was just the package tests.  This does bring up an issue that I have definitely been sweeping under the rug.  These functions are inherently going to be integrated and there are not a lot of unit friendly integration routines out there.  Thus, I can definitely see the benefit of a unit-less version.

We could possibly do something like...

```python
def Maxellian_1D_unitless(v, ...):
    # do stuff
    return distribution

Maxwellian_1D = validate_quantities(v=_velocity_units)(Maxwellian_1D_unitless)
```

The main issue with this approach is documentation.

But now my inner @namurphy is screaming out "ABSTRACT BASE CLASSES" and we should convert everything to a class based structure instead.  Then we could add a builtin integrator.

Sorry @FilipBolt, I'm definitly diverging off the original path.  I think there's a lot to be discussed here, but my primary issue for this PR is how the arguments are setup (see my comments bellow...https://github.com/PlasmaPy/PlasmaPy/pull/816#discussion_r432645641 and https://github.com/PlasmaPy/PlasmaPy/pull/816#discussion_r432673709).

rocco8773: I could see a potential class structure looking like...

```python
class MaxwellianDistribution:
    def __init__(self, vTh: _velocity_units = None, particle='e', order=1, type='velocity'):
        self._vTh = vTh
        self._particle = particle
        self._T = None
        self._order = order
        self._type = type

        if vTh is None:
            warn("not temperature or thermal velocity set")

    @property
    def temperature(self) -> u.K:
        return self._T

    @temperature.setter
    def temperature(self, val: u.K, particle=None)
        # do all your conditioning, unit validations, calc vTh, etc.

    # make properties & setters for vTh and particle

    def _probability_density_unitless(self, v):
        # no units involved, assumed to be correct
        # do calculations
        return prob_dens

    @validate_quantities
    def probability_density(self, v: _velocity_units):
        # this would be unit flavored
        # this is analogous to our current functions
        # do your unit conditioning and call _probability_density_unitless()
        # add units back on
        return prob_dens

    @validate_quantities
    def probability(self, vstart: _velocity_units, vstop: _velocity_units):
        # this would perform the integral from vstart to vstop
        # condition the units and strip
        # integrate using _probability_density_unitless
        return prob
```

use would be like...

```python
>>> dist = MaxwellianDistribution()
>>> dist.temperature = 100 * u.K
>>> dist.probability_density(20 * u.m / u.s)
some value
>>> dist.probability(20 * u.m / u.s, 40 * u.m / u.s)
some value
```

FilipBolt: Ok, does this mean we are not comfortable with letting go of units for the sake of only calculating ```probability``` or in general? 

Everyone that does probability via integration themselves would need to migrate to a new interface to have it as efficient as now. 

I completely understand your concerns @rocco8773 and they I would say it is vital to have them resolved to before discussing a solution for a way forward.

plasmapy/formulary/distribution.py
FilipBolt: I was thinking to keep the functional version just for backwards compatibility, otherwise I'm   against the fully functional approach mostly because of a clean code approach. 

This is the idea what existing functions would do:
```
def Maxwellian_speed_1D():
    dist = DistributionFactory.createDistribution(type='maxwellian', dimension=1, property='speed')
    return dist.from_thermal_velocity(v, vTh=vTh, v_drift=0)
```

but the idea is to actually avoid allowing clients to do this in practice, but actually create the distribution via the factory, mostly because of software maintenance reasons. You can see that right now, you have 6 veeery similar docstrings and methods in ```distribution.py``` that can be difficult to review if you're making a simple change that might effect all of them (like changing all the signatures).

FilipBolt:
> All right, then, if I'm correct this looks mergeable. Am I right, @FilipBolt? :)

I'd like to have a conclusion over units/unitless before going forward, and to make sure we're all happy with it :).

plasmapy/formulary/distribution.py
FilipBolt: Resolving this :).

plasmapy/formulary/distribution.py
FilipBolt: Resolving this, reopen if you feel like it :).

rocco8773:
> Ok, does this mean we are not comfortable with letting go of units for the sake of only calculating probability or in general?

IMO, I think the public API should still have units.  That is consistent with the rest of our API and reduces chances of errors from miss-matched units.  I do understand a unit-less version has significant advantages to users, which can be provided with a scheme similar to the [class I mentioned above](https://github.com/PlasmaPy/PlasmaPy/pull/816#discussion_r432699884).

> I was thinking to keep the functional version just for backwards compatibility, otherwise I'm against the fully functional approach mostly because of a clean code approach.

I wouldn't worry too much about backwards compatibility at the moment.  The package is under significant development so backwards compatibility is not a requirement until we get to `v1.0.0`.  Of course we don't want to go randomly changing things, but I think this change is worthy.

@FilipBolt Would you be available for our [community discussion tomorrow](https://www.plasmapy.org/meetings/weekly/)?  I think this would be a good topic to cover during that meeting.

FilipBolt:
> IMO, I think the public API should still have units

Ok, sounds to me like enforcing units might not be the best option. In your experience, how to other parts of the library deal with units? Do they enforce units or offer unitless interfaces? Or is it tackled on a case-by-case basis (perhaps others have no need of unitless interfaces)?

@rocco8773  I'm happy to attend the community discussion tomorrow and hope to hear from you guys live :).

rocco8773:
@FilipBolt It's always nice to put a face and voice behind the avatar! Hope to see you tomorrow.

The rest of the `plasmapy.formulary` handles units very much like it's currently done here.  For the most part this execution time "issue" is not an issue since these functions are only being called once.  The ~300 us time penalty (see below) is not significant for a single use case compared to the added convenience/functionality.  These distributions functions are a special case because of the required integration, the functions are called enough times that the penalty becomes a problem.

image

PR #822, opened 2020-06-02
Officially delete `plasmapy.examples` (patch to PR 821)

rocco8773:
The `plasmapy.examples` sub-package became obsolete when PR #792 migrated the examples from `.py` files to jupyter notebooks and placed them in `./docs/notebooks`.  This PR deletes the essentially empty `plasmapy.examples` directory that was left over, as well as, a few references throughout the package.

This is a patch to PR #821, since not all need changes were implement when the merge happened.

PR #823, opened 2020-06-02
Move plasmapy.data to plasmapy.particles.data

### LINKED ISSUE #817: Move `plasmapy.data` to `plasmapy.particles.data`
> @rocco8773  will initiate PR on moving plasmapy.data to plasmapy.particles.data

https://docs.google.com/document/d/14ofIBJlXIoxYjtlGPiXeUWKIp7Ynull5Q5K8eeIDyWg/edit#

You signed up for it, mate :smile:

rocco8773:
Move `plasmapy.data` to `plasmapy.particles.data`.  `plasmapy.particles.data` contains particle data used to by various package modules and tests.

This is part of the PLEP7 migration.

Closes #817


rocco8773:
I will remove the draft mark when PR #822 is merged and this branch is updated.

plasmapy/particles/data/test/__init__.py
```diff
@@ -3,5 +3,5 @@
 
 import plasmapy
 
-rootdir = os.path.join(os.path.dirname(plasmapy.__file__), "data", "test")
+rootdir = os.path.join(os.path.dirname(plasmapy.__file__), "particles", "data", "test")
 file_list = glob.glob(os.path.join(rootdir, "*.[!p]*"))
```
StanczakDominik, line 7: Looking at this now, I bet this would fail with wheels, eggs or whatever... we should probably be using something like `pkgutil.get_data("plasmapy.particles.data.test", filename)` with an explicit list of files to be loaded.

Although... maybe I'm wrong, but it looks like we're not using this `file_list` line at all. Only `rootdir` is used; and that's in `plasmapy/classes/sources/tests/test_openpmd_hdf5.py`, where we test `HDF5Reader`.

Nevermind for now; I'll refactor this later.

rocco8773: I made changes changes to how `rootdir` is defined, which I think addresses your issues here and #824.  I also renamed `rootdir` -> `data_dir` to be a little more explicit.

PR #828, opened 2020-06-02
Repair notebook prolog

rocco8773:
1. Repair the link directing to the notebooks on GitHub
    * original link was directing to `doc/notebook/` instead of `docs/notebook/`
    * original link was trying to find a GitHub branch corresponding to the `release` variable defined in `conf.py`...now it's pointing directly to `master`
1. Repair the binder link
    * The original link was not properly finding the notebooks on GitHub
    * This was a useful tool for figuring out the correct link...https://mybinder.org/
1. Remove the `nbviewer` link.  The RTD page is already behaving as a viewer.


docs/conf.py
```diff
@@ -198,15 +198,19 @@
...
-      <p>This page was generated from
-        <a class="reference external" href="https://github.com/PlasmaPy/PlasmaPy/blob/{{ env.config.release|e }}/{{ docname|e }}">{{ docname|e }}</a>.
+      <p style="margin-bottom:0px">
+        This page was generated by 
+        <a href="https://nbsphinx.readthedocs.io/">nbsphinx</a> from
+        <a class="reference external" href="https://github.com/PlasmaPy/PlasmaPy/blob/master/{{ docname|e }}">{{ docname|e }}</a>.
+        <br>
         Interactive online version:
-        <a href="https://mybinder.org/v2/gh/PlasmaPy/PlasmaPy/{{ env.config.release|e }}?filepath={{ docname|e }}"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg" style="vertical-align:text-bottom"></a>.
+        <a href="https://mybinder.org/v2/gh/PlasmaPy/PlasmaPy/master/?filepath={{ docname|e }}"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg" style="vertical-align:text-bottom"></a>.
```
StanczakDominik, line 211: Admittedly this was mostly copypasted from the doc examples, but I think it'd make sense to - if we were to release 0.4 tomorrow - link from 0.4 docs to 0.4 examples. This way, you're linking from every version's docs to the master branch, which seems like an antipattern. 

Perhaps there's a way around this? If there's any module in the docs that'll have version information, that'd be `conf.py`!

docs/conf.py
```diff
@@ -222,6 +226,7 @@
           p.appendChild(document.createTextNode('.'));
         }
       </script>
+      -->
```
StanczakDominik, line 229: Let's just get rid of this whole `<script>` block rather than commenting it out!

docs/conf.py
rocco8773: In general and practice I'd agree that would be the better option.  However, we literally do not have `release` branch to point back to on GitHub.  The only options I can think of that don't require manual work during each deployment are...

1. Pointing back to a commit tree like this (https://github.com/PlasmaPy/PlasmaPy/blob/c1f9482e030beda87f65b5558d40322536d36dd0/docs/notebooks/physics.ipynb), but I don't know If sphinx knows anything about the commit ids.
1. If RTD the has any stored pacakge data and if we can point to that instead.

In order to get a fix in now, I say we merge this PR and open to improve the link back.

StanczakDominik: I'm willing to bet I can take https://github.com/PlasmaPy/PlasmaPy/blob/25c81278e74d6e765e738a973975d790e6b83865/docs/conf.py#L79-L85 adapt it to our release schema from https://github.com/PlasmaPy/PlasmaPy/releases and put it in there (via an f-string, to boot), then https://mybinder.org/ claims to take release tags too, and it's gonna work splendidly. But I do have to do the dishes now, so I'll do it in an hour or so.

Look to my commit at... wow, what time is it over there... er... look to my commit before the first light of the fourth day of June, at dawn, look to the, er, the commit's gonna come from your east so I guess that checks out.

StanczakDominik: Gah, versioning got in my way. When working off 0.3.1, we get 0.3.2 dev for some crazy reason... I'll have to do a couple of RTD tests to get this done. Future issue, definitely.

rocco8773: Ok, I think I go the notebooks to point to the git tree, but I setup a fallback to master.  Here's the current doc build where it seems to be working...https://plasmapy--828.org.readthedocs.build/en/828/notebooks/physics.html

rocco8773:
I thought you'd get a kick out of this one!

StanczakDominik:
I mean, yeah, looking at you delving into the depths of Jinja2 was... something. :laughing:

Issue #59, opened 2017-07-20
Some NumPy and SciPy operations drop units from arrays


namurphy:
A [known issue with `astropy.units`](http://docs.astropy.org/en/latest/known_issues.html#quantities-lose-their-units-with-some-operations) is that some NumPy operations take in arguments that have units, but return arguments that do not have units even though they should.  This issue requires an upstream fix in NumPy (e.g., https://github.com/numpy/numpy/issues/4164). The SciPy 2017 talk on [MetPy's Choice of Unit Support](https://www.youtube.com/watch?v=qCo9bkT9sow&index=49&list=PLYx7XA2nY5GfdAFycPLBdUDOUtdQIVoMf) delves into this in great detail.  An example from the talk:
```Python
>>> import numpy as np
>>> from astropy import units
>>> a = np.array([3.]) * units.m
>>> b = np.array([4.]) * units.m
>>> np.concatenate((a, b))
<Quantity [ 3., 4.] (Unit not initialised)>
```
There are [some workarounds](http://docs.astropy.org/en/latest/known_issues.html#quantities-float-comparison-with-np-isclose-fails) to some of these problems.
```Python
>>> np.isclose(500* u.km/u.s, 300 * u.km / u.s
UnitsError: Can only apply 'add' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
>>> np.isclose(500* u.km/u.s, 300 * u.km / u.s, atol=1e-8 * u.mm / u.s)
array([False], dtype=bool)
```
We cannot do anything from within PlasmaPy, but will need to wait for fixes within NumPy (e.g., new support for `__array_ufunc__` will improve overriding behavior for ufuncs).

namurphy:
A takeaway point: when writing functions that take Astropy Quantities as arguments, always include a unit test to test units.
```Python
>>> density_of_wombats = 18.7 * u.parsec**-3
>>> velocity_of_wombats = 0.04 * c
>>> flux_of_wombats = density_of_wombats * velocity_of_wombats  # second golden rule of astrophysics
>>> assert flux_of_wombats.si.unit == units.m**-2 * units.s**-1
```

Cadair:
iirc the `__array_ufunc__` stuff is in astropy 2.0 / numpy 1.13 so you could require these versions to avoid some issues.

There is a quantity aware version of `isclose` and `assert_all_close` in `astropy.test.helpers` I think.

[...]

PR #831, opened 2020-06-03
Thomson Scattering Module Initial Commit

pheuer:
This PR adds a file plasmapy/diagnostics/thomson.py which contain a code for simulating Thomson scattering spectra for Maxwellian plasmas in both the collective and non-collective regimes.

An associated file of tests is also added: plasmapy/diagnostics/tests/test_thomson.py. This file contains two tests that run the program in two regimes (collective and non-collective) and make sure the results match their current output values.


rocco8773:
Hi @pheuer, thanks again for doing this.  Like I said in our emails, I'll try to give a proper review by the end of the week.  To get started, look into...

1. Replacing the use of `quantity_input` with [`validate_quantities`](http://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.validate_quantities.html#plasmapy.utils.decorators.validate_quantities).
2. The [documentation page for diagnostics](https://github.com/pheuer/PlasmaPy/blob/master/docs/diagnostics/index.rst) will need to be updated.
3. It might be useful to create a jupyter notebook to show how the package works.  This would go in `./docs/notebooks` and we could add it to the `nbgallery` you see on the documentation page.

StanczakDominik:
I'd definitely like to also take a look at this functionality :)

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,187 @@
...
+    w = ws - wl
+
+    # Compute the wavenumbers in the plasma
+    # See Sheffield Sec. 1.8.1 and Eqs. 5.4.1 and 5.4.2
+    ks = np.sqrt(ws**2 - wpe**2)/C
+    kl = np.sqrt(wl**2 - wpe**2)/C
+
+    # Compute the wavenumber shift (required by momentum conservation)
+    sa = np.arccos(np.dot(probe_n, scatter_n))  # The scattering angle
+    k = np.sqrt(ks**2 + kl**2 - 2*ks*kl*np.cos(sa))  # Eq. 1.7.10 in Sheffield
```
StanczakDominik, line 145: Comments like this one tend to be better as variable names:
```suggestion
    scattering_angle = np.arccos(np.dot(probe_n, scatter_n))
    k = np.sqrt(ks**2 + kl**2 - 2*ks*kl*np.cos(scattering_angle))  # Eq. 1.7.10 in Sheffield
```

And since this variable is used in a single expression, if you're worried about line length, I'm fine even with going full Unicode:
```suggestion
    scattering_φ = np.arccos(np.dot(probe_n, scatter_n))
    k = np.sqrt(ks**2 + kl**2 - 2*ks*kl*np.cos(scattering_φ))  # Eq. 1.7.10 in Sheffield
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,187 @@
...
+    w_e = w - k*np.dot(fluid_vel, k_n)
+    w_i = w - np.matmul(ion_vel, np.outer(k, k_n).T)
+
+    # Compute the scattering parameter alpha
+    # expressed here using the fact that v_th/w_p = root(2) * Debye length
+    alpha = wpe/(np.sqrt(2)*k*vTe)
+
+    # Calculate the normalized phase velocities (Sec. 3.4.2 in Sheffield)
+    xe = w_e/(k*np.sqrt(2)*vTe)
+    xi = 1/np.sqrt(2)*np.outer(1/vTi, 1/k)*w_i
```
StanczakDominik, line 160: I'd kill for cleaner variable naming here :smile: no good ideas yet, though.

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,187 @@
...
+    # Sheffield Sec. 5.1
+    chiE = -0.5*np.power(alpha, 2)*ZPrime(xe)
+
+    chiI = np.zeros([fract.size, w.size], dtype=np.complex128)
+    for m in range(fract.size):
+        chiI[m, :] = -0.5*fract[m]*alpha**2*ion_z[m]*(Te/Ti[m])* \
+            ZPrime(xi[m, :])
+
+    # Calculate the logitudinal dielectric function
+    epsilon = 1 + chiE + np.sum(chiI, axis=0)
```
StanczakDominik, line 173: As a relative newbie to plasma theory... this would not be the same as http://docs.plasmapy.org/en/latest/api/plasmapy.formulary.dispersionfunction.plasma_dispersion_func.html#plasmapy.formulary.dispersionfunction.plasma_dispersion_func, right?

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,187 @@
...
+
+@u.quantity_input(wavelength=u.nm, probe_wavelength=u.nm, ne=u.cm**-3,
+                  Te=u.eV, Ti=u.eV, ion_vel=u.cm/u.s, e_drift=u.cm/u.s)
+def spectral_density(wavelength, probe_wavelength=532*u.nm, ne=1e15*u.cm**-3,
+                     fract=np.ones(1), Te=1*u.eV, Ti=np.ones(1)*u.eV,
+                     ion_z=np.ones(1), ion_mu=np.ones(1),
+                     fluid_vel=np.zeros([3])*u.cm/u.s,
+                     ion_vel=np.zeros([1, 3])*u.cm/u.s,
+                     probe_n=np.array([1, 0, 0]),
+                     scatter_n=np.array([0, 1, 0])):
```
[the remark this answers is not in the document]
pheuer: Thank you for the suggestion about the decorator: I've changed it to now use validate_quantities. 

Most of the keyword arguments are intentional: 

- The probe wavelength could in principle be anything, but frequency multiples of the typical 1064 nm lasers are commonly used: 532 is probably the most common in practice. 
- The fract=[1] for a single species should be the default.
- The ion_vel and fluid_vel corresponding to no drift velocity seems reasonable as a default.
- The probe_n and scatter_n correspond to a 90 degree scattering geometry that is the simplest setup. 

There is no good reason for ne, Te, or Ti to have default values: what is the best practice in this case? Should I set the default value to None or something?

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,187 @@
...
+    fluid_vel = fluid_vel.to(u.cm/u.s)
+
+    # Define some constants
+    C = 2.99792458e10*u.cm/u.s  # speed of light
+    Me = 5.685e-16*u.eV/(u.cm/u.s)**2  # Electron mass in eV per (cm/s)^2
+    Mp = Me*1836.1  # Proton mass in same units
+    Mi = ion_mu*Mp  # mass of each ion species in same units
+    vTe = np.sqrt(Te/Me)  # Electron thermal velocity
+    vTi = np.sqrt(Ti/Mi)  # Ion thermal velocity
+    wpe = 5.64e4*(u.cm**1.5*u.rad/u.s)*np.sqrt(ne)
```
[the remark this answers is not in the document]
pheuer: Great! I've updated the code to use these (with the exception of the plasma frequency, which still has a numeric constant hard-coded)

plasmapy/diagnostics/thomson.py
pheuer: I see your point: if it makes it any better x_e and x_i are the conventional notation in the literature (including the Sheffield book I cite). I can't think of a more concise way to say "phase velocities normalized to thermal velocities"

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
StanczakDominik: Yeah, keyword args are definitely the way to go here. Perhaps we don't need the defaults, though: we can use [keyword-only arguments from PEP 3102!](https://www.python.org/dev/peps/pep-3102/#id4)

> The probe wavelength could in principle be anything, but frequency multiples of the typical 1064 nm lasers are commonly used: 532 is probably the most common in practice.

Thanks, today I learned something :laughing: ! In that case, I'm fine with leaving that in. Same goes for fract, ion_vel, probe_n and scatter_n - though for the latter two, more descriptive names would be nice to have.

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
StanczakDominik: http://docs.plasmapy.org/en/latest/api/plasmapy.formulary.parameters.plasma_frequency.html#plasmapy.formulary.parameters.plasma_frequency this function is sad now :( Would it be inappropriate to use it here?

plasmapy/diagnostics/thomson.py
pheuer: ZPrime is just the derivative of that function! I'm actually importing it from that same package. 

The susceptibilities can be expressed directly in terms of Z, but expressing them in terms of Z' is more compact.

plasmapy/diagnostics/thomson.py
StanczakDominik: ... `normalized_phase_ve`... nah. Okay, I give up.

pheuer:
Thanks for the comments! In addition to the changes noted above, I have also created an example Jupyter notebook and *attempted* to edit the diagnostics page in the documentation (although I'm not sure if there's more I need to do in the actual function docstrings to make that work properly?).

StanczakDominik:
I'm also getting some errors with tests on older Astropy versions at https://dev.azure.com/plasmapy/PlasmaPy/_build/results?buildId=993&view=logs&j=8955e93f-e95e-5ac2-4406-a311ed91aa3e&t=adda2fec-11c4-56e4-c47d-199954e67231&l=5683 (seems to point to some `rad/cm` vs `rad/s` inconsistency in the code?), and the documentation build points to warnings on section titles (as mentioned before): https://app.circleci.com/pipelines/github/PlasmaPy/PlasmaPy/287/workflows/044fac81-3940-4f64-8970-dda8018fbd7d/jobs/6586/parallel-runs/0/steps/0-107. I'm not sure Numpydoc has a separate header for keyword-only arguments, but just cleaning up Arguments may help.

StanczakDominik:
Oh yeah, if the fix to the units thing turns out to be hard to figure out, I think I'm fine with bumping up the minimal astropy version. They've been pumping those out like crazy.

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
pheuer: Oh, yes, that works! nice.

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
pheuer: How about scatter_vec and probe_vec?

docs/diagnostics/index.rst
```diff
@@ -8,10 +8,17 @@ Plasma diagnostics (`plasmapy.diagnostics`)
 
 .. automodapi:: plasmapy.diagnostics.langmuir
    :no-heading:
+   
+.. currentmodule:: plasmapy.diagnostics.thomson
+
+.. automodapi:: plasmapy.diagnostics.thomson
+   :no-heading:
```
StanczakDominik, line 15: Oh right! I missed it, but this is not going to be necessary; it'll be simpler for me to send you a PR to this PR :laughing: Expect it soon!

pheuer:
Ok I'm still looking into the astropy.units issue: sifting through the logs the issue seems to be at (currently) line 146: w_e = w - k*np.dot(fluid_vel, k_vec). The units are w->rad/s, k->rad/cm, fluid_vel->cm/s, and k_vec is dimensionless. It looks like, on the older astropy version, the second term is ending up with units of rad/cm and then the error throws when it tries to do the subtraction?

docs/diagnostics/index.rst
pheuer: Done!

StanczakDominik:
> The numpydoc issue should be resolved: There is now only one heading for "parameters" which is their standard. Although, sidebar, I at least find it hard to distinguish between positional and keyword args in this case in the documentation?

Yeah, it is pretty hard to distinguish... @rocco8773, given that you usually have Opinions about documentation layout (and good ones at that!) - any suggestions here?

StanczakDominik:
Wrong hotkey, posted too quickly...

> Ok I'm still looking into the astropy.units issue: sifting through the logs the issue seems to be at (currently) line 146: w_e = w - k*np.dot(fluid_vel, k_vec). The units are w->rad/s, k->rad/cm, fluid_vel->cm/s, and k_vec is dimensionless. It looks like, on the older astropy version, the second term is ending up with units of rad/cm and then the error throws when it tries to do the subtraction?

It rang no bells at first, but looking at the formula and seeing `np.dot` reminded me about #59, coming back to bite us in the posterior ~700 issues/PRs later! Given that this was fixed in some later Astropy version, I vote to just bump the minimal req for that. I'm still going to test it to make sure this is it.

StanczakDominik:
Hahahaha, well done, good timing :rofl: 4.0 it is, then.

rocco8773:
> > The numpydoc issue should be resolved: There is now only one heading for "parameters" which is their standard. Although, sidebar, I at least find it hard to distinguish between positional and keyword args in this case in the documentation?
> 
> Yeah, it is pretty hard to distinguish... @rocco8773, given that you usually have Opinions about documentation layout (and good ones at that!) - any suggestions here?

Yeah, this list is of parameters is fairly long.  If there's some rarely used keywords, then you can add them into an [**Other Parameters**](https://numpydoc.readthedocs.io/en/latest/format.html) section.  Otherwise, I'd stick to just the **Parameters** section for consistency.  To distiguish keywords from argument, I usually explicitly state the default value in their definition.

@pheuer To help yourself distinguish arguments vs keywords in the future, look at the function signature.  Arguments will always be first in line and keywords will always have default values.

StanczakDominik:
> > > The numpydoc issue should be resolved: There is now only one heading for "parameters" which is their standard. Although, sidebar, I at least find it hard to distinguish between positional and keyword args in this case in the documentation?
> > 
> > 
> > Yeah, it is pretty hard to distinguish... @rocco8773, given that you usually have Opinions about documentation layout (and good ones at that!) - any suggestions here?
> 
> Yeah, this list is of parameters is fairly long. If there's some rarely used keywords, then you can add them into an [**Other Parameters**](https://numpydoc.readthedocs.io/en/latest/format.html) section. Otherwise, I'd stick to just the **Parameters** section for consistency. To distiguish keywords from argument, I usually explicitly state the default value in their definition.
> 
> @pheuer To help yourself distinguish arguments vs keywords in the future, look at the function signature. Arguments will always be first in line and keywords will always have default values.

To clarify - we're talking about having a single positional argument here (wavelength) and a few being required (non-optional) keyword-only arguments. What would you recommend for that?

Re: astropy bump, I'm making a separate PR for that. Looking at the test results, it'll definitely fix the bug here.

pheuer:
Great, thanks for resolving that issue. 

Unfortunately all of the arguments for this function are required (except those with default arguments I suppose). Now that I've corrected the numpydoc formatting the documentation is a little more clear online so I think wavelength is now distinguishable as a positional argument there.

rocco8773:
> Hahahaha, well done, good timing 🤣 4.0 it is, then.

My original take was to setup a `try-except` statement to handle to the two version, but the more I think about it I'm good with changing our dependencies to `astropy >= 4.0`.  It's been out for nearly half a year already.

rocco8773:
> > > > The numpydoc issue should be resolved: There is now only one heading for "parameters" which is their standard. Although, sidebar, I at least find it hard to distinguish between positional and keyword args in this case in the documentation?
> > > 
> > > 
> > > Yeah, it is pretty hard to distinguish... @rocco8773, given that you usually have Opinions about documentation layout (and good ones at that!) - any suggestions here?
> > 
> > 
> > Yeah, this list is of parameters is fairly long. If there's some rarely used keywords, then you can add them into an [**Other Parameters**](https://numpydoc.readthedocs.io/en/latest/format.html) section. Otherwise, I'd stick to just the **Parameters** section for consistency. To distiguish keywords from argument, I usually explicitly state the default value in their definition.
> > @pheuer To help yourself distinguish arguments vs keywords in the future, look at the function signature. Arguments will always be first in line and keywords will always have default values.
> 
> To clarify - we're talking about having a single positional argument here (wavelength) and a few being required (non-optional) keyword-only arguments. What would you recommend for that?

My take, keyword parameters should NEVER be required.  Their implementation makes them by default optional.  So, something like this is just bad practice...

```python
def foo(x=None):
    if x is None:
        raise ValueError("x can not be a None value")
```
Note though, in the current `spectral_density` function all arguments are positional and keyword, so you can still call the function like `spectral_density(wavelength=wavelength)`.

In summary, if a parameter is required then make it a positional argument (which can still be called like a keyword argument).  If a parameter has default behavior if not given, then make it a keyword argument.

StanczakDominik:
https://github.com/PlasmaPy/PlasmaPy/pull/831#discussion_r435447584 I should probably have linked this in the first place, resolved discussions can be hard to find and easy to miss, sorry :( https://www.python.org/dev/peps/pep-3102/#id4 has this sort of thing:

```python
def compare(a, b, *, key):
    ...
```
No default value on `key`, but it can still only be input explicitly with the variable name. This is what I was going for in review, and completely forgot about it later...

rocco8773:
> [#831 (comment)](https://github.com/PlasmaPy/PlasmaPy/pull/831#discussion_r435447584) I should probably have linked this in the first place, resolved discussions can be hard to find and easy to miss, sorry :( https://www.python.org/dev/peps/pep-3102/#id4 has this sort of thing:
> 
> ```python
> def compare(a, b, *, key):
>     ...
> ```
> 
> No default value on `key`, but it can still only be input explicitly with the variable name. This is what I was going for in review, and completely forgot about it later...

?? Do you mean?

```python
def compare(a, b, *, key=None):
    ...
```

This syntax just enforces that everything after `*` is a keyword-only argument. ([types of paramters](https://docs.python.org/3/glossary.html#term-parameter))

StanczakDominik:
Nope, actually, it gets cooler than that!
```python
[ins] In [1]: def compare(a, b, *, key): 
         ...:     print(key) 
         ...:     return a ==b 
         ...:                                                                                          

[ins] In [2]: compare(1, 2)                                                                            
<ipython-input-2-5c7da606c74f> in <module>
----> 1 compare(1, 2)

TypeError: compare() missing 1 required keyword-only argument: 'key'

[ins] In [3]: compare(1, 2, 3)                                                                         
<ipython-input-3-b0c1df51f863> in <module>
----> 1 compare(1, 2, 3)

TypeError: compare() takes 2 positional arguments but 3 were given

[ins] In [4]: compare(1, 2, key = "I AM A POTATO")                                                     
I AM A POTATO
Out[4]: False

```

rocco8773:
Oh weird! I've never seen that before.  Not sure how I feel about it, but I'm warming up to it.  I'm surprised Python does not showcase that better.  This is the only line from [PEP 3102](https://www.python.org/dev/peps/pep-3102/#specification) I've ever seen on this...

> Keyword-only arguments are not required to have a default value. Since Python requires that all arguments be bound to a value, and since the only way to bind a value to a keyword-only argument is via keyword, such arguments are therefore 'required keyword' arguments. Such arguments must be supplied by the caller, and they must be supplied via keyword.

rocco8773:
@pheuer How do you typically use this in your workflow?  Do you manually generate it to overlay onto collected data or do you try to fit it?

I'm trying to decide if we should be a classed based approach, kinda like this [comment](https://github.com/PlasmaPy/PlasmaPy/pull/816#discussion_r432699884).  If there is ever a need for fitting or integrating, then a unit-less access could be useful while maintaining the common API as a unit based one.

pheuer:
@rocco8773 

*Disclaimer: I've never actually processed Thomson scattering data in the wild, I've been using this to help plan out a new LAPD diagnostic* 

I think the use-case is 50/50 predictive and fitting. I have done some experimenting fitting synthetic data (using curve_fit and a lambda function to isolate just a few fitting parameters), so that would be good functionality to support so I guess  a unit-less function would be worth having. 

I'm not sure I understand the advantage of the class structure you linked to - why not just include a wrapper  function that takes float inputs and assumes some units when it calls the main function (or vice-versa)?

StanczakDominik:
Maybe let's keep it simple for now? You could apply that logic to most of `formulary`. It probably doesn't need to be that hyper optimized.

Besides, if you're integrating this density, seems like you can quickly calculate it first, then dump units and integrate numerically. It's only a problem for automatic integration routines, and I'm sure we'll figure that out sooner or later.

Off to bed for me! Sorry if I seem grumpy.

lemmatum:
Hey, just chiming in. You can greatly simplify the code in `thomson.py` by using `permittivity_1D_Maxwellian` from `dielectric.py`. This calculates your chi_e and chi_i for you. Thomson scattering simulations is one of the reasons I wrote that code up... 

Also, I've got a little Thomson scattering code I've been working on. It might help add functionality to this, whenever this merge is done. Only issue with it is I've got a bug somewhere which is breaking detailed balance (the upshifted and downshifted plasmon peaks don't look right).

I agree with @pheuer that this module can be used for both prediction and fitting. I mostly see people doing Thomson scattering fits "by hand", meaning they vary the parameters and eyeball the best fit. Sometimes they do a chi-squared. Ideally, they should be doing something that can handle multi-dimensional fits and give error bars, like Markov Chain Monte Carlo. It might be too much to ask for fitting right now... having the code simulate spectra is already quite powerful and users can build their own fitting routines around it as necessary.

Last point: I see you have an IPython notebook. I think you should translate that into an example for the gallery.

Let me know if you guys need any more help with this. I've done Thomson scattering experiments and analysis before (mainly x-ray).

rocco8773:
Sleep tight Mr Grumpy. 🤣

If you're going to have numerous functions and data that are highly dependent on each other, then classes do a better job of encapsulating everything (both for functionality and readability).  I think that's the direction this would go in the future, but, as @StanczakDominik said, it does not need to happen for this PR.

Additionally, I'm currently prototyping the Diagnostic Framework, which will hopefully come online in a few months.  Once that does, this will need to be refactored to merge into the framework.

rocco8773:
> I agree with @pheuer that this module can be used for both prediction and fitting. I mostly see people doing Thomson scattering fits "by hand", meaning they vary the parameters and eyeball the best fit. Sometimes they do a chi-squared. Ideally, they should be doing something that can handle multi-dimensional fits and give error bars, like Markov Chain Monte Carlo. It might be too much to ask for fitting right now... having the code simulate spectra is already quite powerful and users can build their own fitting routines around it as necessary.

Yeah, this is what I was trying to get at.  I think functionality like this would be very useful to the community, but does not need to be implemented now.

rocco8773:
> Last point: I see you have an IPython notebook. I think you should translate that into an example for the gallery.

Yes, it is currently setup like that.

https://plasmapy--831.org.readthedocs.build/en/831/examples.html
https://plasmapy--831.org.readthedocs.build/en/831/diagnostics/thomson.html

lemmatum:
> > Last point: I see you have an IPython notebook. I think you should translate that into an example for the gallery.
> 
> Yes, it is currently setup like that.
> 
> https://plasmapy--831.org.readthedocs.build/en/831/examples.html
> https://plasmapy--831.org.readthedocs.build/en/831/diagnostics/thomson.html

Ah sweet! I missed that!

lemmatum:
Looking through that example in the gallery, it looks like detailed balance isn't implemented in this. The upshifted and downshifted peaks are symmetric (same shape and signal level) when in reality the downshifted peak is stronger than the upshifted peak. We don't need to add this now, but it definitely needs to be added in the future.

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,181 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For each example, plot the the spectral distribution function over a large range to show the broad electron scattering feature (top row) and a narrow range around the probe wavelength to show the ion scattering feature (bottom row)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
```
rocco8773, line 106: In order to get a thumbnail for the `nbsphinx` gallery, you need add metadata to a specific cell to tell `nbsphinx` where it should extract the image from.  This cell is probably the best candidate.  The metadata would look something like...

```suggestion
   "metadata": {
    "nbsphinx-thumbnail": {
     "tooltip": "I am a tool tip.",
     "output-index": 3
       }
   },
```
https://nbsphinx.readthedocs.io/en/0.7.0/gallery/cell-metadata.html
https://nbsphinx.readthedocs.io/en/0.7.0/gallery/multiple-outputs.html

docs/diagnostics/index.rst
```diff
@@ -4,14 +4,8 @@
 Plasma diagnostics (`plasmapy.diagnostics`)
 *******************************************
 
-.. currentmodule:: plasmapy.diagnostics.langmuir
+.. toctree::
```
rocco8773, line 7:
```suggestion
.. automodapi:: plasmapy.diagnostics
    :no-heading:

.. toctree::
```

This will add the docstring from `plasmapy/diagnostics/__init__.py`.

docs/diagnostics/index.rst
```diff
@@ -4,14 +4,8 @@
...
+   :maxdepth: 1
```
rocco8773, line 8:
```suggestion
   :maxdepth: 1
   :caption: Sub-Packages and Modules
```

This will give the table-of-contents a title.

pheuer:
@lemmatum Thanks for pointing out the dielectric functions! They didn't shorten the code much (because I still need to compute the phase velocities), but it's better to not re-invent the wheel anyway.  I also think I found a sqrt(2) error in my previous code while converting over to your functions.

While I was changing that, I realized it was more convenient to use the plasmapy.Particle class to allow users to input ion species, so I've changed over to doing that.

Regarding the height of the peaks: my understanding is that this arises from the geometric factor G(k,w) ( [https://aip.scitation.org/doi/10.1063/1.3478975](as described here)). I didn't include this in S(k,w) because it seems to be written separately in most of theory and, in the regimes I'm used to working in, it's effectively unity.  However I know for x-ray TS this factor becomes very important (as v/c gets larger), so we should find a way to deal with it. I could include the first-order correction easily, but I think for x-ray spectra you really need a fully relativistic treatment of everything?

I hadn't heard the connection to detailed balance before: if you have a reference on that I'd be interested!

pheuer:
@rocco8773 Thanks for the suggestions: I've accepted the ones i could and manually edited the __init__.py file. I did add the metadata to the notebook, but I don't see a thumbnail yet.

rocco8773:
> @rocco8773 Thanks for the suggestions: I've accepted the ones i could and manually edited the **init**.py file. I did add the metadata to the notebook, but I don't see a thumbnail yet.

Yeah, I was playing around with that myself.  It looks like the output is being treated as one image, versus separate images for each sub-figure.  You can either remove the `"output-index"` metadata or re-work the plots so each sub-figure is its own image.  The former sounds way easier at the moment.

lemmatum:
@pheuer Oh yea, the phase velocities and all that stuff does take up a bunch of space... Now that I think about it, perhaps those should be separated out into their own functions instead of hiding in them in one monolithic function? They may be useful on their own. Same goes for the scatter vectors.

Yea separating it into a geometrical factor or something else makes sense. I just meant that we should implement detailed balance somewhere, and not that it necessarily has to go into the DSF. The Ross paper shows optical Thomson scattering spectra with asymmetric peaks and explicitly mentions that the relativistic corrections are important in this regime, even for low temperatures, so this is not just an XRTS term.

Equation 11.2.12 in the Froula, Glenzer, Luhmann, Sheffield book shows the detailed balance term along with a couple references particular to XRTS. The gist of it is that in LTE you have detailed balance and that balance means that the probe beam is more likely to deposit energy into the plasma than the plasma organizing itself in such a way that it donates energy to the scattered probe beam. This can break when you don't have LTE (I wrote a paper on that if you're interested).

The geometrical factor presented in Ross looks a little different from the detailed balance factor for XRTS, so whenever we implement this we should give the user options :smile:

docs/notebooks/thomson.ipynb
StanczakDominik: And we should really document that as well, shouldn't we...

StanczakDominik:
Nice to see you back here, @lemmatum :tada: 

>  I mostly see people doing Thomson scattering fits "by hand", meaning they vary the parameters and eyeball the best fit. Sometimes they do a chi-squared. Ideally, they should be doing something that can handle multi-dimensional fits and give error bars, like Markov Chain Monte Carlo. It might be too much to ask for fitting right now..

That's kind of sort of what my masters' thesis is turning out to be. [PyMC3](https://docs.pymc.io/) is a neat library! I should be able to contribute this back here soon-ish, and for now.

plasmapy/diagnostics/thomson.py
```diff
@@ -11,75 +11,78 @@
...
 # atomic species.
 
-@validate_quantities(wavelength=u.nm, probe_wavelength=u.nm, ne=u.cm**-3,
-                  Te=u.eV, Ti=u.eV, ion_vel=u.cm/u.s, fluid_vel=u.cm/u.s)
-def spectral_density(wavelength, probe_wavelength=532*u.nm, ne=1e15*u.cm**-3,
-                     fract=np.ones(1), Te=1*u.eV, Ti=np.ones(1)*u.eV,
-                     ion_z=np.ones(1), ion_mu=np.ones(1),
+@validate_quantities(wavelengths=u.nm, probe_wavelength=u.nm, ne=u.cm**-3,
+                  kTe=u.eV, kTi=u.eV, ion_vel=u.cm/u.s, fluid_vel=u.cm/u.s)
+def spectral_density(wavelengths, probe_wavelength, ne, kTe, kTi,
```
StanczakDominik, line 29: I completely forgot about the possibility of converting the input to `Particle`s! If you're doing it that way, you might find http://docs.plasmapy.org/en/latest/api/plasmapy.particles.particle_input.html helpful!

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Notes
+    -----
+    For details, see "Plasma Scattering of Electromagnetic Radiation" by
+    Sheffield et al. ISBN 978-0123748775. This code is a modified version of
+    the program described therein.
+
+    For a concise summary of the relevant physics, see Chapter 5 of Derek
+    Schaeffer's thesis, DOI: 10.5281/zenodo.3766933
+
```
rocco8773, line 105:
```suggestion
    For details, see "Plasma Scattering of Electromagnetic Radiation" by
    Sheffield et al. `ISBN 978\\-0123748775`_. This code is a modified version of
    the program described therein.

    For a concise summary of the relevant physics, see Chapter 5 of Derek
    Schaeffer's thesis, DOI: `10.5281/zenodo.3766933`_.
    
    .. _`ISBN 978\\-0123748775`: https://www.sciencedirect.com/book/9780123748775/plasma-scattering-of-electromagnetic-radiation
    .. _`10.5281/zenodo.3766933`: https://doi.org/10.5281/zenodo.3766933
```

Add links for resources.

rocco8773:
I think the notebook is a good start, but it needs to me more than just informational.  It also needs to be educational. Can you updated it with some images and more physics?

plasmapy/diagnostics/thomson.py
rocco8773: I made a [comment](https://github.com/PlasmaPy/PlasmaPy/pull/831#discussion_r436305743) to move towards using the `Particle` class.  Unfortunately, the `ion_species` argument can be a list of of `Particles`, which `particle_input` can not handle (Correct me if I"m wrong on that).

StanczakDominik: It's a tough little decorator, it'll handle it :)

lemmatum:
> That's kind of sort of what my masters' thesis is turning out to be. [PyMC3](https://docs.pymc.io/) is a neat library! I should be able to contribute this back here soon-ish.

Sweet! Teach me your ways! :mage_man:

plasmapy/diagnostics/thomson.py
rocco8773: The issue is...`particle_input` does not recognize more detailed typing annotations.  For example, it will know a typing of just `Particle` but not `List[Particle]`.

image

pheuer:
@rocco8773 

Thanks for the suggestions, here are a few questions/comments:

1) validate_quantities throws an ImplicitUnitWarningConversion warning if you use units if eV: is there a way to silence this? Entering temperatures in eV is almost definitely going to be the norm, so it should not print warnings. 

2) I've added some more details to the docstring, including an equation that I hope helps make clear what S(k,w) is. The full derivation of this expression is the subject of several chapters in the Sheffield book, so ultimately users will have to look there for a fundamental understanding. 

3) I have added some more explanation to the notebook: I cannot currently think of any other plots that I think would be informative. Eventually it would be nice to have other functions in this module that treat other cases (magnetized plasmas, relativistic effects, calculating power spectra) which would then invite a more detailed explanation of how all these parts fit together.

pheuer:
@lemmatum Thank you for your explanation: that makes perfect sense. I will check out your paper too. 

I think the G(k,w) in Ross is a 1st order limit in v/c, but I gather that a more complete treatment is necessary for XRTS. Assuming the full version is not too computationally taxing and reduces correctly in the non-relativistic limit, I suppose that would be the thing to implement, but I have a lot more to learn before I could start such a project...

Ideally I think this module would have different functions for each term in the scattered power spectrum (as well as a wrapper function that uses them all to produce a predicted spectrum).

rocco8773:
> 1. validate_quantities throws an ImplicitUnitWarningConversion warning if you use units if eV: is there a way to silence this? Entering temperatures in eV is almost definitely going to be the norm, so it should not print warnings.

The is put there on purpose to indicate that a non-standard unit conversion is occurring.  With the way warnings work, that will only appear for the first occurrence in a python session.  It's up to the user to filter that out if he or she doesn't want it. If you don't want it to occur in your own scripts then you can add something like `warnings.simplefilter("ignore", category=ImplicitUnitConversionWarning)` to the top of your script. (see [Python's warning suppression](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)).

rocco8773:
> 2. I've added some more details to the docstring, including an equation that I hope helps make clear what S(k,w) is. The full derivation of this expression is the subject of several chapters in the Sheffield book, so ultimately users will have to look there for a fundamental understanding.

Ok, I'll look it over.  I'm not looking for anything detailed, just enough so users understand what's being calculated and a few references so users have somewhere to start searching for more.

StanczakDominik:
I've merged #835 in, so that should solve the minimal-conda test failure. Please ignore the flake8 and pre-commit test failures; they're a side effect of work in #845/#846.

I wouldn't want to make this review cycle too long (ekhm transport - @pheuer if you don't understand this little in-joke, be very glad that you don't :laughing:), so I think we should be heading towards merging this, and possibly adding additional fixes/features in follow-ups. I'll give this another look-through now.

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,257 @@
...
+from plasmapy.particles import Particle
+from plasmapy.utils.decorators import validate_quantities
+from typing import List, Tuple, Union
+from warnings import warn
+
+
+# TODO: interface for inputting a multi-species configuration could be
+# simplified using the plasmapy.classes.plasma_base class if that class
+# included ion and electron drift velocities and information about the ion
+# atomic species.
```
StanczakDominik, line 25: This is the kind of comment I like to see!

I think we should have a general plasma class design issue to track this sort of thing. Note to self...

pheuer:
@StanczakDominik thanks, I updated the notebook as you suggested. I also merged the master branch in, but it doesn't look like it solved the py36-minimal-conda test problem? 

Regarding the module overall, I agree that it would be best to merge this function into master before contemplating further functionality. There's a whole mess of Thomson-related functionality that could be added, but that will take time and should mostly be independent of this function. Maybe we can create a new PR for supporting the relativistic correction/detailed balance that @lemmatum was mentioning after this one is resolved.

StanczakDominik:
> I also merged the master branch in, but it doesn't look like it solved the py36-minimal-conda test problem?

Okay, that's pretty weird. I'll give it another look.

StanczakDominik:
I guess the problem was in Numpy, and my previous attempt at bumping its requirements aimed too low.

I also merged `master` again (something seems to have gone wrong and you didn't have the actual current version) and applied `pre-commit run --all-files` to shut the style linter up. It applies the `black` code formatter.

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,257 @@
...
+    "For each example, plot the the spectral distribution function over a large range to show the broad electron scattering feature (top row) and a narrow range around the probe wavelength to show the ion scattering feature (bottom row)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "nbsphinx-thumbnail": {
+     "tooltip": "I am a tool tip."
+    }
```
StanczakDominik, line 181: Indeed you are. :laughing:

rocco8773:
@StanczakDominik Yeah, I still want to do another once over before merging.  I've been busy the last few days, but I'll trying to review by end of Monday.

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+
+    ion_species : str or Particle class list, shape [N]
+        A list of either instances of the Particle class or strings
+        interpretable by that class representing each ion species.
+        Default is ['H+'] corresponding to a single species of hydrogen ions.
+
+    fluid_vel : astropy.units.Quantity ndarray shape [3]
+        Velocity of the fluid or electron component in the rest frame in
+        units convertable to cm/s. Defaults to [0, 0, 0], representing
+        a stationary plasma.
```
rocco8773, line 105:
```suggestion
    fluid_vel : `~astropy.units.Quantity`, shape (3, ), optional
        Electron fluid velocity in the rest frame. (convertible to cm/s) 
        Defaults to a stationary plasma [0, 0, 0] m/s.
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+
+    ion_vel : astropy.units.Quantity ndarray, shape [N,3]
+        Velocity vectors for each ion population relative to the
+        fluid velocit in units convertable to cm/s. Defaults zero drift
+        for all specified ion species.
```
rocco8773, line 110:
```suggestion
    ion_vel : `~astropy.units.Quantity`, shape (N, 3), optional
        Velocity vectors for each ion population relative to the
        fluid velocity. (convertible to cm/s) Defaults zero drift
        for all specified ion species.
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+
+    probe_vec : float ndarray, shape [3]
```
rocco8773, line 112:
```suggestion
    probe_vec : float `~numpy.ndarray`, shape (3, )
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+        Unit vector in the direction of the probe laser. Defaults to
+        [1, 0, 0].
+
+    scatter_vec : float ndarray, shape [3]
```
rocco8773, line 116:
```suggestion
    scatter_vec : float `~numpy.ndarray`, shape (3, )
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+        Unit vector pointing from the scattering volume to the detector.
+        Defaults to [0, 1, 0] which, along with the default probe_vec,
+        corresponds to a 90 degree scattering angle geometry.
+
+    Returns
+    -------
+    alpha : float
+        mean scattering parameter. alpha > 1 corresponds to collective
+        scattering, while alpha < 1 indicates non-collective scattering.
```
rocco8773, line 125:
```suggestion
        Mean scattering parameter, where `alpha` > 1 corresponds to collective
        scattering and `alpha` < 1 indicates non-collective scattering.
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Skw : astropy.units.Quantity ndarray
+        Computed spectral density function over the input wavelength array
+        with units of s/rad.
```
rocco8773, line 129:
```suggestion
    Skw : `~astropy.units.Quantity`
        Computed spectral density function over the input `wavelengths` array
        with units of s/rad.
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+
+    .. _`ISBN 978\\-0123748775`: https://www.sciencedirect.com/book/9780123748775/plasma-scattering-of-electromagnetic-radiation
+    .. _`10.5281/zenodo.3766933`: https://doi.org/10.5281/zenodo.3766933
+
+    """
+
+    # If ion drift velocity is not specified, create an array corresponding
+    # to zero drift
+    if ion_vel is None:
+        ion_vel = np.zeros([fract.size, 3]) * u.cm / u.s
```
rocco8773, line 149:
```suggestion
        ion_vel = np.zeros([fract.size, 3]) * u.m / u.s
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,298 @@
...
+    "### Example Cases in Different Scattering Regimes\n",
+    "\n",
+    "We will now consider several example cases in different scattering regimes. In order to facilitate this, we'll set up each example as a dictionary of plasma parameters:"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A single-species and stationary hydrogen plasma with densities and temperatures that mean scattering is dominated by single-electron scattering (rather than collective multi-particle scattering)"
```
rocco8773, line 116:
```suggestion
    "A single-species, stationary hydrogen plasma with a density and temperature that results in the mean scattering being dominated by scattering off of single electrons."
```

The descriptions for each regime need to be cleaned-up a bit so they're easier to read.

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,298 @@
...
+    "\n",
+    "Thomson scattering can be either non-collective (the scattered spectrum is a linear sum of the light scattered by individual particles) or collective (the scattered spectrum is dominated by scattering off of collective plasma waves). In the collective case, separate electron and ion features appear on either side of the probe laser wavelength corresponding to scattering off of electron and ion plasma oscillations. Depending on which of these regimes applies, fitting the scattered spectrum can provide the electron (and sometimes ion) density and temperature. Doppler shifting of the spectrum can also provide a measurement of the drift velocity of each plasma species.\n",
+    "\n",
+    "For a detailed explanation of the underlying physics (and derivations of these expressions), see \"Plasma Scattering of Electromagnetic Radiation\" by Sheffield et al."
```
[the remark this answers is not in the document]
pheuer: Ok, I've tried to improve the notebook explanatory text and included some of the equations solved. 

I can't get the links to the other documentation pages to work: do those only work once the documentation has been generated?

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,275 @@
...
+        \sum_i \frac{2\pi Z_i}{k}
+        \bigg |\frac{\chi_e}{\epsilon} \bigg |^2 f_{io}
+        \bigg ( \frac{\omega}{k} \bigg )
+
+    where:
+
+    .. math::
+        \epsilon = 1 + \chi_e + \sum_i \chi_i
+
+    In this function the electron and ion distribution functions
```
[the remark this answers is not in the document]
pheuer: Agreed, added

docs/notebooks/thomson.ipynb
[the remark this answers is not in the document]
rocco8773: The links seem to be working on RTD.  Most of the links will work locally without building, except for the `spectral-density` link.  The spectral density `.rst` file is automatically generated when the documentation ins built.

rocco8773:
I haven't gone through the changes in detail yet, but I like what I'm seeing.  Good job.

I'm about to go into a meeting, so I'll try to go over the code one last time after lunch.

If you're happy with your code, can you remove the "draft" mark?  Once that is removed I'll be able to merge the code.

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,278 @@
...
+        of the ion component number density to the total ion number density.
+        Must sum to 1.0. Default is a single ion species.
+
+    ion_species : str or `~plasmapy.particles.Particle`, shape (N, ), optional
+        A list or single instance of `~plasmapy.particles.Particle`, or strings
+        convertible to `~plasmapy.particles.Particle`. Default is `'H+'`
+        corresponding to a single species of hydrogen ions.
+
+    fluid_vel : `~astropy.units.Quantity`, shape (3, ), optional
+        Electron fluid velocity in the rest frame. (convertible to cm/s)
```
rocco8773, line 106:
```suggestion
        Electron fluid velocity in the rest frame. (convertible to m/s)
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,278 @@
...
+        fluid velocit in units convertable to cm/s. Defaults zero drift
+        for all specified ion species.
+
+    probe_vec : float `~numpy.ndarray`, shape (3, )
+        Unit vector in the direction of the probe laser. Defaults to
+        [1, 0, 0].
+
+    scatter_vec : float `~numpy.ndarray`, shape (3, )
+        Unit vector pointing from the scattering volume to the detector.
+        Defaults to [0, 1, 0] which, along with the default probe_vec,
```
rocco8773, line 120:
```suggestion
        Defaults to [0, 1, 0] which, along with the default `probe_vec`,
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,278 @@
...
+
+    # Treatment of multiple species is an extension of the discussion in
+    # Sheffield Sec. 5.1
+    chiI = np.zeros([fract.size, w.size], dtype=np.complex128)
+    for i, ion in enumerate(ion_species):
+        chiI[i, :] = permittivity_1D_Maxwellian(
+            w_i[i, :], k, Ti[i], ni[i], ion, z_mean=ion_z[i]
+        )
+
+    # Calculate the logitudinal dielectric function
```
rocco8773, line 250:
```suggestion
    # Calculate the longitudinal dielectric function
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,278 @@
...
+import astropy.constants as const
+import astropy.units as u
+import numpy as np
+
+from plasmapy.formulary.parameters import plasma_frequency, thermal_speed
+from plasmapy.formulary.dielectric import permittivity_1D_Maxwellian
+from plasmapy.particles import Particle
+from plasmapy.utils.decorators import validate_quantities
+from typing import List, Tuple, Union
+from warnings import warn
```
rocco8773, line 19:
```suggestion
```

`warn` is no longer used

plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -0,0 +1,251 @@
...
+    ne = 5e17 * u.cm ** -3
+    probe_vec = np.array([1, 0, 0])
+    scatter_vec = np.array([0, 1, 0])
+    fract = np.array([0.7, 0.3])
+    Te = 10 * u.eV
+    Ti = np.array([5, 5]) * u.eV
+    fluid_vel = np.array([300, 0, 0]) * u.km / u.s
+    ion_vel = np.array([[-500, 0, 0], [0, 500, 0]]) * u.km / u.s
+
+    # Use this to also test inputing ion species as Particle objects
```
rocco8773, line 67:
```suggestion
    # Use this to also test passing in ion species as Particle objects
```

plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -0,0 +1,251 @@
...
+    # Define some constants
+    wavelengths = np.arange(520, 545, 0.01) * u.nm
+    probe_wavelength = 532 * u.nm
+    ne = 5e17 * u.cm ** -3
+    probe_vec = np.array([1, 0, 0])
+    scatter_vec = np.array([0, 1, 0])
+    fract = np.array([1.0])
+    Te = 10 * u.eV
+    Ti = np.array([10]) * u.eV
+    ion_species = ["C-12 5+"]
```
rocco8773, line 128:
```suggestion
    ion_species = "C-12 5+"
```

This will allow the tests to hit the following if-clause...
https://github.com/PlasmaPy/PlasmaPy/blob/b9552f4a5f1da5178b9100c12824664a53d15c7e/plasmapy/diagnostics/thomson.py#L155-L156

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[thomson]: ../diagnostics/thomson.rst\n",
+    "[spectral-density]: ../api/plasmapy.diagnostics.thomson.spectral_density.rst#spectral-density\n",
+    "[sheffield]: https://www.sciencedirect.com/book/9780123748775/plasma-scattering-of-electromagnetic-radiation\n",
+    "\n",
+    "The [thomson.spectral_density][spectral-density] function calculates the [spectral density function S(k,w)][sheffield], which is one of several terms that determine the scattered power spectrum for the Thomson scattering of a probe laser beam by a plasma. In particular, this function calculates S(k,w) for the case of a plasma consisting of one or more ion species and a neutralizing electron fluid under the assumption that all of the ion species and the electron fluid have Maxwellian velocity distribution functions. In this regime, the spectral density is given by the equation:\n",
```
rocco8773, line 18:
```suggestion
    "The [thomson.spectral_density][spectral-density] function calculates the [spectral density function S(k,w)][sheffield], which is one of several terms that determine the scattered power spectrum for the Thomson scattering of a probe laser beam by a plasma. In particular, this function calculates $S(k,w)$ for the case of a plasma consisting of one or more ion species and a neutralizing electron fluid under the assumption that all of the ion species and the electron fluid have Maxwellian velocity distribution functions. In this regime, the spectral density is given by the equation:\n",
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+    "\n",
+    "\\begin{equation}\n",
+    "S(k,\\omega) = \\frac{2\\pi}{k} \\bigg |1 - \\frac{\\chi_e}{\\epsilon} \\bigg |^2 f_{e0}\\bigg ( \\frac{\\omega}{k} \\bigg ) + \\sum_i \\frac{2\\pi Z_i}{k} \\bigg | \\frac{\\chi_e}{\\epsilon} \\bigg |^2 f_{i0, i} \\bigg ( \\frac{\\omega}{k} \\bigg )\n",
+    "\\end{equation}\n",
+    "\n",
+    "where $\\chi_e$ is the electron component succeptibility of the plasma and $\\epsilon = 1 + \\chi_e + \\sum_i \\chi_i$ is the total plasma dielectric function (with $\\chi_i$ being the ion component of the succeptibility), $Z_i$ is the charge of each ion, $k$ is the scattering wavenumber, $\\omega$ is the scattering frequency, and the functions $f_{e0}$ and $f_{i0,i}$ are the Maxwellian velocity distributions for the electrons and ion species respectively.\n",
```
rocco8773, line 24:
```suggestion
    "where $\\chi_e$ is the electron component susceptibility of the plasma and $\\epsilon = 1 + \\chi_e + \\sum_i \\chi_i$ is the total plasma dielectric function (with $\\chi_i$ being the ion component of the susceptibility), $Z_i$ is the charge of each ion, $k$ is the scattering wavenumber, $\\omega$ is the scattering frequency, and the functions $f_{e0}$ and $f_{i0,i}$ are the Maxwellian velocity distributions for the electrons and ion species respectively.\n",
```

plasmapy/diagnostics/thomson.py
```diff
@@ -0,0 +1,278 @@
...
+    This function calculates the spectral density function for Thomson
+    scattering of a probe laser beam by a plasma consisting of one or more ion
+    species and a neutralizing electron fluid:
+
+    .. math::
+        S(k,\omega) = 1 - \frac{2\pi}{k}
+        \bigg |1 - \frac{\chi_e}{\epsilon} \bigg |^2
+        f_{e0} \bigg (\frac{\omega}{k} \bigg ) +
+        \sum_i \frac{2\pi Z_i}{k}
+        \bigg |\frac{\chi_e}{\epsilon} \bigg |^2 f_{io}
```
rocco8773, line 61:
```suggestion
        \bigg |\frac{\chi_e}{\epsilon} \bigg |^2 f_{i0,i}
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "non_collective = {'name':'Non-Collective Regime', 'ne':5e15*u.cm**-3, 'Te':40*u.eV, 'Ti':np.array([10])*u.eV,\n",
+    "                  'fract':np.array([1]), 'ion_species':['H+'],\n",
+    "                  'fluid_vel':np.array([0, 0, 0])*u.km/u.s, 'ion_vel':np.array([[0, 0, 0]])*u.km/u.s}\n"
```
rocco8773, line 145:
```suggestion
    "non_collective = {\n",
    "    'name':'Non-Collective Regime', \n",
    "    'ne':5e15*u.cm**-3, \n",
    "    'Te':40*u.eV, \n",
    "    'Ti':np.array([10])*u.eV,\n",
    "    'fract':np.array([1]), \n",
    "    'ion_species':['H+'],\n",
    "    'fluid_vel':np.array([0, 0, 0])*u.km/u.s, \n",
    "    'ion_vel':np.array([[0, 0, 0]])*u.km/u.s,\n",
    "}"
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "collective = {'name':'Collective Regime', 'ne':5e17*u.cm**-3, 'Te':10*u.eV, 'Ti':np.array([4])*u.eV,\n",
+    "              'fract':np.array([1]), 'ion_species':['H+'],\n",
+    "              'fluid_vel':np.array([0, 0, 0])*u.km/u.s, 'ion_vel':np.array([[0, 0, 0]])*u.km/u.s}"
```
rocco8773, line 181:
```suggestion
    "collective = {\n",
    "    'name':'Collective Regime',\n",
    "    'ne':5e17*u.cm**-3,\n",
    "    'Te':10*u.eV,\n",
    "    'Ti':np.array([4])*u.eV,\n",
    "    'fract':np.array([1]),\n",
    "    'ion_species':['H+'],\n",
    "    'fluid_vel':np.array([0, 0, 0])*u.km/u.s,\n",
    "    'ion_vel':np.array([[0, 0, 0]])*u.km/u.s,\n",
    "}"
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "drifts = {'name':'Drift Velocities', 'ne':5e17*u.cm**-3, 'Te':10*u.eV, 'Ti':np.array([10])*u.eV,\n",
+    "          'fract':np.array([1]), 'ion_species':['H+'],\n",
+    "          'fluid_vel':np.array([700, 0, 0])*u.km/u.s, 'ion_vel':np.array([[-600, -100, 0]])*u.km/u.s}"
```
rocco8773, line 199:
```suggestion
    "drifts = {\n",
    "    'name':'Drift Velocities',\n",
    "    'ne':5e17*u.cm**-3,\n",
    "    'Te':10*u.eV,\n",
    "    'Ti':np.array([10])*u.eV,\n",
    "    'fract':np.array([1]),\n",
    "    'ion_species':['H+'],\n",
    "    'fluid_vel':np.array([700, 0, 0])*u.km/u.s,\n",
    "    'ion_vel':np.array([[-600, -100, 0]])*u.km/u.s,\n",
    "}"
```

docs/notebooks/thomson.ipynb
```diff
@@ -0,0 +1,309 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "two_species = {'name':'Two Ion Species', 'ne':5e17*u.cm**-3, 'Te':10*u.eV, 'Ti':np.array([10, 50])*u.eV,\n",
+    "               'fract':np.array([.25, .75]), 'ion_species':['He-4 1+', 'C-12 5+'],\n",
+    "               'fluid_vel':np.array([0, 0, 0])*u.km/u.s, 'ion_vel':np.array([[0, 0, 0],[0, 0, 0]])*u.km/u.s}"
```
rocco8773, line 217:
```suggestion
    "two_species = {\n",
    "    'name':'Two Ion Species',\n",
    "    'ne':5e17*u.cm**-3,\n",
    "    'Te':10*u.eV,\n",
    "    'Ti':np.array([10, 50])*u.eV,\n",
    "    'fract':np.array([.25, .75]),\n",
    "    'ion_species':['He-4 1+', 'C-12 5+'],\n",
    "    'fluid_vel':np.array([0, 0, 0])*u.km/u.s,\n",
    "    'ion_vel':np.array([[0, 0, 0],[0, 0, 0]])*u.km/u.s,\n",
    "}"
```

pheuer:
@rocco8773 Great, thank you for catching the spelling/formatting errors!

plasmapy/diagnostics/thomson.py
rocco8773: @pheuer Do you have a reason for not adding this suggestion, or did it just get lost in the noise?

pheuer: @rocco8773  Oops no, just got lost amongst all the other commits

StanczakDominik:
Woo! Thanks @pheuer for this PR, and thanks @rocco8773 for carrying the review :D

namurphy:
Yes, thank you!  This is really exciting!

pheuer:
Thank you all for your help! Looking forward to adding more in the future.

Issue #759, opened 2020-02-08
Serializable `Particle`s


StanczakDominik:
I'd like to be able to put (some form of) `Particle`s in a dict and dump it to JSON, or put them in an `xarray` attributes dictionary, save an xarray, load it from file and (possibly with some convenient additional steps to handle any necessary conversion) get the particle back.

A basic test-ish use case:
```python
import xarray
array = xarray.DataSet()
array.attrs['particle'] = Particle('tritium')
array.to_zarr("~/storage")       # without crashing
# store for 12.3 years
array2 = xarray.read_zarr("~/storage")
# optional stuff happens here
assert array2.attrs['particle'] == array.attrs['particle']   # not just a string, but a Particle!
```

I have no idea how best to achieve this, though.

namurphy:
I had to look up what [serializable](https://en.wikipedia.org/wiki/Serialization) means in this context.  Oh, the joys of not being a formally trained computer scientist or software engineer...  :joy_cat:

StanczakDominik:
Glad I could help you pick that up, then! :D I actually found a way to do that - I just put str(p) in attrs, and when I need the particle, I use Particle on that!

Of course, it's going to be more difficult once we have CustomParticles. Maybe a from_str factory method on a base class that assigns either a Particle or CustomParticle?

diego7319:
We could use the library "pickle" to save it as a binary file, the downside would be if in the future the class particle change, the file would not be useful. Also it can't be readed as a json file because it s a  binary file. I guess that a function to_json and from_json would be useful in this case.

````
import pickle

f = open('store.pckl', 'wb')
pickle.dump(obj, f)
f.close()

f = open('store.pckl', 'rb')
obj = pickle.load(f)
f.close()

StanczakDominik:
I'd definitely prefer the `to_json` approach.

DarkAEther:
Hey @StanczakDominik, if this issue is still up for grabs, I'd like to have a go at trying to create a function that will produce a JSON representation of a Particle object. I'm a total newbie to Open Source contributions. Any advice and/or suggestions are appreciated.

StanczakDominik:
It absolutely is! Go for it! One piece of advice - open up a PR with your experiments early, so suggestions can be made early on :)

DarkAEther:
[...]

PR #832, opened 2020-06-04
Issue #759 Experiment 1 - Representing Particles in JSON

DarkAEther:
Contains a basic implementation of the `get_json` method for objects of the Particle class that will return a dictionary representation of a Particle.

In addition, this includes a basic set of examples in the docstring of the method as well as a pytest module that tests some basic cases as a proof of concept.

- [ ] I have added a changelog entry for this pull request (please see
  they're failing, ask!).

Any suggestions and/or advice is appreciated. Thank you.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -466,6 +466,79 @@ class Particle(AbstractParticle):
         """Return the particle's symbol."""
         return self.particle
 
+    def get_json(self) -> dict:
+        """
+        Returns a dictionary object
+        """
+        particle_json = dict()
+        particle_json["symbol"] = self.particle
+        try:
+            particle_json["antiparticle"] = self.antiparticle
+        except AtomicError:
+            # Does not have an antiparticle
+            pass
+        particle_json["element"] = self.element
+        particle_json["isotope"] = self.isotope
+        particle_json["ionic_symbol"] = self.ionic_symbol
+        try:
+            particle_json["roman_symbol"] = self.roman_symbol
+        except (roman.OutOfRangeError, ChargeError):
+            # negative charge or no charge specified
+            pass
+        try:
+            particle_json["element_name"] = self.element_name
+        except InvalidElementError:
+            # not an element
+            pass
+        # TODO: isotope name
+        try:
+            particle_json["integer_charge"] = self.integer_charge
+        except ChargeError:
+            # Unknown Charge
+            pass
+        # TODO: charge in u.Quantity
+        # TODO: st. atomic weight in u.Quantity
+        # TODO: nuclide_mass in u.Quantity
+        # TODO: mass in u.Quantity
+        # TODO: mass_energy in u.Quantity
+        # TODO: binding_energy in u.Quantity
+        try:
+            particle_json["atomic_number"] = self.atomic_number
+        except InvalidElementError:
+            # Not an element
+            pass
+        try:
+            particle_json["mass_number"] = self.mass_number
+        except InvalidIsotopeError:
+            # not an isotope
+            pass
+        # TODO: Neutron_number
+        # TODO: Electron_number
+        # TODO: Isotopic abundance
+        # TODO: baryon number
+        # TODO: lepton number
+        # TODO: half_life
+        try:
+            particle_json["spin"] = self.spin
+        except MissingAtomicDataError:
+            # no data for spin
+            pass
+        try:
+            particle_json["periodic_table_entry"] = dict(self.periodic_table._asdict())
+        except InvalidElementError:
+            # not an element
+            pass
+        particle_json["categories"] = self.categories
+        # TODO: is_electron
+        # TODO: is_ion
+        return particle_json
+
+    def get_json_string(self) -> str:
+        """
+        Returns a stringified JSON representation of a Particle
+        """
+        return str(self.get_json())
+
     def __eq__(self, other) -> bool:
         """
         Determine if two objects correspond to the same particle.
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -468,7 +468,38 @@ class Particle(AbstractParticle):
 
     def get_json(self) -> dict:
         """
-        Returns a dictionary object
+        Returns a dictionary object representing the
+        Particle's data
+
+        Examples
+        --------
+
+        >>> import pprint
+        >>> lead = Particle('lead')
+        >>> pprint.pprint(lead.get_json()) #pretty-printed for ease of viewing
+        {'atomic_number': 82,
+         'categories': {'post-transition metal', 'element'},
+         'element': 'Pb',
+         'element_name': 'lead',
+         'ionic_symbol': None,
+         'isotope': None,
+         'periodic_table_entry': {'block': 'p',
+                                 'category': 'post-transition metal',
+                                 'group': 14,
+                                 'period': 6},
+         'symbol': 'Pb'}
+        >>> electron = Particle('e-')
+        >>> pprint.pprint(electron.get_json()) #pretty-printed for ease of viewing
+        {'antiparticle': Particle("e+"),
+         'categories': {'lepton', 'fermion', 'matter', 'electron', 'charged', 'stable'},
+         'element': None,
+         'integer_charge': -1,
+         'ionic_symbol': None,
+         'isotope': None,
+         'roman_symbol': None,
+         'spin': 0.5,
+         'symbol': 'e-'}
+
         """
         particle_json = dict()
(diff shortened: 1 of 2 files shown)


DarkAEther:
@StanczakDominik Could I get some assistance with the Azure pipelines? I'm not too sure why they are failing.

In addition, the test checks for an entry in the changelog, however, I was under the impression that the entry would be made after the PR was merged. May I know at what stage the entry should be made?

I have fixed the coding style errors. I'd like to apologise for that. It was an oversight on my part.

StanczakDominik:
> @StanczakDominik Could I get some assistance with the Azure pipelines? I'm not too sure why they are failing.

Well, looking at https://dev.azure.com/plasmapy/PlasmaPy/_build/results?buildId=986&view=logs&j=7eff0d3e-d379-542c-e552-ec7f38b615ef&t=0c2c247d-9f40-506b-745d-314bd76fc96b&l=5528 and comparing the expected and actual output:

```python
    - 'categories': {'lepton', 'fermion', 'matter', 'electron', 'charged', 'stable'},
    + 'categories': {'electron', 'lepton', 'fermion', 'charged', 'matter', 'stable'},
```

`categories` is a `set`, and those aren't guaranteed to be ordered! So this docstring will not reproduce. I think you could dodge that by replacing the set contents with an ellipsis, `{...}`, but that may ultimately not be necessary - more on that in a few...

> In addition, the test checks for an entry in the changelog, however, I was under the impression that the entry would be made after the PR was merged. May I know at what stage the entry should be made?

Yeah, I generally submit a PR, get the PR number, then add another commit with the changelog. It turns out to be easier to maintain the package that way, rather than adding changelogs post-factum (because we tend to forget to do that!). I now realize that the instructions aren't really clear on this first part.

> I have fixed the coding style errors. I'd like to apologise for that. It was an oversight on my part.

No worries! We are all united as a community of people being shouted at by autopep8 :laughing:

plasmapy/particles/particle_class.py
```diff
@@ -466,6 +466,110 @@ def __str__(self) -> str:
...
+        >>> pprint.pprint(electron.get_json()) #pretty-printed for ease of viewing
+        {'antiparticle': Particle("e+"),
+         'categories': {'lepton', 'fermion', 'matter', 'electron', 'charged', 'stable'},
+         'element': None,
+         'integer_charge': -1,
+         'ionic_symbol': None,
+         'isotope': None,
+         'roman_symbol': None,
+         'spin': 0.5,
+         'symbol': 'e-'}
```
StanczakDominik, line 501: Perhaps we don't need all of this information here? What I had in mind for the issue was something like this workflow:

1. I need to store simulation data that contains information on the `particle`s in that simulation
2. For each particle type, I dump its data to some machine-readable form of text, that could be a str(dictionary) (JSON);
3. I dump the stringified particle data alongside my simulation output, say, in an [`attrs` dictionary of an xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.attrs.html)
4. I load the simulation output from disk, grabbing the string data
5. I turn it back into a dict
6. I use some function to turn that dict back into a `Particle`

So the `dict` does not actually need to contain all the information a `Particle` carries with it; it just needs as much as is necessary to restore the original `Particle` object.

Looking at http://docs.plasmapy.org/en/latest/particles/index.html#classes, this means that for `Particle`s we'd probably just need the `symbol`. It gets a bit trickier with `DimensionlessParticle`s, which will need to store their two numbers, and a bit more trickier with `CustomParticle`s, which will also need their units [(shouldn't be too complex if using existing Astropy functionality)](https://docs.astropy.org/en/stable/units/format.html#string-representations-of-units).

So I think a good direction here would be to make `get_json()` as minimal as possible, then try to implement `parse_json()` that would create the appropriate Particle class from the string produced by `get_json()`. It would also make the docstrings cleaner and testing much simpler.

What do you think?

I'm also sorry I probably misspecified the original issue! Stuff feels obvious when you're typing it down into the void.

DarkAEther:
> Well, looking at https://dev.azure.com/plasmapy/PlasmaPy/_build/results?buildId=986&view=logs&j=7eff0d3e-d379-542c-e552-ec7f38b615ef&t=0c2c247d-9f40-506b-745d-314bd76fc96b&l=5528 and comparing the expected and actual output:

    - 'categories': {'lepton', 'fermion', 'matter', 'electron', 'charged', 'stable'},
    + 'categories': {'electron', 'lepton', 'fermion', 'charged', 'matter', 'stable'},

Ah, I see. Thank you for mentioning the link. I realised I didn't notice it at all while looking at the error list. Whoops xD.

> categories is a set, and those aren't guaranteed to be ordered! So this docstring will not reproduce. I think you could dodge that by replacing the set contents with an ellipsis, {...}, but that may ultimately not be necessary - more on that in a few...

Oh wow, total noob mistake. Did not realise that at all. What do you think about an approach that converts the set to an alphabetically sorted list for the JSON? Something like this:
`'categories': ['charged', 'electron', 'fermion', 'lepton', 'matter', 'stable']`
This could potentially solve the order of the items problem. However, I just saw your mention about the requirements of the `get_json` method. I'll follow up on it in a response to the requested change. 

> Yeah, I generally submit a PR, get the PR number, then add another commit with the changelog. It turns out to be easier to maintain the package that way, rather than adding changelogs post-factum (because we tend to forget to do that!). I now realize that the instructions aren't really clear on this first part.

Thanks for the info. I'll be sure to follow the same :)

> No worries! We are all united as a community of people being shouted at by autopep8

That's true. It is my first time working with pep8 standards. I tried out the pycodestyle tool. Color me impressed. Didn't expect to get a shouting in my inbox though. That was fun xD.

plasmapy/particles/particle_class.py
DarkAEther: > So the dict does not actually need to contain all the information a Particle carries with it; it just needs as much as is necessary to restore the original Particle object.

> Looking at http://docs.plasmapy.org/en/latest/particles/index.html#classes, this means that for Particles we'd probably just need the symbol. It gets a bit trickier with DimensionlessParticles, which will need to store their two numbers, and a bit more trickier with CustomParticles, which will also need their units (shouldn't be too complex if using existing Astropy functionality).

Now that I understand the requirement, this clears up a lot of doubts xD. I agree that since the purpose is to store the particles and recover them later, storing the bare minimum is definitely a good way forward as opposed to dumping the entire data regarding the particle. 
Will this report be read only by a machine? If so, then we can probably work only with the bare minimum, however, if there may be a human who will peer at the report at some point, it might be a good idea to have a few additional bits of information included in the JSON (though we can only use the bare minimum to transform the dict back into its respective object).

> I turn it back into a dict

There is a library that does this : https://docs.python.org/3/library/ast.html#ast.literal_eval 
This will come at a cost of having an additional dependency which is not always a good thing. Moreover, parsing the string to a dictionary manually is fairly straightforward.

> I'm also sorry I probably misspecified the original issue! Stuff feels obvious when you're typing it down into the void.

No problem at all :)

rocco8773:
I'm no JSON expert, and my experience with it is limited.  However, I think looking into Python's [`json` module](https://docs.python.org/3/library/json.html#module-json) would be useful before getting too far into this.  Specifically the [`JSONDecoder`](https://docs.python.org/3/library/json.html#json.JSONDecoder) and [`JSONEncoder`](https://docs.python.org/3/library/json.html#json.JSONEncoder) classes might be useful here.

DarkAEther:
@rocco8773 Thank you for the suggestion. I'll definitely have a look at it :)

StanczakDominik:
Yeah, json.dumps and .loads will be helpful here!

DarkAEther:
> Yeah, json.dumps and .loads will be helpful here!

I just had a look at these. I think they are exactly what we need. Thanks for the suggestion.

I had a quick question. If I want to change my code and make a PR, do I close this one and open a new PR or is there a way to update the PR with the new commits?

rocco8773:
> I had a quick question. If I want to change my code and make a PR, do I close this one and open a new PR or is there a way to update the PR with the new commits?

That's up to you.  Typically I keep everything in the same PR, unless there's a drastic shift in direction.  You are kinda at that inflection point now, but either option is suitable.  If you do create a new PR, then at least mention this PR in the initial comment of the new PR.

DarkAEther:
Thanks for the response! I feel that the direction has changed from the idea I had in mind of a full particle export to a more machine oriented representation, hence I think I will open a new PR for the next set of commits. I will definitely reference this PR in the comment of the new one.

DarkAEther:
The actual approach used to solve the requirement diverged from the solution used here.
Please refer to PRs #833 and #836 for the follow up solutions.
Closing PR.

PR #833, opened 2020-06-04
Issue #759 Serialize particles - Experiment 2

DarkAEther:
Addressing Issue #759
Referencing previous PR #832 

This PR includes implementation of the `to_json` method for objects of the Particle, CustomParticle and DimensionlessParticle. This method produces a JSON string representation of the respective particle keeping in mind the stated requirements in PR #832 .

Additionally, it includes a simple pytest module for the feature. Methods contain docstrings with examples. 

I might get scolded again by autopep8 due to some of the docstrings exceeding the character limit xD.

- [ ] I have added a changelog entry for this pull request (please see
- [ ] I have fixed any new failing tests (if you're unsure why
  they're failing, ask!).

Any suggestions and/or advice are appreciated.

Thank you.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -111,9 +111,17 @@ class AbstractParticle(ABC):
     def charge(self) -> Union[u.Quantity, Real]:
         raise NotImplementedError
 
-    @abstractmethod
-    def to_dict(self) -> dict:
-        raise NotImplementedError
+    @property
+    def particle_dict(self) -> dict:
+        particle_dictionary = {
+            "plasmapy_particle": {
+                "type": type(self).__name__,
+                "description": {
+                    # subclass specific description
+                }
+            }
+        }
+        return particle_dictionary
 
     def __bool__(self):
         """
@@ -126,7 +134,7 @@ class AbstractParticle(ABC):
         """
         Return a JSON string representation of the minimum description of a particle.
         """
-        return json.dumps(self.to_dict())
+        return json.dumps(self.particle_dict())
 
 
 class Particle(AbstractParticle):
@@ -476,24 +484,25 @@ class Particle(AbstractParticle):
         """Return the particle's symbol."""
         return self.particle
 
-    def to_dict(self) -> dict:
+    def particle_dict(self) -> dict:
         """
         Return a dictionary representation of the minimum description of a particle.
 
         Examples
         --------
         >>> lead = Particle('lead')
-        >>> lead.to_dict()
-        {'type': 'Particle', 'symbol': 'Pb'}
+        >>> lead.particle_dict()
+        {'plasmapy_particle': {'type': 'Particle', 'description': {'argument': 'Pb'}}}
         >>> electron = Particle('e-')
-        >>> electron.to_dict()
-        {'type': 'Particle', 'symbol': 'e-'}
+        >>> electron.particle_dict()
+        {'plasmapy_particle': {'type': 'Particle', 'description': {'argument': 'e-'}}}
 
         """
-        particle_dict = dict()
-        particle_dict['type'] = 'Particle'
-        particle_dict['symbol'] = self.particle
-        return particle_dict
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "argument": self.particle
+        }
+        return particle_dictionary
 
     def __eq__(self, other) -> bool:
         """
@@ -1714,7 +1723,7 @@ class DimensionlessParticle(AbstractParticle):
         """
         return f"DimensionlessParticle(mass={self.mass}, charge={self.charge})"
 
-    def to_dict(self) -> dict:
+    def particle_dict(self) -> dict:
         """
         Return a dictionary representation of the minimum description of a dimensionless particle.
 
@@ -1722,17 +1731,18 @@ class DimensionlessParticle(AbstractParticle):
         --------
         >>> from plasmapy.particles import DimensionlessParticle
         >>> dimensionless_particle = DimensionlessParticle(mass=1.0, charge=-1.0)
-        >>> dimensionless_particle.to_dict()
-        {'type': 'DimensionlessParticle', 'mass': '1.0', 'charge': '-1.0'}
+        >>> dimensionless_particle.particle_dict()
+        {'plasmapy_particle': {'type': 'DimensionlessParticle', 'description': {'mass': '1.0', 'charge': '-1.0'}}}
         >>> dimensionless_particle = DimensionlessParticle(mass=1.0)
-        >>> dimensionless_particle.to_dict()
-        {'type': 'DimensionlessParticle', 'mass': '1.0', 'charge': 'nan'}
+        >>> dimensionless_particle.particle_dict()
+        {'plasmapy_particle': {'type': 'DimensionlessParticle', 'description': {'mass': '1.0', 'charge': 'nan'}}}
         """
-        particle_dict = dict()
-        particle_dict['type'] = 'DimensionlessParticle'
-        particle_dict['mass'] = str(self.mass)
-        particle_dict['charge'] = str(self.charge)
-        return particle_dict
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "mass": str(self.mass),
+            "charge": str(self.charge)
+        }
+        return particle_dictionary
 
     @staticmethod
     def _validate_parameter(obj, can_be_negative=True) -> np.float64:
@@ -1864,25 +1874,26 @@ class CustomParticle(AbstractParticle):
         """
         return f"CustomParticle(mass={self.mass}, charge={self.charge})"
 
-    def to_dict(self) -> dict:
+    def particle_dict(self) -> dict:
         """
         Return a dictionary representation of the minimum description of a CustomParticle.
 
         Examples
         --------
-        >>> custom_particle = CustomParticle(mass=1.5e-26 * u.kg, charge=-1)
-        >>> custom_particle.to_dict()
-        {'type': 'CustomParticle', 'mass': '1.5e-26 kg', 'charge': '-1.6021...e-19 C'}
+        >>> custom_particle = CustomParticle(mass=5.12 * u.kg, charge=6.2 * u.C)
+        >>> custom_particle.particle_dict()
+        {'plasmapy_particle': {'type': 'CustomParticle', 'description': {'mass': '5.12 kg', 'charge': '6.2 C'}}}
         >>> custom_particle = CustomParticle(mass=1.5e-26 * u.kg)
-        >>> custom_particle.to_dict()
-        {'type': 'CustomParticle', 'mass': '1.5e-26 kg', 'charge': 'nan C'}
+        >>> custom_particle.particle_dict()
+        {'plasmapy_particle': {'type': 'CustomParticle', 'description': {'mass': '1.5e-26 kg', 'charge': 'nan C'}}}
 
         """
-        particle_dict = dict()
-        particle_dict['type'] = 'CustomParticle'
-        particle_dict['mass'] = str(self.mass)
-        particle_dict['charge'] = str(self.charge)
-        return particle_dict
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "mass": str(self.mass),
+            "charge": str(self.charge)
+        }
+        return particle_dictionary
 
     @property
     def mass(self) -> u.kg:
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -11,7 +11,7 @@ from typing import List, Optional, Set, Tuple, Union
 import astropy.constants as const
 import astropy.units as u
 import numpy as np
-
+import json
 import plasmapy.utils.roman as roman
 from plasmapy.particles.elements import _Elements, _PeriodicTable
 from plasmapy.particles.exceptions import (
@@ -117,6 +117,12 @@ class AbstractParticle(ABC):
         do not have a truth value.
         """
         raise AtomicError("The truth value of a particle is not defined.")
+
+    def to_json(self, dict_representation) -> str:
+        """
+        Return a JSON string representation of the minimum description of a particle.
+        """
+        return json.dumps(dict_representation)
 
 
 class Particle(AbstractParticle):
@@ -465,6 +471,41 @@ class Particle(AbstractParticle):
     def __str__(self) -> str:
         """Return the particle's symbol."""
         return self.particle
+
+    def to_dict(self) -> dict:
+        """
+        Return a dictionary representation of the minimum description of a particle.
+
+        Examples
+        --------
+        >>> lead = Particle('lead')
+        >>> lead.to_dict()
+        {'type': 'Particle', 'symbol': 'Pb'}
+        >>> electron = Particle('e-')
+        >>> electron.to_dict()
+        {'type': 'Particle', 'symbol': 'e-'}
+
+        """
+        particle_dict = dict()
+        particle_dict['type'] = 'Particle'
+        particle_dict['symbol'] = self.particle
+        return particle_dict
+
+    def to_json(self) -> str:
+        """
+        Return a JSON string representation of the minimum description of a particle.
+
+        Examples
+        --------
+        >>> lead = Particle('lead')
+        >>> lead.to_json()
+        '{"type": "Particle", "symbol": "Pb"}'
+        >>> electron = Particle('e-')
+        >>> electron.to_json()
+        '{"type": "Particle", "symbol": "e-"}'
+
+        """
+        return super().to_json(self.to_dict())
 
     def __eq__(self, other) -> bool:
         """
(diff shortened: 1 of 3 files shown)


DarkAEther:
More green than red! Definitely a good sign xD.

I noticed that the error is a mismatch in the floating value of the charge.
https://github.com/PlasmaPy/PlasmaPy/pull/833/checks?check_run_id=740174437

Is it possible that this may change depending on the system and platform being used? I have heard somewhere that floating points are not represented the same way in all machines and platforms. Any suggestions on how we can potentially achieve value consistency across devices?

Another query was the docstring examples - autopep8 mentioned that the maximum limit was exceeded on the docstring lines. Would moving the remainder of the string to the next line be an acceptable approach to rectifying this problem?

StanczakDominik:
> More green than red! Definitely a good sign xD.

Yeah, always happy to see that :D 

> I noticed that the error is a mismatch in the floating value of the charge.
> [#833 (checks)](https://github.com/PlasmaPy/PlasmaPy/pull/833/checks?check_run_id=740174437)
> 
> Is it possible that this may change depending on the system and platform being used? I have heard somewhere that floating points are not represented the same way in all machines and platforms. Any suggestions on how we can potentially achieve value consistency across devices?

Good intuition! But I've seen this bug before (and should definitely have documented it) - it's actually a case of new Astropy versions having newer and more accurate values of physical constants (most likely, here - the mass of a proton and electron)! 

Take a look at how doctests in `parameters` use ellipsis to handle this and skip checking in significant digits:

https://github.com/PlasmaPy/PlasmaPy/blob/c03d8c999c35ca55adce8ac577a668f8ae8870da/plasmapy/formulary/parameters.py#L108-L114

<details>
<summary>If you care about how to set up pytest to do this kind of skipping:</summary>

https://github.com/PlasmaPy/PlasmaPy/blob/c03d8c999c35ca55adce8ac577a668f8ae8870da/setup.cfg#L76-L79

</details>

Also, thanks for pointing out that interface, it's cooler than going through the logs and scrolling past all the warnings as I usually do :D

> Another query was the docstring examples - autopep8 mentioned that the maximum limit was exceeded on the docstring lines. Would moving the remainder of the string to the next line be an acceptable approach to rectifying this problem?

Well, it's the outputs, and since they're most likely compared as strings, it'll be hard to tell pytest that multiline versions of the same string are the same versions as the same string on a single line. If you can find any way to do that, sure, that'd be very nice to pick up - but I'm completely fine with leaving it be.

plasmapy/particles/particle_class.py
```diff
@@ -118,6 +118,12 @@ def __bool__(self):
         """
         raise AtomicError("The truth value of a particle is not defined.")
 
+    def to_json(self, dict_representation) -> str:
+        """
+        Return a JSON string representation of the minimum description of a particle.
+        """
+        return json.dumps(dict_representation)
```
StanczakDominik, line 125: I think there's some rule in programming that I'm forgetting the name of that says if a subclass implements the same method as a parent class, it should have the same calling signature. I think it'd be nice to make this have no arguments and call `self.to_dict()`, which could `raise NotImplementedError` in the base class. That could also be an `@abstractmethod`, so that if you subclass it, you have to remember to implement `to_dict`.

plasmapy/particles/particle_class.py
```diff
@@ -466,6 +472,41 @@ def __str__(self) -> str:
...
+        --------
+        >>> lead = Particle('lead')
+        >>> lead.to_json()
+        '{"type": "Particle", "symbol": "Pb"}'
+        >>> electron = Particle('e-')
+        >>> electron.to_json()
+        '{"type": "Particle", "symbol": "e-"}'
+
+        """
+        return super().to_json(self.to_dict())
```
StanczakDominik, line 508: Then (see above) you wouldn't need to implement this `to_json` method in each subclass - you just create `to_dict`.

plasmapy/particles/particle_class.py
```diff
@@ -1815,6 +1892,41 @@ def __repr__(self):
...
+        {'type': 'CustomParticle', 'mass': {'quantity': 1.5e-26, 'unit': 'kg'}, 'charge': {'quantity': -1.602176634e-19, 'unit': 'C'}}
+        >>> custom_particle = CustomParticle(mass=1.5e-26 * u.kg)
+        >>> custom_particle.to_dict()
+        {'type': 'CustomParticle', 'mass': {'quantity': 1.5e-26, 'unit': 'kg'}, 'charge': {'quantity': nan, 'unit': 'C'}}
+
+        """
+        particle_dict = dict()
+        particle_dict['type'] = 'CustomParticle'
+        particle_dict['mass'] = {"quantity": self.mass.value, "unit": str(self.mass.unit)}
+        particle_dict['charge'] = {"quantity": self.charge.value, "unit": str(self.charge.unit)}
```
StanczakDominik, line 1912: Looks like this could be a bit simpler:

```python
[ins] In [1]: u                                                                                                                           
Out[1]: <module 'astropy.units' from '/usr/lib/python3.8/site-packages/astropy/units/__init__.py'>

[ins] In [2]: 2 * u.kg                                                                                                                    
Out[2]: <Quantity 2. kg>

[ins] In [3]: mass = 2 * u.kg                                                                                                             

[ins] In [4]: str(mass)                                                                                                                   
Out[4]: '2.0 kg'

[ins] In [5]: u.Quantity(str(mass))                                                                                                       
Out[5]: <Quantity 2. kg>

[ins] In [6]: _.unit                                                                                                                      
Out[6]: Unit("kg")

[ins] In [7]: __.value                                                                                                                    
Out[7]: 2.0
```

But I suppose it's fine the way it is, as well! As long as we're consistent. I just want to make sure that we don't end up introducing custom conventions separate from Astropy's.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -997,3 +997,36 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+    ),
+]
+
+
+@pytest.mark.parametrize("cls, kwargs, expected_repr", particle_json_repr_table)
+def test_customized_particle_repr(cls, kwargs, expected_repr):
+    """Test the JSON representations of normal, dimensionless and custom particles."""
+    instance = cls(**kwargs)
+    json_repr = instance.to_json()
+    if not expected_repr == json_repr:
```
StanczakDominik, line 1026: I think the standard pytest idiom would be `assert expected_repr == json_repr, "optional_description:`.

plasmapy/particles/particle_class.py
DarkAEther: Thank you for the pointer. I'll make the necessary changes for it.

plasmapy/particles/particle_class.py
DarkAEther: Yeah, that definitely helps. It will also improve testing since we don't need to add additional (and unnecessary) code in the class. Thanks for the tip :)

plasmapy/particles/particle_class.py
DarkAEther: > ```python
> [ins] In [5]: u.Quantity(str(mass))                                                                                                      
> Out[5]: <Quantity 2. kg> ```

Thank you for suggesting this. I was under the impression that we'd need to manually reform the quantity by doing `magnitude * u.Quantity`. The method you have suggested is definitely easier to code as well as maintain in future, so I shall adopt that into my code.

As an added bonus, this is also likely to remove the pep8 warnings about the extra long lines xD.

plasmapy/particles/tests/test_particle_class.py
DarkAEther: Ah I see! Thanks for the tip. I'll make that change :)

DarkAEther:
> Take a look at how doctests in parameters use ellipsis to handle this and skip checking in significant digits:

Thanks for the hint. I'll incorporate the changes into the pytest method and the docstrings.

DarkAEther:
Awesome! The build and tests are passing this time round :)
I couldn't get much info about the the codecov/patch check fail. May I know what it checks for?

StanczakDominik:
It tests which of the lines affected by the PR are being tested. In this case, it's the `NotImplementedError` in the abstract class - since the PR suddenly has few lines, that one line suddenly starts mattering - according to that metric, at least. I'm fine with ignoring it. :)

Stick a changelog entry on it and I'd say this is good :)

@namurphy, for a second opinion - whatcha think about these changes?

DarkAEther:
> Stick a changelog entry on it and I'd say this is good :)

Awesome! Should I open a separate PR for the conversion from JSON to its respective particle?

StanczakDominik:
As you prefer :) While we're waiting on Nick's review, it would probably make sense to start here. It'd also make the tests are lot more confidence-inspiring :laughing:

StanczakDominik:
But I've been known to make overlarge PRs, so I'm probably not the best person to ask for advice on that :laughing:

DarkAEther:
> But I've been known to make overlarge PRs, so I'm probably not the best person to ask for advice on that laughing

I think I'd prefer to make a separate PR for it. In future if anyone needs to modify either the `to_json` or the incoming `from_json`, they can refer to it a bit easier with 2 different PRs ... hopefully xD.

DarkAEther:
The new set of commits are in reference to the excellent suggestions made by @rocco8773 in PR #836 to improve the approach taken to solve the problem.

Any further suggestions are appreciated :)
Thank you!

StanczakDominik:
It might take me a while to get back to this, I'll have to focus mostly on work for a couple of days.

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +111,31 @@ def mass(self) -> Union[u.Quantity, Real]:
...
+    @property
+    def particle_dict(self) -> dict:
+        particle_dictionary = {
+            "plasmapy_particle": {
+                "type": type(self).__name__,
+                "description": {
+                    # subclass specific description
+                }
+            }
+        }
```
rocco8773, line 123:
```suggestion
        particle_dictionary = {
            "plasmapy_particle": {
                "type": type(self).__name__,
                "module": self.__module__,
                "mass": self.mass.__str__(),
                "charge": self.charge.__str__(),
            }
        }
```

Since `"description"` is not universal among all particle classes and is not used in reconstructing the object, I would leave it out.   Every particle will have `mass` and `charge` so this should be here by default.  `module` is for origin tracing.  You could also add things similar to `created_date`.  Basically, this should only contain data that helps trace the source/generation of the json file/particle and data needed to reconstruct the particle.

plasmapy/particles/particle_class.py
```diff
@@ -1685,6 +1723,27 @@ def __repr__(self):
...
+        >>> dimensionless_particle.particle_dict()
+        {'plasmapy_particle': {'type': 'DimensionlessParticle', 'description': {'mass': '1.0', 'charge': 'nan'}}}
+        """
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "mass": str(self.mass),
+            "charge": str(self.charge)
+        }
+        return particle_dictionary
+
```
rocco8773, line 1746:
```suggestion
```

This is all part of the default behavior for a particle class, so it should be apart of `AbstractParticle` and not required to be overridden by its sub-classes.  Sub-classes should only override this if they are doing something abnormal, like the `Particle` class.  `Particle` has a different `__init__` structure than all the other sub-classes, that's why it's the alternative case in my [example](https://github.com/PlasmaPy/PlasmaPy/pull/836#issuecomment-640249264).

plasmapy/particles/particle_class.py
```diff
@@ -1815,6 +1874,27 @@ def __repr__(self):
...
+        {'plasmapy_particle': {'type': 'CustomParticle', 'description': {'mass': '1.5e-26 kg', 'charge': 'nan C'}}}
+
+        """
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "mass": str(self.mass),
+            "charge": str(self.charge)
+        }
+        return particle_dictionary
+
```
rocco8773, line 1897:
```suggestion
```

see above comment

plasmapy/particles/particle_class.py
```diff
@@ -466,6 +484,26 @@ def __str__(self) -> str:
...
+        {'plasmapy_particle': {'type': 'Particle', 'description': {'argument': 'Pb'}}}
+        >>> electron = Particle('e-')
+        >>> electron.particle_dict()
+        {'plasmapy_particle': {'type': 'Particle', 'description': {'argument': 'e-'}}}
+
+        """
+        particle_dictionary = super().particle_dict
+        particle_dictionary["plasmapy_particle"]["description"] = {
+            "argument": self.particle
+        }
```
rocco8773, line 504:
```suggestion
        particle_dictionary["plasmapy_particle"]["init_params"] = {
            "argument": self.particle
        }
```

DarkAEther:
> I know you separated the JSON encoding and decoding (PR #836) into separate PRs, but they are so relevant to each other that all the code should be in this PR. Basically the design of the encoding and decoding are so coupled to each other that they should not be split. Having them split make it difficult to traceback what is going on.

I definitely regret doing this. I should have taken the advice given by @StanczakDominik and kept everything in the same PR. This was an oversight on my part. I guess it didn't click at the time. Any advice on how to move forward with this? I could probably close both PRs as experimental PRs and open a final PR that has both the serialization and deserialization together and continue any further discussions there. Or we could continue working with this in a split manner. I'm not very sure on what to do :/ .

plasmapy/particles/particle_class.py
```diff
@@ -11,7 +11,7 @@
 import astropy.constants as const
 import astropy.units as u
 import numpy as np
-
+import json
```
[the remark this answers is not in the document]
DarkAEther: Thank you for the suggestion! Will make this change.

plasmapy/particles/particle_class.py
```diff
@@ -11,7 +11,7 @@
...
 import plasmapy.utils.roman as roman
```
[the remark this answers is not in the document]
DarkAEther: Sure, I can make that change. Thanks!

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +111,31 @@ def mass(self) -> Union[u.Quantity, Real]:
     def charge(self) -> Union[u.Quantity, Real]:
         raise NotImplementedError
 
+    @property
+    def particle_dict(self) -> dict:
```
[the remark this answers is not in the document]
DarkAEther: I agree, `json_dict` is more informative to the user regarding the purpose it serves. Will make that change. Thank you for the suggestion.

plasmapy/particles/particle_class.py
DarkAEther: In the implementation of the deserialization, which I have, unfortunately, kept separately in PR #836 , I use the `description` key as the index for the initialization parameters.

The reason I haven't kept  `mass` or `charge` as default is because the Particle class throws a ChargeError when there is an attempt to retrieve its charge. I could make `mass` default, but reasoned that it may be easier for developers to understand the concept for the overriding of the `particle_dict` property. (i.e overriding the CustomParticle's `particle_dict` property to add `mass` and `charge` versus just `charge). I'd like to ask for advice regarding this bit.

Thanks for the suggestion regarding the date and `module`. I can include them in the JSON representation.

plasmapy/particles/particle_class.py
DarkAEther: I see. This makes sense. I think I had misunderstood the previous comment - sorry about that. I agree with overriding for the `Particle` instead of for `CustomParticle` and `DimensionlessParticle`. I still have the question about `Particle` throwing an exception when we try and get it's charge for something like an elemental particle.

plasmapy/particles/particle_class.py
DarkAEther: Thanks for the suggestion!

plasmapy/particles/particle_class.py
DarkAEther: `"init_params"` seems to be the more suitable key. I shall make this change.

rocco8773:
@DarkAEther Don't beat yourself up about it.  I can't tell you how many rabbit holes I've gone down and had to crawl back out of.  It's all part of the process.

Since the majority of the framework is discussed in PR #836, I'd probably move all the changes over to that PR and close this one.  We can rename PR #836 to better describe what it is doing.

Again, the choice is really up to you.  Just make sure if a new PR is created, then all the relevant old PRs are referenced along with all the key comments.

plasmapy/particles/particle_class.py
rocco8773: Interesting...  Once you get everything into a single PR I'll take a look into where the error is coming from.

DarkAEther:
> @DarkAEther Don't beat yourself up about it. I can't tell you how many rabbit holes I've gone down and had to crawl back out of. It's all part of the process.
> 
> Since the majority of the framework is discussed in PR #836, I'd probably move all the changes over to that PR and close this one. We can rename PR #836 to better describe what it is doing.
> 
> Again, the choice is really up to you. Just make sure if a new PR is created, then all the relevant old PRs are referenced along with all the key comments.

Thank you for the suggestion. I've dug myself into a nice rabbit hole xD. 
I have decided to close this PR, moving all further discussions and commits to PR #836

PR #836, opened 2020-06-06
Issue #759 Serializing and Deserializing particles

DarkAEther:
Related to Issue #759 
Extension on to functionality introduced in PR #833 

This introduces the `from_json` method for the Particle, CustomParticle, and DimensionlessParticle classes that converts from a JSON representation to the appropriate particle object.

  they're failing, ask!).

Suggestions and advice is appreciated :)

Thank you!


DarkAEther:
Codecov/patch test isn't looking too good :(
I'd love some advice on how I could improve that score :)

StanczakDominik:
Looks like it's mostly exceptions. Look into `pytest.raises` - it's a context manager that lets you check whether your code does actually raise your expected exception on invalid inputs, and if not, fails the test.

DarkAEther:
> One concern I have about this design is that I have to know in advance the type of `Particle` I'm going to be restoring from stringified JSON - or I have to manually implement a check for each type. Whereas, since we're storing the class name in the JSON, the dream here would be something like this:
> 
> ```python
> >>> particle_dict = {"type": "Particle", "symbol": "e-"}
> >>> plasmapy.particles.from_json(particle_dict)
> Particle('e-')
> ```
> 
> I thought initially about having it be a class method on the abstract base class - an alternative constructor, if you will, but access through `plasmapy.particles.AbstractParticle.from_json` seems awkward. Having all the subclasses refer to a common method, on the other hand, could lead to `plasmapy.particles.DimensionlessParticle.from_json` returning `CustomParticle`, which may be confusing. I don't know what the best design here would be, to be frank.

Potentially, we could make the `from_json` a method inside `plasmapy.particles` as you've suggested. This method can check the particle type and call the appropriate particle class's to_json method internally and return the particle object. 

This solves the fact that we may not know the particle beforehand as well as developers knowing that they have to implement the method since it is an abstractmethod for `AbstractParticle`.

What do you think?

StanczakDominik:
Sounds good!

rocco8773:
*Disclaimer*: Again, I'm no expert in json and have limited experience with it.

What's the advantage of the current approach over utilizing the [`JSONEncoder`](https://docs.python.org/3/library/json.html#json.JSONEncoder) and [`JSONDecoder`](https://docs.python.org/3/library/json.html#json.JSONDecoder) classes.  As I understand the encoder/decoder framework, an implementation would look something like...

```python
# this does all the translation form Particle to JSON object
class ParticleJSONEncoder(JSONEncoder):
    def default(self, obj):
        # this overrides the JSONEncoder.default method
        # this knows how to convert the Particle class into a JSON format
        return json_particle

# this does all the translation form JSON object to Particle
class ParticleJSONDecoder(JSONDecoder):
    def __init__(self, *args, **kwargs):
        json.JSONDecoder.__init__(self, object_hook=self.object_hook, **kwargs)

    def object_hook(self, dictionary):
        if 'Particle' in dictionary:
            # gather all the relevant info from the dictionary
            return Particle(*args, **kwargs)
        return dictionary
```

Then a use case would look like...

```python
>>> # to save a Particle
>>> p = Particle("e-")
>>> with open('myparticle.json', 'w') as fp:
...     json.dump(p, fp, cls=ParticleJSONEncoder)
...
...     # or define a member that does the above line
...     p.dump(fp)

>>> # to read a particle
>>> with open('myparticle.json', 'r') as fp:
...     p = json.load(fp, cls=ParticleJSONDecoder)
```

Now, the loading is a little trickier than the saving.  I don't think there should be a member on `Particle` that loads; just does not make sense  to me.  We could define a `json_load()` function in `plasmapy/particles/__init__.py`.

----

Now, minus all the effort it takes to figure out the translation (not trivial by any means!), this seems like a logical approach to me.  It allows for expand-ability as the Particle class (and it's sisters) evolve while maintaining a constant UI.

Additionally, this encoder and decoder should be designed to work with any subclass of `AbstractParticle` (i.e. `Particle`, `CustomParticle`, etc.).

DarkAEther:
Thank you for your feedback and suggestions @rocco8773 :)
> What's the advantage of the current approach over utilizing the JSONEncoder and JSONDecoder classes.

The reasoning I had in mind to use the current method was that in future, if another class inheriting `AbstractParticle` was to be created, the developer would have to implement the `to_json` method for it specifically (since the fields needed in JSON could be different for the new type of particle). In that case, they could implement it as method in the class. 

For the conversion of JSON to particle, I followed a similar reasoning - the developer will implement the `from_json` method in the class and then add it to a dictionary in the implementation of `plasmapy.particles.from_json` which will internally call the appropriate class (using the dictionary mentioned earlier as a hash table of sorts) method to return the appropriate particle object. Something like shown below:

```python
particles = {"Particle":Particle,
                  "CustomParticle": CustomParticle,
                  "DimensionlessParticle": DimensionlessParticle
                }
particles['Particle'].from_json(string)
```
While I did explain my reasoning, I'd like to say that I definitely agree that the code can be simplified (maintaining the interfaces) and made easier to maintain by using a custom `JSONEncoder` and `JSONDecoder` class instead of the current method. I'd definitely like to change the implementation accordingly.

There is one question I have though - is it possible to inform future developers that they need to go to the JSONEncoder and JSONDecoder classes and add the implementation for encode and decode for their new type of Particle? If I am not mistaken, the current approach will throw a `NotImplementedError` Exception if a developer forgets to add `to_json` and `for_json` into the class definition. I have basic OOPS knowledge and don't have much experience with it. Any advice here would be appreciated :)

I'm not sure if I've understood the requirements thoroughly. I've been assuming that every class that inherits from `AbstractParticle` (i.e every type of particle) will have to have implementations of `to_json` and `from_json` (i.e every particle can be represented as JSON and be created from a JSON). If the serialization of a particle can be "opt-in" so to speak, we can probably remove the class methods altogether, thereby keeping the UI consistent for the particles.

Furthermore, I just realised that any subclass of Particle, CustomParticle and DimensionlessParticle may not work with the current approach (the methods of `to_json` and `from_json` are already implemented in the super class). This defeats the purpose of the current approach, which will only work for subclasses of the `AbstractParticle`.

What do you think?

Thank you!

rocco8773:
> The reasoning I had in mind to use the current method was that in future, if another class inheriting `AbstractParticle` was to be created, the developer would have to implement the `to_json` method for it specifically (since the fields needed in JSON could be different for the new type of particle). In that case, they could implement it as method in the class.

This is a good point.  Unfortunately all the particle classes do not share the same `__init__` so it's difficult to make something that is completely universal.  It will also be significantly more difficult to generalize loading versus dumping.

I think there could be a hybrid approach to this.  Something like...

```python
class AbstraceParticle(ABC):

    @property
    def json_dict(self):
        # this has default values for all particle types but can be overwritten by subclasses
        json_particle = {
            "type": type(self).__name__,
            "module": self.__module__,
            "mass": p.mass.__str__(),
            "charge: p.charge.__str__(),
        }
        return json_particle

    def json_dump(self, fp):
        json.dump({'plasmapy_particle": self.json_particle})
        # this is where we could apply a JSONEncoder, but if we've already
        # generated a json friendly dictionary then I don't see the point

class Particle(AbstractParticle):
    @property
    def json_particle(self):
        json_particle = super().json_particle
        json_particle["symbol"] = self.particle
        json_particle["init_params"] = {
            "argument": self.particle,
        }
     return json_particle
```
So, that would handle the saving.  Loading would be more difficult and the `JSONDecoder` might be handy.

```python
class ParticleJSONDecoder(JSONDecoder):
    _particle_classes = {
        "CustomParticle": CustomParticle,
        "DimensionlessParticle": DimentionlessParticle,
        "Particle": Particle,
    }

    def object_hook(self, dictionary):
        if 'plasmapy_particle' in dictionary:
            json_particle = dictionary['plasmapy_particle']
            cls_name = json_particle["type"]
            cls = _particle_classes[cls_name]

            if 'init_params' in json_particle:
                # this would not work as cleanly if init parameters 
                # were custom objects (i.e. not a default decoding)
                params = json_particle["init_params"]
                return cls(**params)
            else:
                p = cls()
                p.mass = u.Quantity(json_particle['mass'])
                p.charge = u.Quantity(json_particle['charge'])
                return p

        return dictionary

# In plasmapy/particles/__init__.py define
def json_load_particle(fp):
    return json.load(fp, cls=ParticleJSONDecoder)
```

Sorry for throwing a bunch of code your way.  It's just easier to write it out than try to explain it in a paragraph.

In this approach, any subclassed particle of `AbstractParticle` should be JSON serializable.  If a programmer wants to change the serialized values, then he or she would only need to override the `json_particle` property.

rocco8773:
> I'm not sure if I've understood the requirements thoroughly. I've been assuming that every class that inherits from `AbstractParticle` (i.e every type of particle) will have to have implementations of `to_json` and `from_json` (i.e every particle can be represented as JSON and be created from a JSON).

Yes, every subclass will inherit the members defined in its parent class.

DarkAEther:
> I think there could be a hybrid approach to this. Something like...
> 
> ```python
> class AbstraceParticle(ABC):
> 
>     @property
>     def json_dict(self):
>         # this has default values for all particle types but can be overwritten by subclasses
>         json_particle = {
>             "type": type(self).__name__,
>             "module": self.__module__,
>             "mass": p.mass.__str__(),
>             "charge: p.charge.__str__(),
>         }
>         return json_particle
> 
>     def json_dump(self, fp):
>         json.dump({'plasmapy_particle": self.json_particle})
>         # this is where we could apply a JSONEncoder, but if we've already
>         # generated a json friendly dictionary then I don't see the point
> 
> class Particle(AbstractParticle):
>     @property
>     def json_particle(self):
>         json_particle = super().json_particle
>         json_particle["symbol"] = self.particle
>         json_particle["init_params"] = {
>             "argument": self.particle,
>         }
>      return json_particle
> ```

Thank you for the suggestion! This looks like a really good approach. I shall adopt this technique accordingly. 

> Sorry for throwing a bunch of code your way. It's just easier to write it out than try to explain it in a paragraph.

No issues at all. Infact, it helps a lot. As developers, speaking in code is often more efficient that putting it out in a natural language xD

> So, that would handle the saving. Loading would be more difficult and the JSONDecoder might be handy.

I agree, the `JSONDecoder` approach seems to be the best for converting from JSON to particle. It is elegant and easier to maintain. 

Thank you for all the suggestions!

rocco8773:
I look forward to seeing how this turns out!  Thank you for putting in the time to tackle this.  I probably wouldn't have thought about it until someone brought it up and I'm learning the new stuff now. :)

I did make a typo in my example code, which I have fixed.  The overriding property in `Particle` should have been `json_dict` not `json_particle`.

DarkAEther:
This PR will now include the code for serialization of particles carried from PR #833

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +97,56 @@ def _category_errmsg(particle, category: str) -> str:
...
+    def object_hook(self, json_dict):
+        """
+        Returns the appropriate particle object corresponding to the input JSON
+        string. If not a particle, returns a dictionary object. Logic Implementation for
+        plasmapy.particles.from_json method.
+
+        Examples
+        --------
+        >>> import plasmapy
+        >>> input_str = '{"plasmapy_particle": {"type": "Particle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"argument": "Pb"}}}'
```
StanczakDominik, line 114: Including `date` is an interesting idea. This is the exact kind of safeguard against future functionality changes that I would forget if writing this myself.

Are you using the `module` anywhere?

plasmapy/particles/particle_class.py
```diff
@@ -1685,6 +1776,27 @@ def __repr__(self):
         """
         return f"DimensionlessParticle(mass={self.mass}, charge={self.charge})"
 
+    def json_dict(self) -> dict:
```
StanczakDominik, line 1779: If the superclass has `json_dict` as a property, shouldn't this also be a property?

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -997,3 +998,110 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+    instance = cls(**kwargs)
+    instance_from_json = from_json(json_string)
+    assert u.isclose(instance.mass, instance_from_json.mass, equal_nan=True), pytest.fail(
+            f"Expected a mass value of {instance.mass}\n"
+            f"Received a mass value of {instance_from_json.mass}"
+        )
+    assert u.isclose(instance.charge, instance_from_json.charge, equal_nan=True), pytest.fail(
+            f"Expected a charge value of {instance.charge}\n"
+            f"Received a charge value of {instance_from_json.charge}"
+        )
```
StanczakDominik, line 1040: Could be good to also compare `instance == instance_from_json` here, just like you do below!

plasmapy/particles/particle_class.py
DarkAEther: > Including date is an interesting idea. This is the exact kind of safeguard against future functionality changes that I would forget if writing this myself.

Credits to @rocco8773 for the suggestion. :)

> Are you using the module anywhere?

Not for right now. The `module` stores where the class was defined and is useful for tracking in future if particles do stem from other source code files. It doesn't play a role in the deserialization. Another interesting suggestion from @rocco8773 :)

For reference, the suggestions can be found here: https://github.com/PlasmaPy/PlasmaPy/pull/833#discussion_r436424242

plasmapy/particles/particle_class.py
DarkAEther: Ah, that's right. Thank you for pointing it out!

plasmapy/particles/tests/test_particle_class.py
DarkAEther: Thank you for the suggestion. I will look into adding that check as well.

plasmapy/particles/particle_class.py
StanczakDominik: It gets so hard to follow discussions in previous reviews sometimes... :(

DarkAEther: Indeed it does. Unfortunately, I made the problem worse by splitting the PRs. Oops :(

StanczakDominik: Nah, no worries. My bad, honestly :)

plasmapy/particles/tests/test_particle_class.py
DarkAEther: Interestingly, I tried this check and it lead to some interesting behaviour:
```python
>>> c = CustomParticle(mass = 5.2 * u.kg, charge = 5.2 * u.C)
>>> c1 = from_json('{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.2 kg", "charge": "5.2 C"}}}')
>>> c == c1
False
>>> c.mass == c1.mass
True
>>> c.charge == c1.charge
True
```
Any thoughts?

StanczakDominik: Oh yeah, floating point comparisons, especially when compared to the truncated version saved in JSON? That's a bummer, but I guess it is what it is...

StanczakDominik: Oh wait, I scrolled to the right. Still, I'd expect this to be a floating point thing. I can run a debugger to make sure.

DarkAEther: Just a thought, but could it be because there is no `__eq__` defined for the `CustomParticle` class? I noticed that `__eq__` is defined for `Particle` and this seems to pass the test. I'm guessing that it may be using the default behaviour for equality of `CustomParticle`.

StanczakDominik: Welp. I just realized `Particle` has `__eq__`, but `CustomParticle` doesn't! Definitely out of scope for this issue :laughing:

plasmapy/particles/particle_class.py
StanczakDominik:
```suggestion
    @property
    def json_dict(self) -> dict: 
```
halping! :laughing:

plasmapy/particles/particle_class.py
```diff
@@ -466,6 +537,26 @@ def __str__(self) -> str:
         """Return the particle's symbol."""
         return self.particle
 
+    def json_dict(self) -> dict:
```
StanczakDominik, line 540:
```suggestion
    @property
    def json_dict(self) -> dict:
```

DarkAEther:
Any hints on the flake8 and the pre-commit errors? It is the first time I'm encountering them.

StanczakDominik:
Please ignore them, they're #845 gone wrong. :laughing:

DarkAEther:
> Please ignore them, they're #845 gone wrong. laughing
> And I did just add them yesterday!

No issues! Also, I'd like to apologise. I think I accidentally committed the same changes over your commits. Oops xD. Thanks for adding in the changes though!

StanczakDominik:
OK, if you merge current master in, it's gonna fix the linters :)

DarkAEther:
One left :D The precommit Style. Looks like I'll be making some style changes xD. Any way to integrate this into `pycodestyle` or another checker tool that runs locally? Would be helpful to have these checks done before commits are made and tested out.

DarkAEther:
> Any way to integrate this into pycodestyle or another checker tool that runs locally?

Yes, yes there is. Use the black formatter https://black.readthedocs.io/en/stable/

DarkAEther:
Awesome! The new checks passed too!
Any further feedback is welcome :)

StanczakDominik:
> > Any way to integrate this into pycodestyle or another checker tool that runs locally?
> 
> Yes, yes there is. Use the black formatter https://black.readthedocs.io/en/stable/

The checks have a message in logs that says using pre-commit will run black automatically, with instructions on installing. I'm guessing it'd be better to put that in the PR template as well? :D

DarkAEther:
> > > Any way to integrate this into pycodestyle or another checker tool that runs locally?
> > 
> > 
> > Yes, yes there is. Use the black formatter https://black.readthedocs.io/en/stable/
> 
> The checks have a message in logs that says using pre-commit will run black automatically, with instructions on installing. I'm guessing it'd be better to put that in the PR template as well? :D

True. I didn't realise that black was actually a tool that we could download and run locally. I initially thought it was a GitHub specific thing. Adding it into the PR template would definitely help clarify :)

DarkAEther:
@StanczakDominik I added my details to `credits.rst` in the last commit. I hope that's alright. Let me know if there are any further suggestions.

Thank you!

rocco8773:
@DarkAEther Sorry for the slow review on this, I've been pretty busy the last week or so.  You should see me review on this in the next day or so.

plasmapy/particles/__init__.py
```diff
@@ -46,3 +50,7 @@
 deuteron = Particle("D 1+")
 triton = Particle("T 1+")
 alpha = Particle("He-4 2+")
+
+
+def from_json(json_string):
```
rocco8773, line 55: Needs docstrings.

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +97,59 @@ def _category_errmsg(particle, category: str) -> str:
...
+        >>> import plasmapy
+        >>> input_str = '{"plasmapy_particle": {"type": "Particle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"argument": "Pb"}}}'
+        >>> plasmapy.particles.from_json(input_str)
+        Particle("Pb")
+        >>> input_str = '{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.12 kg", "charge": "6.2 C"}}}'
+        >>> plasmapy.particles.from_json(input_str)
+        CustomParticle(mass=5.12 kg, charge=6.2 C)
+        >>> input_str = '{"plasmapy_particle": {"type": "DimensionlessParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.2", "charge": "6.3"}}}'
+        >>> plasmapy.particles.from_json(input_str)
+        DimensionlessParticle(mass=5.2, charge=6.3)
```
rocco8773, line 121: I don't think you need to have examples here, it's not something user's will be playing with and the `pytests` cover its functionality.

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +97,59 @@ def _category_errmsg(particle, category: str) -> str:
...
 
+class ParticleJSONDecoder(json.JSONDecoder):
+    def __init__(self, *args, **kwargs):
+        json.JSONDecoder.__init__(self, object_hook=self.object_hook, *args, **kwargs)
+
+    def object_hook(self, json_dict):
+        """
+        Returns the appropriate particle object corresponding to the input JSON
+        string. If not a particle, returns a dictionary object. Logic Implementation for
+        plasmapy.particles.from_json method.
```
rocco8773, line 108:
```suggestion
        An `object_hook` utilized by the `json` deserialization processes to decode
        json strings into a `plasmapy` particle class (`AbstractParticle`,
        `CustomParticle`, `DimensionlessParticle`, `Particle`).
```

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +97,59 @@ def _category_errmsg(particle, category: str) -> str:
...
+                    particle_init_args["mass"] = np.float64(particle_init_args["mass"])
+                    particle_init_args["charge"] = np.float64(
+                        particle_init_args["charge"]
+                    )
+                return particle_types[particle_class](**particle_init_args)
+            else:
+                raise InvalidParticleError(
+                    f"Particle with type {particle_class} is not "
+                    f"present in the _particle_types dictionary or is undefined."
+                )
```
rocco8773, line 148:
```suggestion
            try:
                pardict = json_dict["plasmapy_particle"]
                partype = pardict["type"]
                args = pardict["__init__"]["args"]
                kwargs = pardict["__init__"]["kwargs"]
                particle = particle_types[partype](*args, **kwargs)
            except KeyError:
                raise InvalidElementError(
                    f"json file does not define a valid plasmapy particle"
                )
            return particle
```

With the updated `json_dict`, this can run much more succinctly.  However, in-order for this to work with `CustomParticle` the parameter validations need to be updated so they can handle string representations of astropy `Quantity`'s.  I've already figured this out and will open a separate PR on it.  Basically we want the following to work...

```python
par = CustomeParticle(mass="5.2 kg")
```

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -995,3 +998,113 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
+        '{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.12 kg", "charge": "6.2 C"}}}',
+    ),
+    (
+        CustomParticle,
+        {"mass": 5.12 * u.kg},
+        '{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.12 kg", "charge": "nan C"}}}',
+    ),
+]
```
rocco8773, line 1024:
```suggestion
custom_particles_from_json_tests = [
    (
        DimensionlessParticle,
        {"mass": 5.2, "charge": 6.3},
        '{"plasmapy_particle": {"type": "DimensionlessParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": [], "kwargs": {"mass": 5.2, "charge": 6.3}}}}',
    ),
    (
        DimensionlessParticle,
        {"mass": 5.2},
        '{"plasmapy_particle": {"type": "DimensionlessParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": [], "kwargs": {"mass": 5.2, "charge": NaN}}}}',
    ),
    (
        CustomParticle,
        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
        '{"plasmapy_particle": {"type": "CustomParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": [], "kwargs": {"mass": "5.12 kg", "charge": "6.2 C"}}}}',
    ),
    (
        CustomParticle,
        {"mass": 5.12 * u.kg},
        '{"plasmapy_particle": {"type": "CustomParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": [], "kwargs": {"mass": "5.12 kg", "charge": "nan C"}}}}',
    ),
]
```

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -995,3 +998,113 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+        Particle,
+        {"argument": "Pb"},
+        '{"plasmapy_particle": {"type": "Particle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"argument": "Pb"}}}',
+    ),
+    (
+        Particle,
+        {"argument": "e-"},
+        '{"plasmapy_particle": {"type": "Particle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"argument": "e-"}}}',
+    ),
+]
```
rocco8773, line 1058:
```suggestion
particles_from_json_tests = [
    (
        Particle,
        {"argument": "Pb"},
        '{"plasmapy_particle": {"type": "Particle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": ["Pb"], "kwargs": {}}}}',
    ),
    (
        Particle,
        {"argument": "e-"},
        '{"plasmapy_particle": {"type": "Particle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-08", 
        "__init__": {"args": ["e-"], "kwargs": {}}}}',
    ),
]
```

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -995,3 +998,113 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
+        '{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-09", "init_args": {"mass": "5.12 kg", "charge": "6.2 C"}}}',
+    ),
+    (
+        DimensionlessParticle,
+        {"mass": 5.2, "charge": 6.3},
+        '{"plasmapy_particle": {"type": "DimensionlessParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-09", "init_args": {"mass": "5.2", "charge": "6.3"}}}',
+    ),
+]
```
rocco8773, line 1089:
```suggestion
particle_json_repr_table = [
    (
        Particle,
        {"argument": "lead"},
        '{"plasmapy_particle": {"type": "Particle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-09", 
        "__init__": {"args": ["Pb"], "kwargs": {}}}}',
    ),
    (
        CustomParticle,
        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
        '{"plasmapy_particle": {"type": "CustomParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-09", 
        "__init__": {"args": [], "kwargs": {"mass": "5.12 kg", "charge": "6.2 C"}}}}',
    ),
    (
        DimensionlessParticle,
        {"mass": 5.2, "charge": 6.3},
        '{"plasmapy_particle": {"type": "DimensionlessParticle", 
        "module": "plasmapy.particles.particle_class", 
        "date_created": "2020-06-09", 
        "__init__": {"args": [], "kwargs": {"mass": 5.2, "charge": 6.3}}}}',
    ),
]
```

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -995,3 +998,113 @@ def test_customized_particle_repr(cls, kwargs, expected_repr):
...
+    json_repr = instance.to_json()
+    test_dict = json.loads(json_repr)["plasmapy_particle"]
+    expected_repr = json.loads(expected_repr)["plasmapy_particle"]
+    assert test_dict["type"] == expected_repr["type"], pytest.fail(
+        f"Problem with JSON representation of {cls.__name__} "
+        f"with kwargs = {kwargs}.\n\n"
+        f"expected type = {expected_repr['type']}\n\n"
+        f"got type: {test_dict['type']}"
+    )
+    assert expected_repr["init_args"] == test_dict["init_args"], pytest.fail(
```
rocco8773, line 1105:
```suggestion
    assert expected_repr["__init__"] == test_dict["__init__"], pytest.fail(
```

plasmapy/particles/particle_class.py
```diff
@@ -10,9 +10,9 @@
 
 import astropy.constants as const
```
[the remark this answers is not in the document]
StanczakDominik: Which is why I really, really, really want to automate this with `isort` ASAP...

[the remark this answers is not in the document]
rocco8773: Can’t 👍 this enough!

plasmapy/particles/__init__.py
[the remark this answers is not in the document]
DarkAEther: Thank you for the suggestion! Could I ask for a clarification on this? I don't think I'm too clear on the change to be made here.
Thank you!

plasmapy/particles/__init__.py
```diff
@@ -37,6 +37,10 @@
     particle_symbol,
 )
 
+# load the JSON decoder
+import json
+from plasmapy.particles.particle_class import ParticleJSONDecoder
+
```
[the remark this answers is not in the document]
DarkAEther: Sounds good! Changes made.

plasmapy/particles/particle_class.py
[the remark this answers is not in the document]
DarkAEther: I've sorted the imports. (I hope I got it right xD) Can't wait for an automated solution for this :)

plasmapy/particles/particle_class.py
DarkAEther: Thank you for suggesting this! It looks much more elegant now.

plasmapy/particles/tests/test_particle_class.py
DarkAEther: I tried this one out, but ended up with an `EOL while scanning string literal` error. Apparently these strings can't be split across multiple lines? Not very sure about this one.

plasmapy/particles/__init__.py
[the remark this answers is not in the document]
rocco8773: The top of the `__init__.py` file should follow a template that looks like...

```python
"""
This is the module level docstring.
"""
__all__ = ["from_json"]

# do all of your imports
from plasmapy.particles.atomic import (...)
...

# now write your code
...
```

plasmapy/particles/tests/test_particle_class.py
rocco8773: It should work just fine.  I tested all these suggestions before suggesting them.

Note: You don't have to hand copy these suggestions to your local code.  You can directly commit the suggestion to your branch from the PR.  You can commit one suggestion at a time by using the "Commit suggestion" button.  Or, you can commit a batch of suggestions by going to the "Files changed" tab and add each suggest to a batch with the "Add suggestion to batch" button.

DarkAEther: Oh I see. Thank you for the suggestion. I wasn't aware of this feature. Definitely convenient.

DarkAEther:
@rocco8773 Thank you so much for all the suggestions. I realised just how much time and effort goes into making them. I introduced a bunch of conflicts while working without knowing about the direct way of making commits from suggestions. For this set of commits, I've decided to work locally. From the next set of suggestions, I'll be sure to use the suggestions to commit feature. 

The last remaining suggestions that I need to work on are the `json.dumps` vs `json.dump` and the long lines in  `test_particle_class.py`. I will try and directly use your suggestions for the `test_particle_class.py` file.

Thank you!

DarkAEther:
> I will try and directly use your suggestions for the test_particle_class.py file.

Okay, looks like I completely messed up. It won't let me use suggestions which show up as outdated :( 
I tried copy pasting the suggestion as it appears into my editor, but it throws `SyntaxError: EOL while scanning string literal`.

@rocco8773 I'd like to apologise for taking up more of your time for any repeat suggestions. I'll use them directly from the PR from now on.

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +165,33 @@ def mass(self) -> Union[u.Quantity, Real]:
     def charge(self) -> Union[u.Quantity, Real]:
         raise NotImplementedError
 
+    @property
+    def json_dict(self) -> dict:
+        json_dictionary = {
+            "plasmapy_particle": {
+                "type": type(self).__name__,
+                "module": self.__module__,
+                "date_created": str(datetime.now().date()),
```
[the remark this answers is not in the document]
DarkAEther: Thank you! Added this into the code.

plasmapy/particles/particle_class.py
DarkAEther: Makes sense. I have removed the examples from this bit.

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +165,33 @@ def mass(self) -> Union[u.Quantity, Real]:
...
         Raise an `~plasmapy.utils.AtomicError` because particles
         do not have a truth value.
         """
         raise AtomicError("The truth value of a particle is not defined.")
 
+    def to_json(self) -> str:
+        """
+        Return a JSON string representation of the minimum description of a particle.
+        """
+        return json.dumps(self.json_dict)
```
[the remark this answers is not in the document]
DarkAEther: I feel that Option 1 works in favour of the requirements of storing the details of a simulation in a file.

plasmapy/particles/__init__.py
```diff
@@ -46,3 +50,10 @@
 deuteron = Particle("D 1+")
 triton = Particle("T 1+")
 alpha = Particle("He-4 2+")
+
+
+def from_json(json_string):
+    """
+    Returns the appropriate object from its JSON string representation
+    """
+    return json.loads(json_string, cls=ParticleJSONDecoder)
```
rocco8773, line 59: I just realized this is a `json.loads`, while it should be `json.load`.  `json.load` will load from a json text file, while `json.loads` loads from a json string.  My understanding of this serialization functionality is to let a user save a particle object and reload it in 10 years.  In that case, `from_json` should definitely utilize `json.load` over `json.loads`.

StanczakDominik: And then you can sort of get the `json.loads` functionality back through https://docs.python.org/3/library/io.html#text-i-o  !

DarkAEther: Ah right! That makes sense. However, how do we take care of multiple JSONs in the same file? This would cause an error since the `ParticleJSONDecoder` is only designed to process a single particle's JSON string at a time. We'd need some way of giving the `ParticleJSONDecoder` one string at a time. Any thoughts on this?

DarkAEther: Another thing that came to mind is that the user might want to store the file as they please. If I remember, @StanczakDominik had mentioned serializing particles that were in an xarray after a simulation. Using loads and dumps could potentially give users the benefit of flexibility to store the file as they please together with other simulation data in future. 
A possible function the user would write to store the particles would be something similar to:
```python
def put_my_particles_into_a_file(particle_list, filepath):
    particle_strings = list(map(lambda x: x.to_json(),particle_list))
    with open(filepath,"w") as particle_file:
         for particle in particle_strings:
              particle_file.write(particle)
```

A possible function the user would write to reload the particles would be something like this:
```python
def get_my_particles_back(filepath):
    particles_in_file = None
    particle_objects = []
    with open(filepath,"r") as particle_file:
         particles_in_file = particle_file.readlines()
    particle_objects = list(map(from_json,particles_in_file)) #converts each particle JSON string to respective particle object
    return particle_objects
```
A sample file would look like:
```txt
{"plasmapy_particle" : {..}}
{"plasmapy_particle" : {..}}
{"plasmapy_particle" : {..}}
```
Just a thought from my side. I'd love feedback on whether this is easy to use for the user.

rocco8773: I think saving and loading multiple particles is another project in itself, and should be given it's own PR.  From my limited research, it looks like writing multiple python objects to a single json file is not a trivial task.  It looks like you have to bundle it all into one json object and write that.  Then we have to address the issue with unique json keys, since we're currently saving all particles with the `'plasmapy_particle'` key.  I do think we can leverage our `ParticleJSONDecoder` to pull out all the particle objects and return a list of particles; thus, no need to have anything else than `plasmapy.particles.from_json` to implement this functionality.

DarkAEther: I have a few ideas for this. I would love to share them and perhaps work on the one we agree upon to move forward with. I agree, a new PR is a good idea. 

Any further suggestions and advice are welcome! 

Thank you!

DarkAEther:
I'd love some advice regarding the `codecov/patch` check. Couldn't figure out how to fix that.
According to the details, the lines of code in `particle_class.py` that were highlighted in red were:

```python

except KeyError:
--
  | 127 | + | raise InvalidElementError(
  | 128 | + | f"json file does not define a valid plasmapy particle"
  | 129 | + | )

```

Thank you!

StanczakDominik:
The `raise InvalidParticleError` can be tested using the `pytest.raises` context manager (grep for other uses in the code base here!) with an invalid particle input - for example, try to have it parse a particle with the symbol "potato".

plasmapy/particles/__init__.py
StanczakDominik: Hey, I'm sorry I've been away from the project for a bit! I'm absolutely down to break accomplishing my xarray dreams into a few PRs!

DarkAEther:
An interesting thing popped up. For reference, here is the code for the `ParticleJSONDecoder`:
```python
if "plasmapy_particle" in json_dict:
            try:
                pardict = json_dict["plasmapy_particle"]
                partype = pardict["type"]
                args = pardict["__init__"]["args"]
                kwargs = pardict["__init__"]["kwargs"]
                particle = particle_types[partype](*args, **kwargs)
                return particle
            except KeyError:
                raise InvalidElementError(
                    f"json file does not define a valid plasmapy particle"
                )
else:
            return json_dict
```
The reason I need to check for `if "plasmapy_particle" in json_dict:` is because, apparently, `JSONDecoder` builds up its strings recursively in bottom up fashion, which means that this section of code is actually executed multiple times until the highest level key `'plasmapy_particle'` is found. This defeats the purpose of the exception handling part. Any thoughts on this?

DarkAEther:
> An interesting thing popped up. For reference, here is the code for the `ParticleJSONDecoder`:
> 
> ```python
> if "plasmapy_particle" in json_dict:
>             try:
>                 pardict = json_dict["plasmapy_particle"]
>                 partype = pardict["type"]
>                 args = pardict["__init__"]["args"]
>                 kwargs = pardict["__init__"]["kwargs"]
>                 particle = particle_types[partype](*args, **kwargs)
>                 return particle
>             except KeyError:
>                 raise InvalidElementError(
>                     f"json file does not define a valid plasmapy particle"
>                 )
> else:
>             return json_dict
> ```
> 
> The reason I need to check for `if "plasmapy_particle" in json_dict:` is because, apparently, `JSONDecoder` builds up its strings recursively in bottom up fashion, which means that this section of code is actually executed multiple times until the highest level key `'plasmapy_particle'` is found. This defeats the purpose of the exception handling part. Any thoughts on this?

This unfortunately makes it rather challenging to do the intended task of checking whether the input is a valid particle JSON and not something else.

rocco8773:
> The reason I need to check for `if "plasmapy_particle" in json_dict:` is because, apparently, `JSONDecoder` builds up its strings recursively in bottom up fashion, which means that this section of code is actually executed multiple times until the highest level key `'plasmapy_particle'` is found. This defeats the purpose of the exception handling part. Any thoughts on this?

No...no...the way this works is that the error is only raised if a particle is in the json file and it's json object is not properly defined.  This is why the try-statement is nested in the if-statement.  An error is never raised if `'plasmapy_particle'` is not in the json file, it is only thrown if `'plasmapy_particle'` is in the json file and the associated json object is not structured properly.

If `'plasmapy_particle'` is not in the json file, then the normal json decoder is used to generate the object.

DarkAEther:
> > The reason I need to check for `if "plasmapy_particle" in json_dict:` is because, apparently, `JSONDecoder` builds up its strings recursively in bottom up fashion, which means that this section of code is actually executed multiple times until the highest level key `'plasmapy_particle'` is found. This defeats the purpose of the exception handling part. Any thoughts on this?
> 
> No...no...the way this works is that the error is only raised if a particle is in the json file and it's json object is not properly defined. This is why the try-statement is nested in the if-statement. An error is never raised if `'plasmapy_particle'` is not in the json file, it is only thrown if `'plasmapy_particle'` is in the json file and the associated json object is not structured properly.
> 
> If `'plasmapy_particle'` is not in the json file, then the normal json decoder is used to generate the object.

Ahh, I see. Thank you! I'll add some testcases for that exception.

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +98,38 @@ def _category_errmsg(particle, category: str) -> str:
     return errmsg
 
 
+class ParticleJSONDecoder(json.JSONDecoder):
+    def __init__(self, *args, **kwargs):
+        json.JSONDecoder.__init__(self, object_hook=self.object_hook, *args, **kwargs)
+
+    def object_hook(self, json_dict):
```
rocco8773, line 105:
```suggestion
    def __init__(self, *, object_hook=None, parse_float=None,
                 parse_int=None, parse_constant=None, strict=True,
                 object_pairs_hook=None):

        if object_hook is None:
            object_hook = self.particle_hook

        json.JSONDecoder.__init__(self,
                                  object_hook=object_hook,
                                  parse_float=parse_float,
                                  parse_int=parse_int,
                                  parse_constant=parse_constant,
                                  strict=strict,
                                  object_pairs_hook=object_pairs_hook)

    @staticmethod
    def particle_hook(json_dict):
```

This structure is more friendly with the internal code of `JSONDecode`.   Not so simply put, `JSONDecode` sets the `object_hook` method from the equally named input parameter (i.e. `self.object_hook = object_hook`).  If we define our hook method as `object_hook` then the `JSONDecode` statement will always be overridden by our method.

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +145,44 @@ def mass(self) -> Union[u.Quantity, Real]:
...
+        return json_dictionary
+
     def __bool__(self):
         """
         Raise an `~plasmapy.utils.AtomicError` because particles
         do not have a truth value.
         """
         raise AtomicError("The truth value of a particle is not defined.")
 
+    def to_json(self) -> str:
```
rocco8773, line 180:
```suggestion
    def json_dumps(self) -> str:
```

As I've mentioned in previous comments, a `json.dumps` method should be in addition to a `json.dump` method.  These method should also be explicitly named to reflect dumps vs dump.

changelog/836.feature.rst
```diff
@@ -0,0 +1,2 @@
+Introduced `to_json` method for AbstractParticleClass that converts a particle into a JSON representation
+Introduced `plasmapy.particles.from_json` method to convert JSON strings to corresponding particle
```
rocco8773, line 2: Will need to be updated when issue around dumps vs dump and loads vs load is resolved.

plasmapy/particles/particle_class.py
```diff
@@ -111,13 +145,44 @@ def mass(self) -> Union[u.Quantity, Real]:
...
+        inherit from AbstractParticle.
+
+        Keys:
+            plasmapy_particle - Specifies that this JSON is for some particle
+                type - The type of the particle (e.g Particle or CustomParticle)
+                module - Specifies the module where the class was declared
+                date_created - Presents the UTC date and time of serialization
+                __init__ - contains a dictionary of the arguments used to create the particle
+                    args - non keyword arguments used to create a particle
+                    kwargs - keyword arguments used to create a particle
```
rocco8773, line 161: I don't think this will render well in the RTD documentation, but I also just realized Sphinx isn't even rendering this for RTD. 😳 I'll look into this and will probably need push through a new PR for it.

plasmapy/particles/__init__.py
```diff
@@ -2,6 +2,9 @@
 The `plasmapy.particles` subpackage provides access to information about
 atoms, isotopes, ions, and other particles.
 """
+__all__ = ["from_json"]
```
rocco8773, line 5:
```suggestion
__all__ = [
    "from_json",
    "AbstractParticle",
    "CustomParticle",
    "DimensionlessParticle",
    "Particle",
]
```

plasmapy/particles/particle_class.py
```diff
@@ -2,17 +2,18 @@
 
 __all__ = ["AbstractParticle", "Particle", "DimensionlessParticle", "CustomParticle"]
```
rocco8773, line 3:
```suggestion
__all__ = [
    "AbstractParticle",
    "CustomParticle",
    "DimensionlessParticle", 
    "Particle",
    "ParticleJSONDecoder",
]
```

plasmapy/particles/__init__.py
```diff
@@ -46,3 +50,10 @@
 deuteron = Particle("D 1+")
 triton = Particle("T 1+")
 alpha = Particle("He-4 2+")
+
+
+def from_json(json_string):
+    """
+    Returns the appropriate object from its JSON string representation
```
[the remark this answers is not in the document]
DarkAEther: Just to clarify, the documentation is for the `json_string` parameter right? Would a one liner explaining what it is (i.e a string that represents the JSON) suffice?

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +98,38 @@ def _category_errmsg(particle, category: str) -> str:
     return errmsg
 
 
+class ParticleJSONDecoder(json.JSONDecoder):
```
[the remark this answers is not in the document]
DarkAEther: Thank you for the suggestion. Will do that.

plasmapy/particles/particle_class.py
DarkAEther: I see. If possible, could I get a sample format for how I could structure the keys of the JSON?

DarkAEther:
@rocco8773  Regarding JSON dumps vs dump and loads vs load, I think a possible way forward (based on your suggestion), could be:
Dumps and Dump:
```python
def json_dumps(self): #from your suggestion

def json_dump(self,dest_file_path):
        return json.dump(self.json_dict,dest_file_path) #Since we are working with only one particle per file at the moment
```
Loads and Load:
```python
def from_json(json_string):
    """
    Returns the appropriate object from input JSON string representation

    Parameters
    --------
    `json_string` is the JSON string that represents a particle
    """
    return json.loads(json_string, cls=ParticleJSONDecoder)

def from_json_file(json_file_path):
    """
    Returns the appropriate object from input JSON file

    Parameters
    --------
    `json_file_path` is the file path to the JSON file that contains the JSON
    string describing the particle
    """
    json_string = None
    with open(json_file_path,'r') as json_file:
        json_string = json_file.read()
    return from_json(json_string)
```
I'd love to hear your thoughts on this potential approach.

plasmapy/particles/particle_class.py
DarkAEther: I checked out the details at [Sphinx RTD Docstring Format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) 

Is this suitable for use here?

DarkAEther: > I checked out the details at [Sphinx RTD Docstring Format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
> 
> Is this suitable for use here?

Probably better for use with methods and class docstrings as opposed to describing the dictionary though

rocco8773:
I think the json functionality we provide should match the naming scheme and signatures of the `json` functionality.  I say this because it make the functionality more intuitive to the user.  Thus, ...

* `json_dumps` should be the method we already have on the particle class.

* `json_dump` should be another method on the particle class, but its signature should reflect `json.dump`.  Thus, it would be used like...

    ```python
    par = Particle("e-")
    with open("my_particle.json", "w") as fp:
        par.json_dump(fp)
    ```

* `from_json` -> `json_loads_particle` and the signature can remain as is.  I like `json_loads_particle` because it matches the `json` naming scheme and it's more explicit. @StanczakDominik What do you think?

* `from_json_file` -> `json_load_particle` and it should match the signature of `json.load`.  Thus, it would be used like ...

    ```python
    with open("my_particle.json", "r") as fp:
        par = json_load_particle(fp)
    ```

StanczakDominik:
Naming scheme looks okay; I don't really mind either way.

Would you mind if I added a few changes to this, via a separate pull request to your repo? I've got a few ideas that would make the json file opener more useful than just doing single particle stuff, and it'll be faster to code them up, honestly!

DarkAEther:
> Naming scheme looks okay; I don't really mind either way.
> 
> Would you mind if I added a few changes to this, via a separate pull request to your repo? I've got a few ideas that would make the json file opener more useful than just doing single particle stuff, and it'll be faster to code them up, honestly!

This sounds awesome! Please go ahead. 
Thank you :)

DarkAEther:
> I think the json functionality we provide should match the naming scheme and signatures of the `json` functionality. I say this because it make the functionality more intuitive to the user. Thus, ...
> 
> * `json_dumps` should be the method we already have on the particle class.
> * `json_dump` should be another method on the particle class, but its signature should reflect `json.dump`.  Thus, it would be used like...
>   ```python
>   par = Particle("e-")
>   with open("my_particle.json", "w") as fp:
>       par.json_dump(fp)
>   ```
> * `from_json` -> `json_loads_particle` and the signature can remain as is.  I like `json_loads_particle` because it matches the `json` naming scheme and it's more explicit. @StanczakDominik What do you think?
> * `from_json_file` -> `json_load_particle` and it should match the signature of `json.load`.  Thus, it would be used like ...
>   ```python
>   with open("my_particle.json", "r") as fp:
>       par = json_load_particle(fp)
>   ```

Thank you for your inputs. I'll follow this approach and make changes.

DarkAEther:
Could I get help with the `test-documentation` check. Couldn't find any details why it is failing? Thank you.

rocco8773:
The errors are mostly likely due to you using [reST style docstrings](https://www.python.org/dev/peps/pep-0287/), while we use [numpy style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html).  ([styles of docstrings](https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format)).  You'll need to edit all your docstrings to match the numpydoc format.

DarkAEther:
> The errors are mostly likely due to you using [reST style docstrings](https://www.python.org/dev/peps/pep-0287/), while we use [numpy style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html). ([styles of docstrings](https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format)). You'll need to edit all your docstrings to match the numpydoc format.

Interesting. I wasn't aware that there were so many styles out there. I'll make the change to numpy style. Thank you for the information!

DarkAEther:
Any further comments and feedback are welcome. Thank you!

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +104,91 @@ def _category_errmsg(particle, category: str) -> str:
...
+                args = pardict["__init__"]["args"]
+                kwargs = pardict["__init__"]["kwargs"]
+                particle = particle_types[partype](*args, **kwargs)
+                return particle
+            except KeyError:
+                raise InvalidElementError(
+                    f"json file does not define a valid plasmapy particle"
+                )
+        else:
+            return json_dict
```
StanczakDominik, line 189: I think we could extract this whole staticmethod to be a function, and then we get a programmatic way of converting a dictionary into a Particle for free! Then we could simply have this method be a thin wrapper on top of that function.

plasmapy/particles/__init__.py
```diff
@@ -46,3 +57,30 @@
...
+def json_loads_particle(json_string):
+    """
+    Returns the appropriate object from input JSON string representation
+
+    Parameters
+    ----------
+    json_string: string
+        JSON string that represents a particle
+    """
+    return json.loads(json_string, cls=ParticleJSONDecoder)
```
StanczakDominik, line 86: I think this deserves a separate file - it's a good practice to keep code in `__init__.py` to a minimum.

plasmapy/particles/__init__.py
```diff
@@ -46,3 +57,30 @@
 deuteron = Particle("D 1+")
 triton = Particle("T 1+")
 alpha = Particle("He-4 2+")
+
+
+def json_load_particle(json_file_object):
```
rocco8773, line 62: Use the same [signature as `json.load`](https://docs.python.org/3/library/json.html#json.load).

plasmapy/particles/__init__.py
```diff
@@ -46,3 +57,30 @@
...
+
+    Parameters
+    ----------
+    json_file_object: file object
+        file object of the JSON file that contains the JSON string describing
+        the particle
+    """
+    json_string = None
+    json_string = json_file_object.read()
+    return json_loads_particle(json_string)
```
rocco8773, line 74:
```suggestion
    return json.load(fp, cls=ParticleJSONDecoder)
```

But you should also pass-through all of the `json.load` parameters.

plasmapy/particles/__init__.py
```diff
@@ -46,3 +57,30 @@
...
+
+
+def json_loads_particle(json_string):
```
rocco8773, line 77: Use the same [signature as `json.loads`](https://docs.python.org/3/library/json.html#json.loads).

plasmapy/particles/__init__.py
rocco8773, line 86: Pass-through all of the `json.loads` parameters.

plasmapy/particles/particle_class.py
rocco8773: I debated this myself.  I had if left like this because it's consistent with the `JSONDecoder` framework.  We could drop the `JOSNDeocder` for something more like...

```python
def particle_hook(....):
    ...

with open('particle.json', 'r') as fp:
    par = json.load(fp, object_hook=particle_hook, ...)
```

I could go either way on this, but I think I slightly prefer the `JSONDecoder` framework.   In the end, I say we leave the PR as it is and, if the need arises to convert dictionaries into particles, then we can modify with a separate PR.

plasmapy/particles/__init__.py
DarkAEther: Thank you! That is a good idea. Any hints on a suitable file name? I was thinking something like `particle_json_utils.py`

DarkAEther: Thank you! Will include those.

rocco8773: `json.py` or `serialization.py` are the first two ideas that come to mind. I think I would lean towards the later.

DarkAEther: Thank you! Will change the name to `serialization.py`

DarkAEther:
Could I ask for feedback regarding the docstrings for the `ParticleJSONDecoder` class and the `json_load_particle` and `json_loads_particle` functions? 

Thank you!

plasmapy/particles/particle_class.py
```diff
@@ -96,6 +104,91 @@ def _category_errmsg(particle, category: str) -> str:
...
+
+        json.JSONDecoder.__init__(
+            self,
+            object_hook=object_hook,
+            parse_float=parse_float,
+            parse_int=parse_int,
+            parse_constant=parse_constant,
+            strict=strict,
+            object_pairs_hook=object_pairs_hook,
+        )
```
rocco8773, line 161:
```suggestion
    """
    A custom `~json.JSONDecoder` class to deserialize JSON objects into
    PlasmaPy particle objects.

    Parameters
    ----------
    object_hook:
        If specified, will be called with the result of every JSON object
        decoded and its return value will be used in place of the given `dict`.
        This can be used to provide custom deserializations (e.g. to support
        JSON-RPC class hinting)  (If ot specified, then defaults
        `particle_hook`.)

    **kwargs:
        Any keyword accepted by `~json.JSONDecoder`.
    """

    def __init__(self, *, object_hook=None, **kwargs):
        if object_hook is None:
            object_hook = self.particle_hook

        json.JSONDecoder.__init__(self, object_hook=object_hook, **kwargs)
```

I know I suggested a lot of the original code for this, and I apologize changing this up on you.  I think we should only document what we're doing differently from `JSONDecoder` and heavily rely on `JSONDecoder`'s own documentation.  No reason to reinvent the wheel.

DarkAEther: No worries! I agree that this will make it easier to understand for future reference as well. Thanks for the suggestion :)

changelog/836.feature.rst
```diff
@@ -0,0 +1,4 @@
+Introduced `json_dumps` method for AbstractParticleClass that converts a particle into a JSON string representation
+Introduced `json_dump` method for AbstractParticleClass that converts a particle into a JSON string representation and writes that string into a specified file object
+Introduced `plasmapy.particles.json_loads_particle` method to convert JSON strings to corresponding particle
+Introduced `plasmapy.particles.json_load_particle` method to convert JSON from file object to corresponding particle
```
rocco8773, line 4:
```suggestion
Add `json_dumps` method to `~plasmapy.particles.particle_class.AbstractParticle` to 
convert a particle object into a JSON string.  Add `json_dump` method to 
`~plasmapy.particles.particle_class.AbstractParticle` to serialize a particle
object and writes it to a file.  Add JSON decoder 
`~plasmapy.particles.serialization.ParticleJSONDecoder` to deserialized JSON objects
into particle objects.  Add `plasmapy.particles.serialization.json_loads_particle` 
function to convert JSON strings to particle objects (using 
`~plasmapy.particles.serialization.ParticleJSONDecoder`).  Add 
`plasmapy.particles.json_load_particle` function deserialized a JSON file into a
particle object (using `~plasmapy.particles.serialization.ParticleJSONDecoder`).
```

DarkAEther:
> I think that does it for me then. @DarkAEther Nice work! Thank you for hanging in there. This turned out being longer than expected.

@rocco8773 Thank you for taking the time to make suggestions! :) 

@StanczakDominik Thank you for all your help getting me started with the PR and guiding me through the CI checks. :)

 I'm looking forward to making more contributions to this project wherever I can. :)

StanczakDominik:
Thanks for doing this @DarkAEther :)

If you believe it's fine @rocco8773 it's fine for me as well! :)

PR #850, opened 2020-06-09
Update parameters.thermal_speed 

### LINKED ISSUE #427: 1D Maxwellian example update
The 1D Maxwellian [example](http://docs.plasmapy.org/en/stable/auto_examples/plot_distribution.html#sphx-glr-auto-examples-plot-distribution-py) in our docs should have a plot for 1D speed distribution with lines that demarcate the various thermal velocities (i.e. "most probable", "mean magnitude", etc.).

pheuer:
Scrolling through the issue log, I saw #427 and thought it was an easy addition. However, while making the example I thought that since 1D, 2D, and 3D Maxwellians are included in the distribution.py module, thermal speeds for different dimensionalities should also be supported (if for no other reason than to remind people to use the right one). Does everyone agree this is worthwhile functionality?

In this PR I've added this functionality and wrote appropriate tests. I also created a thermal_speed example notebook that contains the plot originally requested in #427 and have added this example to the documentation.

Closes #427


docs/formulary/distribution.rst
```diff
@@ -13,3 +13,4 @@ Distribution functions (`plasmapy.formulary.distribution`)
     :caption: Examples
 
     /notebooks/distribution
+	/notebooks/thermal_speed
```
StanczakDominik, line 16: Looks like a mixed space-tab issue is causing this to throw warnings...
```suggestion
    /notebooks/thermal_speed 
```

docs/formulary/parameters.rst
```diff
@@ -11,3 +11,4 @@ Plasma parameters (`plasmapy.formulary.parameters`)
     :caption: Examples
 
     /notebooks/distribution
+	/notebooks/thermal_speed
```
StanczakDominik, line 14:
```suggestion
    /notebooks/thermal_speed 
```

plasmapy/formulary/parameters.py
```diff
@@ -478,10 +483,34 @@ def thermal_speed(
...
+         - 1
+         - :math:`2/\pi`
+       * - 2D
+         - 1
+         - 2
+         - :math:`\pi/2`
+       * - 3D
+         - 2
+         - 3
+         - :math:`8/\pi`
```
StanczakDominik, line 511: This is going to be dreadful to read in jupyter/ipython/other shells, but wow, in the rendered docs that is one pretty table. :smile: I'm completely down for that.

plasmapy/formulary/parameters.py
```diff
@@ -507,15 +536,43 @@ def thermal_speed(
     """
     m = mass if np.isfinite(mass) else particles.particle_mass(particle)
 
+    # TODO: Could implement this more cleanly for any integer ndim
+    # at the cost of using the gamma function in the expression for the
+    # mean magnitude velocity.
+
```
StanczakDominik, line 542: Sure, but it'd be slower, too.

plasmapy/formulary/parameters.py
```diff
@@ -507,15 +536,43 @@ def thermal_speed(
...
+        elif method == "rms":
+            V = np.sqrt(3 * k_B * T / m)
+        elif method == "mean_magnitude":
+            V = np.sqrt(8 * k_B * T / (m * np.pi))
+        else:
+            raise ValueError("Method {method} not supported in thermal_speed")
     else:
-        raise ValueError("Method {method} not supported in thermal_speed")
+        raise ValueError("{ndim} is not a supported value for ndim in "
+                         "thermal_speed")
```
StanczakDominik, line 575: Actually, you know what a nice refactor could be? A dictionary!

```python
coefficients = {1: {"most_probable": 0, ...}, ...}
try:
   coefficient = coefficients[ndim][method]
except KeyError:
    raise ValueError("Method {method} not supported in thermal_speed") 
    # could even break this down into two dict-indexing steps to handle both the exceptions
V = np.sqrt(coefficient * k_B * T / m)
```

The dictionary could even be created outside the function to save some completely unnoticeable amount of time on instantiating it on each call, if we care about that.

plasmapy/formulary/parameters.py
pheuer: Yeah - one advantage of calculating these with an explicit dependence on ndim (rather than the if/else tree) would be the ability to more directly state the formulas in the docstring simply for all ndim. 

Side question: how do you get to the docs for a PR (without typing out the link and changing the PR number)?

plasmapy/formulary/parameters.py
pheuer: Fair enough - 3D plasmas are more than complicated enough for me.

plasmapy/formulary/parameters.py
pheuer: Well that's a much, much better way to do that! Done.

plasmapy/formulary/parameters.py
StanczakDominik: If you look for `docs/readthedocs.org:plasmapy` in the `Checks` tab below, you can click on Details and that should point you to the built docs :)

pheuer: Awesome, thanks!

pheuer:
@StanczakDominik Thanks for the suggestions! I've implemented the dictionary refactor: that's very elegant. A few other changes: 

- I ran pre-commit and it worked well (once I updated the fork to the current master: before that it wanted to fix lots of files). 

- I updated the example notebook to have a thumbnail. I ran into the same problem here as I did with the Thomson notebook where multi-figure plots are required but make bad thumbnails, so a single plot must also be included. Maybe theres a better way to handle this through a more sophisticated use of nbsphinx?

StanczakDominik:
> I updated the example notebook to have a thumbnail. I ran into the same problem here as I did with the Thomson notebook where multi-figure plots are required but make bad thumbnails, so a single plot must also be included. Maybe theres a better way to handle this through a more sophisticated use of nbsphinx?

I don't think there is, or at least I'm not aware of it yet :(

 I'll give this a quick once-over again and merge it, most likely :)

plasmapy/formulary/parameters.py
```diff
@@ -412,6 +412,16 @@ def ion_sound_speed(
     return V_S
 
 
+# This dictionary defines coefficients for thermal speeds
+# calculated for different methods and values of ndim.
+# Created here to avoid re-instantiating on each call
+coefficients = {
```
StanczakDominik, line 418: I think I'd like to make these a private variable:
```suggestion
_coefficients = {
```

The underscore is sort of a convention saying "you probably shouldn't need to use this variable outside", since Python lacks `private`s.

plasmapy/formulary/parameters.py
```diff
@@ -509,16 +548,16 @@ def thermal_speed(
...
     # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
-    if method == "most_probable":
-        V = np.sqrt(2 * k_B * T / m)
-    elif method == "rms":
-        V = np.sqrt(3 * k_B * T / m)
-    elif method == "mean_magnitude":
-        V = np.sqrt(8 * k_B * T / (m * np.pi))
-    else:
+    try:
+        coef = coefficients[ndim]
```
StanczakDominik, line 552:
```suggestion
        coef = _coefficients[ndim] 
```

plasmapy/formulary/tests/test_parameters.py
```diff
@@ -365,6 +365,44 @@ def test_thermal_speed():
         thermal_speed(1 * u.MK, particle="p").si.value, 128486.56960876315
     )
 
+    # Explicitly check all three modes and dimensionalities
+    # ndim = 1
+    assert np.isclose(thermal_speed(T_e, method="most_probable", ndim=1).si.value, 0.0)
+
```
StanczakDominik, line 371:
```suggestion

    # Regression tests start here!
```
I'm bad at keeping track of which of our tests are known value comparisons (probably a minuscule amount) and which are just regression tests, so while I'm looking at this file, lemme add this note. This sort of thing hurt us a bit when astropy changed their current constant values. The implementation is fine, though.

pheuer:
@StanczakDominik  Looks good - most of the tests I've written for both this and the Thomson module have been known value comparisons: I can see the potential problems with this approach but I can't think of a better way to ensure the physics equations haven't been modified?

StanczakDominik:
> @StanczakDominik  Looks good - most of the tests I've written for both this and the Thomson module have been known value comparisons: I can see the potential problems with this approach but I can't think of a better way to ensure the physics equations haven't been modified?

Me neither, unless the original derivation provided value samples... Which they usually don't. Oh well.

StanczakDominik:
We could potentially try property based testing one of these days... I'd have to pick it up from scratch though.

No matter! Merging :)

PR #846, opened 2020-06-08
Fix issues arising from github actions

StanczakDominik:
- [x] I have added a changelog entry for this pull request.

- [ ] If adding new functionality, I have added tests and
      docstrings.

- [ ] I have fixed any newly failing tests.

PR #857, opened 2020-06-20
Restore isort to pre-commit check; apply to entire codebase

StanczakDominik:
#846 removed isort from the pre-commit suite because it didn't play nicely
with CI. Well, according to
https://github.com/timothycrosley/isort/issues/1147#issuecomment-646905036
isort will work well now. Let's see if that works.

- [x] This should probably not get merged yet; let's give isort a few days to put
a release out.

I also brought back the 88 line limit for consistency between the two
(it got the suite stuck in a permanent 2-cycle of 2 character line length
changes) - it turns out it's the default
in `black`, and I figure if Łukasz Langa believes in that default,
it's a sane default.

I also added a file that lets you ignore isort/black commits in git blame locally.

- [ ] make sure the commit hash for this PR is correct upon squashing; this will take a manual local merge.

### DIFF
--- a/plasmapy/__init__.py
+++ b/plasmapy/__init__.py
@@ -28,14 +28,7 @@ if sys.version_info < tuple((int(val) for val in "3.6".split("."))):
 # ----------------------------------------------------------------------------
 import pkg_resources
 
-from plasmapy import (
-    diagnostics,
-    formulary,
-    particles,
-    plasma,
-    simulation,
-    utils,
-)
+from plasmapy import diagnostics, formulary, particles, plasma, simulation, utils
 
 # define version
 try:
@@ -100,8 +93,9 @@ def online_help(query):
     query : str
         The search query.
     """
-    from urllib.parse import urlencode
     import webbrowser
+
+    from urllib.parse import urlencode
 
     url = (
         "http://docs.plasmapy.org/en/stable/search.html?"
--- a/plasmapy/particles/tests/test_parsing.py
+++ b/plasmapy/particles/tests/test_parsing.py
@@ -6,8 +6,7 @@ from plasmapy.particles.exceptions import (
     InvalidElementError,
     InvalidParticleError,
 )
-from plasmapy.particles.parsing import (
-    # duplicate with utils.pytest_helpers.error_messages.call_string?
+from plasmapy.particles.parsing import (  # duplicate with utils.pytest_helpers.error_messages.call_string?
     _case_insensitive_aliases,
     _case_sensitive_aliases,
     _dealias_particle_aliases,
--- a/setup.cfg
+++ b/setup.cfg
@@ -93,7 +93,7 @@ addopts = --doctest-modules --doctest-continue-on-failure --ignore=docs/conf.py
 # E902 - IOError
 # select = E226,E241,E242,E704,W504
 exclude = version.py,build
-max-line-length = 90
+max-line-length = 88
 
 [pydocstyle]
 # D302 is unnecessary as we are using Python 3.6+. Ignoring D202 allows blank
@@ -115,7 +115,7 @@ convention = numpy
 select = D402,D413
 ignore = D202,D205,D302,D400,D105,D107,F401
 exclude = extern,sphinx,*parsetab.py,conftest.py,docs/conf.py,setup.py
-max-line-length = 90
+max-line-length = 88
 
 [coverage:run]
 omit =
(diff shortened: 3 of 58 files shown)


plasmapy/diagnostics/langmuir.py
```diff
@@ -17,17 +17,15 @@
...
-import astropy.constants.si as const
 import numpy as np
 from astropy import units as u
+from astropy.constants import si as const
 from astropy.visualization import quantity_support
-from scipy.optimize import curve_fit
-
 from plasmapy.particles import Particle
 from plasmapy.utils.decorators import validate_quantities
+from scipy.optimize import curve_fit
```
StanczakDominik, line 28: More tinkering with config 

- [ ] separate plasmapy imports by newline

StanczakDominik:
I'll have to try this again now that isort 5 is out

StanczakDominik:
Okay. I updated isort to 5.2.2 (the current stable release), merged in current master and #878 because that's soon to be merged, then applied the full pre-commit suite to all of those changes. The commit to really look through here would thus be 18b60ba . I think the changes it applies look fine now.

plasmapy/__init__.py
```diff
@@ -28,14 +28,7 @@
...
 
-from plasmapy import (
-    diagnostics,
-    formulary,
-    particles,
-    plasma,
-    simulation,
-    utils,
-)
+from plasmapy import diagnostics, formulary, particles, plasma, simulation, utils
```
StanczakDominik, line 31: This looks a bit weird but I guess it's just below the line length limit, and I'm not going to argue with autotools on that point.

rocco8773: Agreed! I guess it's practice in bowing down to our robotic overlords.

StanczakDominik:
Oh right, squash merges do that.

I'll rebase later.

StanczakDominik:
Okay, fixed the conflicts!

setup.cfg
```diff
@@ -132,6 +132,8 @@ exclude_lines =
 
 [isort]
 # Set sorting of imports to be consistent with black formatting
-line_length=90
 multi_line_output=3
-include_trailing_comma: True
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=88
```
rocco8773, line 139: After playing around with `isort`, I'm really excited to have this implemented!!

According to `isort`, the [preferred configuration files are `.isort.cfg` or `pyproject.toml`](https://timothycrosley.github.io/isort/docs/configuration/config_files/).  Since we already have a `pyproject.toml`, I say we migrate the configuration to that file.

After playing with the `isort` options, I think this would be the closes to what we've been implementing so far...

``` toml
[tool.isort]
line_length = 88
wrap_length = 80
sections = ["FUTURE", "STDLIB", "FIRSTPARTY", "LOCALFOLDER"]
known_first_party = ["plasmapy", ]
default_section = "STDLIB"
multi_line_output = 3
force_single_line = true
use_parentheses = true
include_trailing_comma = true
force_alphabetical_sort_within_sections = true
lines_between_types = 1
```

Note, this will force third party packages and standard library packages into the same import section.  That is how we've been organizing imports thus far.  If we want to split up standard library and third party imports, then that's fine too.  I just figure we stick with what we've been doing.

StanczakDominik: I don't really see that as something worth setting up custom settings for - I'd be happy to go for isort defaults. But since you've gone to the trouble of doing it, and you likely have Opinions about this, I'll add this config and rerun the full isort pass.

StanczakDominik:
@rocco8773 I've applied the changes. If they look okay, hit approve - but please don't merge, I'll do it manually locally to get .git-blame-ignore-revs right in a single pass.

plasmapy/particles/tests/test_parsing.py
```diff
@@ -6,8 +6,7 @@
     InvalidElementError,
     InvalidParticleError,
 )
-from plasmapy.particles.parsing import (
-    # duplicate with utils.pytest_helpers.error_messages.call_string?
+from plasmapy.particles.parsing import (  # duplicate with utils.pytest_helpers.error_messages.call_string?
```
rocco8773, line 9: This is odd.  I've confirmed `isort` is placing the comment here, which then violates PEP8 max line-length.  I've played with the configuration options, but I couldn't find a solution.  Maybe run `black` after `isort`.

In the end, this comment does not make much sense to me and can probably be deleted.

StanczakDominik: Okay, I'll remove that one. Anything else that needs changing?

rocco8773: Nope, everything else looks good to me.

PR #859, opened 2020-06-23
Create abstract base class as interface for plasma normalizations

### LINKED ISSUE #808: Create classes to represent normalizations in simulation subpackage
Numerical simulations of plasma phenomena are frequently normalized to characteristic values so that the corresponding equations and fields become dimensionless.  A strategy for keeping track of normalizations is therefore a necessary building block for a simulation framework.  At present, there is no standard method keeping track of normalizations.

Moreover, the normalizations for fluid and particle-in-cell simulations are often done differently.  For example, time in a particle-in-cell simulation is often normalized to the ion cyclotron timescale and length is often normalized to the ion inertial length.  Length and time in a fluid simulation are often normalized to arbitrary values, but typically related such that the Alfvén velocity is one.

It would be helpful to have a standard way of saving normalizations outside of the package.  This could be done for example through a YAML, TOML, or HDF5 file, though there should be an open metadata standard for it.  

One headache when dealing with normalizations is figuring out the normalizations for things like diffusion coefficients or viscosities, which are derived from the primary normalizations (e.g., magnetic field, etc.).  It would be helpful for the normalization class (or classes) to have a method that takes a set of units that is derived from the base units, and then return the normalization for quantities measured in those units.  For example, if we want to return a value that is in units of `kg * m ** 2 / s ** 2`, we would pass in that set of units, and get a `Quantity` with those units that corresponds to the normalization.  We'd have to be careful for this though, and make sure there aren't any extra factors of two that we would be missing.  This wouldn't necessarily need to be created in the first pull request, and could instead be raised as its own issue if we decide to go in that direction.

There are some straightforward cases that could be put in analytically, like the typical normalizations for the MHD equations.

namurphy:
This pull request includes new classes that describe standard normalizations that are used for ideal and extended MHD.  The purpose of these classes is to have a standard way to represent normalization information that can be used for different codes and groups.  Actually, I would have found these classes to be really helpful for a simulation benchmark project that I'm working on so that I didn't need to create them from scratch.  I'm most familiar with the fluid approximation normalizations so I'm starting with those, but we'll also need to create normalizations for PIC simulation (where length is often normalized to the ion inertial length and time to the inverse of the ion gyrofrequency).  I still have a lot to do for this, so it's lower priority for code review at the moment.

      docstrings.

Closes #808.


StanczakDominik:
This pull request has been mentioned on **PlasmaPy Community Discourse Forum**. There might be relevant details there:

https://plasmapy.discourse.group/t/ideas-for-how-we-can-use-normalizations-objects/44/1

namurphy:
For this pull request, I'm thinking that I should focus on getting the normalizations in there in the proper form with sufficient documentation.  It'll probably be best to have follow-up pull requests that deal with how an object containing normalizations can be used in operations.  Best to limit the scope of pull requests, as we've learned...

- [ ] Create an issue on dunder operations for normalizations objects after this pull request is merged.

StanczakDominik:
> This pull request has been mentioned on **PlasmaPy Community Discourse Forum**. There might be relevant details there:
> 
> [plasmapy.discourse.group/t/ideas-for-how-we-can-use-normalizations-objects/44/1](https://plasmapy.discourse.group/t/ideas-for-how-we-can-use-normalizations-objects/44/1)

I would just like to note that this was an automation that I forgot I set up and holy wow, did I just give myself a brief moment of existential doubt.

namurphy:
I decided to downscope this pull request.  This PR now creates a first attempt at an abstract base class, `AbstractNormalizations`, that is to be used in the creation of classes to represent normalizations for plasmas.  This PR does not include tests since this is the first draft, and because I'm not sure of a meaningful way of testing abstract classes.  I'm ready for this PR to be merged when it's ready.

I'm thinking it would be really helpful to have a function that takes a unit and returns a list of corresponding physical types, and a function that takes a physical type and then returns the appropriate SI unit.  I'd probably use this functionality a lot since I often have to figure out what units are for a particular quantity (like thermal conduction).  This is related to what I brought up in https://github.com/astropy/astropy/issues/11202.

In any case, I'm still thinking about the best strategy for creating classes like `MHDNormalizations`, and I think it'll be better to spread the creation of new machinery out over several pull requests.

plasmapy/simulation/abstractions.py
```diff
@@ -54,3 +54,142 @@ class AbstractTimeDependentSimulation(AbstractSimulation):
...
+    Warnings
+    --------
+    This interface is unstable and is subject to change.
+    """
+
+    @property
+    @abstractmethod
+    def current_density(self) -> u.A / u.m ** 2:
+        """The current density normalization."""
+        pass
```
rocco8773, line 73: Just a friendly reminder that we decided to go with Python's standard of using `...` with abstract methods. :)

```suggestion
    @property
    @abstractmethod
    def current_density(self) -> u.A / u.m ** 2:
        """The current density normalization."""
        ...
```

plasmapy/simulation/abstractions.py
```diff
@@ -54,3 +54,142 @@ class AbstractTimeDependentSimulation(AbstractSimulation):
...
+
+
+class AbstractNormalizations(ABC):
+    """
+    An abstract base class to represent the normalizations of systems of
+    equations describing plasmas.
+
+    Warnings
+    --------
+    This interface is unstable and is subject to change.
```
rocco8773, line 66: Might be better to use the `reST` [warning admonition](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions) here so this does not get confused with `numpydoc`'s warning documentation.  This should work, but I've never done it like this before.

```suggestion
    .. warning::
       This interface is unstable and is subject to change.
```

plasmapy/simulation/abstractions.py
```diff
@@ -54,3 +54,142 @@ class AbstractTimeDependentSimulation(AbstractSimulation):
...
+    """
+
+    @property
+    @abstractmethod
+    def current_density(self) -> u.A / u.m ** 2:
```
rocco8773, line 71: Are these methods suppose to be the normalized value, un-normalized value, or the the normalizing constant?

plasmapy/simulation/abstractions.py
namurphy: Good idea!  One thing I read in the numpydoc docs is that they have sections named "Warns" which is for warnings that are issued, and "Warnings" which are intended for readers.

plasmapy/simulation/abstractions.py
namurphy: These are to be the normalizing constant, which I've most often heard referred to as the normalization.  Maybe I should clarify that in the class docstring...

plasmapy/simulation/abstractions.py
namurphy: And if I remember correctly, it was me who suggested that, wasn't it...  :grimacing:

namurphy:
Also, you should be proud that I alphabetized the attributes contained in `AbstractNormalizations`.  :1st_place_medal: :100: :mortar_board: :abc:

rocco8773:
> Also, you should be proud that I alphabetized the attributes contained in `AbstractNormalizations`. 🥇 💯 🎓 🔤

It is beautifully organized! 🔤🤣

plasmapy/simulation/abstractions.py
rocco8773: It was.  I tried to find the old comment thread, but no such luck. 🤣

plasmapy/simulation/abstractions.py
```diff
@@ -2,9 +2,13 @@
 
 __all__ = ["AbstractSimulation", "AbstractTimeDependentSimulation"]
```
rocco8773, line 3: Add `AbstractNormalizations` to `__all__`?

plasmapy/simulation/abstractions.py
rocco8773: I did not know that.  I wonder if it renders like  the reST admonition. 🤔  I went to check the old RTD builds, but it wasn't rendered because `AbstractNormalizations` was not included in the `__all__`.

rocco8773:
Something to think about for the future... This feels like a [`dataclass`](https://docs.python.org/3/library/dataclasses.html) to me and, I'm guessing, that is how it will be predominately used.  I'm assuming it's basically a warehouse for normalization constants that are then used to normalize (`array / norm.length`) or de-normalize (`array * norm.length`) physical parameters.

changelog/859.feature.rst
```diff
@@ -0,0 +1,2 @@
+Created the `~plasmapy.simulation.normalizations.AbstractNormalizations` class
+to serve as an abstract interface for future classes that represent normalizations.
```
rocco8773, line 2:
```suggestion
Created the `~plasmapy.simulation.abstractions.AbstractNormalizations` class
to serve as an abstract interface for future classes that represent normalizations.
```

plasmapy/simulation/abstractions.py
```diff
@@ -53,4 +57,142 @@ class AbstractTimeDependentSimulation(AbstractSimulation):
...
+        ...
+
+    @property
+    @abstractmethod
+    def ion(self) -> Particle:
+        """
+        The ion in the plasma, which could be a positron in the case
+        of a pair plasma.
+        """
+        ...
```
rocco8773, line 109: An `ion` isn't really a normalization.  It's mass could be a normalization, or it's gyrofrequency could be.  I would remove this method and leave it up to the user to manually define/calculate the other normalizations.

namurphy: Yeah...I've been wondering how to handle ions in the normalization, since how to handle ion stuff make normalization stuff kind of messy...in particular for how we handle multi-component plasmas.  Ultimately I'd like for there to be normalization classes that allow operations like:
```Python
>>> dimensionless_plasma = dimensional_plasma / normalizations
>>> dimensional_plasma = normalizations * dimensionless_plasma
```
For these to work we'd need information on the ion(s) somehow, so the choice of ion(s) do end up weirdly more or less being part of the normalization...in particular for the normalization between number density and mass density.  I'll remove the `ion` attribute for now, and worry about how to deal with this next year.

ugh why won't 2016 just end already

plasmapy/simulation/abstractions.py
```diff
@@ -53,4 +57,142 @@ class AbstractTimeDependentSimulation(AbstractSimulation):
...
+    ...
+
+
+class AbstractNormalizations(ABC):
+    """
+    An abstract base class to represent the normalization constants for
+    systems of equations describing plasmas.
+
+    ... warning::
+        This interface is unstable and is subject to change.
```
rocco8773, line 67: Oops! My suggestion had an extra dot in it. 🤦🏻‍♂️

```suggestion
    .. warning::
       This interface is unstable and is subject to change.
```

plasmapy/simulation/abstractions.py
rocco8773: > ugh why won't 2016 just end already

I know! Longest year eveeeeeeeeeer!  Completely feels like 4+ years.

---

Well, that will be an interesting implementation.  I like it, but I have no clue how to go about implementing it.  Might be easier to have something like `dimensionless_plasma.reform(normalizations)` and `dimensional_plasma.reform(normalizations)`.  I only used `reform` to keep the name consistent across all plasma classes.

rocco8773 (approved):
The RTD finally finished and it looks good.  I'm curious where this functionality goes and how its UI develops.  I can see it being useful in several places.

PR #864, opened 2020-07-03
Shocks

KhalilBryant:

      docstrings.


KhalilBryant:
I want to add to the tests that ensure that when ds is negative, an error occurs and tells you that entropy change can't be negative, but I'm not sure how to go about it

StanczakDominik:
[`pytest.raises`!](https://docs.pytest.org/en/stable/reference.html#pytest-raises)

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,92 @@
...
+
+validate_quantities(
+    rho_1={"can_be_negative": False},
+    rho_2={"can_be_negative": False},
+    ds={"can_be_negative": False},
+    c_v={"can_be_negative": False},
+)
+
+
+def entropy_across_shock_polytropic(
```
rocco8773, line 22:
```suggestion
@validate_quantities(
    rho_1={"can_be_negative": False},
    rho_2={"can_be_negative": False},
    ds={"can_be_negative": False},
    c_v={"can_be_negative": False},
)
def entropy_across_shock_polytropic(
```

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    pressure_ratio = p_2 / p_1
+    density_ratio = rho_1 / rho_2
+
+    ds = c_v * log(pressure_ratio * density_ratio ** gamma)
+
+    if ds < 0:
+        raise Exception("Entropy change cannot be negative")
+    else:
+        return ds
```
StanczakDominik, line 86: 1. TODO me: make this numpy capable
2. if ds < 0: warn with PhysicsWarning that the regions are swapped, return `u.Quantity(np.nan, unit=c_v.unit)`

StanczakDominik: `warnings.warn(PhysicsError, "Entropy does not decrease lol")`

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,104 @@
...
+    if ds < 0:
+        warnings.warn(
+            "Entropy change cannot be negative, perhaps regions 1 and 2 are switched",
+            PhysicsWarning,
+        )
+        return u.Quantity(nan, unit=c_v.unit)
+
+    elif isclose(ds, 0, rtol=1e-4, atol=1e-8) is True:
+        warnings.warn(
+            "Irreversible process, entropy change cannot be 0", PhysicsWarning
```
StanczakDominik, line 95: I would suggest being very, very explicit here about what's happening. As it is, this warning is not particularly helpful, it doesn't explain why we're mangling the user's output and how to stop it. Multiline strings are your friend!

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,106 @@
...
+    check_relativistic,
+    validate_quantities,
+)
+from astropy.constants import c, a0, k_B
+from numpy import pi, exp, sqrt, log, nan, isclose
+
+
+@validate_quantities(
+    rho_1={"can_be_negative": False},
+    rho_2={"can_be_negative": False},
```
rocco8773, line 20: Since you're immediately making ratios from `p_1`, `p_2`, `rho_1`, and  `rho_2`, this is a good instance to use `pass_equivalent_units=True`.  This will keep your function from doing 2 extra conversions.  This is not a big deal for `p`, but can be for `rho` because the the conversion to `u.m ** -3` can cause non-negligible floating point errors.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,106 @@
...
+    >>> p_1 = 101325 * u.Pa
+    >>> p_2 = 2 * p_1
+    >>> rho_1 = 2 * u.kg / u.m ** 3
+    >>> rho_2 = 3 * u.kg / u.m ** 3
+    >>> entropy_across_shock_polytropic(c_v, p_1, p_2, rho_1, rho_2, gamma)
+    <Quantity 0.2605... J / K>
+
+    """
+    pressure_ratio = p_2 / p_1
+    density_ratio = rho_1 / rho_2
```
rocco8773, line 82:
```suggestion
    pressure_ratio = (p_2 / p_1).to(u.dimensionless_unscaled)
    density_ratio = (rho_1 / rho_2).to(u.dimensionless_unscaled)
```

This should be done assuming you implement `pass_equivalent_units=True`.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,106 @@
...
+            irreversible, so the entropy change must be nonzero.""", PhysicsWarning
+        )
+        return u.Quantity(nan, unit=c_v.unit)
+
+    else:
+        return ds
+
+
+ds_ = entropy_across_shock_polytropic
+"""alias for :func:`entropy_across_shock_polytropic`"""
```
rocco8773, line 106: No offense to `entropy_across_shock_polytropic()`, but I'm not sure it's common enough to deserve an alias.  If it does get an aliases, then it needs to be more explicit than `ds_`.  `ds_` could be confused for inertial length of species `s`.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,106 @@
...
+    ds = c_v * log(pressure_ratio * density_ratio ** gamma)
+
+    if ds < 0:
+        warnings.warn(
+            "Entropy change cannot be negative, perhaps regions 1 and 2 are switched",
+            PhysicsWarning,
+        )
+        return u.Quantity(nan, unit=c_v.unit)
+
+    elif isclose(ds, 0, rtol=1e-4, atol=1e-8):
```
rocco8773, line 93: I'd actually put this as your first if statement.  If the value is close to zero, then it probably has a +/- floating point "fuzz" on it from floating point rounding errors (from the calculation).  I'd do an if-statement like...

```python
np.isclose(np.abs(ds), 0.0, rtol=1e-4, atol=1e08)
```

`ds < 0` would then be the 2nd clause in the if-statement

plasmapy/formulary/shocks.py
StanczakDominik: I agree that this seems niche enough not to have an alias, though obviously I don't know much about the field itself.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,103 @@
...
+    Entropy is not conserved across a shock, since a shock is
+    an irreversible, non-adiabatic process. This equation is based on the specific entropy
+    of a polytropic gas before and after the shock has passed.
+
+     .. math::
+
+        s_2 - s_1 =
+        c_v ln\left[ \frac{p_2}{p_1} \left( \frac{\rho_1}{\rho_2} \right)^{\gamma} \right]
+    Here variables indexed by 1 and 2 are upstream (pre shock)
+    and downstream (post shock) respectively.
```
rocco8773, line 41: If I remember correctly, this is derived from a non-magnetic, two-fluid approach.  What I'm trying to get at, you should put a little backstory into the docstring so users know the equations assumptions and "valid" conditions.

Also, don't forget to cite reference material.

plasmapy/formulary/shocks.py
KhalilBryant: do I just use `pass_equivalent_units=True` in `@validate_quantities`?

rocco8773: You add the key-value pair to the dictionary.  There's an example in the [`validate_quantities` documentation](https://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.validate_quantities.html#plasmapy.utils.decorators.validate_quantities).

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@validate_quantities(
+    rho_1={"can_be_negative": False, "pass_equivalent_units": True},
+    rho_2={"can_be_negative": False, "pass_equivalent_units": True},
+    p_1={"can_be_negative": True, "pass_equivalent_units": True},
+    p_2={"can_be_negative": True, "pass_equivalent_units": True},
+    c_v={"can_be_negative": False},
+)
+def entropy_across_shock_polytropic(
```
rocco8773, line 25: This is a hard function to name!  While current name is okay, I still feel it's a little vague.  How about something like `entropy_jump_polytropic`, `entropy_jump_polytropic_shock`, `entropy_jump_polytropic_gas_shock`, etc.  I moved to the "jump" nomenclature, because in the shock world the jump conditions refers to the relations between the pre- and post-shock.  I think I'm leaning towards `entropy_jump_polytropic_gas_shock`, but there's probably still room for improvement.

KhalilBryant:
anyone know how I can get rid of these AttributeError: __enter__ messages I get?

rocco8773:
> anyone know how I can get rid of these `AttributeError: __enter__` messages I get?

You're using `warnings.warn` instead of `pytest.warns` in your tests.

StanczakDominik:
Looks okay to me, but that doc build error looks weird. The langmuir notebook should not be affected by this PR. I'm trying to replicate it locally.

StanczakDominik:
We can ignore the `test_documentation` warning for this PR; it also happens on master right now. @KhalilBryant, @rocco8773, anything more to add here? Otherwise we could probably merge :)

rocco8773:
@StanczakDominik Not yet, the `shock` package still needs to be added to the formulary documentation.

@KhalilBryant Once you remove the draft mark on this PR, I'll give it a closer review.

KhalilBryant:
> @StanczakDominik Not yet, the `shock` package still needs to be added to the formulary documentation.
> 
> @KhalilBryant Once you remove the draft mark on this PR, I'll give it a closer review.

How exactly do I do that again? Sorry, I forgot. As soon as I do that I can take the draft mark off

rocco8773:
@KhalilBryant You'll need to add a `shocks.rst` to the [`./docs/formulary`](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/formulary) directory, which you can probably model after the [`ionization.rst`](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/formulary/ionization.rst).  After that you'll need to make references to the `shocks.rst` in the [`./docs/formulary/index.rst`](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/formulary/index.rst) file, both in the [`toctree` directive](https://github.com/PlasmaPy/PlasmaPy/blame/master/docs/formulary/index.rst#L12-L28) and the [`table`](https://github.com/PlasmaPy/PlasmaPy/blame/master/docs/formulary/index.rst#L30-L59).

KhalilBryant:
> @KhalilBryant You'll need to add a `shocks.rst` to the [`./docs/formulary`](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/formulary) directory, which you can probably model after the [`ionization.rst`](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/formulary/ionization.rst). After that you'll need to make references to the `shocks.rst` in the [`./docs/formulary/index.rst`](https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/formulary/index.rst) file, both in the [`toctree` directive](https://github.com/PlasmaPy/PlasmaPy/blame/master/docs/formulary/index.rst#L12-L28) and the [`table`](https://github.com/PlasmaPy/PlasmaPy/blame/master/docs/formulary/index.rst#L30-L59).

How do I add it to the table? I can't seem to find it by going to `plasmapy/docs/index.rst`

rocco8773:
You need to merge the current `master` into your branch. The table was added the the recent PR #867.

plasmapy/formulary/tests/test_shocks.py
```diff
@@ -0,0 +1,27 @@
+import warnings
+import astropy.units as u
+import numpy as np
+import pytest
+from astropy.constants import c
```
rocco8773, line 5:
```suggestion
import astropy.units as u
import numpy as np
import pytest
```

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@validate_quantities(
+    rho_1={"can_be_negative": False, "pass_equivalent_units": True},
+    rho_2={"can_be_negative": False, "pass_equivalent_units": True},
+    p_1={"can_be_negative": True, "pass_equivalent_units": True},
+    p_2={"can_be_negative": True, "pass_equivalent_units": True},
+    c_v={"can_be_negative": False},
+)
+def entropy_jump_polytropic(
```
rocco8773, line 25: I have problem with this name, it's not explicit enough when imported from the `fomulary` module.  That is, this makes sense...

```python
from plasmapy.formulary import shocks

val = shocks.entropy_jump_polytropic(...)
```
but this does not make sense
```python
from plasmapy.formulary import entropy_jump_polytropic

val = entropy_jump_polytropic(...)
```

In the later case, we have no indication that the function is purposefully for shocks.  I see three possible solutions...

1. Rename the function to be more explicit. (e.g `entropy_jumpy_polytropic_gas_shock`)
1. Keep the function name  as is, but give it an more explicit name when imported into `plasmapy/formulary/__init__.py`
1. Only expose `shocks` to the formulary namespace. (i.e. you would have to import like `from plasmapy.formulary.shocks import entropy_jumpy_polytropic`)

My opinion, option 3 is the best choice.  @StanczakDominik @namurphy do you have any thoughts about this?

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+    of a polytropic gas before and after the shock has passed. The shock
+    is an isolated steady, perpendicular shock, flowing through a non magnetized fluid.
+
+     .. math::
+
+        s_2 - s_1 =
+        c_v ln\left[ \frac{p_2}{p_1} \left( \frac{\rho_1}{\rho_2} \right)^{\gamma} \right]
+    Here variables indexed by 1 and 2 are upstream (pre shock)
+    and downstream (post shock) respectively.
+    This function is equivalent to Eq. 4.24 in `Drake`_.
```
rocco8773, line 45: This needs rewording.  For example, there is no magnetic field in this shock formulation so there's no sense of it being perpendicular or parallel.  Additionally, the derivation uses the assumption of a planar shock.  ...I'll need to think about this more, so I'll get back to you with additional thoughts on verbiage.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+    Entropy is not conserved across a shock, since a shock is
+    an irreversible, non-adiabatic process. This equation is based on the specific entropy
+    of a polytropic gas before and after the shock has passed. The shock
+    is an isolated steady, perpendicular shock, flowing through a non magnetized fluid.
+
+     .. math::
+
+        s_2 - s_1 =
+        c_v ln\left[ \frac{p_2}{p_1} \left( \frac{\rho_1}{\rho_2} \right)^{\gamma} \right]
+    Here variables indexed by 1 and 2 are upstream (pre shock)
```
rocco8773, line 43:
```suggestion
    is an isolated steady, perpendicular shock, flowing through a non magnetized fluid.

    .. math::

         s_2 - s_1 = c_v
                     \ln{\left[
                         \frac{p_2}{p_1}
                         \left( \frac{\rho_1}{\rho_2} \right)^{\gamma}
                     \right]}

    Here variables indexed by 1 and 2 are upstream (pre shock)
```

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+    gamma : `float`
+        Polytropic index of the gas.
+
+    Returns
+    -------
+    ds: `~astropy.units.Quantity`
+        The change of entropy due to the shock on the gas.
+
+     Examples
+    --------
```
rocco8773, line 73:
```suggestion
    Examples
    --------
```

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,116 @@
...
+    >>> gamma = 5 / 3 #ideal gas approx
+    >>> p_1 = 101325 * u.Pa
+    >>> p_2 = 2 * p_1
+    >>> rho_1 = 2 * u.kg / u.m ** 3
+    >>> rho_2 = 3 * u.kg / u.m ** 3
+    >>> entropy_jump_polytropic(c_v, p_1, p_2, rho_1, rho_2, gamma)
+    <Quantity 0.2605... J / K>
+
+     Notes
+    -----
```
rocco8773, line 85:
```suggestion
    Notes
    -----
```

plasmapy/formulary/tests/test_shocks.py
```diff
@@ -0,0 +1,27 @@
...
+    p_2 = 2 * p_1
+    rho_1 = 2 * u.kg / u.m ** 3
+    rho_2 = 3 * u.kg / u.m ** 3
+    val = entropy_jump_polytropic(c_v, p_1, p_2, rho_1, rho_2, gamma)
+
+    assert val.unit == u.J / u.K
+    assert np.isclose(val.value, 0.26058001)
+
+    with pytest.warns(PhysicsWarning):
+        entropy_jump_polytropic(c_v, p_1, 0.5 * p_1, rho_1, rho_2, gamma)
```
rocco8773, line 24:
```suggestion
    with pytest.warns(PhysicsWarning):
        # entropy jump can not be negative
        val = entropy_jump_polytropic(c_v, p_1, 0.5 * p_1, rho_1, rho_2, gamma)
        assert np.isnan(val)
```

plasmapy/formulary/tests/test_shocks.py
```diff
@@ -0,0 +1,27 @@
...
+
+    with pytest.warns(PhysicsWarning):
+        entropy_jump_polytropic(c_v, p_1, p_1, rho_1, rho_1, gamma)
```
rocco8773, line 27:
```suggestion
    with pytest.warns(PhysicsWarning):
        # entropy jump can not be zero (shocks are irreversible)
        val = entropy_jump_polytropic(c_v, p_1, p_1, rho_1, rho_1, gamma)
        assert np.isnan(val)
```

plasmapy/formulary/shocks.py
pheuer: @rocco8773 This is resolved now, right?

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+
+@validate_quantities(
+    rho_1={"can_be_negative": False, "pass_equivalent_units": True},
+    rho_2={"can_be_negative": False, "pass_equivalent_units": True},
+    p_1={"can_be_negative": True, "pass_equivalent_units": True},
+    p_2={"can_be_negative": True, "pass_equivalent_units": True},
+    c_v={"can_be_negative": False},
+)
+def entropy_jump_polytropic(
+    c_v: u.J / u.K,
```
pheuer, line 27: Personally I rarely know the specific heat of the plasmas I'm dealing with, so using this function would always require some extra leg work. I would definitely suggest substituting Eq. 4.26 in Drake (which is equivalent to Eq. 3.3) to eliminate c_v, then replacing this input with a `plasmapy.Particle`  instance that represents the ion species to get the atomic number and charge. You can then present Eq. 4.24 and Eq. 4.26 in the documentation to make clear what's happening.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+    p_1: u.Pa,
+    p_2: u.Pa,
+    rho_1: u.kg / u.m ** 3,
```
pheuer, line 30: The choice of mass density vs. number density is definitely field-specific: I think using rho is more common in ICF while in space physics/lab astro I see number density more. I'll just point out that if you do use a `Particle` input for the species, this _could_ be changed to number density and then converted to mass density within the function.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+    rho_2: u.kg / u.m ** 3,
+    gamma=5 / 3,
+) -> u.J / u.K:
+    r"""
+    Entropy is not conserved across a shock, since a shock is
+    an irreversible, non-adiabatic process. This equation is based on the specific entropy
+    of a polytropic gas before and after the shock has passed. The shock
```
pheuer, line 37: It wouldn't hurt to define polytropic here or somewhere (since this documentation is meant to be educational)

plasmapy/formulary/shocks.py
pheuer: Actually, maybe it's worth going one further: since this function only applies to shocks described by a certain model, maybe the namespace should be more like: `formulary.shocks.polytropic.entropy_jump`?

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+    For reference to this function and for more information regarding shocks and rarefactions,
+    see chapter 4 of R Paul Drake's book, "High-Energy-Density Physics: Foundation of
+    Inertial Fusion and Experimental Astrophysics" (`DOI: 10.1007/978-3-319-67711-8_3`_).
+
+    .. _`Drake`: https://doi.org/10.1007/978-3-319-67711-8
+    .. _`DOI: 10.1007/978-3-319-67711-8_3`: https://doi.org/10.1007/978-3-319-67711-8_3
+
+    """
+    pressure_ratio = (p_2 / p_1).to(u.dimensionless_unscaled)
+    density_ratio = (rho_1 / rho_2).to(u.dimensionless_unscaled)
```
pheuer, line 101: Since this only depends on the pressure and density ratios, you could potentially just take those ratios as arguments and then avoid potential confusion about units or the mass density vs. number density question I raised above.

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+    if isclose(abs(ds), 0.0, rtol=1e-4, atol=1e-8):
+        warn(
+            """Entropy change cannot be 0, as this would imply that
+            a shock is a reversible process. Shocks are
+            irreversible, so the entropy change must be nonzero.""",
+            PhysicsWarning,
+        )
+        return u.Quantity(nan, unit=c_v.unit)
+
+    elif ds < 0:
```
pheuer, line 114: You could check for this earlier by checking that the density and pressure ratios are > or < than 1 as expected. The benefit to doing so would be giving better error messages (you could tell the user which of the two they switched).

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+    angular_freq_to_hz,
+    check_relativistic,
+    validate_quantities,
+)
+
+
+@validate_quantities(
+    rho_1={"can_be_negative": False, "pass_equivalent_units": True},
+    rho_2={"can_be_negative": False, "pass_equivalent_units": True},
+    p_1={"can_be_negative": True, "pass_equivalent_units": True},
```
pheuer, line 22: Why can the pressures be negative?

pheuer (commented):
@KhalilBryant Thanks for starting this shocks package! Here's a quick review of this function.

plasmapy/formulary/shocks.py
rocco8773: I would keep `c_v` as is because people could use different methods to calculate it.  Instead, we should add a separate function (in a separate PR) to calculate `c_v`.  In the 2006 version of drake `c_v` is given in eqn 3.5.

plasmapy/formulary/shocks.py
rocco8773: If we're going that route, I'd go with `shocks.gas.entropy_jump_polytropic` since the expression is derived from a purely thermodynamic viewpoint.

This has always been one of areas of confusion on this topic, at least for me.  We have sets of equations that are derived from a purely thermodynamic viewpoint, which means the fluid used in the derivation is gas-like.  And then we have equations that take into consideration charges, currents, and fields, these are more representative of a plasma.  I think the field has always done a bad job of distinguishing the difference between these expressions, and that's why I'm so stuck up on the name here.  I'd like to take this opportunity to use `plasmapy` to help clarify this point and reduce the confusion in the community.

With that said, I still think the best name so far is `entropy_jumpy_polytropic_gas`.  @pheuer If you have arguments for or against this, then I'd love to hear them!?

plasmapy/formulary/shocks.py
rocco8773: This is a good point.  You can probably get way with something like...

```
    Entropy is not conserved across a shock, since a shock is
    an irreversible, non-adiabatic process. This equation is based on the specific
    entropy of a polytropic gas ( :math:`p \propto \rho^\gamma`) before and after
    the shock has passed. The shock is an isolated steady, perpendicular shock,
    flowing through a non magnetized, polytropic gas.
```

plasmapy/formulary/shocks.py
```diff
@@ -0,0 +1,122 @@
...
+
+         s_2 - s_1 = c_v
+                     \ln{\left[
+                         \frac{p_2}{p_1}
+                         \left( \frac{\rho_1}{\rho_2} \right)^{\gamma}
+                     \right]}
+
+    Here variables indexed by 1 and 2 are upstream (pre shock)
+    and downstream (post shock) respectively.
+    This function is equivalent to Eq. 4.24 in `Drake`_.
```
rocco8773, line 51: I'd also add here descriptions for s, p, rho, and gamma, must like you would do in a journal article.

plasmapy/formulary/shocks.py
rocco8773: I think you can just get away with annotating with `[u.kg / u.m ** 3, u.m ** -3]`.  This should allow either number density or mass density to pass through. Then below when you to calculate the ratio you can do something like...

```python
try:
    pressure_ratio = (p_2 / p_1).to(u.dimensionless_unscaled)
    density_ratio = (rho_1 / rho_2).to(u.dimensionless_unscaled)
except u.UnitConversionError as err:
    raise u.UnitConversionError(
        f"Upstream and downstream parameters must have the same units, {err}"
    )
```

plasmapy/formulary/shocks.py
rocco8773: I still think keeping the individual parameters is better, only because I think it's a more explicit signature.  However, I'm open to be convinced otherwise.

plasmapy/formulary/shocks.py
pheuer: Are there different methods that would be valid under this polytropic gas assumption? I haven't been able to find much information on this outside of Drake. Unless there are obvious, well-known models (like there are for resistivity, for example), I think that at the very least the documentation for this function should provide an appropriate equation or point to at least one function in PlasmaPy that will calculate it.

plasmapy/formulary/shocks.py
pheuer: ^Yes, this is the right answer.

plasmapy/formulary/shocks.py
pheuer: I agree with you now: the best solution is to take the individual parameters, make no requirements on their units except that they be identical, then construct the ratios (as you suggested above). That way users can use any pressure or density units they want.

plasmapy/formulary/shocks.py
rocco8773: I'd like to note...Drake's derivation of this uses mass density for rho.  Since I can't think of a scenario where you'd have two different species for upstream/downstream, the mass ratio is equal to the number ratio.

PR #869, opened 2020-07-19
Fix npsphinx-thumbnail for ./docs/notebooks/langmuir_analysis.ipynb

rocco8773:
The original "output-index" (1) for "nbsphinx-thumbnail" was pointing to an invalid MIME (a python dictionary instead of an image).  I changed the "output-index" to 2, so the first image is take as the thumbnail.

The documentation build failure was first noted by @StanczakDominik ([comment](https://github.com/PlasmaPy/PlasmaPy/pull/864#issuecomment-660463536)).

### DIFF
--- a/docs/notebooks/langmuir_analysis.ipynb
+++ b/docs/notebooks/langmuir_analysis.ipynb
@@ -94,7 +94,7 @@
      "outputs_hidden": false
     },
     "nbsphinx-thumbnail": {
-     "output-index": 1
+     "output-index": 2
     }
    },
    "outputs": [],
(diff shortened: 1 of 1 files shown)


StanczakDominik:
The changes look good to me, thanks a ton for figuring this out. I'm surprised nbsphinx would take a dictionary as a valid thumbnail, though! We might want to file an issue upstream.

rocco8773:
It was really an us problem. The cell executed with three outputs, the first being the dictionary and the later two being the images. We were specifically selecting the dictionary. 🤦🏻‍♂️ Oh well, live and learn.

StanczakDominik:
Ooooh, okay! That makes sense.

PR #873, opened 2020-07-23
Implement `__eq__` for `AbstractParticle`

### LINKED ISSUE #806: Implement __eq__ for DimensionlessParticle and CustomParticle
The `__eq__` dunder method for each of `CustomParticle` and `DimensionlessParticle` in `plasmapy.particles` has not yet been defined.  It would be reasonable to have `__eq__` return `True` if both the mass and charge attributes are the same for both.

When a mass and/or charge is not provided to `CustomParticle` or `DimensionlessParticle`, the corresponding attribute is set to `numpy.nan` (with the appropriate units).  I would suggest using `astropy.units.isclose` on each of the attributes, using `equal_nan = True` to account for this case.  In order to avoid issues with floating point comparisons, it may be best to set the relative tolerance `rtol` in `isclose` to something like 1e-14.

A possibility would be to implement this as `AbstractParticle.__eq__`.

### LINKED ISSUE #848: Implement `__eq__` and other utility methods for `Custom`- and `DimensionlessParticle`s
*Something is amiss in `particle_class`.*

Because `Particle` defines `__eq__`, `AbstractParticle` does not, and `CustomParticle` and `DimensionlessParticle`, which inherit from `AbstractParticle`, do not share the same functionality. This results in

```python
>>> c = CustomParticle(mass = 5.2 * u.kg, charge = 5.2 * u.C)
>>> c1 = from_json('{"plasmapy_particle": {"type": "CustomParticle", "module": "plasmapy.particles.particle_class", "date_created": "2020-06-08", "init_args": {"mass": "5.2 kg", "charge": "5.2 C"}}}')
>>> c == c1
False
>>> c.mass == c1.mass
True
>>> c.charge == c1.charge
True
```

The same will probably apply to other utility methods. 

Sleuthed out by @DarkAEther in https://github.com/PlasmaPy/PlasmaPy/pull/836#discussion_r437471882 :)

DarkAEther:

      docstrings.

Closes #806 
May close #848 

Implements `__eq__` into `AbstractParticle` class, allowing comparisons of `CustomParticle` and `DimensionlessParticle` using `==` operator.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -169,6 +169,45 @@ class AbstractParticle(ABC):
         """
         raise AtomicError("The truth value of a particle is not defined.")
 
+    def __eq__(self, other) -> bool:
+        """
+        Determine if two objects correspond to the same particle.
+
+        This method will return `True` if ``other`` is an identical
+        `~plasmapy.particles.Particle`, `~plasmapy.particles.CustomParticle`
+        , `~plasmapy.particles.DimensionlessParticle` instance, and return
+        `False` if ``other`` is a different particle.
+
+        If ``other`` is not an instance amongst `~plasmapy.particles.Particle`,
+        `~plasmapy.particles.CustomParticle`, or
+        `~plasmapy.particles.DimensionlessParticle`,
+        then this method will raise a `TypeError`.
+
+        """
+        # other_particle = None
+        # if isinstance(other, str):
+        #     try:
+        #         other_particle = serialization.json_loads_particle(other)
+        #     except InvalidElementError as exc:
+        #         raise InvalidParticleError(
+        #             f"{other} is not a particle and cannot be compared to {self}."
+        #         )
+        # else:
+        #     other_particle = other
+
+        if not isinstance(other, self.__class__):
+            raise TypeError(
+                f"The equality of a {type(self)} object with a {type(other)} is undefined."
+            )
+
+        # Check properties mass and charge for equivalence
+        if u.isclose(self.mass, other.mass, equal_nan=True, rtol=1e-14) and u.isclose(
+            self.charge, other.charge, equal_nan=True, rtol=1e-14
+        ):
+            return True
+        else:
+            return False
+
     def json_dump(self, fp, **kwargs):
         """
         Writes the particle's `json_dict` to the `fp` file object using `json.dump`.
--- a/plasmapy/particles/tests/test_particle_class.py
+++ b/plasmapy/particles/tests/test_particle_class.py
@@ -1318,3 +1318,135 @@ def test_particle_to_json_file(cls, kwargs, expected_repr):
         f"expected_repr = {expected_repr['__init__']}.\n\n"
         f"json_repr: {test_dict['__init__']}"
     )
+
+
+data_equivalence_custom_dimensionless = [
+    (
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
+        True,
+        None,
+    ),
+    (
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.2 * u.C},
+        CustomParticle,
+        {"mass": 5.12 * u.kg, "charge": 6.3 * u.C},
+        False,
+        None,
+    ),
+    (
+        DimensionlessParticle,
+        {"mass": 5.2, "charge": 6.3},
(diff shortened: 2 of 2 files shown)


DarkAEther:
This commit is a proof of concept and experiment for the mentioned issues.

A few queries I had in mind:
- I had a look at `__eq__` implemented for `Particle` and noticed that there is support for comparing against strings. Should that be supported in this case as well?
- Would it be possible to remove the `__eq__` implementation for `Particle` in favour of the one implemented for `AbstractParticle` (involves generalizing the `__eq__` implementation in 'AbstractParticle` to fit the needs of `Particle`)?

Thank you!

StanczakDominik:
> I had a look at __eq__ implemented for Particle and noticed that there is support for comparing against strings. Should that be supported in this case as well?

Well... for a `Particle("e-")` you can definitely state that it represents the same particle as the string `"e-"`. For `CustomParticles`, you're getting into float comparison... I'd say nah.


> Would it be possible to remove the __eq__ implementation for Particle in favour of the one implemented for AbstractParticle (involves generalizing the __eq__ implementation in 'AbstractParticleto fit the needs ofParticle`)?

Ideally, yeah! That seems like a good idea.

DarkAEther:
> Well... for a Particle("e-") you can definitely state that it represents the same particle as the string "e-". For CustomParticles, you're getting into float comparison... I'd say nah.

I agree. Best to avoid checking for equivalence between particles and string when the particle has a charge and a mass.

>>Would it be possible to remove the eq implementation for Particle in favour of the one implemented for AbstractParticle (involves generalizing the eq implementation in 'AbstractParticleto fit the needs ofParticle`)?

>Ideally, yeah! That seems like a good idea.

This might be tricky to generalise though. For `CustomParticle` and `DimensionlessParticle`, we can generalise because we can compare masses and charges, however, we can't do the same for `Particle` since a mass and charge comparison would not be sufficient to call them equivalent. Any suggestions on how we could handle this particular case are welcome ! 

Thank you!

namurphy:
I'm wondering a bit more about how we should treat `__eq__` specifically for a `DimensionlessParticle`.  To nondimensionalize a particle for a real or a theoretical application (ignoring implementation in PlasmaPy completely), we divide the mass by a characteristic mass, and divide the charge by a characteristic charge.  In those applications, the true value of the mass and charge depends on the normalization, which is not contained within the `DimensionlessParticle` class.  Strictly speaking, two dimensionless particles would only be equal if the normalized mass and charge are the same.  

With all that said, I still think it's reasonable to define `DimensionlessParticle` as having the same dimensionless charge and mass, since that's the behavior that most people would expect.  We'd just want to make sure to document what it's doing.

DarkAEther:
> To nondimensionalize a particle for a real or a theoretical application (ignoring implementation in PlasmaPy completely), we divide the mass by a characteristic mass, and divide the charge by a characteristic charge. In those applications, the true value of the mass and charge depends on the normalization, which is not contained within the DimensionlessParticle class. Strictly speaking, two dimensionless particles would only be equal if the normalized mass and charge are the same.

This is interesting and raises a question. At the moment, the 3 classes inheriting from `AbstractParticle` have 3 different implementations of `__eq__`. Could another approach be to make `__eq__` an abstract method in `AbstractParticle` and implement inside each class separately? The benefit of this approach is that it keeps the comparison consistent with standard comparisons for the particles that are currently there as well as for any future particles that might be added.

>With all that said, I still think it's reasonable to define DimensionlessParticle as having the same dimensionless charge and mass, since that's the behavior that most people would expect. We'd just want to make sure to document what it's doing.

This works too, however, may require redesigning and potential deprecation of some features from `__eq__` for `Particle`. For example, equality with strings may need to be deprecated in order to implement the method effectively in `AbstractParticle`.

rocco8773:
My opinion, defining `__eq__` and `__ne__` for `AbstractParticle` and then letting the sub-classes override or `super()` `__eq__`, if necessary, is the way to go.  This approach is better at standardizing the comparison and automatically makes the comparison available for all sub-classes.  `__eq__` for `AbstractParticle` should focus only on the methods defined within the `AbstractParticle` scope, i.e.

1. check for class instance `isinstance(other, self.__class__)`
1. check for equal masses `self.mass == other.mass`
1. check for equal charges `self.charge == other.charge`

With this approach, I don't think the flavor of `mass` or `charge` will be an issue because if they're a float, int, astropy Quantity, etc. they all have functionality for comparisons.

For the `Particle` class you would then `super()` this and add additional comparisons, e.g.

```python
class Particle(AbstractParticle):
    def __eq__(self, other):
        super().__eq__(other)

        # add additional comparisons
        # basically a deep comparison of the self._attributes dictionary
```

PR #875, opened 2020-07-26
Various improvements/updates to the documentation

### LINKED ISSUE #874: Thumbnails missing/broken on plasmapy documentation webpage home
As the title suggests, and can be seen on the screenshot, two of the three thumbnails
are missing from the PlasmaPy documentation front page at https://docs.plasmapy.org/en/latest/
Untitled

rocco8773:
I originally started this PR to fix #874 and, as always, I went down the rabbit hole.  When will I learn!!

Improvement/Updates made in this PR...

1. Replace the "home" text/image in the RTD sidebar with the plasmapy logo. (see pic below)
1. Use the `datatime` package to define the copyright years in the `copyright` configuration variable (in `conf.py`).  One less thing to maintain!!!
1. Make it so [`:hidden:` `toctree`'s do NOT show up in the navigation sidebar](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-includehidden).
1. Un-hide all `toctree` directives in `./docs/index.rst`.  This make the `toctree`'s visible on the landing page and in the sidebar.
1. Replace the "html `nbgallery`" on the landing page with a real `nbgallery` that is marked `:hidden:` (fixes #874).  `nbgallery` behaves a little differently than `toctree` in that the gallery itself is not hidden when marked with `:hidden:` but it is still hidden from the sidebar.  Another thing we do not have to manually maintain!!
1. Add a custom CSS stylesheet.  This is used in this PR to...
    1. Re-style `toctree` directives when used in a table.
    1. Create styles for adding the *module* and *index* navigation links to the sidebar.
1. Add navigation links *modules* and *index* to the sidebar "header". (see pic below)
1. Move the link to view page source code from the header of the page to the footer of the page.
1. Add to the foot a link to jump back to the top of the page.

image

Closes #874


StanczakDominik:
Oh wow.

StanczakDominik:
image
Interestingly, the PR builder now shows the PR number instead of the version. I wonder whether that's intended? Giles shows 0.4, as it should, so I assume this is just a quirk of RTD's test pr builder and should work fine.

docs/conf.py
```diff
@@ -72,8 +73,8 @@
 
 # General information about the project.
 project = "PlasmaPy"
-copyright = "2015-2020, PlasmaPy Community"
 author = "PlasmaPy Community"
+copyright = f"2015-{datetime.utcnow().year}, {author}"
```
StanczakDominik, line 77: Good hack! :laughing:

docs/conf.py
```diff
@@ -131,13 +132,24 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 #
-# html_theme_options = {}
+html_logo = "./_static/with-text-light-190px.png"
+html_theme_options = {
+    "logo_only": True,
```
StanczakDominik, line 137: Okay, that's pretty darn lovely!

docs/conf.py
```diff
@@ -131,13 +132,24 @@
...
+    #
+    # TOC options
+    #   https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#theme-options
+    "includehidden": False,
```
StanczakDominik, line 141: I was going to ask what the reasoning behind this is, but I see later on that it allowed you to refactor the formulary index table, for which, :+1: !

docs/conf.py
```diff
@@ -131,13 +132,24 @@
...
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-# html_static_path = ['_static']
+html_static_path = ["_static"]
 
+# A list of prefixes that are ignored for sorting the Python module
+# index (e.g., if this is set to ['foo.'], then foo.bar is shown under
+# B, not F).
+modindex_common_prefix = ["plasmapy."]
```
StanczakDominik, line 152: Ah, I see this fixed https://plasmapy--875.org.readthedocs.build/en/875/py-modindex.html - awesome! :)

One question - it looks like a few items are just straight up missing from there. Especially, `plasmapy.simulation`. Do you have any idea what could be causing that?

docs/_templates/breadcrumbs.html
```diff
@@ -0,0 +1,4 @@
+{%- extends "!breadcrumbs.html" %}
+
+{% block breadcrumbs_aside %}
+{% endblock %}
```
StanczakDominik, line 4: I see this is a work in progress, but: what would this do?

docs/_templates/footer.html
```diff
@@ -0,0 +1,41 @@
...
+                </a>
+            {% elif show_source and has_source and sourcename %}
+                <a href="{{ pathto('_sources/' + sourcename, true)|e }}" rel="nofollow">
+                    {{ _('Page Source') }}{{ reldelim2 }}
+                </a>
+            {% endif %}
+        {% endif %}
+        <a href="#">Back to Top</a>
+    </p>
+{% endblock %}
```
StanczakDominik, line 47: Full disclosure: I'm just going to skip looking closely through this one - the result looks great :D

docs/index.rst
```diff
@@ -6,45 +6,12 @@
...
-      </p>
-    </div>
-  </div>
-  <div class="sphx-glr-clear"></div>
+.. nbgallery::
+   :hidden:
+
+   notebooks/physics
+   notebooks/dispersion_function
+   notebooks/thomson
```
StanczakDominik, line 14: [I'll let Gimli, son of Gloin, speak for me here.](https://www.youtube.com/watch?v=7qF13urEo_Y)

docs/index.rst
```diff
@@ -99,26 +58,16 @@ package for plasma physics currently under development.
...
    be helpful to have more readily accessible from the main doc index
    page.
 
-*****
-Index
-*****
-
-* :ref:`genindex`
-* :ref:`modindex`
-* :ref:`search`
```
StanczakDominik, line 122: Just curious - do you think there's no place for the module index besides the button in the top left?

rocco8773:
> image
> Interestingly, the PR builder now shows the PR number instead of the version. I wonder whether that's intended? Giles shows 0.4, as it should, so I assume this is just a quirk of RTD's test pr builder and should work fine.

I do think it's an RTD things.  This [little bit of code](https://github.com/readthedocs/sphinx_rtd_theme/blob/1d856256676d2b801a12c30e4a849c5abc466da9/sphinx_rtd_theme/layout.html#L132-L142) is what defines the behavior.

docs/conf.py
rocco8773: I actually saw it in astropy's conf.py while search for something else.  And I was, "Wow! Why have we never thought of that.  We must have it!"

docs/conf.py
rocco8773: 😁

docs/conf.py
rocco8773: That, and it allows use to use the `nbgallery` directive in `./docs/index.rst` without the gallery titles showing up in the sidebar.

docs/conf.py
rocco8773: Yes, it's a larger issue I've been aware of for awhile.  Basically, not all of our modules are indexed because we basically don't `automodule` (or `:py:module:`) everything.  I've been testing a few ideas to solve this, but I'm not fully happy with any of the results yet.  It's another reason why I was diving so far into `sphinx_automodapi`.

docs/_templates/breadcrumbs.html
rocco8773: This part is complete.  This bit of code removes the view page source code link from the header.

I basically copied the [`breadcrumbs_aside` block](https://github.com/readthedocs/sphinx_rtd_theme/blob/1d856256676d2b801a12c30e4a849c5abc466da9/sphinx_rtd_theme/breadcrumbs.html#L39-L70) from the [`breadcrumbs.html` template](https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/breadcrumbs.html) and moved it into the [our `footer.html`](https://github.com/rocco8773/PlasmaPy/blob/ec0e100da6de1be968466ff3f242d6b5948c065f/docs/_templates/footer.html#L23-L44).  Then this bit of code wipes out the `breadcrumbs_aside` block so the link does not appear twice.

docs/index.rst
rocco8773: I don't think it needs to be anywhere else.  Where it is now will make it visible on all pages.  The only other location I can think to put the links would be in the top-right corner where the view page source link currently is.

I personally like it better in the sidebar because in makes more logical sense to me and would not be crowded by the navigation path (i.e. the `Docs » Formulary (plasmapy.formulary) » Distribution functions (plasmapy.formulary.distribution)`).

docs/conf.py
StanczakDominik: OH SO THAT'S IT!!!!!!!

docs/conf.py
StanczakDominik: Oh right, it's that. Thanks!

docs/index.rst
StanczakDominik: Okay then :)

StanczakDominik:
All right, could merge it whenever you like :)

rocco8773:
Awesome! Thanks for the quick review.

Issue #844, opened 2020-06-07
Fix the many warnings emitted during tests


StanczakDominik:
There's just so many, whyyyyyyyyyyy did we let it get like this

As of 2021-02-24, https://github.com/PlasmaPy/PlasmaPy/runs/1963982310 shows the following tests throw warnings:

<details>
<summary>All the warnings</summary>
plasmapy/analysis/tests/test_fit_functions.py::TestFFExponentialPlusOffset::test_root_solve
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_bias_and_current_length_mismatch
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_duplicate_bias_values
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_duplicate_bias_values
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_getpadded_limit
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_inplace_unique_bias
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_invalid_bias_dimensions
plasmapy/diagnostics/tests/test_langmuir.py::Test__Characteristic_inherited_methods::test_invalid_current_dimensions
plasmapy/diagnostics/tests/test_langmuir.py::Test__characteristic_errors::test_addition
plasmapy/diagnostics/tests/test_langmuir.py::Test__characteristic_errors::test_invalid_dimensions
plasmapy/diagnostics/tests/test_langmuir.py::Test__characteristic_errors::test_subtraction
plasmapy/diagnostics/tests/test_langmuir.py::Test__characteristic_errors::test_unequal_arrays
plasmapy/diagnostics/tests/test_langmuir.py::Test__swept_probe_analysis::test_ordering_invariance[False]
plasmapy/diagnostics/tests/test_langmuir.py::Test__swept_probe_analysis::test_ordering_invariance[True]
plasmapy/diagnostics/tests/test_langmuir.py::test_get_EEDF
plasmapy/diagnostics/tests/test_langmuir.py::test_get_floating_potential_with_return_arg
plasmapy/diagnostics/tests/test_langmuir.py::test_get_ion_density_OML_without_return_fit
[...]

PR #884, opened 2020-08-03
Remove some unit-related warnings in tests

StanczakDominik:
Tackles #844 a bit, cleaning up some of the warnings and bringing us down to 103 from 172. I did not yet touch docstrings, which are responsible for 27 of those still there.

This kind of thing is easier if you use `-Werror` with pytest and remove the doctests flags in setup.cfg.


StanczakDominik:
I forgot #886 removed most of these warnings. This PR is, thus, now obsolete.

PR #889, opened 2020-08-10
Create plasmapy.analysis.swept_langmuir & initiate w/ functionality for calculating the floating potential

rocco8773:
This PR creates the `plasmapy.analysis` sub-package and initializes it with the functionality to calculate the floating potential from a swept langmuir trace.

What's going on...

- [x] Creates the `plasmapy.analysis` sub-package
- [x] Creates the `plasmapy.analysis.swept_langmuir` sub-package to house functionality associated with analyzing swept langmuir traces.
    - [x] create module `floating_potential` for associated functionality
    - [x] create module `fit_functions` for specialized `FitFunction` classes to aid in curve fitting
- [x] Add function `find_floating_potential()` for calculating the floating potential of a swept langmuir trace.
- [x] Update documentation to...
    - [x] Replace the "Diagnostics" page with the "Analyses & Diagnostics" page.  This outlines the differences between diagnostics and analyses, as well as, branches out to the associated documentation page.
    - [x] Create admonitions warning about the active development of the diagnostic and analysis code.
    - [x] Add documentation page for `plasmapy.analysis.swept_langmuir`.
    - [x] Create notebook demonstrating use of `plasmpy.analysis.swept_langmuir.find_floating_potential()`.
- [x] Add `FutureWarnings` to the public facing API of `plasmapy.diagnostics.langmuir` to indicate its future deprication/migration to `plasmapy.analysis.swept_langmuir`.
- [ ] Add tests


docs/ad/ad_attention.inc
```diff
@@ -0,0 +1,6 @@
+.. important::
+   The Analysis and Diagnostic framework is in active development at the moment.
+   For the foreseeable future, the api will be in continuous flux as functionality
```
StanczakDominik, line 3:
```suggestion
   For the foreseeable future, the API will be in continuous flux as functionality
```

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,152 @@
...
+
+    How it works
+    ------------
+    #. The current array ``current` is searched for all points equal to zero and
+       point pairs that straddle  ``current = 0`` to form a set of crossing-points.
+    #. Crossing-points are then grouped into crossing-islands in accordence to
+       the ``threshold`` keyword.
+    #. If multiple crossing-islands are found, then an warning issued and
+       `~numpy.nan` is returned.
+    #. To calculated the floating potential, a `~scipy.stats.linregress` is applied
```
StanczakDominik, line 28:
```suggestion
    #. To calculate the floating potential, a `~scipy.stats.linregress` is applied
```

docs/ad/analysis/index.rst
```diff
@@ -0,0 +1,32 @@
+.. include:: /ad/ad_attention.inc
+
+.. py:currentmodule:: plasmapy.analysis
+
+.. _analysis:
+
+==============================
+Analyses (`plasmapy.analysis`)
```
namurphy, line 8:
```suggestion
Analysis tools (`plasmapy.analysis`)
```

docs/ad/index.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Analyses and diagnostics go hand-in-hand, but have subtle differences.  Thus,
+PlasmaPy gives each their own sub-packages, `plasmapy.analysis` and
+`plasmapy.diagnostics` respectively.
+
+
+Think of the `plasmapy.analysis` as your toolbox.  It has all the tools
+(functionality) you need to analyze your data.  Functionality is built around
+`numpy` arrays and each function has a well-defined, focused task.  For example,
+:func:`numpy.fft.fft` does one specific task, compute the one-dimensional
```
namurphy, line 17:
```suggestion
:func:`numpy.fft.fft` does one specific task: compute the one-dimensional
```

docs/ad/index.rst
```diff
@@ -0,0 +1,48 @@
...
+discrete Fourier Transform.  Similarly,
+:func:`plasmapy.analysis.swept_langmuir.find_floating_potential` only finds the
+floating potential for a given langmuir trace.  It does not have smoothing.
```
namurphy, line 20:
```suggestion
floating potential for a given Langmuir trace.  It does not have smoothing.
```

docs/ad/index.rst
```diff
@@ -0,0 +1,48 @@
...
+It does not do any filtering.  It does not do any signal conditioning.  It has a
+singular task to find the floating potential of a single langmuir trace.
```
namurphy, line 22:
```suggestion
It does not do any filtering.  It does not do any signal conditioning.  Its sole
responsibility is to find the floating potential of a single Langmuir trace.
```

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Parameters
+    ----------
+
+    voltage: ~numpy.ndarray
+        1-D numpy array of monotonically ascending/descending probe biases
+        (in Volts)
+
+    current: ~numpy.ndarray
+        1-D numpy array of probe current (in A) corresponding to the
```
namurphy, line 66:
```suggestion
        1-D numpy array of probe current (in amperes) corresponding to the
```

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,275 @@
...
+        are two types of curves ``"linear"`` and ``"exponential"`` (Default).
+
+    Returns
+    -------
+    vf: `float` or `numpy.nan`
+        The calculated floating potential (in Volts).  Returns `numpy.nan` if the
+        floating potential can not be determined.
+
+    vf_err: `float` or `numpy.nan`
+        The error associated with the floating potential calculation (in Volts).
```
namurphy, line 99:
```suggestion
        The calculated floating potential (in volts).  Returns `numpy.nan` if the
        floating potential can not be determined.

    vf_err: `float` or `numpy.nan`
        The error associated with the floating potential calculation (in volts).
```

plasmapy/diagnostics/langmuir.py
```diff
@@ -25,11 +25,19 @@
...
 
 from plasmapy.particles import Particle
 from plasmapy.utils.decorators import validate_quantities
 
 
+def _langmuir_futurewarning() -> None:
+    warn("The plasmapy.diagnostics.langmuir module will be deprecated in favor of "
+         "the plasmapy.analysis.swept_langmuir sub-package.  The analysis package "
+         " is in active development and is scheduled to be released with v0.5.0.",
+         FutureWarning)
```
namurphy, line 38: In the future I wonder if it would be helpful to have a decorator to handle deprecation warnings like this.  I'm not suggesting that for this PR though, since this approach is sufficient.

plasmapy/analysis/swept_langmuir/floating_potential.py
namurphy: I looked this up a while back and SI units are lower case even when they're based off of a name, except for Celsius.

namurphy (changes requested):
It's looking really good!  Thank you for doing this!  My suggested changes are all pretty minor.

One thought is that we should look into [`astropy.modeling`](https://docs.astropy.org/en/stable/modeling/) as a model fitting tool.  I only used it for the first time a month or two ago when working with a summer student who needed to fit some 1D Gaussians but it looks pretty general and like it has a lot of really useful capabilities.

docs/ad/analysis/index.rst
rocco8773: I wanted to keep "Analyses" because it better paired with "Diagnostics" in the sidebar and TOC.  I honestly prefer "Analysis Tools" as a heading, but felt the Analyses-Diagnostics pair look and feel in the sidebar and TOC beat out my preference.

plasmapy/analysis/swept_langmuir/floating_potential.py
rocco8773: I don't know why I did it like that.   Probably just doing "in V...oh, that's not descriptive enough, Let's complete that...in Volts".

rocco8773:
> One thought is that we should look into [`astropy.modeling`](https://docs.astropy.org/en/stable/modeling/) as a model fitting tool. I only used it for the first time a month or two ago when working with a summer student who needed to fit some 1D Gaussians but it looks pretty general and like it has a lot of really useful capabilities.

That would be useful.  Did not know about that package.  I would much prefer to rely on "more thought out" work like that.

I did a quick scan of the `astropy.modeling` package and my only concern with using it is error propagation.  It does not seem to report any statistics on the fit (errors, covariance, etc.).  I'll need to dig a little deeper though.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,273 @@
+{
```
StanczakDominik, line 1: <ul><li>processng -&gt; processing</li><li>"scanned for points that equal zero" - risky for floats? Maybe "up to a tolerance", if that's what happens there?</li><li>expoenetial -&gt; exponential</li></ul><p>I'm just glossing over whether the algorithm itself is sensible, though.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

StanczakDominik, line 1: <p>See my comments for the other notebook on pathlib! <code>join</code>  and <code>abspath</code>  are just darn ugly.</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

StanczakDominik, line 1: <p>"objec representing" -&gt; "object representing". Also, should vf and vf_err have units attached?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

StanczakDominik, line 1: <p>This could be then broken into a single cell with a variable call for each. It'll render prettier in a notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

StanczakDominik, line 1: <p>resulst -&gt; results</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

StanczakDominik, line 1: <p>Same here! Generally try to avoid this kind of <code>print</code>  in notebooks... take advantage of those __html__repr__'s!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

StanczakDominik, line 1: <p>Maybe change the vertical line color from orange? The fits are displayed in orange already.</p><p> </p><p>Also, one thing I found out about recently - xlim, ylim, xlabel and ylabel (and some others, but probably not ones here) could be wrapped into a single <code>ax.set</code>.  Try it out! It's less code.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

docs/notebooks/langmuir_samples/_generate_noisy.ipynb
```diff
@@ -0,0 +1,104 @@
+{
```
StanczakDominik, line 1: <p>Because of course you would <code>validate_quantities</code> on a one-off noise addition :D I laugh, but it's inspiring and a good habit to incorporate.</p><p> </p><p>In that case, I want a one-liner docstring though ;)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

StanczakDominik, line 1: <p>Just thought I'd let you know how much nicer <code>pathlib</code>  is:</p><p> </p><p><code>import pathlib</code></p><p><code>samples_path = pathlib.Path("./") # not sure what you're going for with ./ here</code></p><p><code>path = samples_path / filename</code></p><p><code>#then in the next cell:</code></p><p><code>save_path = path.with_name(path.stem + "_noisy.npz")</code></p><p> </p><p><code><span class="ql-cursor">﻿</span></code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/889/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

StanczakDominik (commented):
A partial review; I haven't yet looked through `analysis/swept_langmuir`. Those codecov warnings look threatening, though.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
rocco8773: >     * "scanned for points that equal zero" - risky for floats? Maybe "up to a tolerance", if that's what happens there?

In general I agree with you, but I don't think it necessary in this case.  Since the algorithm also collects point pairs that straddle `I=0`, those points that are really close to zero will still be collected.

rocco8773: I always seem to mix up those last three letters.  I can't tell you how many times I've corrected that. Arg!

docs/notebooks/langmuir_samples/_generate_noisy.ipynb
rocco8773: I laughed out loud at the "one-liner docstring" comment.  I totally deserve that. 🤣

rocco8773:
They're only there because I haven't gotten around to writing the tests yet.  Hopefully they won't be too terrible.

docs/notebooks/langmuir_samples/_generate_noisy.ipynb
rocco8773: Ahhh yeah, that is off.  I originally had this code in the `find_floating_potential` notebook and I didn't properly update everything when I copied it over.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
rocco8773: Ok, I implemented `pathlib`.  Not sure which I prefer, but I might side w/ `pathlib` after I used it more.

rocco8773: Good thing to note.  `ax.set` does not seem to be very robust though.  I can't even set font sizes with it.

rocco8773: I'm not sure what you mean by this.  Can you provide an example?

rocco8773: 🤔 I could give them units...

rocco8773: Since the input `current` and `voltage` arrays are not conditioned for units, I don't think `vf` and `vf_err` should be either.  If we put units on `vf` and `vf_err`, then I think units should be enforced on `current` and `voltage`.  I'm between two minds on which way to go with this...

StanczakDominik: Oh, I get what you mean now! Okay, that's fine, then.

StanczakDominik: Instead of putting this single large print call in a single cell, notebooks are better suited for just

```
>>> variable
value_of_variable
>>> other variable
value of other variable
```

It's cleaner that way, and if `value_of_variable` defines how to represent itself in HTML outputs, it'll render better in notebooks. That's true for Quantities, for example.

rocco8773: Ahh...that makes more sense. I thought you were talking about embedding it in the markdown cell.

I generally don't like using `print` in notebooks either. I'll see if your suggested way looks cleaner.

rocco8773: This is why I'm in two minds about having units on this function. I like enforcing units because it creates less chances for improper use, but it will slow the function down.  Like the [distribution discussion](https://github.com/PlasmaPy/PlasmaPy/pull/816#discussion_r432489113), this function will be applied repeatedly to to a series of arrays and checking for units will significantly hamper the performance.  Thus, I went with no units to start.

docs/ad/analysis/index.rst
StanczakDominik: To be honest, I think I still prefer "Analysis tools" despite the nice pairing in the sidebar: it's more explicit. "Analyses" makes me think of something I might do using a framework, not the toolset I'd expect to find in that framework. Closer to "use cases", probably, just stated weirdly.

I second Nick's suggestion here.

docs/ad/index.rst
```diff
@@ -0,0 +1,48 @@
...
+
+   - `xarray` provides labelled multi-dimensional arrays and datasets.
+   - Diagnostics self-manage the computed analysis data within a `xarray`
+     dataset while maintaining the computed data's relation to the original
+     data.
+
+#. Quick viewing of analyzed data with default plotting routines.
+#. Fully defining the physical parameters of a diagnostic with purposely
+   designed Probe classes that are integrated into the analysis workflow.
+#. Adding GUI's to the analysis workflow via notebook widgets.
```
StanczakDominik, line 40: Should we say that this part is a work in progress?

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
StanczakDominik: Okay, in that case, it's completely fine to keep this unit-less so long as the documentation states it clearly.

Meanwhile, look at [how Unitful.jl over in Julia-land affects performance... :sob:](https://github.com/JuliaAstro/Cosmology.jl/pull/16)

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,366 @@
...
+    "    The max allowed index distance between crossing-points before a new crossing-island is formed."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Calculate the Floating Potential\n",
+    "\n",
+    "Below we'll compute the floaing potential using the default fitting behavior (`fit_type=\"exponential\"`) and a linear fit (`fit_type=\"linear\"`)."
```
StanczakDominik, line 101:
```suggestion
    "Below we'll compute the floating potential using the default fitting behavior (`fit_type=\"exponential\"`) and a linear fit (`fit_type=\"linear\"`)."
```

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+        min_points = int(np.round(min_points * voltage.size))
+    elif min_points >= 1:
+        min_points = int(np.round(min_points))
+    else:
+        raise ValueError(f"Got {min_points}, but 'min_points' must be an int or float "
+                         f"greater than or equal to 0.")
+
+    # find possible crossing points (cp)
+    lower_vals = np.where(current <= 0, True, False)
+    upper_vals = np.where(0 <= current, True, False)
```
StanczakDominik, line 201: Aren't these just

```suggestion
    lower_vals = current <= 0
    upper_vals = 0 <= current
```

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+    cp_exact = np.logical_and(lower_vals, upper_vals).nonzero()[0]
```
StanczakDominik, line 202:
```suggestion
    cp_exact = (lower_vals & upper_vals).nonzero()[0]
```
0, because...?

StanczakDominik: Note to self: quitting here because the warnings are becoming too dense. Test it out, please!

StanczakDominik (commented):
A partial review of `floating_potential.py`:

docs/ad/index.rst
rocco8773: I kinda already am.  As part of this PR I'm placing an ["Important" admonition at the top](https://plasmapy--889.org.readthedocs.build/en/889/ad/index.html) of the analysis and diagnostic pages.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+    fit_type: str = "exponential",
+):
+    """
+    Determines the floating potential (:math:`V_f`) for a given Current-Voltage
+    (IV) curve obtained from a swept Langmuir probe.  The floating potential is
+    the probe bias where the collected current equals to zero :math:`I = 0`.
+
+    **How it works**
+
+    #. The current array ``current`` is scanned for all points equal to zero and
```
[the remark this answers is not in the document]
rocco8773: I don't get the issue here, this isn't part of any notebook and does not affect linking from a notebook.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+from collections import namedtuple
+from warnings import warn
+from typing import Union
+
+from plasmapy.analysis import fit_functions as ffuncs
+
+FloatingPotentialResults = namedtuple(
+    "FloatingPotentialResults",
+    ("vf", "vf_err", "rsq", "func", "islands", "indices"),
+)
```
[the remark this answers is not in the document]
rocco8773: Noted.  While slower in creation than `NamedTuple`, `Dataclass` does beat `NamedTuple` in all other metrics.  However, we have not officially dropped Python 3.6, so I think it should stay a `NamedTuple`.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+       the ``threshold`` keyword.
+
+       - A new island is formed when a successive crossing-point is more (index)
+         steps away from the previous crossing-point than allowed by
+         ``threshold``.
+       - If multiple crossing-islands are identified, then the total span of all
+         crossing-islands is compared to ``min_points``.  If the span is greater
+         than ``min_points`` then the function is incapable of identifying
+         :math:`V_f` and will return `numpy.nan` values; otherwise, the span
+         will form one larger crossing-island.
```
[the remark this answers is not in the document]
rocco8773: No, that's not quite right.  The islands are only "merged" if the span of the first point in the first island to the last point in the last island is less than ``min_points``.  For example,

```python
islands = [slice(4,10), slice(40,50), slice(60, 70)]
if (islands[-1].stop - islands[0].start) > min_points:
    # unable to find vf
    return np.nan
else:
    # data points to use for fit
    indices = slice(islands[0].start, islands[-1].strop)
```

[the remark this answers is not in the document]
rocco8773: I'll yield that the wording can use improvement, so I'll take another stab at it.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+    fit_funcs = {
+        "linear": {
+            "func": ffuncs.Linear(),
+            "min_point_factor": 0.1,
+        },
+        "exponential": {
+            "func": ffuncs.ExponentialPlusOffset(),
+            "min_point_factor": 0.2,
+        },
+    }
```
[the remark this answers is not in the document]
rocco8773: I don't like the idea of defining functionality specific to that function outside of the function, but I see your point.  I can make a compromise and not instantiate the classes inside the dict.  I can do that at the time of selection.

docs/ad/index.rst
StanczakDominik: Oh right, yeah. :+1:

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
StanczakDominik: Numpy's NEP 37 or whichever it was does suggest dropping 3.6 by now iirc. Since we depend on Numpy, we're softly tied to their release schedule, even if we haven't officially dropped 3.6.

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
StanczakDominik: I meant that if I do `sweeping_potential?` from jupyter or ipython, I'll get the raw docstring, not the rendered one. Thus, hashes in place of list numbers.

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
rocco8773: I like the concept of a `Dataclass`, but I still think a `NamedTuple` is more appropriate in this use case. (1) You can still index a named tuple and I can see users just wanting to take the first element and running with it.  I see all the swept Langmuir analysis functionality having a return similar to this, so users can always get the first element by doing `results[0]` versus `results.vf`, `results.vp`, `results.te`, etc. (2) Function returns of multiple values are always a tuple, so a `NamedTuple` is more inline with what is expected. (3)  I actually got this idea from `numpy` so, if it's good for them, it's good for me.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+    (IV) curve obtained from a swept Langmuir probe.  The floating potential is
+    the probe bias where the collected current equals to zero :math:`I = 0`.
+
+    **How it works**
+
+    #. The current array ``current`` is scanned for all points equal to zero and
+       point pairs that straddle :math:`I = 0`.  This forms an array of
+       "crossing-points."
+    #. The crossing-points are then grouped into "crossing-islands" in based on
+       the ``threshold`` keyword.
```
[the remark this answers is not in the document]
rocco8773: Okay, I get what you mean now. 👍

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+         crossing-islands is compared to ``min_points``.  If the span is greater
+         than ``min_points`` then the function is incapable of identifying
+         :math:`V_f` and will return `numpy.nan` values; otherwise, the span
+         will form one larger crossing-island.
+
+    #. To calculate the floating potential...
+
+       - If the crossing-island contains less points than ``min_points``, then
+         each side of the crossing-island is equally padded with the nearest
+         neighbor points until ``min_points`` is satisfied.
```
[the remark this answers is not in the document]
rocco8773: It's to document, in RTD, for users how the function works.  A lot of users will not scan the code to get this info. This is for them.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,303 @@
...
+
+    """
+    rtn = FloatingPotentialResults(
+        vf=np.nan,
+        vf_err=np.nan,
+        rsq=None,
+        func=None,
+        islands=None,
+        indices=None
+    )._asdict()
```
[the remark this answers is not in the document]
rocco8773: The default values here are the "failure" values.

I do think the `tuple[2:5]` values are necessary.  It's typical to make plots with the fit overlays on your experimental data.  It's a bit of a sanity check because swept Langmuir traces are not typically "ideal."  It also helps for debugging your fits.  I'm also thinking of writing a function that would optimize `vf_err` and `rsq`, just not sure how to do it yet.  Having this info already in the return will help with that optimization.

plasmapy/analysis/swept_langmuir/floating_potential.py
rocco8773: Yeah, it's the same thing.  I've just always done it with `np.where` because that's how I first learned it.

rocco8773: But my way is significantly slower, so I'll change it up...

image

plasmapy/analysis/swept_langmuir/floating_potential.py
rocco8773: There does not seem to be a significant difference in performance between the two options.  I feel `np.logical_and` is more explicit, so I'm going to stick with that.

image

plasmapy/analysis/swept_langmuir/floating_potential.py
svincena: It's unlikely that anything equals zero, but should you use, for example ,
lower_vals = current <= 0
upper_vals = 0 < current

rocco8773: That's a fringe condition.  The current implementation could accidentally collect points around `current == 0`.  Taking your suggestion in mind, I think this would be the right approach...

```python
lower_vals = current < 0
upper_vals = 0 < current
cp_exact = (current == 0).nonzero()[0]
```

docs/_static/rtd_theme_overrides.css
```diff
@@ -1,3 +1,29 @@
+/* -----------------------------------------------------------------------------
+ * RTD Overrides
```
namurphy, line 2: This makes me think that RTD Override would be a good robot boss name in a mid-1990s video game trying to be cool.  Thus ends the most useful code review comment in the history of the universe.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,295 @@
...
+        1-D numpy array of monotonically ascending/descending probe biases
+        (should be in volts)
+
+    current: `numpy.ndarray`
+        1-D numpy array of probe current (should be in amperes) corresponding
+        to the :data:`voltage` array
+
+    threshold: positive, non-zero `int`
+        Max allowed index distance between crossing-points before a new
+        crossing-island is formed.  That is, if `threshold=5` then consecutive
```
namurphy, line 71:
```suggestion
        crossing-island is formed.  That is, if ``threshold=5`` then consecutive
```

plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+def check_sweep(voltage: np.ndarray, current: np.ndarray) -> None:
+    """
+    Function for checking that the voltage and current arrays are properly
+    formatted for analysis by `plasmapy.analysis.swept_langmuir`.
+
+    Parameters
+    ----------
+    voltage: `numpy.ndarray`
+        1D `numpy.ndarray` representing the voltage of the swept Langmuir trace.
+        Voltage should be monotonically increasing.
```
namurphy, line 16:
```suggestion
        Voltage should be monotonically increasing.  Assumes units of
        volts.
```
Assuming this does actually assume units of volts.

plasmapy/analysis/swept_langmuir/tests/test_helpers.py
```diff
@@ -0,0 +1,71 @@
+"""Tests for `plasmapy.analysis.swept_langmuir.helpers`."""
+
+import numpy as np
+import pytest
+
+from contextlib import ExitStack as does_not_raise
+
+from plasmapy.analysis.swept_langmuir.helpers import check_sweep
+
+# TODO: replace ExitStack with nullcontext when we require >= python 3.7
```
namurphy, line 10: This makes me want to require python >= 3.7 even though I don't know what `ExitStack` and `nullcontext` are.

namurphy (commented):
Here's a review that I apparently started months ago but seemingly never submitted because of how awesome I am at life, so some of the comments might be a little bit dated based on conversations we've had since that I can't quite remember.  Feel free to ignore any suggestions that are irrelevant or superficial.  I do like the direction that this is going in, so thank you once again for working on it.

plasmapy/analysis/swept_langmuir/tests/test_helpers.py
rocco8773: It's a conditional for `with` the does nothing.  It's useful when you have a test function that mixes cases that raises and does not raise `Exception`'s/`Warning`'s.  For example,

```python
@pytest.mark.parametrize(
    "args, expected, conditional",
    [
        (1, None, pytest.mark.raises(PlasmaPyError)),
        ("hello", "goodbye", ExitStack()),
    ],
)
def test_something(args, expected, conditional):
    with conditional:
        results = foo(*args)
        assert results == expected
```

namurphy: Oh!  Thanks!

plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: I'm actually not enforcing units at this point.  I'm primarily doing this for speed reasons.  When analyzing data this routine will be called hundreds of times and I don't want to enforce unit validations every single time.  I'm thinking the `plasmapy.diagnostic` (aka `xarray`) version will have the appropriate unit handling associated with it.

plasmapy/analysis/swept_langmuir/tests/test_floating_potential.py
```diff
@@ -16,13 +17,35 @@
...
+        "vf_err",
+        "rsq",
+        "func",
+        "islands",
+        "indices",
+    )
+    if sys.version_info >= (3, 7):
+        # TODO: remove if clause when python 3.6 support is dropped
+        assert hasattr(FloatingPotentialResults, "_field_defaults")
+        assert FloatingPotentialResults._field_defaults == {}
```
namurphy, line 39: I would totally be okay with dropping 3.6 support for our next development release, especially following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).  Plus, Astropy 4.2 requires Python 3.7, so we would likely effectively be dropping Python 3.6 support in the next year anyway.

rocco8773: Likewise.  I'm trying to leave these TODOs in key places so it will be easier go back and update when the py36 drop PR arrives.

plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+        If either the ``voltage`` or ``current`` arrays are not instances of a
+        `numpy.ndarray`.
+
+    `ValueError`
+        If either the ``voltage`` or ``current`` arrays are not 1D.
+
+        If the ``voltage`` array is not monotonically increasing.
+
+        If the ``current`` array never cross zero (i.e. has not floating
+        potential).
```
namurphy, line 35:
```suggestion
        If the ``current`` array never crosses zero (i.e. has no floating
        potential).
```

plasmapy/analysis/tests/test_fit_functions.py
```diff
@@ -9,7 +9,7 @@
 
 import plasmapy.analysis.fit_functions as ffuncs
 
-# ^ ExitStack can be replaced with nullcontext when we require >= python 3.7
+# TODO: replace ExitStack with nullcontext when we require >= python 3.7
```
namurphy, line 12: We should be able to do this now, or at least soon!

plasmapy/diagnostics/langmuir.py
```diff
@@ -25,11 +25,21 @@
...
 
 from plasmapy.particles import Particle
 from plasmapy.utils.decorators import validate_quantities
 
 
+def _langmuir_futurewarning() -> None:
+    warn(
+        "The plasmapy.diagnostics.langmuir module will be deprecated in favor of "
+        "the plasmapy.analysis.swept_langmuir sub-package.  The analysis package "
+        " is in active development and is scheduled to be released with v0.5.0.",
```
namurphy, line 38: I think this warning message will need some updating.

namurphy (changes requested):
Thanks again for working on this!  I looked through the production code but not the notebooks or tests this time.  My main comment on the code structure is still that there are a few places where using the "extract method" refactoring pattern could work pretty well.

I'm still wondering if we could make this compatible with `Quantity` objects, so that the analysis could be run using either `ndarray` objects or `Quantity` objects.  Pretty much for any analysis I do I'd prefer using `Quantity` objects until speed becomes a serious issue.  I do know that you were planning on having some of this user facing functionality in `diagnostics` though.

In any case, this is looking to be in quite good shape and I'm looking forward to it getting merged!

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,368 @@
+{
```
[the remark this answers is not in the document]
rocco8773: I say the answer to that is, it depends.  From a reviewer standpoint, a notebook can be beneficial to understanding the functionality being added.  I do have a concern that if a notebook is postponed, then it just won't happen.  I often will develop a notebook in parallel with my functionality to help test/refine the functionality, so I already have a notebook ready for the PR in review.

With that said, I do think there will be situations where reviewing a notebook separately would help get functionality integrated quicker.  I think the default practice should be contributing the notebook with the main PR, but be open to postponing the notebook when necessary.

plasmapy/analysis/tests/test_fit_functions.py
rocco8773: Done!

plasmapy/diagnostics/langmuir.py
rocco8773: Done!

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,297 @@
...
+            threshold = 1
+            warn(
+                f"threshold ({threshold}) is less than 1 and needs to be"
+                f" an int >= 1, using a value of 1",
+                PlasmaPyWarning,
+            )
+
+    else:
+        threshold = 1
+        warn(f"threshold is NOT a integer >= 1, using a value of 1", PlasmaPyWarning)
```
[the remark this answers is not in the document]
rocco8773: Done!

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,295 @@
...
+            f"Got {min_points}, but 'min_points' must be an int or float "
+            f"greater than or equal to 0."
+        )
+
+    # find possible crossing points (cp)
+    lower_vals = current < 0
+    upper_vals = 0 < current
+    cp_exact = (current == 0.0).nonzero()[0]
+    cp_low2high = np.logical_and(np.roll(lower_vals, 1), upper_vals).nonzero()[0]
+    cp_high2low = np.logical_and(np.roll(lower_vals, -1), upper_vals).nonzero()[0]
```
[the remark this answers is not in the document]
rocco8773: While I'm usually supportive of this idea, I don't think the code here is complex or long enough to warrant this approach.  I feel the "extract method refactoring` reduces readability when the code is not complex or long, since the user/programmer has to woner away from the current line to find out what the method does.

plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: Yes and no,  there is no unit enforcement in this functionality.  In practice, the voltage array should be in volts and the current array should be in amperes.

I have added functionality to these checks that will strip units if a `u.Quantity` arrays is passed.

When it comes to the `xarray` version, I will enforce `volts` for the voltage array and `amperes` for the current array.

rocco8773:
@namurphy Alright, the tests are finally complete.  I do not have anything else to add to this, so a red pen review will be highly appreciated! :)

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
namurphy: Huh...I've lately been thinking that 20-line functions are long!  I'll still encourage this, but let you make the final decision.

namurphy (commented):
Looked over everything except notebook stuff for the moment.  It's looking quite good, and all of my suggestions are optional.  Will try to get to the notebooks soon, in between dealing with plumbers coming by and carpenters cutting big holes in my kitchen wall.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,291 @@
...
+
+def find_floating_potential(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    threshold: int = 1,
+    min_points: Union[int, float] = None,
+    fit_type: str = "exponential",
+):
+    """
+    Determines the floating potential (:math:`V_f`) for a given Current-Voltage
```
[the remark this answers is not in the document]
rocco8773: I Don't Know What That Was About Either.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,297 @@
...
+          ``voltage`` and ``factor = 0.1`` for ``fit_type = "linear"`` and
+          ``0.2`` for ``"exponential"``.
+        - ``min_points = 0`` The entire passed array is fitted.
+        - ``min_points >= 1`` Exact minimum number of points.
+        - ``0 < min_points < 0`` The minimum number of points is taken as
+          ``min_points * array_size``.
+
+    fit_type: str
+        The type of curve to be fitted to the Langmuir trace,  ``"linear"`` or
+        ``"exponential"`` (Default).  This selects which `FitFunction` class
```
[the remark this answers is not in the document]
rocco8773: With regards to `default` vs `Default`, I generally prefer `Default` because I feel `default` is more easily lost among the rest of the parameter description.  I am also open to standardizing this across `plasmapy` and just adopting that standard.

I see the possible ambiguity with using `"exponential"`, but I feel it's not too bad here since `find_floating_potential` only uses the options `"linear"` and `"exponential"` for `fit_type` and only one version of the exponential fit functions is deployed.  I am prioritizing user convenience over explicitness here, since I don't feel understanding is lost by going from `"exponential_plus_offset"` -> `"exponential"`.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,291 @@
...
+
+    rsq: `float`
+        The coefficient of determination (r-squared) value of the fit.  See the
+        documentation of the ``rsq`` property on the associated fit function
+        (e.g. the
+        `~plasmapy.analysis.fit_functions.ExponentialPlusOffset.rsq`
+        property of
+        `~plasmapy.analysis.fit_functions.ExponentialPlusOffset`).
+
+    func: sub-class of `~plasmapy.analysis.fit_functions.AbstractFitFunction`
```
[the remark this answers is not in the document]
rocco8773: I still prefer `func`, because I don't think end users will be using the full functionality of the fit functions here.  The returned fit function is callable so users will typically just call the "function" to plot against their data.  So I am kinda hiding the complex machinery behind the scenes, but that's intentional.

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
rocco8773: Oh...I don't really consider the length/complexity of the "parent" function when I break off the method, but more the length/complexity of the code that would go into the "method".  If the "method" code easily fits within 2/3 of my screen (i.e. I can still easily read the pre- and post-sections around the "method" code) then I leave it in place.  If I lose the readability of the pre- and post-sections around the "method" code, then I'll break that code-block off into it's own method.  I do it this way, because I find when the code is broken up too much then it's really hard for me to follow it, even when the method names are "explicit."

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,291 @@
...
+
+    threshold: positive, non-zero `int`
+        Max allowed index distance between crossing-points before a new
+        crossing-island is formed.  That is, if ``threshold=5`` then consecutive
+        crossing-points are considered to be in the same crossing-island if
+        they are within 5 index steps of each other. (Default: 1)
+
+    min_points: positive `int` or `float`
+        Minimum number of data points required for the fitting to be applied to.
+        See **How it works** above for additional details.  The following list
```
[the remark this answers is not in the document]
rocco8773: Unfortunately, `numpydoc` does not allow for docstring headings that are not part of their default headings.  Additionally, `numpydoc` does not generate linkable headings, so ...

As an alternative, I've moved the **How it works** "section" into the `numpydoc` **Notes** section and moved it to the bottom of he docstring.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,291 @@
...
+    threshold: int = 1,
+    min_points: Union[int, float] = None,
+    fit_type: str = "exponential",
+):
+    """
+    Determines the floating potential (:math:`V_f`) for a given Current-Voltage
+    (IV) curve obtained from a swept Langmuir probe.  The floating potential is
+    the probe bias where the collected current equals zero :math:`I = 0`.
+
+    **How it works**
```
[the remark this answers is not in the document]
rocco8773: Unfortunately not possible, see my https://github.com/PlasmaPy/PlasmaPy/pull/889#discussion_r568963823.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -0,0 +1,291 @@
...
+    if min_points is None:
+        min_points = int(np.max([5, np.around(min_point_factor * voltage.size)]))
+    elif not isinstance(min_points, (float, np.floating, int, np.integer)):
+        raise TypeError(
+            f"Argument 'min_points' is wrong type '{type(min_points)}', expecting "
+            f"an int or float."
+        )
+    elif min_points == 0:
+        # this signals to use all points
+        pass
```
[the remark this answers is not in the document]
rocco8773: I'm going to keep the default functionality as is.  By default users are going to be passing in a full swept Langmuir trace and not an appropriately cropped one.  As such, I want the function to automatically identify a "reasonable" window to perform the fit on.  `min_points=0` is intended for users that do pass in a cropped trace.

----

`min_points=0` does seem weird as a mechanism to select all points, I think I'll replace this with `np.inf` instead.

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
namurphy: Sad kittens! Now that you mention it, I think I do remember you mentioning this before.  I like the idea of it being in the `Notes` section.

plasmapy/analysis/swept_langmuir/floating_potential.py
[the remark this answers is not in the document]
svincena: Neither voltage nor current are capitalized.

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -297,4 +297,4 @@ def find_floating_potential(
 
 
 find_vf_ = find_floating_potential
-"""Alias to :func:`find_floating_potential`. """
+"""Alias to :func:`find_floating_potential`."""
```
namurphy, line 300: Sorry I missed this...it appears I was spacing out.  😬

rocco8773: Looks like we found it independently at about roughly the same time. 🤣

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Swept Langmuir Analysis: Floating Potential\n",
+    "\n",
+    "This notebook covers the use of the [**find_floating_potential()**](../../../api/plasmapy.analysis.swept_langmuir.floating_potential.find_floating_potential.rst#find-floating-potential) function to determining the floating potential from a swept Langmuir trace."
```
namurphy, line 9:
```suggestion
    "This notebook covers the use of the [**find_floating_potential()**](../../../api/plasmapy.analysis.swept_langmuir.floating_potential.find_floating_potential.rst#find-floating-potential) function to determine the floating potential from a swept Langmuir trace."
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "    1. [Plotting results](#Plotting-results)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## How `find_floating_potential()` works\n",
+    "\n",
+    "1. The passed current array is scanned for points that equal zero and point-pairs that straddle $I = 0$.  This forms an collection of \"crossing-points.\"\n",
```
namurphy, line 51:
```suggestion
    "1. The passed current array is scanned for points that equal zero and point-pairs that straddle where the current equals zero.  This forms an collection of \"crossing-points.\"\n",
```
Or, alternatively, define `I`.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "1. The crossing-points are then grouped into \"crossing-islands\" based on the `threshold` keyword.\n",
+    "    - A new island is formed when a successive crossing-point is more (index) steps away from the previous corssing-point than defined by `threshold`.  For example, if `threshld=4` then an new island is formed if a crossing-point candidate is more than 4 steps away from the previous candidate.\n",
```
namurphy, line 53:
```suggestion
    "    - A new island is formed when a successive crossing-point is more (index) steps away from the previous crossing-point than defined by `threshold`.  For example, if `threshold=4` then an new island is formed if a crossing-point candidate is more than 4 steps away from the previous candidate.\n",
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Swept Langmuir Analysis: Floating Potential\n",
```
namurphy, line 7: I think it would be helpful to start off the notebook with some higher level context, like what a floating potential is, what a swept Langmuir trace is (and is for), and why we want to go about finding it.  It needn't be super detailed, but just enough for someone to have an idea of the big picture.

namurphy: This might also be a good place to define V_f as the floating potential.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "    - If `fit_type=\"exponential\"`, then a `scipy.optimize.curve_fit` fit is applied to the points that make up the crossing-island."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Notes about usage\n",
+    "\n",
+    "- The function provides no signal processing, so, if needed, the user must smooth, sort, crop, etc. the arrays before passing to the function.\n",
```
namurphy, line 67:
```suggestion
    "- The function provides no signal processing.  If needed, the user must smooth, sort, crop, or process the arrays before passing them to the function.\n",
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Knobs to turn\n",
+    "\n",
+    "- `fit_type`\n",
+    "\n",
+    "    There are two types of curves that can be fitted to the identified crossing point data, `\"linear\"` and `\"exponential\"`.  The former will fit fit a line to the data, whereas, the later will fit and exponential curve with an offset.  The default curve is `\"exponential\"` since swept Langmuir data is not typically linear as it passes through $I=0$.\n",
```
namurphy, line 80:
```suggestion
    "    There are two types of curves that can be fitted to the identified crossing point data: `\"linear\"` and `\"exponential\"`.  The former will fit fit a line to the data, whereas, the later will fit and exponential curve with an offset.  The default curve is `\"exponential\"` since swept Langmuir data is not typically linear as it passes through $I=0$.\n",
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "\n",
+    "- `min_points`\n",
+    "\n",
+    "    This specifies the minimum number of points that will be used in the curve fitting.  As mentioned above,  the crossing-islands are identified and then padded until `min_points` is satisfied.\n",
```
namurphy, line 84:
```suggestion
    "    This variable specifies the minimum number of points that will be used in the curve fitting.  As mentioned above, the crossing-islands are identified and then padded until `min_points` is satisfied.\n",
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+   "outputs": [],
+   "source": [
+    "(results[1], results.vf_err)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "- `results[2]` = `results.rsq` = the coeficient of determination (r-squared) value of the fit"
```
namurphy, line 169:
```suggestion
    "- `results[2]` = `results.rsq` = the coefficient of determination (r-squared) value of the fit"
```

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
```diff
@@ -0,0 +1,390 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "- `results[3]` = `results.func` = the resulting fitted function\n",
+    "\n",
+    "    - `results.func` is a callable representation of the fitted function `I = results.func(V)`.\n",
+    "    - `resulst.func` is an instance of a sub-class of `AbstractFitFunction`. ([FitFuction classes](../../../api_static/plasmapy.analysis.fit_functions.rst))\n",
+    "    - Since `results.func` is a class instance, there are many other attribures available.  For example,\n",
```
namurphy, line 189:
```suggestion
    "    - Since `results.func` is a class instance, there are many other attributes available.  For example,\n",
```

namurphy (changes requested):
First notebook reviewed!  Most of my comments are inconsequential, with a bunch of minor wording or speeling changes.  The main changes I'm suggesting are:
 - Add a small numbers of sentences to describe the overall context at the beginning of the notebook to give an idea of the big picture, in particular to newcomers to plasma science who might be doing this as their first research project.
 - It'd be helpful to define variables like `$I$` and `$V_f$` since the notation might not be familiar to newcomers.

docs/notebooks/langmuir_samples/_generate_noisy.ipynb
```diff
@@ -0,0 +1,171 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# noise generator\n",
+    "@validate_quantities\n",
+    "def add_noise(signal, snr_db: u.dB = 10 * u.dB):\n",
```
namurphy, line 36: Logarithmic units! 💯   Oh wait, I meant 2️⃣.

namurphy: Somewhat more seriously, I wonder if we should have functions in PlasmaPy that can be used to add different kinds of noise.  That could potentially go in utils.  No need to do anything about that now...perhaps we could wait until later when we have need for something like this again.

namurphy (commented):
And no substantive comments on the second notebook.  This is really close, so I'm looking forward to it being merged!

docs/notebooks/langmuir_samples/_generate_noisy.ipynb
rocco8773: I was thinking the function might be useful in plasmapy, specifically in `plasmapy.analysis.signal` (similar in scope to `scipy.signal`).  I wasn't sure though, so I just left the notebook where it is for now.  The notebook isn't actually exposed anywhere in the documentation (on purpose).  It's only being used to make a Langmuir sweep with noise, which is then used in the Langmuir notebooks.  I just left the notebook in place for record purposes.

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
rocco8773: Yeah, I've been going back and forth about this.  I'm debating about organizing the notebooks in one of two ways:

1. Have a notebook for each swept Langmuir function that both describes the function's behavior and the physics behind the parameter being calculated.
2. Have a note book for each swept Langmuir functionality that only describes the functions behavior.  Then have a separate notebook that walks through the whole swept Langmuir analysis from beginning to end that focuses more on the swept Langmuir physics and analysis workflow.

I haven't decided which I prefer, but I'm leaning more towards the later.  I won't really know until I get more functionality incorporated.  For expidient reasons, I kept this notebook based more on functionality and plan to come back after I decide on which of the above options I want to go with.  So, I'd rather keep notebook's scope as is and come back in a followup PR.

namurphy: Sounds good!  For the notebooks for each swept Langmuir function, I think it'd still be helpful to have a sentence or two to connect the notebook to the big picture, but that'll probably be easier to do once the notebook that goes through the whole swept Langmuir analysis is done.

namurphy (approved):

docs/notebooks/analysis/swept_langmuir/find_floating_potential.ipynb
svincena: Or, leave documentation of the low-level functionality to the documentation contained within the functions themselves; definitely keep the high level notebook for the whole walkthrough process; and have short notebooks that describe the physics of why one needs to fit  various portions of the curve in detail with examples of how this is accomplished with the various functions. I think this translates to 'notebooks for physics' and 'no notebooks dedicated to functions'

namurphy:
Congratulations!  I'm excited for this to be merged.  Today has been a good day for pull requests, hasn't it?

PR #893, opened 2020-08-26
Thomson Scattering Multiple Electron Populations (and a bug fix)

pheuer:
This PR adds support to the diagnostics.Thomson.spectral_density() function for multiple thermal electron populations (still under the requirement that the resulting plasma is quasi-neutral). The electron keywords are adjusted to use the same syntax as the ion keywords (the function already supported multiple ion species).

This PR also fixes a bug for multiple ion species by adding a factor of ifract to the calculation of the ion contribution.  Physically, this corresponds to modifying the magnitude of the Maxwellian distribution function in S(k,w). Without this, running the code for two identical ion populations with ifract=[0.5,0.5] gives twice the result as running it for ifract=[1.0]. The same correction factor has been added for efract and both now pass this test (which has been added explicitly as a unit test).


plasmapy/diagnostics/thomson.py
```diff
@@ -160,6 +175,16 @@ def spectral_density(
             continue
         ion_species[ii] = Particle(ion)
 
+    # Condition Te
+    if Te.size == 1:
+        # If a single quantity is given, put it in an array so it's iterable
+        # If Te.size != len(efract), assume same temp. for all species
+        Te = [Te.value] * len(efract) * Te.unit
```
StanczakDominik, line 182:
```suggestion
        Te = np.repeat(Te, len(efract))
```
A little simpler!

plasmapy/diagnostics/thomson.py
```diff
@@ -224,54 +257,58 @@ def spectral_density(
...
+    alpha = np.sqrt(2) * wpe / np.outer(k, vTe)
 
     # Calculate the normalized phase velocities (Sec. 3.4.2 in Sheffield)
-    xe = (w_e / (k * vTe)).to(u.dimensionless_unscaled)
+    xe = (np.outer(1 / vTe, 1 / k) * w_e).to(u.dimensionless_unscaled)
     xi = (np.outer(1 / vTi, 1 / k) * w_i).to(u.dimensionless_unscaled)
 
     # Calculate the susceptibilities
-    chiE = permittivity_1D_Maxwellian(w_e, k, Te, ne, "e-")
+    # chiE = permittivity_1D_Maxwellian(w_e, k, Te, ne, "e-")
```
StanczakDominik, line 272:
```suggestion
    
```

StanczakDominik (approved):
Good, clean changes! Sorry for taking a while to get to them.

Do you think it might make sense to showcase the multiple electron populations in the notebook as well?

A few minor fixes:

pheuer:
@StanczakDominik Thank you for the edits! I've made a slight change to the notebook example to include multiple electron populations in the last set of plots (along with multiple ion populations).

docs/notebooks/thomson.ipynb
```diff
@@ -136,8 +136,9 @@
     "non_collective = {\n",
     "    'name': 'Non-Collective Regime', \n",
     "    'n': 5e15*u.cm**-3, \n",
-    "    'Te': 40*u.eV, \n",
+    "    'Te': np.array([40])*u.eV,\n",
     "    'Ti': np.array([10])*u.eV,\n",
+    "    'efract': np.array([1]), \n",
```
StanczakDominik, line 141: I just realized that this is a bit concerning - if there's a single electron population, wouldn't it make sense to allow passing a scalar Te and skipping putting in efract?

Then we could simplify the notebook by putting scalar values here and calling `x.get('efract')` where we currently have `efract=x['efract']`. If we make `efract` default to `None`, then catch it as None and replace it with `np.ones(1)` in the function's body, it's gonna work nicely.

Come to think of it, we could - and should, because having mutable lists (arrays are probably the same thing) as arguments is a bad call - replace that for `ifract` as well.

Does that sound reasonable to you?

pheuer: @StanczakDominik Sure that makes sense: the first part of what you asked (scalar Te) was actually already implemented and I just forgot to update it in the notebook. I've changed that now.

I've also made the change you suggested for efract and ifract: they now default to None and in that case are replaced with np.ones(1). I change the notebook to use this capability as you suggested, too.

StanczakDominik: Thank you! I'll look through tomorrow (SLEEP HYGIENE, I SHALL NOT ABANDON YOU FOR PURPOSES OF WATCHING COOL CODE) and probably merge soon :)

StanczakDominik (approved):
Looks great! :+1: Thanks, that's a merge!

PR #895, opened 2020-09-08
Proton Radiography Module

pheuer:
Proton radiography is an [increasingly common diagnostic technique](https://doi.org/10.1063/1.4750234) used to interrogate the electric and magnetic fields inside high energy density plasmas. The plasma of interest is positioned between a bright pulsed source of protons and a detector plane. Electric and magnetic fields in the plasma deflect the protons, producing patterns on the detector. Since this represents a non-linear and line-integrated measurement of the fields, the interpretation of these "proton radiographs" is a complicated business. The goal of this module is to provide tools for planning these experiments and analyzing their results.

Eventually, this new module should contain two type different types of functions:

1) Functions that construct synthetic proton radiographs given simulated or calculated fields, either for planning an experiment or analyzing the results of one.

2) Functions that use assumptions about symmetries etc. to "invert" experimental radiographs to recover the magnitudes of
the fields that created them.

The SyntheticProtonRadiograph object contained in this initial PR belongs to the first category. The SyntheticProtonRadiograph object is initialized with a `plasmapy.plasma.grid.AbstractGrid` subclass containing electric and magnetic fields, as well as the locations of the proton source and detector. A particle tracing algorithm is then run to determine the trajectories of the protons, and the resulting positions in the detector plane are used to create a synthetic radiograph (2D histogram). 

This function can currently support field data on either regular or non-uniform (although the former is much faster) to support complicated simulation meshes. Time steps are computed in an adaptive manner based on both the fields currently experienced by the particles and the grid resolution. Source and detector positions can be specified anywhere in Cartesian, cylindrical, or spherical coordinates systems to re-create geometric effects. 

The example fields generated within the function for testing and demonstration are intentionally simple, but this technique can re-create complicated structures, such as this off-axis radiograph of a cylindrical, radial electric field: 

SimPrad


StanczakDominik:
Whooaaaaa.

I'll take a closer look today, but for now I'd just like to say,

whooaaaaaaa.

StanczakDominik (commented):
Hey, sorry for the delay. Here's a partial review of just the notebook for now. I'll get to the code ASAP.

docs/notebooks/proton_radiography_particle_tracing.ipynb
```diff
@@ -0,0 +1,227 @@
+{
```
StanczakDominik, line 1: <p>exampels -&gt; examples</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

StanczakDominik, line 1: <p>Eventually I think it'd make a ton of sense to have <code>prad.test_fields</code>  return some sort of <code>Plasma</code>  object - if it's just a grid with an electric and magnetic fields... I think this might finally push me to reimplement  our gridded<code>Plasma</code> class through xarray. It'd simplify this nicely. Nothing to worry about; just pointing this out for future reference.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

StanczakDominik, line 1: <p>The formatting for the list is a bit off here. I guess you wanted to have a 3-element unordered list?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

StanczakDominik, line 1: <p>I would not be against actually calling it <code>SimulatedProtonRadiography</code> in the module! It's more explicit that way. @namurphy, @rocco8773, whatcha think ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

StanczakDominik, line 1: <p>It'd be nice to explain <code>max_theta</code>  (the angle variation) here as well. It just sort of pops out out of nowhere.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

StanczakDominik, line 1: <p>This is definitely going to be better with tqdm as a progress bar. I'll go back to that during code review.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

StanczakDominik, line 1: <p>This sounds confusing, I guess it'll make more sense once I've gone through the code...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

StanczakDominik, line 1: <p>I don't get the point of the <code>Grid</code>  scatter plot here. Is it meant just to display the origin of the grid?</p><p> </p><p>Also, the protons could definitely use a <code>label</code>.  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

StanczakDominik, line 1: <p>histograpm -&gt; histogram</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

StanczakDominik, line 1: <p><code>sim.synthetic_radiograph</code>could definitely use xarray outputs... I'll tinker with it a little.</p><p> </p><p>I guess <code>bins</code>  is the number of bins in the horizontal and vertical directions?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

StanczakDominik, line 1: <p>It'd be really cool to show radiographs from other example fields here, so that people can see that this doesn't just sample a circle!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/895/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

pheuer: <p>Fixed - apparently indenting it was the mistake</p>

pheuer: Sure, that sounds great! If there is a gridded plasma object that includes a spatial grid as well as arrays of E and B, that type could also be used as input for the SimPrad class.

pheuer: Good point - done

pheuer: It will - but that sort of defeats the point of an example I guess! I've added a brief summary of the steps taken by the algorithm to the notebook.

pheuer: Yes, that keyword is just directly passed on to np.histogram2d but I can provide more explanation here if that's not a commonly used fcn.

pheuer: Added a label and changed "Grid" to "Simulated fields" to be more clear.

What I _wanted_ to do was draw a little volume for the simulated fields grid, but that proved complicated and I judged it not worth the space. The point of this graphic is supposed to be just to show that the protons have traveled from the source through the simulated field grid to the detector, and that they all line up nicely in the detector plane at the end.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,928 @@
...
+
+    Parameters
+    ----------
+    grid : `~astropy.units.Quantity` array shape [nx,ny,nz,3], optional
+        A grid of positions on which to calculate fields. If no grid is
+        specified, one will be created.
+
+    mode : str, optional
+        The field model to load. The default represents the electrostatic
+        E-field created by a spherical Gaussian potential.
```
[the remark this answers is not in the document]
pheuer: I agree - maybe that's best handled by linking to the docs? For some reason I'm getting a build-the-docs error now: once that's resolved I'll add a link here.

docs/notebooks/proton_radiography_particle_tracing.ipynb
pheuer: I agree (and I need to include more good examples to showcase), but I do have one concern: running the code to create nice pictures for non-trivial examples take non-trivial computational resources, and I wasn't sure we wanted that happening on the back end every time the notebook is updated. Thoughts?

pheuer: I agree that would be nicer - the problem is that the simulation loop is actually a while loop, so I don't know ahead of time how many iterations to expect. Is there a way to modify tqdm to print the current number of particles on the grid, like I'm doing now, as a measure of progress?

StanczakDominik: In that case we can just run it once and remember not to clear the outputs for this particular notebook. That way, the notebook won't get executed on CI builds. That's exactly why I added nbsphinx, and I'm happy it's finally being useful!

pheuer: Perfect - I've added a couple more examples, then, and I will no longer clear the output so that it doesn't eat up time on CI.

pheuer:
@StanczakDominik  The failing Read the Docs build might be related to this open issue on sphinx: https://github.com/sphinx-doc/sphinx/issues/8198

namurphy:
Is this the error related to what @rocco8773 [mentioned on Element](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/$1599798873267478AMdOq:matrix.org?via=matrix.org&via=openastronomy.org&via=nitro.chat) recently about nbconvert?

One other thing is that we need to make sure that the [`api_static`](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/api_static) directory is up-to-date too, I think.  We need to improve the documentation on the `api_static` directory too, or find a way to automate it.

docs/notebooks/proton_radiography_particle_tracing.ipynb
namurphy: I generally support longer class names that are clear and unambiguous, so `SimulatedProtonRadiography` sounds good to me.

rocco8773:
> Is this the error related to what @rocco8773 [mentioned on Element](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/$1599798873267478AMdOq:matrix.org?via=matrix.org&via=openastronomy.org&via=nitro.chat) recently about nbconvert?

Yes, this is the same error.  I was hoping `nbconvert` would produce a patch, but nothing yet.  To get around this we need to add `nbconvert < 6.0.0` to our dependencies.  Should I open a PR for this?

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,955 @@
...
+
+        # Compute the timestep indicated by the grid resolution
+        # min is taken for irregular grids, in which different particles
+        # have different local grid resolutions
+        gridstep = (np.min(self.ds) / self.v0).to(u.s)
+
+        # If not, compute a number of possible timesteps
+        # Compute the cyclotron gyroperiod
+        Bmag = np.linalg.norm(B, axis=1)  # B is [nparticles,3] here
+        gyroperiod = (2 * np.pi * const.m_p.si / (const.e.si * np.max(Bmag))).to(u.s)
```
namurphy, line 677: I'm wondering if we'd want to make this a `plasmapy.formulary` function since I expect it'll come up fairly often.

rocco8773:
> One other thing is that we need to make sure that the [`api_static`](https://github.com/PlasmaPy/PlasmaPy/tree/master/docs/api_static) directory is up-to-date too, I think. We need to improve the documentation on the `api_static` directory too, or find a way to automate it.

I agree it would be nice to have `api_static` be dropped for a completely automatic setup that only populates `api`.  Unfortunately, the way `automodapi` is designed makes it non-trivial to implement that setup.  For us to get there we would need to design our own "documentation helpers."  I think that's low-priority at the moment and we can live with the minor inconvenience of `api_static`.

namurphy (commented):
> Yes, this is the same error. I was hoping `nbconvert` would produce a patch, but nothing yet. To get around this we need to add `nbconvert < 6.0.0` to our dependencies. Should I open a PR for this?

If you could, that would be much appreciated!

plasmapy/diagnostics/proton_radiography.py
rocco8773: It would probably be easier to do something like...

```python
gyroperiod = (1. / gyrofrequency(np.max(Bmag), particle="p", to_hz=True)).decompose()
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,955 @@
...
+    pradius = np.sqrt(xarr ** 2 + yarr ** 2)
+
+    # Create blank arrays for fields
+    E = np.zeros(grid.shape) * u.V / u.m
+    B = np.zeros(grid.shape) * u.T
+
+    if mode == "no fields":
+        pass
+
+    elif mode == "electrostatic gaussian sphere":
```
namurphy, line 164: I was reading really recently about a refactoring strategy on [replacing conditional statements with polymorphism](https://www.codementor.io/@uditrastogi/replace-conditional-statements-if-else-or-switch-with-polymorphism-ryl8mx4ns).  One difficulty with `if`/`elif`/`else` blocks is that, if there end up being a lot of options, they can lead to code that is harder to maintain (as I've been repeatedly discovering for another code I've been using with conditional blocks that are hundreds of lines long...and that I've added hundreds of lines to...).  The link above goes through an interesting example of how object-oriented programming allows us to avoid that with [polymorphism](https://www.educative.io/edpresso/what-is-polymorphism)...in particular, creating different classes to represent the different situations.  I'm wondering if that would be a worthwhile approach here.  Hm...

docs/notebooks/proton_radiography_particle_tracing.ipynb
rocco8773: I'm all for more explicit names!  My only issue is using "Simulation" outside the simulation sub-pkg, but I'll get over that.

namurphy:
Oh!  And I forgot to mention that when I first saw this pull request, it made my day.  Thank you for doing this!

plasmapy/diagnostics/proton_radiography.py
StanczakDominik: It comes up often enough that I wouldn't mind a thin wrapper for it, just for the convenience, even with a oneliner docstring: inverse gyro frequency, see here: gyro frequency.

docs/notebooks/proton_radiography_particle_tracing.ipynb
pheuer: How about `SyntheticProtonRadiograph`? I'd like to eventually include other methods of predicting radiographs from field grids to the same class (for example, there is a method using integration that works for linear small-deflection cases) and those methods are not all "simulations".

plasmapy/diagnostics/proton_radiography.py
pheuer: That sounds like a good solution to me: should we open a separate PR for that or can I just add that wrapper function as part of this PR?

plasmapy/diagnostics/proton_radiography.py
pheuer: This sounds like it could work: I'll play around with it.

rocco8773: If you want a python example, I basically did this with the fit functions for my `find_floating_potential()` in PR #889. The floating potential function allows you to choose two different curves to fit to the data.  There's a lot of common functionality needed between the two fits (fitted parameters, rsq, the func, a solver, etc.) but not all functionality is performed the same way.  For example, the solver and root finder for each curve use different methods. Thus, I package the fit functions into a set of parent, child classes that share a common API.  You can find the `FitFunctions` classes [here](https://github.com/rocco8773/PlasmaPy/blob/ad-lang-vf/plasmapy/analysis/swept_langmuir/fit_functions.py).  The "conditional selection" of the functions inside `find_floating_potential()` are [here](https://github.com/rocco8773/PlasmaPy/blob/e17b15aaa8a2f57e0f8c891c1caf2ded87415d9c/plasmapy/analysis/swept_langmuir/floating_potential.py#L146-L164).  And, the execution of the `FitFunctions` inside `find_floating_potential()` is [here](https://github.com/rocco8773/PlasmaPy/blob/e17b15aaa8a2f57e0f8c891c1caf2ded87415d9c/plasmapy/analysis/swept_langmuir/floating_potential.py#L296-L304).

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
StanczakDominik: I think Erik's yesterday pr should fix that. If you merge master in, it should work!

docs/notebooks/proton_radiography_particle_tracing.ipynb
StanczakDominik: I discussed this with Erik yesterday at the telecon; it's a decent enough idea but let's do it this way for now and adjust it later. :)

StanczakDominik: Sure, Radiograph sounds awesome. And I'm completely fine with it staying out of Simulation.

StanczakDominik: Yeah! I'll add that when I'm back to code review for this, okay? :)

StanczakDominik: Note to self: look at what fields do the particles see when outside the grid: also, find a civilized way to ask this question.

pheuer: Sure, that'd be great :)

pheuer: Great, class name updated to "SyntheticProtonRadiograph"

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
pheuer: Done, and seems to have worked - thanks!

plasmapy/diagnostics/proton_radiography.py
pheuer: @namurphy @rocco8773  Thanks for the suggestion and example: I've implemented something like this in the code so let me know what you think of the new structure.

rocco8773: I'll take a look at it.  I'm aiming to start a review later today.

docs/notebooks/proton_radiography_particle_tracing.ipynb
pheuer: The answer to that question is in this snippet in _push():

`E = self.E[self.xi, self.yi, self.zi, :] * np.outer(self.on_grid, np.ones(3))
        B = self.B[self.xi, self.yi, self.zi, :] * np.outer(self.on_grid, np.ones(3))`

'self.on_grid' is 1 for particles that are on the grid and 0 for particles that are off the grid. So, any particle off the grid sees zero field. 'self.on_grid' itself is calculated right above that: the distance from each particle to its interpolated grid point is calculated and the particle is "on the grid" if it's less than the local grid resolution 'self.ds'.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1203 @@
...
+
+
+class TestField:
+    """
+    Base class for example fields for testing SyntheticProtonRadiograph
+    functions.
+    """
+
+    def __init__(self, grid, emax=0 * u.V / u.m, bmax=0 * u.T, regular_grid=None):
+        self.grid = grid
```
rocco8773, line 34: Add doctrings.  You don't need to include a 2nd description of the class, just the parameter descriptions.

rocco8773 (changes requested):
This is a partial review of just he field generating classes.  I think this a good foundation, but we need to evolve it more.

1. @namurphy @StanczakDominik This field generating functionality should actually be placed within `plasmapy.plasma` based on our [PLEP7](https://github.com/PlasmaPy/PlasmaPy-PLEPs/blob/master/PLEP-0007.rst#plasma).  I think we should create a `plasmapy/plasma/fields.py` module and move this functionality there.

1. Docstrings need to be added to all the methods defined in the abstract class.  Note that if a sub-class does not override the docstrings, then the abstract class's docstrings will be used for the sub-class docstring.

1. The sub-class class level docstrings need to be expanded to include more detail on how the fields are calculated.

1. Methods `validate`, `gen_fields`, and `norm` don't feel like they're intended to be public facing API, as such, they should receive a leading underscore.

1. I'm surprised we haven't done this yet, but we should define the sphinx configuration variable `autoclass_content = "both"` in `/docs/conf.py`.  This will essentially combine the class level and `__init__` docstrings into one when generating the documentation.  This is needed in this case, so you don't have to re-document the parameter stuff when you override the class level docstring.

1. @pheuer Would you expect any of the abstract class variables (e.g. `radius`, `pradius`, `emax`, `bmax`, etc.) to be changed after the initial instantiation?  If so, they should probably be made into class properties that also includes regeneration of the fields when the value is changed.  For example,

```python
class AbstractField(ABC):
    def __init__(self, bmax, ...):
        self._bmax = bmax

    @property
    def bmax(self):
        return self._bmax

    @bmax.setter
    def bmax(self, val):
        self._bmax = val
        self.rerun_field_calc()
```

docs/notebooks/proton_radiography_particle_tracing.ipynb
StanczakDominik: <p>Clever! I like that :)</p>

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1203 @@
...
+
+    def validate(self):
+        if not self.regular_grid:
+            raise ValueError(
+                "ElectrostaticPlanarShock can only be created "
+                "on a regularly spaced grid."
+            )
+
+
+def test_fields(
```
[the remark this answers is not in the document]
pheuer: That makes sense - I'll change this function to "example_fields".

pheuer:
@rocco8773 Thanks for the review & lesson on abstract classes! I've added docstrings and underscores as you described.

The AbstractField subclasses are currently not user-facing (users interact with example_fields) but the setter methods for emax and bmax are a good idea so I've added them in anticipation of maybe moving this function to 'plasma/plasma/fields.py'. Changing the grid variable essentially means re-generating the entire field, so I think in that case the user should be directed to just instantiate a new object.

I think moving the Field classes to 'plasma/plasma.fields.py' makes a lot of sense: let me know once you've decided how you want that class structured and I can modify 'SyntheticProtonRadiograph' to take that object as input.

rocco8773:
Sorry for the delay in getting back to you.  After discussing during the community meeting, we decided that it's best to move the field stuff to the `plasmapy.plasma` sub-package.  It could most likely fit within a `fields.py` file for now, but if you feel like it deserves a "fields" sub-package within `plasmapy.plasma` then give it a shot.

Once you migrate the code, I'll give it another review.  I just don't want to give you suggestions/changes before the move.

pheuer:
@rocco8773 Great - I've moved the AbstractField class and subclasses over to fields.py

I also created an object for the field grid ("PosGrid" for now) that provides a natural home for many calculations done with the grid (all of which previously lived in proton_radiography.py but didn't really belong there). Let me know what you think of the structure etc. of this class.

rocco8773:
Cool!  I'll take a look at it.

@pheuer If you don't mind me pushing to your branch, I'll take care of the `index.rst` and any associated Sphinx magic.

pheuer:
@rocco8773 Sounds good to me!

plasmapy/plasma/fields.py
```diff
@@ -0,0 +1,700 @@
...
+
+        L = np.zeros([2, 3]) * length.unit
+        if length.size == 1:
+            L[:, 0] = np.array([-length.value, length.value]) * length.unit
+            L[:, 1] = np.array([-length.value, length.value]) * length.unit
+            L[:, 2] = np.array([-length.value, length.value]) * length.unit
+        elif length.size == 3:
+            L[:, 0] = np.array([-length[0].value, length[0].value]) * length.unit
+            L[:, 1] = np.array([-length[1].value, length[1].value]) * length.unit
+            L[:, 2] = np.array([-length[2].value, length[2].value]) * length.unit
```
rocco8773, line 261:
```suggestion
        if length.size == 1:
            L[:, 0] = [-1.0, 1.0] * length
            L[:, 1] = [-1.0, 1.0] * length
            L[:, 2] = [-1.0, 1.0] * length
        elif length.size == 3:
            L[:, 0] = [-1.0, 1.0] * length[0]
            L[:, 1] = [-1.0, 1.0] * length[1]
            L[:, 2] = [-1.0, 1.0] * length[2]
```

rocco8773 (changes requested):
This is a partial review of `PosGrid`.

1. I really think you should restrict kwarg `grid` to being an instance of `PosGrid`.  Why? (1) It can be really hard to condition `grid` otherwise. (2) Users should be forced into using the class API to ensure the grid it's formatted correctly.  My comments/suggestions in this review will reflect this point.

1. I'd like to see the functionality expanded so you can defined 1D, 2D, and 3D grids.  But this PR is already getting long, so that functionality would be better for a follow up PR.

1. I've made the suggestion a few times bellow that I think the keyword arguments should reflect the keyword arguments of `numpy.linspace`, since that is what is being used to generate the grid and it would be more familiar to users.

----

I realize this PR is starting to diverge significantly from its original intent.  How comfortable would you be at splitting off the field stuff into a separate PR?  We can focus on getting the field stuff complete and then come back to the proton radiography functionality.

plasmapy/plasma/fields.py
```diff
@@ -0,0 +1,700 @@
...
+        # Create a blank grid
+        grid = np.zeros([num[0], num[1], num[2], 3]) * u.cm
+
+        # If regulr_grid keyword is False, create a non-uniform grid with
+        # random point spacing in all directions
+        if not regular_grid:
+            for d in [0, 1, 2]:
+                ax = np.random.uniform(low=L[0, d].value, high=L[1, d].value, size=num)
+                ax = np.sort(ax, axis=d)
+                grid[..., d] = ax * L.unit
```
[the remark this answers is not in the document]
pheuer: This is a pretty important option for testing the proton radiography module: irregularly gridded fields are a commonly produced by Lagrangian fluid simulations, (or Eulerian codes with adaptive mesh refinement), so the ability to use that output without computationally expensive re-gridding is very helpful. If you have a better way of generating such a mesh I'm all ears though: this was a quick way of doing it. 

Implementing dimensions is a good idea. I'm not entirely clear what implementing non-Cartesian coordinate systems would entail: can you elaborate on that?

plasmapy/plasma/fields.py
```diff
@@ -0,0 +1,700 @@
...
+
+        if "length" in kwargs.keys():
+            length = kwargs["length"]
+        else:
+            length = 1 * u.cm
+
+        if "regular_grid" in kwargs.keys():
+            regular_grid = kwargs["regular_grid"]
+        else:
+            regular_grid = True
```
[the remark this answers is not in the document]
pheuer: I like that idea - would you agree that each keyword should be either a single value (applied across the number of dimensions) or a tuple of ndim values specifying separate values for each axis?

pheuer:
@rocco8773 Thanks for the edits! I agree that this Fields module is out of the scope of this PR (but necessarily must be merged before it). How about we finish addressing your suggestions above, then I will move it to its own PR and we can continue there? 

While we're separating things: I think there's a strong case for splitting PosGrid into its own file as well. The same framework would be useful for creating grids to hold any number of plasma parameters (densities, flow velocities, etc.) and should therefore
be its own entity.

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
pheuer: So, implementing non-Cartesian coordinate systems would involve a substantial change: self._grid could not be a u.Quantity because the different axes would have different units.

rocco8773:
> @rocco8773 Thanks for the edits! I agree that this Fields module is out of the scope of this PR (but necessarily must be merged before it). How about we finish addressing your suggestions above, then I will move it to its own PR and we can continue there?

Agreed! Let's finish up the current comments on `PosGrid` and then split it off in a new PR to continue.  It will best to have most of the reviews in the new PR so the discussion isn't "lost."

If you don't field like handling pkg/documentation elements,  then just split of the `fields.py` and open a new PR.  Once that is done, I can go through and cherry pick the pkg/documentation commits.

rocco8773:
> While we're separating things: I think there's a strong case for splitting PosGrid into its own file as well. The same framework would be useful for creating grids to hold any number of plasma parameters (densities, flow velocities, etc.) and should therefore be its own entity.

I agree with this too.  I'm not sure where to place it because I can see it being used in all sorts of places without any clear winner for its home.  Most logical to me seems to be in the simulation package.  So...how to proceed?  I think we should just keep it where it is and then when we expand the "Grid" functionality we can find it's permanent home.

pheuer:
@rocco8773 I think it makes sense to have it in the Plasma package (in a grids.py file at the same level as fields.py): then the way it could enter into the simulation package would be as part of the input plasma for the simulation? 

Put another way: simulations contain plasmas, but plasmas don't contain simulations.

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
rocco8773: Agreed.  To ensure we're on the same page then...

1. if `start` and `stop` are single values then they'll be applied to all dims.
    - `num` would then have to be single valued or a 3-elemnet array_like object
1. If `start` and `stop` are not single valued.
    - They must be a 3-element array_like object.
    - if `num` is single valued then it'll will be applied to all dims
    - if `num` is NOT single valued then it must be a 3-element array_like object

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
rocco8773: > So, implementing non-Cartesian coordinate systems would involve a substantial change: self._grid could not be a u.Quantity because the different axes would have different units.

Fair.  However I feel the discuss is till warranted. But the will be in the future for another PR.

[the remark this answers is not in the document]
rocco8773: > This is a pretty important option for testing the proton radiography module: irregularly gridded fields are a commonly produced by Lagrangian fluid simulations, ...

I see.  I'll need to think about this more then.  My first concern is with using `np.sort` because you risk shrinking `ax`.  `np.sort` performs a sort and a unique operation.  This is not my forte, but @namurphy might have some good suggestions.

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
pheuer: I think I agree: isn't that equivalent to:

Start, stop and num may each be either a 3-element array or a single value. If they are single_values, they are expanded like np.array( [single_value]  * 3).

This is a little more complicated in the case of supporting 1D and 2D arrays: we need to decide if there are actually use-cases for that.

rocco8773:
Sure, but simulations will often require gridding/meshing and then population of said grid. That's essentially what `PosGrid` is doing for position values. In the future I see a class specifically for grid creation and then subclasses or user classes to create and populate.

With all that said, I'm fine with putting `PosGrid` in a `plasmapy/plasma/grid.py` file.

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
rocco8773: I think there will be, but I'm ok with leaving it for a future PR. I fully expect this functionality to expand and evolve as the need arises. I'm most concerned with getting a nice initial product so we can get back to your proton stuff.

pheuer:
@rocco8773 Great: please see PR #909 .

plasmapy/plasma/fields.py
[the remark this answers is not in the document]
pheuer: I tested this and it seems to be that `np.sort` does not return a unique list? I'm currently still using that over in #909 but open to other suggestions.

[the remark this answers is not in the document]
rocco8773: @pheuer You're right. I was mistaken.

pheuer:
@StanczakDominik I added a `tqdm` progress meter as you suggested (instead of the print() explosion I had before), but it looks like CI doesn't have `tqdm` installed? I get a bunch of errors like this: 

```
ERROR ../.tox/py38-numpydev/lib/python3.8/site-packages/plasmapy/diagnostics/proton_radiography.py - ModuleNotFoundError: No module named 'tqdm'
```

StanczakDominik:
The current failures look intermittent. Looks like you added tqdm properly to dependencies; I'll take a look after taking this increasingly annoyed doggo out, if you don't mind!

pheuer:
@StanczakDominik Sounds good, thanks!

StanczakDominik:
Looks like it's working now :) I cherry-picked a minor config fix from #996 to remove the intermittent issues, and it seems you've indeed correctly added tqdm. So that should be fine :)

pheuer:
@StanczakDominik  Thanks! Any insight on the test-documentation error? Do I have a docstring somewhere it's not supposed to be?

```
/root/project/plasmapy/diagnostics/proton_radiography.py:docstring of plasmapy.diagnostics.proton_radiography:1:duplicate object description of plasmapy.diagnostics.proton_radiography, other instance in ad/diagnostics/proton_radiography, use :noindex: for one of them
```

StanczakDominik:
Looks like we've got a duplicate .rst file:

* `docs/diagnostics/proton_radiography.py`
* `docs/ad/diagnostics/proton_radiography`

StanczakDominik:
By the way, we've just (#996) transitioned to github actions from circleci, so you might wanna merge master :)

pheuer:
OH, thank you! I still don't understand the structure of the docs directories...

StanczakDominik:
I think I'm just as surprised as you were about that! Must have been something born of Erik's diagnostics suite, which I haven't had much time to follow recently :(

pheuer:
@StanczakDominik @rocco8773 I've marked this as 'ready to review` because it is, but since this PR depends on #981, that one definitely needs to be reviewed and merged first. 

I know this is a big commit, so maybe I can go through it at our next LineIntegratedDiagnostics meeting to get you oriented?

Also, just FYI, the style linters wanted all sorts of corrections to my .ipynb notebooks that pre-commit doesn't seem to touch. I eventually solved the problem using [this helpful package](https://pypi.org/project/black-nbconvert/), but is it supposed to be doing that?

StanczakDominik:
That's a result of just having merged nbqa into pre-commit and not having corrected the docs, sorry...

Basically, I'm pushing for people to actually start using pre-commit with a configuration mirroring the CI one so that these formatting issues automatically solve themselves, and we don't need to pay attention to them.

I'll expand the docs and maybe the PR template...

pheuer:
@rocco8773  and @StanczakDominik I think I've made all of the changes we discussed in our meeting. A couple notes:

- I haven't implemented the different time steps in `coast_to_grid` for each particle: that's a bit complicated and represents only a small time savings, so I'd like to leave it for a future PR. 
- I wasn't sure how to add the setup fig I made: right now it's just sitting in the notebook directory.
- I can put together a separate PDF explaining the algorithm if you think that's advisable, but I think the notebook does a pretty good job? Let me know if you think a separate PDF is necessary (separate from improvements to the notebook).

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -0,0 +1,340 @@
...
+    sim with a quasi-1D field grid that can then be summed to get good
+    enough statistics to use as a test.
+    """
+    grid = _test_grid(name, L=1 * u.mm, num=50)
+
+    # Cartesian
+    source = (0 * u.mm, -10 * u.mm, 0 * u.mm)
+    detector = (0 * u.mm, 200 * u.mm, 0 * u.mm)
+
+    # Catch warnings (test fields aren't well behaved at edges)
```
StanczakDominik, line 125: Should that be a TODO entry?
```suggestion
    # TODO: Catch warnings (test fields aren't well behaved at edges)
```

I like leaving these explicitly in the code, makes grepping easier :)

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -0,0 +1,340 @@
...
+
+    sim = prad.SyntheticProtonRadiograph(grid, source, detector, verbose=False)
+    sim.create_particles(1e4, 3 * u.MeV, max_theta=0.1 * u.deg)
+    sim.run()
+
+    size = np.array([[-1, 1], [-1, 1]]) * 10 * u.cm
+    bins = [200, 60]
+    hax, vax, values = sim.synthetic_radiograph(size=size, bins=bins)
```
StanczakDominik, line 133: this implementation is officially a hack now :stuck_out_tongue_winking_eye:

plasmapy/plasma/grids.py
```diff
@@ -1118,12 +1118,23 @@ def volume_averaged_interpolator(
...
     spaced grid.
     """
 
+    def _validate(self):
+        # Check that all units are lengths
+        for i in range(3):
+            try:
+                self.units[i].to(u.m)
+            except u.UnitConversionError:
+                raise ValueError(
```
StanczakDominik, line 1133: Could do with a test for this one! If it's not too much effort, that is.

StanczakDominik (commented):
Partial review; not much of note yet, haven't dug through the main dish of `proton_radiography.py`, but I wanted to communicate I'm doing a nonzero amount of work on this topic! :sweat_smile: 

The notebook looks great, as well! I also take it this is going to be one of the notebooks that carries execution outputs with it, right? Perfectly fine with me. We could probably give them a `_with_output.ipynb` suffix, so we remember to rerun them some of the time to ensure they're still working as expected...

plasmapy/diagnostics/tests/test_proton_radiography.py
pheuer: Does this satisfy the 1 pun/1000 lines of code requirement, or do I need a more obvious one? :P

plasmapy/plasma/grids.py
pheuer: Oops, not at all (already had it implemented for CartesianGrid)

pheuer:
@StanczakDominik I've cleared the output on the notebook and it ran in a pretty reasonable amount of time (a few minutes). That's comparable to some of the tests, so maybe it's easiest to just let it run?

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+                RuntimeWarning,
+            )
+
+    # *************************************************************************
+    # Run/push loop methods
+    # *************************************************************************
+
+    def _adaptive_dt(self, Ex, Ey, Ez, Bx, By, Bz):
+        r"""
+        Calculate the appropraite dt based on a number of considerations
```
StanczakDominik, line 550:
```suggestion
        Calculate the appropriate dt based on a number of considerations
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+
+        # Calculate the remaining distance each particle needs to travel
+        # along that unit vector
+        remaining = np.dot(self.source, uvec)
+
+        # Calculate the time remaining to reach that plane and push
+        t = (dist - remaining) / np.dot(self.v, uvec)
+
+        # Calculate the particle positions for that case
+        self.x0 = self.source + self.v * np.outer(t, np.ones(3))
```
StanczakDominik, line 621: Since v already has the required dimensions, you can just broadcast:
```suggestion
        self.x0 = self.source + self.v * t[:, np.newaxis]
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+        v_towards_det = np.dot(self.v, -self.det_n)
+
+        # Time remaining for each particle to reach detector plane
+        t = dist_remaining / v_towards_det
+
+        # If particles have not yet reached the detector plane and are moving
+        # away from it, they will never reach the detector.
+        # So, we can remove them from the arrays
+        condition = np.logical_and(v_towards_det < 0, dist_remaining > 0)
+        ind = np.nonzero(np.where(condition, 0, 1))[0]
```
StanczakDominik, line 642: If we want to pick particles for which both conditions are true, this simplifies to...

```suggestion
        ind = (v_towards_det < 0) & (dist_remaining > 0)
```
And boolean indexing does the rest!

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+        if self.fract_deflected > 0.05:
+            warnings.warn(
+                f"{100*self.fract_deflected:.1f}% particles have been "
+                "deflected away from the detector plane. The fields "
+                "provided may be too high to successfully radiograph "
+                "with this particle energy.",
+                RuntimeWarning,
+            )
+
+        self.x += self.v * np.outer(t, np.ones(3))
```
StanczakDominik, line 661:
```suggestion
        self.x += self.v * t[:, np.newaxis]
```
Definitely don't do `outer` here, that incurs additional multiplications.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+
+        # Check that all points are now in the detector plane
+        # (Eq. of a plane is nhat*x + d = 0)
+        plane_eq = np.dot(self.x, self.det_n) + np.linalg.norm(self.detector)
+        assert np.allclose(plane_eq, np.zeros(self.nparticles_grid), atol=1e-6)
```
StanczakDominik, line 666:
```suggestion
        assert np.allclose(plane_eq, 0, atol=1e-6)
```
pretty sure this should work! ;)

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+        self.v[self.grid_ind, :] = v
+
+    def _stop_condition(self):
+        r"""
+        The stop condition is that most of the particles have entered the grid
+        and almost all have now left it.
+        """
+        # Count the number of particles who have entered, which is the
+        # number of non-zero entries in entered_grid
+        self.num_entered = np.nonzero(self.entered_grid)[0].size
```
StanczakDominik, line 723: If `self.entered_grid` is an array of booleans, you can just do `.sum()`, no? Oh right, no, this is the counter for how many times each one has entered the grid. This looks fine, then.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+        # Throw a warning if < 50% of the particles are included on the
+        # histogram
+        percentage = 100 * np.sum(intensity) / self.nparticles
+        if percentage < 50:
+            warnings.warn(
+                f"Only {percentage:.2f}% of the particles are shown "
+                " on this synthetic radiograph. Consider increasing "
+                " the size to include more.",
+                RuntimeWarning,
+            )
```
StanczakDominik, line 1006: This is one of my new favorite tricks:

```suggestion
        # Throw a warning if < 50% of the particles are included on the
        # histogram
        percentage = np.sum(intensity) / self.nparticles
        if percentage < 0.5:
            warnings.warn(
                f"Only {percentage:.2%} of the particles are shown "
                " on this synthetic radiograph. Consider increasing "
                " the size to include more.",
                RuntimeWarning,
            )
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+
+        if optical_density:
+            # Generate the null radiograph
+            x, y, I0 = self.synthetic_radiograph(size=size, bins=bins, ignore_grid=True)
+
+            # Calcualte I0 as the mean of the non-zero values in the null
```
StanczakDominik, line 1012:
```suggestion
            # Calculate I0 as the mean of the non-zero values in the null
```

StanczakDominik (commented):
Phew, done! Looks great! I mostly have a bunch of minor fixes and suggestions. Once we've got those out of the way, it's smooth sailing to merge :)

plasmapy/diagnostics/proton_radiography.py
pheuer: I didn't know that was an option!

plasmapy/diagnostics/proton_radiography.py
pheuer: Yes, your understanding there is right.

plasmapy/diagnostics/proton_radiography.py
pheuer: Not sure why I wasn't doing that earlier, but seems to be working!

plasmapy/diagnostics/proton_radiography.py
pheuer: Ooooooooooohhhh

I KNEW there had to be a better way to do this! Thanks!!

plasmapy/diagnostics/proton_radiography.py
pheuer: Hm, for some reason I didn't think these logical operators worked element-wise on arrays...but I guess they do! 

The actual one-line replacement should be 
```
ind = np.logical_not((v_towards_det < 0) & (dist_remaining > 0)).nonzero()[0]
```
though, since `ind` should be the indices of all of the values where the statement is False.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+
+        # Also generate a uniform phi distribution
+        phi = np.random.uniform(size=self.nparticles) * 2 * np.pi
+
+        return theta, phi
+
+    def _angles_uniform(self):
+        """
+        Generates angles for each particle such that their velocities are
+        uniformly distributed on a grid in theta and phi.
```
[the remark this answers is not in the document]
pheuer: Oh, good point, I've added this to the docstring for create_particles:

```
This method
        requires that `nparticles` be a perfect square. If it is not, 
        `nparticles` will be set as the largest perfect square smaller
        than the provided `nparticles`.
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+            Units must be convertable to radians.
+
+        charge : `~astropy.units.Quantity`
+            The charge of the particle, in units convertable to Columbs.
+            The default is the proton charge.
+
+
+        mass : `~astropy.units.Quantity`
+            The mass of the particle, in units convertable to kg.
+            The default is the proton mass.
```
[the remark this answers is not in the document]
pheuer: Yes you're right, that's much cleaner. I'll do that.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+                        np.dot(vec, self.src_det)
+                        / np.linalg.norm(vec)
+                        / np.linalg.norm(self.src_det)
+                    )
+                    ind += 1
+        return np.max(theta)
+
+    def _log(self, msg):
+        if self.verbose:
+            print(msg)
```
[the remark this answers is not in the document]
pheuer: Sounds like a good idea to me! I'll have to learn more about the `logging` module.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+                unit = self.grid._recognized_quantities[rq].unit
+                arg = {rq: np.zeros(self.grid.shape) * unit}
+                self.grid.add_quantities(**arg)
+
+            # Check that there are no infinite values
+            if not np.isfinite(self.grid[rq]).all():
+                raise ValueError(
+                    f"Input arrays must be finite: {rq} contains "
+                    "either NaN or infinite values."
+                )
```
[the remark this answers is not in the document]
pheuer: Interesting question: maybe? I don't think we should necessarily prevent all grids from having infinite values or any set of required quantities. But I could add methods to `grid.AbstractGrid` that checks for these conditions and returns a Boolean, then use those functions here to raise the exceptions?

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,1022 @@
...
+
+        if charge is not None:
+            self.q = charge.to(u.C).value
+        else:
+            self.q = self._e
+
+        if mass is not None:
+            self.m = mass.to(u.kg).value
+        else:
+            self.m = self._m_p
```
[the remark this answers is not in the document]
pheuer: Done

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+            )
+
+            if edge_max > 1e-3 * np.max(arr):
+                unit = grid.recognized_quantities[rq].unit
+                warnings.warn(
+                    "Fields should go to zero at edges of grid to avoid "
+                    f"non-physical effects, but a value of {edge_max:.2E} {unit} was "
+                    f"found on the edge of the {rq} array. Consider applying a "
+                    "envelope function to force the fields at the edge to go to "
+                    "zero.",
```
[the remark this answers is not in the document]
pheuer: I guess, if you have the ability to expand the grid to cover a larger area? I was looking at it from the perspective of having a grid of positions that is already specified.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+        # Vector directly from source to detector
+        self.src_det = self.detector - self.source
+
+        # Magnification
+        self.mag = 1 + np.linalg.norm(self.detector) / np.linalg.norm(self.source)
+        self._log(f"Magnification: {self.mag}")
+
+        # Check that source-detector vector actually passes through the grid
+        if not self.grid.vector_intersects(self.source * u.m, self.detector * u.m):
+            raise ValueError(
```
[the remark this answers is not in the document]
pheuer: Yes that's quite a good idea. Could just be a generalized `GeometryError` since the same sort of error could come up in simulations?

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
StanczakDominik: Yeah, that's what I meant to and failed to communicate, sorry! :D

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
StanczakDominik: But the grid, as i understand it, is mostly a computational object onto which you project your simulated or experimental fields, so it would not in principle sound impossible to make the grid larger?

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+    The algorithm is based on this discussion on StackExchange:
+    https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d/476311#476311
+    """
+    # Normalize both vectors
+    a = a / np.linalg.norm(a)
+    b = b / np.linalg.norm(b)
+
+    # Manually handle the case where a and b point in opposite directions
+    if np.dot(a, b) == -1:
+        return -np.identity(3)
```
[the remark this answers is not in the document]
pheuer: I think I've added tests to address most of these case now :)

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -0,0 +1,941 @@
...
+        quantity will be assumed to be zero everywhere.
+
+    source : `~astropy.units.Quantity`, shape (3)
+        A vector pointing from the origin of the grid to the location
+        of the proton point source. This vector will be interpreted as
+        being in either cartesian, cylindrical, or spherical coordinates
+        based on its units. Valid geometries are:
+        * Cartesian (x,y,z) : (meters, meters, meters)
+        * cylindrical (r, theta, z) : (meters, radians, meters)
+        * spherical (r, theta, phi) : (meters, radians, radians)
```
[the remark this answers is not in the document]
pheuer: Actually I solved it! Looks like Sphinx just wants a blank line before and after the list:

```
based on its units. Valid geometries are:

        * Cartesian (x,y,z) : (meters, meters, meters)
        * cylindrical (r, theta, z) : (meters, radians, meters)
        * spherical (r, theta, phi) : (meters, radians, radians)

        In spherical coordinates theta is the polar angle.
```

plasmapy/diagnostics/proton_radiography.py
StanczakDominik: I learned it a few days ago! :D

docs/notebooks/proton_radiography_particle_tracing.ipynb
```diff
@@ -310,7 +312,7 @@
    "outputs": [],
    "source": [
     "max_deflection = sim.max_deflection\n",
-    "print(f\"Maximum deflection α = {np.rad2deg(max_deflection):.2f}\")"
+    "print(f\"Maximum deflection ÃŽÂ± = {np.rad2deg(max_deflection):.2f}\")"
```
StanczakDominik, line 315:
```suggestion
    "print(f\"Maximum deflection α = {np.rad2deg(max_deflection):.2f}\")"
```

docs/notebooks/proton_radiography_particle_tracing.ipynb
```diff
@@ -331,7 +333,7 @@
     "mu = l * max_deflection.value / a\n",
     "print(f\"a = {a}\")\n",
     "print(f\"l = {l:.1f}\")\n",
-    "print(f\"μ = {mu:.2f}\")"
+    "print(f\"ÃŽÂ¼ = {mu:.2f}\")"
```
StanczakDominik, line 336:
```suggestion
    "print(f\"μ = {mu:.2f}\")"
```

StanczakDominik (approved):
LGTM! Something got messed up with unicode in the notebook, which is weird; but we can merge it with these two fixes :)

StanczakDominik:
This feature is large enough that hey, might warrant 0.6 in a few weeks' time (I'm swamped until then)!

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
pheuer: Ok, so I just did this for checking required quantities (`grid.require_quantities`) since that seemed like a pretty common problem. 

Upon further reflection, I'm not convinced that checking for infinite values or large value boundaries deserve their own functions in grids yet? Just weighing the value of moving the code there vs. the cost of adding even more user-facing methods to clog up the documentation?

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
pheuer: Fair enough - if you have the option (eg. re-run a bigger simulation or create a larger analytical field) then that would work too. I've changed the text accordingly: 

```
"Fields should go to zero at edges of grid to avoid "
                    f"non-physical effects, but a value of {edge_max:.2E} {unit} was "
                    f"found on the edge of the {rq} array. Consider using a "
                    "larger grid (with lower fields at the edges), or applying an "
                    "envelope function to force the fields at the edge to go to "
                    "zero.",
```

pheuer:
@StanczakDominik YAY! *clinks virtual champagne* Thanks for all your help!

Getting this functionality onto the release branch sounds great to me: I have lots of people I'm planning on sharing it with, but it's definitely easier for people to install from release!

StanczakDominik:
Sure, but unless you wanna try handling the release... hope you can wait!

I should be able to get it out between March 6 and 14.

pheuer:
Hahaha no no no sometime in March sounds good to me!

PR #909, opened 2020-10-02
Grids

pheuer:
This PR creates a new sets of objects within the 'plasmapy.plasma' package:

Child classes of 'grids.AbstractGrid' represent 3D grids of positions in various coordinate systems. Further subclasses can provide special types of grids - for example 'IrregularCartesianGrid' is a non-uniformly spaced Cartesian grid for testing post-processing scripts meant for irregularly gridded simulation outputs.

This PR originated as part of the proton radiography module (PR #895 ) and must be merged before that module can be finalized. In the meantime, that PR serves as an example of one usecase for these objects. The same grid objects could potentially also be useful input forms for simulations, etc.

- [x] Add methods for calculating volume-averaging values over the grid at a position (at least for CartesianGrid with uniform spacing...)


pheuer:
This is currently just a rough draft - I'd welcome any comments on the structure or features included.

rocco8773:
Ok, I migrated over all the old documentation items, added documentation for `grids.py`, and fixed some PEP8 styling issues.

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    Abstract grid represents a 3D grid of positions. The grid is stored as an
+    np.ndarray, while the units associated with each dimension are stored
+    separately.
+    """
+
+    # TODO: add appropriate typing here on start, stop, num
+    # start/stop can be -> number(int,float), ndarray, u.quantity, list of numbers
+    # or u.quantities units can be astropy.units.core.Unit or list of same
+    def __init__(self, start=0, stop=1, num=100, grid=None, units=None):
```
rocco8773, line 25: `start` and `stop` should be required arguments and, thus, no default values.

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,330 @@
...
+        else:
+            self._units = [u.dimensionless_unscaled] * 3
+
+        self._grid = grid
+
+        # Check to make sure that the object created satisfies any
+        # requirements: eg. units correspond to the coordinate system
+        self._validate()
+
+    def _make_grid(self, start=0, stop=1, num=100):
```
rocco8773, line 148: Some comment as I made for `__init__`, `start` and `stop` should be required arguments.

rocco8773 (changes requested):

plasmapy/plasma/grids.py
pheuer: As currently written they cannot be required because they should be optional in the case that the 'grid' keyword is set (loading a user-created grid rather than creating a new one). Should we have them all default to None then enforce this via a check, or is there a better way?

rocco8773: How about something like...

```python
def __init__(self, *seeds, num=100, units=None, **kwargs):
    if len(seeds) == 1:
        # assume grid
        self._load_grid(seeds[0], units=units)
    elif len(seeds) == 2:
        # assume start-stop
        self._make_grid(seeds[0], sees[1], num=num, **kwargs)
    else:
        raise TypeError(f"{self.__class__.__name}() takes 1 or 2 positional arguments but {len(seeds)} were given")

def _load_grid(self, grid, units=None):
    if not isinstance(grid, self.__class__):
        raise VelueError(f"Expected instance of {self.__class__} and got {grid.__class__}."

def _make_grid(self, start, stop, num=100, **kwargs):

```
Notes:  I did drop the `grid` kwarg on purpose and the `grid` input to `_load_grid` should be a required argument.

pheuer: Sounds good: done

pheuer:
@rocco8773 I've finished implementing the four grid types I originally planned and updated `fields.py` to match. There are a few very specific functions I'm using in the proton radiography module (such as nearest_neighbor and vector_intersects_grid) that could technically be methods of the CartesianGrid class, but right now I'm leaning towards just including them in the proton radiography code directly since they seem unlikely to be used elsewhere.

What do you think?

pheuer:
@rocco8773 Actually, maybe I'm not quite done with this... I think it makes sense to incorporate some more things like the nearest_neighbor interpolator (added now) and maybe some methods for calculating volume-averaging coefficients. Those sorts of things really only involve the grid and will potentially be useful in lots of simulation/synthetic diagnostic contexts.

rocco8773:
@pheuer Sounds good!  I'll take a look at it.  I'm currently in the middle of a collaboration meeting, so I'm unlikely to be able to do a thorough review until after Wednesday.

plasmapy/plasma/fields.py
```diff
@@ -0,0 +1,326 @@
...
+        grid : `plasmapy.plasma.grids.AbstractGrid` or child class thereof
+            Positions of grid points in space.
+
+        emax : `~astropy.units.Quantity`
+            Maximum electric field to normalize example field to. The default
+            is 0 * u.V / u.m.
+
+        bmax : `~astropy.units.Quantity`
+            Maximum magnetic field to normalize the example field to. The
+            default is 0 * u.T.
```
StanczakDominik, line 44: Suppose I didn't understand what this did and passed in negative values, here? Would it break?

plasmapy/plasma/fields.py
```diff
@@ -0,0 +1,326 @@
...
+    blob of potential with a Gaussian radial distribution:
+
+    .. math::
+        \phi = e^{-(r/a)^2}
+    .. math::
+        E = -\nabla \phi \text{, for }  (r < L/2)
+    .. math::
+        E = 0 \text{, for } (r < L/2)
+    .. math::
+        B = 0
```
StanczakDominik, line 166: I'm no .rst wizard, but - couldn't this go into a single `.. math::` block?

StanczakDominik (commented):
Partial review! A bit of refactoring in suggestions. I still need to look at `grids.py`.

plasmapy/plasma/grids.py
StanczakDominik: Then we can get rid of the TODO note, right?

StanczakDominik (commented):
I'm looking into the `grids` functionality atm, and I'm concerned that this seems to reimplement a whole lot of functionality that xarray already provides. I'm not sure I feel comfortable merging that in and maintaining essentially the same functionality they do. I was not previously aware of the scope this has.

What would you say if I were able to rewrite a not insignificant portion of `grids` using xarray?

cc @rocco8773 for a second opinion!

plasmapy/plasma/tests/test_fields.py
```diff
@@ -0,0 +1,39 @@
...
+    model = fields.ElectrostaticGaussianSphere(gridobj, emax=1e9 * u.V / u.m)
+    model.emax = 2e9 * u.V / u.m
+    model.bmax = 1 * u.T
+    a = model.emax
+    a = model.bmax
+
+    E, B = fields.example_fields(gridobj, model="electrostatic planar shock")
+
+
+test_example_fields()
```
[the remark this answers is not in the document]
pheuer: Oops, I just forgot to delete this line!

plasmapy/plasma/fields.py
pheuer: I think it would just give you back -B or -E: the routine just normalizes the array to the maximum value then multiplies by emax and bmax respectively.

plasmapy/plasma/fields.py
pheuer: I had it like that initially and had some formatting problems. It makes sense it should work that way, though. If you're not quite an `.rst` wizard I'm more like an `.rst` gnome or something.

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,896 @@
...
+            self._make_grid(seeds[0], seeds[1], num=num, units=units, **kwargs)
+
+        else:
+            raise TypeError(
+                f"{self.__class__.__name__} takes 1 or 2 "
+                f"positional arguments but {len(seeds)} were given"
+            )
+
+        # Initialize some variables
+        self._uniform_grid = None
```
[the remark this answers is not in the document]
pheuer: Yes, that's definitely better!

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,896 @@
...
+        if hasattr(pos, "unit"):
+            pos2 = pos.si.value
+        else:
+            for i in range(3):
+                pos2[:, i] = (pos[:, i] * self.units[i]).si.value
+
+        # Interpolate indices
+        i = self.interpolator(pos2)
+        # Convert any non-NaN values to ints
+        i = i.astype(np.int32)
```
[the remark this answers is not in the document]
pheuer: Oh that comment may be out of date - previously I was filling out-of-bound values in the interpolator with NaN but then running into this issue. How would you suggest handling this? 

I think the ideal solution would be to fill with np.nan but then convert everything else into int32?

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,896 @@
...
+    @property
+    def ax2(self):
+        """Axis 3"""
+        return self._grid[0, 0, :, 2] * self.unit2
+
+    @property
+    def dax0(self):
+        """
+        Grid step size along axis 1
+        Only valid if grid is uniform: otherwise an exception is raised
```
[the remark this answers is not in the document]
pheuer: Yes, that's true. We have to weigh that against potential confusion if a user expects an array but gets a single value when calling for a uniform array.

pheuer:
@StanczakDominik I was not previously aware of xarray, but I agree there seems to be a lot of overlap. If you can refactor this to be more concise by employing xarray I think that's a great idea!

 I will note that although xarray does have interpolation (calling the scipy module), the way I've implemented it here is optimized for getting values from multiple other arrays (eg. E and B) that match the grid, so I think that functionality is added.

plasmapy/plasma/grids.py
```diff
@@ -0,0 +1,896 @@
...
+            self.ax2.si.value,
+        )
+
+        # Create the interpolator
+        # Note that this regular grid interpolator requires points be a regular
+        # grid but does not require uniform spacing
+
+        self._interpolator = interp.RegularGridInterpolator(
+            pts, indgrid, method="nearest", bounds_error=False, fill_value=np.nan
+        )
```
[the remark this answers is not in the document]
namurphy: !!!!!!!!!!

If we require Python 3.8+ then we'd no longer be following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)...but this is a pretty compelling use case given that we could potentially use this feature a lot, and in the development series I'm less concerned about supporting older versions of Python.  In the excitement about walrus operators I missed this new feature!

StanczakDominik:
That's true, but I think we could handle that - I think if we store the Fields object as an xarray.DataSet, with the current Grids as, basically, its coords field, we can handle that np.

I mean no problem, wow, the dot at the end really changes the optics of that sentence.

I'll try to start tinkering on that tomorrow.

StanczakDominik:
Here's where I'm at with a little experimenting in xarray:

https://gist.github.com/StanczakDominik/15dc0fd81495e1702535716169fd3c19

It renders better locally.

I can easily create field objects called `grid.Ex` through `grid.Ez`; I'm not yet sure how to elegantly stack them into a shared `grid.E`.

pheuer:
@StanczakDominik This looks very good! Can you initialize with a full grid (rather than with 1D vectors) so it can in theory support unstructured grids?

Reporting vector quantities as three scalar quantities might be ok: it would keep things simple when we have both scalar and vector quantities on the grid (eg. density and E field).

I like that the quantities (eg. E, B) are part of the grid object: that would provide a neat interface for diagnostics that require a bunch of quantities (prad with solid interactions for example would require E, B, and solid density).

StanczakDominik:
> Can you initialize with a full grid (rather than with 1D vectors) so it can in theory support unstructured grids?

AFAIK no, but I know fairly little about unstructured. Might be possible if we meshgrid, then unravel, I guess? Or not. I'm not sure.

> Reporting vector quantities as three scalar quantities might be ok: it would keep things simple when we have both scalar and vector quantities on the grid (eg. density and E field).

That might indeed make sense, since then in other cases we can have Etheta, Bphi... depending on the coordinate system.

---

One idea I've had that didn't go into the notebook - the current split into Cartesian, Spherical, etc... we could really go by the labels in the `coords` field for the dataset, and that's sufficient to identify it!

pheuer:
> > Can you initialize with a full grid (rather than with 1D vectors) so it can in theory support unstructured grids?

The main reason I ask is that I want routines like SyntheticProtonRadiograph to be able to take input directly from a simulation (presumably a set of arrays) and turn it into a grid object behind the scenes, eg.
```
if not isinstance(grid, AbstractGrid): 
    grid = AbstractGrid(grid)
```
I guess we could consider allowing only uniformly-spaced grids: that certainly makes higher-order interpolation schemes easier too...


> One idea I've had that didn't go into the notebook - the current split into Cartesian, Spherical, etc... we could really go by the labels in the `coords` field for the dataset, and that's sufficient to identify it!

Yes this is a good idea: however they should probably still have separate sub-classes because some methods might be instantiated differently in one vs. the other. For example, I've implemented a volume-weighted interpolator for CartesianGrid, but the same function would be different on a CylindricalGrid or SphericalGrid.

pheuer:
@rocco8773 Looking at this and the `Plasma3D` class, I think there's a lot of overlap (which makes sense, since @StanczakDominik was using Plasma3D as input to the current ParticleTracker module). 

Maybe it makes sense to have Grids hold a dictionary of different properties stored on them (E, B, n, T, etc.) and then have that basically replace the framework of Plasma3D.

StanczakDominik:
I'll get back to you on this tomorrow. Things happened and I'm out protesting, lol.

rocco8773:
I would side with what @StanczakDominik thinks is appropriate.  With that said...My intuition tells me to keep grids and fields as independent functionality, where grids is the parent class to the "fields" classes.  Each "field" class (i say "field" because this can be extended to temperatures, densities, etc.) represents just that parameter.  Then, we have the `Plasma3D` class handle/mangage the relationships between the fields.  I would envision `Plasma3D` calling on the fields classes to construct the fields and then store them in a self-managed `xarray.Dataset`.

pheuer:
@StanczakDominik @rocco8773 

So, I've been playing a bit with xarray and it pretty much requires that the grids be uniform (each coordinate axis represented by a 1D array). Without this, you can't create `coords`  for the `xarray.DataArray`, which are essential to most of what the package does for data manipulation. This is probably ok for most applications, but at least in principle it would be nice to support grids with non-uniform spacings (say, the results of a Lagrangian fluid simulation). 

Is there really an advantage to xarray that justifies that sacrifice? It seems to me we can replicate most of the nice features of xarray by just including a dictionary of arrays in the grid object that represent quantities on the grid (like E and B)?

pheuer:
Actually, I think I've found a way to use xarray: although many of its nicer features won't be usable for non-uniform grids it should still be possible to use those features for the uniform ones. I'll let you know once I have a review-able version based on xarray.

pheuer:
@rocco8773 I've mostly re-implemented grids.py using xarray. Note that I trimmed a lot of unnecessary functionality (eg. different grid types) to speed up the review process: I'll re-implement them in another PR once we get the basics figured out. 

@StanczakDominik What do you think about the possibility of speeding up the interpolators here with numba? I'm interpolating indices so that the actual interpolation is only done once across multiple quantities (better for nearest-neighbor schemes) but maybe there's a faster, simpler way to do it? I think having the interpolators reside in this object makes sense (as opposed to in ParticleTracker) since there may be other applications for interpolating data on a grid.

One observation based on your meeting minutes: `MultiIndex` only takes 1D arrays, so that would work only if we treated as non-uniform grids as entirely unstructured (just 3x 1D lists of points). Maybe that is the best way to go?

StanczakDominik:
> @rocco8773 I've mostly re-implemented grids.py using xarray. Note that I trimmed a lot of unnecessary functionality (eg. different grid types) to speed up the review process: I'll re-implement them in another PR once we get the basics figured out.

Taking a look now!

> @StanczakDominik What do you think about the possibility of speeding up the interpolators here with numba? I'm interpolating indices so that the actual interpolation is only done once across multiple quantities (better for nearest-neighbor schemes) but maybe there's a faster, simpler way to do it? I think having the interpolators reside in this object makes sense (as opposed to in ParticleTracker) since there may be other applications for interpolating data on a grid.

Initial reaction: numba sounds like a great fit but I'll have to refresh my memory of how the interpolators are implemented here. It does make sense to keep them in Grids.
 
> One observation based on your meeting minutes: `MultiIndex` only takes 1D arrays, so that would work only if we treated as non-uniform grids as entirely unstructured (just 3x 1D lists of points). Maybe that is the best way to go?

That sounds like a good idea! I wish I had more experience with unstructured grids :( I was able to find https://github.com/NOAA-ORR-ERD/gridded, though, that seems to have some xarray-like structured/unstructured grid functionality. Could be good to take a look...

pheuer:
@StanczakDominik Great! Grab the most recent commit: I just implemented the nonuniform grids as unstructured arrays idea (except for in the interpolators, where it gets a bit tricky).

StanczakDominik:
By the way, could you grab the lastest `master`? I just fixed github actions tests being cancelled if one of them fails.

StanczakDominik:
Tests seem to fail because xarray needs to be added to setup.cfg as a dependency. Let's just run with whatever version is latest for now.

pheuer:
> Tests seem to fail because xarray needs to be added to setup.cfg as a dependency. Let's just run with whatever version is latest for now.

Ok I just tried adding it: let me know if I put it in the wrong place

StanczakDominik:
Nah, it's [the good place](https://www.youtube.com/watch?v=wDnGoucD9KY) :)

One worry with the interpolators-to-numba idea - I see they're currently implemented as methods, which numba will have some trouble handling in the optimal nopython mode of compilation (because you can't turn `self` into a low-level object without also turning it into a `numba.@jitclass`. We can either profile this, then find the slowest parts and numbify those, **or** we can turn the entire methods into functions and numbify those.

pheuer:
> Nah, it's [the good place](https://www.youtube.com/watch?v=wDnGoucD9KY) :)

Love that show ;)

> One worry with the interpolators-to-numba idea - I see they're currently implemented as methods, which numba will have some trouble handling in the optimal nopython mode of compilation (because you can't turn `self` into a low-level object without also turning it into a `numba.@jitclass`. We can either profile this, then find the slowest parts and numbify those, **or** we can turn the entire methods into functions and numbify those.

I think turning them into functions should be no problem: I can work on that. Basically it's just a wrapper on Scipy.Interpolate functions: all the slow stuff is coming from within scipy so can numba still help with that?

StanczakDominik:
Crud, if it's scipy, numba might fail there - numba compiles pure Python code, and if it's scipy functionality it's likely cythonized/straight up Fortran/C++. It won't compile that.

A few relevant links:

* https://github.com/dbstein/fast_interp
* https://stackoverflow.com/questions/62579492/interpolation-numba
* https://gist.github.com/shoyer/c8bb9e4d689b297a170296426a08aef7

pheuer:
> Crud, if it's scipy, numba might fail there - numba compiles pure Python code, and if it's scipy functionality it's likely cythonized/straight up Fortran/C++. It won't compile that.

Darn - ok. We definitely need full 3D interpolation so the Numba interpolator is probably out, but that fast_interp package sounds promising.

StanczakDominik:
Might honestly be best to run our own interpolator... the code would be pretty clean if we were to use numba.

StanczakDominik:
Or we could accept the runtime penalty for Scipy usage? How are the timings on that?

plasmapy/plasma/grids.py
```diff
@@ -289,81 +250,96 @@ def dax2(self):
...
+                [pts0.flatten(), pts1.flatten(), pts2.flatten()],
+                names=["ax0", "ax1", "ax2"],
+            )
+            self.ds.coords["ax"] = mdx
+
+        # Add quantities
+        for qk in kwargs.keys():
+            q = kwargs[qk]
+
+            self.add_quantity(qk, q)
```
StanczakDominik: I'm having trouble concentrating on this right now, but I'm guessing this is the meat and potatoes of the new idea? I don't see anything immediately wrong with it, but I'll have to take a closer look. I take it the `is_uniform_grid` expects the `numpy.meshgrid` kind of arrays?

plasmapy/plasma/tests/test_grids.py
```diff
@@ -11,190 +11,117 @@
...
-    # Test a u-quantity input
-    i = grid.interpolate_indices(pos.value)
-    pout = grid.nearest_neighbor_interpolator(pos.value, grid.grid)
     # Test multiple grid points
     pos = np.array([[0, 0, 0], [0.5, 0.2, 0]]) * u.cm
     i = grid.interpolate_indices(pos)
-    pout = grid.nearest_neighbor_interpolator(pos.value, grid.grid)
-    pout = grid.volume_averaged_interpolator(pos, grid.grid)
+    pout = grid.nearest_neighbor_interpolator(pos.value, "pts0")
+    pout = grid.volume_averaged_interpolator(pos, "pts0")
```
StanczakDominik, line 112: I'm starting to think it might be best to implement those ourselves using jitclass. That way, the API could be kept similar to scipy's.

StanczakDominik (commented):
Looks like we're hitting the exact same problem I saw in #675 - all the nice interpolators are in scipy and numba can't use them, and you need to interpolate the fields at each step of the simulation, so you're doing this weird dance between calling numba jit code and compiled code from scipy. Groan.

plasmapy/plasma/grids.py
pheuer: Yes the idea is that the data can be structured in two ways: 

Uniform grids: Quantities on the grid have shapes [nx,ny,nz], normal 1D coordinate axes are stored. The relationship between these axes and the grid points is as in np.meshgrid, as you said.

Non-uniform grids: Treated as completely unstructed data: grids are flattened to 1D and a MultiIndex coordinate is provided whose three elements give the x,y,z positions of each data point.

plasmapy/plasma/tests/test_grids.py
pheuer: Sure - note that both are ultimately relying on interpolate_indices right now, so the root of the problem still lies there.

pheuer:
> Might honestly be best to run our own interpolator... the code would be pretty clean if we were to use numba, and efficiency would not be much of an issue.

I absolutely agree, but writing one sounds like a big project! 

> Or we could accept the runtime penalty for Scipy usage? How are the timings on that?

Here's a quick summary: 

```
Grid shape: (250, 250, 250)
Number of positions: 1000000
Interpolate Indices (+Creating interpolator)
--- 0.5229988098144531 seconds ---
Interpolate Indices (Interpolator already exists)
--- 0.38499951362609863 seconds ---
Nearest-Neighbor Interpolation (Interpolator already exists)
--- 0.4419996738433838 seconds ---
Volume-Averaged Interpolation (Interpolator already exists)
--- 1.2230005264282227 seconds ---
```

So, not bad not great? This is just for the nearest-neighbor scheme (with a first-order correction in the volume averaged scheme case): the performance on any linear or higher order interpolation is awful. 

Note that for regular grids the interpolator is being initialized the first time it's called, so subsequent calls are faster (hence the extra 120 ms on the first `interpolate_indices` call).

plasmapy/plasma/grids.py
StanczakDominik: I'll have to run this through a debugger and get a feel for how the MultiIndex variant works. The idea sounds reasonable though!

pheuer:
@StanczakDominik I looked further into the fast_interp repository to linked to earlier: it looks like it will only work well for smooth data, which might not be good for this application. 

Maybe the best thing to do is to just move forward using the scipy interpolators, focus on getting the right API etc. then come back in the future and write our own numba nearest-neighbor interpolator to speed things up under the hood?

StanczakDominik:
Sounds good to me!

StanczakDominik:
In that case, I'll try to give this a fuller review by tomorrow, if that's okay with you? :)

pheuer:
> In that case, I'll try to give this a fuller review by tomorrow, if that's okay with you? :)

Sounds good!

StanczakDominik:
Hey, I started looking at the failing tests, found out the output after the recent transition to github actions was absolutely unreadable, and birthed #956. So I haven't been able to carry out the review yet.

Either way, I have some errors in `fields`:
```
FAILED plasmapy/plasma/tests/test_fields.py::test_example_fields[no fields] - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
FAILED plasmapy/plasma/tests/test_fields.py::test_example_fields[electrostatic gaussian sphere] - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
FAILED plasmapy/plasma/tests/test_fields.py::test_example_fields[electrostatic planar shock] - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
FAILED plasmapy/plasma/tests/test_fields.py::test_example_fields[axial magnetic field] - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
FAILED plasmapy/plasma/tests/test_fields.py::test_setting_emax_bmax - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
FAILED plasmapy/plasma/tests/test_fields.py::test_planar_shock - AttributeError: 'CartesianGrid' object has no attribute 'x_arr'
```

pheuer:
@StanczakDominik  Oops, sorry! I keep forgetting that fields.py is part of this PR since we've been just focusing on grids. I've removed `fields.py` for now: once we've finalized the grid class I'll open a new PR for fields. 

I was able to resolve all the failed tests except for test-documentation: that seems like something external.

StanczakDominik:
There goes the doc test :)

StanczakDominik (approved):
LGTM, and I've put off dealing with this for way, way too long now. I apologize, and let's get this in, then adjust as needed later on! :)

Issue #807, opened 2020-05-10
Make @particle_input compatible with CustomParticle and ParticleLike


namurphy:
The `@particle_input` decorator currently transforms something particle-like (e.g., a string representation of a particle or an atomic number) into a `Particle` object.  However, `@particle_input` is not yet compatible with `CustomParticle` objects.  Here are my suggestions for how to make these @particle_input` and `CustomParticle` compatible.

1. Allow `CustomParticle` objects to get passed through the decorator just like a `Particle` object would.
2. If a mass or a charge is provided for the parameter annotated with `Particle`, then pass through a `CustomParticle` using the provided mass or charge.
3. If a mass *and* a charge is provided (e.g., as in a `tuple` or `list`, or possibly even as two separate parameters passed in sequentually) in the place of the parameter annotated with `Particle`, create a `CustomParticle` object based on the provided mass and charge.  

Here are a few tests that could be used for this.  For a few parts of this, I assume that #806 is implemented.  The tests could also be simplified using the test helpers from #728.

```Python
from astropy import units as u
from plasmapy.particles import particle_input, Particle, CustomParticle

mass = 1e-26 * u.kg
charge = 16e-19 * u.C

@particle_input
def get_particle_input_result(particle: Particle):
    """
    Return the object decorated with Particle that has been acted upon by 
    the `particle_input` decorator.
    """
    return particle

def test_that_particle_input_allows_custom_particles():
    custom_particle = CustomParticle(mass=mass, charge=charge)
    new_custom_particle = get_particle_input_result(custom_particle)
    assert isinstance(new_custom_particle, CustomParticle)
    assert custom_particle == new_custom_particle  # assumes #806 is implemented

def test_that_particle_input_creates_custom_particle_from_mass():
    custom_particle = CustomParticle(mass=mass)
    new_custom_particle = get_particle_input_result(mass)
    assert isinstance(new_custom_particle, CustomParticle)
    assert custom_particle == new_custom_particle  # assumes #806 is implemented
    
def test_that_particle_input_creates_custom_particle_from_charge():
    custom_particle = CustomParticle(charge=charge)
    new_custom_particle = get_particle_input_result(charge)
    assert isinstance(new_custom_particle, CustomParticle)
    assert custom_particle == new_custom_particle  # assumes #806 is implemented

def test_that_particle_input_creates_custom_particle_from_mass_and_charge():
    custom_particle = CustomParticle(mass=mass, charge=charge)

[...]

PR #911, opened 2020-10-05
Starting an attempt to remove default particle assumptions

arichar6:
This is my attempt to remove default values from functions that take a particle as input in `formulary/parameters.py`.

This is described in issue #453.

- [x] If adding new functionality, I have added tests and docstrings. (Well, mostly I changed the docstrings to reflect the fact that some arguments that used to be optional now are required.


rocco8773:
@arichar6 Thanks for taking this on!  Issue #453 was created before I started working on `plasmapy`, but I absolutely agree with it.  With that said,  we shouldn't even give the `ion` and `particle` keywords are default value of `None`.  Instead we should make them a required argument like...

```python
def Alfven_speed(B, density, ion, z_mean=None):
    ...
```

or


```python
def Alfven_speed(B, density, *, ion, z_mean=None):
    ...
```

The latter keeps `ion` as a keyword-only argument, whereas, the former makes it a keyword and positional argument.

Additionally, we should add a proper annotation `Union[str, Particle]`, which would look something like...

```python
def Alfven_speed(B: u.T, density: [u.m ** -3, u.kg / u.m ** 3], *, ion: Union[str, Particle], z_mean=None):
    ...
```

arichar6:
Thanks for the feedback!

I'll remove the `None` defaults, and add the type annotations.

arichar6:
I still have to fix the tests that now fail. But the defaults should be fixed now, and type annotations added.

arichar6:
As I was working on the tests, I realized that one change might introduce a backwards-incompatibility. 
I made the `ion` argument in the `ion_sound_speed` function a keyword-only argument. I'm not sure how to remove the default value without making it keyword-only, since arguments that come before it have default values. 

An alternative would be to change the order of the arguments to the function, but that seems worse somehow.

Thoughts?

plasmapy/formulary/parameters.py
```diff
@@ -177,12 +182,11 @@ def Alfven_speed(
         Either the ion number density in units convertible to 1 / m**3,
         or the mass density in units convertible to kg / m**3.
 
-    ion : str, optional
+    ion : str
```
namurphy, line 185: For `Alfven_speed`, `density` can be either a mass density or a number density, so the argument `ion` should be optional since it's not needed when a mass density is provided.

plasmapy/formulary/parameters.py
```diff
@@ -1676,7 +1683,10 @@ def upper_hybrid_frequency(B: u.T, n_e: u.m ** -3) -> u.rad / u.s:
     },
 )
 @angular_freq_to_hz
-def lower_hybrid_frequency(B: u.T, n_i: u.m ** -3, ion="p+") -> u.rad / u.s:
+def lower_hybrid_frequency(
+    B: u.T,
+    n_i: u.m ** -3,
+    ion: Union[str, Particle]) -> u.rad / u.s:
```
namurphy, line 1689:
```suggestion
    ion: Particle -> u.rad / u.s:
```

plasmapy/formulary/parameters.py
```diff
@@ -275,7 +279,8 @@ def ion_sound_speed(
     k: u.m ** -1 = None,
     gamma_e=1,
     gamma_i=3,
-    ion="p+",
+    *,
+    ion: Union[str, Particle],
```
namurphy, line 283:
```suggestion
    ion: Particle,
```

plasmapy/formulary/parameters.py
```diff
@@ -664,7 +668,7 @@ def thermal_pressure(T: u.K, n: u.m ** -3) -> u.Pa:
     T={"can_be_negative": False, "equivalencies": u.temperature_energy()}
 )
 def kappa_thermal_speed(
-    T: u.K, kappa, particle="e-", method="most_probable"
+    T: u.K, kappa, particle: Union[str, Particle], method="most_probable"
```
namurphy, line 671:
```suggestion
    T: u.K, kappa, particle: Particle, method="most_probable"
```

plasmapy/formulary/parameters.py
```diff
@@ -780,8 +784,8 @@ def Hall_parameter(
     n: u.m ** -3,
     T: u.K,
     B: u.T,
-    ion,
-    particle="e-",
+    ion: Union[str, Particle],
+    particle: Union[str, Particle],
```
namurphy, line 788:
```suggestion
    ion: Particle,
    particle: Particle,
```

namurphy (changes requested):
Thank you for working on this!   Most of my suggestions are related to annotations for particle inputs, in particular that the annotation will need to be `Particle` for when we eventually implement the `@particle_input` decorator to this file (#341).  Otherwise, all of the changes look great.  Thank you again!

Many of the suggested changes are related to the `@particle_input` decorator which still needs to implemented in the formulary

Otherwise, all of the changes look great.

plasmapy/formulary/parameters.py
```diff
@@ -161,7 +163,10 @@ def mass_density(
 @check_relativistic
 @validate_quantities(density={"can_be_negative": False})
 def Alfven_speed(
-    B: u.T, density: [u.m ** -3, u.kg / u.m ** 3], ion="p+", z_mean=None
+    B: u.T,
+    density: [u.m ** -3, u.kg / u.m ** 3],
+    ion: Union[str, Particle],
```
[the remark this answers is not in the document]
rocco8773: @arichar6 Sorry for the back-and-forth on this, but, as @namurphy says, we should just annotate with `Particle` since that is needed for `@particle_input` to work correctly.

With that said, `@particle_input` should be updated to work with `Union[str, Particle]` since those are the actual types that are expected.  This drives my OCD nuts since when annotated with just `Particle` my IDE throws warnings all over the place when I try to input with a string.  I do have a scheme that I think will accomplish this, so I'll open a separate issue on it.

plasmapy/formulary/parameters.py
arichar6: Ok, I took a stab at correcting this. Since it's optional, that means I had to put back the default value of `None`, so that the call to `mass_density` would work.

plasmapy/formulary/parameters.py
```diff
@@ -161,7 +163,10 @@ def mass_density(
 @check_relativistic
 @validate_quantities(density={"can_be_negative": False})
 def Alfven_speed(
-    B: u.T, density: [u.m ** -3, u.kg / u.m ** 3], ion="p+", z_mean=None
+    B: u.T,
+    density: [u.m ** -3, u.kg / u.m ** 3],
+    ion: Optional[Particle] = None,
```
arichar6, line 168: I'm not sure if I did this annotation right (I'm pretty new to using type annotations). I followed the example of the `mass_density` function.

Also, question. Should the previous line actually be using `Union`? My IDE is complaining about using a list for the type annotation.
```suggestion
    density: Union[u.m ** -3, u.kg / u.m ** 3],
    ion: Optional[Particle] = None,
```

arichar6:
One more question. Now that we are using `Particle` as the type annotation, the docstrings should probably be updated from saying something like `ion : str` to `ion : Particle`, correct?

plasmapy/formulary/parameters.py
namurphy: This should still be `[u.m ** -3, ...]` since unit annotations aren't *type hint* annotations, strictly speaking, but are rather used by `plasmapy.utils.validate_quantities` and `astropy.units.quantity_input` to check that arguments have the correct units.  Last I heard, it's still an open issue in Astropy to allow unit annotations that also serve as type hint annotations (e.g., https://github.com/astropy/astropy/pull/10662).  For the moment, we should not use `Union` here, but I'm hoping that this inconsistency (that annotations are used for things besides type hints) eventually gets resolved.

namurphy:
> One more question. Now that we are using `Particle` as the type annotation, the docstrings should probably be updated from saying something like `ion : str` to `ion : Particle`, correct?

Hm...for the moment, I'd suggest using `ion:  Particle`, but we're having a bit of discussion in #912 about how to best do this.

plasmapy/formulary/parameters.py
nstarman: Once 10662 is merged then, for python 3.7+,`astropy.units.quantity_input ` can work with
```python
density: Union[Quantity[u.m ** -3], Quantity[u.kg / u.m ** 3]]
```
For python 3.6+ the manual construction would have to be used
```python
density: Union[Annotated[Quantity, u.m ** -3], Annotated[Quantity, u.kg / u.m ** 3]]
```

plasmapy/formulary/parameters.py
```diff
@@ -61,12 +62,12 @@
...
 
 
-def _grab_charge(ion, z_mean=None):
+def _grab_charge(ion: Particle, z_mean=None):
     """Utility function to merge two possible inputs for particle charge.
 
     Parameters
     ----------
-    ion : str or `plasmapy.particles.Particle`
+    ion : `plasmapy.particles.Particle`
```
arichar6, line 70: Ok, I have a couple more questions (sorry... I'm pretty new to using sphinx and docstrings). 
1) Is it ok to use the fully qualified name here (`plasmapy.particles.Particle`) instead of just `Particle`?
2) Should this be `` `plasmapy.particles.Particle` `` (with the backticks) or `~plasmapy.particles.Particle` (without the backticks, but with a preceeding tilde)? 

And what does that tilde even do? Is that part of the sphinx toolchain that makes clickable documentation links?

rocco8773: It's better to use the fully qualified name, since not doing so could create linking issues when building the documentation.  The `~` indicates to `Sphinx` to just display the last element in the path.  So `` `~plasmapy.particles.Particle` `` will still link the fully qualified path but the link in the documentation will look like `Particle` instead of `plasmapy.particles.Particle`.

When to use `~` and not?  Well, it depends on the context of the current spot in the documentation.  If `Particle` makes sense within the current context then use it, otherwise use `plasmapy.particles.Particle`.  It's kind like using UCLA vs. University of California, Los Angeles.  When in context UCLA is fine.

arichar6: Thanks!

rocco8773:
@arichar6 It appears the documentation build errors are due to the jupyter notebooks not being updated, specifically `/docs/notebooks/physics.ipynb`.  Do you want to give fixing this a try?  Just make sure before committing your changes that you clear all cell outputs.

plasmapy/formulary/parameters.py
```diff
@@ -275,7 +281,8 @@ def ion_sound_speed(
     k: u.m ** -1 = None,
     gamma_e=1,
     gamma_i=3,
-    ion="p+",
+    *,
+    ion: Particle,
```
rocco8773, line 285: This should look like....

```python
def ion_sound_speed(
    T_e: u.K,
    T_i: u.K,
    ion: Particle,
    n_e: u.m ** -3 = None,
    k: u.m ** -1 = None,
    gamma_e=1,
    gamma_i=3,
    z_mean=None,
) -> u.m / u.s:
```

`*`, overly simplified, indicates the end to the positional arguments, so keyword-only arguments should not be defined before it.

plasmapy/formulary/parameters.py
```diff
@@ -468,7 +474,7 @@ def ion_sound_speed(
 @particles.particle_input
 def thermal_speed(
     T: u.K,
-    particle: particles.Particle = "e-",
+    particle: Particle = "e-",
```
rocco8773, line 477:
```suggestion
    particle: Particle,
```

The test associated with this will need to be updated.

changelog/911.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Removed default Particle types for functions in `formulary/parameters.py`. Any
+code which had previously used the default value for these arguments will need
+to be updated by passing the appropriate particle type as a parameter.
+
+Documentation and tests for these functions have been updated as needed.
```
rocco8773, line 5:
```suggestion
Removed default values for the `ion` and `particle` arguments of functions contained in `plasmapy.formulary.parameters`, in accordance with issue [#453](https://github.com/PlasmaPy/PlasmaPy/issues/453).
```

You want to keep this as one paragraph because it gets translated to a single bullet point in a reST unordered list.

rocco8773 (changes requested):
This is looking good!  I only have a few minor changes plus the update of the jupyter notebook.

----

The rest of the `plasmapy.formulary` needs this treatment too, but I'm ok keeping that to a separate/followup PR.

plasmapy/formulary/parameters.py
arichar6: I was going to leave this optional, because if the `mass` argument is specified, then the `particle` argument is unused. But of course having the `particle` argument sometimes ignored is very different than giving it a default value.

I'll change this function, too.

Thinking about this more, does it still make since to keep the `mass` argument for this function? I'm probably missing something, but overriding the mass is not something that seems like a common, or expected thing to do. I guess it doesn't hurt to leave it.

arichar6: See also `formulary.collisions._replaceNanVwithThermalV` for a case where the `mass` argument is specified, but now I have to add something like `particle='e-'` for the code to work. But in the context of that function it seems strange to specify a particle.

Maybe an alternative to having the `mass` parameter in the `thermal_speed` function would be to just use a `CustomParticle` object? Something like
```
particle = CustomParticle(mass=m, charge=0)
v = thermal_speed(T, particle)
```

rocco8773: I get you.  In that case I think we should drop the `particle` argument altogether and make `mass` a required argument, but reinforce in the examples that you can do `thermal_speed(5*u.ev, mass=Particle("e").mass)`.  A function should always require what it needs to run and having `particle` and `mass` has optional keywords does not satisfy that.

@namurphy What do you think is the best way to proceed?

plasmapy/formulary/parameters.py
arichar6: So you are ok with changing the order of the arguments in this function call? I was using the `*` entry so that I could keep the order of arguments the same, but have `ion` (with no default) come after the other arguments which do have defaults.

rocco8773: Keywords have no order associated with them, so doing this does not change the order at all.  It does make `ion` a required argument that is either positional or keyword.  So, the suggested makes the interface behave like...

```python
ion_sound_speed(2*u.eV, 3*u.eV, "He+")

# or

ion_sound_speed(2*u.eV, 3*u.eV, ion="He+")
```

All the other arguments are keyword-only, so they can be specified in any order.

plasmapy/formulary/parameters.py
namurphy: Hm...I prefer the approach of using `CustomParticle`, though this will eventually require finishing #807.  Maybe that's the next issue I should work on...

arichar6: Maybe for #807 you can make `@particle_input` check for `AbstractParticle`? Isn't that the advantage of having an abstract base class... so that any checks for that will still work with concrete subclasses?

namurphy: Yep, that's pretty close to what I was thinking.  We'd also need to make sure that `DimensionlessParticle` does not included, since pretty much everything in `plasmapy.formulary` requires dimensional particles.  We could have a hierarchy like `AbstractParticle` → `DimensionalParticle` → `Particle` & `CustomParticle`.

plasmapy/formulary/parameters.py
arichar6: I'm not sure why you say that all other arguments are keyword-only. I just tried, and you can specify an electron density and k by just giving them as the next arguments. No keywords needed.
```python
>>> ion_sound_speed(2*u.eV, 3*u.eV, "He+", 1 * u.m**-3, 1 * u.m ** -1)
<Quantity 1.54899987 m / s>
```

plasmapy/formulary/parameters.py
rocco8773: Idk, I'm still not convinced about this.  I feel requiring a user to do `thermal_speed(5 * u.eV, particle=CustomeParticle(mass=5. * u.g))` is just too far away from `thermal_speed(5 * u.eV, mass=5. * u.g)` and will cause user frustration.  It's not as intuitive as `thermal_speed(5 * u.eV, mass=Particle("e").mass)`.  I don't really like either scenario though.

I think it might be cooler to have a decorator do something like...

```python
@from_particle(mass={"grab": mass})
themal_speed(T: u.K, mass: u.g):
    ...
```

Then when doing `thermal_speed(5*u.eV, mass="e")` the decorator knows to grab `Particle("e").mass` and pass that in.  Additionally, `thermal_speed(5 * u.eV, mass=5. * u.g)` would still work as expected.

arichar6:
Any insights into why the codecov test is failing now? I'm don't really quite know what it is check for, exactly.

rocco8773:
> Any insights into why the codecov test is failing now? I'm don't really quite know what it is check for, exactly.

Looks like line 339 in `collisions.py` is not getting hit by tests.  Without digging more, I'm not sure why this is the case.

plasmapy/formulary/parameters.py
arichar6: In my mind, there will be more times that I want the thermal speed of a particle than I will want the thermal speed for a given mass. I think that the point of having a package like this is to make it easy to do the common things.
```python
thermal_speed(T, "e-")
```
is easier than
```python
thermal_speed(T, Particle("e-").mass)
```
The second form requires that either:
1. you remember the electron's mass of the top of your head so you can directly enter it 
```python
thermal_speed(T, 9.1094e-28 * u.g)
```
or

2. remember that the thermal speed depends only on temperature and mass, and write
```python
thermal_speed(T, Particle("e-").mass)
```

Not that option (2) is hard to remember, but the point of the package is so that you don't even *need* to remember.

And for those few times where you do want to use a specific mass, it's pretty easy to write 
```python
thermal_speed(T, CustomParticle(mass=5. * u.g))
```

namurphy: With respect to the decorator...one possibility would be to have `@particle_input` pass through a `CustomParticle` when supplied with a mass or charge.

rocco8773: Ideally, I'd like us to be able to do both

```python
thermal_speed(5*u.eV, "e")
thermal_speed(5 * u.eV, 5. * u.g)
```

With that said, I'll back that

```python
thermal_speed(5*u.eV, "e")
thermal_speed(5 * u.eV, CustomParticle(mass=5. * u.g))
```

is the best option for now, where we drop the `mass` kwarg altogether and just use `particle`.  I still think we should try to achieve the ideal case in the future with some kind of decorator.

rocco8773: > With respect to the decorator...one possibility would be to have `@particle_input` pass through a `CustomParticle` when supplied with a mass or charge.

That could be a possibility, but the decorator might be doing too many things at that point.  Let's see when/if we get there.

arichar6:
> 
> Looks like line 339 in `collisions.py` is not getting hit by tests. Without digging more, I'm not sure why this is the case.

I might be wrong, but it looks like line 339 in `collisions.py` is not accessible. Here are the precedding lines:
```python
elif np.any(np.isnan(V)):
    if np.isscalar(V.value) and np.isscalar(T.value):
        V = parameters.thermal_speed(T, "e-", mass=m)
    elif np.isscalar(V.value):
        V = parameters.thermal_speed(T, "e-", mass=m)
```
I just tried this, and if `np.any(np.isnan(V))` returns `True` for a numpy array (with astropy units), then `np.isscalar(V.value)` returns `False`. Was there some other case that this was trying to test for?

namurphy:
> I might be wrong, but it looks like line 339 in `collisions.py` is not accessible. Here are the preceding lines:

> I just tried this, and if `np.any(np.isnan(V))` returns `True` for a numpy array (with astropy units), then `np.isscalar(V.value)` returns `False`. Was there some other case that this was trying to test for?

Hm...that's a good question.  That's a bit strange so it would be worth looking into it further.  It does seem outside of the intended scope of this pull request, so my suggestion would be to create a brief issue on it so that we can address it in a different pull request.  I'm not a fan of long extensive `if`/`elif` statements so I'm wondering if there's a way to [replace these `if`/`elif` blocks with polymorphism](https://www.codementor.io/@uditrastogi/replace-conditional-statements-if-else-or-switch-with-polymorphism-ryl8mx4ns)...but that is definitely for another day.

namurphy:
To follow up: there are a few other things in `collisions.py` that should probably be cleaned up, so I can create the other issue.

rocco8773 (changes requested):

plasmapy/formulary/collisions.py
```diff
@@ -331,18 +331,18 @@ def _replaceNanVwithThermalV(V, T, m):
...
         if np.isscalar(V.value) and np.isscalar(T.value):
-            V = parameters.thermal_speed(T, mass=m)
+            V = parameters.thermal_speed(T, "e-", mass=m)
         elif np.isscalar(V.value):
-            V = parameters.thermal_speed(T, mass=m)
+            V = parameters.thermal_speed(T, "e-", mass=m)
         elif np.isscalar(T.value):
             V = V.copy()
-            V[np.isnan(V)] = parameters.thermal_speed(T, mass=m)
+            V[np.isnan(V)] = parameters.thermal_speed(T, "e-", mass=m)
```
rocco8773, line 342: This should fix the coverage issues...

```suggestion
        if np.isscalar(V.value) or np.isscalar(T.value):
            if np.isscalar(V.value):
                V = parameters.thermal_speed(T, "e-", mass=m)

            if np.isscalar(T.value):
                V[np.isnan(V)] = parameters.thermal_speed(T, "e-", mass=m)
```

changelog/911.breaking.rst
```diff
@@ -0,0 +1 @@
+Removed default values for the `ion` and `particle` arguments of functions contained in `plasmapy.formulary.parameters`, in accordance with issue [#453](https://github.com/PlasmaPy/PlasmaPy/issues/453).
```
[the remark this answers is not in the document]
rocco8773: Include something like... *"..., and updated all relevant calls to modified functionality."*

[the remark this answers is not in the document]
arichar6: Will do.

plasmapy/formulary/collisions.py
arichar6: Looks good. I'll make this change.

arichar6:
Ok. I think this PR is finally done. While this does not complete the changes required for #453, it does complete the changes needed in `parameters.py`.

namurphy (approved):
Thank you again for doing this!  It looks good to me so I think it's ready to merge.

PR #919, opened 2020-10-11
Test helper refactoring 1: Move test helper exceptions from utils to tests.helpers

### LINKED ISSUE #746: Create plasmapy.tests [PLEP7]
As part of the changes associated with the forthcoming PLEP 7 (PlasmaPy/PlasmaPy-PLEPs#26), we decided create a ``plasmapy.tests`` sub-package.

Tasks:

- [ ] Create ``plasmapy.tests``
- [ ] Move ``plasmapy.utils.pytests_helpers`` to ``plasmapy.tests.pytest_helpers``

namurphy:
This is the first in a series of multiple pull requests in an effort to refactor PlasmaPy's test helper functionality.  These changes were originally part of #728, which got much too big and became stuck in code review.  This PR will need to be merged before most subsequent PRs.

This pull request:
 1. Creates `plasmapy.tests.helpers`.
 2. Moves exceptions currently used in `plasmapy.utils.pytest_helpers` to `plasmapy.tests.helpers`.
 3. Puts commented out import statements in `plasmapy.tests.helpers.__init__.py`, that will be uncommented in subsequent PRs.
 4. Updates files in `docs/api_static` that correspond to these files.

Closes #746


namurphy:
This pull request is a prime number!  :smile_cat:  :calendar: :smile_cat:  We seriously need to add a bot that labels prime number pull requests so I don't waste time doing web searches for "is 919 prime"...

rocco8773:
I will note that by moving `pytest_helpters` -> `tests.helpers` the functionality is not automatically imported when `plasmapy` is imported.  That is...

```python
>>> import plasmapy as pp
>>> hasattr(pp.utils, "tests")
False
```

However, they are still importable when explicitly imported.  That is, `from plasmapy.utils import tests` works.

This is probably a good thing as long as this functionality is only used within the `tests` directories.

rocco8773:
On another note, I'm not sold on having test helpers within `plasmapy.tests` because a `tests` folder is for testing, not its utilities.  Plus, it goes against what we agreed on with [PLEP 7,](https://github.com/PlasmaPy/PlasmaPy-PLEPs/blob/master/PLEP-0007.rst#tests) so we'd have to amend PLEP7 is we go against it.

docs/api_static/plasmapy.tests.helpers.exceptions.rst
```diff
@@ -0,0 +1,10 @@
+:orphan:
+
+`plasmapy.tests.helpers.exceptions`
+==============================================
```
rocco8773, line 4:
```suggestion
`plasmapy.tests.helpers.exceptions`
===================================
```

docs/api_static/plasmapy.tests.helpers.rst
```diff
@@ -0,0 +1,10 @@
+:orphan:
+
+`plasmapy.tests.helpers`
+==============================================
```
rocco8773, line 4:
```suggestion
`plasmapy.tests.helpers`
========================
```

docs/api_static/plasmapy.tests.helpers.rst
```diff
@@ -0,0 +1,10 @@
...
+
+.. currentmodule:: plasmapy.tests.helpers
+
+.. automodapi::  plasmapy.tests.helpers
+   :include-all-objects:
+   :no-heading:
```
rocco8773, line 10:
```suggestion
.. automodule:: plasmapy.tests.helpers

Sub-Packages & Modules
----------------------

.. autosummary::

   exceptions

.. automodapi::  plasmapy.tests.helpers
   :include-all-objects:
   :no-main-docstr:
   :no-heading:
   :headings: "-^
```

plasmapy/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""PlasmaPy tests and test helpers."""
```
rocco8773, line 1: I'll note this will never make it into the documentation since we do not have a `plasmapy.tests` stub file, and I think it should stay that way.

plasmapy/tests/helpers/__init__.py
```diff
@@ -0,0 +1,32 @@
+"""Test helper functionality for PlasmaPy and affiliated packages."""
+
+# This file contains several commented out import statements.  These
+# statements will be uncommented out over the course of several pull
+# requests that were each originally part of #728.  The blank lines
+# between the import statements will hopefully simplify automatic merging.
+
+# from plasmapy.tests.helpers.actual import ActualTestOutcome
+
+# from plasmapy.tests.helpers.cases import AttrTestCase, FunctionTestCase, MethodTestCase
```
rocco8773, line 10: I think this is fine for your working branch, but should be removed for the PR.  It refers to code that does not exist yet and that shouldn't happen in the `master` branch.

plasmapy/tests/helpers/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+#  from plasmapy.tests.helpers.inputs import (
+#      AbstractTestInputs,
+#      ClassAttributeTestInputs,
+#     ClassMethodTestInputs,
+#     FunctionTestInputs,
+#     GenericClassTestInputs,
+# )
+
+# from plasmapy.tests.helpers.runner import test_runner
```
rocco8773, line 32: Same as above comment.

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+    "UnexpectedResultError",
+    "InconsistentTypeError",
+    "MissingExceptionError",
+    "UnexpectedExceptionError",
+    "MissingExceptionError",
+    "UnexpectedWarningError",
+    "MissingWarningError",
+    "InvalidTestError",
+    "ExceptionMismatchError",
+    "WarningMismatchError",
```
rocco8773, line 14: Alphabetize.  Makes the documentation easier to read.  ... There also seems to be duplicates in here. eg `MissingExceptionError`.

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+    Notes
+    -----
+    This exception was derived from `~pytest.fail.Exception`, which in
+    turn was derived from `BaseException` (not `Exception`).
+    """
+
+    pass
+
+
+class InconsistentTypeError(Failed):
```
rocco8773, line 36: Should this also inherit from `TypeError`, to be consistent?

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class ExceptionMismatchError(UnexpectedExceptionError, MissingExceptionError):
+    """
+    Exception for when an exception is expected, but a different
+    exception is raised.
+    """
+
+    pass
+
```
rocco8773, line 68: This feels a little to "fine" detail to me.  Why not have a `RaisesError` for any wrong `Exception` type failures?  The description string will do all the necessary identification of mismatches, unexpected, etc.

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+
+class WarningMismatchError(UnexpectedWarningError, MissingWarningError):
+    """
+    Exception for when a warning is expected, but one or more
+    different warnings were issued instead.
+    """
+
+    pass
```
rocco8773, line 92: Same comment as to the exception errors above.  Why not have a `WarnsError` for everything?

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+class UnexpectedResultError(Failed):
+    """
+    Exception for when the actual result differs from the expected
+    result by more than the allowed tolerance.
+    """
+
+    pass
+
+
+class InvalidTestError(Exception):
```
rocco8773, line 104: I'm assuming you're inheriting form `Exception`, and not `Failed`, on purpose?  This is a weird one since the test fails, but not due to the tested code failing.  Not sure which inheritance is more appropriate.

plasmapy/utils/pytest_helpers/pytest_helpers.py
```diff
@@ -18,16 +18,16 @@
 
 from typing import Any, Callable, Dict
 
-from plasmapy.utils.error_messages import _exc_str, _represent_result, call_string
-from plasmapy.utils.exceptions import PlasmaPyWarning
-from plasmapy.utils.pytest_helpers.exceptions import (
+from plasmapy.tests.helpers.exceptions import (
```
rocco8773, line 21: Note: This import will cause `plasmapy.tests` to be automatically import when `plasmapy` is imported.  This will stop happening then your full architecture is implemented, assuming `plasmapy.tests.helpers` are only used within test directories.  (**Probably good to make a note about this on the "Testing Guidelines" page.**)

rocco8773 (changes requested):
Do you want to make a reference to `plasmapy.tests.helpers` within the "Writing Tests" section of the "Testing Guidelines" page?

Do you know if a `tests` directory within a `tests` directory will still run?  I'm assuming so, but I've never tested it.

There are docstring references to `` `~plasmapy.utils.pytest_helpers.<Exception>` `` in `pytest_helpers.run_test()` that needs updating.

plasmapy/tests/helpers/exceptions.py
namurphy: There are a couple of reasons why I chose to have a different exception for each of the different types of test failures.
 - I want the error messages to be as informative as possible, as quickly as possible.  I chose the names of these exceptions to be highly descriptive so that developers can get a good idea of what the problem is just by reading the name of the exception.  As someone who has a bad habit of skimming through error messages without actually reading them, a descriptive name really does help.
 - Having more specific exceptions allows more specific exception handling.  There are some possibilities for extending additional test runners (e.g., to replace `run_test_equivalent_calls`) that may need to handle different exceptions differently.

plasmapy/tests/helpers/exceptions.py
namurphy: Yeah...this is different than a test failure, because the test was not actually able to be performed.  It's kind of like how pytest treats errors and failures differently.  I've been using `Failed` to represent failures where the problem is with the production code, and `InvalidTestError` to represent errors where the problem is the test code.

plasmapy/utils/pytest_helpers/pytest_helpers.py
namurphy: Good point!  I hadn't realized that, though it's good that this problem will go away once the whole refactoring is done.

plasmapy/tests/helpers/exceptions.py
rocco8773: I'm still not sold on this, but I won't argue it since I see the reasoning.  With that said, they should probably all inherit from a `RaisesError` exception, then you can catch all variations in the try-statement.

plasmapy/tests/helpers/exceptions.py
rocco8773: Is there currently a use case for this?  Does `pytest` already have something like this in it?  Just curious if we're duplicating functionality that already exists.

plasmapy/tests/helpers/__init__.py
namurphy: The reason why I have this in here is because I have a whole bunch of simultaneous closely related pull requests that are each going to need to make changes to this file, and I'm hoping that this will prevent the need to resolve merge conflicts across ~7 PRs and allow the PRs to be cleanly separated from each other.  Pretty much...if you let me bend the rules a little bit just this once...it'll make my life a little easier!

namurphy:
> Do you want to make a reference to `plasmapy.tests.helpers` within the "Writing Tests" section of the "Testing Guidelines" page?

That's the plan!  Later pull requests extracted from #728 will deal with the narrative documentation.

> Do you know if a `tests` directory within a `tests` directory will still run? I'm assuming so, but I've never tested it.

Yep.  I've tried this and it works.

plasmapy/tests/helpers/__init__.py
rocco8773: Fine, but it will cost you a beer at the next in-person APS. 😁

namurphy: > Fine, but it will cost you a beer at the next in-person APS. grin

Thanks!  I'll try to remember that for 2026.

rocco8773:
> > Do you want to make a reference to `plasmapy.tests.helpers` within the "Writing Tests" section of the "Testing Guidelines" page?
> 
> That's the plan! Later pull requests extracted from #728 will deal with the narrative documentation.

It might be a good idea to create a mini-project to gather all these PRs, issues, and little tidbits. I like this smaller PR approach, but a project could be helpful with the overall concept and workflow.

namurphy:
> It might be a good idea to create a mini-project to gather all these PRs, issues, and little tidbits. I like this smaller PR approach, but a project could be helpful with the overall concept and workflow.

Yes!  Great idea!

plasmapy/tests/helpers/exceptions.py
namurphy: Despite the name, this is different from a `TypeError`.  A `TypeError` is to be (from Python 3 docs): "raised when an operation or function is applied to an object of inappropriate type"...which is not happening here.  In this case, the types of two objects are being checked against each other, and are found to not be the same.  I wonder if we should rename this to something like `InconsistentTypes`...

plasmapy/tests/helpers/exceptions.py
namurphy: The closest thing is a `UsageError` in pytest...though the docstring specifically refers to pytest so I don't think it's quite apropos to inherit from it:  
```Python
@final
class UsageError(Exception):
    """Error in pytest usage or invocation"""
```

namurphy: The use case is for a developer who is writing tests, but makes an error in setting the test up.  That has happened to me...dare I say...more than once.

plasmapy/tests/helpers/exceptions.py
rocco8773: Maybe instead of calling these `<something>Error` we should call them `<something>Failure`.  Since this is really a failure opposed to an error.  For reference, `pytest.fail` does...

```python
>>> pytest.fail()
Failed: <Failed instance>
```

plasmapy/tests/helpers/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    InvalidTest,
+    MissingException,
+    MissingWarning,
+    TestFailure,
+    TypeMismatch,
+    UnexpectedException,
+    UnexpectedResult,
+    UnexpectedWarning,
+    WarningMismatch,
+)
```
rocco8773, line 23: I feel this naming convention is a bit vague.  Why not do something like I suggested in https://github.com/PlasmaPy/PlasmaPy/pull/919#discussion_r504986757.  I would even consider renaming the module to `plasmapy.test.helpers.failures`.

> Maybe instead of calling these `<something>Error` we should call them `<something>Failure`. Since this is really a failure opposed to an error. For reference, `pytest.fail` does...
> 
> ```python
> >>> pytest.fail()
> Failed: <Failed instance>
> ```

----

Additionally, if we're continuing with this explicit failure route, then we might want to create a `plasmapy.fail()` version of [`pytest.fail()`](https://docs.pytest.org/en/stable/_modules/_pytest/outcomes.html#fail), which would allow us fail with our custom failures.

namurphy: I'm hesitant to add `Failure` to the end of the exception names because it should be clear from context that it's a test failure.  Hence, I don't think `Failure` would add any additional information to distinguish between the different modes of test failures.  I plan on getting a lot of test failure messages in the future :joy_cat: and I think the shorter names will help us distinguish between the causes of test failures a little bit more quickly, especially once a developer gets used to them.  Ideally, the test failure name should get to the root of the problem as quickly as possible without unnecessary information.

Another reason is that the word `Failure` can be just a little bit discouraging...at least that's how it affected me when I was first starting out.  

> I would even consider renaming the module to `plasmapy.test.helpers.failures`.

I like this idea!  That is a really good fit except for `InvalidTest` which is more of an error than a test in pytestspeak, but I don't think it matters too much.  The path also shouldn't show up to users during actual test failures, so I don't think it'd have the slight psychological discouragement I mentioned above.

> Additionally, if we're continuing with this explicit failure route, then we might want to create a `plasmapy.fail()` version of [`pytest.fail()`](https://docs.pytest.org/en/stable/_modules/_pytest/outcomes.html#fail), which would allow us fail with our custom failures. ~I think this would be more appropriate than raising custom `Exceptions`.~ `pytest.fail()` is just functionality wrapped around a `raises Failed`.

I'm trying to remember if that's what I did way back in #728!  I don't think I ended up doing this, though.  This is outside the scope of this PR, but I'll make a comment in one of the subsequent pull requests so we don't forget.

rocco8773: > I'm hesitant to add `Failure` to the end of the exception names because it should be clear from context that it's a test failure. Hence, I don't think `Failure` would add any additional information to distinguish between the different modes of test failures. I plan on getting a lot of test failure messages in the future 😹 and I think the shorter names will help us distinguish between the causes of test failures a little bit more quickly, especially once a developer gets used to them. Ideally, the test failure name should get to the root of the problem as quickly as possible without unnecessary information.

I disagree with this.  **This goes against our whole concept of explicit is better than implicit.**  All exceptions, except `Exception`, end with `Error` for this reason, likewise for `Warnings`.  Now, `Failure` might not be the best extension, maybe `Failed` since that matches better with `pytest.Failed`...or `Fail`.  When outside the context of `plasmapy.test.helpers.exceptions`, the current variable names would have no context.

Having `Error` at the end of an exception does not hinder its readability, so I doubt having something like `Fail` would hinder its readability.

> Another reason is that the word Failure can be just a little bit discouraging...at least that's how it affected me when I was first starting out.

This is something we all have to get use to when we start testing.  However, it should not supersede the need for explicitly over implicity.  It's on us to encourage new contributors when they're getting test failures, not the code.

namurphy: I renamed the test failure exceptions to have `Fail` at the end, except for `InvalidTestError` which is different than a normal test failure.  I decided against renaming `exceptions.py` to `failures.py` since in pretty much every other package we use `exceptions.py`, and it's less needed since we have `Fail` at the end of the exception names now.

namurphy:
@rocco8773 - I think we've now resolved everything we talked about so far.  Let me know if you have any other suggestions, and otherwise feel free to merge if it looks good to you.  Thanks for the review!

plasmapy/tests/helpers/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    InvalidTestError,
+    MissingExceptionFail,
+    MissingWarningFail,
+    TestFailure,
+    TypeMismatchFail,
+    UnexpectedExceptionFail,
+    UnexpectedResultFail,
+    UnexpectedWarningFail,
+    WarningMismatchFail,
+)
```
rocco8773, line 23: I'm thinking we should add an `__all__` with these exceptions in it.

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,105 @@
...
+    "UnexpectedExceptionFail",
+    "UnexpectedResultFail",
+    "UnexpectedWarningFail",
+    "WarningMismatchFail",
+]
+
+import pytest
+
+
+class TestFailure(pytest.fail.Exception):
```
rocco8773, line 19: This is the first time I've looked into `pytest`'s file structure and man is it weird!

Using `pytest.fail.Exception` is very round about way of getting the `Exception`.  It's not a literal exception definition, but a class attribute of `_WithException` which is used in decorating `pytest.fail()`.  Instead, you should use `_pytest.outcomes.Failed`, which equates to the parent class of `pytest.fail.Excpetion`.

rocco8773: You'll enjoy this one...How about `TestFailed` instead of `TestFailure` to better mirror `_pytest.outcome.Failed`?  That is all I will say and I'll let you choose whichever you prefer. 😆

namurphy: I had thought about that, but was hesitant to because it didn't seem like part of the public API.  I'm also not sure if it adds much to inherit from the pytest exception.  In any case, I'm okay with any of these possibilities.

And yeah...the inside of pytest is pretty different from what I'm used to.

namurphy: > You'll enjoy this one...How about `TestFailed` instead of `TestFailure` to better mirror `_pytest.outcome.Failed`? That is all I will say and I'll let you choose whichever you prefer. laughing

This I'll agree with you on!

namurphy: > Instead, you should use `_pytest.outcomes.Failed`, which equates to the parent class of `pytest.fail.Excpetion`.

Hm...it appears that `pytest._pytest` cannot be imported.  I think that means we'll need to go with `pytest.fail.Exception` unless I'm missing something.

```Python
In [8]: import pytest._pytest
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-8-148f78f1fe80> in <module>
----> 1 import pytest._pytest

ModuleNotFoundError: No module named 'pytest._pytest'
```

rocco8773: > Shall I go ahead with inheriting from _pytest.outcomes.Failed, given the uncertainty about whether or not it's part of the public API?

I'd say yes.  I think that approach would be more stable than the other route.  My only concern is that by sub-classing `_pytest.outcomes.Failed` our custom failures might start interacting with the internals of `pytest`.  If we find that this happens in an undesirable way, then we make our own `Failed` class that inherits from `BaseException` or `Exception`.

plasmapy/tests/helpers/exceptions.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    pass
+
+
+class UnexpectedResultFail(TestFailed):
+    """
+    Exception for when the actual result differs from the expected
+    result by more than the allowed tolerance.
+    """
```
rocco8773, line 95: Might want to generalize the wording a little more, since I can imagine an `UnexpectedResultFail` that is not involve `isclose` or `allclose`.  For example, arrays of differing sizes.

rocco8773:
It looks like the failed coverage check is related to lines the weren't hit before this PR.  I'm assuming these will be address in the following bazillion PRs associated with the test helper overhaul.

plasmapy/tests/helpers/exceptions.py
rocco8773: You should do...

```python
from _pytest.outcomes import Failed

class TestFailed(Failed):
   ...
```

`_pytest` is installed with `pytest` because `pytest` basically calls on `_pytest`.

namurphy:
> It looks like the failed coverage check is related to lines the weren't hit before this PR. I'm assuming these will be address in the following bazillion PRs associated with the test helper overhaul.

The lines that aren't covered are coming from the old test runner, which I'm planning to delete soon after the new test runner is implemented.

rocco8773 (approved):
Looks good to me!  That was fun! :)

namurphy:
> Looks good to me! That was fun! :)

That is good because we have nine more of these to go. :eyes: :joy_cat: :seedling: :zzz: :dragon: :floppy_disk: :space_invader: :tophat: :interrobang:

Issue #726, opened 2019-12-03
`plasmapy.formulary.quantum.chemical_potential` bug with powers


StanczakDominik:
The current doctest error for `plasmapy.formulary.quantum.chemical_formula()` is a result of a floating point arithmetic error...

<img width="805" alt="image" src="https://user-images.githubusercontent.com/29869348/69690641-5d474980-1081-11ea-95fc-86d0d29e34fa.png">

There is a large difference in `n_e` when we convert from `cm ** -3` to `m ** -3`...

```python
>>> (1.e27*u.m**-3) - (1.e21*u.cm**-3).to(u.m**-3)
<Quantity 1.37438953e+11 1 / m3>
```

Thus, even though the inputs are essentially identical, the following yields...

```python
>>> chemical_potential(1.e27 ** u.m ** -3, 11000 * u.K)
<Quantity 2.00039985e-12>

>>> chemical_potential(1.e21 ** u.cm ** -3, 11000 * u.K)
<Quantity 5.1487703e-12>
```

@namurphy @StanczakDominik Thoughts???  If we can't figure out a quick fix, then I feel like we should exclude this code from the source until we determine a solution.

_Originally posted by @rocco8773 in https://github.com/PlasmaPy/PlasmaPy/pull/722#issuecomment-558917491_

rocco8773:
This issue has been raised on AstroPy ([#9721](https://github.com/astropy/astropy/issues/9721)).

JaydenR2305:
@StanczakDominik I think this is working now, I get identical outputs for your given example case

StanczakDominik:
Huh! Can you try removing `@pytest.mark.xfail` from the related tests in `quantum` to trigger CI runs for this?

JaydenR2305:
I'm getting an assertion error for the known test case... Was the test case itself calculated incorrectly?

StanczakDominik:
Could be! Can you post that as a PR?

PR #920, opened 2020-10-12
Test helper refactoring 2: Refactor and move functionality for generating call strings used in error messages

namurphy:
The second spin-off PR from #726.  This should be reviewed and merged after #919.

We currently have a function `call_string` which generates a string that can reproduce a particular call.  This function is currently used in test helper functionality when generating error messages.  Its potential uses go beyond creating error messages, though, such as in `__repr__` methods.  This PR moves that functionality to `plasmapy.utils.formatting`, and creates functions for reproducing the call strings of functions and attributes/methods of class instances.


namurphy:
- I removed colorama as a dependency, in part because it sometimes led to weird formatting in terminals.  
 - I sorted imports with `isort` and reformatted with `black` for a small number of files elsewhere in the package.

plasmapy/utils/exceptions.py
```diff
@@ -93,6 +94,14 @@ class PlasmaPyWarning(Warning):
...
 
 
+class PlasmaPyDeprecationWarning(DeprecationWarning, PlasmaPyWarning):
+    """
+    A warning for features in PlasmaPy that have been deprecated.
+    """
+
+    pass
+
+
```
rocco8773, line 104: What's the motivation for this, versus just using `DeprecationWarning`?  If we really want our own `DeprecationWarning`, then I would encourage considering an approach similar to how `sphinx` does it.   They do this [this](https://github.com/sphinx-doc/sphinx/blob/64498ad0dbab86fa42f5556d7bc84baf719cefda/sphinx/deprecation.py#L21-L33)...

```python
class RemovedInSphinx40Warning(DeprecationWarning):
    pass


class RemovedInSphinx50Warning(PendingDeprecationWarning):
    pass


class RemovedInSphinx60Warning(PendingDeprecationWarning):
    pass
```

So, our version would look like...

```python
class RemovedInPlasmaPy05Warning(DeprecationWarning, PlasmaPyWarning):
    pass


class RemovedInPlasmaPy06Warning(PendingDeprecationWarning, PlasmaPyWarning):
    pass
```

Now, I also think we should consider `FutureWarning` versus `DeprecationWarning`.  According to [Python's documentation](https://docs.python.org/3/library/warnings.html?highlight=warning#warning-categories), `DeprecationWarning` and `FutureWarnings` both announce deprecated features, but `DeprecationWarning` is intended for developers and `FutureWarning` is intended for end users.  Since we are constantly changing and deprecating end user features, we might want to consider using `FutureWarning` instead.

docs/utils/index.rst
```diff
@@ -17,16 +17,17 @@ means:
 
  * the many kinds of warnings and exceptions you may (hopefully not!)
    encounter while working with PlasmaPy, such as
-   `plasmapy.utils.RelativityWarning` or `plasmapy.utils.PhysicsError`.
+   `~plasmapy.utils.RelativityWarning` or `~plasmapy.utils.PhysicsError`.
```
rocco8773, line 20:
```suggestion
   `~plasmapy.utils.exceptions.RelativityWarning` or `~plasmapy.utils.exceptions.PhysicsError`.
```

Since we do not `automodapi` or `automodule` `plasmapy.utils`,  nothing defined in the `__all__` of `plasmapy/utils/__init__.py` is documented; thus, there is no stub file for things like `plasmapy.utils.RelativityWarning`.

docs/utils/index.rst
```diff
@@ -17,16 +17,17 @@ means:
...
  * decorators we use for reusable physical quantity computation and
-   checking, such as `plasmapy.utils.decorators.validate_quantities` and
-   `plasmapy.utils.check_relativistic`.
+   checking, such as `~plasmapy.utils.decorators.validate_quantities` and
+   `~plasmapy.utils.check_relativistic`.
```
rocco8773, line 23:
```suggestion
   `~plasmapy.utils.decorators.check_relativistic`.
```

rocco8773 (changes requested):
This is just a partial review.  I'm only focusing on the documentation aspect of the functionality.

plasmapy/utils/exceptions.py
namurphy: My original motivation was to follow Astropy's strategy for handling deprecations.  The benefit of having `PlasmaPyDeprecationWarning` is to have a bit more specificity than `DeprecationWarning`.  I'm more inclined to specify the version in which something is deprecated in a `@deprecated` decorator (#945) rather than having a series of different classes.  I don't have a strong opinion though.

I think I'll back off on the deprecation stuff in this PR since it's off-topic, but I'll link to this discussion in #945 since it's really relevant there.

plasmapy/utils/formatting/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Tools for formatting strings, including for error messages."""
```
rocco8773, line 1: I would argue that this sub-directory could/should be decomposed into a single `formatting.py` file, much like we did with `roman.py`.  I'd back off this if you see a significant expansion of the functionality contained within this sub-directory, but I do not see that at the moment.

I also don't think the name `formatting` quite represents what this functionality is doing.  Something like `repr` would better reflect what the functionality is doing.  Obviously, `repr` would be a poor name due to naming conflicts. ...`code_repr`, `obj2str`, `obj2repr`, `obj2code`, `obj2codestr` ...idk, none of these seem to be elegant or representative.

rocco8773 (changes requested):
Another partial review since GitHub doesn't let me respond to previous comments while I have a pending review.  Come on GitHub, it's almost 2021!!

plasmapy/utils/formatting/formatting.py
```diff
@@ -0,0 +1,255 @@
+import inspect
+import numpy as np
+
+from astropy import units as u
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+__all__ = ["call_string", "attribute_call_string", "method_call_string"]
```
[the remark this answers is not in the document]
namurphy: If there ever was a code review comment that captures the essence of @rocco8773 reviewing @namurphy's pull request, this is it. :1st_place_medal: :100: :+1:  :joy_cat: 

I better go check the other PRs in this series!

plasmapy/utils/formatting/__init__.py
namurphy: Good idea...I'll move this into a single file.  I can't think of anything else what would go into `formatting` at the moment, though I expect some things here and there will come up.  

I'll think about the name...thus far some variation of `code_repr` seems best.

namurphy:
My to do list...
 - [ ] Move `formatting.py` to its own file in utils, rather than in a directory in utils
 - [ ] Decide on a name for `formatting.py`
 - [ ] Remove deprecation stuff from this PR as its out-of-scope
 - [ ] Put `__all__` above imports everywhere

namurphy:
@rocco8773 — thanks again for the review!  I applied most of the suggestions so it's ready for another round when you have a chance.  It'd be particularly helpful if you could double check that the `docs/api_static` stuff is being handled correctly.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+    followed by the corresponding message, separated by a full line.
+    """
+    warnings_with_messages = [
+        _object_name(warning, showmodule=False) + ": " + message
+        for warning, message in zip(warning_types, warning_messages)
+    ]
+    return "\n\n".join(warnings_with_messages)
+
+
+def call_string(f: Callable, args: Any = tuple(), kwargs: Dict = {}) -> str:
```
rocco8773, line 121: I'm trying to understand the purpose of `call_string`.  Is it used for actual `assert` tests or is used only for reporting info associated with test failures?  For example,

```python
pytest.fail(f"Call {cal_string(...)} was suppose to do blah, but did doogah.")
```

I don't know, this seems like a lot of complexity and we're traveling rather far from [KISS](https://en.wikipedia.org/wiki/KISS_principle).  This functionality works beautifully for simple function calls, but would get out of hand if you were trying to use it on a function with a signature like [`scipy.signal.find_peaks`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html) (random example that would plausible in a `plasmapy.analysis` function).

rocco8773 (commented):
Only a review on `plamsapy.utils.code_repr`.


This is a general statement to consider for all functionality in `code_repr.py`.  Since the `formatting` lingo is being replaced with the `code_repr` lingo, any naming or reference to `format` should probably also be replaced with `code_repr`.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+    Transform a `~astropy.units.Quantity` into a format as would appear
+    in a function call.
+    """
+    if not isinstance(arg, u.Quantity):
+        raise TypeError("Expecting a Quantity.")
+
+    formatted = f"{repr(arg.value)}"
+
+    if isinstance(arg.value, np.ndarray):
+        formatted = "np." + formatted
```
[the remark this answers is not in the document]
namurphy: Yes, agreed.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+            else:
+                formatted += f"*u.{base}**{power}"
+
+    return formatted
+
+
+def _format_arg(arg) -> str:
+    """Transform an argument into a format as would appear in a function call."""
+    if hasattr(arg, "__name__"):
+        return arg.__name__
```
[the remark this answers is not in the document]
namurphy: From the [built-in types page](https://docs.python.org/3/library/stdtypes.html#definition.__name__):

> `definition.__name__`
> 
>     The name of the class, function, method, descriptor, or generator instance.

I wasn't sure about this, so it was helpful to look it up.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+    if arg.unit == u.dimensionless_unscaled:
+        formatted += "*u.dimensionless_unscaled"
+    else:
+        for base, power in zip(arg.unit.bases, arg.unit.powers):
+            if power == -1:
+                formatted += f"/u.{base}"
+            elif power == 1:
+                formatted += f"*u.{base}"
+            else:
+                formatted += f"*u.{base}**{power}"
```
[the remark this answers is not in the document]
namurphy: This seems like a good case to create a test for, too.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+    Return the name of an `object`.  If the `object` has a `__name__`
+    attribute and ``showmodule`` is `True`, then prepend the module
+    name if not in `builtins`.
+    """
+    obj_name = obj.__name__ if hasattr(obj, "__name__") else repr(obj)
+
+    if hasattr(obj, "__name__") and showmodule is True:
+        module_name = inspect.getmodule(obj).__name__  # ADD TEST
+        if module_name != "builtins":
+            obj_name = f"{module_name}.{obj_name}"  # ADD TEST
```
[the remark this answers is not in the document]
namurphy: Looking back at this, maybe I could also simplify this using `__qualname__`.  Will refactor...

plasmapy/utils/code_repr.py
namurphy: The main purpose of `call_string`, etc. is for reporting info associated with test failures.  Reading the actual function call that led to a test failure can be extremely helpful for figuring out what the problem is, and it can help speed up debugging too, especially for newcomers to the project.  These functions could also be used in `__repr__`, and I think there would likely be times when these could be used directly in tests too.  

The point you made about managing complexity is important.  I'll need to think about that more...

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+
+from astropy import units as u
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+
+def _format_quantity(arg: u.Quantity) -> str:
+    """
+    Transform a `~astropy.units.Quantity` into a format as would appear
+    in a function call.
+    """
```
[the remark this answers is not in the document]
namurphy: I decided to tentatively go with names like `_code_repr_of_lihjdfkh`.  Thanks!

plasmapy/utils/code_repr.py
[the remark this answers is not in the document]
namurphy: Thinking more, this shouldn't come up since `u.some_unit ** 0` turns into `u.dimensionless_unscaled`.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,257 @@
...
+    string as they would appear in a function call.
+    """
+    args = args if isinstance(args, tuple) else (args,)
+    args_and_kwargs = ""
+
+    for arg in args:
+        args_and_kwargs += f"{_format_arg(arg)}, "
+
+    for kwarg in kwargs:
+        args_and_kwargs += f"{_format_kw(kwarg)}={_format_arg(kwargs[kwarg])}, "
```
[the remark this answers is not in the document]
namurphy: Tentatively renaming it to `_format_kw_name`.

plasmapy/utils/code_repr.py
rocco8773: I'm curious if you could sidestep the complexity if you incorporate some functionality form `inspect` to do the heavy lifting, like [`getcallargs`](https://docs.python.org/3/library/inspect.html#inspect.getcallargs) in combination with [`formatargspec`](https://docs.python.org/3/library/inspect.html#inspect.formatargspec).

namurphy: That looks promising!  I'll look into it!

namurphy: I'm not sure if what's in `inspect` will be able to take care of all the use cases that are needed.  Since `call_string` is being used to report testing information, it needs to be able to handle situations where the call is invalid, since invalid calls will sometimes need to be reconstructed and reported back to the user.  That is apparently out of scope for `inspect`.  Hm...

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,323 @@
...
+
+    method: `str`
+        The name of the method in class ``cls``
+
+    cls_args : `tuple`, `list`, or any; optional
+        A `tuple` or `list` containing positional arguments, or any
+        other type of `object` if there is only one positional argument,
+        to be used during instantiation of ``cls``
+
+
```
namurphy, line 288:
```suggestion

```

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+    def replace_excess_items_with_ellipsis(s, max_items):
+        substrings_between_commas = s.split(",")
+        to_comma_before_ellipsis = ",".join(substrings_between_commas[0:max_items])
+        closing_brackets = "]" * substrings_between_commas[-1].count("]")
+        closing = f", ...{closing_brackets})"
+        return to_comma_before_ellipsis + closing
+
+    if not isinstance(array, np.ndarray):
+        raise TypeError("Expecting an ndarray.")
```
rocco8773, line 41: This scenario is not covered in the tests.

plasmapy/utils/tests/test_code_repr.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+@pytest.mark.parametrize(
+    "quantity, expected",
+    [
+        (5.3 * u.m, "5.3*u.m"),
+        (5.4 / u.m, "5.4/u.m"),
+        (5.5 * u.m ** -2, "5.5*u.m**-2"),
+        (5.0 * u.dimensionless_unscaled, "5.0*u.dimensionless_unscaled"),
+        (np.array([3.5, 4.2]) * u.m ** -2.5, "np.array([3.5, 4.2])*u.m**-2.5"),
```
rocco8773, line 223: For completeness, maybe throw in a case that looks like `5.0 / u.m ** 3`.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+
+def _code_repr_of_arg(arg) -> str:
+    """Transform an argument into a format as would appear in a function call."""
+    if hasattr(arg, "__name__"):
+        return arg.__name__
+
+    function_for_type = {
+        u.Quantity: _code_repr_of_quantity,
+        np.ndarray: _code_repr_of_ndarray,
+    }
```
rocco8773, line 86: Should probably add `_code_repr_...`'s for long lists, tuples, and dicts.  I'm guessing these cases will turn up quickly.

plasmapy/utils/tests/test_code_repr.py
```diff
@@ -0,0 +1,370 @@
...
+    "arg, expected",
+    [
+        (1, "1"),
+        ("asdf", "'asdf'"),
+        (3.42, "3.42"),
+        ([3.42, 3.84], "[3.42, 3.84]"),
+        (4.2 * u.m, "4.2*u.m"),
+        (np.array([1, 2, 3]), "np.array([1, 2, 3])"),
+        (np.array([[1, 2], [3, 4]]), "np.array([[1, 2], [3, 4]])"),
+        (np.array([1.0, 2.0, 3.0]) * u.m / u.s, "np.array([1., 2., 3.])*u.m/u.s"),
```
rocco8773, line 276: How about test cases for lists, tuples, and dicts?

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+    args_collection = args if isinstance(args, tuple) else (args,)
+
+    args_and_kwargs = ""
+
+    for arg in args_collection:
+        args_and_kwargs += f"{_code_repr_of_arg(arg)}, "
+
+    for kwarg in kwargs.keys():
+        args_and_kwargs += f"{kwarg}={_code_repr_of_arg(kwargs[kwarg])}, "
```
rocco8773, line 109:
```suggestion
    for kw_name, kw_val in kwargs.items():
        args_and_kwargs += f"{kw_name}={_code_repr_of_arg(kw_val)}, "
```

Just think this is a little more readable...but your decision to take it or leave it.

plasmapy/utils/tests/test_code_repr.py
```diff
@@ -0,0 +1,370 @@
...
+    args_and_kwargs = _code_repr_of_args_and_kwargs(args, kwargs)
+    if args_and_kwargs != expected:
+        pytest.fail(
+            f"_code_repr_of_args_and_kwargs with the following arguments:\n"
+            f"  args:   {args}\n"
+            f"  kwargs: {kwargs}\n"
+            f"is not returning the expected string:\n"
+            f"  expected: {repr(expected)}\n"
+            f"  actual:   {repr(args_and_kwargs)}"
+        )
```
rocco8773, line 319: Testing and reporting failures this way seems like it's sidestepping the [`assert` mechanics](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).  Why not do something more like...

```suggestion
    args_and_kwargs = _code_repr_of_args_and_kwargs(args, kwargs)
    assert args_and_kwargs == expected, (
        f"_code_repr_of_args_and_kwargs with the following arguments:\n"
        f"  args:   {args}\n"
        f"  kwargs: {kwargs}\n"
        f"is not returning the expected string:\n"
        f"  expected: {repr(expected)}\n"
    )
```

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+    if args_and_kwargs[-2:] == ", ":
+        args_and_kwargs = args_and_kwargs[:-2]
+
+    return args_and_kwargs
+
+
+def _name_with_article(ex: Exception) -> str:
+    """
+    Return a string with an indefinite article and name of
+    exception ``ex``.
```
rocco8773, line 120: Very few people here are going to be this familiar with grammar syntax, so this probably won't be clear for many.

```suggestion
    Return a string with an indefinite article and the name of
    the exception (e.g. `Exception` -> ``'an Exception'``.
```

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+        "builtins": "",
+        "astropy.units.core": "u",
+        "astropy.units.quantity": "u",
+        "astropy.units.decorators": "u",
+        "astropy.units": "u",
+    }
+    for old, new in replacements.items():
+        if module_name.startswith(old):
+            module_name = module_name.replace(old, new, 1)
+    return module_name
```
rocco8773, line 153: If `_substitute_module_shortcuts()` is only going to be used by `_object_name()` then it's code should be contained with it `_object_name()`; otherwise, `_substitute_module_shortcuts()` should have its own dedicated tests.  If neither of those approaches are taken then these two functions are not 100% independent of each other can lead to problems down the line.

plasmapy/utils/tests/test_code_repr.py
```diff
@@ -0,0 +1,370 @@
...
+        generic_function,
+        (1, "b"),
+        {"b": 42, "R2": "D2"},
+        "generic_function(1, 'b', b=42, R2='D2')",
+    ),
+]
+
+
+@pytest.mark.parametrize("function,args,kwargs,expected", call_string_table)
+def test_call_string(function, args, kwargs, expected):
```
rocco8773, line 73: Unless `call_string_table` (and the like below) is (are) going to be called in several locations, I find that this disassociation of the parametrized list and the test function to significantly hinder readability of tests.  I mean, I start reading the test and then I have to wonder away from the test to get the cases that are being tested.

plasmapy/utils/code_repr.py
```diff
@@ -0,0 +1,321 @@
...
+        The class to be used in the string representation
+
+    method: `str`
+        The name of the method in class ``cls``
+
+    cls_args : `tuple`, `list`, or any; optional
+        A `tuple` or `list` containing positional arguments, or any
+        other type of `object` if there is only one positional argument,
+        to be used during instantiation of ``cls``
+    cls_kwargs: `dict`, optional
```
rocco8773, line 287:
```suggestion
        to be used during instantiation of ``cls``

    cls_kwargs: `dict`, optional
```

plasmapy/utils/tests/test_code_repr.py
namurphy: I think the main problem with readability here is that the when we parametrize tests using a list containing a series of tuples which might contain inner tuples, the parameters get detached from the corresponding semantic information/metadata.  It'd be so much more readable if the arguments were defined like keywords.  In #925 I define classes that are intended to address this problem...without me having to rewrite essentially the same code over and over again.  I might try doing something like that here, though independent of #925, since I'm trying to avoid using the test helpers to test the test helpers.

plasmapy/utils/code_repr.py
namurphy: Since these are private functions, it might be better to do EAFP rather than LBYL (following a code review comment I received on Astropy).  I think I'll delete these lines.

plasmapy/utils/tests/test_code_repr.py
rocco8773: I agree with that sentiment.  `pytest.mark.parametrize`'s separation of names and values does make it a bit difficult to follow.

My point here is that when someone reads `@pytest.mark.parametrize("function,args,kwargs,expected", call_string_table)` they have to immediately search the code to find `call_string_table`.  The fact that I have to leave the test to find out what is being tested is what reduces readability for me.

plasmapy/utils/tests/test_code_repr.py
rocco8773: I do think this approach should be applied to all similar cases.

namurphy:
Changes:
 - I renamed `cls_args` to `args_to_cls` (and same to similar arguments)
 - I created namedtuples to make it easier to understand the things in parametrize
   - I thought...hey!  This is a job for dataclasses!  But it turns out, it wasn't.  Keep it simple splendorously.
 - I added some tests with a list or tuple as the argument.  I found a bug.  I fixed it.
 - I made some optional arguments keyword-only in order to enforce
 - I added a parameter `max_items` that can limit the number of items that will get printed out in an `ndarray` or a `Quantity`
 - Otherwise, I adopted most @rocco8773's suggestions.

rocco8773 (approved):
I'm satisfied. 😁 ... Looks good.  The `namedtuple` is an interesting approach to the problem, would not have thought about it myself.  I'm guessing you decided to leave the `code_repr` functions for lists, tuples, and dicts for a separate PR, correct?

PR #923, opened 2020-10-12
Test helper refactoring 4: Add class to represent expected outcomes of tests

namurphy:
This PR adds a class that represents the expected outcomes of tests.  This class, `ExpectedTestOutcome`, can account for test outcomes such as returned values and/or a warning being issued, or an exception being raised.  This accounts for `Quantity` objects too.  This class originally came from #728, which is being split into multiple PRs.  The motivation for the test helper refactoring, as well as the development history of this class, is discussed thoroughly in #728. 

Remaining tasks

 - [ ] Import this class in `plasmapy/tests/helpers/__init__.py` (after #919 is merged)
 - [ ] Decide if this should be a private class or not


plasmapy/tests/__init__.py
```diff
@@ -1,3 +1,5 @@
 """PlasmaPy tests and test helpers."""
 
+__all__ = ["helpers"]  # is this needed?
```
StanczakDominik, line 3: Probably not really. https://plasmapy--923.org.readthedocs.build/en/923/api_static/plasmapy.tests.html isn't a thing, and https://plasmapy--923.org.readthedocs.build/en/923/api_static/plasmapy.tests.helpers.html#module-plasmapy.tests.helpers probably doesn't care about this `__all__` statement.
```suggestion
```

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    Case(expected_warning_and_value, "expecting_a_value", True),
+    Case(expected_warning_and_value, "expected_outcome", expected_warning_and_value),
+]
+
+
+@pytest.mark.parametrize("case", expected_outcome_test_cases)
+def test_expected_test_outcome_attributes(case: Case):
+    """Test that `ExpectedTestOutcome` attributes each return the expected values."""
+    expected_outcome = ExpectedTestOutcome(case.argument)
+    result = expected_outcome.__getattribute__(case.attribute)
```
StanczakDominik, line 58: Could simply be a
```suggestion
    result = getattr(expected_outcome, case.attribute)
```

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    if result is not case.correct_outcome and result != case.correct_outcome:
+        pytest.fail(
+            f"ExpectedTestOutcome({case.argument}) results in {repr(result)} "
+            f"but should result in {case.correct_outcome}."
+        )
```
StanczakDominik, line 63: I keep running into these blocks. I honestly don't think they're helpful.

Consider this basic test:


```python
import pytest

@pytest.mark.parametrize("answer", [42, 44])
def test_answer_to_everything(answer):
    assert 41 < answer < 43, "Answer to everything out of bounds (cf. Douglas, 1979)"
```

When I run this, I get 

```
🕙[ 11:09:22 ] ❯ pytest test_pytest.py                                                                                                                                  [1]
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.9.2, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
Matplotlib: 3.4.0
Freetype: 2.10.4
rootdir: /home/dominik
plugins: datadir-1.3.1, profiling-1.7.0, mpl-0.12, ordering-0.6, regressions-2.2.0, pudb-0.7.0, testmon-1.0.3, hypothesis-6.3.4, cov-2.11.1, anyio-2.2.0, Faker-6.6.3, mock-3.1.1, xdist-2.2.1, forked-1.3.0, black-0.3.12, xonsh-0.9.24
collected 2 items

test_pytest.py .F                                                                                                                                                    [100%]

================================================================================= FAILURES =================================================================================
______________________________________________________________________ test_answer_to_everything[44] _______________________________________________________________________

answer = 44

    @pytest.mark.parametrize("answer", [42, 44])
    def test_answer_to_everything(answer):
>       assert 41 < answer < 43, "Answer to everything out of bounds (cf. Douglas, 1979)"
E       AssertionError: Answer to everything out of bounds (cf. Douglas, 1979)
E       assert 44 < 43

test_pytest.py:5: AssertionError
========================================================================= short test summary info ==========================================================================
FAILED test_pytest.py::test_answer_to_everything[44] - AssertionError: Answer to everything out of bounds (cf. Douglas, 1979)
======================================================================= 1 failed, 1 passed in 0.61s ========================================================================
```

The values are shown and I didn't need to hand write any f-strings.

So I would prefer to change those (and stick to this style in our tests, and ensure the same is stated in the documentation guide) to *(trying to invert your condition...)*... and the reasoning is obvious enough from the simple test, so the message is unnecessary...

```suggestion
    assert result is case.correct_outcome
    assert result == case.correct_outcome
```

And now I honestly wonder what the point is of having both `is` and `==` checks. That sounds off. I haven't looked into the class itself yet, just trying to familiarize myself with how it works on the surface level; but if the correct outcome of a test were to be a string...

```python
[ins] In [3]: A = "Engage."

[ins] In [4]: B = "Engage!"

[ins] In [5]: A[:-1] == B[:-1]
Out[5]: True

[ins] In [7]: A[:-1] is B[:-1]
Out[7]: False
```

That latter condition does not sound all that helpful. Are you 100% sure we need it?

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+]
+
+
+@pytest.mark.parametrize("case", exception_raising_cases)
+def test_expected_test_outcome_exceptions(case: Case):
+    """Test that attributes of `ExpectedTestOutcome` raise exceptions as needed."""
+    if not issubclass(case.correct_outcome, Exception):
+        raise TypeError(
+            "Incorrect test setup: the expected outcome must be an exception."
+        )
```
StanczakDominik, line 86:
```suggestion
	assert issubclass(case.correct_outcome, Exception)
```

Same logic as above. When you simplify the checks to their simplest form, they usually become self-documenting.

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    with pytest.raises(case.correct_outcome):
+        expected_outcome = ExpectedTestOutcome(case.argument)
+        result = expected_outcome.__getattribute__(case.attribute)
+        pytest.fail(
+            f"The ExpectedTestOutcome instance for {case.argument} did not "
+            f"raise the expected exception but instead returned {result}."
+        )
```
StanczakDominik, line 93: I'm trying and failing to wrap my head around this.

So this is in a `raises` block, but your error message says it **did not** raise the expected exception.

... wha?

Can we rewrite this to avoid the manual `pytest.fail`?

At the very least,
```suggestion
    with pytest.raises(case.correct_outcome):
        expected_outcome = ExpectedTestOutcome(case.argument)
        result = getattr(expected_outcome, case.attribute)
        pytest.fail(
            f"The ExpectedTestOutcome instance for {case.argument} did not "
            f"raise the expected exception but instead returned {result}."
        )
```

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    Test that `~plasmapy.utils.pytest_helpers.expected._is_warning`
+    returns `True` for warnings and `False` for other objects.
+    """
+    interpreted_as_warning = _is_warning(possible_warning)
+    if interpreted_as_warning != actually_a_warning:
+        pytest.fail(
+            f"_is_warning({repr(possible_warning)}) should return "
+            f"{actually_a_warning}, but is instead returning "
+            f"{interpreted_as_warning}."
+        )
```
StanczakDominik, line 119:
```suggestion
    assert interpreted_as_warning == actually_a_warning
```

It's also usually simpler to think in terms of true statements rather than their negations.

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    Test that `~plasmapy.utils.pytest_helpers.expected._is_exception`
+    returns `True` for exceptions and `False` for other objects.
+    """
+    interpreted_as_exception = _is_exception(possible_exception)
+    if interpreted_as_exception != actually_an_exception:
+        pytest.fail(
+            f"_is_exception({repr(possible_exception)} should return "
+            f"{actually_an_exception}, but is instead returning "
+            f"{interpreted_as_exception}."
+        )
```
StanczakDominik, line 145:
```suggestion
    assert interpreted_as_exception == actually_an_exception
```

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    `list` containing a warning and an `object` that is not a `Warning`,
+    and `False` for anything else.
+    """
+    interpreted_as_warning_and_value = _is_warning_and_value(possible_warning_and_value)
+    if interpreted_as_warning_and_value != actually_warning_and_value:
+        pytest.fail(
+            f"_is_warning_and_value returns {interpreted_as_warning_and_value} "
+            f"with {repr(possible_warning_and_value)} as an argument, but is "
+            f"expected to return {actually_warning_and_value}."
+        )
```
StanczakDominik, line 178:
```suggestion
    assert interpreted_as_warning_and_value == actually_warning_and_value
```
Seriously, make these fail on your local setup and see what gets printed out!

plasmapy/tests/helpers/tests/test_expected.py
```diff
@@ -0,0 +1,178 @@
...
+    is_warning_and_value_test_inputs,
+)
+def test__is_warning_and_value(
+    possible_warning_and_value, actually_warning_and_value: bool
+):
+    """
+    Test that `_is_warning_and_value` returns `True` for a `tuple` or
+    `list` containing a warning and an `object` that is not a `Warning`,
+    and `False` for anything else.
+    """
```
StanczakDominik, line 171: To be honest, I tend to zone out and gloss over docstrings while reading simple tests like these. I would not mind not requiring those unless the test has complex logic (e.g. we need to simulate something to compare the results, or we want to provide a source for an analytical solution).

plasmapy/tests/helpers/expected.py
```diff
@@ -0,0 +1,198 @@
...
+import inspect
+
+from typing import Any
+
+from plasmapy.utils.code_repr import call_string
+
+
+def _is_warning(obj) -> bool:
+    """Return `True` if the argument is a warning, and `False` otherwise."""
+    return inspect.isclass(obj) and issubclass(obj, Warning)
```
StanczakDominik, line 14: Curious - doesn't the latter imply the former? I mean, everything in Python derives from `Object`, no?

plasmapy/tests/helpers/expected.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def _is_exception(obj) -> bool:
+    """Return `True` if the argument is an exception, and `False` otherwise."""
+    if not inspect.isclass(obj):
+        return False
+    else:
+        return issubclass(obj, BaseException) and not issubclass(obj, Warning)
```
StanczakDominik, line 22: You'd have to have some really messed up design to have a warning that derives from BaseException, no? :sweat_smile:

StanczakDominik (changes requested):
Starting to get to this, finally! This is a partial review; I got through everything but the class itself, and I've got a bunch of questions in the comments below.

For now, I think I should point out these two:

* https://docs.pytest.org/en/stable/reference.html#pytest.raises - you know this one, but note the `match` regex functionality and the `exc_info` - I think there might be a little bit of scope overlap between this and that functionality.
* https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising - I ran into this just now and it probably does have a whole lot of scope overlap with your ExpectedTestOutcome class, except for the not-class-oriented design. But in a real world use case I'm still likely to use that one, I think.

My general thinking is that we should try to leverage as much pytest or other existing functionality as possible, rather than hand write our own. Otherwise it'd be a little like, y'know someone doing plasma research in python and writing their own functions for the plasma frequency when [there's now a great package that does that for you](http://docs.plasmapy.org/) :wink:  and can be assumed to have put more thought into it than goes into a hand written implementation.

namurphy:

PR #924, opened 2020-10-12
Test helper refactoring 5: Add classes to validate, store, and call inputs for test cases

namurphy:
This PR (which has been extracted from #728) creates three classes to be used to store the callable and arguments, and then call it.  These classes include all of the processing to handle and validate positional and keyword arguments, and include additional checks like making sure that functions are indeed functions.  The goal is to hide all of this low-level stuff from the higher-level test runner functionality during the refactoring of `run_test`.
    
The three variations of the result test pattern are: (1) calling a function with `args` & `kwargs`, (2) calling a class instance method, and (3) accessing a class instance attribute to make sure that the result matches the correct value.  The classes corresponding to this include `FunctionTestInputs`, `ClassMethodTestInputs`, and `ClassAttributeTestInputs`.
    
The lines that say `__tracebackhide__ = True` are instructions for pytest to consider this a test runner function and not include traceback information when doing the error report.  (I still need to make sure that these are effective and work.)
    
An alternative to the `call` method on these classes is to use `functools.partial`, which binds arguments to a callable and creates a new function-like thing.  I think the code is more understandable as the pull request currently stands since it spells out how the function, method, or attribute is being called.


plasmapy/tests/helpers/inputs.py
```diff
@@ -0,0 +1,444 @@
...
+    else:
+        return (args,)
+
+
+def _validate_kwargs(kwargs: Optional[Dict[str, Any]]) -> Dict[str, Any]:
+    """
+    Check that the argument is a valid `dict` to represent keyword
+    arguments to be provided to some function.  If the input is `None`,
+    then this function will return an empty `dict`.  This function does
+    not check that
```
namurphy, line 44: I should probably finish this sentence, shouldn't I?

plasmapy/tests/helpers/inputs.py
```diff
@@ -0,0 +1,444 @@
+"""Classes and functions that validate, store, and call inputs for test cases."""
+
+__all__ = [
+    "AbstractTestInputs",
+    "FunctionTestInputs",
+    "GenericClassTestInputs",
+    "ClassAttributeTestInputs",
+    "ClassMethodTestInputs",
```
namurphy, line 8: I better alphabetize these before @rocco8773 sees this...

namurphy:

PR #925, opened 2020-10-12
Test helper refactoring 6: Add classes to handle test cases

namurphy:
This PR adds classes (split off from #728) that contain all inputs for a test case and the expected outcome.  

The three use cases are:
 1. Providing a function with `args` and `kwargs`, along with the expected outcome
 2. Providing a class (instantiated with `cls_args` and `cls_kwargs`) and the name of an attribute (labelled with `@property`), along with the expected outcome of trying to access that attribute.
 3. Providing a class (instantiated with `cls_args` and `cls_kwargs`) and the name of a method (to be called with `method_args` and `method_kwargs), along with the expected outcome of trying to call that method.

These classes are intended to be what developers use to set up test cases.  Instances of these classes are to be passed to the test runner from #728 that will be added in a subsequent pull request.


namurphy:

PR #926, opened 2020-10-13
Test helper refactoring 7: Add class that represents an actual test outcome

namurphy:
This PR (originally came from #728) adds `ActualTestOutcome` to `plasmapy.tests.helpers`, which calls the function/method or accesses the class attribute and records the outcome of that call (be it a returned value and/or a warning or an exception).  The idea is to consolidate all of the exception handling and context environments into one place so that it can be isolated from the high level calls.


namurphy:

PR #927, opened 2020-10-13
Test helper refactoring 8: Add classes that perform comparisons between actual & expected outcomes

namurphy:
When writing tests for `plasmapy.particles` or `plasmapy.formulary`, I found myself repeatedly rewriting essentially the same error messages over and over for different tests.  The goal of this PR is to isolate all this boilerplate stuff into specific classes so that we can avoid that unnecessary duplication of code.  

This PR, which originally came from #728, adds classes that are responsible for determining whether or not a test passes, selecting appropriate exceptions to raise, and generating appropriate error messages.  For example, the `CompareActualExpected` class makes the comparison between an actual test outcome and the expected test outcome.  This class includes the machinery necessary to make these comparisons for situations where the expected outcome is a returned value (including `Quantity` instances, etc.), warnings being issued (possibly along with a returned value), or exceptions being raised.  

In the current implementation of `run_test`, a whole bunch of this stuff is done directly in that function.  This refactoring of test helper functionality makes the approach much more modular and extensible.


namurphy:

PR #928, opened 2020-10-13
Test helper refactoring 9: Add test runner functionality

namurphy:
This PR, originally from #728, adds the `test_runner` function that replaces `plasmapy.utils.pytest_helpers.run_test`.  The current implementation of `run_test` is run-on spaghetti code that lacks modularity.  The modularity that comes from other pull requests in this series enables `test_runner` to be significantly shorter and cleaner than `run_test`.  

The goal of `test_runner` is to simplify the process of writing tests that fall under the result test pattern: for example, when you call a function with particular arguments, and then you see if the actual outcome matches the expected outcome.  In practice this gets annoyingly complicated due to a whole bunch of different cases...like when you expect a warning to be issued (possibly along with a returned value) or an exception to be raised, or if you're comparing `Quantity` objects, or NumPy arrays, or other things that make my brain hurt.  `test_runner` provides a unified and extensible interface for performing a wide range of these tests.  All of the usages have documentation in either the docstring of `test_runner`, or the narrative documentation to be added to a new pull request shortly.

`test_runner` accepts an instance of `FunctionTestInputs`, `ClassAttributeTestInputs`, or `ClassMethodTestInputs` as its sole argument.  It then runs the test and generates a test failure message when necessary.  These are the only four objects that most users will need to interact with, and possibly only `test_runner` and `FunctionTestInputs`.  For the most part, users will not need to interface with the classes created in #923, #925, #926, and #927. 

My best guess is that ~90% of tests in `plasmapy.particles` and `plasmapy.formulary` fall into the result test pattern (including warnings and exceptions) and could be performed using `test_runner`.


namurphy:
- [ ] As per the [comment](https://github.com/PlasmaPy/PlasmaPy/pull/919#discussion_r505646404) by @rocco8773: in #919, consider adding `plasmapy.tests.helpers.fail` as an analog of `pytest.fail`.

A benefit of this is that we could further customize test failure stuff for use in PlasmaPy, such as adding stuff that could be done during every test failure.

namurphy:

PR #932, opened 2020-10-13
Create plasmapy.dispersion.dispersionsolver & associated functionality

ahmadryan:

      docstrings.


plasmapy/formulary/ionization.py
```diff
@@ -102,13 +102,15 @@ def Saha(g_j, g_k, n_e: u.m ** -3, E_jk: u.J, T_e: u.K) -> u.dimensionless_unsca
     and recombination of ions with electrons are balanced.
 
     .. math::
-        \frac{N_j}{N_k} = \frac{1}{n_e} \frac{g_j}{4 g_k a_0^{3}} \left( \frac{k_B T_e}{\pi E_H} \right)^{\frac{3}{2}}
+        \frac{N_j}{N_k} = \frac{1}{n_e} \frac{g_j}{4 g_k a_0^{3}} \left(
+            \frac{k_B T_e}{\pi E_H} \right)^{\frac{3}{2}}
          \exp\left( \frac{-E_{jk}}{k_B T_e} \right)
```
rocco8773, line 107:
```suggestion
        \frac{N_j}{N_k} = \frac{1}{n_e} \frac{g_j}{4 g_k a_0^{3}} \left(
                              \frac{k_B T_e}{\pi E_H}
                          \right)^{\frac{3}{2}}
                          \exp\left( \frac{-E_{jk}}{k_B T_e} \right)
```

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+    gamma_i=3,
+    ion="p+",
+    m_e: u.kg = m_e,
+    m_i: u.kg = m_p,
+    theta: u.deg = 45 * u.deg,
+    z=1,
+):
+    r"""
+    Return a dictionary of frequencies corresponding to the
+    solutions of dispersion relation in low frequency regime.
```
rocco8773, line 39:
```suggestion
    Return a dictionary of frequencies corresponding to the solutions of
    dispersion relation in low frequency regime.
```

Our current max line length is 88, so don't worry about wrapping text until you're past 80 characters.

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+    theta : ~astropy.units.Quantity
+        Angle of propagation which defaults to 45 degrees.
+    z : float or integer
+        Average ionization number which defaults to 1 for protons.
+
+    Returns
+    -------
+    omega : ~astropy.units.Quantity
+        A dictionary of Wave frequencies corresponding to three modes namely
+        a) Ion-acoustic mode, b) Alfven mode and c) Fast mode, in 1/s units.
```
rocco8773, line 83:
```suggestion
    omega : `~astropy.units.Quantity`
        A dictionary of Wave frequencies corresponding to three modes:
        (1) Ion-acoustic mode, (2) Alfven mode, and (3) Fast mode, in 1/s units.
```

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+
+    .. math::
+        \frac{\omega}{\omega_{ci}} = \sqrt(2\Lambda \sqrt(-\frac{p}{3}) cos
+        \left( \frac{1}{3} cos^{-1}\left( \frac{3q}{2p} \sqrt(-\frac{3}{p}
+        \right) - \frac{2\pi}{3}j \right) + \frac{A\Lambda}{3})
+
+    Where,
+        j = 0 ==> fast mode
+        j = 1 ==> Alfven mode
+        j = 2 ==> Acoustic mode
```
rocco8773, line 111:
```suggestion
    where :math:`j = 0` represents the fast mode, :math:`j = 1` represents the
    Alfven mode, and :math:`j = 2` represents the Acoustic mode.
```

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+
+    The above equation is derived from the general wave equation in the low
+    frequency regime, where both electrons and ions play significant role (in
+    the high frequency regime ions do not play any significant role and the
+    process is dominated by electron dynamics).
+
+    The complete dispersion equation is thus written as (from equation (1) of
+    [1]):
```
rocco8773, line 119:
```suggestion
    [1]_):
```

The trailing underscore tells `Sphinx` this is suppose to be a link.

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+
+    .. math::
+        \left( cos^2\theta - Q\frac{\omega^2}{k^2 {v_A}^2} \right) \left[
+        \left( cos^2\theta - \frac{\omega^2}{k^2 {c_s}^2 \right) -
+        Q\frac{\omega^2}{k^2 {v_A}^2 \left( 1 - \frac{\omega^2}{k^2 {c_s}^2
+        \right) \right] = \left(1 - \frac{\omega^2}{k^2 {c_s}^2 \right)
+        \frac{\omega^2}{{\omega_{ci}}^2}cos^2\theta
```
rocco8773, line 126:
```suggestion
 
        \left( \cos^2 \theta - Q \frac{\omega^2}{k^2 {v_A}^2} \right) &
        \left[
            \left( \cos^2 \theta - \frac{\omega^2}{k^2 {c_s}^2} \right)
            - Q \frac{\omega^2}{k^2 {v_A}^2} \left(
                1 - \frac{\omega^2}{k^2 {c_s}^2}
            \right)
        \right] \\
            &= \left(1 - \frac{\omega^2}{k^2 {c_s}^2} \right)
              \frac{\omega^2}{{\omega_{ci}}^2} \cos^2 \theta
```

I reformatted this for two reasons: (1) there were missing closing curly brackets `}` that were keeping the equation from being rendered correctly and (2) improve readability for the coder.

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Here,
+    .. math::
+        Q = 1 + k^2 c^2/{\omega_{pe}}^2
+    \omega_{ci} is the proton gyrofrequency
```
rocco8773, line 131:
```suggestion
    Here,

    .. math::
        Q = 1 + k^2 c^2/{\omega_{pe}}^2

    :math:`\omega_{ci}` is the proton gyrofrequency.
```

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+
+    References
+    ----------
+    .. [1] PM bellan, Improved basis set for low frequency plasma
+    waves, 2012, JGR, 117, A12219, doi:10.1029/2012JA017856
+    .. [2] TE Stringer, Low-frequency waves in an unbounded
+    plasma, 1963, JNE, Part C, doi:10.1088/0368-3281/5/2/304
+    .. [3] Rogers, B. N.; Denton, R. E.; Drake, J. F. & Shay, M. A.
+    Role of Dispersive Waves in Collisionless Magnetic Reconnection, prl, 2001,
+    87, 195004
```
rocco8773, line 141:
```suggestion
    .. [1] PM bellan, Improved basis set for low frequency plasma waves, 2012,
           JGR, 117, A12219, doi: `10.1029/2012JA017856
           <https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2012JA017856>`_.

    .. [2] TE Stringer, Low-frequency waves in an unbounded plasma, 1963, JNE,
           Part C, doi: `10.1088/0368-3281/5/2/304
           <https://doi.org/10.1088/0368-3281/5/2/304>`_

    .. [3] Rogers, B. N.; Denton, R. E.; Drake, J. F. & Shay, M. A. Role of
           Dispersive Waves in Collisionless Magnetic Reconnection, prl, 2001,
           87, 195004, doi: `10.1103/PhysRevLett.87.195004
           <https://doi.org/10.1103/PhysRevLett.87.195004>`_
```

On line continuation the text needs to be indented.  Added links for the DOI references.

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+    >>> k = 0.01 * u.m ** -1
+    >>> theta = 30 * u.deg
+    >>> B = 8.3E-9 * u.T
+    >>> n = 5.e6 * u.m ** -3
+    >>> T_e = 1.6e6 * u.K
+    >>> T_i = 4.e5 * u.K
+    >>> z = 1
+    >>> omega = tfds(B=B, k=k, n=n, T_e=T_e, T_i=T_i, theta=theta, z=z)
+    >>> omega
+    {'fast_mode': <Quantity [[1520.5794506]] rad / s>, 'alfven_mode': <Quantity [[1261.75471561]] rad / s>, 'acoustic_mode': <Quantity [[0.6881521]] rad / s>}
```
rocco8773, line 156:
```suggestion
    {'fast_mode': <Quantity [[1520.5794506]] rad / s>,
     'alfven_mode': <Quantity [[1261.75471561]] rad / s>,
     'acoustic_mode': <Quantity [[0.6881521]] rad / s>}
```

plasmapy/dispersion/two_fluid_dispersion_solver.py
```diff
@@ -0,0 +1,243 @@
...
+    >>> z = 1
+    >>> c = 3.e8 * u.m/u.s
+    >>> c_s = pfp.ion_sound_speed(T_e=T_e, T_i=T_i, n_e=z * n, ion='p+')
+    >>> v_A = pfp.Alfven_speed( B, n, ion='p+')
+    >>> omega_ci = pfp.gyrofrequency(B=B, particle='p+', signed=False, Z=z)
+    >>> omega = tfds(n=n, B=B, T_e=T_e, T_i=T_i, theta=theta, z=z, k=k_arr)
+    >>> omega['fast_mode'][:,40]
+     <Quantity [1.61176312e-02, 1.77335334e-01, 3.38688590e-01, ...,
+	                1.52030361e+03, 1.52045553e+03, 1.52060745e+03] rad / s>
+
```
rocco8773, line 173:
```suggestion
    <Quantity [1.61176312e-02, 1.77335334e-01, 3.38688590e-01, ...,
               1.52030361e+03, 1.52045553e+03, 1.52060745e+03] rad / s>
```

rocco8773 (changes requested):
This is review is intended to get the documentation building correctly...

1. You made changes to `plasmapy.formulary.ionization` that caused the documentation build to stop working.  This is due to indentation issues.  However, this change is not relevant to the current PR, so it should be broken out into its own PR. ... What you were trying to fix (too long lines) was the right thing to do.
1. Think of the documentation in `/docs/api_static/` as the auto-generated documentation and everything else as the narrative documentation.  So, if you're not narrating something, then utilize the stub files contained in `/docs/api_static/`.
    1. Delete `/docs/dispersion/dispersionfunction.rst` and rely on `/docs/api_static/plasmapy.dispersion.dispersionfunction.rst`.
    1. In `plasmapy.dispersion.dispersionfunction.rst`, update the notebook link to `/notebooks/dispersion/dispersion_function`.
    1. Delete `/docs/dispersion/twofluiddispersionsolver.rst` and replace it with `/docs/api_static/plasmapy.dispersion.two_fluid_dispersion_solver.rst`.  The template for this file can be copied from `/docs/api_static/plasmapy.dispersion.dispersionfunction.rst`.
    1. Delete `/docs/api_static/plasmapy.dispersion.rst` and rely on `/docs/dispersion/index.rst`.  This is opposite from the other cases because this seeds the narrative documentation of the dispersion sub-package.
    1. In `/docs/dispersion/index.rst` add `.. py:currentmodule:: plasmapy.dispersion` between the page title and the `.. automodule::` directive.
1. I also have changes for the notebook, but I'll do this with a followup review.

docs/notebooks/dispersion/two_fluid_dispersion_solver.ipynb
```diff
@@ -0,0 +1,271 @@
+{
```
rocco8773, line 1: <p>Replace the top cell with a markdown cell containing the notebook title.  Something like <code>Dispersion: Two Fluid Analytical Solution</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

rocco8773, line 1: <p>Replace the cell with...</p><p> </p><p><code></code><code></p><p>%matplotlib inline</p><p> </p><p>import astropy.units as u</p><p>import matplotlib.pyplot as plt</p><p>import numpy as np</p><p> </p><p>from matplotlib import colors</p><p>from mpl_toolkits.axes_grid1 import make_axes_locatable</p><p> </p><p>from plasmapy.formulary import parameters as pfp</p><p>from plasmapy.dispersion.two_fluid_dispersion_solver import two_fluid_dispersion_solution</p><p></code><code></code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

rocco8773, line 1: <p><strong>Item 1:</strong></p><p> </p><p><code>k = np.linspace(10<strong>-7, 10</strong>-2, int(1E4)) * u.m ** -1</code></p><p> </p><p><code>num</code> for <code>linespace</code> needs to be an integer.</p><p> </p><p><strong>Item 2:</strong></p><p> </p><p>In your parameter list you should define <code>ion = "p+"</code> and pass this into <code>ion_sound_speed</code>, <code>Alfven_speed</code> , and <code>gyrofrequency</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

rocco8773, line 1: <p>The <code>text</code>  variable should be replaced with <code>text = f"$v_A/c_s = {v_A/c_s:.1f}, c/v_A = 10^{np.log10(c/v_A):.0f}, \\theta = {theta.value:.0f}" + "^{\\circ}$"</code> . </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

rocco8773, line 1: <p><strong>Item 1:</strong></p><p> </p><p><code>k = np.linspace(10<strong>-7, 10</strong>-2, int(1E4)) * u.m ** -1</code></p><p> </p><p><code>num</code> for <code>linespace</code> needs to be an integer.</p><p> </p><p><strong>Item 2:</strong></p><p> </p><p>In your parameter list you should define <code>ion = "p+"</code> and pass this into <code>ion_sound_speed</code>, <code>Alfven_speed</code> , and <code>gyrofrequency</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

rocco8773, line 1: <p>A couple of you lines are too long and a few of the indentations don't follow PEP8 standards.  Nothing else is wrong, the formatting just needs to be cleaned up a bit.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/932/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

ahmadryan: In the most recent commit, I have updated the lines so that it is consistent with PEP8 standards and lines are of proper length. Can you please check it once and confirm that that's the case?

ahmadryan: Also...in the documentation part I can't figure out how to display the title and it shows as "no title", can you please look into this as well!

docs/index.rst
```diff
@@ -10,7 +10,8 @@ PlasmaPy Documentation
    :hidden:
 
    notebooks/physics
-   notebooks/dispersion_function
+   notebooks/dispersion/dispersion_function
+   notebooks/dispersion/two_fluid_dispersion
```
rocco8773, line 14:
```suggestion
   notebooks/dispersion/two_fluid_dispersion
```

We want to keep the number of notebooks on the RTD landing page to 3.  We're treating them like the feature cards on plasmapy.org.

docs/dispersion/index.rst
```diff
@@ -2,21 +2,25 @@
...
 .. automodule:: plasmapy.dispersion
 
 Example Notebooks
 -----------------
 
 .. nbgallery::
 
-   /notebooks/dispersion_function
+   /notebooks/dispersion/dispersion_function
+   /notebooks/dispersion/two_fluid_dispersion
```
rocco8773, line 15:
```python
.. nbgallery::
   :glob:
   :reversed:

   /notebooks/dispersion/*
```

This will cause the thumbnail gallery to automatically update whenever a new notebook is added.  And, I definitely see more notebooks being added down the road.

changelog/932.feature.rst
```diff
@@ -0,0 +1 @@
+Added a function to solve for the analytical solution of the dispersion relation in low frequency regime.
```
rocco8773, line 1:
```suggestion
Add the function
:func:`~plasmapy.dispersion.two_fluid_dispersion.two_fluid_dispersion_solution` to
`plasmapy.dispersion`, which gives an analytical solution to the dispersion relation as
derived by P. M. Bellan 2012 (DOI: `10.1029/2012JA017856
<https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2012JA017856>`_).
```

The change log acts as an official log for us, so I like entries to be as specific/explicit as possible without losing readability.

rocco8773 (changes requested):
Partial review of just the documentation related elements.

ahmadryan (commented):
Just went through recent changes. Looks alright to me!

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    T_e: u.K,
+    T_i: u.K,
+    gamma_e=1,
+    gamma_i=3,
+    ion="p+",
+    m_e: u.kg = m_e,
+    m_i: u.kg = m_p,
+    theta: u.deg = 45 * u.deg,
+    z=1,
+):
```
rocco8773, line 36:
```suggestion
@validate_quantities(
    B={"can_be_negative": False},
    n_i={"can_be_negative": False},
    T_e={"can_be_negative": False, "equivalencies": u.temperature_energy()},
    T_i={"can_be_negative": False, "equivalencies": u.temperature_energy()},
)
def two_fluid_dispersion_solution(
    *,
    B: u.T,
    ion: Union[str, Particle],
    k: u.rad / u.m,
    n_i: u.m ** -3,
    T_e: u.K,
    T_i: u.K,
    gamma_e: Union[float, int] = 1,
    gamma_i: Union[float, int] = 3,
    theta: u.deg = 45 * u.deg,
    z_mean: Union[float, int] = None,
):
```

Here I've...

1. removed the `m_e` and `m_i` arguments since they were never used in the function
2. made `ion` a required argument (this as been a recent decision to do across the pkg...see issue #453 and PR #911)
3. require `k` to have units `rad/m` to be consistent with it actual units
4. added some typing annotations
5. renamed `z` -> `z_mean` and created default value of `None` (this is more consistent with the `z_mean` usage across the pkg)
6. removed from `@validation_quantities` the validation for `theta` since the default behavior is `can_be_negative == True`

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    n : ~astropy.units.Quantity
+        Number density.
+    T_e : ~astropy.units.Quantity
+        The electron temperature
+    T_i : ~astropy.units.Quantity
+        The ion temperature
+    theta : ~astropy.units.Quantity
+        Angle of propagation which defaults to 45 degrees.
+    z : float or integer
+        Average ionization number which defaults to 1 for protons.
```
rocco8773, line 81:
```suggestion
    B : `~astropy.units.Quantity`
        The magnetic field magnitude in units convertible to :math:`T`.
    ion : `str` or `~plasmapy.particles.particle_class.Particle`
        Representation of the ion species (e.g., ``'p'`` for protons, ``'D+'``
        for deuterium, ``'He-4 +1'`` for singly ionized helium-4, etc.). If no
        charge state information is provided, then the ions are assumed to be
        singly ionized.
    k : `~astropy.units.Quantity`
        Wavenumber in units convertible to :math:`rad / m`.  Either single
        valued or 1-D array of length :math:`N`.
    n_i : `~astropy.units.Quantity`
        Ion number density in units convertible to :math:`m^{-3}`.
    T_e : `~astropy.units.Quantity`
        The electron temperature in units of :math:`K` or :math:`eV`.
    T_i : `~astropy.units.Quantity`
        The ion temperature in units of :math:`K` or :math:`eV`.
    gamma_e : `float` or `int`, optional
        The adiabatic index for electrons, which defaults to 1.  This
        value assumes that the electrons are able to equalize their
        temperature rapidly enough that the electrons are effectively
        isothermal.
    gamma_i : `float` or `int`, optional
        The adiabatic index for ions, which defaults to 3. This value
        assumes that ion motion has only one degree of freedom, namely
        along magnetic field lines.
    theta : `~astropy.units.Quantity`
        The angle of propagation of the wave with respect to the magnetic field,
        :math:`\cos^{-1}(k_z / k)`, in units must be convertible to :math:`deg`.
        Either single valued or 1-D array of size :math:`M`. (Default
        ``45 * u.deg``)
    z_mean : `float` or int, optional
        The average ionization state (arithmetic mean) of the ``ion`` composing
        the plasma.  Will override any charge state defined by argument ``ion``.
```

Here I've updated the docstring to reflect the updates to the function signature and updated wording to be more consistent with the wording across the pkg.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+
+    Returns
+    -------
+    omega : `~astropy.units.Quantity`
+        A dictionary of Wave frequencies corresponding to three modes:
+        (1) Ion-acoustic mode, (2) Alfven mode, and (3) Fast mode, in 1/s units.
```
rocco8773, line 87:
```suggestion
    omega : Dict[str, `~astropy.units.Quantity`]
        A dictionary of computed wave frequencies in units :math:`rad/s`.  The
        dictionary contains three keys: ``'fast_mode'`` for the fast mode,
        ``'alfven_mode'`` for the Alfvén mode, and ``'acoustic_mode'`` for the
        ion-acoustic mode.  The value for each key will be a :math:`N x M` array.
```

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    TypeError
+        The magnetic field and density arguments are not instances of
+        `~astropy.units.Quantity` and cannot be converted into those.
+
+    ~astropy.units.UnitConversionError
+        If the magnetic field or density is not in appropriate units.
+
+    ValueError
+        If the magnetic field, density or wavenumber is negative, or the ion
+        mass or charge state cannot be found.
```
rocco8773, line 100:
```suggestion
    TypeError
        If applicable arguments are not instances of `~astropy.units.Quantity` or
        cannot be converted into one.

    TypeError
        If ``ion`` is not of type or convertible to `~plasmapy.particles.Particle`.

    TypeError
        If ``gamma_e``, ``gamma_i``, or``z_mean`` are not of type `int` or `float`.

    ~astropy.units.UnitConversionError
        If applicable arguments do not have units convertible to the expected
        units.

    ValueError
        If the ``B``, ``k``, ``n_i``, ``T_e``, or ``T_i`` is negative.

    ValueError
        If ``ion`` is not of category ion or element.

    ValueError
        If ``B``, ``n_i``, ``T_e``, or ``T_I`` are not single valued
        `astropy.units.Quantity` (i.e. an array).

    ValueError
        If ``k`` or ``theta`` are not single valued or a 1-D array.
```

Update to include all raised `Exception`s introduced by the argument conditioning (see below review comments).

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+
```
rocco8773, line 101:
```suggestion

    Warns
    -----
    : `~plasmapy.utils.exceptions.PhysicsWarning`
        When the computed wave frequencies violate the low-frequency
        (:math:`\omega/kc \ll 1`) assumption of the dispersion relation.

```

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    >>> c = 3.e8 * u.m/u.s
+    >>> c_s = pfp.ion_sound_speed(T_e=T_e, T_i=T_i, n_e=z * n, ion='p+')
+    >>> v_A = pfp.Alfven_speed( B, n, ion='p+')
+    >>> omega_ci = pfp.gyrofrequency(B=B, particle='p+', signed=False, Z=z)
+    >>> omega = tfds(n=n, B=B, T_e=T_e, T_i=T_i, theta=theta, z=z, k=k_arr)
+    >>> omega['fast_mode'][:,40]
+    <Quantity [1.61176312e-02, 1.77335334e-01, 3.38688590e-01, ...,
+               1.52030361e+03, 1.52045553e+03, 1.52060745e+03] rad / s>
+    """
+
```
rocco8773, line 203:
```suggestion

    # validate argument ion
    if not isinstance(ion, Particle):
        try:
            ion = Particle(ion)
        except TypeError:
            raise TypeError(
                f"For argument 'ion' expected type {Particle} but got {type(ion)}."
            )
    if not (ion.is_ion or ion.is_category("element")):
        raise ValueError(f"The particle passed for 'ion' must be an ion or element.")

    # validate z_mean
    if z_mean is None:
        try:
            z_mean = abs(ion.integer_charge)
        except ChargeError:
            z_mean = 1
    else:
        if not isinstance(z_mean, (int, np.integer, float, np.floating)):
            raise TypeError(
                f"Expected int or float for argument 'z_mean', but got {type(z_mean)}."
            )
        z_mean = abs(z_mean)

    # validate arguments
    for arg_name in ("B", "n_i", "T_e", "T_i"):
        val = locals()[arg_name].squeeze()
        if val.shape != ():
            raise ValueError(
                f"Argument '{arg_name}' must a single value and not an array of "
                f"shape {val.shape}."
            )
        locals()[arg_name] = val

    # validate arguments
    for arg_name in ("gamma_e", "gamma_i"):
        if not isinstance(locals()[arg_name], (int, np.integer, float, np.floating)):
            raise TypeError(
                f"Expected int or float for argument '{arg_name}', but got "
                f"{type(locals()[arg_name])}.")

    # validate argument k
    k = k.squeeze()
    if not (k.ndim == 0 or k.ndim == 1):
        raise ValueError(
            f"Argument 'k' needs to be a single valued or 1D array astropy Quantity,"
            f" got array of shape {k.shape}."
        )
    if np.any(k < 0):
        raise ValueError(f"Argument 'k' can not be a or have negative values.")

    # validate argument theta
    theta = theta.squeeze()
    theta = theta.to(u.radian)
    if not (theta.ndim == 0 or theta.ndim == 1):
        raise ValueError(
            f"Argument 'theta' needs to be a single valued or 1D array astropy "
            f"Quantity, got array of shape {k.shape}."
        )

```

Add in conditioning/validation of input arguements.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    # plasmapy.formulary.parameters
+    v_A = pfp.Alfven_speed(B, n, ion=ion)
+
+    # Compute the ion gyro frequency using the function from
+    # plasmapy.formulary.parameters
+    omega_ci = pfp.gyrofrequency(B=B, particle="p+", signed=False, Z=z)
+
+    # Compute the electron plasma frequency using the function from
+    # plasmapy.formulary.parameters
+    omega_pe = pfp.plasma_frequency(n=n, particle="e-", z_mean=z)
```
rocco8773, line 221:
```suggestion
    # Calc needed plasma parameters
    n_e = z_mean * n_i
    c_s = pfp.ion_sound_speed(
        T_e=T_e,
        T_i=T_i,
        ion=ion,
        n_e=n_e,
        gamma_e=gamma_e,
        gamma_i=gamma_i,
        z_mean=z_mean,
    )
    v_A = pfp.Alfven_speed(B, n_i, ion=ion, z_mean=z_mean)
    omega_ci = pfp.gyrofrequency(B=B, particle=ion, signed=False, Z=z_mean)
    omega_pe = pfp.plasma_frequency(n=n_e, particle="e-")
```

This does not alter what the original code was doing, but updates the variable names to match what the new function signature is.   Additionally, I'm reducing the comments to improve readability.  I've personally been someone that over comments, but I've learned over time that comments should be reserved for what is not directly apparent from reading the code itself.  Otherwise, you have the risk of having conflicting statements between how your code reads and what the comments say.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    # Compute the dimensionless parameters corresponding to equation 35 of
+    # Bellan2012JGR
+    A = (Q + Q ** 2 * beta + Q * alphav + alphav * Lambda) / Q ** 2
+    B = alphav * (1 + 2 * Q * beta + Lambda * beta) / Q ** 2
+    C = alphav ** 2 * beta / Q ** 2
+
+    # Compute the dimensionless parameters corresponding to equation 36 of
+    # Bellan2012JGR
+    p = (3 * B - A ** 2) / 3
+    q = (9 * A * B - 2 * A ** 3 - 27 * C) / 27
```
rocco8773, line 246:
```suggestion
    # Bellan2012JGR params equation 32
    alpha = np.cos(theta.value) ** 2
    beta = (c_s / v_A).to(u.dimensionless_unscaled).value ** 2
    alphav, kv = np.meshgrid(alpha, k.value)  # create grid
    Lambda = (kv * v_A.value / omega_ci.value) ** 2

    # Bellan2012JGR params equation 2
    Q = 1 + (kv * c.value / omega_pe.value) ** 2

    # Bellan2012JGR params equation 35
    A = ((1 + alphav) / Q) + beta + (alphav * Lambda / Q ** 2)
    B = alphav * (1 + 2 * Q * beta + Lambda * beta) / Q ** 2
    C = beta * (alphav / Q) ** 2

    # Bellan2012JGR params equation 36
    p = (3 * B - A ** 2) / 3
    q = (9 * A * B - 2 * A ** 3 - 27 * C) / 27
```

Like above, this does not change the basic functionality but does...

1. reduce comments to improve readability
2. augments the order of operations to reduce the number of operations and avoid squaring large or small number before dividing
3. extracting the astropy `Quantity` values earlier to avoid issues around unit wrangling

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,273 @@
...
+    # Create a dictionary with the wave mode names as its keys
+    omega = dict.fromkeys(keys)
+
+    # Compute the value of  omega for each key and for each value of wavenumber
+    # and direction of propagation
+    for (ind, key) in zip(range(3), keys):
+        # The solution corresponding to equation 38
+        omega[key] = omega_ci * np.lib.scimath.sqrt(
+            R * np.cos(1 / 3 * np.lib.scimath.arccos(S) - 2 * np.pi / 3 * ind) + T
+        )
```
rocco8773, line 267:
```suggestion
    # Bellan2012JGR params equation 38
    R = 2 * Lambda * np.emath.sqrt(-p / 3)
    S = 3 * q / (2 * p) * np.emath.sqrt(-3 / p)
    T = Lambda * A / 3
    omega = {}
    for ind, key in enumerate(("fast_mode", "alfven_mode", "acoustic_mode")):
        # The solution corresponding to equation 38
        w = omega_ci * np.emath.sqrt(
            R * np.cos(1 / 3 * np.emath.arccos(S) - 2 * np.pi / 3 * ind) + T
        )
        omega[key] = w.squeeze()

        # check for violation of dispersion relation assumptions
        # (i.e. low-frequency, w/kc << 0.1)
        wkc_max = np.max(w.value / (kv * c.value))
        if wkc_max > 0.1:
            warn(
                f"The {key} calculation produced a high-frequency wave (w/kc == "
                f"{wkc_max}), which violates the low-frequency (w/kc << 1) assumption "
                f"of the dispersion relation.",
                PhysicsWarning,
            )
```

Again, this does not replace the original functionality of the code but does...

1. replace use of `numpy.lib.scimath.*` with `numpy.emath.*` as per [`numpy`'s suggestion](https://numpy.org/devdocs/reference/routines.emath.html?highlight=emath#module-numpy.lib.scimath)
2. streamlines the definition of the dictionary to improve readability...this part is really a personal so choose the method you prefer
3. add a warning for when/if violates w/kc << 1 (a key assumption behind the dispersion relation)

rocco8773 (changes requested):
This is nearly a complete review of `two_fluid_dispersion.py`.  I've covered everything except the examples in the docstrings.  The basic functionality looks great!  My comments primarily focus on the UI, how the user interacts with the function and what the function reports back (via exceptions and warnings) to the user.

Once this part is complete, I'll look at the tests followed by the notebook.  Things are looking good and I think we'll be able to finish it up soon.

rocco8773:
@ahmadryan Ok, I think I've finished updating the tests and the jupyter notebook.  For the notebook I added more background of the dispersion relation, updated the plots so they work better with RTD, and added a section that duplicates Figure 1 of [Bellan 2012](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2012JA017856).  If you don't mind giving it a read through to ensure everything makes sense and reads correctly, then that would be much appreciated.  No matter how many time I reread the things I write I always seem to miss wordage that make me go 🤦🏻‍♂️.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,352 @@
...
+            gamma_i=gamma_i,
+            z_mean=z_mean,
+        )
+    v_A = pfp.Alfven_speed(B, n_i, ion=ion, z_mean=z_mean)
+    omega_ci = pfp.gyrofrequency(B=B, particle=ion, signed=False, Z=z_mean)
+    omega_pe = pfp.plasma_frequency(n=n_e, particle="e-")
+
+    # Bellan2012JGR params equation 32
+    alpha = np.cos(theta.value) ** 2
+    beta = (c_s / v_A).to(u.dimensionless_unscaled).value ** 2
```
namurphy, line 309: If I remember correctly this isn't strictly the standard plasma beta (gas pressure to magnetic pressure ratio), but rather that there's a factor of order unity that's dependent on what `gamma_i` is (I think).  I'm wondering if we would want to call this `beta_prime`.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,352 @@
...
+            gamma_e=gamma_e,
+            gamma_i=gamma_i,
+            z_mean=z_mean,
+        )
+    v_A = pfp.Alfven_speed(B, n_i, ion=ion, z_mean=z_mean)
+    omega_ci = pfp.gyrofrequency(B=B, particle=ion, signed=False, Z=z_mean)
+    omega_pe = pfp.plasma_frequency(n=n_e, particle="e-")
+
+    # Bellan2012JGR params equation 32
+    alpha = np.cos(theta.value) ** 2
```
namurphy, line 308: We don't need to change anything, but I thought I'd mention that Python lets us use a wide change of unicode characters in code.  If we wanted to, we could have variables named `α` instead of `alpha`.  We wouldn't want to do that for code that needs to be revised a lot, but would be a way to help improve math readability.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,352 @@
...
+
+    .. math::
+        Q &= 1 + k^2 c^2/{\omega_{pe}}^2 \\
+        \cos \theta &= \frac{k_z}{k} \\
+        \mathbf{B_o} &= B_{o} \mathbf{\hat{z}}
+
+    :math:`\omega` is the wave frequency, :math:`k` is the wavenumber, :math:`v_A`
+    is the Alfvén velocity, :math:`c_s` is the sound speed, :math:`\omega_{ci}` is
+    the ion gyrofrequency, and :math:`\omega_{pe}` is the electron plasma frequency.
+    This relation does additionally assumme low-frequency waves
```
namurphy, line 154:
```suggestion
    This relation does additionally assume low-frequency waves
```

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,352 @@
...
+                \frac{1}{3} \cos^{-1}\left(
+                    \frac{3q}{2p} \sqrt{-\frac{3}{p}}
+                \right)
+                - \frac{2 \pi}{3}j
+            \right)
+            + \frac{\Lambda A}{3}
+        }
+
+    where :math:`j = 0` represents the fast mode, :math:`j = 1` represents the
+    Alfvén mode, and :math:`j = 2` represents the Acoustic mode.  Additionally,
```
namurphy, line 174:
```suggestion
    Alfvén mode, and :math:`j = 2` represents the acoustic mode.  Additionally,
```
I usually see acoustic/magnetoacoustic/magnetosonic lower case, though as long as it's consistent it would be fine with me.

namurphy (commented):
Looking good!  And thank you for working on this!  Most of my comments are pretty minor, like spelling or capitalization.  The only substantive comment is that it'd be helpful to use the [`@particle_input`](https://docs.plasmapy.org/en/stable/particles/decorators.html#passing-particle-instances-to-functions-and-methods) decorator instead of having code in the function that does the processing and checking of particles.

plasmapy/dispersion/two_fluid_dispersion.py
```diff
@@ -0,0 +1,352 @@
...
+    # validate argument ion
+    if not isinstance(ion, Particle):
+        try:
+            ion = Particle(ion)
+        except TypeError:
+            raise TypeError(
+                f"For argument 'ion' expected type {Particle} but got {type(ion)}."
+            )
+    if not (ion.is_ion or ion.is_category("element")):
+        raise ValueError(f"The particle passed for 'ion' must be an ion or element.")
```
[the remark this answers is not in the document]
rocco8773: If you don't mind, I'd like to add `@particle_input` after we develop and deploy `particle_like`.  `@particle_input` just doesn't play well with other decorators so I'd like to avoid unforeseen complications until `@particle_input` goes through its update.

plasmapy/dispersion/two_fluid_dispersion.py
rocco8773: Unless I'm missing something, I believe this is the plasma beta.  Bellan and Stringer use the beta nomenclature, so I think it's better to keep `beta` to be consistent with the literature, and the docstrings.

rocco8773:
@namurphy Thanks for the quick read through!  I'm going to leave the PR open for now to give @ahmadryan a chance to read through and comment.  I'm hoping we'll merge Monday or Tuesday.

plasmapy/dispersion/two_fluid_dispersion.py
ahmadryan: Yes, I got to know that recently too. Will keep this in mind going forward!

plasmapy/dispersion/two_fluid_dispersion.py
ahmadryan: I agree with @rocco8773, for the sake of consistency with the original document, let's just leave it as `beta`. Even in literature we always find two different definitions of `beta` which differ from each other by a factor of 2, so there is that!

ahmadryan:
@rocco8773 @namurphy I guess we can merge this to the main branch of PlasmaPy during today's meeting!
Thank you for the code review and all the suggestions!

plasmapy/dispersion/two_fluid_dispersion.py
rocco8773: Correct.  The Bellan beta is a factor of 2 off from the typical beta, `beta = nkT/(B/(2 mu0)) = 2 beta_bellan`.

rocco8773:
@ahmadryan Sure, I can do it during our call. 👍

rocco8773 (approved):
👍

PR #933, opened 2020-10-15
Formatting fixes such as for when multiple strings are on one line

namurphy:
We use [black](https://black.readthedocs.io/en/stable/) as a code formatter.  One of black's idiosyncrasies is that it sometimes puts multiple strings on one line if they'll both fit.  This PR mostly includes formatting fixes, such as the change below:
```Python
        f"Unknown field orientation " f"'{self.field_orientation}'"   # two strings
        f"Unknown field orientation '{self.field_orientation}'"  # replace with one string
 ```
I'm skipping `ionization_state.py`, etc. because I'm doing these fixes in #796 for those files.

### DIFF
--- a/plasmapy/diagnostics/tests/test_langmuir.py
+++ b/plasmapy/diagnostics/tests/test_langmuir.py
@@ -1,4 +1,3 @@
-# coding=utf-8
 """Tests for Langmuir probe analysis functions."""
 
 import astropy.constants.si as const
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

**Comment by namurphy** [comment]
Oh goodness.  I just found an inconsistency between when we run black locally and our pre-commit thing.  

The pre-formatter wants it like this:
```Python
    assert np.isclose(
        ion_sound_speed(1.831 * u.MK, 1.3232 * u.MK, "p").value, 218816.06086407552,
    )
```
...but running it locally changes it to this:
```Python
    assert np.isclose(
        ion_sound_speed(1.831 * u.MK, 1.3232 * u.MK, "p").value,
       218816.06086407552,
    )
```

 - [ ] Change the settings for black or our pre-commits to remove this inconsistency

rocco8773:
I'd like to figure out how to tweak `black` a little bit.  There are somethings it does, which I feel reduces readability of the code.  For example, the two examples you show are valid PEP8 formats, but I think your local version is more readable.  I'd like to figure out how to make the `pre-commit` version of `black` allow both versions.

rocco8773 (approved):
All looks good to me.  Can I merge, or do plan to make additional changes?

namurphy:
Feel free to merge!  The code style guide in the development docs needs an update, but that's for a different series of ten pull requests in a row.

PR #929, opened 2020-10-13
Test helper refactoring 10: Update narrative docs to discuss usage of test_runner

namurphy:
This PR, originally from #728, updates the narrative documentation in the development guide to discuss the usage of `test_runner`, which is being implemented in #928.

PR #908, opened 2020-10-02
Create plasmapy.analysis sub-pkg & add fit functions as initial functionality

rocco8773:
Fit functions are a set of callable classes that wraps together the following functionality to perform curve fitting...

1. A callable version of the fitted function.
1. A curve fitting method to define the fit function parameters and associated errors.
1. Error propagation for the fitted curve
1. Root solver for the fitted function

This functionality was originally developed for the swept Langmuir analysis in PR #889, but soon went beyond the PR's scope so I broke it out into this PR.  This PR has to be merged before PR #889.

----

This PR does the following...

- [x] Creates `plasmapy.analysis`.
- [x] Initialize `plasmapy.analysis` with `fit_functions.py`.
- [x] Updates the documentation to incorporate the "Analysis" documentation with the "Diagnostics" documentation.
- [x] Adds associated tests.
- [x] Add a jupyter notebook example.

PR #943, opened 2020-10-23
Decorator @modify_docstring

rocco8773:
This is a simple decorator that allows for the programmatic prepending and/or appending of docstrings.  The original docstring is stored in a new `__original_doc__` attribute so it can still be retrieved.

This is useful in a scenario where you are overriding a method and the signature does not change, but you want add a little more info to the original docstring.

### DIFF
--- a/plasmapy/utils/decorators/helpers.py
+++ b/plasmapy/utils/decorators/helpers.py
@@ -37,9 +37,9 @@ def modify_docstring(func=None, prepend: str = None, append: str = None):
         ...     '''Beautiful'''
         ...     pass
         >>> foo.__original_doc__
-        "Beautiful"
+        'Beautiful'
         >>> foo.__doc__
-        "Hello\\n\\nBeautiful\\n\\nWorld"
+        'Hello\\n\\nBeautiful\\n\\nWorld'
 
     """
 
--- a/plasmapy/utils/decorators/helpers.py
+++ b/plasmapy/utils/decorators/helpers.py
@@ -87,6 +87,7 @@ def modify_docstring(func=None, prepend: str = None, append: str = None):
                 f"Expected type str for argument 'append', got {type(append)}."
             )
 
+        # define new docstring
         wrapper.__doc__ = "\n".join(prependlines + doclines + appendlines)
 
         return wrapper
--- a/plasmapy/utils/decorators/helpers.py
+++ b/plasmapy/utils/decorators/helpers.py
@@ -7,7 +7,7 @@ import functools
 import inspect
 
 
-def modify_docstring(func=None, prepend=None, append=None):
+def modify_docstring(func=None, prepend: str = None, append: str = None):
     """
     A decorator which programmatically prepends and/or appends the docstring
     of the decorated method/function.  The unmodified/original docstring is
--- a/plasmapy/utils/decorators/helpers.py
+++ b/plasmapy/utils/decorators/helpers.py
@@ -32,15 +32,17 @@ def modify_docstring(func=None, prepend: str = None, append: str = None):
     Examples
     --------
 
-    >>> @modify_docstring(prepend='''Hello''', append='''World''')
-    ... def foo():
-    ...     '''Beautiful'''
-    ...     pass
-    >>> foo.__original_doc__
-    'Beautiful'
-    >>> foo.__doc__
-    'Hello\n\nBeautiful\n\nWorld'
+        >>> @modify_docstring(prepend='''Hello''', append='''World''')
+        ... def foo():
+        ...     '''Beautiful'''
+        ...     pass
+        >>> foo.__original_doc__
+        "Beautiful"
+        >>> foo.__doc__
+        "Hello\\n\\nBeautiful\\n\\nWorld"
+
     """
+
     def decorator(f):
         sig = inspect.signature(f)
 
--- a/plasmapy/utils/decorators/helpers.py
+++ b/plasmapy/utils/decorators/helpers.py
@@ -1,10 +1,100 @@
 """
 Miscellaneous decorators for various package uses.
 """
-__all__ = ["preserve_signature"]
+__all__ = ["modify_docstring", "preserve_signature"]
 
 import functools
 import inspect
+
+
+def modify_docstring(func=None, prepend=None, append=None):
+    """
+    A decorator which programmatically prepends and/or appends the docstring
+    of the decorated method/function.  The unmodified/original docstring is
+    saved as the ``__original_doc__`` attribute.
+
+    Parameters
+    ----------
+    func: callable
+        The method/function to be decorated
+
+    prepend: `str`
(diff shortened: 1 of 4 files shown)


plasmapy/utils/decorators/helpers.py
```diff
@@ -1,12 +1,105 @@
...
+    Parameters
+    ----------
+    func: callable
+        The method/function to be decorated.
+
+    prepend: `str`
+        The string to be prepended to the ``func``'s docstring.
+
+    append: `str`
+        The string to be appended to the ``func``'s docstring.
```
namurphy, line 25:
```suggestion
        The string to be prepended to ``func``'s docstring.

    append: `str`
        The string to be appended to ``func``'s docstring.
```
Slight wording change

plasmapy/utils/decorators/helpers.py
```diff
@@ -1,12 +1,105 @@
...
+    Returns
+    -------
+    callable
+        Wrapped version of the function.
+
+    Examples
+    --------
+
+        >>> @modify_docstring(prepend='''Hello''', append='''World''')
+        ... def foo():
```
namurphy, line 36:
```suggestion
        ... def vegan_spam():
```

plasmapy/utils/decorators/helpers.py
```diff
@@ -1,12 +1,105 @@
...
+        wrapper.__doc__ = "\n".join(prependlines + doclines + appendlines)
+
+        return wrapper
+
+    if func is not None:
+        # `modify_docstring` called as a function
+        return decorator(func)
+    else:
+        # `modify_docstring` called as a decorator "sugar-syntax"
+        return decorator
```
namurphy, line 100: This comes up for a whole bunch of decorators that have optional arguments, so I'm wondering if we should create a function for it.  Doesn't have to be now, though.

plasmapy/utils/decorators/tests/test_helpers.py
```diff
@@ -1,14 +1,114 @@
...
+# --------------------------------------------------------------------------------------
+class TestModifyDocstring:
+    # create function to test on
+    @staticmethod
+    def foo_simple(x: float, y: float) -> float:
+        """A simple docstring."""
+        return x + y
+
+    @staticmethod
+    def foo_complex(x: float, y: float) -> float:
```
namurphy, line 21:
```suggestion
    def simple_docstring(x: float, y: float) -> float:
        """A simple docstring."""
        return x + y

    @staticmethod
    def complex_docstring(x: float, y: float) -> float:
```
I was thinking that the intent would be more clear with something like these names.

plasmapy/utils/decorators/helpers.py
```diff
@@ -1,12 +1,105 @@
...
+            appendlines = [""] + appendlines
+        elif append is None:
+            appendlines = []
+        else:
+            raise TypeError(
+                f"Expected type str for argument 'append', got {type(append)}."
+            )
+
+        # define new docstring
+        wrapper.__doc__ = "\n".join(prependlines + doclines + appendlines)
```
namurphy, line 91: This decorator has the potential to get significantly more complex, so I would suggest break up the prepending and appending (prependation and appendation?) into different private functions that can be called inside the decorator. 

It could be something vaguely along the lines of:
```Python
doclines = _save_original_docstring(...)
prependlines = _prepend_docstring_lines(prepend)
appendlines = _append_docstring_lines(append)
wrapper.__doc__ = _define_new_docstring(prependlines, appendlines, doclines)
```
where the comments describing the section of code become the function names (though there's probably a better way to break it up than the above code block).  I think of this pattern of refactoring when there are sections of code separated by comments.  I think I originally read it in _Clean Code_ by R. Martin.  We could also use the strategy of dropping one level of abstraction per function.

I'm also mentioning this because if I had followed this advice with the original implementation of `plasmapy.utils.pytest_helpers.run_test`, I wouldn't have needed to go through with the test helper refactoring that I'm doing in #919 through #929...oof.

namurphy (changes requested):
I do very much like this decorator!  As we talked about earlier, we can wait until a future pull request to expand the functionality for this decorator...for example to inject the descriptions of common arguments like `B` that we keep having to repeat between functions.

The only substantive comment I have is to break out the different tasks in the decorator into private functions...since this will make it easier to work with the decorator later on should we expand its functionality.

Also...the file name `test_helpers.py` has the potential for ambiguity with the test helpers that I'm working on starting in #919.  I'm wondering if we could change names to avoid this overlap.  It's not important, but it did cause me a moment of confusion.

Thank you for doing this!

plasmapy/utils/decorators/helpers.py
rocco8773: Ha!  This is a Southern California thing.  We are know for putting "the" in front instances like this.  It's believed to come from the fact we've named all of our interstates.  For example, 405 is known as "the San Diego freeway" so we say "the 405."

Since this is so ingrained in me, I can't bring myself to change it.  If it really bothers you, then you'll have sneak the change in yourself. 🤣

plasmapy/utils/decorators/helpers.py
rocco8773: Maybe for the future.  I do always look up one of our old decorators to remind me how to do it.

plasmapy/utils/decorators/helpers.py
rocco8773: Yeah, I do have several ideas to tackle more complex scenarios.  I'll probably migrate this to a class style decorator, much like we do for `validate_quantities` and `ValidateQuantites`.  Once there, I'll have functionality for injecting into specific `numpydoc` sections.

However,  I want to leave that for the future.  I want to get this through so I can get PR #908 merged and move on with the diagnostics.  That is much more important to me at this point.

plasmapy/utils/decorators/helpers.py
rocco8773: This is staying `foo`. 😁

rocco8773:
> Also...the file name `test_helpers.py` has the potential for ambiguity with the test helpers that I'm working on starting in #919. I'm wondering if we could change names to avoid this overlap. It's not important, but it did cause me a moment of confusion.

If we rename `test_helpers.py`, then we have to rename `plasmapy/utils/decorators/helpers.py` too.  However, I don't think this is needed.  Functionality in #919 is in a completely different location to the decorators, it doesn't name any `test_helpers.py` file, and it's standard practice the `test_<something>.py` are tests for the `<something>.py` file one level up.  We can't us unique names for every part of the package.

plasmapy/utils/decorators/helpers.py
namurphy: Sure!  I'm happy to save that for the future, in particular if we create an issue or two on GitHub so that we can keep track of it for the future.

namurphy (approved):

PR #952, opened 2020-11-22
Move tests from Travis to Github Actions

StanczakDominik:
I'll explain later, need tests to start working first...


StanczakDominik:
Okay, this is beginning to look good.

The reason for this is that Travis is beginning to retreat from the open source market because people were abusing it to mine crypto etc. I thus replaced it with GitHub actions (all hail to our new Microsoft overlords!).

In our setup Travis doesn't do much that Azure didn't, so it got relegated to cron jobs a while back. I'm pretty sure I've got that covered. Come to think of it, though, the Actions set up does look nicer than the Azure one (it seems like it's going to be easier to parse errors this way), so **I'd actually be in favor of using Actions for PRs instead**.

The only thing I'd keep from Azure (for now) is wheel builds for releases, but I'm pretty sure I'm going to be able to automate that soon as well. I'm still figuring out code coverage.

cc @rocco8773 @namurphy

StanczakDominik:
wrt code coverage, I'm still not sure whether it works correctly but it'd be much easier to set up "on the spot" rather than semi-blindly as I'm trying to do now.

I also have another [PR on my fork](https://github.com/StanczakDominik/PlasmaPy/pull/3) that disables PR builds on Azure. Basically, all the Azure check entries would disappear. This is nice because the tests become much more user friendly.

tox.ini
```diff
@@ -1,7 +1,14 @@
 [tox]
-envlist = py37,build_docs
+envlist = py36,py37,py37-astropydev,py36-minimal,py38,py38-numpydev,py39,build_docs
```
namurphy, line 2: Thanks for adding in Python 3.9!  I'm also inclined to drop Python 3.6 support soon, though that's probably out of scope for this PR.

tox.ini
```diff
@@ -1,7 +1,14 @@
...
 isolated_build = True
 
+[gh-actions]
+python =
+    3.7: py37, py37-astropydev
+    3.6: py36, py36-minimal
+    3.8: py38, py38-numpydev
```
namurphy, line 9: I like how the `numpydev` and `astropydev` test runs are staggered with respect to Python versions.  That keeps the tests happening, while not making our test suite excessively long.

namurphy (commented):
This is looking good so far!  I like the idea of using GitHub Actions for most of our tests.  I only took a quick look at this today and will try to look at it again before the meeting tomorrow when I have fewer things due by 5 pm Pacific time.

tox.ini
StanczakDominik: Admittedly, come to think of it, we might prefer not to run those on every PR, just on master.

StanczakDominik: Erik pointed out that that approach would delay figuring out that things break on dev versions, and if two PRs get merged between the weekly cron run, we would not be able to easily identify the offender. It's thus better to keep them here.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches: [ master ]
+  pull_request:
+  schedule:
+    # runs these tests every Sunday, at 12 UTC
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onschedule
+    - cron:  '0 12 * * 0'
+
+jobs:
+  build:
+    if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.event.head_commit.message, '[skip github]')"
```
namurphy, line 14: We should add to the developer docs that we can put `[ci skip]`, etc. in commit messages to skip the CI stuff.

namurphy (approved):

Issue #936, opened 2020-10-16
Improvements to ParticleTracker


pheuer:
With some improvements/additions, the ParticleTracker module could be really useful for the LineIntegratedDiagnostic project. I'm starting this issue as a place to keep track of some possible improvements. 

Here's a list of necessary changes:
- [x] Modify the ParticleTracker to  take `grid` objects that contain E and B fields. Make use of the interpolators built into those objects for the interpolation step.

- [x] Allow particles to be initialized with positions, velocities, and species specified by the user. Based on our discussion, we'll probably use a `ParticleSpecies` class that represents a bunch of particles all of the same species.

- [ ] Provide an alternative `boris_push` that's correct for relativistic particles (for example, protons in proton radiography are often moving at 20% c). 

- [x] Make recording the position and velocity histories optional: that's a big performance/memory hit for large numbers of particles that may not be necessary. Also have an option to write that stuff out to a file instead of RAM. Can we use `h5py` for that?

- [ ] Implement a more sophisticated auto-timestep function with conditions based on E, B, and the grid size. Also, the user should still have the ability to set their own timestep.

- [ ] Allow the push functions to take arrays of q and m (so that particles don't all have to be the same species)

Stretch goals (not necessary now, but would be really helpful in the future):

- [ ] Use cython or numba to improve performance of all the interpolator and particle_pusher methods.

- [ ] Provide an option to keep a log of any grid quantities that the particles encounter at each time step, Eg. record a time trace of the B-field felt by a particle. Should be an option to record either the full time series or a cumulative sum.

StanczakDominik:
Quick note, haven't had much time to think about the rest yet:

>  Allow particles to be initialized with positions, velocities, and species specified by the user. I propose that species be a list of particle types (nspecies long) and the position and velocity arrays x,y,z,vx,vy,vz be of shape [nspecies, nparticles].

For a simulation I think I'd rather see it as this:

```python
electrons = Species('e', 500)
electrons.r = np.random.random((500, 3)
electrons.v = # likewise

# ions likewise

ParticleTracker([electrons, ions], ...)
```

pheuer:
That sounds good too - is the `Species` class already defined somewhere? Maybe `Species` or `Population` or something could be a subclass of `BasePlasma` in the `Plasma` module?

pheuer:
[...]

PR #953, opened 2020-11-22
ParticleTracker refactor: Bring in particle_integrators.py

StanczakDominik:
This is (finally!) a follow-up to the closed #675. It brings in a separated implementation of the Boris algorithm, which I was hoping would solve the issue raised in https://plasmapy.discourse.group/t/e-cross-b-drift/; we're not there yet, but this is a change in the right direction and should be easier to review.

7f20252 is marked [REVERT], but I haven't yet found a way to combine astropy.units with parallel numba, and in my previious tests the parallelization did not bring huge speedups, so it's fine to keep it that way for now.

This separate implementation will also make #895 easier.


StanczakDominik:
`black`'s failure gets fixed in #952.

pheuer:
This PR would also address a couple of the points raised in issue #936

StanczakDominik:
Merging this now would break Python 3.9 support, because that version broke Numba (https://github.com/numba/numba/issues/6345) and we won't get that fixed until RC 2021.

rocco8773:
Could you possibly have a `numba` and non-`numba` version by doing something like....

```python

try:
    import numba
    _range = numba.prange
except ImportError:
    warn(
        "If running Python < 3.9 consider installing numba.  This will allow parallel "
        "computing and speed up ParticleTracker"
    )
    _range = np.arange

def foo(*args):
    # use _range for the loops so the function is automatically a numba or 
    # non-numba version
    ...

if "numba" in globals:
    foo = numba.njit()(foo)
```

Then we might be able to use [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) so `numba` is only installed on `python < 3.9`. (see also https://stackoverflow.com/a/32643122)  I've never done this, so I'm not sure how well it will work.

StanczakDominik:
It's still a for loop. If we do "optional Numba" here, the performance will suffer greatly over the current implementation.

rocco8773:
> It's still a for loop. If we do "optional Numba" here, the performance will suffer greatly over the current implementation.

I get that.  The idea is that, with the environmental markers we can make `numba` a dependency for installs on `python < 3.9` and thus utilized the `numba` functionality you're integrating here.  If someone is using `python >= 3.9`, then `numba` will not be installed, the particle tracker functionality will revert to a slower form, and the user is warned about what is happening.  However, the UI is made consistent across the python versions so when the `numba` with `python39` issue is resolved the user does not have to change anything when make the patch on our end.

StanczakDominik:
I went without numba for now. I figure we can wait until Q1 2021 and introduce less complexity.

In the meantime, I went with refactoring the numpy version out of the class, and lo and behold, I fixed the ExB drift direction bug. I've got the numba version backed up, but for now, this should be good.

StanczakDominik:
Looks ready to review! MacOS build cancellation smells of intermittency.

changelog/953.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug in the Particle Tracker simulation. The integrator had a sign
+error, thus the ExB drift came out in the incorrect direction.
```
namurphy, line 2:
```suggestion
error, thus the E×B drift came out in the incorrect direction.
```
Unicode!

plasmapy/simulation/particle_integrators.py
```diff
@@ -0,0 +1,76 @@
...
+    Parameters
+    ----------
+    x : np.ndarray
+        particle position at full timestep
+    v : np.ndarray
+        particle velocity at half timestep
+    b : np.ndarray
+        magnetic field at full timestep
+    e : float
+        electric field at full timestep
```
namurphy, line 20: I'm wondering about adding some more detail to the arguments.  Reading this, I could end up being confused about the `ndarray` objects...like if they are three-vectors corresponding to x, y, z.  I could also get confused about what units are assumed.

plasmapy/simulation/particle_integrators.py
```diff
@@ -0,0 +1,76 @@
...
+    vminus = v + hqmdt * e
+
+    # rotate to add magnetic field
+    t = b * hqmdt
+    s = 2 * t / (1 + (t * t).sum(axis=1, keepdims=True))
+    vprime = vminus + np.cross(vminus, t)
+    vplus = vminus + np.cross(vprime, s)
+
+    # add second half of electric impulse
+    v[...] = vplus + hqmdt * e
```
namurphy, line 61: Huh, I can't remember seeing this notation before.  I've pretty much always done `v[:] =`.  Interesting to learn!

namurphy (changes requested):
Looking good to me so far!  My main comment is to add a little bit more detail in the `boris_push` docstring to reduce the chance of ambiguity.  Thanks!

pheuer:
@StanczakDominik As expected I found another sign error elsewhere in my proton radiography code, so I'm now pretty sure you have the signs right in this version of `boris_push` (eg. no minus sign in the definition of `t`).

changelog/953.bugfix.rst
StanczakDominik: That's some good application of unicode, I love it! But I think it's confusing enough for the mobile GitHub app that I can't merge the suggestion without a crash... Will merge after reporting it:D

plasmapy/simulation/particle_integrators.py
StanczakDominik: Yeah, that would have been fine as well! In fact it might have some subtle differences, I'll check back on that - thanks for pointing it out...

plasmapy/simulation/particle_integrators.py
```diff
@@ -0,0 +1,76 @@
...
+
+    x += v * dt
+
+
+c = constants.c.si.value
+
+
+def _Lorentz_factor(u):
```
pheuer, line 69: The Lorentz factor is already implemented in `formulary.relativity' (although that version is for velocities). Maybe it would be better to modify that version to accept momentum as well (calculating gamma as appropriate) and then import that here?

StanczakDominik: Fair point, though I did want to have a version without the overhead on units in here. For now, it's not being used (I had another pusher in here that did incorporate it) so I'll just remove it for now.

StanczakDominik:
Should be good to go!

StanczakDominik:
@namurphy here's the promised reminder :)

plasmapy/simulation/particle_integrators.py
```diff
@@ -0,0 +1,71 @@
...
+    time, and the algorithm, being a symplectic integrator, conserves energy.
+
+    References
+    ----------
+    .. [1] C. K. Birdsall, A. B. Langdon, "Plasma Physics via Computer
+           Simulation", 2004, p. 58-63
+    .. [2] L. Brieda, "Particle Push in Magnetic Field (Boris Method)",
+           https://www.particleincell.com/2011/vxb-rotation/
+    """
+    hqmdt = 0.5 * dt * q / m
```
namurphy, line 59: After working on a pull request on physical types in Astropy (https://github.com/astropy/astropy/pull/11204), it's really bugging me to not be able to figure out what sort of physical type corresponds to units of C · s / m!

namurphy (approved):
I like this refactoring, so thank you!  My final suggestions are pretty minor and mostly aren't even necessary.  I think it's ready to be merged whenever you're ready to.

plasmapy/simulation/particle_integrators.py
StanczakDominik: I'll have to take a closer look at that one!

PR #956, opened 2020-12-02
Pick up astropy's GitHub Actions config

StanczakDominik:

      docstrings.

I started reading Astropy's github-actions config and picked up a ton of useful stuff. For example:

* I did some light refactoring of tox.ini and the configs.
* I split up tox envs between different github runners. I thus removed my previous idea of using tox-gh-actions, which I was dissatisfied with.
* Most importantly, the errors now pop up right away when looking at tests, making test output cleaner. and easier to understand.

Follow up to #952.

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -1,10 +1,18 @@
+name: Style linters
+
 on:
   pull_request:
   push:
     branches: [master]
 
-name: Style linters
 jobs:
+  pre-commit:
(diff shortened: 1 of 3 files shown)


namurphy (approved):
I'm only moderately familiar with these test configurations, but all of the changes look like helpful improvements as far as I can tell.  Thank you for doing this!

StanczakDominik:
Thanks a lot!

PR #959, opened 2020-12-15
Patch root_solve() of fit function Linear()

rocco8773:
This patch updates the `root_solve()` method of `plasmapy.analysis.fit_function.Linear` to handle the case where the slope is zero.  If this scenario occurs, then there are not finite roots and `(np.nan, np.nan)` is returned.

### DIFF
--- a/plasmapy/analysis/fit_functions.py
+++ b/plasmapy/analysis/fit_functions.py
@@ -594,7 +594,7 @@ class Linear(AbstractFitFunction):
 
         if m == 0.0:
             warn(
-                f"Slope of Linear fit function is zero so not finite root exists. ",
+                f"Slope of Linear fit function is zero so no finite root exists. ",
                 RuntimeWarning,
             )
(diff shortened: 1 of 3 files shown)


plasmapy/analysis/fit_functions.py
```diff
@@ -590,10 +591,20 @@ def root_solve(self, *args, **kwargs):
             :attr:`parameters` and :attr:`parameters_err`.
         """
         m, b = self.params
+
+        if m == 0.0:
+            warn(
+                f"Slope of Linear fit function is zero so not finite root exists. ",
```
namurphy, line 597:
```suggestion
                f"Slope of Linear fit function is zero so no finite root exists. ",
```

namurphy (approved):
One slight change but otherwise looks pretty good and straightforward to me.  Thanks!

rocco8773:
@namurphy This is ready to merge if you have no other comments.

StanczakDominik (approved):

PR #961, opened 2020-12-16
Update development guide

namurphy:
A few parts of our development guide have become obsolete, so the purpose of this pull request is to provide some updates for those relevant parts.  There are a few broader things I'd like to do to revamp different parts of our development guide, but I think that would be best done incrementally over a few pull requests.

### DIFF
--- a/docs/development/release_guide.rst
+++ b/docs/development/release_guide.rst
@@ -84,7 +84,7 @@ Release
 
 * Push the tagged commit to the version's branch on GitHub: ``git push --force
   --follow-tags upstream v0.6.x``. Note that ``--force`` is necessary to trigger
-  a rebuild with the tagged version. This kicked us in the posterior for ``0.6.0``.
+  a rebuild with the tagged version. This kicked us in the posterior for ``0.4.0``.
 
 At this point, `the OpenAstronomy Azure Pipelines
 <https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html>`_
--- a/docs/development/release_guide.rst
+++ b/docs/development/release_guide.rst
@@ -12,19 +12,19 @@ and may refer to `Astropy's release procedures
 <http://docs.astropy.org/en/stable/development/releasing.html>`_ for
 guidance.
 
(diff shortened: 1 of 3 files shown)


docs/development/release_guide.rst
```diff
@@ -69,22 +69,22 @@ Release
...
 
 * Tag the new version with ``git tag -s v<version> -m "Version v<version>"``
 
   * Note that ``-s`` signs the commit with your GPG key.
 
 * Push the tagged commit to the version's branch on GitHub: ``git push --force
-  --follow-tags upstream v0.5.x``. Note that ``--force`` is necessary to trigger
-  a rebuild with the tagged version. This kicked us in the posterior for ``0.5.0``.
+  --follow-tags upstream v0.6.x``. Note that ``--force`` is necessary to trigger
+  a rebuild with the tagged version. This kicked us in the posterior for ``0.6.0``.
```
StanczakDominik, line 87: It's probably unnecessary to bump all of these up for each release - and this line was not even historically accurate in the first place, we had issues with the tagged version not being rebuilt in 0.4.0 :smile:

namurphy: Ah, oops!  I've had enough off-by-one errors in my life so I figured it'd be worth the 20 seconds to find and replace since I was in the area anyway, but this led to an off-by-one error that became an off-by-two error.  Thank you for finding this!

StanczakDominik (approved):

StanczakDominik:
Hey, I was going to add some more documentation on pre-commit. Since you're touching the same files, and in fact writing a little on the same subject, is it fine if I build on this branch as well?

StanczakDominik:
Groan. I just realized we literally have a paragraph on pre-commit just below. I'll submit my changes (rethought from the above a little) in a separate PR, and let's merge this for now, as a straight-up improvement over the previous version :)

namurphy:
Sure, that's fine.  It'll probably be more effective to make incremental changes to the development guide than one large pull request to overhaul it...especially since I've let this PR sit idle for a while.

PR #963, opened 2020-12-16
Grids (part 2)

pheuer:
This PR further develops the grids functionality, including documentation notebooks. 

A summary change log:

**Major**
- A list of `recognized_keys` in `grids.py` has been started to standardize dictionary keys for different quantities (eg. E, B, n_e)
- Created two documentation notebooks for `grids.py`

**Minor**
- Nearest-neighbor interpolation on non-uniform grids is now implemented
- Grids now have a `__repr__` string that summarizes their contents (like xarray)
- Fixed a bug when interpolating outside of the grid (now returns zero instead of crashing).
- Created many new tests.

Apologize for the long list of old commits carried over from the last grids PR: I probably should have started a new branch for this...


plasmapy/plasma/grids.py
```diff
@@ -336,6 +419,29 @@ def add_quantity(self, key: str, quantity: u.Quantity):
...
+        r"""
+        Adds a list of keys and quantities to the grid. See "add_quantity"
+
+        Parameters
+        ----------
+        keys : list
+            A list of str keys for quantities to add
+        quantities : list
+            A list of quantities to add. Each must be an array that matches
+            the size of the grid
```
StanczakDominik, line 432: This smells like it should take a single dictionary argument instead, doesn't it? Maybe even something like
```
    def add_quantities(self, **quantities: dict):
        r"""
        Adds a list of keys and quantities to the grid. See "add_quantity"

        Parameters
        ----------
        quantities :
            quantities to add. Each must be an array that matches
            the size of the grid. The argument names should fit `recognized_keys`
```

But I'm not sure how much I like that versus the single dictionary option, so I'm not making this a "suggestion" yet. I'll note that a single dictionary could then be passed in as `grid.add_quantities(**my_awesome_dictionary)`.

... come to think of it, I think this option would make `add_quantity` redundant; `grid.add_quantity("E_x", E_x_array)` would be equivalent to `grid.add_quantities(E_x=E_x_array)`. The naming might then use adjustment, of course.

plasmapy/plasma/grids.py
```diff
@@ -734,3 +904,25 @@ def _make_mesh(self, start, stop, num, **kwargs):
...
+
+        Ex, Ey, Ez = np.gradient(potential, grid.dax0, grid.dax1, grid.dax2)
+
+        Ex = -np.where(radius < b, Ex, 0)
+        Ey = -np.where(radius < b, Ey, 0)
+        Ez = -np.where(radius < b, Ez, 0)
+
+        grid.add_quantities(["E_x", "E_y", "E_z"], [Ex, Ey, Ez])
+
+    return grid
```
StanczakDominik, line 928: I would definitely suggest putting a `Value` or `KeyError` in an `else` clause here. Otherwise you'll see mildly confused people asking, hey, why is my `guasian_sphere` an empty grid? :)

StanczakDominik (changes requested):
Looks great! Some comments, though:

* the notebooks should probably be linked from somewhere around https://plasmapy--963.org.readthedocs.build/en/963/api_static/plasmapy.plasma.grids.html
* a couple notes on the notebooks themselves, coming right up through reviewnb!

docs/notebooks/grids_cartesian.ipynb
```diff
@@ -0,0 +1,354 @@
+{
```
StanczakDominik, line 1: <p>minor typo: verticies -&gt; vertices here and in a couple other places :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/963/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

StanczakDominik, line 1: <p>Maybe something like:</p><p>"Here, for example, we create a..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/963/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

plasmapy/plasma/grids.py
pheuer: You're absolutely right - consider it done.  I'm going with the option `grid.add_quantities(E_x=E_x_array)` for now.

plasmapy/plasma/grids.py
pheuer: Yes, definitely. Done.

pheuer:
The `grids_examples` notebook still needs some polishing. I'm not sure why the test-documentation error is being raised: the file seems to be where it should be?

StanczakDominik:
Looking into it...

StanczakDominik (approved):
Looks great! I believe @rocco8773 still wanted to take a look, so I'm not merging it just yet; but it does look quite mergeable :)

changelog/963.feature.rst
```diff
@@ -0,0 +1 @@
+Further developed grid objects, improved documentation.
```
rocco8773, line 1: This should be a little more descriptive.  At least mention that you're adding the Jupyter notebooks in this PR.

rocco8773 (changes requested):
`pandas` and `xarray` need to be added to the package dependencies.   This is done my updating files `requirements/requirements.txt` and `requirements/environment.yml`, and `setup.cfg`.  I see `xarray` is already in the `setup.cfg`, but we should add `pandas` to make the dependency explicit (since we do explicitly import `pandas`).

plasmapy/plasma/grids.py
```diff
@@ -734,3 +888,88 @@ def _make_mesh(self, start, stop, num, **kwargs):
         arr0, arr1, arr2 = np.meshgrid(ax0, ax1, ax2, indexing="ij")
 
         return arr0, arr1, arr2
+
+
+def example_grid(name, L=1 * u.cm, num=100):
```
rocco8773, line 893: I'm having a hard time wrapping my head around this function.  It feels more like a "field generator" than an example grid.  And, as an example grid it's opaque since it does not demonstrate how to build a grid.  I'd rather see this function postponed to the fields PR and have a notebook that builds a grid for an electrostatic sphere than one the uses `example_grid()` to get the electrostatic sphere grid and plot it.  That approach would be a lot more informative for the user.  Now, when the field functionality is added down the road we could then update the notebook to compare constructing the grid to the built-in functionality (not absolutely necessary though).

StanczakDominik: Could be good to rename it, say, to example field generator, but if it's there already, achieves what it's supposed to add is going to be helpful later I don't see any value in postponing it. That just creates administrative overhead, I think.

pheuer: I agree with Dominik: I'm fine with renaming this function but it already accomplishes everything I was previously doing with `field.py`, so I think having another type of object is redundant at this point. 

I wouldn't call it `field_generator`, though, because it could just as easily construct grids with more than just fields on them, eg. density profiles for testing interferometry synthetic diagnostics. Maybe `template_grids` or `test_grids`?

changelog/963.feature.rst
pheuer: I've updated the changelog to mention the notebooks. I could add more (akin to what's listed in the text at the top of this PR), but that will end up being a pretty long changelog entry since the changes were both broad and deep in scope.

pheuer:
@rocco8773 I updated the requirements to include pandas (1.1.0) and xarray (0.16.0). Let me know if I did it wrong or if you think a different version is better.

requirements/requirements.txt
```diff
@@ -6,5 +6,7 @@ lmfit >= 1.0.1
 matplotlib >= 2.0
 mpmath >= 1.0.0
 numpy >= 1.18.1
+pandas >= 1.1.0
 scipy >= 1.2
 setuptools >= 41.2
+xarray  >= 0.16.0
```
rocco8773, line 12: @StanczakDominik You could answer this better than me, do we need to have version dependencies on `pandas` and/or `xarray`?  My intuition says to have no version dependency for `pandas` but have one for `xarray`.  Then, `xarray` will always grab the `pandas` version it needs.  For the `xarray` version, I'd grab the one where Python `>= 3.6` was first supported, so `xarray >= 0.14.0`.

docs/examples.rst
```diff
@@ -18,3 +18,4 @@ General-purpose and introductory examples from PlasmaPy.
     :reversed:
 
     notebooks/*
+    notebooks/*/*
```
rocco8773, line 21: @StanczakDominik I think we're starting to get enough examples that we should start breaking these up into sub-sections of notebooks.  What do you think?  (Obviously we'd do this in a separate PR.)

plasmapy/plasma/grids.py
rocco8773: I'm still having a hard time with this.  I don't see when the current form would be used.  I think a more usefully form would be a `populate_grid()` where a user would request a basic shape/type/template (i.e. hard sphere, gaussian sphere, cylinder, etc.) and the variable type (i.e. density, b-field, e-field, etc.).  I'd see this working something like...

```python
grid = grids.CartesianGrid(x, y, z)

# add a hard sphere of density
grid = populate_grid(grid, template="hard sphere", quantity_type="density", quantity_units=u.cm**-3)

# add a hard cylinder of magnetic field along the x-axis
grid = populate_grid(
    grid, 
    template="hard cylinder",
    quantity_type="magnetic field",
    quantity_units=u.T,
    align_on=ax,
)
```

This is just a concept off the top of my head.  I think dealing with 1D,  2D, and 3D data in one function `populate_gird()` might be a bit difficult.

----

With that said, I think your "Example Grids" notebook would be more useful to a user if you, say, went through the process of constructing an electrostatic Gaussian sphere first and then was like, "Ah ah, we do the hard work for you and you can use `example_grids()` instead."

plasmapy/plasma/grids.py
```diff
@@ -76,6 +95,46 @@ def _validate(self):
     # Fundamental properties of the grid
     # *************************************************************************
 
+    def __repr__(self):
```
rocco8773, line 98: Do you plan to have all sub-classes of `AbstractGrid` build and manage data around a `xarray.Dataset`?  If so, I would leverage `xarray.Dataset`'s `__repr__` because most have what you're doing here is already by the dataset's repr.

requirements/requirements.txt
StanczakDominik: No idea, honestly! Pandas 1.0.0 has been out for a while, doesn't sound like a bad idea to have that as a lower bound; but I suspect xarray depends on it pandas, so we might as well go with their minimum dependency. For xarray, no idea.

Whatever we do with, the ultimate condition is that it works; so the minimum version should be added to the minimum- or minimal- or something environment in tox.ini, so that they're tested.

Sorry for lack of precision, just had a very efficient waking up and am a bit unfocused :sweat_smile:

docs/examples.rst
StanczakDominik: As in, in the main notebook list? Yeah, definitely; definitely calling that out of pr scope as well though.

plasmapy/plasma/grids.py
StanczakDominik: Fair point, and i didn't think of that :(

pheuer: So, I tried doing that, but I think the dataset `__repr__` is a lot less informative than the one I created. Here's the `dataset.__repr__` for an example grid:

```
<xarray.Dataset>
Dimensions:               (ax0: 10, ax1: 10, ax2: 10)
Coordinates:
  * ax0                   (ax0) float64 -1.0 -0.7778 -0.5556 ... 0.7778 1.0
  * ax1                   (ax1) float64 -1.0 -0.7778 -0.5556 ... 0.7778 1.0
  * ax2                   (ax2) float64 -1.0 -0.7778 -0.5556 ... 0.7778 1.0
Data variables:
    B_x                   (ax0, ax1, ax2) float64 0.4924 -0.8735 ... -0.1033
    not_a_recognized_key  (ax0, ax1, ax2) float64 -0.04441 -0.2404 ... -0.3236
    B_y                   (ax0, ax1, ax2) float64 0.4924 -0.8735 ... -0.1033
    B_z                   (ax0, ax1, ax2) float64 0.4924 -0.8735 ... -0.1033
Attributes:
    axis_units:  [Unit("cm"), Unit("cm"), Unit("cm")]
``` 

And here's my `__repr__` for the same dataset:

```
*** Grid summary ***
<class 'plasmapy.plasma.grids.AbstractGrid'>
Shape: (10, 10, 10)
Units: [Unit("cm"), Unit("cm"), Unit("cm")]
Uniformly Spaced, dx,dy,dz = (0.222 cm,0.222 cm,0.222 cm)
-----------------------------
Recognized Quantities:
-> B_x (T)
-> B_y (T)
-> B_z (T)
-----------------------------
Unrecognized Quantities:
-> not_a_recognized_key (kg)
```

There's lots of additional information in the second one:
* Quantities separated by whether or not they were recognized, which is important for debugging
* Units for quantities
* The type of grid (in this case, AbstractGrid)
* Whether the grid is uniformly spaced, and the grid spacing.

So I think this printout is doing substantial work above the `dataset` one. Also, implementing this ourselves allows us to add more to it (eg. we could print the min and max of each axis and/or quantity).

plasmapy/plasma/grids.py
pheuer: So the only use-case I had in mind was for example notebooks, and I see your point that manually creating these grids in those notebooks is more instructive. So, I have removed the `example_grids` function and corresponding notebook from this PR. I will create these in the notebook examples for different synthetic diagnostics as they come up. 

I like the idea of a `populate_grid` function, but that sounds like a considerable project. Let's leave it for another PR - we don't need it for the current synthetic diagnostics project.

requirements/requirements.txt
pheuer: That makes sense - to do that I assume I just remove the >= #.#.# from the `setup.cfg` file. Do I delete the pandas line altogether from `requirements.txt` or just delete the version requirement?

plasmapy/plasma/grids.py
rocco8773: I think that's a good route to take, and thank you for putting up with my hardheadedness on this.  Your original concept behind `example_grids` is a very good one and, if we generalize it little more, then I think it will be a very useful tool.

requirements/requirements.txt
rocco8773: > No idea, honestly! Pandas 1.0.0 has been out for a while, doesn't sound like a bad idea to have that as a lower bound; but I suspect xarray depends on it pandas, so we might as well go with their minimum dependency. For xarray, no idea.

`xarray` does have a dependency on `pandas`.  I believe it's the main structure they use for indexing their `DataArray`s and `Dataset`s.

> That makes sense - to do that I assume I just remove the >= #.#.# from the setup.cfg file. Do I delete the pandas line altogether from requirements.txt or just delete the version requirement?

I prefer that we leave `pandas` in the `requirements.txt` and `setup.cfg` so it's explicitly clear that `plasmapy` has a dependency on `pandas` (only because we specifically import `pandas`).  I say we do `xarray >= 0.14.0` and `pandas >= 1.0.0` because they are the first  version to roll their dependencies to `python >= 3.6`.

pheuer: Ok, done (I think)

StanczakDominik (approved):

plasmapy/plasma/grids.py
pheuer: Per our discussion in the last meeting, I've fleshed out this `__repr__` a bit so that it now fully replaces the `xarray.dataset.__repr__`, while still including the additional features I wanted:

```
*** Grid Summary ***
<class 'plasmapy.plasma.grids.AbstractGrid'>
Dimensions: (ax0: 10, ax1: 10, ax2: 10)
Uniformly Spaced: (dax0, dax1, dax2) = (0.222 cm, 0.222 cm, 0.222 cm)
-----------------------------
Coordinates:
	-> ax0 (cm) float64 (10,)
	-> ax1 (cm) float64 (10,)
	-> ax2 (cm) float64 (10,)
-----------------------------
Recognized Quantities:
	-> B_x (T) float64 (10, 10, 10) 
	-> B_y (T) float64 (10, 10, 10) 
	-> B_z (T) float64 (10, 10, 10) 
-----------------------------
Unrecognized Quantities:
	-> not_a_recognized_key (kg) float64 (10, 10, 10) 
```

pheuer:
@rocco8773 I've made the changes we discussed Monday - hopefully this should now be good to go

rocco8773:
Could you generated images for the two notebooks and place them in `/docs/_static/notebook_images/` (you'll have to create the folder too)?  After that, I'll do the work to make them show up in the notebook thumbnails.  They don't need to be anything fancy, I'm thinking of something along the lines of...

**Uniform**
image

**Non-Uniform**
image

... but I'll leave it up to you on what the image should be.

plasmapy/plasma/grids.py
```diff
@@ -96,14 +168,13 @@ def grids(self):
...
+                    self.ds["ax0"].data,
+                    self.ds["ax1"].data,
+                    self.ds["ax2"].data,
                 )
 
         return self._grids
 
-    # Note: may remove this function?
     @property
     def grid(self):
```
rocco8773, line 179: There a couple subtle issues with how this method is currently written.

1. The method is actually returning a copy of the data contained in the `xarray.Dataset`.  This is not a problem in and of itself, but the docstring should be updated to specifically state this.  For example, `"Return a copy of the positional data as an (N, N, N, 3) array."`

   This is how the method currently operates.
   ```python
   >>> np.array_equal(grid.ds["ax0"], grid.grid[..., 0, 0, 0])
   True
   >>> grid.grid[0, 0, 0, 0] = 5.0
   >>> np.array_equal(grid.ds["ax0"], grid.grid[..., 0, 0, 0])
   False
   ```
2. Since the returned variable is only defined when `self._grid == None` or `self._grids == None`, this property will never get updated when the `xarray.Dataset` is modified.  There is some annoying overhead with this, but the array should probably be reconstructed every time the property is called.

changelog/963.feature.rst
```diff
@@ -0,0 +1 @@
+Further developed grid objects and improved their documentation, including several example notebooks.
```
rocco8773, line 1:
```suggestion
Further developed grid objects (`plasmapy.plasma.grids`) and improved their
documentation, including several example notebooks.
```

The `plasmapy.plasma.grids` will provide a link back to the documentation of the module when the change log is compiled.

I'd also like to see a few specifics on how the objects where "further developed."  What was actually done?

rocco8773 (changes requested):

rocco8773 (changes requested):
I can't directly comment on these, but the docstrings for `ax0`, `ax1`, `ax2`, `dax0`, `dax1`, and `dax2` are not rendering nicely when the documentation is built.

**For the `ax#`**
* The current documentation looks like...
image
image

* The docstrings should be replaced with something like...
   ```python
   r"""
   First axis of the grid, only valid for uniform girds.

   Raises
   ------
   ValueError
       If grid is non-uniform.
   """
   ```
   which will render like....
image
image

**For the `dax#`**
* These have a similar documentation issue as the `ax#` and should be replaced with something like (for `dax1`) ...
   ```python
   r"""
   Grid step size along axis :attr:`ax1`, only valid for uniform grids.

   Raises
   ------
   ValueError
       If grid is non-uniform.
   """
   ```
* Note the `` :attr:`ax1` ``, this will create a link back to the `ax1` documentation.
* This renders like...
image
image

plasmapy/plasma/grids.py
```diff
@@ -12,11 +12,28 @@
...
+    "rho": ("Mass density", u.kg / u.m ** 3),
+    "E_x": ("Electric field (x component)", u.V / u.m),
+    "E_y": ("Electric field (y component)", u.V / u.m),
+    "E_z": ("Electric field (z component)", u.V / u.m),
+    "B_x": ("Magnetic field (x component)", u.T),
+    "B_y": ("Magnetic field (y component)", u.T),
+    "B_z": ("Magnetic field (z component)", u.T),
+    "phi": ("Electric Scalar Potential", u.V),
+}
+
```
rocco8773, line 36: These are really specific to the grid objects, so it's probably best to add these to `AbstractGrid` as a class attribute.  Then they would be properly documented.  This would look something like...

```python
class AbstractGrid(ABC):
    """..."""

    # quantities recognized by the grid mechanics
    _recognized_quantities = {...}

    @property
    def recognized_quantities(self):
        """Give a description."""
        return self._recognized_quantities

class PlasmaGrid(AbstractGrid):
    """..."""

    _recognized_quantities = {
        **AbstractGrid._recognized_quantities,
        "n_i": ("ion number density", u.m ** -3),
    }
```

In this setup, `recognized_quantities` will be properly documented for `AbstractGrid` and all its sub-classes.  Each sub-class can override or extend the recognized quantities without breaking any of the documentation.

----

This is completely your preference, but implementing a named tuple could be helpful here.  This would look something like...

```python
from collections import namedtuple

RecognizedQuantites = namedtuple("RecognizedQuantities", ["name", "long_descr", "unit"])

# implement in recognized_keys
recognized_keys = [
    RecognizedQuantities("x", "x spacial position", u.m),
    ...,
    RecognizedQuantites("B_Z", "Magnetic field (z componenet)", u.T),
    ...,
]
```

Then,
```python
>>> recognized_keys[0].name == recognized_keys[0][0]
True
>>> recognized_keys[0].name
'x'
>>> recognized_keys[0].long_descr
'x spacial position'
>>> recognized_keys[0].unit
Unit("m")
```

plasmapy/plasma/grids.py
pheuer: I'm aware of this issue: I was worried that the overhead in reconstructing these arrays would be costly in simulation loops. I'll change it to be the way you suggest for now. If it turns out to be a big performance hit, we can try to find another way?

plasmapy/plasma/grids.py
pheuer: Ok, I'll move this into AbstractGrid for now. 

In the future I think there's a case for having a structure like this that's used broadly throughout PlasmaPy. Certainly it would be bad if different modules used different standard key names. I'll tag @namurphy and @StanczakDominik to also think about this possibility in the future.

pheuer: I've chosen to leave this as a regular dictionary for now because it's useful to be able to quickly access the entries by key. I'm constructing this dictionary automatically from the list format you suggested above to minimize repetition.

pheuer:
@rocco8773 Thanks for all the feedback! I've made the changes you suggested - let me know what you think.

StanczakDominik (approved):
There, I applied the thumbnail adjustment! These changes look good, and in the interest of pushing this along, let's get this a merge and possibly adjust later :)

PR #971, opened 2021-01-02
ExB drift notebook

### LINKED ISSUE #965: Notebook for ExB drift
A Jupyter notebook needs to be created in our documentation that not only shows how to use `plamsapy.formulary.drifts.ExB_drift()` but that also describes the physics behind the drift.  The notebook has to be placed in `/docs/notebooks/` or one of it's sub-directories.  Then the notebook needs to be properly reference in the documentation.

kjmontes:
Closes #965

This pull request adds a new file `docs/notebooks/ExB_drift.ipynb` with a Jupyter notebook explaining the ExB and general F drifts. An analytical derivation of single particle motion in a uniform B-field is included, then followed by a derivation of the drift velocity to explain the physics. To facilitate understanding, I then used the analytical solutions to evolve a proton's trajectory. The case with a uniform B field and no external forces is compared to the case where a perpendicular E field is added to the system. 

In addition to the new notebook, a reference to this example was also added in the file `docs/formulary/drifts.rst`

      docstrings.


StanczakDominik:
Hey, thanks for doing this! I'll take a look at it tomorrow :) Sorry for the end-of-holiday-break delay!

docs/notebooks/ExB_drift.ipynb
```diff
@@ -0,0 +1,225 @@
+{
```
StanczakDominik, line 1: <p>I'm going to have to suggest writing it just as "ExB" - while this looks great in the notebook, this first heading is used in the HTML &lt;title&gt; tag, so if you take a look at the &lt;title&gt; of https://plasmapy--971.org.readthedocs.build/en/971/notebooks/ExB_drift.html, you'll see it breaks a little.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/971/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

StanczakDominik, line 1: <p>For v_y I think I've always seen this done with a charge-signed \omega_c, so then there's no need to add the q / |q| part - but this is nice and lucid! :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/971/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

StanczakDominik, line 1: <p>This sort of duplicates the functionality of https://docs.plasmapy.org/en/latest/notebooks/particle_stepper.html, but as I wrote in the docs - highly unstable code etc etc, and at this point in time I don't think it's comfortable enough to use here rather than this short didactic example. So - I do like this; we could maybe add a note that this is a simplified implementation of https://docs.plasmapy.org/en/latest/notebooks/particle_stepper.html for now; and once I've refactored that one, I'll gyrate back here and see whether it makes sense to replace it :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/971/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

StanczakDominik (changes requested):
Looks great!

* I do have a few minor suggestions in the notebooks (see above/ReviewNB);
* we should add a note on this change to `changelog/971.doc.rst`;
* the CircleCI documentation test notes that the new notebook has warnings; if you go to `docs`, then `make html`, you'll get these:

```.
../PlasmaPy/docs/notebooks/ExB_drift.ipynb:38: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:40: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:42: WARNING: Definition list ends without a blank line; unexpected unindent.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:193: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:195: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:197: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:201: WARNING: Inline interpreted text or phrase reference start-string without end-string.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:203: WARNING: Definition list ends without a blank line; unexpected unindent.
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:41: WARNING: Unknown target name: "v".
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:196: WARNING: Unknown target name: "f".
.../PlasmaPy/docs/notebooks/ExB_drift.ipynb:196: WARNING: Unknown target name: "dv".
```

which I haven't seen before. O @PlasmaPy/mighty-wizards-of-sphinx, does that ring any bells?

docs/notebooks/ExB_drift.ipynb
kjmontes: <p>Thanks, I missed that. I'll change it in the next commit. </p><p> </p>

kjmontes: <p>Yes, I figured that keeping \omega_c unsigned would make it consistent with the <code>formulary.parameters.gyrofrequency()</code> function in PlasmaPy. Feel free to modify any of the notation if you think it's appropriate. </p>

kjmontes: <p>Thanks for pointing this out. My understanding is that the particle tracker class uses the Boris algorithm to evolve a particle's position in prescribed E and B fields. Therefore, one could solve the problem using that class (as is done in the particle stepper notebook) but it would not require solving for the drift velocity analytically. Both implementations should yield the same result for this problem, but the Boris algorithm is much more widely applicable because it doesn't require an analytical solution for the velocity. I added a note explaining this at the end of the notebook and pointing users to the particle stepper notebook for comparison.  </p>

kjmontes:
Thanks @StanczakDominik for your suggestions! I replied to each and made a few commits to address them. I also do not know what is causing the warnings in the CircleCI documentation test. Let me know if you think of anything else.

StanczakDominik:
I think I fixed it in 0058862 ! Given tests are passing now, that's a merge. Thanks for doing this, @kjmontes!

@rocco8773, do you have any idea why Sphinx would complain about `\begin{equation}`-style latex in notebooks rather than `$$equation goes here`-style latex?

rocco8773:
@StanczakDominik According to `nbsphinx` using `\begin{equaiton}` should work.  In fact, the current notebook builds just fine when I do a local `sphinx` build, but the [RTD hosted version is not building correctly](https://docs.plasmapy.org/en/latest/notebooks/ExB_drift.html).  ... After a little investigation, my local build is using `sphinx == 2.4.4` and the RTD build is using `v1.8.5`.  I'm guessing this is why the math stuff is not generating correctly online.

It looks like `sphinx-automodapi` came out with a new release a few months ago, so we should update our dependencies, `sphinx-automodapi >= 0.13` and `sphinx >= 2.4.4`.  I tested the build out on my local machine and it seems to be working.

StanczakDominik:
Gah, good catch... I'll fix that ASAP.

PR #978, opened 2021-01-15
Add nuclear cross sections, rates: start to issue #110

cfe316:
This is a "minimum viable product" for cross section and rate coefficient data on the four nuclear reactions involving, D, T, and 3He for which formulas are given in Bosch & Hale's paper. There's certainly work to be done in thinking about the overall interface, abstracting the reaction parser, extending the set, ensuring the right exceptions are thrown. I also need to learn more of the automated-documentation-and-testing frameworks that require things like `def methodname() -> u.Quantity:` to ensure that everything hooks up in the correct manner.

At the moment this allows a user to:
```
import astropy.units as u
from plasmapy.particles.nuclear import FusionReaction

dt = FusionReaction("D + T -> 4He + n")
dt.cross_section(10 * u.keV)
(it returns the cross section in u.barns)
dt.rate_coefficient10 * u.keV)
(it returns the rate coefficient in u.m**3 / u.s)
```

      docstrings.

### DIFF
--- a/plasmapy/particles/nuclear.py
+++ b/plasmapy/particles/nuclear.py
@@ -1,6 +1,8 @@
 """Functions that are related to nuclear reactions."""
 __all__ = ["nuclear_binding_energy", "nuclear_reaction_energy", "mass_energy"]
 
+import numpy as np
+
 import astropy.units as u
 import re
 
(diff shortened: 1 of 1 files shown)


plasmapy/particles/nuclear.py
```diff
@@ -11,6 +13,263 @@
...
+        """
+        self.Bg = Bg
+        self.mrcsq = mrcsq
+        self.C = C
+        self.energy_range = energy_range
+        self.error_max = error_max
+
+    def sigma_v(self, temperature):
+        # TODO return warning if out of energy range?
+        T = temperature.to(u.keV).value
```
namurphy, line 259: We'll have to be careful here since keV are units of energy rather than temperature, which makes it more complicated to use `astropy.units`.  It might be necessary to use the `u.temperature_energy(https://docs.astropy.org/en/stable/units/equivalencies.html?highlight=equivalencies#temperature-energy-equivalency)` equivalency here, though this might also be able to be taken care of with `@plasmapy.utils.validate_quantities`.

namurphy (changes requested):
Thank you for working on this!  #110 has been a longstanding issue so I'm really grateful that you're working this.  

I have a first round of suggestions, including some which deal with conventions for PlasmaPy.  I have a few comments for the future too, which aren't immediately relevant here.  This pull request will need some tests too.

Thank you again!

PR #981, opened 2021-01-22
Grids (part 3)

pheuer:
I'm getting really close to a reviewable proton radiography PR, but I ended up needing some new features (and some other minor improvements) to grids. A few of these are breaking changes, I'll try to mark them below:

**Major**

- Introduced 'persistent interpolators' which assume the grid has not changed between calls (as long as a keyword flag is set). Otherwise, ~25% of the interpolation cycle was being wasted constantly rebuilding some arrays. I tried to make this as elegant as possible, but maybe it can be done better? Something like this has to be done, though.

- Introduced the `grid.grid_resolution` method which determines a single scalar length scale from the grid spacing. I'm using this to determine a simulation timestep. This could be defined in the proton radiography module, but it's generally applicable so I put it here.

- Introduced `grid.on_grid`: a method which determines which of a list of points lies within the region bounded by the grid. Useful for the proton radiography code, likely also elsewhere so I put it here.

- Introduced `grid.vector_intersects` which determines whether a line between two points intersects the grid using a ray-box intersection algorithm. This is broadly useful.

**Minor**
- `grid.grid` and `grid.grids` are now dimensioned. Subsequently, `grid.grids` is only defined for grids with the same units on all three dims. `grid.pts#` are still dimensioned, their API hasn't changed. (BREAKING) 

- `grid.is_uniform_grid` is now just `grid.is_uniform` because the old name was redundant. (BREAKING)

- Added a `grid.quantities` property that returns a list of all the quantities defined on the grid. 

- Other minor speed improvements to interpolators, none of which are breaking. For example, interpolation is now done without units and I was able to collapse some for loops into array ops.

Everything should have associated tests added

Here's an example of the speed enhancement achieved with the `persistent` keyword for 6 quantities at 5e5 positions on a (200x200x200) grid (these are times per cycle, so the extra 300 ms adds up fast).
```
Non-Persistent Volume-Averaged Interpolation (avg over 30)
--- 1.2901666641235352 seconds ---
Persistent Volume-Averaged Interpolation (avg over 30)
--- 1.014957102139791 seconds ---
```


StanczakDominik:
I took the liberty of replacing your implementation of grid_resolution for the non-uniform case with a few lines of scipy code :) I also added a very basic test for it, but it'd be great if you could double-check if it still does the same thing - or add more detailed tests! ;)

plasmapy/plasma/grids.py
```diff
@@ -601,18 +629,100 @@ def _make_mesh(self, start, stop, num, **kwargs):
...
+        on_grid = (
+            np.greater(ax0_min, pos[:, 0]).astype(np.int8)
+            + np.less(ax0_max, pos[:, 0]).astype(np.int8)
+            + np.greater(ax1_min, pos[:, 1]).astype(np.int8)
+            + np.less(ax1_max, pos[:, 1]).astype(np.int8)
+            + np.greater(ax2_min, pos[:, 2]).astype(np.int8)
+            + np.less(ax2_max, pos[:, 2]).astype(np.int8)
+        )
+
+        return np.where(on_grid == 0, True, False)
```
StanczakDominik, line 677: Okay, I'm trying and failing to understand this implementation. 

Am I right in thinking that this method only actually returns indices; not a (n,3) array of float positions but a (n,) array of booleans?

plasmapy/plasma/grids.py
```diff
@@ -601,18 +629,100 @@ def _make_mesh(self, start, stop, num, **kwargs):
...
     # *************************************************************************
     # Interpolators
     # *************************************************************************
 
+    # These arrays are used to enable persistent interpolation (faster repeated
+    # calls with the same grid and arguments)
+
+    _interp_args = []
+    _interp_quantities = None
+    _interp_units = None
```
StanczakDominik, line 716: I wonder if we could use https://docs.python.org/3/library/functools.html#functools.cached_property to simplify this part? I've been using it recently and it simplifies such code by a ton. It is 3.8 only, but there is a backport https://pypi.org/project/cached-property/ that I would be more than happy to add to our requirements until we're 3.8 only!

StanczakDominik (commented):
Partial review for now, I haven't dug into the persistence of interpolators yet. A couple questions! :)

plasmapy/plasma/grids.py
pheuer: Yes, right now it takes an array of particle positions [N,3] and returns an array of Booleans [N] which indicate whether the particles are on or off the grid. 

Really I'm just interested in the total number on the grid, so if you think it would be otherwise more useful we could try to return a list of the indices of particles that are on the grid (then I can just use the length of that array)? Maybe then we could only interpolate and push those particles?

plasmapy/plasma/grids.py
pheuer: I will look into this!! I can always count on you knowing a more efficient way of doing things :)

pheuer:
@StanczakDominik Thanks for the improved `grid_resolution`, that looks great! Effectively the same as what I was doing, just much more efficient.

pheuer:
@StanczakDominik I was able to re-implement persistent interpolation with cached properties: the code is definitely cleaner and, based on a speed test, maybe even a hair faster! 

I tried to add `cached-properties >= 1.5.2` to the dependencies: let me know if I put it all the right places (`setup.cfg` and `requirements.txt`)

StanczakDominik:
Yay, I finally broke tests :smirk_cat: 

The codecov failure displays the tests lines that are not covered by tests in this code. It was supposed to pop up earlier, but I sort of missed that it wasn't working right for the last two months...

Anyway, these are minor and let's fix them later on (#895 ?), so this doesn't get stalled further. On to the final review I go!

plasmapy/plasma/grids.py
StanczakDominik: I went with the less intrusive change and simplified the [N] array output version a little. You should, however, be able to use that array to interpolate/push only those particles, though!

StanczakDominik (approved):
Okay, let's get this done! :) (and hopefully go back to the minor coverage leaks in #895 :) )

pheuer:
Yay, Codecov is back!! Writing tests just isn't much fun without it...

StanczakDominik:
Yeah... sorry for that!

PR #984, opened 2021-01-25
Deprecate `Particle.particle` in favor of `Particle.symbol`

namurphy:
One naming decision I've regretted is for `Particle.particle`.  I would have preferred the name to be `Particle.symbol`, since `particle` does not really represent what the particle is.  I've deprecated `Particle.particle` in favor of `Particle.symbol` for when we release 0.6.0.  We should remove `Particle.particle` in a subsequent release of PlasmaPy, probably 0.8.0 in order to give a bit of time.  Once this is merged, we'll need to create an issue to remove this and add it to the appropriate milestone.


namurphy:
I also added `symbol` as a property of `AbstractParticle` that then gets inherited by `CustomParticle` and `DimensionlessParticle`.  It defaults to what comes from `repr`.

changelog/984.feature.rst
```diff
@@ -0,0 +1,3 @@
+Add a ``symbol`` attribute to `~plasmapy.particles.particle_class.AbstractParticle`
+and therefore `~plasmapy.particles.particle_class.CustomParticle` and
+`~plasmapy.particles.particle_class.DimensionlessParticle`.
```
rocco8773, line 3: Since the `breaking` change log should be removed (we're not breaking yet), update this change log to include info about the `Particle.particle` & `Particle.symbol` updates.

If you want to mention the `FutureWarning` for the deprecation, then the entry might be best suited as a `trivial` change log entry.  Maybe a `removal` change log entry, but I still think that is more suitable for when the feature is actually deleted.

rocco8773 (changes requested):
Looking good! I like these types of reviews.  Nice and straight forward. 🤣

A few things:

1. There's one missed obscure reference to `Particle.particle` in `/docs/particles/decorators.rst` on line 65.
2. On line 421 in `particle_class.py`, to be consistent with this new naming scheming we should refactor `particle` -> `symbol` or `particle_symbol`.
3. On line 425, the `'particle'` key in the `attributes` (`self._attributes`) dict should be refactored to `'symbol'`. Again, being consistent with new nomenclature.

StanczakDominik:
Just to clear this up - this is about the renaming of the `particle` *field* of the `Particle` *object* to be called `symbol`, which sounds very sane and I wholeheartedly agree!

changelog/984.feature.rst
namurphy: This all makes sense.  I wasn't quite sure how to best organize the change logs.

plasmapy/particles/particle_class.py
```diff
@@ -678,20 +683,42 @@ def json_dict(self) -> dict:
...
+        warnings.warn(
+            "Particle.particle has been deprecated and will be removed in "
+            "a subsequent release. Use Particle.symbol instead."
+        )
+        return self.symbol
+
+    @property
+    def symbol(self) -> Optional[str]:
+        """
+        Return the particle's symbol.
```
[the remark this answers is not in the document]
namurphy: Hm... "molecular" typically refers to a particle containing two or more atoms, so it's not quite the word I'm looking for with the current version of `Particle`.  With that said, it is a good point that this could use a bit of clarification, so I'll think about how to reword it.  Maybe `Return the symbol of the particle, atom, isotope, or ion.`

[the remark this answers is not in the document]
rocco8773: > Maybe `Return the symbol of the particle, atom, isotope, or ion.`

I like that suggestion over the original.  Another possible idea, `Return a symbolic representation of the particle (e.g. the atomic symbol).`

changelog/984.feature.rst
namurphy: The [towncrier docs](https://towncrier.readthedocs.io/en/actual-freaking-docs/quickstart.html#creating-news-fragments) say that `removal` signifies "a deprecation or removal of public API" so `removal` does seem most appropriate.

rocco8773 (approved):
LGTM!

Issue #980, opened 2021-01-19
Drop Python 3.6 support for version 0.6.0


namurphy:
[NumPy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) suggests that scientific Python projects support versions of Python released in the last 42 months.  Python 3.6 was released more than 42 months ago, so this suggests that we no longer need to support Python 3.6.  Astropy 4.2 also requires Python 3.7+, so we'll need to go with a more recent version of Python anyway.  There are a few [new features in Python 3.7](https://docs.python.org/3/whatsnew/3.7.html) that look relevant, including dataclasses.  I propose we drop Python 3.6 support with PlasmaPy version 0.6.0.  I think this would require:

- [ ] Updating the requirements declarations
- [ ] Updating our test matrix
- [ ] Updating our documentation where it mentions
- [ ] Updating our website
- [ ] Updating the main `__init__.py` (I think)
- [ ] Discussing this in the changelog

...though it's probable I'm missing a few things.

(NEP 29 also suggests that projects support versions of NumPy from the last 24 months, which would mean that we should support NumPy 1.17+.  However, we've been supporting NumPy 1.18.1+ since #831, since there were apparently issues with units being dropped for some operations in older versions.  This seems to me like a good enough reason to support only more recent versions of NumPy.)

StanczakDominik:
Let's!

PR #985, opened 2021-01-26
Describe valid representations for particles in docstring for `ParticleLike` typing alias

namurphy:
This pull request is to expand the docstring for `particle_like`, which is a typing construct to represent what valid types are for a particle.  The docstring is intended to describe in more detail about what the valid types or particle representations there are.  

Both `CustomParticle` and `DimensionlessParticle` both require a bit more thought, and may need to be handled separately (e.g., have the type in a formulary docstring say `particle_like or CustomParticle` rather than wrap `CustomParticle` inside of `particle_like`).  Right now I'm including `CustomParticle` instances as `particle_like` with the intention that they can be used in `plasmapy.formulary` and elsewhere...but I'm still wondering if this is the right thing to do.  I'm excluding `DimensionlessParticle` because it does not contain physical units (and thus cannot be used in the formulary).  

I added `is_particle_like` to return a `bool` about whether or not an object can uniquely represent a particle.  I'm not sure how useful it would be, so I'd be okay with dropping it if needed.

@rocco8773 — this probably complements what you might have been working on separately, so I'd be interested to chat with you about this.


namurphy:
Also, I did a `from __future__ import annotations` thing at the top, which is a Python 3.7+ feature only.  I think delayed evaluation of annotations is needed if an argument in `Particle` has an annotation of `particle_like`, and `particle_like` depends on `Particle`.

namurphy:
Also, can we require a minimum Python version of 3.10a4?  😹  Instead of having to do an annotation of 
```Python
typing.Union[str, int, float]
```
we could do
```Python
str | int | float
```
And at that time, delayed evaluation of annotations becomes default! 😬

StanczakDominik:
> Also, can we require a minimum Python version of [3.10a4](https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator)?  Instead of having to do an annotation of
> 
> ```python
> typing.Union[str, int, float]
> ```
> 
> we could do
> 
> ```python
> str | int | float
> ```
> 
> And at that time, delayed evaluation of annotations becomes default!

:drooling_face: do want! Probably a little early, though :(

namurphy:
I'm going to undo the change for which I needed the `__future__` import so that this does not need to wait for #980.  That change can wait until later.

plasmapy/particles/particle_class.py
```diff
@@ -2036,7 +2036,111 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
...
+by ``"e"``, ``"e-"``, or ``"electron"``.
+
+Instances of the `~plasmapy.particles.particle_class.CustomParticle`
+class are considered particle-like because the mass and charge of the
+custom particle are uniquely defined in physical units.
+
+`~plasmapy.particles.particle_class.DimensionlessParticle` instances are
+not considered particle-like because (without normalization information)
+they do not *uniquely* identify a particle.
+"""
```
namurphy, line 2103: Since we'll be using `particle_like` as a function annotation in, for example, the formulary...we don't want `DimensionlessParticle` to be included in this.

plasmapy/particles/particle_class.py
```diff
@@ -2036,7 +2036,111 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
...
+
+A **special particle** may be represented by a string that contains a
+the name of the particle or a standard symbol for it.  A neutron can be
+represented as ``"n"`` or ``"neutron"``, a proton can be represented as
+``"p"``, ``"p+"``, or ``"proton"``, and an electron can be represented
+by ``"e"``, ``"e-"``, or ``"electron"``.
+
+Instances of the `~plasmapy.particles.particle_class.CustomParticle`
+class are considered particle-like because the mass and charge of the
+custom particle are uniquely defined in physical units.
```
namurphy, line 2098: I'm still wondering how to best handle `CustomParticle`.  My current plan is to include `CustomParticle` in the definition of `particle_like` so that the annotations to formulary functions can simply be `particle_like` instead of `Union[particle_like, CustomParticle]`.  If we do that, then we should make sure to have the type description for the first input to `Particle` be: "`particle_like`, excluding `CustomParticle` instances".

plasmapy/particles/particle_class.py
```diff
@@ -2036,7 +2036,111 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
...
+`~plasmapy.particles.particle_class.DimensionlessParticle` instances are
+not considered particle-like because (without normalization information)
+they do not *uniquely* identify a particle.
+"""
+
+
+def is_particle_like(obj: Any) -> bool:
+    """
+    Return `True` if an `object` can uniquely represent a physical
+    particle, and `False` otherwise.
```
namurphy, line 2109: I'm still wondering if we want to have `is_particle_like` in here.  I'm tentatively including it since there are times when code like this could end up making things more readable, but I'm open to deleting it.

It would probably be cleaner to make `particle_like` callable (perhaps by setting `particle_like.__call__` to `is_particle_like`), but I ran into some weirdness when trying that.

namurphy:
During the telecon today we decided to defer the decision about whether or not to include `DimensionlessParticle` in `particle_like`.    I tentatively didn't include it because it does not uniquely identify a particle since it does not include information on the normalizations, and because `DimensionlessParticle` instances are not intended for use in `plasmapy.formulary`. 

I'm also going to go back in and do the Python 3.7+ stuff like deferred evaluation of annotations.

changelog/985.doc.rst
```diff
@@ -0,0 +1,2 @@
+Describe what constitutes a valid representation of a particle in the docstring
+for the `plasmapy.particles.particle_class.particle_like` typing construct.
```
namurphy, line 2: Add `is_particle_like`

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2023,118 @@ def mass(self, m: u.kg):
...
+Particle("Fe-56")
+
+An **ion** or **ionic level** is represented as a string that contains a
+representation of an element or isotope, followed by a space, and then
+the charge information (e.g., ``"2+"``).  The charge information in the
+string includes an integer either followed or preceded by a plus or
+minus sign.  For example, Fe:sup:`2+` can be represented as
+``"Fe 2+"``, while :sup:`56`Fe:sup:`0+` can be represented as
+``"iron-56 +0"``.  The integer charge can alternatively be specified
+using the ``Z`` keyword to `~plasmapy.particles.particle_class.Particle`.
```
namurphy, line 2076: Mention Roman notation

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2023,119 @@ def mass(self, m: u.kg):
...
-# TODO: Describe valid particle representations in docstring of particle_like
+# TODO: Need to incorporate particle_like into @particle_input
+
+# TODO: Decide if DimensionlessParticle should be in particle_like
+
+particle_like = Union[str, Integral, Particle, CustomParticle]
+
+particle_like.__doc__ = """
+An `object` is particle-like if it can uniquely identify an ion, atom,
+isotope, special particle, or custom particle.
```
rocco8773, line 2034:
```suggestion
An `object` is particle-like if it can be identified as an instance of 
`~plasmapy.particles.particle_class.Particle` or 
`~plasmapy.particles.particle_class.CustumParticle`, or cast into one.
```

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2023,119 @@ def mass(self, m: u.kg):
...
+
+Notes
+-----
+Particles in PlasmaPy are canonically represented as instances of the
+`~plasmapy.particles.particle_class.Particle` class.
```
rocco8773, line 2039:
```suggestion
Real world particles are typically represented as instances of the
`~plasmapy.particles.particle_class.Particle` class in PlasmaPy.
Custom particles can be constructed with the
`~plasmapy.particles.particle_class.CustomParticle` class
```

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2023,119 @@ def mass(self, m: u.kg):
...
+>>> from plasmapy.particles import Particle
+>>> Particle("proton")
+Particle("p+")
+>>> Particle("neutron")
+Particle("n")
+
+All `~plasmapy.particles.particle_class.Particle` instances are
+particle-like.  Any `object` that can be passed as a sole argument to
+the `~plasmapy.particles.particle_class.Particle` class is considered
+particle-like.
```
rocco8773, line 2050:
```suggestion
All `~plasmapy.particles.particle_class.Particle` instances and variables
that can be cast into a `~plasmapy.particles.particle_class.Particle` 
instances are particle-like.
```

namurphy:
A brainstorm of things to do and/or think about. 

 - [ ] Define the docstring by using `particle_like.__doc__ = """` to fix issue with the docstring for `typing.Union` stuff appearing
 - [ ] Repace separate `is_particle_like` function with doing things like `isinstance(particle, particle_like)`.  We would need to use the [`__instancecheck__` dunder](https://docs.python.org/3/reference/datamodel.html?highlight=instance%20check#class.__instancecheck__) 
 - [ ] Fix lingering ReST error
 - [ ] Figure out if it should be `particle_like` or `ParticleLike`, where the latter would indicate a class.  I'm using lower case because when I do `inspect.isclass(particle_like)` it returns `False`, and I can't get `particle_like` to be instantiated.  But if we use `isinstance(particle, ParticleLike)` then it would be treating it like a class.
 - [ ] Think about using `particle_like.__call__` instead of `is_particle_like`...though I had trouble setting this.

changelog/985.doc.rst
namurphy: Nevermind, since we're removing `is_particle_like`.

namurphy:
So, @rocco8773 and I decided that it would be good to set it up so that we can do `isinstance(something, particle_like)`, rather than have a separate function for `is_particle_like`.  I _think_ to do that, we might need to create a new or custom type rather than using `particle_like = typing.Union[blah]`, since I think `__instancecheck__` needs to be defined on the class rather than an instance of the class.  I'm wondering about the utility of saving that for a follow-up pull request, since my original purpose for this one is the docstring.

rocco8773:
> So, @rocco8773 and I decided that it would be good to set it up so that we can do `isinstance(something, particle_like)`, rather than have a separate function for `is_particle_like`. I _think_ to do that, we might need to create a new or custom type rather than using `particle_like = typing.Union[blah]`, since I think `__instancecheck__` needs to be defined on the class rather than an instance of the class. I'm wondering about the utility of saving that for a follow-up pull request, since my original purpose for this one is the docstring.

Yeah, I'm a big fan of this idea, but Python does not seem to share the sentiment.  `particle_like = typing.Uniion[...]` is an instance of [`typing._GenericAlias`](https://github.com/python/cpython/blob/0869a713f21f4b2fe021d802cf18f1b1af53695f/Lib/typing.py#L650) and [these lines (for py38)](https://github.com/python/cpython/blob/0869a713f21f4b2fe021d802cf18f1b1af53695f/Lib/typing.py#L768-L778) give us fits when using `isinstance()` or `issubclass()`.

namurphy:
Rather than leaving `TODO` comments in there, here's a checklist of things to do after.  Probably would be good to raise as issues.
 - [ ] Need to incorporate `particle_like` into `@particle_input`
 - [ ] Decide if `DimensionlessParticle` should be in `particle_like`
 - [ ] Make `particle_like` a new type using functionality from typing. If it becomes a class when doing this, change to `ParticleLike`.
 - [ ] Enable doing `isinstance(x, particle_like)` with `__instancecheck__`

namurphy:
With respect to the issue about whether or not `DimensionlessParticle` should be in `particle_like`, one possibility would be to rename `particle_like` to `physical_particle`.  That would make clear that `DimensionlessParticle` would not be included.

rocco8773:
With regards to the different types of particles, maybe we can take the approach of your much loved `numbers` pkg and have something w/ `.Particle`, `.Real`, `.Sim`, etc. I think the most difficult aspect of that approach will be the naming scheme.

namurphy:
> With regards to the different types of particles, maybe we can take the approach of your much loved `numbers` pkg and have something w/ `.Particle`, `.Real`, `.Sim`, etc. I think the most difficult aspect of that approach will be the naming scheme.

I think it'll depend on how complicated it ends up being.  Most of the use cases I can think of could either be done with an annotation of `particle_like` (pretty much everything in the formulary) or `DimensionlessParticle` (probably in `plasmapy.simulation`).  

In addition to that, if a argument to a function is named `ion` and annotated with (currently) `Particle`, `@particle_input` will enforce that the particle end up being an ion...and there's also the option of specifying which categories the particle has to fall in.

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2022,88 @@ def mass(self, m: u.kg):
...
+
+>>> Particle("He-4", Z=2)
+Particle("He-4 2+")
+
+A **special particle** may be represented by a string that contains
+the name of the particle (case-insensitive) or a standard symbol for it
+(case-sensitive).  A neutron can be represented as ``"n"`` or
+``"neutron"``; a proton can be represented as ``"p+"``, ``"p"``, or
+``"Proton"``; and an electron can be represented by ``"e-"``, ``"e"``,
+or ``"ELECTRON"``.
```
rocco8773, line 2087: [For future refactoring]  I would say the docstring of `particle_like` shouldn't duplicate, but reference, documentation is other areas.   This  block seems like it should be in the main `Particle` documentation and not here.  However, after looking at the main particle documentation, the documentation does not seem to cleanly cover what is said here.  At some point what is said here should be migrated to the main documentation.

rocco8773 (changes requested):
The bottom of the RTD page has this weird "extra" documentation that you have not specifically written here.  Can you look into why this is happening.

image

----

I feel the most important aspect of the `particle_like` documentation would be describing what is considered particle like, and not so much how to use `Particle`.  I think the notes section would be better served by directly stating what particles are particle like.  For example...

``` restructuredtext
Notes
-----

PlasmaPy offers several ways of creating represenations for real world particles, synthetic paritcles, custom particles, etc.  The following is what we consider as (or not as) `particle_like`:

* All `~plasmapy.particles.particle_class.Particle` instances and variables
that can be cast into a `~plasmapy.particles.particle_class.Particle`
instances are particle-like.
* Instances of the `~plasmapy.particles.particle_class.CustomParticle`
class are considered particle-like because the mass and charge of the
custom particle are provided in physical units.
* `~plasmapy.particles.particle_class.DimensionlessParticle` instances are
**NOT** considered particle-like because, without normalization information,
they do not *uniquely* identify a physical particle and therefore cannot
be used in `plasmapy.formulary`.
```

plasmapy/particles/particle_class.py
```diff
@@ -2023,7 +2022,88 @@ def mass(self, m: u.kg):
...
-# TODO: Describe valid particle representations in docstring of particle_like
-
-particle_like = Union[str, Integral, Particle]
-"""A typing construct for valid representations of a particle."""
+particle_like = Union[str, Integral, Particle, CustomParticle]
+
+particle_like.__doc__ = """
+An `object` is particle-like if it can be identified as an instance of
+`~plasmapy.particles.particle_class.Particle` or
+`~plasmapy.particles.particle_class.CustomParticle`, or cast into one.
```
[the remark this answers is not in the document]
namurphy: At the moment, `particle_like` is not an instance of `GenericAlias`...though I suspect that is the way we should go about doing this in the future.  Good idea to point out that it's for typing annotations.

namurphy:
> The bottom of the RTD page has this weird "extra" documentation that you have not specifically written here. Can you look into why this is happening.

Weird!  I will do my best to try to figure it out, but I don't know why it is like that.  😕 

> I feel the most important aspect of the `particle_like` documentation would be describing what is considered particle like, and not so much how to use `Particle`. I think the notes section would be better served by directly stating what particles are particle like. 

That's a good point.  I wonder if it would suffice to link to the narrative documentation for `plasmapy.particles`.  Hm...

plasmapy/particles/particle_class.py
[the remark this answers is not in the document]
rocco8773: I'll admit that I don't fully understand the mechanics/lexicon behind python's `typing` and `types` pkgs, but this...

image

suggests to me that `particle_like` is a Generic Alias.

rocco8773:
> > I feel the most important aspect of the `particle_like` documentation would be describing what is considered particle like, and not so much how to use `Particle`. I think the notes section would be better served by directly stating what particles are particle like.
> 
> That's a good point. I wonder if it would suffice to link to the narrative documentation for `plasmapy.particles`. Hm...

I think that would be the best approach; specify what it is to be `particle_like` and then link to the narrative documentation for further details on working with particles.

plasmapy/particles/particle_class.py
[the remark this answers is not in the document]
rocco8773: But I really don't care if `GenericAlias` is mentioned at all.  I feel most people will probably not know what that means anyways.  ...I mean, I've read about it and I still don't fully understand it... As you said above, the main point is to say somewhere in the documentation that this variable is intended for typing annotations.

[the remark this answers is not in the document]
namurphy: Yeah...the details of this are not particularly clear...
```Python
>>> typing_alias = typing.Union[int, set]
>>> isinstance(typing_alias, typing.GenericAlias)
False
```
> As you said above, the main point is to say somewhere in the documentation that this variable is intended for typing annotations.

Agreed completely!  In part because that reduces the need for me to actually understand things...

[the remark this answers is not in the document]
rocco8773: Honestly, reading through the `typing` document I get the feeling that the pkg is going through some kind change. Some of the document feels a little out of date and some feels like it's been recently modified.  Look at the Generic Alias section, the link for `GenericAlias` (the class) is nonexistent or broken.

namurphy:
> The bottom of the RTD page has this weird "extra" documentation that you have not specifically written here. Can you look into why this is happening.

Um...I have a feeling that it might be easier to figure out the `typing.NewType` stuff than to figure out why the documentation is doing that.  I think I'll try that route and see if it works, since we were going to do that anyway.

namurphy:
Also, I renamed `particle_like` → `ParticleLike` to reflect [advice in PEP 484](https://www.python.org/dev/peps/pep-0484/#type-aliases).

namurphy:
It's so good to have Codecov back!  😹

namurphy:
This is tricky to write in a way that's both concise and complete!  Especially since I realized I forgot to include some representations that I had previously forgotten about...

I made pseudo-sections for the different categories of particles.  That hopefully makes it easier for people to scan through.  Ideally, I'd like people to be able to get enough from the first sentence of these pseudo-sections that they get the gist of it.

I put the thing about `DimensionlessParticle` as a Sphinx directive, in part because that separates it from what `ParticleLike` _does_ include, and in part because it's fun to use Sphinx directives.  It also seems to fit best as a separate note.  I also figured it'd be worthwhile to have a `See Also` section.

I added an example function at the end that uses `ParticleLike` as a type hint annotation.  I was thinking this would illustrate how it's used for people who are unfamiliar with type hints. 

> The bottom of the RTD page has this weird "extra" documentation that you have not specifically written here. Can you look into why this is happening.
> 
> image

It turns out that this is something Sphinx automatically does for aliases.  There might be some way to override it, but it could potentially be helpful to some people, and it's at the end so it's not too distracting.  I'm okay with leaving it in, at least for the time being.

rocco8773:
Random post without context...

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases

changelog/985.doc.rst
rocco8773, line 2:
```suggestion
Describe what constitutes a valid representation of a particle in the docstring
for the `plasmapy.particles.particle_class.ParticleLike` typing construct.
```

plasmapy/particles/ionization_state.py
```diff
@@ -161,7 +161,7 @@ class IonizationState:
 
     Parameters
     ----------
-    particle: `particle_like`
+    particle: `ParticleLike`
```
rocco8773, line 164:
```suggestion
    particle: `~plasmapy.particles.particle_class.ParticleLike`
```

Otherwise the `intersphinx` linking does not work.

plasmapy/particles/ionization_state_collection.py
```diff
@@ -39,12 +39,12 @@ class IonizationStateCollection:
         instances with units of number density.
 
     abundances: `dict`, optional, keyword-only
-        A `dict` with `particle_like` elements or isotopes as keys and
+        A `dict` with `ParticleLike` elements or isotopes as keys and
         the corresponding relative abundance as values.  The values must
         be positive real numbers.
```
rocco8773, line 44:
```suggestion
        A `dict` with `~plasmapy.particles.particle_class.ParticleLike`
        objects used as the keys and the corresponding relative abundance as the
        values.  The values must be positive real numbers.
```

Fix the `intersphinx` link and re-word a little.  The original wording was really confusing to me so I took a stab clarifying it.  Feel free to do just the `intersphinx` correction if you prefer.

plasmapy/particles/ionization_state_collection.py
```diff
@@ -39,12 +39,12 @@ class IonizationStateCollection:
...
 
     log_abundances: `dict`, optional, keyword-only
-        A `dict` with `particle_like` elements or isotopes as keys and
+        A `dict` with `ParticleLike` elements or isotopes as keys and
         the corresponding base 10 logarithms of their relative
         abundances as values.  The values must be real numbers.
```
rocco8773, line 49:
```suggestion
        A `dict` with `~plasmapy.particles.particle_class.ParticleLike`
        objects used as the keys and the corresponding base 10 logarithms of their
        relative abundances as the values.  The values must be real numbers.
```

Fix the `intersphinx` link and re-word a little.  The original wording was really confusing to me so I took a stab clarifying it.  Feel free to do just the `intersphinx` correction if you prefer.

plasmapy/particles/particle_class.py
```diff
@@ -18,7 +20,7 @@
 from collections import defaultdict, namedtuple
 from datetime import datetime
 from numbers import Integral, Real
-from typing import List, Optional, Set, Tuple, Union
+from typing import List, Optional, Set, Tuple, TypeVar, Union
```
rocco8773, line 23:
```suggestion
from typing import List, Optional, Set, Tuple, Union
```

rocco8773 (changes requested):
Looks good to me.  Just a few minor fixes related mostly to `intersphinx`.

[optional]  In this rare case, I think the "Examples" section of the docstring should directly follow header of the docstring.  Why?  The main purpose of `ParticleLike` is to be a typing annotation, and with a long "Notes" section the user has to scroll down to find the example.  The "Example" section is key to understanding what `ParticleLike` is used for, while the "Note" section is extra info on what is a particle.  I just feel more is communicated withe the header and "Example" section paired together.

rocco8773 (approved):

PR #986, opened 2021-01-26
Delete duplicate attribute in Particle class

namurphy:
The `nuclide_mass` attribute was in `Particle` twice.  Now it is in there once.  And the universe is now safe once again.  But...for how long?


StanczakDominik (approved):
Little to be said about this one, LGTM! :laughing:

PR #987, opened 2021-01-26
Drop python 3.6

### LINKED ISSUE #980: Drop Python 3.6 support for version 0.6.0

StanczakDominik:
Closes #980

      docstrings.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -18,15 +18,10 @@ jobs:
       matrix:
         include:
 
-          - name: Python 3.6 with minimal dependencies
+          - name: Python 3.7 with minimal dependencies
             os: ubuntu-latest
-            python: 3.6
-            toxenv: py36-minimal
-
-          - name: Python 3.6
-            os: ubuntu-latest
-            python: 3.6
-            toxenv: py36
+            python: 3.7
+            toxenv: py37-minimal
 
           - name: Python 3.7
             os: ubuntu-latest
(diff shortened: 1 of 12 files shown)


namurphy (approved):
Looks fantastic!  Thank you again!

PR #977, opened 2021-01-15
Update top-level README file

namurphy:
This PR includes several updates to PlasmaPy's principal README file, including updates to our badges.  I did my best guess about the badges that we dropped, though I wasn't completely sure we have stopped using Codecov.  Any that I shouldn't have dropped we can put back in there.  If we add any new GitHub Actions workflows, we should add a badge for them here too.  I still want to add a badge for the YouTube channel, and then put a bit more information about our weekly events.

PR #988, opened 2021-01-26
Remove lingering git hashes in README.md

namurphy:
This is a quick follow-up to something I neglected to finish in #977.


namurphy:
Using my administrators superpowers to merge this, since these changes were already approved in #977.

PR #998, opened 2021-02-03
Solov'ev equilibria

StanczakDominik:
Very early draft that, as it turns out, doesn't seem to work right just yet; but I wanted to put this out in the open early.

All the new stuff is in the notebook, from which I'll eventually extract the functionality.

The interface is a very initial sketch and will most likely change :)

Here's hoping the plots render right!

Those should show the problem I have with those - the conturs don't align with Xref, Yref, which are supposed to be coordinates for the plasma's surface (as a function of a poloidal-like angle tau). I might have to do some more reading on why that is.

      docstrings.

### DIFF
--- a/docs/notebooks/solovev2.ipynb
+++ b/docs/notebooks/solovev2.ipynb
@@ -1,0 +1,6934 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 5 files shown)


StanczakDominik:
Okay, I'm giving up on doing this, because https://github.com/RemDelaporteMathurin/plasma-boundaries already implements it, and is on PyPI, etc!

I think I'll just wrap this into a plasma class, instead.

PR #1010, opened 2021-02-12
Initial draft of particle species class

### LINKED ISSUE #1008: Create a class to represent an individual particle species
In order to create a representation of a multi-component plasma, we need to have a means of defining a single component.  We could do this by creating a class called `Species` to describe the properties of a single particle species in a plasma.  Instances of this class should have attributes like `temperature`, `number_density`, `pressure`, `mass_density`, and `particle` (which would contain the `Particle` instance).

Eventually we'll want to have a way to combine the different components:
```Python
>>> electrons = Species(n=..., T=..., particle="e-")
>>> protons = Species(n=..., T=..., particle="p+")
>>> neutrals = Species(n=..., T=..., particle="H-1 0+")
>>> partially_ionized_plasma = electrons | protons | neutrals  # union operator, or maybe we could use +
```
This would not be in the first pull request, though.

StanczakDominik:
Closes #1008

quick draft for now, using a dataclass; examples and docs still need to be added

      docstrings.

### DIFF
--- a/plasmapy/particles/species.py
+++ b/plasmapy/particles/species.py
@@ -1,0 +1,35 @@
+import astropy.units as u
+
+from dataclasses import dataclass
+from typing import Optional
+
+import plasmapy
+
+from plasmapy.formulary import gyrofrequency, mass_density, thermal_speed
+
+
+@dataclass
+class Species:
+    """
+    A group of particles in a plasma with associated parameters:
+    (number) density, temperature and/or pressure.
+
+    Not all parameters need be specified at creation time.
+
+    Parameters may be arrays, representing, for example,
+    profiles of electron temperature and density from a
+    stellarator diagnostic.
+    """
+
+    particle: plasmapy.particles.Particle
+    number_density: Optional[u.Quantity] = None
+    temperature: Optional[u.Quantity] = None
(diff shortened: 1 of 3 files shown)


StanczakDominik:
A few things to figure out:

- [ ] how to document fields with a dataclass? Probably just like sphinx docs for variables but I still have to try it out
- [ ] What other `formulary` functions should this support? For now I'm adding them as I need them for thesis, but...
- [ ] In related questions: is there a clean-ish way to systematically, automatically add methods from the formulary that would take quantities Species have access to?
- [ ] Ways to wrangle quantity_input into working with this? particle_input? Or skip for now?

plasmapy/particles/species.py
```diff
@@ -0,0 +1,35 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+from plasmapy.formulary import gyrofrequency, mass_density, thermal_speed
+
+from . import particle_class
+
+
+@dataclass
```
namurphy, line 11: Python 3.7! 😺 😺 😺 😺 😺 

I think a dataclass is a good choice here, at the very least to start out with.

PR #1012, opened 2021-02-14
Revert changes accidentally committed to master

namurphy:
...when I try to use PyCharm's interface instead of the command line like I'm used to...


namurphy:
Should be a quick review, as this is only reverting changes that I accidentally pushed to master.

Alternatively, we could do a reverse code review, and reject this pull request if you like the changes that I made!  😬   (Actually this would not be a good idea.)

StanczakDominik (approved):
I'm conflicted but okay, let's remove these helpful lines of code

StanczakDominik:
yes you are, btw!

Issue #1011, opened 2021-02-13
Pickle-able Particles (and other objects, for parallelization)


pheuer:
My draft Thomson scattering fitting function (#974) is currently restricted to using 1 CPU core because the underlying fit function (`thomson.spectral_density`) contains instances of the `Particle` object, which is not pickle-able because it includes a lambda function. If this issue was resolved, enabling parallelization of the fit function would be as simple as including a keyword to `scipy.optimize.differential_evolution`. 

Here is the error if I try to turn on parallelization now:

```
AttributeError: Can't pickle local object 'Particle.__init__.<locals>.<lambda>
```

and here is (I think) the offending line in `Particle.__init__`:

```
Line 408    -->    self._attributes = defaultdict(lambda: None)
``` 

Pickling a model that's then passed between multiple worker pools is, I gather, a pretty common way of parallelizing Python code, so I think it would be very desirable (maybe even a requirement) for all of these basic classes to be serializable. I know there are other serialization modules that DO support lambdas (eg. dill, cloudpickle), but they don't help since the issue is what's compatible with Scipy.

@namurphy  Is this something that can be easily changed?

namurphy:
This is really helpful, so thank you for bringing this up.  I started to address this in #1013, but removing `lambda` expressions in and of itself is not going to be enough to make `Particle` instances ~~pickleable~~ ~~picklable~~ able to be pickled. 

We might need to switch away from using a `defaultdict` for storing the attributes in `Particle` too.  We could do that if we set the following attributes to default to `None` manually in a regular `dict`, though it'd probably be better to refactor.

```Python
["half-life", "element", "isotope", "mass", "ion", "atomic number", "particle", "spin", "baryon number", "charge", "mass number",  "element name", "periodic table", "generation"]
```

namurphy:
I'm adding some (mostly failing) tests in #1014.  `DimensionlessParticle` and `CustomParticle` can be pickled, but `Particle` and ionization state stuff (which depends on `Particle`) cannot.

pheuer:
@namurphy I've realized that a similar problem will arise if we intend to use `numba` to pre-compile any of these functions, since in that case all of the function inputs need to be `numpy` objects. 

I was thinking that one way of solving this problem would be to split functions we want to optimize into a private optimized numba function and a public high-level wrapper that handles the units and translates objects like Particle. Here's a contrived example: 

```python
# A speedy, numba-ified subfunction whose inputs are all np types
[...]

PR #1013, opened 2021-02-15
Remove lambda functions

namurphy:
This pull request removes some `lambda` expressions from `plasmapy.particles` and `plasmapy.utils` because `lambda` expressions cannot be pickled. This partially address #1011.  However, more would need to be done in order to pickle `Particle` instances.  I left the `lambda` expressions in Langmuir stuff to avoid conflicts with #889.  

I propose a new tongue twister:
> I pickled a Particle of pickled PEP8ers.

### DIFF
--- a/docs/development/code_guide.rst
+++ b/docs/development/code_guide.rst
@@ -49,9 +49,8 @@ Coding Style
 * Use ``Optional[type]`` for type hinted keyword arguments with a
   default value of ``None``.
 
-* Avoid using `lambda`.  Anonymous functions created with `lambda`
-  cannot be pickled, and `lambda` notation may be unfamiliar to
-  newcomers to Python.
+* Avoid using `lambda` to define functions, as this notation may be
+  unfamiliar to newcomers to Python.
 
 * There should be at most one pun per 1284 lines of code.
 
--- a/docs/development/code_guide.rst
+++ b/docs/development/code_guide.rst
@@ -49,6 +49,10 @@ Coding Style
 * Use ``Optional[type]`` for type hinted keyword arguments with a
   default value of ``None``.
 
+* Avoid using `lambda`.  Anonymous functions created with `lambda`
+  cannot be pickled, and `lambda` notation may be unfamiliar to
+  newcomers to Python.
+
 * There should be at most one pun per 1284 lines of code.
 
 pre-commit hooks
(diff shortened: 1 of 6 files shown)


plasmapy/particles/ionization_state_collection.py
```diff
@@ -162,8 +166,11 @@ def __init__(
                     )
                 set_abundances = False
 
+        def return_none():
+            return None
+
         try:
-            self._pars = collections.defaultdict(lambda: None)
+            self._pars = collections.defaultdict(return_none)
```
namurphy, line 173: One of the things I read suggested that `defaultdict` objects cannot be pickled, so we might need to venture away from them.  Out of scope for this PR, though.

docs/development/code_guide.rst
```diff
@@ -49,6 +49,10 @@ Coding Style
 * Use ``Optional[type]`` for type hinted keyword arguments with a
   default value of ``None``.
 
+* Avoid using `lambda`.  Anonymous functions created with `lambda`
+  cannot be pickled, and `lambda` notation may be unfamiliar to
```
StanczakDominik, line 53: That lambdas cannot be pickled is of course true; but as far as I know, and ["Is there an easy way to pickle a python function (or otherwise serialize its code)?" on Stack](https://stackoverflow.com/questions/1253528/is-there-an-easy-way-to-pickle-a-python-function-or-otherwise-serialize-its-cod) seems to confirm, Python functions just can't be pickled at all, lambda or no lambda. ["How to pickle a python function with its dependencies?"](https://stackoverflow.com/questions/10048061/how-to-pickle-a-python-function-with-its-dependencies) seems to show you can do it using `marshal` or `cloudpickle` packages, but that's not exactly the same thing - I don't think scipy multiprocessing will handle that, for example.

StanczakDominik (approved):
The changes are okay, but (see comment):

plasmapy/particles/ionization_state_collection.py
```diff
@@ -162,8 +166,11 @@ def __init__(
                     )
                 set_abundances = False
 
+        def return_none():
+            return None
```
StanczakDominik, line 170: Doesn't seem like that one's actually used - see codecov warning!

namurphy: Oh, cool — that means we can use a regular `dict` instead of a `defaultdict` when storing the parameters for `IonizationStateCollection`.

namurphy:
I revised the statement about avoiding the use of `lambda` in the code guide, and removed the reference to pickling.  If this looks good, I'm happy with it being merged.  Thanks for the quick and helpful review!

StanczakDominik:
Boom! :)

PR #1014, opened 2021-02-15
Add mostly failing tests for pickling various objects in plasmapy.particles

namurphy:
This PR adds new tests in `plasmapy.particles` regarding the pickling of particles.  Both `CustomParticle` and `DimensionlessParticle` instances can be pickled, but `Particle` instances (and objects that contain `Particle` instances) cannot.  Hence, most of these tests are marked as failing.  These tests should help later on when we address #1011 and make particles piclkealble.  (Yeah, I gave up on trying to spell that.)

### DIFF
--- a/plasmapy/particles/tests/test_ionization_state.py
+++ b/plasmapy/particles/tests/test_ionization_state.py
@@ -1,6 +1,8 @@
 import astropy.units as u
 import collections
 import numpy as np
+import os
+import pickle
 import pytest
 
 from plasmapy.particles import (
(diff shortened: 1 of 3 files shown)


StanczakDominik (commented):
Good test! I've simplified the tmp file a little, outsourcing that job to pytest (docs.pytest.org/en/stable/tmpdir.html), and then I added a test that verifies that these particles can then be loaded back from their pickles. Looks like that breaks for Custom and Dimensionless particles, and we should probably investigate that first...

namurphy:
> Good test! I've simplified the tmp file a little, outsourcing that job to pytest (docs.pytest.org/en/stable/tmpdir.html)

Thanks!  Good idea!

> and then I added a test that verifies that these particles can then be loaded back from their pickles. Looks like that breaks for Custom and Dimensionless particles, and we should probably investigate that first...

I changed `open(..., "r")` to `open(..., "rb")` since pickle files are binary.

StanczakDominik:
Oh bloody hell, i completely forgot about that, duh...

StanczakDominik (approved):

PR #1017, opened 2021-02-16
Create custom list for particles

### LINKED ISSUE #948: Create a `ParticleCollection` class
The `Particle` class currently provides a means of representing a single particle, and we also have `CustomParticle` and `DimensionlessParticle`.  It would be helpful to have a `ParticleCollection` class that we could use to represent multiple particles together.  That class could have attributes like `mass` that would return an array with units of mass so that we can do array operations and the like...and the same for other properties like `charge` too.  

Here's a use case that came from a discussion on Element.  Suppose we want to find the thermal velocities of all of the ions in a multi-component plasma.  If we have a `ParticleCollection` class, we could do something like:
```Python
>>> particles = ParticleCollection(['p+', 'He-4 1+', 'O-16 1+'])
>>> v_th = thermal_speed(1 * u.MK, particles)
```
If we make it iterable then we could also do things like:
```Python
particles = ParticleCollection([...])
for particle in particles:
    print(thermal_speed(1 * u.MK, particle)
```
This class _shouldn't_ include transitory properties like temperature, since that's not what the `Particle` class does.  That should be in a different class...perhaps something like `MultiComponentPlasma` and/or we could have something like `Species` that keeps track of the plasma parameters for a single species in a plasma...but that is not for this issue.

Probably the best way to approach this class would be to split up development into multiple pull requests.  The first could create a prototype of the class, the next could add `__iter__` and `__next__` to make it iterable, and so on.  That'd make code review more manageable.

namurphy:
This pull request creates `ParticleList`, which inherits from `collections.UserList`.  The purpose of `ParticleList` is to simplify getting `Quantity` arrays for things like `particles.mass` and `particles.charge`, and lists like `particles.symbols`.  The purpose of this pull request is to establish the essential architecture, rather than make this list fully featured.  In particular, I'm adding the features that I expect to need in `IonizationState`, `IonizationStateCollection`, and PlasmaPy-NEI more generally.  I'll probably use `ParticleList` a lot in those places, in particular to get rid of some code that I keep having to rewrite.

I'd also like to enable the formulary to be used like:
```Python
>>> plasmapy.formulary.gyrofrequency(B=5 * u.T, particle=["p+", "D+", "T+", "alpha"])
```
and then get an array of gyrofrequencies back.

Later on, I'd like to add subclasses for different categories of particles.  For example, an `IonicLevels` list would contain attributes shared by all ions.  It'd be really helpful to have an easy way to create a list of all of the ionic levels of a particular element or isotope...like `IonicLevels("He-4")`.  Or alternatively, all isotopes (common, stable, or unstable) of a particular element.

Closes #948.  Related to #687.


namurphy:
As an attempt at computational efficiency, I'm trying to store retrieved attributes so that they can be used later, so long as the contents of the list does not change.  This could help with computationally intensive things where there are lots of particles, and the arrays of masses or charges get used repeatedly.

namurphy:
I've been starting to think that caching this could lead to some weird and unexpected behaviors later on, so I'm going to drop the caching part here.  The issues are:

1. `ParticleList` is a mutable type
2. `ParticleList` can contain `CustomParticle` instances, which are also mutable.  

Suppose one element of `ParticleList` is a `CustomParticle`.  Then suppose someone modifies the mass of that `CustomParticle` directly, without modifying the identities of the particles within the list.   Then the cached values of mass would no longer be valid.  This will probably not be a common issue, but it could cause unexpected and difficult to track down problems.  It'll be safer and cleaner to avoid caching.  We can revisit caching if/when there's a situation where it ends up being needed.

StanczakDominik:
True, though people changing particle masses after creation seems like a rare use case. Dusty plasmas, maybe... But for e.g. impurity studies I'd just want a bunch of completely static particle types. Then we can take advantage of caching without issues. But keeping PRs small is also a great idea :+1:

namurphy:
Ahhhhhhh scope creep on this pull request  🙀  but check this out 🐍 
```Python
>>> α = Particle("alpha")
>>> D = Particle("D+")
>>> T = Particle("T+")
>>> n = Particle("n")
>>> D + T > n + α
<Quantity 2.81810898e-12 J>
```
I've been meaning to implement this for bleeping years.  I think the idea originally came from @Cadair.  I hereby dedicate the use of unicode characters to @rocco8773.  😬

namurphy:
Later on I think it'd be better if the lines like in my last comment returned an instance of a `NuclearReaction` or `FusionReaction` class, as brought up in #1026 and as started in #978.
```Python
>>> fusion_reaction = D + T > n + α
```
For the moment, returning the nuclear reaction energy is still useful.

plasmapy/particles/tests/test_particle_collections.py
```diff
@@ -0,0 +1,302 @@
...
+    ],
+)
+def test_particle_list_membership(args):
+    """
+    Test that the particles in the `ParticleList` match the particles
+    (or particle-like objects) that are passed to it.
+    """
+    particle_list = ParticleList(args)
+    for arg, particle in zip(args, particle_list):
+        assert particle == arg
```
StanczakDominik, line 76: Just thinking about this: this would work like `particle == arg and particle is not arg`? As in, you don't hold references to particle objects, right? (just now going to dig into the implementation!)

plasmapy/particles/particle_collections.py
```diff
@@ -0,0 +1,207 @@
...
+    >>> particle_list.symbols
+    ['e-', 'e+']
+    >>> particle_list + [Particle("p+")]
+    ParticleList(['e-', 'e+', 'p+'])
+    """
+
+    @staticmethod
+    def _list_of_particles_and_custom_particles(
+        particles: Iterable[ParticleLike],
+    ) -> List[Union[Particle, CustomParticle]]:
```
StanczakDominik, line 49: Wow, now I wish we had https://github.com/PlasmaPy/PlasmaPy/issues/687 for this particular annotation. :sweat_smile: 

In fact, what do you think about tagging this...

```suggestion
    ) -> List[Union[Particle, CustomParticle]]:  # TODO #687
```

I can probably set up a grep-by-issue-numbers script to list such notes.

plasmapy/particles/particle_collections.py
```diff
@@ -0,0 +1,207 @@
...
+
+        try:
+            return ParticleList(other)
+        except (InvalidParticleError, TypeError):
+            pass
+
+        try:
+            return ParticleList([other])
+        except (InvalidParticleError, TypeError):
+            raise InvalidParticleError(f"Cannot cast {other} into a ParticleList")
```
StanczakDominik, line 89: Missing tests here for this edge case (and a couple such edge cases below)!

plasmapy/particles/particle_collections.py
```diff
@@ -0,0 +1,207 @@
...
+    def __repr__(self):
+        return f"ParticleList({repr(self.symbols)})"
+
+    def __gt__(self, other):
+        from plasmapy.particles.nuclear import nuclear_reaction_energy
+
+        other_as_particle_list = self._cast_other_as_particle_list(other)
+        return nuclear_reaction_energy(
+            reactants=self.symbols, products=other_as_particle_list.symbols
+        )
```
StanczakDominik, line 113: This should be documented! The comment on this PR you made earlier would be a great start.

StanczakDominik (commented):
Good stuff! I have mostly minor comments about testing, and one more important note about documentation for the nuclear reaction energy feature.

Oh yeah, and that last one should go into the changelog!

plasmapy/particles/tests/test_particle_collections.py
namurphy: The way it works is that if `ParticleList` is provided with a `Particle` or `CustomParticle`, it will hold the reference to the original particle object.  My motivation to do it this way is to be more consistent with how a regular `list` works.

plasmapy/particles/tests/test_nuclear.py
```diff
@@ -170,3 +170,9 @@ def test_nuclear_reaction_energy_kwargs(reactants, products, expectedMeV, tol):
     energy = nuclear_reaction_energy(reactants=reactants, products=products).si
     expected = (expectedMeV * u.MeV).si
     assert np.isclose(expected.value, energy.value, atol=tol)
+
+
+def test_nuclear_reaction_whatever():
```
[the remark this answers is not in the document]
namurphy: Oops!  Yeah, I probably ought to make that a bit more descriptive.  It's at least an accurate test name though, I suppose?

plasmapy/particles/particle_class.py
```diff
@@ -292,7 +322,7 @@ class Particle(AbstractParticle):
     'D 0+'
 
     If the particle doesn't belong to one of those categories, then
-    these attributes return `None`.
+    these attributes provide `None`.
```
StanczakDominik, line 325: Eh, provide... maybe...
```suggestion
    these attributes are `None`.
```

StanczakDominik (approved):

plasmapy/particles/particle_class.py
```diff
@@ -204,7 +205,32 @@ def json_dumps(self, **kwargs) -> str:
         return json.dumps(self.json_dict, **kwargs)
 
 
-class Particle(AbstractParticle):
+class AbstractPhysicalParticle(AbstractParticle):
+    """Base class for particles that are defined with physical units."""
+
+    @property
+    def _as_particle_list(self):
+        from plasmapy.particles.particle_collections import ParticleList
```
namurphy, line 213: The import statement is here to avoid circular imports.

StanczakDominik: Figured!

StanczakDominik: Probably better to put that in as a comment, rather than here! ;0

plasmapy/particles/particle_collections.py
```diff
@@ -0,0 +1,296 @@
...
+    `~plasmapy.particles.exceptions.InvalidParticleError`
+        If an object supplied to `ParticleList` is not `ParticleLike`.
+
+    TypeError
+        If a `DimensionlessParticle` is provided.
+
+    Examples
+    --------
+    A `ParticleList` can be created by calling it with a `list`, `tuple`,
+    or other iterable that provides `~plasmapy.particles.ParticleLike`
```
StanczakDominik, line 42: So I'm guessing the doc test glitches out on this reference, because it can't find ParticleLike. Do we have it in `particles/__init__.py`?

plasmapy/particles/particle_collections.py
```diff
@@ -0,0 +1,296 @@
...
+Remove and return item at index (default last).  Raise `IndexError` if
+the `ParticleList` is empty or the index is out of range.
+"""
+
+ParticleList.remove.__doc__ = """
+Remove the first occurrence of a `ParticleLike` item.  Raise
+`ValueError` if the value is not present.
+"""
+
+ParticleList.reverse.__doc__ = """Reverse the `ParticleList` in place."""
```
namurphy, line 296: This is to override the docstrings for `collections.UserList` which were either absent or clunky.  The implementations for each of these methods didn't need to be overridden.

plasmapy/particles/particle_collections.py
StanczakDominik: Turns out we do. Note, however, that...

`/home/runner/work/PlasmaPy/PlasmaPy/docs/particles/index.rst.rst:64:autosummary: stub file not found 'plasmapy.particles.ParticleLike'. Check your autosummary_generate setting.`

double `.rst` extension for some reason?

StanczakDominik: Something with https://github.com/namurphy/PlasmaPy/tree/particle-collection/docs/api_static (which I still don't quite get?) ? @rocco8773 do you see anything?

namurphy: It seems the current problem was related to me setting `ParticleLike.__name__` for some reason.  I just removed that line and I'm getting the tests to pass.  I'm going to raise a different issue about the docs for `ParticleLike`, since I started going beyond the scope of the current PR.

StanczakDominik: That is pretty wack, indeed. Looks ready to merge, then!

PR #1018, opened 2021-02-17
Separate tox env for slow tests

StanczakDominik:
~~Okay, admittedly I'm doing a little bit of a trick here: I'm limiting our existing envs to tests NOT marked as slow, and having the slow tests env run either slow or not slow tests. This is because~~

1. ~~the not-slow tests aren't going to impact the runtime of the slow tests that much;~~
2. ~~I don't know how to merge coverage reports from different action runs.~~

cc @pheuer 

Turns out codecov is smarter than that, and doing it the simplest way does work!

I ended up running "not slow" tests in most envs and only the "slow" tests in a single env. This limits the variance in runtimes a little.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,8 +18,9 @@ deps =
     !minimal: pytest-xdist
     pytest-github-actions-annotate-failures
 commands =
-    !cov: {env:PYTEST_COMMAND} {posargs}
-    cov: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml
+    !cov: {env:PYTEST_COMMAND} {posargs} -m 'not slow'
+    cov: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml -m 'not slow'
+    cov-slow: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml -m 'slow'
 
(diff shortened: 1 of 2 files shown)


StanczakDominik:
I left two code coverage tests in place, but I might merge them - just curious as to what the output would be.

StanczakDominik:
Well, what do you know. It does look like codecov merges code coverage from multiple test runs. Cool!

Should be good for a review :)

tox.ini
```diff
@@ -18,8 +18,9 @@ deps =
     !minimal: pytest-xdist
     pytest-github-actions-annotate-failures
 commands =
-    !cov: {env:PYTEST_COMMAND} {posargs}
-    cov: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml
+    !cov: {env:PYTEST_COMMAND} {posargs} -m 'not slow'
+    cov: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml -m 'not slow'
+    cov-slow: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}/setup.cfg --cov-append --cov-report xml:coverage.xml -m 'slow'
```
namurphy, line 23: Reading this at first made me wonder if `coverage.xml` would have been overwritten or something, so it's good to hear that Codecov is able to merge coverage reports from multiple tests.

namurphy (approved):
This looks great.  Thank you!

tox.ini
StanczakDominik: Which is why I initially thought we'd have to manually sidestep the issue! I'll keep an eye on the coverage results for the next few days just to be sure.

PR #1028, opened 2021-02-21
Add an attribute for `mean_particle` to ionization state classes

namurphy:
If we want to make `IonizationState` instances compatible with `plasmapy.formulary`, it would be helpful to have a way to convert `IonizationState` and `IonizationStateCollection` instances to `CustomParticle` instances.  I started this and then realized that it'll be a bit easier with `ParticleList` from #1017, so I'm going to wait until #1017 is done before finishing this.

If the ionization fraction is 1 for a particular ionic level in an `IonizationState` instance, then it should return the corresponding `Particle` rather than the `CustomParticle`.  For example, if 100% of the `He-4` is doubly ionized, then it should return `Particle("alpha")`.  

I'm including a keyword for `include_neutrals` because there may be some instances where only charged particles matter.  

There will probably be applications where users want, e.g., the gyroradius of all of the ions.  Converting all of this into a `CustomParticle` instance will not take care of this use case.  

It'll also be worth thinking about how this might interplay with `@particle_input` and if we could eventually consider `IonizationState` and `IonizationStateCollection` to be `ParticleLike`.

### DIFF
--- a/plasmapy/particles/ionization_state.py
+++ b/plasmapy/particles/ionization_state.py
@@ -8,6 +8,7 @@ __all__ = ["IonizationState", "IonicFraction"]
 import numpy as np
 import warnings
 
+from astropy import constants as const
 from astropy import units as u
 from numbers import Integral, Real
 from typing import List, Optional, Union
@@ -18,7 +19,7 @@ from plasmapy.particles.exceptions import (
     InvalidParticleError,
     ParticleError,
 )
-from plasmapy.particles.particle_class import Particle
+from plasmapy.particles.particle_class import CustomParticle, Particle
 from plasmapy.utils.decorators import validate_quantities
 
 _number_density_errmsg = (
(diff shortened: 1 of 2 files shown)


StanczakDominik:
This might make sense from the implementation side, but physically... in neoclassics I see a lot of "charge weighting factors" (may or may not be the same as Z_eff, haven't figured that out yet) where you weigh particular charge states of impurities by density * charge^2 / sum(all densities * all charges^2). That's also a mean, for a particular definition of one; and the arithmetic one may not always make sense.

namurphy:
>  where you weigh particular charge states of impurities by density * charge^2 / sum(all densities * all charges^2)

Yes, definitely.  I've been worried about this too.  I'm not sure how to best handle this.  One possibility would be to have a keyword argument like `use_rms_charge` in the `mean_particle` method, but I'm definitely open to suggestions.

namurphy:
- [ ] Rename `average_particle` to `average_ion` since I'm completely ignoring electrons since I don't give a quark about them, much less three quarks.

StanczakDominik:
>  - [x] Rename `average_particle` to `average_ion` since I'm completely ignoring electrons since I don't give a quark about them, much less three quarks.

<3 quarks, then?

StanczakDominik (approved):
I've given this a look-through a while back. Finished it just now, LGTM!

PR #1029, opened 2021-02-22
Create class to represent MHD normalizations

namurphy:
This PR will add a class that represents a particular system of normalizations that is commonly used for the equations of MHD, in particular for simulations and theory where we can't be bothered to deal with units.  I'm just getting started on this, but will probably use some commits that I had temporarily had in #859, though I ultimately decided to reduce the scope of that PR to an abstract base class.  This partially addresses #808, but there are other common normalizations that will need to be added (hence the abstract base class).

At the moment I'm thinking that it would be good to allow the inputs to be `Quantity` instances with the value of the normalization coefficient, `Unit` instances (for example if we want to normalize length to `1 * u.mm`, or a `ParticleLike` object (since the choice of ion matters when going between mass density and number density).  It'd also be good to be flexible in what inputs the user may provide.  I usually set `L`, `n`, and `B`, but others might normalize to different things.  I'm also wondering if there would be a good way to allow users to specify that the function to normalize length to would be, e.g., the ion inertial length or gyroradius.


changelog/1029.feature.rst
```diff
@@ -0,0 +1 @@
+Expanded functionality in `plasmapy.simulation.normalizations`.
```
namurphy, line 1: - [ ] Update this when the PR is nearing completion.

namurphy:
After having used these normalizations for nigh on a decade now, one thing I've realized is that we'll need to show the dimensional equations, the normalized/dimensionless equations, and the formulae for the normalization constants in order for this to be useful. Unfortunately, this means that the "Notes" section of the class's docstring is going to be rather long and involved.

Some of the alternatives which I'm considering are:
 - Writing a thing that could be uploaded to Zenodo, though this would be a lot harder to maintain than putting it in the documentation.
 - Writing this up in the narrative documentation instead of in the docstring. However, my suspicion is that it would be better to keep the equations close to this class.
 - Writing an example notebook.
 - Adding glossary entries on different forms of MHD and other equations describing plasmas.

In any case...I'm still working on this, even if slowly!

Issue #1032, opened 2021-02-23
Release version 0.6.0


namurphy:
Following up on the discussion in #895, it'd be good to release version 0.6.0 in March or April.  We can use this issue to discuss what we want to make sure gets into this release.

namurphy:
There are a few things related to `plasmapy.particles` that I think we ought to get into our next release.  

 - [ ] We updated `ParticleLike` in #985 to include a (hopefully) complete description of how particles can be represented in PlasmaPy.  We ought to use `ParticleLike` in docstrings consistently in the formulary and elsewhere to link to that description, especially since particles can be confusing when beginning to use PlasmaPy.  
 - [ ] If we want to use `ParticleLike` in type hint annotations, we would need to update `@particle_input` to accept `ParticleLike` instead of (or in addition to) `Particle`.  Related to: #912, #913.  
 - [ ] Similarly, we should make `@particle_input` accept `CustomParticle` instances (#807).
 - [ ] I'd like to get `ParticleList` (#948, #1017) implemented, and make it such that `plasmapy.formulary` functions can work with lists of particles (i.e., `gyroradius(B=..., particle=["p+", "e-"])` would return an array with the gyroradii for protons and electrons).  This would also require some changes to `@particle_input` too.

StanczakDominik:
Oh wow, you mean you read what's going on in those big PRs? I have to admit I've stopped doing that for PRs I'm not handling, and wanted to discuss this at the telecon today :D but sure...

I'll review the particle list asap!

StanczakDominik:
You know what we should also do?

- [ ] Change the three featured notebooks on the main docs index to showcase the newest and coolest functionality

namurphy:
> Change the three featured notebooks on the main docs index to showcase the newest and coolest functionality

Perhaps we could have the first notebook be one for getting started with PlasmaPy, and the other two could be the newest and coolest functionality?

StanczakDominik:
Quoting myself from chat:

> Honestly I dislike the ITER data analysis notebook, it's too basic; can we just replace that with the proton radiography one and keep thomson up? Make this an actual highlight reel and I promise I'll write a proper "start using plasmapy" notebook for 0.7?

This is what pushed through #1065 and what I commit to doing for 0.7. For now, we're on schedule to release tomorrow, so I'll close this one :)

PR #1033, opened 2021-02-23
Make @particle_input compatible with CustomParticle

namurphy:
This pull request is intended to make `@particle_input` work with `CustomParticle`.  That is to say, the goal is to make `@particle_input` pass through a `CustomParticle` instance as if it were a `Particle` instance.

There will probably be times where a `CustomParticle` is appropriate and other times where it would be inappropriate.  It would probably be good to have an optional keyword argument for `accept_custom_particles`, though I'm not sure what it should default to.  There are a whole lot of functions (e.g., `atomic_symbol`) which will require a real `Particle`, so maybe it should default to `False`.  

This address part of #807.  This is one of the things I discussed in #1032.

So far I'm just adding a failing test.


StanczakDominik:
I think it should default to True; functions that can't take CustomParticles should be the exception, not the norm, or nobody will use them for fear of "it's going to glitch out eventually".

namurphy:
> I think it should default to True; functions that can't take CustomParticles should be the exception, not the norm, or nobody will use them for fear of "it's going to glitch out eventually".

That seems like good reasoning to me.  My main consideration is that there are a whole lot of functions in `plasmapy.particles.atomic` (and probably elsewhere) that assume a `Particle` rather than `CustomParticle` (e.g., `atomic_symbol`).  With that said...maybe we could do some careful exception handling and if the function being decorated raises an `AttributeError`, then we could provide an error message that states that the function needs a `Particle` rather than a `CustomParticle`.  If we do that, then maybe we could take care of this situation automagically without having to add an extra keyword argument.  We would want to document this carefully though...

StanczakDominik:
Or we could just use our test suite to figure out which functions glitch out with custom particles, then fix the default for those!

PR #1035, opened 2021-02-23
PRad Bugfix

pheuer:
I found a few minor bugs in sections of `proton_radiography.SyntheticProtonRadiograph` that I used rarely: 

- Accessing the `max_deflection` property raised an exception if particles had been deflected away from the detector (since `v` and `v_init` were then un-equal in length). I fixed this by removing deflected particles from `v_init` as well in `_coast_to_detector`.

- `_adaptive_dt` depended on the velocity of the particles `v0`, which was calculated in `create_particles` but not defined if the user provided particles in `load_particles`. Furthermore, user-provided particles don't necessarily have the same energy. I solved that by using `max(v)` in `_adaptive_dt`, which I'm calculating once at the beginning of `run()`.

- `_adaptive_dt` now calculates the cyclotron period using the provided particle charge and mass (previously assumed protons).


plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -365,13 +365,15 @@ def test_run_options():
...
+    grid = _test_grid("constant_bz", num=50, B0=250 * u.T)
     source = (0 * u.mm, -10 * u.mm, 0 * u.mm)
     detector = (0 * u.mm, 200 * u.mm, 0 * u.mm)
     sim = prad.SyntheticProtonRadiograph(grid, source, detector, verbose=False)
-    sim.create_particles(1e4, 3 * u.MeV, max_theta=1 * u.deg)
+    sim.create_particles(1e4, 3 * u.MeV, max_theta=10 * u.deg)
     with pytest.warns(RuntimeWarning):
         sim.run(field_weighting="nearest neighbor", dt=1e-12 * u.s)
+        # Calc max deflection
+        sim.max_deflection
```
StanczakDominik, line 376: Would be good to at least have a bounds check (`assert lower < sim.max_deflection < upper`) here, rather than calculating it just for the non-exception-throwing-execution code coverage!

StanczakDominik (approved):

plasmapy/diagnostics/tests/test_proton_radiography.py
pheuer: Good idea - I've set a bound between 0 and pi/2 (which is valid for particles with a small `max_theta`).

PR #1036, opened 2021-02-24
Add some unicode particle aliases

namurphy:
Because I'd apparently rather play with unicode than make the increasingly wise life decision of going to bed, this pull request adds a few additional unicode particle aliases (e.g., we would be able to do `Particle("α")` for an alpha particle).  This could be useful for code and examples that will be read much more frequently than they are written.

At present a lot of the aliases are not documented.  I wonder if it would be helpful to have an ``aliases`` attribute to `Particle` that could list all of the valid aliases for the different particles.  I've also been thinking about ``unicode_symbol``, ``mathml``, and ``latex_str`` attributes too for different representations, but that's not for this PR.


namurphy:
I still need to add tests, which is presently giving me much-needed motivation to go to bed.

StanczakDominik (approved):
Nice and short! Looking forward to the tests, but, seal of approval for now :D

StanczakDominik:
Looks Good To Me Conditional On Tests Passing, But For The Changelog Entry. LGTMCOTPBFTCE. Think I'll need a macro for this...

PR #1039, opened 2021-02-24
Improve docstring for `Particle.is_category`

namurphy:
The purpose of this PR is to ~~procrastinate doing other things~~ make the docstring for `Particle.is_category` consistent with the numpydoc standard.

PR #1040, opened 2021-02-24
Follow-up to #1036

StanczakDominik:
Added a forgotten alias that we should have added ages ago. Noted indirectly by https://github.com/PlasmaPy/PlasmaPy/pull/1039#issuecomment-785295004.


namurphy:
I dare say that there is an element of humor in this.

namurphy (approved):

Issue #935, opened 2020-10-15
Review formulary docstrings for equations and clarity


rocco8773:
The majority of the `plasmapy.formulary` docstrings are in good shape, but I still think there's room for improvement.  Some functional docstrings are missing their associated formulae and some docstrings are difficult to read.  I suggest we do the following...

1. Add formulae to all `plasmapy.forumulary` functionality where it is missing.  (**This should be a default requirement for all formualry items.**)
1. Move/copy the formulae from the `Notes` section to the main body of the docstring.  This would move the equation to just below the function signature in the documentation, which would help with readability and understanding for users.
1. Review body, parameter, etc. descriptions for clarity.  *Keep in mind how these would read if you were new to them.*

Tiger-Du:
Hi, Erik. I'd be happy to work on this issue. Plasma physics is still rather new to me, but being a novice would be useful for this issue I think.

Tiger-Du:
Hi, @rocco8773. I submitted a draft pull request (#951) that adds a section named `Formula` to the top of the docstring of one function, with the formula copied from `Notes`. Please let me know whether it's similar to what you had in mind and whether I should go ahead and add this section in the same way to all other functions in `plasmapy.formulary`.

Also, about the ":math:" parts: Do those render at some point into mathematical symbols? Or is ":math:" just a formatting convention for denoting mathematics in docstrings?

I'll continue to commit to #951, or make a separate pull request, to address 1. and 3. in your description of this issue.

Tiger-Du:
From recent discussion on Item 2:
1. If the formula is exceptionally complex, put the simplest version of the formula (if such a version exists) at the top of the docstring and put the full formula in the "Notes" section.
2. If the variables of the formula are not the same as the parameters of the function, describe the relationship between the parameters and the variables in the "Parameters" section.

Thus, this docstring feature is expected to differ in general from function to function, which we are ok with.

Issue #946, opened 2020-10-27
Create a script to inject standard parameter descriptions into docstrings


namurphy:
There are a whole bunch of functions in `plasmapy.formulary` and `plasmapy.particles` that have the same arguments.  Because they have the same arguments, they should have the same description in the `Parameters` section of the docstring.  But they don't.  If we want to standardize them, then right now we would have to copy and paste the same thing in every function that uses them.  If we want to change the standard description, then we would have to make the change in every docstring that they appear in.  

One possibility that we thought about was to create a decorator akin to `@modify_docstring` in #908 that would inject the standardized parameter descriptions into the docstrings during import.  However, this could slow down imports, and the docstrings in the files would not contain the parameter descriptions.

An alternative solution would be to create a script that goes through each of the requested files, looks for functions and methods that take in standard arguments like `B`, `n_e`, and `ion`, and then inject the standardized docstring directly into the module file, which would then be committed.

namurphy:
Thinking more about this...this could be achieved using reST substitutions such as those defined in `docs/common_links.rst`.  However, the text to be substituted would not appear in the docstrings when using `help`, which would be a significant con in this case.  A possible middle ground would be to define substitutions that have a sufficiently descriptive name such that it'd be good enough for `help`, though the pros and cons of this would still need to be considered.

Another alternative would be to add tests that check that the descriptions for each parameter exactly match the canonical description...though of course this approach has pros and cons too.   

I'm not really sure how to handle this in a way which is a significant improvement to the status quo, so I'll set the priority of this to low.

PR #1041, opened 2021-02-25
Formulary Documentation: Adding Missing Formulas to Docstrings

Tiger-Du:

      docstrings.

This PR is for Issue #935.

I went through the documentation of `plasmapy.formulary`, and from what I can tell, outside of `plasmapy.formulary.braginskii.py`, the only two (user-facing) functions that lack formulas are `dimensionless.quantum_theta` and `dimensionless.beta`. These quantities are simple ratios, but I believe they still should get complete documentation. Formulas for these two functions are added in this PR.

I plan to add formulas to the docstrings in `braginskii.py` in this PR although these may be more complicated to document because of multiple supported models of classical transport.

On a related note, I am not sure whether the methods of classes, such as the Classical Transport class in `braginskii.py`, should be documented in the same way (or at all) as standalone / free functions, and I would like to hear what other contributors think.


StanczakDominik:
> On a related note, I am not sure whether the methods of classes, such as the Classical Transport class in braginskii.py, should be documented in the same way (or at all) as standalone / free functions, and I would like to hear what other contributors think.

If I remember correctly, the functions call the ClassicalTransport methods and not the other way around; I would suggest bluntly referencing the ones with the actual implementation (e.g. "Wraps classicaltransport.this_method") from the wrappers :)

plasmapy/formulary/dimensionless.py
```diff
@@ -31,10 +31,18 @@
...
     Compares Fermi energy to thermal kinetic energy to check if quantum
     effects are important.
 
+    The quantum theta (:math:`\theta`) of a plasma is defined by
+
+    .. math::
+        \theta = \frac{E_{T}}{E_{F}}
+
+    where :math:`E_{T}` is the thermal energy of the plasma
+    and :math:`E_{F}` is the Fermi energy of the plasma.
```
StanczakDominik, line 44: It would probably make sense to add some notes here on how those are defined (e.g. through a See Also section if we implement them elsewhere, or a quick oneliner for the thermal pressure in the other case). The same applies to `beta`.

StanczakDominik (changes requested):
Good stuff! One note, though :)

Tiger-Du:
@StanczakDominik - Hi, Dominik. Thank you for reviewing this PR. I made the changes you suggested in `dimensionless.py` by adding "See Also" sections and the formula for thermal energy in the "Notes" section. I also added formulas (or introduced them) in the docstrings of `braginskii.py`. I am deciding to add formulas to both the methods and the free functions here because even though they are (very) similar, they are separate functionality. I do not think we should spend too much effort editing `braginskii.py` in this PR because it is due for refactoring. But I think the changes made so far to `braginskii.py` make it more presentable for 0.6.0. Hopefully, this PR could be quick to merge! :)

StanczakDominik:
> I am deciding to add formulas to both the methods and the free functions here because even though they are (very) similar, they are separate functionality.

One worry I have about that is that if this works as I remembered and one calls the other, changing - for whatever possible reason, suppose we made a mistake and were off by sqrt2 for concreteness - one implementation will propagate changes in results, but not the documentation, into the other. I know I could and likely would easily forget that, which is why I advocated for the `see also` or `thin wrapper for` (as we do with aliases) approach earlier.

But it's a worry I can live with, I think. I can get you a review sometime after breakfast :)

Tiger-Du:
@StanczakDominik - Good point. I am not very familiar with wrappers, so maybe you could suggest some code for that. But perhaps this problem will end up being moot in the future if some significant refactoring of `braginskii.py` happens. We might also address this problem with something like #946 but for docstring formulas.

Also, forgot to mention, but I plan to add formulas to the docstrings in `formulary.magnetostatics.py` in this PR too.

StanczakDominik:
Gah, I'm sorry, that turned out to be a off-by-one-breakfast error combined with a wrong meal bug... getting to it now!

When I say "wrapper", I mean something like how this:

https://github.com/PlasmaPy/PlasmaPy/blob/6a161f09530ee3d6870ea00a5c95f5ef278e12fb/plasmapy/formulary/braginskii.py#L740-L794

is a very thin wrapper for this:

https://github.com/PlasmaPy/PlasmaPy/blob/6a161f09530ee3d6870ea00a5c95f5ef278e12fb/plasmapy/formulary/braginskii.py#L457-L491

Though yeah, I think any refactor we'd make here would start by deprecating the heck out of these functions, as - looking at this for the first time in a while - that's just terrible design :sweat_smile:

plasmapy/formulary/braginskii.py
```diff
@@ -482,9 +493,11 @@ def resistivity(self) -> u.Ohm * u.m:
 
     @property
     def thermoelectric_conductivity(self):
-        """
+        r"""
         Calculate the thermoelectric conductivity.
 
+        The thermoelectric conductivity (:math:`\hat{\beta}`) of a plasma is defined by...
```
StanczakDominik, line 499: Looks like this one's missing a continuation? I'd be fine merging it as is, since this is, of course, an incremental improvement over the previous state of this module.

Maybe a [TODO note](https://www.sphinx-doc.org/en/master/usage/extensions/todo.html#directive-todo) would be useful, for easier text search of these!

plasmapy/formulary/braginskii.py
```diff
@@ -502,9 +515,21 @@ def thermoelectric_conductivity(self):
     @property
     @validate_quantities
     def ion_thermal_conductivity(self) -> u.W / u.m / u.K:
-        """
+        r"""
         Calculate the thermal conductivity for ions.
 
+        The ion thermal conductivity (:math:`\kappa`) of a plasma is defined by
```
StanczakDominik, line 521: Just a thought - one key piece of intuition I was missing while reviewing these three years ago was the understanding that the way these are *defined* is a transport law like $\vec{q} = -\kappa * \nabla T$, q being the heat flux caused by the temperature gradient. What transport theory actually does is provide estimates, based on reasonable assumptions, for the magnitude of $\kappa$ and other such coefficients. And then you can have different variants, including curvature effects (thus, neoclassical), or turbulence (thus, anomalous).

So yeah, despite wondering a little about the word "defined", I think this is a good change, but I wonder whether we shouldn't rewrite the docs for the whole transport module a little, to account for that...

gah, of course we should do that, we should take a hammer to the whole thing, after all. I'll make a note to do that when we refactor it.

StanczakDominik (approved):
The docstrings look good, I'd be happy to merge it now, but I figured I'd wait and see if you want to discuss the comments a little :) If not (future PRs, and stuff), just say so and I can hit merge any time.

plasmapy/formulary/braginskii.py
StanczakDominik: But at the same time, those transport laws are very general and don't necessarily apply only to classical transport. So it might be a moot point.

Tiger-Du: Interesting points. We could change "defined" if users or contributors might be confused by this term. Perhaps something like "... is related by the formula ..." or "This function computes ... by the formula ...". But I think that we should discuss this and figure this out in an Issue or Discussion so that the documentation can be standardized.

plasmapy/formulary/braginskii.py
Tiger-Du: Yes, I just left off with "..." because I could not parse how the method works with the multiple models supported. But I added todo's for picking up on these formulas later on.

Tiger-Du: The todo extension is not in our conf.py file for the docs. Should I add it there in this PR, or do you think #TODO is fine? The todo directive would not be displayed in the build, but #TODO would be I believe.

StanczakDominik: Completely forgot we didn't have that in our config! Added it, let's see if it works :)

plasmapy/formulary/braginskii.py
StanczakDominik: Good point, let's come back here from #1044 later :)

StanczakDominik:
That should do it, merging! Thanks, @Tiger-Du :)

PR #1046, opened 2021-02-26
IonicFraction refactor

StanczakDominik:
Starting to apply changes to `IonicFraction` from 
https://github.com/PlasmaPy/PlasmaPy/discussions/1020#discussioncomment-408683. Refactoring the implementation a little along the way. Part of #1059.

- [x] ~~docs for T_i~~ will go in at a future time, along with that feature!


plasmapy/particles/ionization_state.py
```diff
@@ -35,7 +35,7 @@ class IonicFraction:
     ion: `~plasmapy.particles.particle_class.ParticleLike`
         The ion for the corresponding ionic fraction.
 
-    ionic_fraction: real number between 0 and 1, optional
+    ionic_level: real number between 0 and 1, optional
```
namurphy, line 38: For the attribute, I'd keep the name as `ionic_fraction` since it's the fraction of the element/isotope that's at this ionic level.

plasmapy/particles/ionization_state.py
```diff
@@ -84,32 +82,32 @@ def __eq__(self, other):
...
-            f"IonicFraction({repr(self.ionic_symbol)}, "
-            f"ionic_fraction={self.ionic_fraction})"
+            f"IonicLevel({repr(self.ionic_symbol)}, " f"ionic_level={self.ionic_level})"
         )
 
     @property
     def ionic_symbol(self) -> str:
         """The symbol of the ion."""
-        return self._particle.ionic_symbol
+        return self.ion.ionic_symbol
```
namurphy, line 102: Yes, this is better!

plasmapy/particles/ionization_state.py
```diff
@@ -166,7 +164,7 @@ class IonizationState:
         an element, isotope, or ion; or an integer representing the
         atomic number of an element.
 
-    ionic_fractions: `~numpy.ndarray`, `list`, `tuple`, or `~astropy.units.Quantity`; optional
+    ionic_levels: `~numpy.ndarray`, `list`, `tuple`, or `~astropy.units.Quantity`; optional
```
namurphy, line 167:
```suggestion
    ionic_fractions: `~numpy.ndarray`, `list`, `tuple`, or `~astropy.units.Quantity`; optional
```
This also should be changed back to `ionic_fractions`.

plasmapy/particles/ionization_state.py
```diff
@@ -701,26 +695,26 @@ def tol(self, atol: Real):
         else:
             raise ValueError("Need 0 <= tol < 1.")
 
-    def _get_states_info(self, minimum_ionic_fraction=0.01) -> List[str]:
+    def _get_states_info(self, minimum_ionic_level=0.01) -> List[str]:
```
namurphy, line 698:
```suggestion
    def _get_states_info(self, minimum_ionic_fraction=0.01) -> List[str]:
```

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -574,25 +560,35 @@ def test_IonizationState_exceptions(test):
...
+
+@pytest.mark.parametrize("key", expected_properties)
+def test_IonizationState_attributes(instance, key):
     """
     Test a specific case that the `IonizationState` attributes are
     working as expected.
     """
     expected = expected_properties[key]
-    actual = eval(f"instance.{key}")
+    actual = getattr(instance, key)
```
namurphy, line 591: Much better!

namurphy (changes requested):
Thank you again for doing this!  

My main comment is that we should keep the same name for the attributes originally named `ionic_fraction`, since the ionic fraction is the fraction of the atoms of an element that are at that ionic level.  Renaming the `IonicFraction` class to `IonicLevel` is what we want to do, since `IonicLevel` contains the ionic fraction as well as some other properties.  For an example of the naming (at least for the way heliophysicists tend to use it)... `He 1+` is the ionic level.  If half of the helium is at that ionic level, then the ionic fraction would be 0.5.  

If you request another review then I should be able to get to this pretty quickly.

StanczakDominik:
Gah, you're of course right; that's just overzealous grepping. Getting to it.

plasmapy/particles/ionization_state.py
StanczakDominik: The argument, but not the field of `IonizationState`, if I understand correctly?

StanczakDominik: Oh, okay, I get it now. Sorry, tired.

StanczakDominik:
That should do it! Sorry I didn't pay enough attention to this.

namurphy (approved):

PR #1049, opened 2021-02-27
Proton radiography with mesh grids

pheuer:
In proton radiography experiments a wire mesh grid is often placed between the proton source and the object of interest, leaving a shadow of the grid in the proton fluence. The displacement of these shadow grid lines can then be used to quantitatively extract the line-integrated force experienced at each grid vertex.

This PR adds a new method `add_wire_mesh()` to `diagnostics.proton_radiography.SyntheticProtonRadiograph` that can be used to create synthetic radiographs with a mesh in place. In short, this is accomplished by calculating the positions of each particle in the mesh plane, figuring out which of those would hit one of the mesh wires, and then removing those particles from the calculation prior to running the tracing algorithm. 

Different combinations of keywords can be used to create circular or rectangular meshes with different numbers of wires. The meshes can be rotated or placed at arbitrary locations or angles relative to the object grid. All of these options are demonstrated in the example notebook.

The end result on a synthetic radiograph looks like this. 

prad_mesh_illustration

Information about the line-integrated field can be extracted from the shifted positions of the grid vertices.


StanczakDominik:
For the name - maybe wire_mesh to be fully explicit?

Looking into it, looks like cool functionality!

pheuer:
> For the name - maybe wire_mesh to be fully explicit?

Good idea: I'll make that change now.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +300,205 @@ def _log(self, msg):
...
+
+            Raises a ValueError if the specified grid removes ALL of the
+            particles.
+
+
+        Returns
+        -------
+
+        nremoved : int
+            The number of particles removed
```
StanczakDominik, line 375: Okay, that's non-intuitive to me (I haven't read the notebook yet, so assume I'm a user who stumbled upon the docstring before the example).

I glanced at the test and you create the simulation, create the particles, add the wire mesh, then run the simulation. How does the wire mesh know how many particles it's going to remove before you run the simulation?

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +300,205 @@ def _log(self, msg):
...
+
+        # Raise exception if mesh is AFTER the field grid
+        if np.linalg.norm(location - self.source) > np.linalg.norm(self.source):
+            raise ValueError(
+                f"The specified mesh location, {location},"
+                "is not between the source and the origin."
+            )
+
+        # Calculate the times required to evolve each particle to the mesh
+        # plane
```
StanczakDominik, line 430: So at this point you stop defining the geometry of this wire mesh and start to do stuff with particles. Doesn't this suggest we should break this up into two functions/methods/whatever? This one seems like it's doing too much.

Also, suppose I ran `sim.run` first, then added the wire mesh? What would happen then?

StanczakDominik (commented):
Partial review!

plasmapy/diagnostics/proton_radiography.py
StanczakDominik: Or suppose my experimental set up had two meshes one after the other?

I think we might want to have a list of (representations of physical) objects for particles to interact with. The mesh is one, the grid is another, and then the detector is a third one. Given their spatial ordering, we could then iterate over them.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +300,205 @@ def _log(self, msg):
     # Define some constants so they don't get constantly re-evaluated
     _c = const.c.si.value
 
+    # *************************************************************************
+    # Create mesh
+    # *************************************************************************
```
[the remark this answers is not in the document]
pheuer: Agreed - I wish there was a version that editors interpreted as sections that could be minimized or expanded (the way many treat function headers).

plasmapy/diagnostics/proton_radiography.py
pheuer: So the wire mesh function works by calculating where particles WILL be in the future (in the wire mesh plane), determining which ones will intersect a wire, then deleting those particles before you begun `run()`. Since the mesh has to be before the grid, the system is entirely linear at that point. 

I thought about having `add_wire_mesh()` store this information in a dict, then having `run()` apply the mesh (or meshes) at the beginning of the `run()` call, but I was worried that that obscured what was happening under the hood too much.

plasmapy/diagnostics/proton_radiography.py
pheuer: So we could replace part of this function with an `coast_to_plane(center, hdir, vdir)` function. However it wouldn't replace `coast_to_grid` (which doesn't assume the grid is rectangular) or `coast_to_detector` (which also handles removal of particles going the wrong way. We could potentially refactor the particle deletion part of `coast_to_detector` out, and then have `coast_to_plane` and `remove_particles_heading_away`?

I could see having a list of meshes? Although I've never seen an experiment that used two, it is in theory possible.

StanczakDominik: I'll draft a refactor sketch later to show you what I mean here - currently a little swamped!

StanczakDominik:
Okay, so how about this. Obviously, very early sketch, naming is dreadful but just to illustrate the idea:
```python
class ProtonRadiographySimulation:
	objects: List[ExperimentalGeometryObject]    # e.g. [wire mesh, grid, detector]
	particles: list of positions/velocities # separate arrays, ofc

	def run(self, ...):
		for object in self.objects:
			self.coast_particles_to(object)
            # calls object.get_coasted(self.particles) or something? I could not quickly come up with a non-Yoda naming for this
            # anyway each object can implement its own version of "where in space do I need these particles"             
			self.simulate_particles(object)
            # similar; non-grid objects just pretty much implement this as pass
            # while the grid does all of the current Boris pushing etc
            self.remove_or_deposit_particles(object)
            # with the wire mesh removing particles and the detector removing them,
            # and the grid passing on this
```

This might be a terrible idea, but I just wanted to get your opinion on this :)

pheuer:
I think that might be generalizing a little too far, but I will implement something like this to handle meshes (to allow users to specify meshes before `create_particles`, then have them applied at runtime). I will also generalize the `coast_to_mesh` and `coast_to_detector` into a `coast_to_plane` function, but I think `coast_to_grid` should be distinct (for example, if the grid corner is pointed at the source the `coast_to_plane` idea won't work).

I'll try this out in the next few days.

StanczakDominik:
I mean, if you think that's generalizing too much - we can absolutely leave it at that! I was just probing your opinion on it. I do think the method as it is might do too much in one go, but I'm not dead set on one particular method of solving that!

pheuer:
@StanczakDominik Check out the refactor I just did and let me know what you think. `_coast_to_plane` replaces the common functionality between `coast_to_grid` and the beginning of `add_wire_mesh`. 

After discussion with @rocco8773 yesterday, I've also changed things so that `add_wire_mesh` just basically does input validation, then stores the results as a dict in a list of meshes that are then processed at the beginning of `run()`. This way you can call `add_wire_mesh` before `create_particles` which makes more intuitive sense. 

It's possible for the user to get into trouble by setting `mesh_hdir` and `mesh_vdir` that are not orthogonal. You can imagine the grid ends up squashed into the angle between them. Maybe the kwargs should be `mesh_hdir` and `rot_around_hdir` or something that allows rotation around `hdir` but inherently keeps them orthogonal?

StanczakDominik:
Will take a look tomorrow, currently in deadline-driven-development mode :smile:

pheuer:
No problem, sounds good to me! Glad you're making progress on the thesis!

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+
+    def add_wire_mesh(
+        self, location, extent, nwires, wire_diameter, mesh_hdir=None, mesh_vdir=None
+    ):
+        """
+        Add a wire mesh grid between the particle source and the object grid
+        that blocks particles which intersect the wires.
```
StanczakDominik, line 317:
```suggestion
        that blocks particles whose paths intersect the wires.
```
Unless those are pretty chunky particles.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+            In spherical coordinates theta is the polar angle.
+
+        extent : Tuple of 1 or 2 `~astropy.units.Quantity`
+            The size of the mesh grid (in the mesh plane). If one value
+            is provided, the mesh is circular and the value provided is
+            interpreted as the diameter. If two values are provided, the
+            mesh is rectangular and they the values are interpreted as the
+            width and height respectively.
+
+        nwires : Tuple of 1 or 2 ints
```
StanczakDominik, line 343: Could also be a single int for convenience, but this is fine. Oh, you cast it to tuple later - great! Just note it down here.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+            The number of wires in the horizontal and vertical directions. If
+            only one value is provided, the number in the two directions is
+            assumed to be equal. Note that a wire will cross the center of the
+            grid only when nwires is odd.
```
StanczakDominik, line 347:
```suggestion
            mesh only when nwires is odd.
```

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+        """
+        x = self._coast_to_plane(location, mesh_hdir, mesh_vdir)
+
+        # Particle positions in 2D on the mesh plane
+        xloc = np.dot(x - location, mesh_hdir)
+        yloc = np.dot(x - location, mesh_vdir)
+
+        # Create an array in which 0 indicates that a particle has hit one of
+        # the wires and any other value indicates that it has not
+        no_hit = np.ones([self.nparticles])
```
StanczakDominik, line 457: If nparticles is the shape you can drop the list, I think; also force a boolean dtype to conserve a little memory :)

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+
+        # Mark particles that overlap vertical or horizontal position with a wire
+        h_centers = np.linspace(-width / 2, width / 2, num=nwires[0])
+        for c in h_centers:
+            no_hit *= np.where(np.isclose(xloc, c, atol=wire_radius), 0, 1)
```
StanczakDominik, line 462: For booleans this will probably simplify to something like... (Not quite this; just sketching)
```suggestion
            no_hit &= ~np.isclose(xloc, c, atol=wire_radius)
```

Probably easier to reason about this if you flip the condition, start with all falses and apply "or" with the isclose for each wire.

StanczakDominik (commented):
Partial look through, found a way to simplify the logic a little...

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -43,6 +43,16 @@ def _test_grid(name, L=1 * u.mm, num=100, B0=10 * u.T):
...
+    defaults = {
+        "B0": 10 * u.T,
+        "E0": 5e8 * u.V / u.m,
+        "phi0": 1.4e5 * u.V,
+    }
+
+    for k in defaults.keys():
+        if k not in kwargs.keys():
+            kwargs[k] = defaults[k]
+
```
StanczakDominik, line 55: Why not just add these to the default arguments? Won't that work the same way?

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -417,6 +460,203 @@ def test_synthetic_radiograph():
...
+    This test runs a known example problem and compares to a theoretical
+    model for small deflections.
+
+    Still under construction (comparing the actual form of the radiograph
+    is possible but tricky to implement).
+    """
+
+    # The Gaussian sphere problem for small deflection potentials
+    # is solved in Kugland2012relation, and the equations referenced
+    # below are from that paper.
```
StanczakDominik, line 474:
```suggestion
    # The Gaussian sphere problem for small deflection potentials
    # is solved in Kugland2012relation, and the equations referenced
    # below are from that paper.
    # https://doi.org/10.1063/1.4750234
```

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -417,6 +460,203 @@ def test_synthetic_radiograph():
     h, v, i = sim.synthetic_radiograph(size=size, bins=bins, optical_density=True)
 
 
+def test_gaussian_sphere_analytical_comparison():
+    """
+    This test runs a known example problem and compares to a theoretical
+    model for small deflections.
+
+    Still under construction (comparing the actual form of the radiograph
+    is possible but tricky to implement).
```
StanczakDominik, line 469: This is a cool idea! I suspect with this sort of thing you'd need to run a bunch of simulations, then average the histograms to limit particle noise?

plasmapy/diagnostics/tests/test_proton_radiography.py
```diff
@@ -417,6 +460,203 @@ def test_synthetic_radiograph():
...
+    l = 10
+    L = 200
+
+    # Define and run the problem
+    grid = _test_grid("electrostatic_gaussian_sphere", num=100, phi0=phi0 * u.V)
+    source = (0 * u.mm, -l * u.mm, 0 * u.mm)
+    detector = (0 * u.mm, L * u.mm, 0 * u.mm)
+
+    with pytest.warns(
+        RuntimeWarning, match="Fields should go to zero at edges of grid to avoid "
```
StanczakDominik, line 489: :heart:

StanczakDominik (commented):
I've finished looking through the tests and the code. A couple more suggestions and this should be on its way to a speedy merge.

I've also read that notebook and **wow**, this one just *clicked* for me. Kudos!

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+
+        if self.nparticles - nremoved <= 0:
+            raise ValueError(
+                "The specified mesh is blocking all of the particles. "
+                f"The wire diameter ({2*wire_radius}) may be too large."
+            )
+
+        self.x = self.x[i, :]
+        self.v = self.v[i, :]
+        self.theta = self.theta[i]  # Importat to apply here to get correct grid_ind
```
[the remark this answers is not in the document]
pheuer: Haha true, I'll add some more

plasmapy/diagnostics/tests/test_proton_radiography.py
pheuer: Well, one simulation is sufficient with enough particles to limit noise, but there are also histogram-boxing effects that you have to be careful about. 

Really a spherical geometry isn't ideal here: the tests in `test_1D_deflections` are also comparisons against an analytical result (sort of) but since they are cylindrical it's easy to sum up one dimension and get a low-noise result with fewer particles.

plasmapy/diagnostics/tests/test_proton_radiography.py
pheuer: Hm... yep you're right. 

I think I've been doing this sometimes when I have an unwieldly long list of possible arguments, but the way you suggest is definitely better here.

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -576,43 +787,43 @@ def _coast_to_grid(self):
...
+        Calculates the positions where the current trajectories of each
+        particle impact a plane, described by the plane's center and
+        horizontal and vertical unit vectors.
+
+        Returns an [nparticles, 3] array of the particle positions in the plane
 
-        These positions are used to quickly compute null radiographs, which are
-        used to determine the degree of deflection.
+        Note that this function does not alter self.x itself. That is done by
+        setting self.x = the output of this function.
```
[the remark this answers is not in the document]
pheuer: That doesn't sound that hard - can we just do that now? Just something like 

```python

def _coast_to_plane( ... , xout = None):
    ... 
    xout  = self.x + self.v * t[:, np.newaxis]

    ...
    return xout
```
?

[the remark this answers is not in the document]
StanczakDominik: Yes, but there's a trick to it: you have to do `xout[...] = rhs` or `xout[:] = rhs`. Otherwise, this will overwrite the name - the pointer to the array, if you will - rather than its contents!

plasmapy/diagnostics/proton_radiography.py
```diff
@@ -300,6 +305,212 @@ def _log(self, msg):
...
+            loc_rad = np.sqrt(xloc ** 2 + yloc ** 2)
+            no_hit = np.where(loc_rad > radius, 1, no_hit)
+
+            # In the case of a circular mesh, also create a round wire along the
+            # outside edge
+            no_hit *= np.where(np.isclose(loc_rad, radius, atol=wire_radius), 0, 1)
+
+        # Identify the particles that have hit something, then remove them from
+        # all of the arrays
+        i = np.argwhere(no_hit)[:, 0]
```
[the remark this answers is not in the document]
pheuer: This part doesn't seem to work: isn't `~hit` still a boolean array? I still need to use `np.argwhere` to get the indices.

plasmapy/diagnostics/proton_radiography.py
pheuer: NICE

pheuer (commented):
Ack accidentally started a review somehow?

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
pheuer: Great, that works! Is there a way to avoid doing something like this?

```python
if x is None:
            x = self.x + self.v * t[:, np.newaxis]
else:
            x[:] = self.x + self.v * t[:, np.newaxis]
```

StanczakDominik:
The capabilities of `ack` truly have no bounds!

StanczakDominik:
This should just about work, but please look through to make sure I didn't mess any logical value up!

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
StanczakDominik: Done in a commit below!

plasmapy/diagnostics/proton_radiography.py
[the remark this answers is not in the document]
StanczakDominik: the only simplification possible here is probably...
```python
if x is None:
    # we didn't get any argument; let's preallocate
    x = np.empty_like(self.x) 
x[:] = self.x + self.v * t[:, np.newaxis]
```

StanczakDominik (approved):

StanczakDominik:
LGTM, merge whenever, basically :)

PR #1054, opened 2021-03-03
Example notebook for proton radiography custom source capability

pheuer:
This PR adds an example notebook demonstrating how users can use `SyntheticProtonRadiograph` to create synthetic proton radiographs using custom lists of initialized particles, and how this capability allows the creation of synthetic radiographs with arbitrary source fluence profiles. 

(Previously `load_particles` wasn't used in any of the example notebooks)

This PR also now includes minor changes to the other PRad notebooks too (mostly adding links). 

TODO: 

- [x] Figure out the reason the links aren't working
- [x] Once links are working, add them to the other two PRad notebooks too
- [x] Create explicit tests for `rot_a_to_b()` in mathematics. Requires making a `test_mathematics.py` file.


pheuer:
@rocco8773 Can you take a look at this one?

rocco8773:
@pheuer Yes, I'll take a look over it.  Give me today/tomorrow to get to it. Thanks.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Proton Radiography: Synthetic Radiographs with Custom Source Profiles"
```
rocco8773, line 7: Can you shorten the title by at least 1 word?  Currently the title is so long that it's extending beyond the boundaries of the thumbnail.

image

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "An experimental reality of charged particle radiography experiments is that the source of the particles has finite size and often generates a non-uniform angular velocity distribution. For this reason, the`plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph` particle tracing algorithm allows users to specify their own initial particle positions and velocities. This example will demonstrate how to use this functionality to create a more realistic synthetic radiograph that includes these source profile effects."
```
rocco8773, line 14:
```suggestion
    "[SyntheticProtronRadiography]: ../../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#syntheticprotonradiograph\n",
    "\n",
    "In the real world the particle source for charged particle radiography experiments is finite in size and often has a non-uniform angular velocity distribution. For this reason, the [SyntheticProtronRadiography] particle tracing algorithm allows users to specify their own initial particle positions and velocities. This example will demonstrate how to use this functionality."
```

Sphinx directives do not work in notebooks, so when referencing code documentation we have to used mardown's linking mechanisms.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   "metadata": {},
+   "source": [
+    "### Creating the Initial Particle Velocities"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First we will create the velocity distribution by generating $\\theta$ and $\\phi$ angles for each particle. Generating $\\phi$ is simple: we will choose the azimuthal angles to just be uniformly distributed"
```
rocco8773, line 71: This notebook does not have any mention on how the xyz coords are defined, so we don't have reference for `phi` and `theta`.  I'd also specifically state what `phi` and `theta` correspond to since they are often flipped between different research fields.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "ax.set_title(\"Mean source profile\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
rocco8773, line 501: Delete empty cell at the end of the notebook.  This is often missed because Jupyter notebook automatically adds the cell when you execute the last cell in the notebook.  It's just something we have to be aware of when committing new commits.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "After this point, the rest of the process for creating the radiograph is unchanged."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "nbsphinx-thumbnail": {
+     "tooltip": "Proton Radiography"
+    }
```
rocco8773, line 432: For some reason this thumbnail tagging is not working.  I'll continue looking into it, but two pairs will be quicker than one.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "v0 = np.random.normal(loc=v_cent, scale=1e6, size=int(nparticles))\n",
+    "v0 *= u.m / u.s\n",
+    "\n",
+    "fig, ax = plt.subplots(figsize=(6, 6))\n",
+    "v_per_bin, bins = np.histogram(v0.si.value, bins=100)\n",
+    "ax.plot(bins[1:], v_per_bin)\n",
+    "ax.set_xlabel(\"v0 (m/s)\")\n",
+    "ax.set_ylabel(\"#\")\n",
+    "ax.axvline(x=1.05 * v_cent, label=\"+5%\", color=\"C1\")\n",
+    "ax.axvline(x=0.96 * v_cent, label=\"-5%\", color=\"C2\")\n",
```
rocco8773, line 189:
```suggestion
    "ax.axvline(x=0.95 * v_cent, label=\"-5%\", color=\"C2\")\n",
```

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "vel[:, 0] = v0 * np.sin(theta) * np.cos(phi)\n",
+    "vel[:, 1] = v0 * np.sin(theta) * np.sin(phi)\n",
+    "vel[:, 2] = v0 * np.cos(theta)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you want a velocity distribution centered on another vector, the function `proton_radiography._rot_a_to_b` can create an appropriate rotation matrix."
```
rocco8773, line 216: If you're going to reference the `_rot_a_to_b` functionality, then it should not be private.   Either remove the reference from the notebook or refactor it to `rot_a_to_b` and properly document.

rocco8773 (changes requested):
@pheuer Here's a partial review of the notebook.  The notebook looks to be in good condition, but there are a few things to address.

* As you can see in my first cell comment, Sphinx's normal linking mechanisms don't work in notebooks.  Instead we have to manually create links using markdown's syntax.  Can you go through your notebook and add these links to documentation whenever you're referencing `plasmapy`'s functionality or any external pkgs like `numpy`.
* This notebook is long enough that it might warrant making a table of contents at the top of the notebook.  If you feel you'd like to go this way, take a look at the [Fit Function notebook](https://docs.plasmapy.org/en/latest/notebooks/analysis/fit_functions.html) and/or [Floating Potential notebook](https://docs.plasmapy.org/en/latest/notebooks/analysis/swept_langmuir/find_floating_potential.html).
* Can you please optimize the plot sizes and font sizes?  More so the font sizes.  The font sizes are a little too small for readability in a web browser.  For tips you can again refer to the Fit Function and Floating Potential notebooks.
* [optional] When plotting the particle distributions, it might be easier for users to interpret the plots if they are normalized to the total number of particles in the distribution `N / No / dSigma`.  If you go this route, `No` should be included somewhere in the plot.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
pheuer: Sure, let's try just "Synthetic Radiographs with Custom Source Profiles"

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
pheuer: I find this pretty confusing, can you elaborate? There is no `/api` folder that I can see, nor a `SyntheticProtonRadiograph.rst`, so I assume these are files that are being created when the documentation is being compiled? How do I reference a method like `create_particles`?

Is there a way to get to this reference path from the documentation URL? They look similar...
```
https://docs.plasmapy.org/en/latest/api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.html#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph
```

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
pheuer: I definitely agree, but it's such a general function that I don't think it belongs in `proton_radiography.py`. Where should we put it?

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
pheuer: I'll specify earlier that the source-detector axis is parallel to Z, and I'll explicitly define theta and phi as the polar and azimuthal angles respectively. I could consider a diagram, but that might be overkill?

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
pheuer: Hm, no idea, I checked and this `metadata` is identical to what I have in the main proton radiography notebook that seems to be working.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
rocco8773: Everything in the `docs/api/` folder is auto generated by the `automodsumm` and `autosummary` directives.  That's why you don't see the files until you do a build of the documentation.  The naming scheme of the `api/` files is the same as what we use for the `api_static/` files, `plasmapy.mod.mod.Class`, `plasmapy.mod.mod.func`, etc.  The `#` at the end tells the build you're trying to reference a specific heading in that file.  To reference the whole page then you just exclude the `#` part.

image

---
Here's the [`nbsphinx` documentation on this topic.](https://nbsphinx.readthedocs.io/en/0.8.2/markdown-cells.html#Links-to-Domain-Objects)

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
rocco8773: I'll leave that decision to you, but when it comes to defining coordinates it's always easier to understand the coord system via a diagram opposed to words.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
rocco8773: I did the same and couldn't see why.  I'll keep investigating.  I'm sure we'll find the culprit as we get through the review.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
rocco8773: 🤔 It could possibly be put in `plasmapy.formulary.mathematics`.  I think that would make sense.

pheuer: Sounds good to me!

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
StanczakDominik: Ha, third pair's the charm! At first I thought this was a question of [multiple outputs from a cell](https://nbsphinx.readthedocs.io/en/0.8.2/gallery/multiple-outputs.html), and it kind of is. tqdm is the issue - "stream" output seems to be the on-the-fly progress bar animation. Every text line of the ASCII progress bar seems to be recorded as an output.

To sidestep this, I just split this cell into one for the simulation and one for the plot.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
StanczakDominik: I'll have to quote that comment into the doc guide...

pheuer: I've tried to do this, but I must still be doing something wrong? I have

```
[SyntheticProtronRadiography]: ../../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph

[SyntheticProtronRadiography] 
```

but CI raises this error

```
Warning, treated as error:
/home/runner/work/PlasmaPy/PlasmaPy/docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb:20:File not found: '../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph'
```

pheuer:
Note that I've made a couple minor edits to another tutorial notebook (`proton_radiography_particle_tracing`) here too:
1) Shortened the title so it fits in the box below the thumbnail
2) Fixed a warning caused by the newest Matplotlib version

I'd like to add nice documentation links to that notebook too, once we figure out how to make them work on the source profile notebook...

StanczakDominik:
@pheuer just noting - this looks on track for 0.6 :) is there anything off with the thumbnails, still? That last commit message looks concerning.

StanczakDominik:
Paging @rocco8773 the link king ;)

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
rocco8773: Your link is going back one directory too far, `../../` is sending you back to the repor root directory instead of `/docs/`.  Just delete one `../` and it should work.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,598 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[SyntheticProtronRadiography]: ../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph\n",
+    "\n",
+    "An experimental reality of charged particle radiography experiments is that the source of the particles has finite size and often generates a non-uniform angular velocity distribution. For this reason, the \n",
+    "[SyntheticProtronRadiography] particle tracing algorithm allows users to specify their own initial particle positions and velocities. This example will demonstrate how to use this functionality to create a more realistic synthetic radiograph that includes these source profile effects."
```
rocco8773, line 17:
```suggestion
    "In the real world the particle source for charged particle radiography experiments is finite in size and often has a non-uniform angular velocity distribution. For this reason, the [SyntheticProtronRadiography] particle tracing algorithm allows users to specify their own initial particle positions and velocities. This example will demonstrate how to use this functionality."
```

I still find this lead in a bit clunky.  Every time I get to the middle of the first sentence I have to re-scan the sentence to determine what is trying to be said.

pheuer: I agree, how about this 

In real charged particle radiography experiments, the finite size of the particle source limits the resolution of the radiograph. Realistic sources may also generate a non-uniform angular distribution of particles, superimposing a large scale non-uniform "source profile" on the radiograph. For these reasons, the 
[SyntheticProtronRadiography] particle tracing algorithm allows users to specify their own initial particle positions and velocities. This example will demonstrate how to use this functionality to create a more realistic synthetic radiograph that includes these finite source and source profile effects.

pheuer:
@StanczakDominik @rocco8773 

I think this is ready! All the links are now working. I created the file `test_mathematics.py` in the formulary test folder to write an explicit test for `rot_a_to_b()`.

I think a developer docs page on how to do the links in notebooks is a good idea, since they really improve the documentation but are annoying to create! The learning curve is steep since you have to commit and wait for CI to run to get feedback on whether you're doing it correctly. For the record, here are three examples that finally worked:

For a function
```
[rot_a_to_b()]: ../api/plasmapy.formulary.mathematics.rot_a_to_b.rst#plasmapy.formulary.mathematics.rot_a_to_b
```

For a class
```
[SyntheticProtronRadiography]: ../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph
```

For a method of a class
```
[create_particles()]: 
../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.create_particles
```

For an external website
```
[np.random.choice()]: https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html
```

It would be worth explaining where the `api/` files are created. These paths are relative to the main `notebooks` directory: if the notebook is in a sub folder (eg. `cartesian_grids.ipynb` which is in `plasma`) an additional `../` is required to get back to the `api/` directory.

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_rot_a_to_b():
+
+    # Test one rotation
+    a = np.array([1, 0, 0])
+    b = np.array([0, 1, 0])
+    R = mathematics.rot_a_to_b(a, b)
+    correct = np.array([[0, 1, 0], [-1, 0, 0], [0, 0, 1]])
+    assert np.allclose(R, correct, atol=1e-6)
```
StanczakDominik, line 19:
```suggestion
    np.testing.assert_allclose(R, correct, atol=1e-6)
```
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html has a bit nicer output when something breaks, so i like it for array comparisons

StanczakDominik (commented):
Looked through non notebook changes; they look great. I'll leave the final call to @rocco8773, though (unless you're short on time, then I can continue here).

rocco8773:
I got a partial review in the works, nothing too crazy. However, I ran out of steam tonight, so I'll post in the morning.

StanczakDominik:
Sleep well!

rocco8773 (changes requested):
This is a review of the custom source notebook.  All in all it looks good and I've confirmed the links are working.   All my suggestions focus on readability and clarity.  I just felt a few spots here and there could be refined a little more to make it easier for user to understand the notebook.

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
         return integral_arr
     else:
         raise TypeError(f"Improper type {type(x)} given for argument x.")
+
+
+def rot_a_to_b(a, b):
```
rocco8773, line 97:
```suggestion
def rot_a_to_b(a: np.ndarray, b: np.ndarray) -> np.ndarray:
```

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
...
+    r"""
+    Calculates the 3D rotation matrix that will rotate vector a to be aligned
+    with vector b.
```
rocco8773, line 100:
```suggestion
    Calculates the 3D rotation matrix that will rotate vector ``a`` to be aligned
    with vector ``b``.
```

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
...
+
+    Parameters
+    ----------
+    a : ~np.ndarray, shape (3)
+        Pre-rotation orientation unit vector
```
rocco8773, line 105:
```suggestion
    a : `~numpy.ndarray`, shape (3,)
        Vector to be rotated.  Should be a 1D, 3-element unit vector.  If ``a``
        is not normalize, then it will be normalized.
```

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
...
+
+    b : ~np.ndarray, shape (3)
+        Post-rotation orientation unit vector
```
rocco8773, line 108:
```suggestion
    b : `~numpy.ndarray`, shape (3,)
        Vector representing the desired orientation after rotation.  Should be
        a 1D, 3-element unit vector.  If ``b`` is not normalized, then it will
        be.
```

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
...
+
+    Returns
+    -------
+    R : ~np.ndarray, shape(3,3)
+        The rotation matrix that will rotate vector a to be parallel to
+        vector b.
```
rocco8773, line 114:
```suggestion
    R : `~numpy.ndarray`, shape (3,3)
        The rotation matrix that will rotate vector ``a`` onto vector ``b``.
```

plasmapy/formulary/mathematics.py
```diff
@@ -92,3 +92,43 @@ def Fermi_integral(
...
+
+
+    The algorithm is based on `this discussion <https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d/476311#476311>` on StackExchange:
```
rocco8773, line 117:
```suggestion
    Notes
    -----
    The algorithm is based on
    `this discussion <https://math.stackexchange.com/a/476311>`_ on StackExchange:
```

Note, for linking in `sphinx` you need to have the trailing underscore `` `label <target>`_ ``.

I think you should also include the associated math in this "Notes" section.

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,40 @@
...
+    R = mathematics.rot_a_to_b(a, b)
+    correct = -np.identity(3)
+    assert np.allclose(R, correct, atol=1e-6)
+
+
+if __name__ == "__main__":
+    """
+    test_rot_a_to_b()
+    """
+    pass
```
rocco8773, line 40:
```suggestion
```
This is not needed.

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,40 @@
...
+    R = mathematics.rot_a_to_b(a, b)
+    correct = np.identity(3)
+    assert np.allclose(R, correct, atol=1e-6)
+
+    # Test the special case where the two vectors are anti-parallel
+    a = np.array([1, 0, 0])
+    b = np.array([-1, 0, 0])
+    R = mathematics.rot_a_to_b(a, b)
+    correct = -np.identity(3)
+    assert np.allclose(R, correct, atol=1e-6)
```
rocco8773, line 33: In practice you want to write your tests so if one fails, then the other will still run.  In how you have the test written here, if the first test fails then the last two never run.  The best way to accomplish this is to use [`pytest.mark.parametrize`](https://docs.pytest.org/en/stable/parametrize.html), which would look a little like...

```python
@pytest.mark.parametrize(
    "a, b, correct",
    [
        # case 1
        (a, b, correct),
        # case 2
        (a, b, correct),
        ...
    ],
)
def test_rot_a_to_b(a, b, correct):
    R = rot_a_to_b(a, b)
    assert np.allclose(R, correct)
```

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
+import pytest
+
+from plasmapy.formulary import mathematics as mathematics
+
+# TODO: Move the Fermi integral tests over to this file?
+
+
+def test_rot_a_to_b():
```
rocco8773, line 12: When you add the shape conditioning for `a` and `b`, you'll also need to add a test for it.  I would typically to this like....

```python
@pytest.mark.parametrize(
    "a, b, _raises",
[
        # case 1
        (a, b, ValueError),
        # case 2
        (a, b, ValueError),
        ...
    ],
)
def test_rot_a_to_b_raises(a, b, _raises):
    with pytest.raises(_raises):
        rot_a_to_b(a, b)
```

rocco8773 (changes requested):
This is a review of `rot_a_to_b` and its tests.

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,598 @@
...
+    "ax.set_title(\"Synthetic Proton Radiograph\", fontsize=14);"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[synthetic_radiograph()]: ../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.synthetic_radiograph\n",
+    "\n",
+    "Calling the [synthetic_radiograph()] function with the `ignore_grid` keyword set will display the source profile (the radiograph calculated in the absence of any grid fields)."
```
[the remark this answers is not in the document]
pheuer: Made a slight variation: 
```
Calling the [synthetic_radiograph()] function with the `ignore_grid` keyword will produce the synthetic radiograph corresponding to the source profile propagated freely through space (i.e. in the absence of any grid fields).
```

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,598 @@
...
+    ")\n",
+    "\n",
+    "fig, ax = plt.subplots(ncols=2, figsize=(12, 6))\n",
+    "ax[0].pcolormesh(xpos, zpos, hist.T)\n",
+    "ax[0].set_xlabel(\"x ($\\\\mu m$)\", fontsize=14)\n",
+    "ax[0].set_ylabel(\"z ($\\\\mu m$)\", fontsize=14)\n",
+    "\n",
+    "ax[1].pcolormesh(xpos2, ypos, hist2.T)\n",
+    "ax[1].set_xlabel(\"x ($\\\\mu m$)\", fontsize=14)\n",
+    "ax[1].set_ylabel(\"y - $y_0$ ($\\\\mu m$)\", fontsize=14)"
```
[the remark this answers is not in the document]
pheuer: Sure, done

docs/notebooks/proton_radiography_particle_tracing_custom_source.ipynb
```diff
@@ -0,0 +1,598 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "with warnings.catch_warnings():\n",
+    "    warnings.simplefilter(\"ignore\")\n",
+    "    sim = prad.SyntheticProtonRadiograph(grid, source, detector, verbose=True)"
```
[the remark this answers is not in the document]
pheuer: That's just one of the default printouts from `run()`: I agree it's not really relevant. How about I just set `verbose=False` and hide all of that?

pheuer:
@rocco8773 Great, thanks for the review. I think I've addressed all your comments

plasmapy/formulary/mathematics.py
```diff
@@ -88,3 +88,77 @@ def Fermi_integral(
...
+                v_3 & 0 & -v_1 \\
+                -v_2 & v_1 & 0
+            \end{bmatrix}
+
+    Note that this algorithm fails when :math:`1+c=0`, which occurs when :math:`a` and
+    :math:`b` are anti-parallel. However, since the correct rotation matrix
+    in this case is simply :math:`R=-I`, this function just handles this
+    special case explicitly.
+
+    This algorithm is based on `this discussion <https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d/476311#476311>`_ on StackExchange:
```
rocco8773, line 129:
```suggestion
    The algorithm is based on
    `this discussion <https://math.stackexchange.com/a/476311>`_ on StackExchange.
```
Use the shorter link so the docstrings stay within our max line width.

plasmapy/formulary/mathematics.py
```diff
@@ -88,3 +88,77 @@ def Fermi_integral(
...
+        The rotation matrix that will rotate vector ``a`` onto vector ``b``.
+
+    """
+
+    # Normalize both vectors
+    a = a / np.linalg.norm(a)
+    b = b / np.linalg.norm(b)
+
+    # Manually handle the case where a and b point in opposite directions
+    if np.dot(a, b) == -1:
```
rocco8773, line 155: It's not sufficient to rely on `np.dot` to condition `a` and `b` since it will still allow any N-element arrays to pass as long as `a.shape == b.shape`.  The functionality you've coded here only works for 1D, 3-element arrays so you need a workflow like...

```python
a = np.squeeze(a)  # will convert list to array and reduce any single element dimensions
if a.shape != (3,):
    raise ValueError("give message")
a = a / np.linalg.norm(a)
```

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,43 @@
...
+    np.testing.assert_allclose(R, correct, atol=1e-6)
+
+
+@pytest.mark.parametrize(
+    "a, b, _raises",
+    [
+        # a is the wrong length
+        (np.array([1, 0]), np.array([1, 0, 0]), ValueError),
+        # b is the wrong length
+        (np.array([1, 0, 0]), np.array([1, 0]), ValueError),
```
rocco8773, line 38: Add a case where both are the wrong length.

plasmapy/formulary/mathematics.py
pheuer: Ok, done

plasmapy/formulary/tests/test_mathematics.py
pheuer: Done

plasmapy/formulary/mathematics.py
```diff
@@ -88,3 +88,86 @@ def Fermi_integral(
...
+    R : `~numpy.ndarray`, shape (3,3)
+        The rotation matrix that will rotate vector ``a`` onto vector ``b``.
+
+    """
+
+    # Normalize and validate both vectors
+
+    a = np.squeeze(a)
+    if a.shape != (3,):
+        raise ValueError(f"a must have shape (3,) but input has shape {a.shape}.")
```
rocco8773, line 155:
```suggestion
        raise ValueError(
            f"Argument 'a' must have shape (3,) but input has shape {a.shape}."
        )
```

plasmapy/formulary/mathematics.py
```diff
@@ -88,3 +88,86 @@ def Fermi_integral(
...
+    a = a / np.linalg.norm(a)
+
+    b = np.squeeze(b)
+    if b.shape != (3,):
+        raise ValueError(f"b must have shape (3,) but input has shape {b.shape}.")
```
rocco8773, line 160:
```suggestion
        raise ValueError(
            f"Argument 'b' must have shape (3,) but input has shape {b.shape}."
        )
```

plasmapy/formulary/tests/test_mathematics.py
```diff
@@ -0,0 +1,45 @@
...
+        # both are the wrong length
+        (np.array([1, 0, 0, 0]), np.array([1, 0]), ValueError),
```
rocco8773, line 40: Do both are wrong length but equal length.  This is the case that should have failed, but would have passed if you were just depending on `np.dot`.

plasmapy/formulary/mathematics.py
```diff
@@ -88,3 +88,90 @@ def Fermi_integral(
...
+
+    Note that this algorithm fails when :math:`1+c=0`, which occurs when :math:`a` and
+    :math:`b` are anti-parallel. However, since the correct rotation matrix
+    in this case is simply :math:`R=-I`, this function just handles this
+    special case explicitly.
+
+    This algorithm is based on
+    `this discussion <https://math.stackexchange.com/a/476311>`_ on StackExchange.
+
+
```
rocco8773, line 132:
```suggestion
```

Delete extra space.  Sorry for throwing all this little things at you.  Just trying to catch it now so we don't have to come back.

plasmapy/formulary/tests/test_mathematics.py
pheuer: Got it- done

docs/notebooks/proton_radiography_particle_tracing.ipynb
```diff
@@ -244,6 +257,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "[create_particles()]: ../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.create_particles\n",
+    "\n",
+    "\n",
```
rocco8773, line 262:
```suggestion
```

This link is never used so you can delete it.

changelog/1054.doc.rst
```diff
@@ -0,0 +1 @@
+Created an example notebook demonstrating how the `SyntheticProtonRadiograph` class can be used to generate synthetic proton radiographs with arbitrary source profiles.
```
rocco8773, line 1:
```suggestion
Created an example notebook demonstrating how the
`~plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph` class can be
used to generate synthetic proton radiographs with arbitrary source profiles.  Add
code documentation links to all proton radiograph notebooks.
```

This will allow `intersphinx` to properly generate the link for `SyntheticProtonRadioraph`.

rocco8773 (changes requested):
Thanks for sticking with this.  Everything looks really good!

That last thing that's needed is a change log entry for `rot_a_to_b`, so create a `1054.feature.rst` for this.

pheuer:
No problem - change log entry created!

rocco8773 (approved):
All looks good.  I'm going to wait for the checks to pass and then I'll merge. Thanks for the hard work!

PR #1059, opened 2021-03-10
[VERY EARLY WIP] Neoclassics

StanczakDominik:
This branch is absolutely not ready for most things, but it's getting there. Much like myself, in a way. Figured I'd practice what I preach and work in the open, though. I'll keep breaking off bits of functionality that are reviewable into smaller PRs.


pheuer:
@StanczakDominik I noticed you're storing data in `.csv` files for this. I don't know if performance is a concern at all here, but [this article](https://towardsdatascience.com/what-is-npy-files-and-why-you-should-use-them-603373c78883) makes a compelling case for using `.npy` files instead.

StanczakDominik:
The .csv's are actually test files generated by https://pytest-regressions.readthedocs.io/en/latest/api.html#pytest_regressions.num_regression.NumericRegressionFixture.check so it's not quite myself picking the format here :D so it's not quite myself picking the format here. If I had any larger data to store, I would likely indeed use npy files.

PR #1060, opened 2021-03-10
Put back import of __version__ from plasmapy in conf.py

rocco8773:
Using `pkg_resources.get_distribution("plasmapy")` severely limits the use of `make html` to build documentation.  This forces `plasmapy` to be installed when building docs versus just being in the python path.  The later is my preferred workflow since editable installs don't properly reflect version increments unless you reinstall (version is frozen in the EGG info).  I don't want to reinstall `plasmapy` for each documentation build.

https://github.com/PlasmaPy/PlasmaPy/blob/d9f3b197fbaca9e875ce5ea34e79ef7aeb020159/plasmapy/__init__.py#L37-L42

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ from sphinx.application import Sphinx
 
 sys.path.insert(0, os.path.abspath(".."))
 
-from plasmpy import __version__ as release
+from plasmapy import __version__ as release
 
 # -- General configuration ------------------------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,9 +27,7 @@ from sphinx.application import Sphinx
 
 sys.path.insert(0, os.path.abspath(".."))
 
-from pkg_resources import get_distribution
-
-release = get_distribution("plasmapy").version
+from plasmpy import __version__ as release
 
 # -- General configuration ------------------------------------------------
 


StanczakDominik (approved):
I'm a bit torn on this. The pkgtools solution [is the recommended workflow as per setuptools-scm](https://github.com/pypa/setuptools_scm#usage-from-sphinx), and it seems to me that you wouldn't really need to reinstall on each build unless you actually wanted the version tag to be pitch-perfect correct on each build, which seems to me like overdoing it.

Also, relevant xkcd:

1172: Workflow

But I don't mind it either way too much, so long as nothing breaks.

StanczakDominik:
Okay, that sounds better, but I do have a question:

> I don't like pkg metadata being installed in my development repo (which editable installs do) because it tends to cause unexpected issue. I've been caught off guard by this several times in the past

If you don't install it, how do you use PlasmaPy from outside the repository directory? Do you literally install the release version?

rocco8773:
> If you don't install it, how do you use PlasmaPy from outside the repository directory? Do you literally install the release version?

I have a "development" virtual environment that when activated automatically adds the `plasmapy` and `bapsflib` repo paths to my `PYTHONPATH`, and removes the paths when I deactivate the venv.  I find this setup up propagates changes much quicker for prototyping work.

StanczakDominik:
Well, if you're happy with that :D

PR #1058, opened 2021-03-09
Minor docstring fixes in formulary

namurphy:
Here's a PR of some minor fixes mostly related to reStructuredText formatting that I've been meaning to work on a while now.  My main motivation is that there are some exceptions and warnings that show up like **~astropy.whatever.SomeError** rather than as a linked object.  I might make a few minor PEP 8 fixes here and there too, such as for variable names.

I don't know if we've settled on conventions regarding the specific formatting of references, warnings, etc.  I'm mostly concerned about consistency and links here, so I'm open to different conventions.

PR #1062, opened 2021-03-11
Add pydocstyle to CI in the flake8 runner; start fixing flake8 being effectively disabled

StanczakDominik:
This is a draft pull request inspired by @Tiger-Du's recent massive review of formulary changes https://github.com/PlasmaPy/PlasmaPy/pull/1058#pullrequestreview-609352996. I looked at it and figured hey, there must be a way to point these out automatically.

So I used pydocstyle, for which - funnily enough - we have configuration in setup.cfg:

https://github.com/StanczakDominik/PlasmaPy/blob/cae0a8387159a3f4a411a6843991edba3687f1e2/setup.cfg#L107-L115

 but which does not currently fix issues in docstrings (see https://github.com/PyCQA/pydocstyle/issues/249). So this should be taken as nothing but an overview of how much goes wrong with our docstrings right now. I also suspect this should not actually run in pre-commit, but on the side in CI like flake8 does. I'd rather reserve pre-commit for stuff that can happen automagically, so it doesn't slow down committing. We want to promote frequent commits, after all!


StanczakDominik:
The error output is from line 84 to 2377, so that's about... 1146 little docstring errors and inconsistencies.

namurphy:
Thank you, this is great!  I remember really wanting a pydocstyle check ages ago, long before the days of GitHub Actions.  I'm thinking it would be helpful to not require this for merging until we get the pydocstyle configuration stabilized (since our style for docstrings has evolved) and fix the existing issues.

StanczakDominik:
Yup, completely agreed, but https://github.com/actions/toolkit/issues/399 github actions doesn't yet have an "allow failure" setting. So this might take a while.

I also started messing with the flake8 config we had (turns out it was getting overridden), and we will first have to make sure that doesn't complain on stuff black accepts. Probably gonna have to limit it to quick checks for unfound variables etc...

namurphy:
I'd be inclined to drop rule `D413` from `add-select` in the `pydocstyle` settings for `setup.cfg`.  This rule requires that there be an extra blank line before the end of the docstring (i.e. a blank line between the end of the `Examples` section and the `"""`).  We've tended to not have that extra blank line in there.  With that said, I'm okay either way — so long as it's done consistently!  💤

StanczakDominik:
Okay. To summarize this long journey:

* I don't think fixing all the errors raised by pydocstyle and flake8 is in scope for this pull request. Thus, I'm **adding them to CI, but disabling everything** that they complained about in `setup.cfg`. Iteratively enabling them again them is a great idea for first contributions or PDD.
* pydocstyle actually provides a flake8 plugin called flake8-docstrings. This means we could have **a single linter** for those two, and the pydocstyle checks would be run from flake8. I decided to go with this approach.
    * This would be preferable because the flake8 plugin ecosystem looks rich as heck. https://pypi.org/project/flake8-rst-docstrings/ in particular looks handy - this checks and points out invalid rst formatting in docstrings, which would be much easier to find that way than by iteratiely checking docstrings. So I've gone ahead and added that.
* pydocstyle by default ignores test files. I've thus excluded test files from flake8 for now, as well, because it was complaining about docstring errors in tests a lot. This should be removed in the future.

This is all I have strength to do for this PR, I think :sweat_smile:

setup.cfg
```diff
@@ -90,43 +90,96 @@ looponfailroots =
...
+    D302,
+    # D205 and D400 are ignored to allow the "one-liner" to exceed one
+    # line, which is sometimes necessary for even concise descriptions of plasma
+    # physics functions and classes.
+    D400,
+    # TODO D401: First line should be in imperative mood
+    D401,
+    # D403: First word of the first line should be properly capitalized ('Latex', not 'LaTeX') # Dominik: good job, pydocstyle /s
+    D403,
+    # TOOD D407: Missing dashed underline after section ('References')
```
namurphy, line 133:
```suggestion
    # TODO D407: Missing dashed underline after section ('References')
```

namurphy (approved):

PR #1064, opened 2021-03-12
Minor docstring fixes in particles

namurphy:
Here are a bunch of more docstring fixes using my new favorite programming paradigm of PDD.¹  This time these are for `plasmapy.particles`, and were primarily inspired by running `pydocstyle`.  If this looks good, I'm okay with it being merged.

I also renamed two private dictionaries to be more PEP8ish: `_Elements` to `_elements` and `_Isotopes` to `_isotopes` (since these aren't classes).  

¹Procrastination-driven development.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -35,7 +35,7 @@ from plasmapy.particles.exceptions import (
     ParticleError,
     ParticleWarning,
 )
-from plasmapy.particles.isotopes import _Isotopes
+from plasmapy.particles.isotopes import _isotopes
 from plasmapy.particles.parsing import (
     _dealias_particle_aliases,
     _invalid_particle_errmsg,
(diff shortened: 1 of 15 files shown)


plasmapy/particles/particle_class.py
```diff
@@ -108,29 +108,29 @@ def _category_errmsg(particle, category: str) -> str:
...
         """
-        A dictionary representation of the particle object that is JSON friendly
-        (i.e. convertible to a JSON object).
+        Provide a dictionary representation of the particle object that
+        is JSON friendly (i.e. convertible to a JSON object).
 
         The dictionary should maintain the following format so
-        `~plasmapy.particles.ParticleJSONDecoder` knows how to decoded the resulting
-        JSON object.
+        `~plasmapy.particles.ParticleJSONDecoder` knows how to decoded
```
StanczakDominik, line 132: Note to self- I've got a great example for hypothesis testing here. Since I'm using that in neoclassics it's gonna be a handy introduction.

StanczakDominik (approved):
I have mixed feelings about whatever docstyle directive adds "returns..." to properties. Properties are supposed to mirror the behavior of fields; the end user is not going to be aware they're actually using a method. Meanwhile calling help on that property will show "returns..."

I'm dead tired but I think you see my point?

But I'm also too tired to mind much!

namurphy:
> I have mixed feelings about whatever docstyle directive adds "returns..." to properties.

Yeah...I had mixed thoughts about that too.  It's a good practice for functions and methods, I think, but not as well suited to properties.  I'll undo those particular changes, and then we could merge this.

PR #1070, opened 2021-03-13
Review changelog for v0.6.0 release

rocco8773:
This is a review of the change log entries from our last release and ahead of the v0.6.0 release.

PR #1072, opened 2021-03-13
Made `is_uniform` a documented public attribute of AbstractGrid (closes #1071)

### LINKED ISSUE #1071: Document attribute `is_uniform` for AbstractGrid
@pheuer It appears that the `is_uniform` attribute on `AbstractGrid` is semi-public.  How do you feel about making this an actual property of the the class so it appears in the RTD documentation?

pheuer:
Properly documented this property

### DIFF
--- a/plasmapy/plasma/grids.py
+++ b/plasmapy/plasma/grids.py
@@ -243,7 +243,7 @@ class AbstractGrid(ABC):
         return u.Quantity(self.ds[key].data, self.ds[key].attrs["unit"], copy=False)
 
     @property
-    def is_uniform(self):
+    def is_uniform(self) -> bool:
         """
         A boolean value reflecting whether or not the grid points are
         uniformly spaced.
--- a/plasmapy/plasma/grids.py
+++ b/plasmapy/plasma/grids.py
@@ -248,13 +248,6 @@ class AbstractGrid(ABC):
         A boolean value reflecting whether or not the grid points are
         uniformly spaced.
         """
-
-        if self._is_uniform is None:
-            raise ValueError(
-                "The `is_uniform` attribute is not accessible "
-                "before a grid has been loaded."
-            )
-
         return self._is_uniform
 
     @property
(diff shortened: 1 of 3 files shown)


pheuer:
@rocco8773 Here is the implementation

```
@property
    def is_uniform(self):
        """
        A boolean value reflecting whether or not the grid points are
        uniformly spaced.
        """

        if self._is_uniform is None:
            raise ValueError(
                "The `is_uniform` attribute is not accessible "
                "before a grid has been loaded."
            )

        return self._is_uniform
```

However, it is physically impossible to access the `if self._is_uniform is None`. The private variable is set in `load_grid`, which gets called as part of any of the possible outcomes of `__init__`. So, maybe we can drop the error check (since I can't test it...)?

rocco8773:
If you add...

```python
if self._is_uniform is None:   # coverage: ignore
    ...
```

then he the coverage statistics should ignore this block.  Since this scenario should never happen and is non-trivial to test, I would have no problem doing this.  In reality, you should be able to test the scenario using the `unittest.mock` functionality.

plasmapy/plasma/grids.py
```diff
@@ -241,6 +242,14 @@ def __getitem__(self, key):
         """
         return u.Quantity(self.ds[key].data, self.ds[key].attrs["unit"], copy=False)
 
+    @property
+    def is_uniform(self):
```
namurphy, line 246:
```suggestion
    def is_uniform(self) -> bool:
```
Might be helpful to add a type hint annotation.

namurphy (approved):
Looks good!  I have one suggestion, though this looks pretty straightforward.

pheuer:
@rocco8773 I'm done, so you can merge this whenever you're ready

rocco8773:
@pheuer Ok, I'll get to it in a bit.  My priority right now is to get #1070 ready for review and merge.

rocco8773 (approved):
Looks good to me! Thanks!

PR #1077, opened 2021-03-14
Release version 0.6.0

StanczakDominik:
What it says on the tin!


StanczakDominik:
Will continue soon, brb, doggo!

StanczakDominik:
All right, this should be good to go for a minor review. Definitely skip c1425e0. I think we could squash merge this one and see what happens; the tag should still be there, so I don't think anything will break.

Once that's merged, I'll try adding a v0.7.dev tag to master, so that the changelog entry works nicely. This will take an additional change to CI, which I'm adding in 31684cce.

docs/development/release_guide.rst
```diff
@@ -73,7 +75,9 @@ Release
 
 * Make sure that tests pass and that documentation builds without issue.
 
-* No, really, check twice.
+  * No, really, check twice. Let the tests do their thing. You want things tip
+    top, and by now, you want that cuppa tea anyway. Treat yourself! Celebrate
+    the new release and let the darn tests pass.
```
rocco8773, line 80: ☕️☺️

rocco8773 (approved):
Looks good to me! Thank you for all the hard work!

PR #1080, opened 2021-03-16
Add github action to automatically label PRs

StanczakDominik:
This adds https://github.com/actions/labeler. PRs that modify, say, plasmapy.particles should now automatically get the Particles label. I mirrored the existing labels; we can add more if needed.

I'm not 100% sure I've got the YAML set up properly, but I don't want to duplicate our labels on my own fork, so I'll have to adjust things here afterwards, anyway.

This PR is more of a "do we want this cool new thing that astropy's already trying out?". It's also a prelude to [finally adding the new contributor greeter bot, which can be done here in a basic form](https://github.com/PlasmaPy/PlasmaPy/new/master?filename=.github%2Fworkflows%2Fgreetings.yml&workflow_template=greetings) but I need to write some concise greetings first.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,17 @@
+Diagnostics: plasmapy/diagnostics/**/*
+
+Formulary: plasmapy/formulary/**/*
+
+Mathematics: plasmapy/formulary/mathematics.py
+
+parameters.quantum: plasmapy/formulary/quantum.py
+
+Particles: plasmapy/particles/**/*
+
+Documentation: docs/**/*
+
+Plasma class: plasmapy/plasma/**/*
+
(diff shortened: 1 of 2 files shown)


namurphy (approved):
This pull request makes me happy!  I don't know GitHub Actions well enough to understand if the syntax is right, but as you said, it is no problem to update this after.  

Some labels that seem good to add would be for:
 - `plasmapy.analysis`
 - `plasmapy.tests.helpers`
 - Transport 🚆 ✈️ 🚀 ⛵ 🚗 🚌 🛵 

Also...there's an electron emoji on here now! :electron:

StanczakDominik:
Let's try it out, then :)

PR #1083, opened 2021-03-17
Various administrative cleanups

StanczakDominik:
This, done along #1059, cleans up a bunch of left-over minor issues in the repo and removes a couple warnings. Take a look at the commit messages for details.

I also suspect these are trivial enough for changelog entries to be unnecessary. Correct me if I'm wrong!

I suspect I'll have a few more of these PRs in the works along work on neoclassical transport. Look to their coming, given the timezones of most of you, at dawn, look to the east.


StanczakDominik:
... good bot, kind of? :D

StanczakDominik:
Well, that's probably one issue with the code owners stuff. It can't tell whether we're making a huge change to Particles or whether we're just removing a warning...

namurphy (approved):
LGTM.  All pretty straightforward.

From my own view of changelogs, I don't think this rises to the level of needing a change log entry.

namurphy:
> Well, that's probably one issue with the code owners stuff. It can't tell whether we're making a huge change to Particles or whether we're just removing a warning...

Oh...that's a good point.  There will probably still be some need for us to adjust things manually, but I think the code owners and automatic labeling of stuff will be a closer approximation to what we want it to be than what we were doing before.  So, it's at least a statistical improvement.

StanczakDominik:
Thanks for the quick look-through! :)

namurphy:
> Thanks for the quick look-through! :)

Not a problem!  Always happy to engage in procrastination-driven code review.

PR #1088, opened 2021-03-26
Trying to fix very minor error in ExB_drift.ipynb

pheuer:
Started getting a CI error related to the thumbnail in this notebook. Trying to fix it in this PR just by splitting up some cells.

The error:
```
/home/runner/work/PlasmaPy/PlasmaPy/docs/notebooks/ExB_drift.ipynb:"nbsphinx-thumbnail": Unsupported output type in cell 11/output 0: "stream"
```

Relevant documentation: https://nbsphinx.readthedocs.io/en/latest/code-cells.html#Plots

### DIFF
--- a/changelog/1088.trivial.rst
+++ b/changelog/1088.trivial.rst
@@ -1,0 +1,1 @@
+Fixed minor error in `ExB_drift.ipynb`
--- a/changelog/1088.trivial.rst
+++ b/changelog/1088.trivial.rst
@@ -1,1 +1,1 @@
-Fixed minor error in `ExB_drift.ipynb`
+Fixed a minor error in the ExB drift notebook.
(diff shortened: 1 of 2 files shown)


pheuer:
Found a solution! Whatever changed in nbsphinx seems to like 

```python
fig = plt.figure(figsize=(6, 6))
ax = fig.add_subplot(111, projection="3d")
```

more than

```python
fig = plt.figure()
ax = fig.gca(projection="3d")
```

Probably only a problem when making 3D plots with matplotlib, which is why we aren't seeing this error everywhere?

StanczakDominik (approved):
If it works, it works! I don't think that's a nbsphinx issue; there have been no recent release there. Matplotlib, on the other hand, [did release 3.4.0 just yesterday](https://pypi.org/project/matplotlib/#history). The gca syntax felt wonky, anyway, so thanks for this cleanup!

PR #1094, opened 2021-04-01
Add stale bot for PR

### LINKED ISSUE #1092: Implement a 🤖 to close stale pull requests
Pull requests that have not had any activity for the past year are probably not going to be finished...and in any case the main branch will probably have had lots of changes in that time period.  I propose creating a bot that will automatically close pull requests that haven't been touched in the last year.  My main reason for suggesting a bot is to use an emoji 🤖 , but a GitHub Action would probably be a good choice too.

I'm less inclined to close issues automatically, since we have requests for large features from our early days that are still relevant.

This is low priority for the time being, but would be more helpful as the contributor community grows.

StanczakDominik:
Closes #1092

Configured as per https://github.com/marketplace/actions/close-stale-issues.

I went with a yearly cutoff; after that, it'll comment that whoa, what's *not* going on here, and then it'll wait 2 more weeks and close the issue.

I do not think this locks discussion by default. If it does and it's just not documented, I'm just going to purge it.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,8 +10,8 @@ jobs:
       - uses: actions/stale@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          stale-pr-message: 'This PR is stale because it has been open for a year with no activity. Remove stale label or comment or this will be closed in 14 days.'
-          close-pr-message: 'This PR was closed because it has been stalled for a year with no activity.'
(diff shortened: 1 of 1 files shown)


namurphy (approved):
This looks great, and thank you for getting to it so quickly!  I have one optional rewording suggestion, though it's no problem to merge it as is.

StanczakDominik:
Thanks for the quick look-through! Let's see if it works :)

PR #1096, opened 2021-04-01
ParticleTracker mechanics refactoring (again!)

StanczakDominik:
It's officially Q2 of 2021 and I already feel the roadmap backlog pressure (thank you once again, @rocco8773 :wink: ), so I'm starting out with the particle tracker refactoring again. I brought over changes from #675 and made them at least a little more self contained. At this point I have a bad reimplementation of a terrible first implementation.

What I intend to do with this from now on:

- [ ] assimilate the changes the wise @pheuer made in `diagnostics/proton_radiography.py` (some great ideas and an actual use case! that really helps flesh out a code :laughing: )
- [ ] make this code actually nice
- [ ] allow using this with the Grids classes; probably refactor some old cruft out (looking at you, Plasma3D, your names would be numbered if I had an ETA for this)
- [ ] make the current proton_radiography use the ParticleTracker class for the actual motion within EM fields part.
- [ ] Merge the ExB drift notebook with the one I've got for this, since the implementation is actually starting to look nice

I also recently had the idea of using google/jax with this. I'm not sure how much of an issue (as usual...) fields calculations would be with that; but I'd like to at least try. Automatically running this on GPUs would be pretty sweet.


pheuer:
Sounds great, looking forward to this! Looking for any easy optimization opportunities in `grids.py` is on my to-do list. We can definitely change or add functionality to `grids` as you need it. 

A few comments/ideas on top of your list above: 

- I would aim for three output modes: full trajectories, partial trajectories (eg. every N time steps), and only the final positions. It would also be good to allow storing trajectories in some way other than memory (eg. an ascii or hdf5 file) which would be necessary for all but trivial calculations. 

- It would be nice for users to be able to provide their own "stop condition" functions (in addition to a few built-in options). I'm thinking the user could pass in a func that takes the particle tracker as its only argument and return a boolean based on whatever internal logic it likes.

- In order to play nicely with `grids`, we should refactor the particle integrators to take arrays of field scalars rather than vectors (eg. Ex, Ey, Ez instead of E). That would avoid doing this on every push:

```python
 # Create arrays of E and B as required by push algorithm
E = np.array([Ex.to(u.V / u.m).value, Ey.to(u.V / u.m).value, Ez.to(u.V / u.m).value] )
E = np.moveaxis(E, 0, -1)
B = np.array([Bx.to(u.T).value, By.to(u.T).value, Bz.to(u.T).value])
B = np.moveaxis(B, 0, -1)
```

- Related to that last point, we should create methods for `grids` that return values without fetching and applying units (just to have the units then stripped off again before going into the particle integrator). For example, the interpolators should have an option to not apply units before returning.

- Don't forget we also have #979 to wrap up at some point!

In addition, here's some longer-term brainstorming. 

- Eventually I'd really like to find (or make) 3D interpolators that are faster than the SciPy ones, and that are parallelizable and numba-compatible. 

- Another idea I've been thinking about is extending `grids` to allow a time axis. This would allow, for example, proton radiographs of fields that evolve on the scale of the proton crossing time (capturing motion blurring, for example). I'd guess this would be helpful for particle tracker too. 

\<sarcasm\> I don't think anyone other than cryptocurrency miners are able to get GPUs anymore. Maybe it would be better to optimize around some more ubiquitous architecture. Maybe smart speakers.\</sarcasm\>

StanczakDominik:
Just don't look at the code yet! :D

pheuer:
@StanczakDominik what's the status of this PR in light of #2245?

Issue #1090, opened 2021-03-27
No `astropy.units.Quantity`ies within actual body of formulary functions?


StanczakDominik:
Crossposted from #1089:

Thank you for trying this! It is good to get some concrete numbers on how much this could improve performance. However, I am still concerned that this current implementation doesn't go far enough. I ran my own execution time tests and replicated your findings:  with validations takes ~ 1 ms, while with the new  flag that's down to 400 us. Using  input and the  keyword, I got that down to 240 us.

The problem is that the simplest possible version of this function executes in only 1 us! 

```python
import numpy as np
from astropy.constants import k_B
k_B = k_B.si.value


_coefficients = {
    1: {"most_probable": 0, "rms": 1, "mean_magnitude": 2 / np.pi},
    2: {"most_probable": 1, "rms": 2, "mean_magnitude": np.pi / 2},
    3: {"most_probable": 2, "rms": 3, "mean_magnitude": 8 / np.pi},
}

def fast_thermal_speed(T, m, ndim, method):
    coeff = _coefficients[ndim][method]
    return np.sqrt(coeff * T / m)
```

The same is true for  and probably many of the other formulary functions. **In my view, in order to be viable for actual computational work, the  functions (in their "fast mode") need to be within a factor of 2-3x times the execution time of this simplest implementation.**  Otherwise, how can we possibly produce codes that can compete in the software ecosystem? Poorly documented, sloppy code will always win if it's 200x faster (or 10x faster, for that matter).

Here's some  output of the new :

```python
Timer unit: 1e-07 s
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   668      1000      44700.0     44.7      2.3      m = mass if np.isfinite(mass) else particles.particle_mass(particle)
   669                                           
   670                                               # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
   671      1000       4999.0      5.0      0.3      try:
   672      1000       5941.0      5.9      0.3          coef = _coefficients[ndim]
   673                                               except KeyError:
   674                                                   raise ValueError(
   675                                                       f"Got unsupported value for argument 'ndim' ({ndim}), expect one of "
   676                                                       f"{list(_coefficients.keys())}."
   677                                                   )
   678      1000       3948.0      3.9      0.2      try:
   679      1000       5550.0      5.5      0.3          coef = coef[method]
   680                                               except KeyError:
   681                                                   raise ValueError(
[...]

PR #1098, opened 2021-04-02
[PROTO] Framework for lite "fast" formulary functionality

rocco8773:
This is a prototype for developing a framework to implement lite ("lightweight") formulary functionality whose design intent is for computation.  The current formulary functionality is constructed around Astropy's units for user convenience and to reduce calculation errors from improper unit use.  This is great, but yields very slow executing code.  The goal here is to develop a framework that still has our traditional formulary functionality but also an intuitive lite version for computation.

This prototype is developed for the `thermal_speed` function.  The approach this prototype takes is...

1. A lightweight thermal speed function is developed as `thermal_speed_lite`.   The formula this function has is based on SI units, but there is no argument validations (values or units).  The assumption here is the user knows what is going and can insure proper values are passed to the function.  The function is also decorated with `numba.njit`.
2. The previous `thermal_speed` function exists as it always has, but now utilizes `thermal_speed_lite` within the code for calculation.
3. A decorator `add_lite` is developed to bind `thermal_speed_lite` to `thermal_speed` as `thermal_speed.lite`, as well as update `thermal_speed`'s docstrings to indicate the addition of `lite`.

This all works something like...

```python
@preserve_signature  # this is needed since njit destroys the signature
@numba.njit
def thermal_speed_lite(...):
    ...

@add_lite(thermal_speed_lite)
@validate_quantities(...)
@particle_inpute
def thermal_speed(...):
    ...
```

The use now looks like...

```python
>>> T = 5 * u.K
>>> par = Particle("He+)
>>> thermal_speed(T, par)
<Quantity 144.13705752 m / s>

>>> %timeit thermal_speed(T, par)
717 µs ± 34.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

>>> # now using lite
>>> coef = thermal_speed.coefficients[3]["most_probable"]  # coefficients is also bound using add_lite
>>> thermal_speed.lite(T.value, par.mass.value, coef)
144.13705751725402

>>> %timeit thermal_speed.lite(T.value, par.mass.value, coef)
71.7 µs ± 3.16 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> # better performance can be achieved if units don't have to be used at all
>>> T_unitless, mass_unitless = T.value, par.mass.value
>>> %timeit thermal_speed.lite(T_unitless, mass_unitless, coef)
397 ns ± 5.09 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

---
See further discussions in #1090

### DIFF
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -537,11 +537,17 @@ cs_ = ion_sound_speed
 # This dictionary defines coefficients for thermal speeds
 # calculated for different methods and values of ndim.
 # Created here to avoid re-instantiating on each call
-_coefficients = {
+#
+thermal_speed_coefficients = {
     1: {"most_probable": 0, "rms": 1, "mean_magnitude": 2 / np.pi},
     2: {"most_probable": 1, "rms": 2, "mean_magnitude": np.pi / 2},
     3: {"most_probable": 2, "rms": 3, "mean_magnitude": 8 / np.pi},
 }
+"""
+Dictionary of various coefficients used in calculating the
+`~plasmapy.formulary.parameters.thermal_speed`.
+"""
+
 
 
 @check_relativistic
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -535,6 +535,7 @@ def ion_sound_speed(
 cs_ = ion_sound_speed
 """ Alias to :func:`ion_sound_speed`. """
 
+# -----                                                             thermal_speed  -----
 
 # This dictionary defines coefficients for thermal speeds
 # calculated for different methods and values of ndim.
@@ -560,6 +561,12 @@ def thermal_speed_lite(T, mass, coeff):
     """
     return np.sqrt(coeff * k_B_si_unitless * T / mass)
 
+
+@add_lite(
+    thermal_speed_lite,
+    attrs=[("coefficients", "thermal_speed_coefficients")],
+    scope=globals(),
+)
 @check_relativistic
 @validate_quantities(
     T={"can_be_negative": False, "equivalencies": u.temperature_energy()},
@@ -577,7 +584,9 @@ def thermal_speed(
     Return the most probable speed for a particle within a Maxwellian
     distribution.
 
-    **Aliases:** `vth_`
+    **Aliases:** `~plasmapy.formulary.parameters.vth_`
+
+    **Lightwieght Version:** `~plasmapy.formulary.parameters.thermal_speed_lite`
 
     Parameters
     ----------
@@ -703,7 +712,9 @@ def thermal_speed(
 
 
 vth_ = thermal_speed
-""" Alias to :func:`thermal_speed`. """
+""" Alias to :func:`~plasmapy.formulary.parameters.thermal_speed`. """
+
+# -----                                                          thermal_pressure  -----
 
 
 @validate_quantities(
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -549,6 +549,14 @@ Dictionary of various coefficients used in calculating the
 """
 
 
+@preserve_signature
+@njit
+def thermal_speed_lite(T, mass, coeff):
+    """
+    A lite weight version of `thermal_speed` intended for computational used
+    and, thus, does not do any argument validation/conditioning.
+    """
+    return np.sqrt(coeff * k_B_si_unitless * T / mass)
 
 @check_relativistic
 @validate_quantities(
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -686,12 +686,12 @@ class _ThermalSpeed:
         2: {"most_probable": 1, "rms": 2, "mean_magnitude": np.pi / 2},
         3: {"most_probable": 2, "rms": 3, "mean_magnitude": 8 / np.pi},
     }
+    """A docstring for coefficients."""
 
-    def __init__(self):
-        ...
 
     @staticmethod
     def unitless(T, mass, coef):
+        """A docstring for unitless."""
         return np.sqrt(coef * k_B_si_unitless * T / mass)
 
     @check_relativistic
@@ -708,6 +708,7 @@ class _ThermalSpeed:
         mass: u.kg = None,
         ndim=3,
     ) -> u.m / u.s:
+        """A docstring for the __call__"""
         if mass is None:
             mass = particles.particle_mass(particle)
 
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -545,7 +545,7 @@ _coefficients = {
     mass={"can_be_negative": False, "can_be_nan": True},
 )
 @particles.particle_input
-def thermal_speed(
+def _thermal_speed(
     T: u.K,
     particle: Particle,
     method="most_probable",
@@ -678,6 +678,48 @@ def thermal_speed(
     return np.sqrt(coef * k_B * T / m)
 
 
+class _ThermalSpeed:
+
+    def __init__(self):
+        ...
+
+    @staticmethod
+    def unitless(T, mass, method="most_probable", ndim=3):
+        # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
+        try:
+            coef = _coefficients[ndim]
+        except KeyError:
+            raise ValueError(
+                "{ndim} is not a supported value for ndim in thermal_speed")
+        try:
+            coef = coef[method]
+        except KeyError:
+            raise ValueError("Method {method} not supported in thermal_speed")
+
+        return np.sqrt(coef * k_B.value * T / mass)
+
+    @check_relativistic
+    @validate_quantities(
+        T={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+        mass={"can_be_negative": False, "can_be_nan": True},
+    )
+    @particles.particle_input
+    def __call__(
+        self,
+        T: u.K,
+        particle: Particle,
+        method="most_probable",
+        mass: u.kg = None,
+        ndim=3,
+    ) -> u.m / u.s:
+        if mass is None:
+            mass = particles.particle_mass(particle)
+
+        speed = self.unitless(T=T.value, mass=mass.value, method=method, ndim=ndim)
+        return speed * u.m / u.s
+
+
+thermal_speed = _ThermalSpeed()
 vth_ = thermal_speed
 """ Alias to :func:`thermal_speed`. """
 
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -689,7 +689,6 @@ class _ThermalSpeed:
     }
     """A docstring for coefficients."""
 
-
     @staticmethod
     @njit
     def unitless(T, mass, coef):
@@ -714,16 +713,16 @@ class _ThermalSpeed:
         if mass is None:
             mass = particles.particle_mass(particle)
 
-            # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
-            try:
-                coef = self.coefficients[ndim]
-            except KeyError:
-                raise ValueError(
-                    "{ndim} is not a supported value for ndim in thermal_speed")
-            try:
-                coef = coef[method]
-            except KeyError:
-                raise ValueError("Method {method} not supported in thermal_speed")
+        # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
+        try:
+            coef = self.coefficients[ndim]
+        except KeyError:
+            raise ValueError(
+                "{ndim} is not a supported value for ndim in thermal_speed")
+        try:
+            coef = coef[method]
+        except KeyError:
+            raise ValueError("Method {method} not supported in thermal_speed")
 
         speed = self.unitless(T=T.value, mass=mass.value, coef=coef)
         return speed * u.m / u.s
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -564,11 +564,11 @@ def thermal_speed_lite(T, mass, coeff):
     mass={"can_be_negative": False, "can_be_nan": True},
 )
 @particles.particle_input
-def _thermal_speed(
+def thermal_speed(
     T: u.K,
     particle: Particle,
     method="most_probable",
-    mass: u.kg = np.nan * u.kg,
+    mass: u.kg = None,
     ndim=3,
 ) -> u.m / u.s:
     r"""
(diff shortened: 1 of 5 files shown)


StanczakDominik:
> posted a minute ago

Good timing on my part! :wink:

plasmapy/formulary/parameters.py
```diff
@@ -528,35 +535,54 @@ def ion_sound_speed(
...
+#
+thermal_speed_coefficients = {
     1: {"most_probable": 0, "rms": 1, "mean_magnitude": 2 / np.pi},
     2: {"most_probable": 1, "rms": 2, "mean_magnitude": np.pi / 2},
     3: {"most_probable": 2, "rms": 3, "mean_magnitude": 8 / np.pi},
 }
+"""
+Dictionary of various coefficients used in calculating the
+`~plasmapy.formulary.parameters.thermal_speed`.
+"""
```
StanczakDominik, line 552: Minor rewrite:
```suggestion
"""
Dictionary of various coefficients used for
`~plasmapy.formulary.parameters.thermal_speed`, with the dimensionality
as the key.
"""
```

plasmapy/formulary/parameters.py
```diff
@@ -528,35 +535,54 @@ def ion_sound_speed(
...
+
+
+@preserve_signature
+@njit
+def thermal_speed_lite(T, mass, coeff):
+    """
+    A lite weight version of `thermal_speed` intended for computational used
+    and, thus, does not do any argument validation/conditioning.
```
StanczakDominik, line 560: Minor rewrite:
```suggestion
    A lightweight version of `thermal_speed` intended for computational use
    which does not do any argument validation/conditioning.
```

plasmapy/formulary/parameters.py
```diff
@@ -528,35 +535,54 @@ def ion_sound_speed(
...
+    T: u.K, particle: Particle, method="most_probable", mass: u.kg = None, ndim=3,
 ) -> u.m / u.s:
     r"""
     Return the most probable speed for a particle within a Maxwellian
     distribution.
 
-    **Aliases:** `vth_`
+    **Aliases:** `~plasmapy.formulary.parameters.vth_`
+
+    **Lightwieght Version:** `~plasmapy.formulary.parameters.thermal_speed_lite`
```
StanczakDominik, line 585: Typo fix
```suggestion
    **Lightweight Version:** `~plasmapy.formulary.parameters.thermal_speed_lite`
```

and wow, with the stars around this really jumps out like a coke ad.

plasmapy/formulary/parameters.py
```diff
@@ -663,23 +689,27 @@ def thermal_speed(
...
     <Quantity 621251... m / s>
     """
-    m = mass if np.isfinite(mass) else particles.particle_mass(particle)
+    if mass is None:
+        mass = particles.particle_mass(particle)
 
     # different methods, as per https://en.wikipedia.org/wiki/Thermal_velocity
     try:
-        coef = _coefficients[ndim]
+        coeff = thermal_speed_coefficients[ndim]
```
StanczakDominik, line 697: Just taking the opportunity to avoid overwriting the var name later with a minor refactor
```suggestion
        coeffs_for_dim = thermal_speed_coefficients[ndim]
```

plasmapy/formulary/parameters.py
```diff
@@ -663,23 +689,27 @@ def thermal_speed(
...
     except KeyError:
         raise ValueError("{ndim} is not a supported value for ndim in thermal_speed")
     try:
-        coef = coef[method]
+        coeff = coeff[method]
```
StanczakDominik, line 701: continues minor refactor
```suggestion
        coeff = coeffs_for_dim[method]
```

plasmapy/formulary/parameters.py
```diff
@@ -663,23 +689,27 @@ def thermal_speed(
...
     except KeyError:
         raise ValueError("Method {method} not supported in thermal_speed")
 
-    return np.sqrt(coef * k_B * T / m)
+    speed = thermal_speed_lite(T=T.value, mass=mass.value, coeff=coeff)
+    return speed * u.m / u.s
```
StanczakDominik, line 706: I know this is supposed to be the slow version, but on my machine the following is still twice as fast - not reason not to take that:
```suggestion
    return u.Quantity(speed, u.m / u.s)
```

(with a predefined speed unit `mps = u.m / u.s` there's another factor of 2 speed up)

plasmapy/formulary/parameters.py
```diff
@@ -663,23 +689,27 @@ def thermal_speed(
...
 
 
 vth_ = thermal_speed
-""" Alias to :func:`thermal_speed`. """
+""" Alias to :func:`~plasmapy.formulary.parameters.thermal_speed`. """
```
StanczakDominik, line 710: Works nicely with #1099, too! :smile:

plasmapy/utils/decorators/helpers.py
```diff
@@ -142,3 +142,51 @@ def wrapper(*args, **kwargs):
         wrapper.__signature__ = inspect.signature(f)
 
     return wrapper
+
+
+def add_lite(flite, attrs=None, scope=None):
```
StanczakDominik, line 147: Maybe "register_lite", come to think of it? I think "register" has a pretty common meaning for decorators.

plasmapy/utils/decorators/helpers.py
```diff
@@ -142,3 +142,51 @@ def wrapper(*args, **kwargs):
...
+            f" needs to be defined and contain the objects to be bound.  "
+            f"Setting 'scope=globals()' will typically suffice."
+        )
+
+    def decorator(f):
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            return f(*args, **kwargs)
+
+        wrapper_doc = inspect.cleandoc(getattr(wrapper, "__doc__", """"""))
```
StanczakDominik, line 166:
```suggestion
        wrapper_doc = inspect.cleandoc(getattr(wrapper, "__doc__", ""))
```
since
```python
> """""" == ""
True
```

plasmapy/utils/decorators/helpers.py
```diff
@@ -142,3 +142,51 @@ def wrapper(*args, **kwargs):
...
+        )
+
+        for bound_name, attr_name in attrs:
+            attr = scope[attr_name]
+
+            setattr(wrapper, bound_name, attr)
+
+            wrapper_doc += (
+                f"\n   - `{f.__name__}.{bound_name}` <--> `~{f.__module__}.{attr_name}`"
+            )
```
StanczakDominik, line 186: Looks great [in the docs!](https://plasmapy--1098.org.readthedocs.build/en/1098/api/plasmapy.formulary.parameters.thermal_speed.html#codecell0:~:text=To%20increase%20usability%20of%20thermal_speed%20several,are%20manually%20bound%20to%20this%20function.) :+1:

requirements/requirements.txt
```diff
@@ -5,6 +5,7 @@ h5py >= 2.8
 lmfit >= 1.0.1
 matplotlib >= 2.0
 mpmath >= 1.0.0
+numba
```
StanczakDominik, line 8: Once you start adding tests for this, tox-minimal-37 will complain since it has its own list of dependencies. I'm adding that in `tox.ini` for you.

StanczakDominik (approved):
> The goal here is to develop a framework that still has our traditional formulary functionality but also an intuitive lite version for computation.

I personally hate the fact that we're still enjoying zero-ver, but already have traditions around functionality :wink:

Looks great! A bunch of minor text fix suggestions, but I have no complaints about the implementation!

plasmapy/formulary/parameters.py
```diff
@@ -528,35 +535,54 @@ def ion_sound_speed(
...
+    attrs=[("coefficients", "thermal_speed_coefficients")],
+    scope=globals(),
+)
 @check_relativistic
 @validate_quantities(
     T={"can_be_negative": False, "equivalencies": u.temperature_energy()},
     mass={"can_be_negative": False, "can_be_nan": True},
 )
 @particles.particle_input
 def thermal_speed(
```
pheuer, line 576: The docstring for `thermal_speed` includes a table of the coefficients: maybe we should add a link next to that to the `thermal_speed_coefficients` variable to make more clear that these contain the same information.

plasmapy/utils/decorators/helpers.py
pheuer, line 147: I think this might be better split into two functions, eg. `register_lite` and `register_attrs`. Or, maybe `flite` should also be a kwarg? I'm just thinking that currently attaching an attribute without also attaching a lite function would not be supported, but that seems like it should be possible.

plasmapy/utils/decorators/helpers.py
```diff
@@ -142,3 +142,51 @@ def wrapper(*args, **kwargs):
...
+
+
+def add_lite(flite, attrs=None, scope=None):
+    """
+    Decorator to bind lightweight "lite" versions of formulary functions to the full
+    formulary function.
+    """
+    if attrs is None:
+        attrs = []
+    elif scope is None:
```
pheuer, line 154: Is there a reason to not just set scope=globals() by default?

plasmapy/formulary/parameters.py
```diff
@@ -528,35 +535,54 @@ def ion_sound_speed(
...
+"""
+Dictionary of various coefficients used in calculating the
+`~plasmapy.formulary.parameters.thermal_speed`.
+"""
+
+
+@preserve_signature
+@njit
+def thermal_speed_lite(T, mass, coeff):
+    """
```
pheuer, line 558: This docstring should still provide information on the parameters and returns so those appear in the documentation. Right now in the docs there's no indication that this function doesn't take units, or where one should look to determine `coeff`

pheuer (changes requested):
This looks great @rocco8773 !! I'm so excited about this. I've suggested a few changes - let me know what you think.

The documentation looks great: the only part I don't love is that the lite versions of the functions appear alongside the main functions and the aliases in the list of functions. But, I think we should just accept that for now. When we figure out a solution for the aliases, hopefully that will improve this situation too.

plasmapy/utils/decorators/helpers.py
StanczakDominik: Hmm, what use case would you see for that?

plasmapy/utils/decorators/helpers.py
StanczakDominik: I suspect this part could be optimized out given some tinkering. Maybe `sys.modules(__module__)` combined with `importlib` and `inspect.getmembers` in some funny way...

But at this point, if you just bluntly set `globals` as a default from `helpers.py`, it's going to "inherit the scope of helpers.py" rather than the local environment of the function that's being registered.

plasmapy/utils/decorators/helpers.py
pheuer: I don't see one now: just following the "one function one job" principle.

plasmapy/utils/decorators/helpers.py
pheuer: Got it - that makes sense (as much as anything regarding scopes does)

pheuer:
@StanczakDominik Let's continue the conversation from #1099 about separately documenting lite functions here. It sounds like we have three possible solutions:

1) Leave lite functions alone: let them appear right next to the parent functions in the docs.
    - Pros: Less work, no additional code
    - Cons: Potentially doubles the number of visible functions. Could be confusing when some functions have lite versions but others don't yet.
    
2) Use "semi-dynamically generated module" wizardry to sort "_lite" functions into a separate module under the hood (like #1099 does for aliases). 
    - Pros: Separates the lite functions visually in the docs.
    - Cons: Possible unintended consequences? No one understands sphinx. 
 
3) Remove lite functions from all, document them manually in the docstrings of their main functions (possibly with a decorator that automatically inserts the lite function docstring at the end of the main function). 
    - Pros: Cleaner documentation (for the main functions)
    - Cons: Harder to find documentation on lite functions? Sphinx might complain about non-private functions being left out of `__all__`?

Obligatory tag @namurphy because trade-off analysis 😃

StanczakDominik:
See, Nick? We're learning! :smile_cat:

plasmapy/utils/decorators/helpers.py
rocco8773: It is doing "one job"  I added the ability to add other attributes than "lite", because sometimes you need additional attributes to properly use "lite".  For example, for `thermal_speed.lite` to properly work you also need the coefficients.  So, think of `add_lite` as binding all required "lite" functionality (not just the lite function) to get `<func>.lite` to work.

rocco8773: Additionally, to get the documentation to come out correctly and cleanly then the lite function and required attributes needs to be added at the same time.

pheuer: I see that. I guess I'm fine with this implementation now: if we need to add attributes separately in the future I guess we can always break this function apart then.

rocco8773:
> > The goal here is to develop a framework that still has our traditional formulary functionality but also an intuitive lite version for computation.
>
> I personally hate the fact that we're still enjoying zero-ver, but already have traditions around functionality 😉

I think this is why I've become more in favor of the date base versioning.  I think plasmapy will always have a combination of mature and development functionality.  That's why I consider it perpetual beta.

> Looks great! A bunch of minor text fix suggestions, but I have no complaints about the implementation!

Great!  I wanted to get everyone's opinion on the framework before spending the time to put the bells and whistles on all the functionality and docstrings.  I can start refining everything now.

plasmapy/formulary/parameters.py
rocco8773: Yeah, I ran across this same thing when working on this PR.  I actually thought I already made this change.  I think we need to get in the practice of using the explicit link since `` `thermal_speed` `` would just look for the variable in the current `global()` scope.

plasmapy/utils/decorators/helpers.py
rocco8773: When I think "register" I'm think it's registering the functionality with some kind of table/database.  How about `bind_lite`?  That's closer to what it is actually doing.

requirements/requirements.txt
rocco8773: I thought we already had `numba` for some reason, so I was surprised when all the tests started failing because of this.  Do you have any restrictions related to which `numba` version we should take?

plasmapy/formulary/parameters.py
rocco8773: Agreed! I'll get to this.  I haven't refined and the docstrings because I didn't want to spend the time on it until the framework was approved.

plasmapy/utils/decorators/helpers.py
rocco8773: > But at this point, if you just bluntly set globals as a default from helpers.py, it's going to "inherit the scope of helpers.py" rather than the local environment of the function that's being registered.

Yes, this is what was happening when `globals()` was the default value.

rocco8773:
> 1. Leave lite functions alone: let them appear right next to the parent functions in the docs.
>
>    * Pros: Less work, no additional code
>    * Cons: Potentially doubles the number of visible functions. Could be confusing when some functions have lite versions but others don't yet.

I don't see that as a con.  `<func>_lite` is part of the public api and thus should be documented.  Hiding it from the documentation significantly reduces discoverability, which is a major pain for users.  Since `<func>_lite` is apart of the public api, a use would see it, go look for it in the documentation, and then not be able to find it.  This is BAD!  I'd rather have everything listed than not have something discoverable.

StanczakDominik:
Well, given #1100, let's definitely go with the automodsumm one! I would like to try my hand at adapting the solution from there to the `_lite` functions case - I guess the whole of #1099 resulted from me not knowing enough sphinx and I'd like to fix that! So I'd be completely fine with leaving them in the main table for now, with the understanding that I'll fix them later.

plasmapy/utils/decorators/helpers.py
StanczakDominik: That sounds great!

requirements/requirements.txt
StanczakDominik: Yeah, we experimented with adding it in #953; at the time, Py3.9 was released, numba completely broke on it with its bytecode changes, and we didn't want to add version-dependent numba-or-nonumba logic. 0.53 is out now; things should be fine, as py39 will just (I think...) default to that as the only compatible dependency. Though we should look at the CI logs and check which runners get which version, just in case.

I don't have a clean lower bound on which numba versions are fine, though.

pheuer:
> Since <func>_lite is apart of the public api, a use would see it, go look for it in the documentation

I definitely agree with this! My only concern is that I think having one very long list of functions in the documentation makes the package less approachable, especially to students new to plasma physics.

"Who are these crazy people and why do they have SO MANY DIFFERENT FREQUENCIES? Maybe I should transfer to a condensed matter group..."

I agree the automodsum implementation a la #1100 is the ideal way of handling this. I'd like to see a list of "Functions", followed by "Lite Functions" and "Aliases" (in whatever order). Then the first thing the user sees is a more manageable list of the main functions, and they have to scroll down to see the different versions of those functions.

rocco8773:
As I stated in https://github.com/PlasmaPy/PlasmaPy/pull/1100#issuecomment-812910309, I think I have a solution for grouping lite functions, so stay tuned for a prototype.

As I stated before, I think the lite functions should be document like everything else and I do agree they cause clutter on the documentation page.  I like the ability to group them in a separate section that explains what they are.

Issue #1097, opened 2021-04-01
Formulary organization


pheuer:
The formulary namespace is getting pretty cluttered, especially `formulary.parameters` which is basically a catch-all category with 38 functions (including aliases) and counting.

I think we should consider taking a cue from the NRL formulary and separating some of these functions into categories based on their units, eg. "frequencies", "lengths", and "velocities". Some other quantities could be easily refactored into this scheme: for example all of the functions in `drifts.py` would fit nicely into "velocities".  Other candidate categories might be "pressures", "densities", and "diffusivities". The functions in `dimensionless.py` could also be made into a "dimensionless" category at the same level. I guess I'm proposing putting these at the same level as `parameters` is now, eg. `formulary.frequencies.py`. I would replace `paramters.py` with `misc.py` in this new scheme.

This would also make it easier to ensure that functions are in the right place. For example, right now `Hall_parameter` is in `parameters` when it definitely belongs in `dimensionless.py`. 

One con is that not all functions make sense in this scheme: for example `collision_frequency` belongs in `collisions` more than it belongs with `gyrofrequency` . But, I'd argue that distinction is no more arbitrary than leaving `collision_frequency` out of `parameters`. 

Finally, I think the number of aliases is cluttering the documentation. ~~Is there a way to hide these in `autodoc` on the module level so that they only appear listed as "aliases" when you click on the individual function?~~ It looks like some approach grouping the aliases together will work better.

rocco8773:
I absolutely agree the `formulary` is starting to get cluttered and could use a little reorganizing.  However, there's no non-trivial way to improve this.  The mechanisms (i.e. `__all__`) used for organizing the package namespace is also used for controlling the documentation.  So, restricting the `formulary` namespace can really hinder the exposure of our functionality.   Additionally, if we want good exposure of functionality on the `formulary` namespace then it will always have a LOT of names.  This is kind of like the `astropy.units` namespace.  It will always be large because a lot of functionality needs to be exposed there.

pheuer:
I don't know much about Sphinx or automodapi, but maybe[ this `:skip:` command](https://sphinx-automodapi.readthedocs.io/en/latest/automodapi.html#module-sphinx_automodapi.automodapi) could be used to hide functions from the docs? If that hides it from the list of module functions but still allows it to be set as an alias, that would be a solution?

Alternately, since the aliases really are one-to-one replacements, they could be hidden and then just mentioned in the docstring of the function they replace?

[...]

PR #1099, opened 2021-04-02
Semi-dynamically generated aliases submodule 

### LINKED ISSUE #866: Have a aliases submodule for formulary aliases
In our recent community meeting ([minutes](https://docs.google.com/document/d/1zVPBZInQ1jEEW059GORpWAszjUD6cgdy_8ca4Ubyezo/edit?usp=sharing)) there was a discussion of creating an alias module (`aliases.py`) within the `plasmapy.formulary` to act as a container for all aliases.  A framework for this approach needs to hashed out, and the framework should not create excess work for maintaining it.

StanczakDominik:
Touches #1097. Closes #866.

---

# Note

Full outline in https://github.com/PlasmaPy/PlasmaPy/pull/1099#issuecomment-812398097.

---

@rocco8773 take a look :) I'm kinda happy with how this works:

```bash
>>> python -c 'from plasmapy.formulary import parameters; print(dir(parameters))'

['Alfven_speed', 'Bohm_diffusion', 'ChargeError', 'DB_', 'Debye_length', 'Debye_number', 'Hall_parameter', 'Optional', 'Particle', 'PhysicsError', 'PhysicsWarning', 'Union', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_coefficients', '_grab_charge', 'aliases', 'angular_freq_to_hz', 'betaH_', 'c', 'check_relativistic', 'cs_', 'cwp_', 'e', 'eps0', 'gyrofrequency', 'gyroradius', 'inertial_length', 'ion_sound_speed', 'k_B', 'kappa_thermal_speed', 'lambdaD_', 'lower_hybrid_frequency', 'magnetic_energy_density', 'magnetic_pressure', 'mass_density', 'mu0', 'nD_', 'np', 'numbers', 'oc_', 'particles', 'plasma_frequency', 'pmag_', 'pth_', 'rc_', 'rho_', 'rhoc_', 'thermal_pressure', 'thermal_speed', 'u', 'ub_', 'upper_hybrid_frequency', 'va_', 'validate_quantities', 'vth_', 'vth_kappa_', 'warnings', 'wc_', 'wlh_', 'wp_', 'wuh_']

>>> python -c 'from plasmapy.formulary.aliases import parameters; print(dir(parameters))'
['DB_', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'betaH_', 'cs_', 'cwp_', 'function', 'importlib', 'inspect', 'lambdaD_', 'module', 'nD_', 'name', 'oc_', 'pmag_', 'pth_', 'rc_', 'rho_', 'rhoc_', 'sys', 'ub_', 'va_', 'vth_', 'vth_kappa_', 'wc_', 'wlh_', 'wp_', 'wuh_']
```

And at the same time, the main parameters page doesn't show the aliases. Because they're not in `__all__` anymore. But you can still import them; just not with a star import from parameters.

Of course the naming and interface is an initial draft.

I have not yet gotten documentation for the aliases submodule to work. I suspect that's going to be the largest hurdle here, compared to the relatively simple implementation.

Whatcha think about the general direction?

---

# TODO

- [x] apply alias function to other modules of the formulary
- [x] cut the aliases from `__all__` in those, too
- [x] fix linkbacks for each alias
- [x] ~~replace alias detection condition from "does not start with `_`, ends with `_`" with something more sensible~~ fine for now?


rocco8773:
> I have not yet gotten documentation for the aliases submodule to work. I suspect that's going to be the largest hurdle here, compared to the relatively simple implementation.

`automodapi` generates the documentation from `__all__`, so populate that and the documentation will show up.  You can probably even make `aliases` into a py file instead of a directory, depending on how you intend to organize it.  You'll likely need to may and `api_static` stub file as well, `plasmapy.formulary.aliases.rst`.

---

I'm heading off to bed, so I'll have to think about this a little more.  However, I'm still hesitant on having aliases in a different location than their parent functions.  Logically it makes sense that they should live in the same namespace.  Moving the aliases to a different namespace breaks the parent-alias connection, which forces a user to put more effort forward in investigating the documentation to locate functionality and understand what's happening.  For same reason, and as I said in #866 (https://github.com/PlasmaPy/PlasmaPy/issues/866#issuecomment-812305434), I don't like the idea of hiding the alias documentation because you're hiding document to public api features.

I think I'd be more inclined to do something closer to what @pheuer suggested in #1097, with some modification.  So,

1. We better catagorized the sub-modules in `formulary` to provide more organization.  That is, have your `plasmapy.formulary.speed`, `.frequencies`, `.lengths`, etc.
2. Each of those sub-modules would still define and document the aliases.
3. Remove all imports from `formulary.__init__` except for just the module imports.  That is, include `from . import speed` but not `from .speed import *`.

I see strong pros and cons with point 3, and I'm sure everyone will have an opinion on it.  This approach will better control namespace pollution, but hinder discoverability since not all functionality is sitting on `plasmapy.formulary`.  However, I think this would be my preferred route at the moment.

StanczakDominik:
> I'm heading off to bed, so I'll have to think about this a little more. However, I'm still hesitant on having aliases in a different location than their parent functions. Logically it makes sense that they should live in the same namespace. Moving the aliases to a different namespace breaks the parent-alias connection, which forces a user to put more effort forward in investigating the documentation to locate functionality and understand what's happening. For same reason, and as I said in #866 (#866 (comment)), I don't like the idea of hiding the alias documentation because you're hiding document to public api features.

That description is not entirely what I had in mind. My plan was:

* `parameters.py` defines both the functions and their aliases, just like you said; nothing changes on that front
* `parameters.__all__` includes just the functions; star imports from `parameters` do not include aliases, but module imports do (I checked, it works).
* `aliases/parameters.py` imports the aliases from `parameters.py`
* the documentation in `parameters.rst` displays first an automodapi directive for `parameters.py`, including the main functions (as the current doc build successfully does), and then an automodapi directive for `aliases/parameters.py`. The aliases are not hidden, they're just not littering the same table of functions.
* (this might be weird/contentious `parameters.py` actually imports `aliases/parameters.py as aliases`, so the aliases are still within the parameters.py namespace; just grouped as `parameters.aliases.rho_` etc.

@pheuer's proposed #1097 rework of the formulary organization might be a good idea, but I still think it does not take precedence over separating the functions and their aliases in the very-hard-to-visually-grep table of functions provided by `parameters.rst`.

@pheuer, what do you think about that?

StanczakDominik:
Ha, got it to work! Take a look at the doc build for parameters.

docs/formulary/parameters.rst
```diff
@@ -4,9 +4,23 @@
...
+=======
+Aliases
+=======
+
+These are not star-imported when you do `from plasmapy.formulary.parameters import *` by default;
+you can do that through `from plasmapy.formulary.aliases.parameters import *`.
+
+.. automodsumm:: plasmapy.formulary.aliases.parameters
+
+
```
StanczakDominik, line 23: For maximum DRY this could be a custom sphinx directive. But I'm not that kind of crazy.

StanczakDominik (commented):
This looks closer to prod-ready, with remaining issues I can see:

- [ ] apply alias function to other modules of the formulary
- [ ] cut the aliases from `__all__` in those, too
- [ ] fix linkbacks for each alias

And of course, review and discussion still pending!

StanczakDominik:
Also paging @namurphy for opinions about the parameters doc page's look with this change! (link two comments above)!

pheuer:
I'm torn: I really don't like the clutter of aliases in the current documentation, but I also agree with @rocco8773 that aliases should exist in the same namespace and file as their parent function. With this change, would you still be able to use `import plasmapy.formulary.parameters.DB_`? If so, I'm all for it. If not, I have some reservations about that. 


Comments: 

- The documentation looks SLICK!

- Is the `name.endswith("_")` convention really the best way to identify which functions are aliases? Is it possible to use a decorator or set some property of the variable that more explicitly flags `DB_` as an alias? Idk, `__isalias__` as a boolean attribute of the function? 

- If we can figure out a way of doing this that doesn't rely on the naming convention, I'd love to explore extending this in #1098 to make a similar separate list for lite functions.

It sounds like there is a tradeoff here (easier to read documentation vs. slightly more complicated aliases). My vote is that this is a good tradeoff, since beginner users are unlikely to be using aliases but very likely to be trying to read the docs!

StanczakDominik:
> With this change, would you still be able to use import plasmapy.formulary.parameters.DB_? If so, I'm all for it.

Then you're all for it :) I'm gonna try recording an example:

StanczakDominik:
There we go, this is how it works: (click for video, I think)

[asciicast](https://asciinema.org/a/404634)

Also,

> Is the name.endswith("_") convention really the best way to identify which functions are aliases? Is it possible to use a decorator or set some property of the variable that more explicitly flags DB_ as an alias? Idk, __isalias__ as a boolean attribute of the function?

No, that was just the first workable strategy that came to mind, a placeholder, essentially! I'm all for replacing that with something more sensible. I'm also all for idea suggestions :smile:

StanczakDominik:
I honestly think that, for now, it's completely fine to go with something like "name is a short string, ends with `_`" rather than adding attributes to each alias... oh right, that wouldn't work either. Both the alias and the original function point to the same function object in memory. You cannot modify one without modifying the other.

I honestly think that this is a problem that we can figure out if/when it actually starts to become a problem, and at this point, we'd be overengineering it.

pheuer:
Awesome, looks great! 

Ok, I see your point: probably any solution there for aliases would introduce unnecessary complexity at this stage. 

That just means that, in order to do this with the lite functions we'd need a different approach (unless you want to use the naming convention `endsin("_lite")` there?).

StanczakDominik:
Well, you still have to make different files, which is a bit clunky, and I suspect this solution (especially the part where I mess with `currentmodule`) might have Sphinx caveats that my wizardry is not powerful enough to grasp. For example, the aliases do not show up in the [generated variables index](https://plasmapy--1099.org.readthedocs.build/en/1099/genindex.html). I don't think that's a bad thing; they're not different objects, after all.

pheuer:
Conversation about the lite functions continued over on #1098

StanczakDominik:
I just realized that `find_floating_potential` in `analysis` and `two_fluid_dispersion_solution` in `dispersion` also have aliases defined. I have opinions about that, but for now let's just say I'm limiting myself to the formulary. :)

rocco8773:
I still really don't like the aliases being moved to a different file/namespace. (I really regret making that issue now.)  A user should be able to easily transition from `from plasmapy.formulary.parameters import thermal_speed` to `from plasmapy.formulary.paramaters import vth_`.  Additionally, `vth_` is defined in `plasmapy/formulary/parameters.py` and is a public API (regardless of being in `__all__` or not) and, thus, should be documented in that place.  Re-importing it into `aliases` does not side step that fact.  I could support having the `aliases` module in addition to what we currently have, but this does add to the namespace.

Additionally, `import *` should not be apart of this conversation because `import *` should really never happen, because it is a terrible practice that is consider a terrible practice by the python community.  If a user `import *`'s and doesn't like the results, then that's their own fault.

To clean up the namespace, I'd rather see `parameters.py` converted into a directory containing `speed.py`, `frequency.py`, etc.  Then aliases can live and be documented next to each other while mitigating file length.

I'll also say, having a long table displayed on `plasmapy.formulary` documentation page is not a bad thing, even if it's hard to search.  It does wonders for discoverability, which is a primary goal of documentation.  Just look at Astropy's units table,  It's large and hard to search, but I'll deal with that for the benefit of searching for all functionality in one place.

> I just realized that find_floating_potential in analysis and two_fluid_dispersion_solution in dispersion also have aliases defined. I have opinions about that, but for now let's just say I'm limiting myself to the formulary. :)

I'm find with having aliases for functionality outside of the formulary.  The idea is that we adopt a naming convention `<name>_` so users of plasmapy automatically know that functionality is an alias for something else.  This idea should not be restricted to just the formulary.

rocco8773:
I'll mention, even though `__all__` controls the `*` imports I don't define it or use for that context.  I used because it controls what is documented by `automodapi`.  For me, it's primarily a documentation control.

rocco8773:
> That just means that, in order to do this with the lite functions we'd need a different approach (unless you want to use the naming convention endsin("_lite") there?). I actually don't hate that.

I don't want to exclude `<name>_lite` from the documentation.  In fact, it is need!.  Bound attributes to a function, i.e. `thermal_speed.lite`, are note documented by sphinx.  This is why my `add_lite` decorator in #1098 manually adds the admonition stating `thermal_speed.lite` maps to `thermal_speed_lite`.  That admonition also serves to link to the `thermal_speed_lite` documentation.  If `thermal_speed_lite` is excluded, then the lite functionality will never be documented.

rocco8773:
> Well, you still have to make different files, which is a bit clunky. I wouldn't think we'd want a different submodule for the lite functions.

I would not want a separate module for lite functions for the same reason I don't want a separate module for aliases.  The functionality is inherently linked to each other and, thus, should be defined next to each other.

rocco8773:
I do like the idea of aliases being in a separate table in the documentation, but I don't like all the namespace tricks we have to do to get that to happen.  I'd rather created a sphinx directive much like [`automodsumm`](https://sphinx-automodapi.readthedocs.io/en/latest/automodsumm.html) (and [code on GitHub](https://github.com/astropy/sphinx-automodapi/blob/main/sphinx_automodapi/automodsumm.py)) that would do this for us.

What I would envision,  we can defined an `__aliases__` dunder much like `__all__` to contain the alias names.  This means we could also move the alias names out of `__all__`.  The directive (say `autoaliases`) would read `__aliases__` and constructed the `autosummary` table, just like `automodsumm` does with `__all__`.  Then writing the `plasmapy.formulary.rst` `api_static` page it would look something like...

```restructuredtext

.. automodapi:: plasmapy.formulary
   :no-main-docstr:
   :no-heading:

.. autoaliases:: plasmapy.formulary
```

Now, all you'd have to do is defined and maintain `__aliases__` to get the documentation to auto update.  Ideally, I wish we could have this in `automodapi`, but I don't think it belongs there either.  This is a reason why I wanted to make a plasmapy version of `automodapi`, but it hasn't moved up the priority list.

StanczakDominik:
We could fork automodapi and depend on the fork (via a github sourced pip dependency) while we wait for them to accept the PR. Just pointing out that's a possibility.

Anyway like I said in the other PR, I like the custom automodsumm approach much more and I should definitely have waited with the other files besides just parameters on further discussion. My bad! No harm done, though. I'll close this one in favor of that PR.

rocco8773:
> Anyway like I said in the other PR, I like the custom automodsumm approach much more and I should definitely have waited with the other files besides just parameters on further discussion. My bad! No harm done, though. I'll close this one in favor of that PR.

Nah, I like that you opened this PR.  Thanks you!  I think it really drove the conversation so we got to commonly accepted approach sooner.  It definitely helped me think out he problem.

StanczakDominik:
Nah, I do think the PR itself was a good call :) just went about it suboptimally in the later stages. But like I said, no harm done. Glad it helped!

PR #1100, opened 2021-04-03
[PROTO] Create sphinx directive autosum_aliases for documenting aliases separately

rocco8773:
This is a quick-and-dirty prototype to illustrate the Sphinx directive `autosum_aliases` I was suggesting in https://github.com/PlasmaPy/PlasmaPy/pull/1099#issuecomment-812673370.  This is quick-and-dirty for illustration purposes only, so there's still refinement to be had.  It is adapted from [`shpinx_autmodapi.automodsumm` code](https://github.com/astropy/sphinx-automodapi/blob/main/sphinx_automodapi/automodsumm.py).   I applied the `autosum_aliases` directive to the `formulary.parameters` documentation page so we all can see how the aliases are still automatically documented.  All aliases are move to the `__aliases__` dunder so they will not be import with `*` imports.


StanczakDominik:
At a glance this looks way better from my yesterday's prototype. I can definitely get behind this, so I'll close the previous PR :)

StanczakDominik:
I added a little CI magic to ignore the new docs/.py file in pytest. This should fix the tests. I originally also ran pre-commit but wouldn't want to cause you merge conflicts with anything you might have locally, so I reverted that.

plasmapy/formulary/parameters.py
```diff
@@ -20,18 +16,24 @@
     "magnetic_energy_density",
     "magnetic_pressure",
     "mass_density",
+    "plasma_frequency",
+    "thermal_pressure",
+    "thermal_speed",
+    "upper_hybrid_frequency",
+]
+__aliases__ = [
```
pheuer, line 24: Missed one! `lambdaD_`

docs/autosum_aliases.py
```diff
@@ -0,0 +1,465 @@
...
+
+        # loop over all autosum_aliases in this document
+        for i, (i1, i2, modnm, ops, rem) in enumerate(zip(indent1s, indent2s, mods,
+                                                          opssecs, remainders)):
+            allindent = i1 + ('    ' if i2 is None else i2)
+
+            # filter out functions-only, classes-only, and ariables-only
+            # options if present.
+            option_lines = ops.split('\n')
+            # toskip = []
```
pheuer, line 128: Probably best to remove all this commented-out code prior to merging

docs/conf.py
```diff
@@ -263,5 +263,11 @@
 
 
 def setup(app: Sphinx) -> None:
+
+    import autosum_aliases
```
pheuer, line 267: This is resulting in a CI error:

```
File "/home/runner/work/PlasmaPy/PlasmaPy/docs/conf.py", line 267, in setup
import autosum_aliases
ModuleNotFoundError: No module named 'autosum_aliases'
```

pheuer (commented):
I won't pretend to understand how most of this works, but I'm very impressed with the results! Bravo! This is absolutely the best way to do this, and the documentation looks so much cleaner already.

Would it be possible to make this more general and allow subclasses of this class that just define which list of functions to look for (eg. `__aliases__`)? It seems like this would be the ideal solution to any case where we want to divide up a long list of functions into conceptually separate lists in the docs without actually changing the package structure. But, we obviously wouldn't want to reproduce this 500 lines of code.

I'll hold off on approving this, in case anyone who DOES understand what's going on with Sphinx wants to do a more complete review. But my vote is, if it works, we should merge it.

StanczakDominik:
> I'll hold off on approving this, in case anyone who DOES understand what's going on with Sphinx wants to do a more complete review. But my vote is, if it works (and it appears to), we should merge it.

Well put! I'm in the same boat :D

rocco8773:
Ok, I actually have an idea for making this work for grouping "aliases" and "lite functions", as well as making it more expandable in case of future wants.  I'm also hoping to do this such that we don't have to deviate from using just `automodapi` or `automodsumm`.   I'll get a prototype up hopefully in a few days.

rocco8773:
Replaced by PR #1105

Issue #856, opened 2020-06-16
Create aliases for common formulary functions.


rocco8773:
### Update (2020-12-17)

Most of the common `plasmapy.formulary` functionality has been given aliases.  This issue is being left open since aliases may be desired on future functionality.  If there's functionality that does not have an alias and you'd like one, then please leave a comment.

----

Explicit function names are great for readability and clarity, and the primary functions should remain like that.  However, explicit names can be long and become burdensome for users.  Users either work with the long names or have to individually import each function and rename it to a more manageable name.  Thus, I propose we make aliases for, at least, the most common formulary functionality.  Implementation would looks something like...

```python
__all__ = ["Alfven_speed", "va_"]

...

def Alfven_speed(...):
    ...
    return speed

va_ = Alfven_speed
"""Alias to :func:`Alfven_speed`."""

# or

#: Alias for :func:`Alfven_speed`.
va_ = Alfven_speed
```
### Development points

* Aliases will have a trailing underscore to signify they are aliases and allow users to still use the shortest abbreviations for their code. For example,

  ```python
  from plasmapy.formulary import va_

  va = va_(...)
  ```
* Aliases shout be documented in the docstring of the parent function after the main description but before the parameters list.

  ```python
  def Alfven_speed(...):
      """
      Main function description

      **Aliases:**  `va_`

      Parameters
      ----------
      ...
      """
  ```

* Aliases should be tested in the the parent function `pytest`s.  This can be done with the following simple `assert` statement.

  ```python
  assert va_ is Alfven_speed
  ```

### List of functions to make aliases for...

(By no means is this a complete list and will be updated as needed.)

- [x] `Alfven_speec` -> `va_` (PR #865)
- [x] `Bohm_diffusion` -> `DB_`  (PR #865)
- [x] `deBroglie_wavelength` -> `lambdaDB_` (PR #865)
- [x] `Debye_length` -> `lambdaD_` (PR #865)
- [x] `Debye_number` -> `nD_` (PR #865)
- [x] `ExB_drift` -> `veb_` (PR #865)
- [x] `Fermi_energy` -> `Ef_` (PR #865)
- [x] `force_drift` -> `vfd_` (PR #865)
- [x] `gyrofrequency` -> `oc_` `wc_` (PR #865)
- [x] `gyroradius` -> `rc_` `rhoc_` (PR #865)
- [x] `inertial_length` -> `cwp_` (PR #865)
[...]

Issue #855, opened 2020-06-16
Aliases for common formulary function


StanczakDominik:
We've decided [at the telecon](https://docs.google.com/document/d/1H_TiB3mv1UKRdi6PNGS7nNUh8SYnpk8rDpu07jVU5gk/edit#) it would be handy to built-in aliases to the formulary functions:

```python
def plasma_frequency(...):
    return 42

wp = plasma_frequency
"""Alias for :fund:`plasma_frequency`."""
```

Because apparently people at BAPSF use it that way, and we'd like to make canonical aliases a thing!

This will mean we'll have to come up with aliases for the following functions:

- [ ] List the relevant functions!

rocco8773:
Whoops! Didn’t even see this here. You’re trigger finger is faster than mine.

StanczakDominik:
Glad to see all that starcraft practice paying off! :D

PR #1101, opened 2021-04-03
Limit aliases to formulary

StanczakDominik:
When we first discussed aliases in #855 and #856, I thought that we'd add them for the formulary only. In https://github.com/PlasmaPy/PlasmaPy/issues/856#issue-639988545, Erik wrote (emphasis mine):

> Explicit function names are great for readability and clarity, and the **primary functions should remain like that**. However, explicit names can be long and become burdensome for users. Users either work with the long names or have to individually import each function and rename it to a more manageable name. Thus, I propose we make **aliases for, at least, the most common formulary functionality**.

But while working on https://github.com/PlasmaPy/PlasmaPy/pull/1099 I found out about [`analysis.find_vf_`](https://docs.plasmapy.org/en/stable/api/plasmapy.analysis.swept_langmuir.find_vf_.html#plasmapy.analysis.swept_langmuir.find_vf_) and [`dispersion.tfds_`](https://docs.plasmapy.org/en/stable/api/plasmapy.dispersion.tfds_.html#plasmapy.dispersion.tfds_).

### Why I like formulary aliases and dislike them elsewhere

When working with the Formulary, we conceptually dive into the space of plasma physics concepts. It is reasonable to use a name such as `wp_` for the plasma frequency ω<sub>p</sub> because we're already in the mind state of expecting those. These expressions are likely to repeat in multiple other equations. https://github.com/PlasmaPy/PlasmaPy/pull/1059 or other equation-heavy codes are examples of a place where I might want to use the aliases. Reading and writing code like this, under such conditions, can be easier than with the full and explicit names. 
But these conditions are not fulfilled for general, "primary" functions, as noted by Erik. Neither `analysis` (yet?) nor `dispersion` induce a headspace of associations; and I don't think there is a potential context that induces you to say *"yes, it is obvious at first glance what tfds is"*.

I also suspect that these two functions would be unlikely to repeat in code as densely as formulary functions - sure, you might loop over parameters or datasets with them, but it's going to be rare to have multiple occurrences side by side.

Also, code is read more often than it is written; and ― this may not be codified anywhere, but I always thought one of the principles of our little project here was to promote people writing clean and readable research code. By making those aliases there, we directly promote their usage; and this works against that goal. People can, of course, define their own aliases; but I'd rather we didn't do much to promote it unless we have clear reasons to.

To sum up - I don't think it was the right call to introduce them. I thus propose their removal, and that we limit ourselves to the formulary from now on with officially sanctioned aliases.

The removal part is, of course, a breaking change; but I think it's better to cause their direct users some annoyance now rather than to doom generations of grad students to come to scratch their heads at what the heck `tfds` is... forever. This is, of course, sarcastic, but also not really. :wink:

### DIFF
--- a/changelog/1101.breaking.rst
+++ b/changelog/1101.breaking.rst
@@ -1,3 +1,3 @@
-Remove alias: `tfds` to `~plasmapy.dispersion.two_fluid_dispersion_solution`,
+Remove alias: `tfds_` to `~plasmapy.dispersion.two_fluid_dispersion_solution`,
 with the reasoning behind the removal outlined in `#1101
 <https://github.com/PlasmaPy/PlasmaPy/pull/1101#issue-608337879>`_
--- a/changelog/1101.breaking.rst
+++ b/changelog/1101.breaking.rst
@@ -1,4 +1,3 @@
-Remove aliases: `tfds` to `~plasmapy.dispersion.two_fluid_dispersion_solution`
-and `find_vf_` to `~plasmapy.analysis.swept_langmuir.find_floating_potential`,
+Remove alias: `tfds` to `~plasmapy.dispersion.two_fluid_dispersion_solution`,
 with the reasoning behind the removal outlined in `#1101
 <https://github.com/PlasmaPy/PlasmaPy/pull/1101#issue-608337879>`_
--- a/changelog/1101.breaking.rst
+++ b/changelog/1101.breaking.rst
@@ -1,0 +1,4 @@
+Remove aliases: `tfds` to `~plasmapy.dispersion.two_fluid_dispersion_solution`
+and `find_vf_` to `~plasmapy.analysis.swept_langmuir.find_floating_potential`,
+with the reasoning behind the removal outlined in `#1101
+<https://github.com/PlasmaPy/PlasmaPy/pull/1101#issue-608337879>`_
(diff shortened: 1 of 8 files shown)


rocco8773:
Ok, I can get behind this argument, especially for `tfds_`.  However, `vf` (floating potential), `vp` (plasma potential), `isat` (ion-saturation current), etc. are very common nomenclatures in the Langmuir analysis world.  They are used extensively in literature and are usually used in place of their long forms.  So, I feel these are still appropriate because they still satisfy the statement, "It is reasonable to use a name such as `wp_` for the plasma frequency ωp because we're already in the mind state of expecting those."

With that said, I'm okay with putting a restriction on aliases such that aliases are only given for functionality that could see frequent use and the aliases names can only use shorthand nomenclature associated with that field/context.

StanczakDominik:
Okay, I'm obviously a foreigner to that world, so I'll take your word on that. I'll revert the langmuir changes. Does that sound good?

changelog/1101.breaking.rst
```diff
@@ -0,0 +1,3 @@
+Remove alias: `tfds` to `~plasmapy.dispersion.two_fluid_dispersion_solution`,
```
rocco8773, line 1:
```suggestion
Remove alias: `tfds_` to `~plasmapy.dispersion.two_fluid_dispersion_solution`,
```

rocco8773 (approved):
Just that one typo in the change log, otherwise looks good to me.

PR #1111, opened 2021-04-15
Towncrier >= 19.2.0

rocco8773:
The ability to have multiple changelog entries for a given PR and changelog type was introduced in `v19.2.0`, so I'm limiting our requirements to that.  Additionally, `towncrier` uses that 3rd "entry" in `1062.trivial.1` as the counter, so we ought to included numbering on all `towncrier` changelog files if we're adding multiple ones.


StanczakDominik (approved):
Sure, let's do that :)

PR #1114, opened 2021-04-15
Index page updates

### LINKED ISSUE #1112: More examples on documentation main page?
The documentation main page currently only highlights three notebooks, but we have many more cool notebooks that could be featured here. I'd propose changing up the layout to include a grid of 3x3 examples. The "PlasmaPy is..." text could be larger font and go above them, along with the logo. The remaining table-of-contents is redundant with the sidebar, so I think it's fine for it to be pushed down out of view. I would also add a label "Highlights" or something to make clear that the examples don't represent all of the functionality in the package. 

The goal of this change would be to emphasize the diversity of functionality that already exists, similar to the way retail workers are taught to spread products out and pull them forward on the shelves to make them look really full.

image

StanczakDominik:
Closes #1112. Reviewable, but absolutely up for discussion on how this should look. I have not actually looked at the doc build yet; just made the changes I hope should work.

I don't actually know how to make the font size on "PlasmaPy is..." larger, though.


pheuer:
Maybe the PlasmaPy logo needs to be smaller so it sits inline with the header paragraph? But other than that, unsurprisingly, I like it!

index

pheuer:
I think that looks great! 

Non-Uniform grids is maybe not the best example of functionality: I'd vote for replacing that with the plasma dispersion function or something. But other than that I'm all for this change!

namurphy:
My main suggestion is that since mobile devices might show the example notebooks in a single column, we might want to have ∼3–6 example notebooks instead of 9.  In that case, we could include only the first proton radiography notebook.  It'd also particularly like to keep the E×B drift notebook in there too since it's something that comes up in intro to plasma courses pretty ubiquitously.  In any case, this is definitely an improvement!

docs/index.rst
```diff
@@ -2,26 +2,34 @@
...
-
-`PlasmaPy <http://www.plasmapy.org/>`_ is an open source
-community-developed core `Python <https://www.python.org/>`_ 3.7+
-package for plasma physics currently under development.
+   notebooks/analysis/swept_langmuir/find_floating_potential
+   notebooks/diagnostics/proton_radiography_particle_tracing_wire_mesh
+   notebooks/formulary/thermal_bremsstrahlung
+   notebooks/plasma/grids_nonuniform
+   notebooks/ExB_drift
+   notebooks/simulation/particle_stepper
```
namurphy, line 32:
```suggestion
   notebooks/analysis/swept_langmuir/find_floating_potential
   notebooks/ExB_drift
   notebooks/simulation/particle_stepper
```
In office hours today, @Tiger-Du, @svincena, and I talked about which notebooks to drop if we went down to six, and we're thinking these three...though it was a tough decision!  

The reason was that if viewed on a mobile device, the notebooks often show up all in a single column so you have to scroll down past all of them.

namurphy (approved):
LGTM!

PR #1121, opened 2021-04-20
Classical Transport 2.0 (WIP)

pheuer:
Work in progress, long-term project, help welcome. As may be obvious, I'm just learning transport theory now so please let me know if anything seems iffy. 

Development of a new classical transport module that re-organizes the existing `braginskii.py` file into something a little more extendable to include more models etc. The principles guiding development currently are: 

- Code
  - Compartmentalize each model in a separate file to keep the code compact.
  - Maintain identical (when possible) interface between models to facilitate comparisons.
  - Keep data out of the methods! All constants should be either organized into arrays at the beginning of the file or stored in separate `.npy` files as appropriate. 

- Documentation
  - Make clear in the docstring for each model sub-class what fluid equations are being solved and the corresponding Ohm's law and heat flow equations with the coefficients as given. 
  - Add warnings about the limitations/assumptions behind each model to the module docstring.

TODO: 
- [ ] Add support for ion thermal conductivity and viscosities for Braginskii 
- [ ] Re-implement Braginskii transport
- [ ] Re-implement Ji-Held
- [ ] Re-implement Spitzer-Harm
- [ ] Create an example notebook (or several) explaining the relationship between the models, the physical meaning of each coefficient, and comparing the models coefficient-by-coefficient. 
- [ ] Write tests that compare the models to each other (eg. to Spitzer-Harm at B=0) and to limiting cases.

### DIFF
--- a/plasmapy/transport/classical/__init__.py
+++ b/plasmapy/transport/classical/__init__.py
@@ -1,0 +1,16 @@
+"""
+The `~plasmapy.formulary` subpackage contains commonly used formulae
+from plasma science.
+"""
+# __all__ will be auto populated below
+__all__ = []
+
+from .base import *
+from .epperlein_haines import *
+
+
+# auto populate __all__
+for obj_name in list(globals()):
+    if not (obj_name.startswith("__") or obj_name.endswith("__")):
+        __all__.append(obj_name)
+__all__.sort()
--- a/plasmapy/transport/classical/epperlein_haines.py
+++ b/plasmapy/transport/classical/epperlein_haines.py
@@ -1,0 +1,241 @@
+"""
+Implementation of the magnetized transport coefficients presented by
+Epperlein and Haines in their 1986 paper "Plasma transport coefficients in a magnetic field by direct numerical solution of
+the Fokker–Planck equation" (Phys. Fluids), doi: 10.1063/1.865901
+"""
+
+__all__ = [
+    "EpperleinHainesCoefficents",
+    ]
+
+
+# TODO: allow options for returning transport coefficients in physical units
+# this will require more information about the plasma
+# look at sec. IIB in Ridgers2008 (pg 3) for relations: these likely hold for EH as well
+# EH equivalent is right at the end of section II (pg. 5). Looks the same.
+
+from abc import ABC
+import astropy.units as u
+import numpy as np
+import warnings
+
+from plasmapy.formulary.parameters import gyrofrequency
+from plasmapy import particles
+from plasmapy.particles import Particle
+
+from plasmapy.transport.classical.base import AbstractClassicalTransportCoefficients
+
+import matplotlib.pyplot as plt
+
+Ztable = np.array([1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 14, 20, 30, 60, np.inf])
+
+# TABLE II
+α0 = np.array([0.5061, 0.4295, 0.3950, 0.3750, 0.3618, 0.3524, 0.3454, 0.3399, 0.3319, 0.3263, 0.3221, 0.3144, 0.3081, 0.3015, 0.2945])
+α0p = np.array([1.37, 1.58, 1.68, 1.74, 1.78, 1.8, 1.82, 1.84, 1.87, 1.88, 1.9, 3.53, 5.49, 7.61, 9.17])
+α1p = np.array([3.03, 3.21, 3.17, 3.15, 3.14, 3.13, 3.12, 3.11, 3.1, 3.1, 3.09, 3.52, 3.97, 4.41, 4.73])
+a0p = np.array([2.77, 2.78, 2.78, 2.78, 2.78, 2.79, 2.79, 2.79, 2.79, 2.8, 2.8, 5.14, 7.94, 10.9, 13])
+a1p = np.array([6.72, 6.7, 6.47, 6.37, 6.33, 6.29, 6.26, 6.23, 6.21, 6.2, 6.19, 7.97, 10, 12.2, 13.8])
+α0x = np.array([0.1989, 0.3285, 0.41457, 0.4794, 0.5285, 0.5676, 0.5996, 0.6264, 0.6686, 0.7004, 0.7254, 0.7757, 0.8209, 0.8727, 0.9328])
+α0pp = np.array([2.66e2, 4.91e2, 6.3e2, 7.06e2, 7.23e2, 7.57e2, 7.94e2, 8.17e2, 8.89e2, 9.62e2, 9.88e2, 1.06e3, 1.17e3, 1.3e3, 1.33e3])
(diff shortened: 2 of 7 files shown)


plasmapy/transport/classical/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from .base import *
+from .epperlein_haines import *
+
+
+# auto populate __all__
+for obj_name in list(globals()):
+    if not (obj_name.startswith("__") or obj_name.endswith("__")):
+        __all__.append(obj_name)
+__all__.sort()
```
StanczakDominik, line 16: Isn't the result of this what you get if you don't manually define `__all__` at all and just do the star imports?

plasmapy/transport/classical/epperlein_haines.py
```diff
@@ -0,0 +1,238 @@
...
+γ1p = np.array([4.66, 3.96, 3.72, 3.6, 3.53, 3.49, 3.49, 3.43, 3.39, 3.37, 3.35, 3.32, 3.3, 3.27, 3.25])
+c0p=np.array([1.93, 1.89, 1.66, 1.31, 1.12, 1.04, 1.02, 0.875, 0.77, 0.722, 0.674, 0.605, 0.566, 0.502, 0.457])
+c1p=np.array([2.31, 3.78, 4.76, 4.63, 4.62, 4.83, 5.19, 4.74, 4.63, 4.7, 4.64, 4.65, 4.81, 4.71, 4.81])
+c2p= np.array([5.3, 7.78, 8.88, 8.8, 8.8, 8.96, 9.24, 8.84, 8.71, 8.73, 8.65, 8.6, 8.66, 8.52, 8.53])
+γ0x = np.array([6.071, 15.75, 25.65, 34.95, 43.45, 51.12, 58.05, 64.29, 75.04, 83.93, 91.38, 107.8, 124.3, 145.2, 172.7])
+γ0pp = np.array([4.01, 2.46, 1.13, 0.628, 0.418, 0.319, 0.268, 0.238, 0.225, 0.212, 0.202, 0.200, 0.194, 0.189, 0.186])
+γ1pp = np.array([2.5] * 15)
+c0pp = np.array([.661, .156, .0442, .018, .00963, .00625, .00461, .00371, .003, .00252, .00221, .00185, .00156, .0013, .00108])
+c1pp = np.array([.931, .398, .175, .101, .0702, .0551, .0465, .041, .0354, .0317, .0291, .0256, .0228, .0202, .018])
+c2pp = np.array([2.5, 1.71, 1.05, .775, .646, .578, .539, .515, .497, .482, .471, .461, .45, .44, .43])
```
StanczakDominik, line 73: Yup, that's definitely transport theory!

plasmapy/transport/classical/epperlein_haines.py
```diff
@@ -0,0 +1,238 @@
...
+        -----
+        For details, see "Plasma transport coefficients in a magnetic field by 
+        direct numerical solution of the Fokker–Planck equation" by
+        E. M. Epperlein and M. G. Haines, DOI: `10.1063/1.865901`
+    
+        .. _`10.1063/1.865901`: https://aip.scitation.org/doi/10.1063/1.865901
+    
+        """
+        i = _find_nearest(self.Z)
+        return α0[i]*np.ones(self.chi.size)
```
StanczakDominik, line 134: So this is essentially nearest neighbor interpolation, right? Might cause nonphysical discrete jumps as you sweep the parameter space. Wouldn't we want something like scipy splines instead? Polynomial fits?

plasmapy/transport/classical/epperlein_haines.py
```diff
@@ -0,0 +1,238 @@
...
+    ax.set_ylim(0.001, 100)
+    ax.plot(chi, coef1.norm_kappa_perp())
+    ax.plot(chi, coef2.norm_kappa_perp())
+    
+    ax = axarr[2][1]
+    ax.set_title('$\\kappa_\\wedge$')
+    ax.set_ylim(0.01, 10)
+    ax.plot(chi, coef1.norm_kappa_cross())
+    ax.plot(chi, coef2.norm_kappa_cross())
+
```
StanczakDominik, line 238: Ah. Looks like a notebook in the `__main__`. :)

StanczakDominik (commented):
Quick look through!

plasmapy/transport/classical/__init__.py
pheuer: I've no idea, I just copied this from the formulary module!

plasmapy/transport/classical/epperlein_haines.py
pheuer: One model down...many to go 🤦

plasmapy/transport/classical/epperlein_haines.py
pheuer: Yeah I wanted to raise that question. I do want there to be some way to extract exactly the data from the paper (for historical purposes?). But some sort of fits in Z would be good. 

For models where we have the actual data (transport coefficients tabulated as functions of chi and Z rather than polynomial fits) I have some code that just interpolates the values from that using scipy's `interp2d`. I suppose we could generate these tables from the polynomial fits and then apply the same method here?

StanczakDominik: As long as that doesn't introduce any heavy error terms... sure, should be good enough!

plasmapy/transport/classical/epperlein_haines.py
pheuer: Yep, eventually I'll move it to one!

plasmapy/transport/classical/epperlein_haines.py
pheuer: Do you mean relative to the original data? Sure, we can check that.

pheuer: Another option would be to implement both the explicit polynomials (exactly as given in the paper) and the fit, eg. `EpperleinHainesCoefficients` and `EpperleinHainesCoefficientsPolynomialFit`. Then we could write tests to ensure that the interpolated values in the former were within some margin of the fits given in the papers?

plasmapy/transport/classical/epperlein_haines.py
```diff
@@ -0,0 +1,805 @@
...
+            )
+        )
+
+
+if __name__ == "__main__":
+
+    import astropy.units as u
+
+    coef = EpperleinHaines.dimensional(
+        particle="H+", B=10 * u.T, n_e=1e19 * u.cm**-3, T_e=100 * u.eV
```
namurphy, line 715:
```suggestion
        particle="p+", B=10 * u.T, n_e=1e19 * u.cm**-3, T_e=100 * u.eV
```

PR #1122, opened 2021-04-21
Pickleable particles

### LINKED ISSUE #1011: Pickle-able Particles (and other objects, for parallelization)

StanczakDominik:
Closes #1011!

I do not fully understand why this works. But it seems that it does.

I was tinkering with slicing of IonizationState as part of #1059 (@namurphy I have actual arguments for that, and it's just the first of many travesties I want to do to IonizationState). I then somehow stumbled upon [this](https://docs.python.org/3/library/pickle.html#pickle-picklable:~:text=instances%20of%20such%20classes%20whose%20__dict__,section%20Pickling%20Class%20Instances%20for%20details), that got me thinking, I started looking at `Particle.__dict__`, noticed that `json.dumps(particle)` complains about `_PeriodicTable` contained within that `__dict__`, realized that's a namedtuple, thought hey, if namedtuples have trouble pickling (I don't understand why they would, that's my whole problem with this solutin!), maybe dataclasses wouldn't. And guess what, they don't. And the xfailing tests started passing.

@pheuer could you look at that issue you had previously with scipy.optimize, maybe even add a smaller test case of that somewhere as a failing (`pytest.mark.xfail`)? I'd like to check again. Maybe this got fixed.

### DIFF
--- a/changelog/1122.bugfix.rst
+++ b/changelog/1122.bugfix.rst
@@ -1,0 +1,1 @@
+Made `~plasmapy.particles.Particle` instances pickleable.
--- a/plasmapy/particles/elements.py
+++ b/plasmapy/particles/elements.py
@@ -14,9 +14,15 @@ import collections
 import json
 import pkgutil
 
-_PeriodicTable = collections.namedtuple(
-    "periodic_table", ["group", "category", "block", "period"]
-)
+from dataclasses import dataclass
+
+
+@dataclass
+class _PeriodicTable:
+    group: int
+    period: int
+    block: str
+    category: str
 
 
 def _element_obj_hook(obj):
--- a/plasmapy/particles/tests/test_pickling.py
+++ b/plasmapy/particles/tests/test_pickling.py
@@ -19,17 +19,15 @@ class TestPickling:
     Test that different objects in `plasmapy.particles` can be pickled.
     """
 
-    xfail = pytest.mark.xfail(reason="see issue #1011")
-
     @pytest.mark.parametrize(
         "instance",
         [
             CustomParticle(mass=1 * u.kg, charge=1 * u.C),
             DimensionlessParticle(mass=5, charge=5),
-            pytest.param(Particle("p+"), marks=xfail),
-            pytest.param(IonicLevel("p+", 0.1, 1e9 * u.m ** -3), marks=xfail),
-            pytest.param(IonizationState("H", [0.5, 0.5]), marks=xfail),
-            pytest.param(IonizationStateCollection({"H": [0.5, 0.5]}), marks=xfail),
+            pytest.param(Particle("p+")),
+            pytest.param(IonicLevel("p+", 0.1, 1e9 * u.m ** -3)),
+            pytest.param(IonizationState("H", [0.5, 0.5])),
+            pytest.param(IonizationStateCollection({"H": [0.5, 0.5]})),
         ],
     )
     def test_pickling(self, instance, tmp_path):
--- a/plasmapy/particles/tests/test_pickling.py
+++ b/plasmapy/particles/tests/test_pickling.py
@@ -24,10 +24,10 @@ class TestPickling:
         [
             CustomParticle(mass=1 * u.kg, charge=1 * u.C),
             DimensionlessParticle(mass=5, charge=5),
-            pytest.param(Particle("p+")),
-            pytest.param(IonicLevel("p+", 0.1, 1e9 * u.m ** -3)),
-            pytest.param(IonizationState("H", [0.5, 0.5])),
-            pytest.param(IonizationStateCollection({"H": [0.5, 0.5]})),
+            Particle("p+"),
+            IonicLevel("p+", 0.1, 1e9 * u.m ** -3),
+            IonizationState("H", [0.5, 0.5]),
+            IonizationStateCollection({"H": [0.5, 0.5]}),
         ],
     )
     def test_pickling(self, instance, tmp_path):


StanczakDominik:
Having had time to think this through on the evening doggo walk,

Nope, I got nothing.

StanczakDominik:
It should! If it's more complex than necessary, don't bother :)

pheuer:
So, my attempts at writing a test function that would demonstrate the issue with `differential_evolution` in `lmfit`, parallelization, and particles didn't work. I did try just merging this over onto the `thomson_fitting` branch. That did clear the old error, which was : 


```
PicklingError: Can't pickle <class 'plasmapy.particles.elements.periodic_table'>: attribute lookup periodic_table on plasmapy.particles.elements failed
```

But resulted in this new error

```
    ret = differential_evolution(self.penalty, _bounds, **fmin_kws)
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\site-packages\scipy\optimize\_differentialevolution.py", line 307, in differential_evolution
    constraints=constraints) as solver:
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\site-packages\scipy\optimize\_differentialevolution.py", line 501, in __init__
    self._mapwrapper = MapWrapper(workers)
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\site-packages\scipy\_lib\_util.py", line 373, in __init__
    self.pool = Pool()
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\context.py", line 119, in Pool
    context=self.get_context())
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\pool.py", line 176, in __init__
    self._repopulate_pool()
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\pool.py", line 241, in _repopulate_pool
    w.start()
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\popen_spawn_win32.py", line 46, in __init__
    prep_data = spawn.get_preparation_data(process_obj._name)
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\spawn.py", line 143, in get_preparation_data
    _check_not_importing_main()
  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\multiprocessing\spawn.py", line 136, in _check_not_importing_main
    is not going to be frozen to produce an executable.''')
RuntimeError: 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...
```

In summary, I think this fixes the pickling issue, although I'm not sure if the new error is related to your fix or was just always waiting after the first error.

StanczakDominik:
I'd still appreciate the code for that! If it's doing complex setup beforehand, that's fine; so long as it reproduces. Maybe I can pinpoint the reason why it's crashing.

pheuer:
To test in-vivo in the thomson fitting, just set `fit_kws={'workers':-1}` in `run_fit()` in `test_thomson.py`. That turns on the built-in parallelization. 

But, I think I got a test working! The code below fits two models: one with a Particle in it and one without. On the main branch version, only the first fit executes, then the same "Not pickleable" error is raised during the second one. But, on your branch, both execute! This could be rearranged into a test, but it's computationally intensive and I think redundant with the pickling tests that already exist so I'd be inclined to leave it out of the official tests? 

Don't worry about that second error above: that was my fault (not a problem with the code).

```python
from lmfit import Parameters, Model

def _mock_model(x, settings=None, **params):
    particle = settings['particle']
    a = params['a']
    b = params['b']
    return particle.integer_charge*np.exp(-(x-a)**2/b**2)


def _mock_model2(x, settings=None, **params):
    a = params['a']
    b = params['b']
    return np.exp(-(x-a)**2/b**2)


def test_differential_evolution_oarallelization_with_particles():
    x = np.linspace(1, 5, num=1000)
    particle = Particle('H+')
    
    a = 3
    b = 1
    
    ydata = particle.integer_charge*np.exp(-(x-a)**2/b**2)*np.random.normal(loc=1.0, scale=0.01, size=x.size)
    
    
    params = Parameters()
    params.add('a', 5, min=0.1, max=10, vary=True)
    params.add('b', 5, min=0.1, max=10, vary=True)
    
    fit_kws = {'workers':-1, 'updating':'deferred'}

    
    # TEST 1: Create model without particle. This should work in parallel
    model = Model(_mock_model2, intependent_vars=["x"],
                  settings=None)
    
    
    res = model.fit(ydata, params,  x=x, fit_kws=fit_kws, method='differential_evolution')
    print(res.best_values)
    
    
    # TEST 2: Create model with the particle. Should fail if particles cannot
    # be pickled
    model = Model(_mock_model, intependent_vars=["x"],
                  settings={'particle':particle})
    
    res = model.fit(ydata, params,  x=x, fit_kws=fit_kws, method='differential_evolution')
    print(res.best_values)
```

StanczakDominik:
Yeah, I suspect it's fine to leave it out. I can confirm that both pass on my branch.

Shall we merge it, then? :)

pheuer:
Sounds good to me! What does @namurphy think?

namurphy (approved):

PR #1105, opened 2021-04-10
A version of automodapi for PlasmaPy

rocco8773:
This PR creates the sphinx extension package `plasmapy_sphinx`.  This package defines two reST directives `automodapi` and `automodsumm`, which are an adaption of the likely named directives from `sphinx_automdoapi`.  There are a few significant distinctions of our directives over the `sphinx_automodapi` directives that made me add the code here and not push it upstream to `sphinx_automodapi`.

1. I introduced the concept of groups for the directives.  So, when an module is inspected for its objects, the found objects are categorized into group types (modules, functions, classes, exceptions, etc.).  As a result, the `sphinx_automodapi` options like `:functions-only:` are replaced by `:groups: functions`.
2. This move to the groups options allows for multiple groups to be displayed at once for `automodsumm` by doing

    ``` rst
    .. automodsumm:: foo.bar
       :groups: functions, classes

    .. automodapi:: foo.bar
       :groups: functions, classes
    ```

    It additionally allows for custom groups to be defined.
3. Configuration values (that are defined in ``conf.py``) are introduced that allows you define custom groups.  To define a custom group the `automodapi_custom_groups` configuration value needs to be defined like

    ```python
    automodapi_custom_groups = {
        "mygroup" : {
            "title": "My Custum Group",
            "dunder": "__mygroup__",
        },
    }
    ```

    In this example `mygroup` would be the name of the custom group, `"title"` would be the title rendered above the `autosummary` table when `automodapi` is used, and `"dunder"` is the name of the dunder variable used to list the custom group objects.  This `__mygroup__` would be placed in the module being documented with the custom group and is defined just like `__all__`.
4. The `automodapi` directive is an official `autodoc` directive and not a function connected to the [`source-read`](https://www.sphinx-doc.org/en/master/extdev/appapi.html?highlight=source-read#event-source-read) event.  This allows `automodapi` to benefit from all the default functionality associated with the `autodoc` directives.

Issue #779, opened 2020-04-07
CI for plots


StanczakDominik:
> An alternative would be to do a regression test and make sure that the hash of the plot object stays the same.  I think this is a strategy that SunPy has used.  The problem is that this seems to be a fragile test, since the hash could change due to upstream changes.  I'm not really sure on the best way to deal with testing plots...

_Originally posted by @namurphy in https://github.com/PlasmaPy/PlasmaPy/pull/675/files#r404113800_

also see the rest of that thread

StanczakDominik:
https://matplotlib.org/3.2.1/devel/testing.html#writing-an-image-comparison-test looks pretty darn helpful

StanczakDominik:
ping @pheuer since we talked about it in #1123 : I took a look at how SunPy does it using https://github.com/matplotlib/pytest-mpl; they've got some semi-complex infrastructure to do it, but seem to be still in the process of figuring this out. For example, one issue is fonts: you need the same font file versions locally as you've got remotely, or any image based comparison will complain about the fonts being different. I'll wait a month or two for it to stabilize, then start tinkering with it.

pheuer:
Comparing the actual plots seems unnecessarily complicated to me (unless there's a use case I'm not seeing). The hash would be extremely sensitive to very small variations, which would be a problem for anything with a stochastic component (like the proton radiographs. 

I think it makes more sense to think of a way of testing 1D and 2D arrays for "near equality" in some sense. For example, maybe spline fitting the data then comparing the spline fits? Or comparing FFT spectra if horizontal shifts are a concern? Something that approximates what we do by eye: comparing the output to previous output and judging that it hasn't substantially changed?

PR #1123, opened 2021-04-23
Path/Line Integrated Diagnostic (Abstract classes + interferometer implementation)

pheuer:
This PR introduces a new abstract class "PathIntegratedDiagnostic" to the diagnostics module. This class abstracts the source/grid/detector geometry already used in `SyntheticProtonRadiography`. I've refactored the `SyntheticProtonRadiograph` module to be a subclass of the new class, but there are no API changes.

This PR also introuduces a subclass of `PathIntegratedDiagnostic`, `LineIntegratedDiagnostic`, which represents the case where the integration paths are straight lines. Given a `grid` object, this class can line-integrate through any quantity on the grid along any arbitrary straight line of sight. The quantity integrated is given by a "_integrand" method, which can be in turn be calculated using any combination of the quantities on the grid. Different diagnostics with non-trivial response functions (like proton radiography) can be implemented by just replacing this `_integrand` in subclasses.

This module could has many applications (x-ray radiography, linear-regime charged particle radiography, etc.), but for a start I've implemented a very basic interferometry function. Given a 1 mm radius sphere of constant density 3e19 cm^-3, the following is the predicted interferogram with a 263 nm probe beam

interferogram

The documentation still needs some work (I'll write one notebook explaining the interferometry subclass and another showing how to create and use custom subclasses). But I'm ready for comments on the basic structure of the `LineIntegratedDiagnostic` class!


pheuer:
This PR has a couple errors in the checks that I don't understand: 

```
./plasmapy/diagnostics/line_integrated_diagnostic.py:62:1: RST301 Unexpected indentation.
    def _log(self, msg):
^
./plasmapy/diagnostics/line_integrated_diagnostic.py:214:1: RST215 Inline interpreted text or phrase reference start-string without end-string.
        return (xax * u.m, yax * u.m, *integral)
^
./plasmapy/diagnostics/line_integrated_diagnostic.py:272:1: RST215 Inline interpreted text or phrase reference start-string without end-string.
        # Continue with the rest of the parent class init
^
2     RST215 Inline interpreted text or phrase reference start-string without end-string.
1     RST301 Unexpected indentation.
```

StanczakDominik:
Hey, sorry, that's flake8 being overzealous after #1062 ; ignore the flake8 for now, please :( I'll get to it when I'm less swamped :(

StanczakDominik:
All right, things have calmed down, I'll give this a thorough review tomorrow. Still need to recharge my batteries a little :(

pheuer:
Sounds good - no rush though! 

Note that I've changed the naming a bit: 

"PathIntegratedDiagnostic" -> Encapsulates the source/detector/grid geometry setup.

"LineIntegratedDiagnostic" -> Subclass for specifically the linear-integration regime, include `_line_integral()`.

docs/notebooks/diagnostics/line_integrated_diagnostic.ipynb
```diff
@@ -0,0 +1,341 @@
+{
```
StanczakDominik, line 1: <p>Just starting out by reading the notebook for now, but - I suppose the use of CartesianGrid here is more along the lines of "it's set up to work with CartesianGrid right now", rather than "It will never take a non-uniform grid"? Perhaps we could reword it to:</p><p> </p><p>"by providing a Grid object, representing the plasma to be measured (currently only CartesianGrid is supported)"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/1123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

StanczakDominik, line 1: <p>My first impression when reading this whole heading was "Oh, I have to create a custom class to use this? I wonder if we could not make this easier to use somehow" - and then the last sentence up-ends my understanding! Perhaps we could reword the title to something like "Example: custom Interferometer class", and in the beginning of this part say something about how this is mostly a didactic example so people can wrap their head around the pre-made <code>Interferometer</code> and the whole LineIntegratedDiagnostic framework easier?</p><p> </p><p>It would also be helpful to point people to a place in the docs where they can find e.g. a schema for the arguments of <code>evaluate</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/1123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

StanczakDominik, line 1: <p>Could do with a title noting that the plot is a projection to a plane perpendicular to the line between the source and detector.</p><p> </p><p>Eh, you know what we should do, actually? Add pyvista as at least a docs dependency. It's 2021, let's do 3D plotting already. I volunteer to do that and modernise our plots after I wrap up with my thesis.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/1123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

StanczakDominik, line 1: <p>It would be very nice to have a repeat of the same thing with a higher resolution detector plate, and an explanation of how the pixelated distribution is the result of the chosen resolution.</p><p> </p><p>By the way, if you end the last line with </p><p><code>cb.ax.set_ylabel("Phase Shift / $\pi$", fontsize=16); #note the semicolon</code></p><p> </p><p>It will suppress the output of <code>Text(0, 0.5, 'Phase Shift / $\\pi$')</code>  as the last object returned from the notebook cell.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/1123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

StanczakDominik, line 1: <p>As a newbie to interferometry, and speaking for all newbies to interferometry to come after me - some more explanation of the unwrapping could be handy! :D I have a hunch about how it works, and I expect diving through the code will confirm it, but just noting a note of confusion down.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/1123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
```diff
@@ -0,0 +1,173 @@
...
+    size = np.array([[-2, 2], [-2, 2]]) * u.mm
+    bins = [350, 350]
+
+    hax, vax, phase = obj.evaluate(
+        1.14e15 * u.Hz,
+        size=size,
+        bins=bins,
+        num=100,
+        unwrapped=False,
+    )
```
StanczakDominik, line 162: Just checking that this evaluates?

Would it help if we had mechanisms in place to verify that our tests keep returning the same image as they did previously? I'm asking because I've been putting #779 off for a while (started dabbling it a little while ~~frustrated with~~ working on https://github.com/PlasmaPy/PlasmaPy/pull/1059 ), and I could get back to it if it made this kind of test easier.

plasmapy/diagnostics/path_integrated_diagnostic.py
```diff
@@ -0,0 +1,412 @@
...
+            Number of bins in each direction of the detector region. The
+            default is [50,50].
+
+        collimated : Boolean, optional
+            If True, the source will be assumed to be collimated. If False,
+            a point source will be used. The default is True (collimated).
+
+        num : int, optional
+            Number of integration points along the line (within the grid region).
+            The default is 100.
```
StanczakDominik, line 185: Seems like a potentially fragile default...? I imagine if someone has steep gradients, it might cause incorrect results.

StanczakDominik (commented):
Looks good on the architecture side! A couple questions above about the notebook, and a few here.

docs/notebooks/diagnostics/line_integrated_diagnostic.ipynb
pheuer: Actually, this should already work for non-uniform grids! CartesianGrids are just so much easier to set up. But maybe I should add an example with a non-uniform grid too...

pheuer: Ah, good point! I've added the following to the beginning of that section to hopefully head off the confusion: 

```
A [LineIntegratedDiagnostic] synthetic diagnostic for interferometry ([Interferometer]) is already implemented in PlasmaPy. However, in this notebook we will re-create this class in order to demonstrate how a custom subclass of [LineIntegratedDiagnostic] can be used to create a new synthetic diagnostic.
```

There should probably be a separate (short) notebook explaining the actual `Interferometer` class? That would hopefully also help dispel the confusion and provide a more natural place to document the `evaluate` method.

pheuer: I'm not sure I follow this comment. I agree making that more clear is a good, but is a plot title the best way? Rather than putting it in the text?

pheuer: Thanks for the semicolon tip! 

Unfortunately this is already pretty high resolution: I'll double it but it doesn't look THAT much better. The restriction is the size of the array that we need to evaluate the integrand at: for a [h#, v#] detector plane and num=[n#] integration points that's an [h#, v#, n#] array, so it can become slow if the resolution gets high.

StanczakDominik: Ah, in that case, let's just reword it! :)

StanczakDominik: It's probably fine to keep the example in this notebook. If/once it grows out, we can add another one.

StanczakDominik: Ah, sorry. The plot title was just the first thing that came to mind. Some way of communicating "this situation is 3D, but the reason we're seeing a 2D plot is that we're projecting to the XZ (IIRC) plane" would be great :)

pheuer: So, basically you just remove discontinuities of 2pi using `np.unwrap`. 

I think this belongs in the Interferometer notebook, so I'm starting that now and I'll manually unwrap the signal there as an example.

StanczakDominik: In that case we could go the other way - start with a smaller resolution and compare with the current one?

plasmapy/diagnostics/path_integrated_diagnostic.py
pheuer: Maybe it makes sense to just make this a required input? 100 could be overkill for some grids or insufficient for others. Or it could default to `npmax(grid.shape)` or something?

docs/notebooks/diagnostics/line_integrated_diagnostic.ipynb
pheuer: Ok, I'll keep it like that for now. Eventually there will be a bunch of these LineIntegratedDiagnostics, so then it may make sense for them each have a notebook?

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
```diff
@@ -0,0 +1,173 @@
...
+
+    assert np.allclose(line, theory, atol=5e-4)
+
+
+def test_constant_box(grid):
+    x, y, z = grid.grid.T
+    t1 = np.where(np.abs(x) < 1 * u.mm, 1, 0)
+    t2 = np.where(np.abs(z) < 1 * u.mm, 1, 0)
+    field = np.where(t1 * t2 != 0, 1, 0) * u.kg / u.m ** 3
+    grid.add_quantities(rho=field)
```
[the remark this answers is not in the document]
pheuer: What method? Seems to be working to me!

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
pheuer: Yes I would use that all the time! Ideally it would include some sort of `np.isclose` rtol keyword or something. As is I've been writing custom checks based on the geometry of each image and I just haven't thought of one for this one yet.

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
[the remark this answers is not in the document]
StanczakDominik: I mean `grid.add_quantities(rho=field)` if rho already exists! Glad it's working, though!

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
StanczakDominik: All right, I'll try to work on that this weekend, then :)

plasmapy/diagnostics/tests/test_path_integrated_diagnostic.py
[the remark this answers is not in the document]
pheuer: Oh, yes I think that should just overwrite whatever was there first!

StanczakDominik:
Whoa. Take a look at the RTD error. The notebook kernel is dying. Looks like it's high time we pre-executed the longer-running simulation-based notebooks...

StanczakDominik:
As a stop gap solution, I commit to rerunning them on a monthly basis to check if anything's broken. I imagine I'll be pissed off enough to find an automatic solution within 3 months or so. :D

pheuer:
Thanks @StanczakDominik. Do you or @rocco8773 know why CI Documentation is raising the following error still? 

```
Warning, treated as error:
/home/runner/work/PlasmaPy/PlasmaPy/docs/ad/diagnostics/path_integrated_diagnostic.rst:9:An option to automodapi is either unknown or has an invalid value: 'no-heading'
```

The offending line 
```
.. automodapi:: plasmapy.diagnostics.path_integrated_diagnostic
   :no-heading:
```

It looks like `:no-heading:` is in all the files, so I don't know why it's causing a problem here?

StanczakDominik:
I don't know why it gave a warning in f714472, but I'm not seeing any of those on master, so I think @rocco8773 dropped it with #1105. I got rid of it; let's see how it renders.

plasmapy/diagnostics/path_integrated_diagnostic.py
```diff
@@ -0,0 +1,627 @@
...
+        location and the size and bins keywords.
+
+        2) For each cell of detector grid, create an array of points of
+        separation ds from there to the source point (or, when collimated=True,
+        the source plane).
+
+        3) Evaluate the integrand function at each point.
+
+        4) Integrate along the line-integrated dimension to obtain the
+        line-integrated quantity in the detector plane.
```
rocco8773, line 498: This is probably causing the new errors since the line continuation is suppose to indent.

plasmapy/diagnostics/path_integrated_diagnostic.py
```diff
@@ -0,0 +1,627 @@
...
+        bins=[200, 200],
+        collimated=True,
+        unwrapped=True,
+    ):
+        r"""
+        Creates an interferogram by calculating the line integral through
+        the electron number density :math:`n_e` provided on the grid. The phase shift
+        is
+
+        .. math:: \Delta \phi = -\frac{\omega_{probe}}{2 c n_c} \int n_e dl
```
rocco8773, line 556: If you don't want to make this  raw string, then you'll have to put an escape character on math functions; that is, 

```rst
 .. math:: \\Delta \\phi = -\\frac{\\omega_{probe}}{2 c n_c} \\int n_e dl
```

plasmapy/diagnostics/path_integrated_diagnostic.py
```diff
@@ -0,0 +1,627 @@
...
+
+    def evaluate(
+        self,
+        num,
+        size=None,
+        bins=[200, 200],
+        collimated=True,
+    ):
+
+        r"""
```
rocco8773, line 484:
```suggestion
    ):
        r"""
```

pheuer:
@StanczakDominik This error is appearing intermittently 
```
FAILED plasmapy/formulary/tests/test_Chandrasekhar_G.py::test_Chandrasekhar_with_hypothesis - assert 0.25 <= 0.21399915915288345
```

StanczakDominik:
Welp that's on me. Feel free to ignore it and I will deal with it tomorrow

plasmapy/diagnostics/path_integrated_diagnostic.py
pheuer: Oh ok, done

docs/notebooks/diagnostics/line_integrated_diagnostic.ipynb
```diff
@@ -0,0 +1,1032 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "weekly-howard",
+   "metadata": {},
+   "source": [
+    "[LineIntegratedDiagnostic]: ../../api/plasmapy.diagnostics.path_integrated_diagnostic.LineIntegratedDiagnostic.rst#plasmapy.diagnostics.path_integrated_diagnostic.LineIntegratedDiagnostic\n",
+    "\n",
+    "[Grid]: ../../api/plasmapy.plasma.grids.rst#plasmapy.plasma.grids\n",
```
rocco8773, line 71:
```suggestion
    "[Grid]: ../../api_static/plasmapy.plasma.grids.rst\n",
```

This will get rid of the next doc build error.

docs/notebooks/diagnostics/line_integrated_diagnostic.ipynb
```diff
@@ -0,0 +1,1032 @@
...
+   "source": [
+    "## Interferometer"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dfaa16bd",
+   "metadata": {},
+   "source": [
+    "[Interferometer]: ../../api/plasmapy.diagnostics.path_integrated_diagnostic.interferometry.Interferometer.rst#plasmapy.diagnostics.path_integrated_diagnostic.interferometry.Interferometer\n",
```
rocco8773, line 99:
```suggestion
    "[Interferometer]: ../../api/plasmapy.diagnostics.path_integrated_diagnostic.Interferometer.rst#plasmapy.diagnostics.path_integrated_diagnostic.Interferometer\n",
```

This should fix the last error.

pheuer:
Thanks for the help, documentation wizard @rocco8773 !

StanczakDominik (approved):
All right, this looks great to me and I could see it being merged - unless @rocco8773 wants to take another sweep, since this falls under diagnostics? (if no, just hit merge!)

Also, is it just me, or does this look like the Batman logo?

pheuer:
Great! I don't see the batman logo even if I squint really really hard...I think you might need some sleep @StanczakDominik  ;)

Issue #1116, opened 2021-04-19
Iteratively enable flake8 checks


StanczakDominik:
#1062 brought back flake8 linter checks, but kept most of them disabled. There are a bunch of them in setup.cfg:

https://github.com/PlasmaPy/PlasmaPy/blob/4a71ade0a7c0e3ef951a7096201f8ab22fc0c7c9/setup.cfg#L96-L185

It'd be nice to have them back. This would mean, for any brave soul willing to undertake this, for every check marked as TODO (and probably some others, too):

- enable the check again by removing it from the exclude list;
- fix all the lint errors that flake8 (in the tests section) starts throwing at you;
- continue down the list for as long as you'd like.

This is, admittedly, not exciting, but quite helpful and simple to execute; would be great for a new contributor.

aksalcido:
Hello Dominik, I emailed you and I would like to work on this :)

StanczakDominik:
Go right ahead! 😄

PR #1127, opened 2021-04-30
Enable flake8 checks

aksalcido:
Removed majority of the flake8 checks listed in #1116 under # TODO. Also edited the corresponding files that were throwing the flake8 checks once they were no longer being ignored. The only error code that was not handled completely was F541, only some of the F541 errors were handled, but not all. The remaining errors for F541 is for strings that begin with fr"string". Was a bit confused on if this was an intentional design so I wanted clarification before I handled them. I wanted feedback on if this seems good since this is my first contribution to Open Source and feel a bit anxious on if I was working on it incorrectly on my end.

      docstrings.

### DIFF
--- a/plasmapy/formulary/__init__.py
+++ b/plasmapy/formulary/__init__.py
@@ -5,18 +5,18 @@ from plasma science.
 # __all__ will be auto populated below
 __all__ = []
 
-from .braginskii import *
-from .collisions import *
-from .dielectric import *
-from .dimensionless import *
-from .distribution import *
-from .drifts import *
-from .ionization import *
-from .magnetostatics import *
-from .mathematics import *
-from .parameters import *
-from .quantum import *
-from .relativity import *
+from .braginskii import *  # noqa: F403
+from .collisions import *  # noqa: F403
+from .dielectric import *  # noqa: F403
+from .dimensionless import *  # noqa: F403
+from .distribution import *  # noqa: F403
+from .drifts import *  # noqa: F403
+from .ionization import *  # noqa: F403
+from .magnetostatics import *  # noqa: F403
+from .mathematics import *  # noqa: F403
+from .parameters import *  # noqa: F403
+from .quantum import *  # noqa: F403
+from .relativity import *  # noqa: F403
 
 # auto populate __all__
 for obj_name in list(globals()):
--- a/plasmapy/formulary/__init__.py
+++ b/plasmapy/formulary/__init__.py
@@ -5,18 +5,18 @@ from plasma science.
 # __all__ will be auto populated below
 __all__ = []
 
-from .braginskii import *  # noqa: F403
-from .collisions import *  # noqa: F403
-from .dielectric import *  # noqa: F403
-from .dimensionless import *  # noqa: F403
-from .distribution import *  # noqa: F403
-from .drifts import *  # noqa: F403
-from .ionization import *  # noqa: F403
-from .magnetostatics import *  # noqa: F403
-from .mathematics import *  # noqa: F403
-from .parameters import *  # noqa: F403
-from .quantum import *  # noqa: F403
-from .relativity import *  # noqa: F403
+from .braginskii import *
+from .collisions import *
+from .dielectric import *
+from .dimensionless import *
+from .distribution import *
+from .drifts import *
+from .ionization import *
+from .magnetostatics import *
+from .mathematics import *
+from .parameters import *
+from .quantum import *
+from .relativity import *
 
 # auto populate __all__
 for obj_name in list(globals()):
--- a/plasmapy/particles/particle_collections.py
+++ b/plasmapy/particles/particle_collections.py
@@ -6,9 +6,9 @@ import astropy.units as u
 import collections
 import numpy as np
 
-from typing import *
+from typing import *  # noqa: F403
 
-from plasmapy.particles.exceptions import *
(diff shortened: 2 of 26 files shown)


StanczakDominik:
Hey, I'll look through this within a few hours :)

plasmapy/plasma/grids.py
```diff
@@ -1184,7 +1184,7 @@ def volume_averaged_interpolator(
                     valid = (
                         (x >= 0) & (x < n0) & (y >= 0) & (y < n1) & (z >= 0) & (z < n2)
                     )
-                    out = np.where(valid == False)
+                    out = np.where(not valid)
```
StanczakDominik, line 1187: This actually crashes the tests ( https://github.com/PlasmaPy/PlasmaPy/pull/1127/checks?check_run_id=2475959379#step:6:184 ), because `valid` is a numpy array and and the usual pythonic negation doesn't work element-wise by default. This should fix it:
```suggestion
                    out = np.where(~valid)
```

plasmapy/utils/decorators/validators.py
```diff
@@ -330,8 +330,8 @@ def _validate_quantity(self, arg, arg_name: str, arg_validations: Dict[str, Any]
             typeerror_msg += f"{unit}"
 
             if ii != len(arg_validations["units"]) - 1:
-                typeerror_msg += f", "
-        typeerror_msg += f"]"
+                typeerror_msg += ", "
+        typeerror_msg += "]"
```
StanczakDominik, line 334: ... okay, we were definitely overdoing these :D

plasmapy/formulary/__init__.py
```diff
@@ -5,18 +5,18 @@
...
+from .dielectric import *  # noqa: F403
+from .dimensionless import *  # noqa: F403
+from .distribution import *  # noqa: F403
+from .drifts import *  # noqa: F403
+from .ionization import *  # noqa: F403
+from .magnetostatics import *  # noqa: F403
+from .mathematics import *  # noqa: F403
+from .parameters import *  # noqa: F403
+from .quantum import *  # noqa: F403
+from .relativity import *  # noqa: F403
```
StanczakDominik, line 19: Yes, star imports are a terrible anti pattern, yes, this file is a little bit of a special case... I think we could however make this a little cleaner by using https://stackoverflow.com/a/54454433/4417567. So we could replace this entire thing by adding
```
per-file-ignores =
    plasmapy/formulary/__init__.py:F403
```
to the flake8 section of setup.cfg. How does that sound?

plasmapy/particles/particle_collections.py
```diff
@@ -6,9 +6,9 @@
 import collections
 import numpy as np
 
-from typing import *
+from typing import *  # noqa: F403
```
StanczakDominik, line 9:
```suggestion
from typing import Callable, List, Iterable, Union
```

and if I missed any, a test failure will tell us about it 😄

StanczakDominik (commented):
Okay, first up, once again - thanks for tackling this! :)

A few notes:

* This got me to realize we have a collective habit of using f-strings everywhere... The code is, of course simpler without them, but I wouldn't mind disabling the check and letting people use them "as defaults" in the future (keeping the removals in this PR as is, of course). I'm not sure which way is better, though, so - this is not a strong opinion.
* A few cases where `variable == False` was replaced with `not variable`, which would make sense in general but doesn't work with numpy arrays in particular. I've made suggestions wherever that happened to replace those. (EDIT: I think I missed a case with the suggestions, but if you apply the one suggestion I made, failing tests should tell us if there's actually any missing case).
* There are a bunch of errors still being raised in https://github.com/PlasmaPy/PlasmaPy/pull/1127/checks?check_run_id=2475959193#step:6:2581. I notice:
    * The RST210 errors are being found by the `flake8-rst-docstrings` plugin (installed in the tests https://github.com/StanczakDominik/PlasmaPy/blob/83e6337031d88380bda740e1ec2f9e71951d9fc6/.github/workflows/linters.yml#L16 ), which I didn't ask you to install through email, as I didn't think you'd be doing them all in one go! So, you've already gone the extra mile here; but we can revert the removals of those exclusions from `setup.cfg` if you'd prefer to keep this PR shorter. We can always tackle those in the future.
    * A couple of missed f-strings (F541)?
    * and a few W605, which I honestly don't understand why it's being raised and I'm absolutely down to keep excluding them unless we gain that understanding somehow! :)

All in all, great first PR, and I'm happy to merge it with minor revisions! ;)

StanczakDominik (approved):
Looks great, let's get it in! :)

StanczakDominik:
I'm in the phone ATM and it looks like the github app is having some issues right now. I suppose this isn't merging because of merge conflicts? If so I'll clean those up later today.

StanczakDominik:
Oooooh. So because of https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/ we now have to have approve every single workflow run on a first contributor's PR, rather than approving it on a per-PR basis. That's... just silly, honestly.

StanczakDominik:
I also untagged Erik and Nick because that's just CODEOWNERS being overzealous, here.

StanczakDominik:
Haha, whoops. My previous failing auto-merge triggered despite flake8 failing. My bad!

I was already going to do some minor fixes to the CI setup, so I'll fix the two remaining RST errors while I'm at it today.

Thanks, @aksalcido! :)

PR #1128, opened 2021-05-01
Minor CI refactor

StanczakDominik:
Follow-up to #1127.

What this mostly does is allow running linters locally through simply `tox -e linters`. I will follow up with updates to the dev guide soon-ish, where I will document that.


StanczakDominik:
This is minor and touches no user facing functionality, so I'll just merge it :)

PR #1130, opened 2021-05-03
Changes to particles from neoclassics

StanczakDominik:
The relatively non-controversial changes to `plasmapy/particles` lifted from #1059. Does not include #1122.

I just cherry-picked the changes to make this simpler. Roughly speaking, I:

1. added an optional `T_i` field to `IonizationState` that works similarly to `ionic_fractions`.
2. Added the same to `IonicLevel` - it turns out I don't really need it as much, having refactored my code away from using it, but it's probably good to have it be accessible from both levels of abstraction, if it's at all defined.
4. Allowed slicing of IonizationState: it returns a list of IonicLevels. As before, given the recent refactor, I'm no longer as attached to this change as I had been before.
5. made `IonizationState("H", ...) == IonizationState("Li", ...)` return `False` rather than raise an exception.
6. Allowed `len(IonizationState(...))` - it returns the number of charge states, and `len(IonizationStateCollection(...))`, likewise, return the number of base particles
7. Added an xfail test for creating a IonizationStateCollection of two isotopes of the same element. I suspect that's going to come in handy, though I don't yet see a way of doing it.
8. Removed some try/except clauses from the particle tests that were making running `--pudb` impossible.
9. Added a test for `one IonizationState is not another IonizationState`. I think I fixed it somewhere along the way, but I'm not sure.

If any of this is controversial, I can split it off into a separate PR. I'll write the changelogs once we agree on what to include and what to skip.

---

Given positive review:

- [ ] changelog needs writing
- [x] coverage needs improving
- [x] TODO note on T_i needs amending


namurphy:
Thanks!  Reading through the changes you mentioned, these all sound worth including.  I'll try to review this before our community meeting in case I have any questions/suggestions.

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -724,3 +722,14 @@ def test_iteration_with_nested_iterator():
         for fraction2 in hydrogen:
             i += 1
     assert i == 4
+
+
+@pytest.mark.xfail
+def test_electrons_can_be_put_into_ionization_state():
+    electrons = hydrogen = IonizationState("e-", n_elem=1e20 * u.m ** -3, T_e=10 * u.eV)
```
namurphy, line 729: I'm a tad confused about this test.  Is the idea to allow cases where the electron density isn't calculated from the ionization state, but rather given by the user without making the assumption of quasineutrality?

namurphy (commented):
Looking good!  I added a couple of suggestions and questions, and the only other thing I'd suggest would be to add an example in the docstring for `IonizationState` of providing `T_i` as either a single value or as an array.  Thank you for doing this!

plasmapy/particles/tests/test_ionization_state.py
StanczakDominik: Oh... yeah. Remember how I said this would be the non-controversial part? This slipped through.

plasmapy/particles/ionization_state.py
```diff
@@ -577,6 +608,45 @@ def T_e(self, value: u.K):
...
+        if self._T_i is None:
+            return self.T_e
+        return self._T_i.to(u.K, equivalencies=u.temperature_energy())
+
+    @T_i.setter
+    @validate_quantities(
+        value=dict(equivalencies=u.temperature_energy(), none_shall_pass=True)
+    )
+    def T_i(self, value: u.K):
+        """Set the electron temperature."""
```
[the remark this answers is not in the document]
StanczakDominik: Well, of course, thanks for catching this! 😅

plasmapy/particles/ionization_state.py
```diff
@@ -235,23 +252,32 @@ class IonizationState:
...
+        },
+    )
     @particle_input(require="element")
     def __init__(
         self,
         particle: Particle,
         ionic_fractions=None,
         *,
         T_e: u.K = np.nan * u.K,
+        T_i: u.K = None,
```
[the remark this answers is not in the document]
StanczakDominik: I thought so, too; but then, if we want to keep the current behavior the same (you can initialize a IonizationState without a temperature, but then it defaults to `nan`), I'd have to move it somewhere else, so there's no gain in it.

I too dislike the lack of consistency between the two, though.

StanczakDominik:
Thanks for the quick review! I'll wrap this up in time for today's meeting.

StanczakDominik:
Should be good to go!

StanczakDominik:
Well, I did wrap it up in time for *a* Tuesday meeting. [The changes since last time](https://github.com/PlasmaPy/PlasmaPy/pull/1130/files/4dd637e770740b8529fb1138c078574625e42bc1..beea5e00ac22981df6709d1c8972dc0c717bc705) are mostly tests, plus I did add `T_i` to the IonizationState summary if it's defined.

plasmapy/particles/ionization_state.py
```diff
@@ -268,10 +290,11 @@ def __init__(
         Initialize an `~plasmapy.particles.ionization_state.IonizationState`
         instance.
         """
+        self._number_particles = particle.atomic_number + 1
```
namurphy, line 293: [optional] Perhaps `self._number_of_particles`?

plasmapy/particles/ionization_state.py
```diff
@@ -767,6 +839,10 @@ def _get_states_info(self, minimum_ionic_fraction=0.01) -> List[str]:
                     value = "{:.2e}".format(state.number_density.si.value)
                     state_info += f"    n_i = {value} m**-3"
 
+                if np.isfinite(state.T_i):
+                    value = "{:.2e}".format(state.T_i.si.value)
```
namurphy, line 843: I keep forgetting that there are other ways to format strings besides f-strings. 😬

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -745,3 +791,9 @@ def test_iteration_with_nested_iterator():
         for fraction2 in hydrogen:
             i += 1
     assert i == 4
+
+
+def test_ionization_state_is_another():
+    hydrogen = IonizationState("D+", n_elem=1e20 * u.m ** -3, T_e=10 * u.eV)
+    deuterium = IonizationState("T+", n_elem=1e20 * u.m ** -3, T_e=10 * u.eV)
+    assert hydrogen is not deuterium
```
namurphy, line 799:
```suggestion
def test_ionization_state_inequality_and_identity():
    deuterium_states = IonizationState("D+", n_elem=1e20 * u.m ** -3, T_e=10 * u.eV)
    tritium_states = IonizationState("T+", n_elem=1e20 * u.m ** -3, T_e=10 * u.eV)
    assert deuterium_states is not tritium_states
    assert deuterium_states != tritium_states
```
Should at least make the variable names consistent with the isotopes.

namurphy (approved):
Looks good!  Some minor suggestions, with the main suggestion being to make the variable names consistent with the isotopes in the final test.  Thanks!

plasmapy/particles/ionization_state.py
StanczakDominik: ... it shows that I copy pasted the block above, doesn't it? :D

plasmapy/particles/tests/test_ionization_state.py
StanczakDominik: Gah. Yeah. Duh. Of course.

StanczakDominik:
All right, let's hit it. Thanks, Nick! :)

PR #1134, opened 2021-05-08
Easier Output Saving for Proton Radiography

pheuer:
Large runs with `SyntheticProtonRadiograph` can take a while to run, so it's often necessary to save the output for later analysis. By design, the way to do this is to save the particle positions in the detector plane, so different histograms can be made from the same dataset. So far, I personally have been doing this by just pickling the entire `SyntheticProtonRadiograph` object. However, I've recently realized that pickling complex objects isn't great practice for long-term storage. 

This PR includes the following changes

- The module `proton_radiography` is now renamed `charged_particle_radiography`

- The class `SyntheticProtonRadiograph` is now renamed `Tracker`, as in 
```python
from plasmapy.diagnostics import charged_particle_radiography as cpr
sim = cpr.Tracker()
```

- `Tracker` now stores the particle positions and some metadata in a dict structure after running.

- the `save_result()` method of `Tracker` can save this dictonary to a `numpy` `npz` file.

- the `synthetic_radiograph`  function is no longer a method of `Tracker`  but instead a stand-alone function in `charged_particle_radiography.py`. This function takes either a `Tracker`  object or an output dictionary (eg. like one loaded from a `npz` file), then creates the synthetic radiograph in the same way as before. 

TODO
- [ ] Update documentation notebook to show how to save the npz file? This is tricky because we can't actually save it in a ipynb?
- [ ] Add an option (maybe new default option) to save in the open metadata PRADFORMAT. *This will be delayed for a future PR*


StanczakDominik:
> I've written tests for most of this, but is there a way I can test writing and reading a file in the CI tests? Is there a way I can demonstrate this capability in the example notebooks? Anything wrong with just saving to os.getcwd()?

You could do:

```python
result.save(filename)
loaded_result = load(filename)
assert result == loaded_result
```

Use https://docs.pytest.org/en/stable/tmpdir.html for the filename.

We could also add a sample dataset as a regression test. It looks like our tests' data files currently live in `plasmapy/particles/data/test`, which is pretty meh (e.g. `plasmapy/plasma/sources` tests depend on it) but we can refactor that later.

namurphy:
Thank you for doing this!  These are very useful capabilities to add.  Do you think it would also be worth having the option to save and load data in HDF5 format?  I'm wondering about that since HDF5 is probably the best format to use for long-term archiving of data sets on sites like Zenodo.

pheuer:
Sure! `h5py` and HDF5 are my go-to formats, so I'd could easily add HDF5 as an option. That said, directly reading out a dict that can be directly passed into `synthetic_radiograph` is quick and convenient, which is why I originally went with `npz` here.

pheuer:
@namurphy There is a nascent effort to[ introduce a community input/output standard for proton radiograph](https://github.com/phyzicist/pradformat). The formats are all based on HDF5. One of the proposed formats for [ParticleLists](https://scott.cikeys.com/prad/formats/particles_list/) could be used for this kind of output. Is there a way we could include this as an optional dependency to allow the PlasmaPy proton radiography particle tracer to read and write these standards?

StanczakDominik:
https://github.com/picmi-standard/picmi may also be relevant-ish

pheuer:
After looking into openPMD, I want to try and rework the pradformat spec to make it consistent. I'll leave HDF saving out of this PR, but I intend to add that once we've made changes to the pradformat spec.

StanczakDominik:
Getting to the review in a few minutes!

pre-commit.ci autofix

StanczakDominik (commented):
Okay, I've briefly looked at this - nothing too deep, just a quick look-through. I'd like to ask you about a slightly different approach.

My main issue here is naming: having a `SyntheticProtonRadiograph` class and a `synthetic_radiograph` function in the same module sounds confusing. Which do I use first?

We could rename the latter; but then I looked at the signature of the function - it takes either an object or a dict representing the object.

But how about something like this (from #1059; nevermind the details, ofc):

https://github.com/StanczakDominik/PlasmaPy/blob/037995509568bb1d802abe03eeb9388a2690f087/plasmapy/transport/flows.py#L37-L89

We could have an alternate constructor that takes a dict or path to npz file (could even have two, one calling the other); then have `proton_radiograph` still be a method of the `SyntheticProtonRadiograph` class.

So, instead of the workflow you're proposing here:

```python
sim = SyntheticProtonRadiograph(...)
# run the simulation

# or

sim = np.load(...)

# then

proton_radiograph(sim)
```

you could do

```python
sim = SyntheticProtonRadiograph(...).run(...)

# or

sim = SyntheticProtonRadiograph.from_dict(...)

# or

sim = SyntheticProtonRadiograph.from_npz(path, ...)

# then

sim.proton_radiograph()
```

which would probably simplify the method/function call by a bunch., since then you don't need to effectively dispatch on the type of argument you're getting.

What this would require is creating the SPR object within the alternate constructor, then setting its fields manually. It looks like the only problematic thing would be this part:

https://github.com/pheuer/PlasmaPy/blob/451d12ede71d71ef88a8b325e04581d6c93bc417/plasmapy/diagnostics/proton_radiography.py#L1100-L1107

... or maybe you don't even need to set the fields; you could just set the dict. Perhaps. Probably better to set the fields while we're at it.

But at the same time, this might be complicated since if I've got this right, you don't seem to be saving `self.x`... wouldn't you, though, want to save it?

What do you think?

pheuer:
> My main issue here is naming: having a `SyntheticProtonRadiograph` class and a `synthetic_radiograph` function in the same module sounds confusing. Which do I use first?
>
> We could rename the latter; but then I looked at the signature of the function - it takes either an object or a dict representing the object.

Yes, I agree that this needs to be changed, but I think it's actually `SyntheticProtonRadiograph` that now has a misleading name. `synthetic_radiograph` is descriptive (it generates a synthetic radiograph from particle positions),  but `SyntheticProtonRadiograph` is (after this PR) actually just a particle tracer, with the one added step of advancing particles to the detector plane at the end.  My intention would be that, after implementing support for pradformat, `synthetic_radiograph` could be used _independent of the plasmapy particle tracer_ to make synthetic radiographs from lists of particle positions provided by some other program. For that reason, I don't want to have to instantiate the particle tracer object if I already have a list of particle positions in the detector plane? 

The other issue is that the input to `synthetic_radiograph` needs a lot less information than the particle tracer: just positions in a 2D plane as opposed to 3D positions and velocities, and the field grids. Storing all that information would be wasteful, but otherwise instantiating `SyntheticProtonRadiograph` with that data would not produce a fully functional object (eg. you couldn't call `run()`). 

I would suggest the cleanest way to do this would be to just rename `SyntheticProtonRadiograph` to something more descriptive - maybe `SyntheticRadiographParticleTracer` or something?

StanczakDominik:
That makes sense, yeah. Sorry! I blame the heat wave we're having 😅 

`SyntheticProtonRadiograph` -> `SyntheticRadiographParticleTracer` sounds reasonable to me, but let's call upon @namurphy and @rocco8773 for ideas :)

pheuer:
From meeting today, this naming scheme is suggested

```python
import plasmapy.diagnostics.charged_particle_radiography as cpr
# then
sim = cpr.ParticleTracker(...)
```

pheuer:
@rocco8773  CI is raising this error: 

`
Warning, treated as error:
/home/runner/work/PlasmaPy/PlasmaPy/docs/api/plasmapy.analysis.fit_functions.Exponential.rst.rst:18:autosummary: failed to import Exponential.FitParamTuple
`

rocco8773:
> @rocco8773 CI is raising this error:
> 
> `Warning, treated as error: /home/runner/work/PlasmaPy/PlasmaPy/docs/api/plasmapy.analysis.fit_functions.Exponential.rst.rst:18:autosummary: failed to import Exponential.FitParamTuple ERROR: InvocationError for command /home/runner/work/PlasmaPy/PlasmaPy/.tox/build_docs/bin/sphinx-build docs docs/_build/html -W -b html (exited with code 2)`

Yes, this is the failed check I mentioned during the call today.  I'm puzzled by it because the import should not even be happening and I don't get the warning on my local builds.  Ignore it for now because I have to do more digging on it.

docs/notebooks/diagnostics/charged_particle_radiography_particle_tracing.ipynb
```diff
@@ -0,0 +1,552 @@
...
+   "metadata": {},
+   "source": [
+    "# Creating Synthetic Charged Particle Radiographs by Particle Tracing"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[charged_particle_radiography]: ../../api/plasmapy.diagnostics.charged_particle_radiography.rst#plasmapy.diagnostics.charged_particle_radiography\n",
```
rocco8773, line 14:
```suggestion
    "[charged_particle_radiography]: ../../ad/diagnostics/charged_particle_radiography.rst\n",
```

docs/notebooks/diagnostics/charged_particle_radiography_particle_tracing.ipynb
```diff
@@ -0,0 +1,552 @@
...
...pPW0BqfPqnY5dUIM4Q0cbHx+nu7qajo+OiKD9tNt9bV1e3pc1Bptw+ODjI0aNH98xEZa/hcDg4ePAgPT09lJWVXRCi0vmQ2RjNz8+ztLSUnRleO5qXGW9aS+LKNTLM//7+/l1Z0W6EvWJN71f82gdik8mE3+/n/vvvz0q+JZNJXvnKV2KxWHjLW96y62Dc0dFBT08PwI79Z1VN58nRBebCqTT72SDyy7ElfE4TrzqUHqdYjsl8s3sGt8WI1SQxH07y9z8cxbFqcTi+FGdyOc5911Rgkk6/FE+NLfK156YRhfR57CYJXUgzn4G0xzA611QVMjI2zjJOwjIsxmReebCEtpI0acJnNzE0EyFDocizGRAEWIjKqJqGuBpWZS1NDpNE8NhNlOZZCCcURufDfPXZKRajKVRtfbklEzx1QeAFdb4NR5fOd/+eOLqQJaWpenoLnJS3byO3H1BRUYHRaLygfcqN+r27me8VRZGWlhaOHTtGX18fra2t+yLD3y7MZjOHDx+mr6+PeDxOdXX1Bd9UqKrKwsIC8/Pz2Y3RZhwWIDtfnEqlkGV5x97G54PJZKKzszN7b3LVK/5NK03/2gdiSZJ4wxvesO7vzGYz3/jGN7j77rsxGo288Y1vzGlmvJM5RFXXiaZUwgmZfHt60XWaJUbno8TlNJEpGE2haWA1rYp0WI38/NgCt7cUIIkiTouB6eUEgXCS0rx0vy6clLn/2Sm8NhM2swS6ztBcBFnVVw0kBCaW4kyvJNEACTfXeiJc3VZHkd+7TvDjVYdL6JsOE4ykCWJleTY+c28tjw8HePL4IsOzEfxGEYMkMLmUQNV1nGaJlWgCXVP55rNj6JpIkdPEbEReZYunoZMOnnU+67aDMMATowusJNZ75urAXc1b05zej8iwjDMzurmWxNxKv3e3yJCfJiYm6Orqor29fV8zpzeDJEl0dHRw9OjRC+aklUqlssIniURiR+5fZ5pG7MU1i6JIWVlZ1tikpaVl1xvH37Kmf0NgsVj41re+xUtf+lKMRiOvfe1rd3W8zIuaCcbbLWGZJJFKr43+6TBOi0ZC1rAYpfSY0OrP2E0SGqcDaDylYJBEMu0PXU9zkzOBbHAmzJefnqBrYoWUrOK2mSj3WHFZDFTkWzkejKHrcHIhTmORA5/DTCyl8vNFByUnx7AbBSxrZPL8jnSPd2A6jChAS4
```
rocco8773, line 484:
```suggestion
    "[create_particles()]: ../../api/plasmapy.diagnostics.charged_particle_radiography.Tracker.rst#plasmapy.diagnostics.charged_particle_radiography.Tracker.create_particles\n",
```

rocco8773 (changes requested):
A few links that still need updating...

1. `charged_particle_radiography_particle_tracing_source.ipynb`
    * The `[create_particles()]` link still points to `SyntheticProtonRadiography`
    
    ```markdown
    # old
    [create_particles()]: ../../api/plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.rst#plasmapy.diagnostics.proton_radiography.SyntheticProtonRadiograph.create_particles

    # needs to be
    [create_particles()]: ../../api/plasmapy.diagnostics.charged_particle_radiography.Tracker.rst#plasmapy.diagnostics.charged_particle_radiography.Tracker.create_particles
    ```
2. Notebook like in `docs/index.rst` needs to be updated

    ```rest
    .. old
    .. nbgallery::
       :hidden:

       notebooks/diagnostics/proton_radiography_particle_tracing

    .. needs to be
    .. nbgallery::
       :hidden:

       notebooks/diagnostics/charged_particle_radiography_particle_tracing
    ```

By applying these changes in my local copy I've been able to clear all the documentation build warnings.

docs/notebooks/diagnostics/charged_particle_radiography_particle_tracing.ipynb
rocco8773: Not sure why GitHub is displaying the wrong line here, but if you go to the "Files" tab you can see this suggestion in the proper context.

rocco8773:
Is it possible to reduce the number of particles `nparticles` used in the notebooks in attempt to speed up the notebook execution?  The execution times is becoming a hindrance in developing/prototyping.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -162,6 +164,16 @@ def __init__(
...
+        # Calculate the size of the grid imaged onto the detector
+        # (used for auto-choosing the size keyword when making histograms)
+        self.grid_scale_length = np.max(
+            [
+                np.max(np.abs(self.grid.pts0.to(u.m).value)),
+                np.max(np.abs(self.grid.pts1.to(u.m).value)),
+                np.max(np.abs(self.grid.pts2.to(u.m).value)),
+            ]
+        )
+
```
rocco8773, line 176: Make this into a property so has proper documentation and is only calculated when needed.  Right now I do not see why it needs to be part of the class `__init__`.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,6 +1104,55 @@ def run(
             f"{self.fract_deflected*100}%"
         )
 
+        return self.output
+
+    def _store_result(self):
```
rocco8773, line 1109: Even if a function/method is private, it still needs a docstring.  In this case the docstring is written for the developer's knowledge and not the end user. ... Always keep in the back of you mind, "Will this be understood by me or anyone else a year from now?"  If no, then the documentation/comments need to be expanded.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -347,6 +352,9 @@ def add_wire_mesh(
 
         """
 
+        # Simulation has not run, because adding mesh changes the simulation
+        self._has_run = False
```
rocco8773, line 356: Are there any other "things" that need to be "reset" when a user modifies `Tracker` (e.g. voiding/deleting results).  This might be a good thing to do in a `_reset_results` method.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,6 +1103,125 @@ def run(
             f"{self.fract_deflected*100}%"
         )
 
+        # Simulation has not run, because creating new particles changes the simulation
+        self._has_run = True
+
+        return self.results_dict
```
rocco8773, line 1109: If I understand correctly, `results_dict` is really only used if someone wants to save the results, otherwise, the user accesses results via the `Tracker` attributes.  If I'm correct in this interpretation, then I wouldn't return `results_dict` here.  If I'm wrong, then I'd rename `results_dict` -> `results` and advertise it as the main way to interface with simulation results.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Not really -  in 99% of use cases I expect that the simulation will only be run once (since the overhead to instantiate a new simulation is so much lower than actually running the simulation). The only ways to access the results (other than going in and pulling out the attributes manually) are as output from `run()` or through `results_dict()` which now checks to make sure the results are current.

rocco8773: I agree that a very few percentage of situation will re-run a simulation, but if it's possible to re-run then we need to properly handle it.  Otherwise, it could lead to unintended consequences.  For now, I would be okay with not allowing any modification to the `Tracker` object after the sim is run.

rocco8773 (changes requested):
The parameters section in `Tracker.create_particles` is missing its section heading.

That is, ...

```rst
Parameters
----------
```

https://github.com/PlasmaPy/PlasmaPy/blob/959133e722dc81cc56eecd9384ee232bdc7312c8/plasmapy/diagnostics/charged_particle_radiography.py#L561-L569

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -19,6 +20,7 @@
 
 from plasmapy import particles
 from plasmapy.formulary.mathematics import rot_a_to_b
+from plasmapy.formulary.relativity import relativistic_energy
```
rocco8773, line 23:
```suggestion
```

Since `relativistic_energy` is no longer used in this module.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Ok, done. 

I think this will be the permeant solution: I tried making it so you could re-run the simulation with changes but this necessitates storing copies of a lot of the arrays in a way that just isn't memory efficient for large calculations.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,6 +1103,125 @@ def run(
...
+            The y-coordinate location where each particle would have hit the
+            detector plane if the grid fields were zero, in meters. Useful
+            for calculating the source profile.
+
+        'v0' : `~numpy.ndarray`, [nparticles, 3]
+            The velocity of each particle when it hit the detector plan if the
+            grid fields were zero, in meters per second. The velocity is in a
+            coordinate system relative to the detector plane. The components are
+            [normal, horizontal, vertical] relative to the detector plane
+            coordinates.
```
rocco8773, line 1161: This was not very readable/clear in RTD build...

image


So, I think it would look better as a table...

image


```suggestion
        .. list-table:: Dictionary keys and descriptions.
           :width: 100%
           :widths: 1 1 4
           :header-rows: 1

           * - Key
             - Type
             - Description
           * - ``"source"``
             - `~numpy.ndarray`
             - The source location vector, in meters.
           * - ``"detector"``
             - `~numpy.ndarray`
             - The detector location vector, in meters.
           * - ``"mag"``
             - `float`
             - The system magnification.
           * - ``"nparticles"``
             - `int`
             - Number of particles in the simulation.
           * - ``"max_deflection"``
             - `~astropy.units.Quantity`
             - The maximum deflection experienced by a particle in the
               simulation, in radians.
           * - ``"xloc"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The x-coordinate location where each particle hit the
               detector plane, in meters.
           * - ``"yloc"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The y-coordinate location where each particle hit the
               detector plane, in meters.
           * - ``"v"``
             - `~numpy.ndarray`, [``nparticles``, 3]
             - The velocity of each particle when it hits the detector
               plane, in meters per second. The velocity is in a
               coordinate system relative to the detector plane. The
               components are [normal, horizontal, vertical] relative
               to the detector plane coordinates.
           * - ``"xloc0"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The x-coordinate location where each particle would have
               hit the detector plane if the grid fields were zero, in
               meters. Useful for calculating the source profile.
           * - ``"yloc0"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The y-coordinate location where each particle would have
               hit the detector plane if the grid fields were zero, in
               meters. Useful for calculating the source profile.
           * - ``"v0"``
             - `~numpy.ndarray`, [``nparticles``, 3]
             - The velocity of each particle when it hit the detector
               plan if the grid fields were zero, in meters per second.
               The velocity is in a coordinate system relative to the
               detector plane. The components are [normal, horizontal,
               vertical] relative to the detector plane coordinates.
```

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Currently the workflow is set up so the user can also directly pass a `Tracker` object into `synthetic_radiograph`. But, I agree that this isn't necessary here in either case, so I'll remove it.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Perfect, thanks, I tried to format this a few ways but kept getting errors.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: :cry: Don't worry #979, we won't forget you!

rocco8773 (changes requested):
To be consistent, choose singular `result` or plural `results` and us that for both the `results_dict` and `save_result`.  I currently don't have a preference, so choose the one you like better.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,11 +1107,137 @@ def run(
...
+            y0=y0loc,
+            v0=v0,
+        )
+
+        return result_dict
+
+    def save_results(self, path):
+        """
+        Save the output dictionary containing the final particle positions in
+        the detector plane into a `numpy.npz` file for later analysis.
```
rocco8773, line 1224:
```suggestion
        Save the simulations results :attr:`results_dict` to a `numpy`
        ``.npz`` file format (see `numpy.lib.format`) using `numpy.savez`.
```

So you don't have to restate things, this will create link-backs for expanded detials.  `numpy.lib.format` actually discusses the different `numpy` file formats.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,11 +1107,137 @@ def run(
...
+
+        Parameters
+        ----------
+
+        path : str or `os.path`
+            The path to save the output file.
```
rocco8773, line 1230:
```suggestion
        path : `str` or `os.path`
            Either the filename (string) or an open file (file-like object)
            where the data will be saved. If file is a string or a Path,
            the ``.npz`` extension will be appended to the filename if
            it is not already there.
```

Since you're leveraging `np.savez` just use there definition for the argument to remain consistent.

rocco8773 (changes requested):

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1260,144 @@ def max_deflection(self):
...
-        image plane).
 
-        Parameters
-        ----------
-        size : `~astropy.units.Quantity`, shape (2,2)
-            The size of the detector array, specified as the minimum
-            and maximum values included in both the horizontal and vertical
-            directions in the detector plane coordinates. Shape is
-            [[hmin,hmax], [vmin, vmax]]. Units must be convertable to meters.
+        The Tracker methods could give strange results if setup methods
```
rocco8773, line 1269:
```suggestion
        The `Tracker` methods could give strange results if setup methods
```

There is also a blank line above this line that needs deleting.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1260,144 @@ def max_deflection(self):
...
+        """
 
-        ignore_grid: bool
-            If True, returns the intensity in the image plane in the absence
-            of simulated fields.
+        if self._has_run:
+            raise RuntimeError(
+                "Adding new particles or meshes after a simulation "
+                "is complete is not supported. Create a new simulation "
+                "object to begin a new run."
```
rocco8773, line 1278:
```suggestion
                "Modifying the `Tracker` object after running the "
                "simulation is not supported. Create a new `Tracker` "
                "object for a new simulation."
```

Generalize the error statement a little more so it's more encompassing of possible future modifications and, if so, would not need rewording.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,143 @@ def max_deflection(self):
...
+# Synthetic diagnostic methods (creating output)
+# *************************************************************************
 
-            .. math::
-                OD = -log_{10}(Intensity/I_0)
 
-            where I_O is the intensity on the detector plane in the absence of
-            simulated fields. Default is False.
+def synthetic_radiograph(
+    obj, size=None, bins=[200, 200], ignore_grid=False, optical_density=False
```
rocco8773, line 1288:
```suggestion
    obj, size=None, bins=None, ignore_grid=False, optical_density=False
```

You don't ever want your default values to be mutable objects because it can lead to some strange behavior.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,143 @@ def max_deflection(self):
...
+            "The first argument of synthetic_radiograph must be "
+            "either a cpr.Tracker or an "
+            "output dictionary from cpr.Tracker"
         )
 
-        # h, v are the bin edges: compute the centers to produce arrays
-        # of the right length (then trim off the extra point)
-        h = ((h + np.roll(h, -1)) / 2)[0:-1]
-        v = ((v + np.roll(v, -1)) / 2)[0:-1]
+    # Note that, at the end of the simulation, all particles were moved
```
rocco8773, line 1350:
```suggestion
    if bins is None:
        bins = [200, 200]

    # Note that, at the end of the simulation, all particles were moved
```

rocco8773 (changes requested):

rocco8773 (changes requested):
This is adding to the `synthetic_radiograph` docstring proper syntax highlighting and cross-referencing.

I'd recommend when you're writing docstrings to start paying attention to the following things...

1. Being consistent with wording, symbols, etc.  For example, `numpy` shapes are tuples so switching between square brackets `[]` and parenthesis `()` can be confusing for new users.
2. Make sure to to add single back ticks to elements that should be cross-linked.  For example, do `` `~numpy.ndarray` `` and not `ndarray`.
3. When writing inline code use double back ticks.  For example, ` ``(200, 200)`` ` and not `(200, 200)`.
4. Make sure your docstrings dot the i's and cross the t's.  For example. the `optical_density` used the variable `Intensity` in the equation which was never defined anywhere in the nearby context.  Things like make it harder for new users to pickup and maintainers to maintain.

These are a lot of annoying little things that we never perfect.  But the goal is keep them in mind so we're good about it 99% of the time.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,149 @@ def max_deflection(self):
...
-            # detector, since particles could be deflected out
-            size = 5 * np.array([[-w, w], [-w, w]]) * u.m
+    # results_dict will raise an error if the simulation has not been run.
+    if isinstance(obj, Tracker):
+        d = obj.results_dict
 
-        # Generate the histogram
-        intensity, h, v = np.histogram2d(
-            xloc, yloc, range=size.to(u.m).value, bins=bins
+    # Check if dictioanry-like (dict or npz file)
```
rocco8773, line 1343:
```suggestion
    # Check if dictionary-like (dict or npz file)
```

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,149 @@ def max_deflection(self):
...
+
+    if size is None:
+        # If a detector size is not given, choose a size based on the
+        # particle positions
+        w = np.max([np.max(np.abs(xloc)), np.max(np.abs(yloc))])
+
+        # The factor of 5 here is somewhat arbitrary: we just want a
+        # region a few times bigger than the image of the grid on the
+        # detector, since particles could be deflected out
+        size = 5 * np.array([[-w, w], [-w, w]]) * u.m
```
rocco8773, line 1376: Maybe add something like a `pad` keyword to `synthetic_radiograph` so the user can adjust this.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,149 @@ def max_deflection(self):
...
+    # If ignore_grid is True, use the predicted positions in the absence of
+    # simulated fields
+    if ignore_grid:
+        xloc = d["x0"]
+        yloc = d["y0"]
+    else:
+        xloc = d["x"]
+        yloc = d["y"]
+
+    if size is None:
```
rocco8773, line 1368: Should add some `size` validations here to handle types and shape.  Something like...

```python
if size is None:
    ...
elif not isinstance(size, np.ndarray):
    raise TypeError
elif size.shape != (2, 2):
    raise ValueError
elif not isinstance(size, u.Quantity):
    warn("Assuming SI")
    size = size * u.m
```

rocco8773: Note that `Quantity` is a sub-class of `ndarray` so `isinstance(5 * u.m, np.ndarray)` will be `True`.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,149 @@ def max_deflection(self):
...
-            # histogram. Zeros are just outside of the illuminate area.
-            I0 = np.mean(I0[I0 != 0])
+        # Calculate I0 as the mean of the non-zero values in the null
+        # histogram. Zeros are just outside of the illuminate area.
+        I0 = np.mean(I0[I0 != 0])
 
-            # Overwrite any zeros in intensity to avoid log10(0)
-            intensity[intensity == 0] = 1
+        # Overwrite any zeros in intensity to avoid log10(0)
+        intensity[intensity == 0] = 1
```
rocco8773, line 1406: Why did you use this approach here?  I feel like this is adding artificial values to the histogram data.  Replacing these points with `np.nan` or just allowing `np.log10` to fill with `-np.inf` seems like the more appropriate route.  To avoid the `RuntimeWarning` in the latter case you can just temporarily activate a warnings filter.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,149 @@ def max_deflection(self):
...
+        # detector, since particles could be deflected out
+        size = 5 * np.array([[-w, w], [-w, w]]) * u.m
+
+    # Generate the histogram
+    intensity, h, v = np.histogram2d(xloc, yloc, range=size.to(u.m).value, bins=bins)
+
+    # h, v are the bin edges: compute the centers to produce arrays
+    # of the right length (then trim off the extra point)
+    h = ((h + np.roll(h, -1)) / 2)[0:-1]
+    v = ((v + np.roll(v, -1)) / 2)[0:-1]
```
rocco8773, line 1384:
```suggestion
    h = ((h + np.roll(h, -1)) / 2)[:-1]
    v = ((v + np.roll(v, -1)) / 2)[:-1]
```

Both of these approaches produce the same results with the same execution time.  My mind always trips out when I see a `0` index before the `-1` index, so I have an easier time reading the code I'm suggesting.  However, this is very subjective and doesn't matter in the end so take my suggestion or leave yours in place.  I'll leave the final decision up to you.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Actually, since we changed the way `w` is calculated to just include all particle positions on the detector, we should just change 5 -> 1 here. There should be no need to pad beyond this, since all particles will be captured by definition.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Done, except I just kept the requirement that `size` be an explicit `u.Quantity`. Since detectors vary in size (but are always much smaller than a meter), allowing raw numbers interpreted as SI units is an unlikely use case.

plasmapy/diagnostics/charged_particle_radiography.py
pheuer: Good point: I've taken this out and added a note in the docstring about how to deal with `np.inf` values if necessary. 

This is how I filtered the warnings:

```python
with np.errstate(divide="ignore"):
            intensity = -np.log10(intensity / I0)
```

changelog/1134.breaking.3.rst
```diff
@@ -1,2 +1,2 @@
-`~plasmapy.diagnostics.charged_particle_radiograph.Tracker()` no longer supports making changes to a `Tracker` object and re-running. 
+`~plasmapy.diagnostics.charged_particle_radiograph.Tracker()` no longer supports making changes to a `Tracker` object and re-running.
```
namurphy, line 1:
```suggestion
`~plasmapy.diagnostics.charged_particle_radiograph.Tracker` no longer supports making changes to a `~plasmapy.diagnostics.charged_particle_radiograph.Tracker` object and re-running.
```

changelog/1134.bug.rst
```diff
@@ -1,2 +1,2 @@
-`~plasmapy.diagnostics.charged_particle_radiograph.synthetic_radiograph` run with the `optical_density` keyword will now return `np.nan` where the 
+`~plasmapy.diagnostics.charged_particle_radiograph.synthetic_radiograph` run with the `optical_density` keyword will now return `np.nan` where the
 source profile `I0` is zero. Previously, an incorrect value was returned.
```
namurphy:
```suggestion
`~plasmapy.diagnostics.charged_particle_radiograph.synthetic_radiograph` run with the ``optical_density`` keyword will now return `numpy.nan` where the
source profile ``I0`` is zero. Previously, an incorrect value was returned.
```

namurphy (changes requested):
A few quick changes in the changelog for reST formatting of links.  Using back ticks makes Sphinx try to link to the relevant code thing, while double back ticks makes it get formatted as literal text that doesn't get linked to.  Thank you again for doing this!

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,11 +1107,140 @@ def run(
...
+
+        Parameters
+        ----------
+
+        path : `str` or `os.path`
+            Either the filename (string) or an open file (file-like object)
+            where the data will be saved. If file is a string or a Path,
+            the ``.npz`` extension will be appended to the filename if
+            it is not already there.
+
```
rocco8773, line 1234:
```suggestion

    Notes
    -----

    Useful for saving the results from a simulation so they can be
    loaded at a later time and passed into
    `~plasmapy.diagnostics.charged_particle_radiography.synthetic_radiograph`.

```

I was thinking that it would be useful to add a little context here so it's more apparent to the user that this can be used in conjunction with `synthetic_radiograph`.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,161 @@ def max_deflection(self):
...
-            The horizontal axis of the synthetic radiograph in meters.
+    Parameters
+    ----------
 
-        vax : `~astropy.units.Quantity` array shape (vbins, )
-            The vertical axis of the synthetic radiograph in meters.
+    obj: `dict` or `~plasmapy.diagnostics.charged_particle_radiography.Tracker`
+        Either a `~plasmapy.diagnostics.charged_particle_radiography.Tracker`
+        object that has been run, or an output dictionary created by
+        running `~plasmapy.diagnostics.charged_particle_radiography.Tracker`.
```
rocco8773, line 1300:
```suggestion
    obj: `dict` or `~plasmapy.diagnostics.charged_particle_radiography.Tracker`
        Either a `~plasmapy.diagnostics.charged_particle_radiography.Tracker`
        object that has been run, or a dictionary equivalent to
        `~plasmapy.diagnostics.charged_particle_radiography.Tracker.results_dict`.
```

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,161 @@ def max_deflection(self):
...
+        plane in the absence of simulated fields. Default is `False`.
 
-        # Generate the histogram
-        intensity, h, v = np.histogram2d(
-            xloc, yloc, range=size.to(u.m).value, bins=bins
+        If the `Intensity` histogram contains zeros, `OD` will contain
+        -`~numpy.inf` values. These can be easily replaced with zeros
+        if desired for plotting using `~numpy.nan_to_num`
+
+        ``OD = np.nan_to_num(OD, neginf=0)``
```
rocco8773, line 1332:
```suggestion
        where :math:`Intensity` is the simulation intensity on the
        detector plane and :math:`I_0` is the intensity on the detector
        plane in the absence of simulated fields. Default is `False`.
        If the :math:`Intensity` histogram contains zeros, then the
        corresponding values in :math:`OD` will be `numpy.inf`.  When
        plotting :math:`OD` the `~numpy.inf` values can be replaced
        using ``numpy.nan_to_num(OD, neginf=0, posinf=0)``.
```

The rendering of parameter definitions is very limited.  You basically need to keep everything in one paragraph for it to render correctly.

I also replaced `-numpy.inf` with `numpy.inf` since the OD expression is `-numpy.log10(I/I0)` which means `0` maps to `numpy.inf`.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,161 @@ def max_deflection(self):
...
+        d = obj.results_dict
+
+    # Check if dictionary-like (dict or npz file)
+    elif isinstance(obj, dict):
+        d = obj
+    else:
+        raise ValueError(
+            "The first argument of synthetic_radiograph must be "
+            "either a cpr.Tracker or an "
+            "output dictionary from cpr.Tracker"
```
rocco8773, line 1357:
```suggestion
    # condition `obj` input
    if isinstance(obj, Tracker):
        # results_dict raises an error if the simulation has not been run.
        d = obj.results_dict
    elif isinstance(obj, dict):
        d = obj
    else:
        raise TypeError(
            f"Expected type dict or {Tracker} for argument `obj`, but "
            f"got type {type(obj)}."
```

This is really a `TypeError` because once you get to the else part of the clause you know `obj` is not the correct type.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1115,118 +1263,161 @@ def max_deflection(self):
...
+    elif not isinstance(size, u.Quantity) or not size.unit.is_equivalent(u.m):
+        raise TypeError(
+            "``size`` must be an `~astropy.units.Quantity` "
+            "object with units convertable to meters."
+        )
+    elif size.shape != (2, 2):
+        raise ValueError(
+            "``size`` must have shape ``(2,2)`` corresponding to "
+            "``[[xmin, xmax], [ymin, ymax]]``."
+        )
```
rocco8773, line 1389:
```suggestion
        elif not isinstance(size, u.Quantity):
        raise TypeError(
            "Argument `size` must be an astropy.units.Quantity object with "
            "units convertable to meters."
        )
    elif not size.unit.is_equivalent(u.m):
        raise ValueError(
            "Argument `size` must have units convertible to meters."
        )
    elif size.shape != (2, 2):
        raise ValueError(
            f"Argument `size` must have shape (2, 2), but got {size.shape}."
        )
```

I broke this up because the not a `Quantity` is a `TypeError` but the rest are `ValueError`'s.

Note that all the back-tick and tilde syntax used in docstrings don't work for warnings or issues.  The syntax is specific to `sphinx` stuff.

rocco8773 (changes requested):
Overall this looks good.  Just a few minor changes needed for the functionality in `charged_particle_radiography.py`.

I'll now go over the test in a separate review.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,11 +1107,140 @@ def run(
...
+             - The y-coordinate location where each particle would have
+               hit the detector plane if the grid fields were zero, in
+               meters. Useful for calculating the source profile.
+           * - ``"v0"``
+             - `~numpy.ndarray`, [``nparticles``, 3]
+             - The velocity of each particle when it hit the detector
+               plan if the grid fields were zero, in meters per second.
+               The velocity is in a coordinate system relative to the
+               detector plane. The components are [normal, horizontal,
+               vertical] relative to the detector plane coordinates.
```
rocco8773, line 1173:
```suggestion
           * - ``"x"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The x-coordinate location where each particle hit the
               detector plane, in meters.
           * - ``"y"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The y-coordinate location where each particle hit the
               detector plane, in meters.
           * - ``"v"``
             - `~numpy.ndarray`, [``nparticles``, 3]
             - The velocity of each particle when it hits the detector
               plane, in meters per second. The velocity is in a
               coordinate system relative to the detector plane. The
               components are [normal, horizontal, vertical] relative
               to the detector plane coordinates.
           * - ``"x0"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The x-coordinate location where each particle would have
               hit the detector plane if the grid fields were zero, in
               meters. Useful for calculating the source profile.
           * - ``"y0"``
             - `~numpy.ndarray`, [``nparticles``,]
             - The y-coordinate location where each particle would have
               hit the detector plane if the grid fields were zero, in
               meters. Useful for calculating the source profile.
           * - ``"v0"``
             - `~numpy.ndarray`, [``nparticles``, 3]
             - The velocity of each particle when it hit the detector
               plan if the grid fields were zero, in meters per second.
               The velocity is in a coordinate system relative to the
               detector plane. The components are [normal, horizontal,
               vertical] relative to the detector plane coordinates.
```

Looks like the dictionary keys were modified at some point...

rocco8773 (changes requested):
I don't like making suggestions like this, but `test_synthetic_radiograph` needs to be replaced with the code I've suggested below.  I'm making the suggestion like this because I want to illustrate what I'm looking for in tests.  The reformatting of the test comes for 2 key reasons...

1. Tests need to be parametrized.  In the original code if there's a failure on one of the initial lines of code, then none of the subsequent lines get executed.  The strategic splitting of the tests and use of `pytest.mark.parametrize` ensures subsequent tests still execute even if there was an earlier failure.
2. Just doing `h, v, i = cpr.synthetic_radiograph()` is not a test.  This line will pass regardless of what happens,  There needs to be specific assertions done on the results of `synthetic_radiograph`.

```python
def create_tracker_obj():
    # CREATE A RADIOGRAPH OBJECT
    grid = _test_grid("electrostatic_gaussian_sphere", num=50)

    sim = cpr.Tracker(grid, source, detector, verbose=False)
    sim.create_particles(int(1e4), 3 * u.MeV, max_theta=10 * u.deg)
    return sim


tracker_obj_simulated = create_tracker_obj().run(field_weighting="nearest neighbor")


@pytest.mark.slow
class TestSyntheticRadiograph:
    """
    Tests for
    `plasmapy.diagnostics.charged_particle_radiography.synthetic_radiograph`.
    """

    tracker_obj_not_simulated = create_tracker_obj()
    tracker_obj_simulated = create_tracker_obj()
    tracker_obj_simulated.run(field_weighting="nearest neighbor")
    sim_results = tracker_obj_simulated.results_dict.copy()

    @pytest.mark.parametrize(
        "args, kwargs, _raises",
        [
            # obj wrong type
            ((5, ), {}, TypeError),
            # size wrong type
            ((tracker_obj_simulated,), {"size": "not a Quantity"}, TypeError),
            # size not convertible to meters
            ((sim_results,), {"size": 5 * u.ms}, ValueError),
            # size wrong shape
            ((sim_results,), {"size": [-1, 1] * u.cm}, ValueError),
            # simulation was never run
            ((tracker_obj_not_simulated, ), {}, RuntimeError),
        ],
    )
    def test_raises(self, args, kwargs, _raises):
        """Test scenarios the raise an Exception."""
        with pytest.raises(_raises):
            cpr.synthetic_radiograph(*args, **kwargs)

    def test_warns(self):
        """
        Test warning when less than half the particles reach the detector plane.
        """
        sim_results = self.sim_results.copy()
        sim_results["nparticles"] = 3 * sim_results["nparticles"]
        with pytest.warns(RuntimeWarning):
            cpr.synthetic_radiograph(sim_results)

    @pytest.mark.parametrize(
        "args, kwargs, expected",
        [
            (
                # From a Tracker object
                (tracker_obj_simulated,),
                {},
                {
                    "xrange": [-0.036934532101889815, 0.03702186098974771] * u.m,
                    "yrange": [-0.03697401811598356, 0.037007901161403144] * u.m,
                    "bins": (200, 200)
                },
            ),
            (
                # From a dict
                (sim_results,),
                {},
                {
                    "xrange": [-0.036934532101889815, 0.03702186098974771] * u.m,
                    "yrange": [-0.03697401811598356, 0.037007901161403144] * u.m,
                    "bins": (200, 200)
                },
            ),
            (
                # From a dict
                (sim_results,),
                {
                    "size": np.array([[-1, 1], [-1, 1]]) * 30 * u.cm,
                    "bins": (200, 60)
                },
                {
                    "xrange": [-0.3, 0.3] * u.m,
                    "yrange": [-0.3, 0.3] * u.m,
                    "bins": (200, 60)
                },
            ),
        ],
    )
    def test_intensity_histogram(self, args, kwargs, expected):
        """Test several valid use cases."""
        results = cpr.synthetic_radiograph(*args, **kwargs)

        assert len(results) == 3

        x = results[0]
        assert isinstance(x, u.Quantity)
        assert x.unit == u.m
        assert x.shape == (expected["bins"][0],)
        assert np.isclose(np.min(x), expected["xrange"][0], rtol=1e4)
        assert np.isclose(np.max(x), expected["xrange"][1], rtol=1e4)

        y = results[1]
        assert isinstance(y, u.Quantity)
        assert y.unit == u.m
        assert y.shape == (expected["bins"][1],)
        assert np.isclose(np.min(y), expected["yrange"][0], rtol=1e4)
        assert np.isclose(np.max(y), expected["yrange"][1], rtol=1e4)

        histogram = results[2]
        assert isinstance(histogram, np.ndarray)
        assert histogram.shape == expected["bins"]

        """
        """
        bins = (200, 60)

        )
        )

        zero_mask = intensity == 0
        od = -np.log10(intensity/i0)

        assert np.allclose(od[~zero_mask], od_results[2][~zero_mask])
        assert np.all(np.isposinf(od_results[2][zero_mask]))
```

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1088,11 +1107,140 @@ def run(
...
+           * - ``"mag"``
+             - `float`
+             - The system magnification.
+           * - ``"nparticles"``
+             - `int`
+             - Number of particles in the simulation.
+           * - ``"max_deflection"``
+             - `~astropy.units.Quantity`
+             - The maximum deflection experienced by a particle in the
+               simulation, in radians.
```
rocco8773, line 1141: Since the rest of the values in this dictionary are numpy arrays, I'm thinking we drop units for ``max_deflection``.  We can do this either at the `max_deflection` property or just when building the `results_dict`.  I feel this will make the whole `results_dict` more consistent and, while reviewing `test_saving_ouput`, the `numpy` saving and loading system basically converts the `Quantity` back to a numpy array.

pheuer:
Thanks @rocco8773, I've implemented the suggestions. I'll try to take advantage of the class structure for testing more.

plasmapy/diagnostics/charged_particle_radiography.py
rocco8773: This section in the docstring needs to be updated to reflect that `max_deflection` is no longer a `Quantity` in the dictionary.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+        intensity = intensity_results[2]
+        zero_mask = intensity == 0
+        i0 = np.mean(intensity[~zero_mask])
+        od = -np.log10(intensity / i0)
+
+        assert np.allclose(od[~zero_mask], od_results[2][~zero_mask])
+        assert np.all(np.isposinf(od_results[2][zero_mask]))
+
+
+def test_saving_output(tmp_path):
```
rocco8773, line 602:
```suggestion
def test_saving_output(tmp_path):
    """Test behavior of Tracker.save_results."""
```

Test functions and classes still need docstrings.  It's just that docstrings are written for the programmer/contributor and not the user.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+
+    path = os.path.join(tmp_path, "temp.npz")
 
-    # CREATE A RADIOGRAPH OBJECT
     grid = _test_grid("electrostatic_gaussian_sphere", num=50)
     source = (0 * u.mm, -10 * u.mm, 0 * u.mm)
     detector = (0 * u.mm, 200 * u.mm, 0 * u.mm)
 
-    sim = prad.SyntheticProtonRadiograph(grid, source, detector, verbose=False)
+    sim = cpr.Tracker(grid, source, detector, verbose=False)
```
rocco8773, line 610:
```suggestion
    sim = create_tracker_obj()
```

Since we've now created the `create_tracker_obj()` function, lets use it instead of duplicating code.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+
+    # Test that output cannot be saved prior to running
+    with pytest.raises(RuntimeError):
+        sim.results_dict
```
rocco8773, line 614:
```suggestion
        _ = sim.results_dict
```

Just so linters and IDE's don't complain about something being returned to nother.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+
     sim.create_particles(1e4, 3 * u.MeV, max_theta=10 * u.deg)
     sim.run(field_weighting="nearest neighbor")
 
-    size = np.array([[-1, 1], [-1, 1]]) * 30 * u.cm
-    bins = [200, 60]
+    res1 = sim.results_dict
```
rocco8773, line 619:
```suggestion
    sim.run(field_weighting="nearest neighbor")
    results_1 = sim.results_dict
```

`create_particles` is encompassed by `create_tracker_obj()`

Also, use `results_1` instead of `res1` for readability and clarity.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+
+    # Save result
+    sim.save_results(path)
```
rocco8773, line 622:
```suggestion
    path = os.path.join(tmp_path, "temp.npz")
    sim.save_results(path)
```

Keep `path` definition near it's use for readability.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+
+    # Load result
+    res2 = dict(np.load(path, "r", allow_pickle=True))
```
rocco8773, line 625:
```suggestion
    results_2 = dict(np.load(path, "r", allow_pickle=True))
```

Use `results_2` for more explicitly.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
 
-    # Test size is None, default bins
-    h, v, i = sim.synthetic_radiograph()
+    assert np.all(res1["x"] == res2["x"])
```
rocco8773, line 627:
```suggestion
    assert set(results_1.keys()) == set(results_2.keys())
    for key in results_1.keys():
        assert np.allclose(results_1[key], results_2[key])
```

Make tests more robust.  Everything about the dictionary needs to be tested.

rocco8773 (changes requested):
Refinement of `test_saving_output`.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,203 @@ def test_run_options():
...
+        sim.load_particles(sim.x, sim.v)
+
+    # Error from adding wire mesh
+    with pytest.raises(RuntimeError):
+        sim.add_wire_mesh(
+            np.array([0, -2, 0]) * u.mm,
+            (2 * u.mm, 1.5 * u.mm),
+            9,
+            20 * u.um,
+        )
```
rocco8773, line 657:
```suggestion
@pytest.mark.parametrize(
    "case",
    ["creating particles", "loading particles", "adding a wire mesh"],
)
def test_cannot_modify_simulation_after_running(case):
    """
    Test that a Tracker objection can not be modified after it is
    run (Tracker.run).
    """

    sim = create_tracker_obj()
    sim.run(field_weighting="nearest neighbor")

    # Error from creating particles
    with pytest.raises(RuntimeError):
        if case == "creating particles":
            sim.create_particles(1e4, 3 * u.MeV, max_theta=10 * u.deg)
        elif case == "loading particles":
            sim.load_particles(sim.x, sim.v)
        elif case == "adding a wire mesh":
            sim.add_wire_mesh(
                np.array([0, -2, 0]) * u.mm,
                (2 * u.mm, 1.5 * u.mm),
                9,
                20 * u.um,
            )
        else:
            pytest.fail(f"Unrecognized test case '{case}'.")
```

`test_cannot_modify_simulation_after_running` needs a docstring and needs to be parametrize.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
pheuer: I've no issue adding this, but I don't think this docstring adds any information that isn't already in the name of the function.

pheuer:
Ok, I made those changes. I think that, if we're at the point of re-naming variables in the test for clarity, it's time to merge this PR and move on to other work.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
rocco8773: It does because the function name does not indicate that you're testing the `Tracker.save_results` method.  This test file is testing several different radiography functions, methods, and classes, so one can not just assume this test function is working on the `Tracker` class.

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -459,26 +458,202 @@ def test_run_options():
...
+        od = -np.log10(intensity / i0)
+
+        assert np.allclose(od[~zero_mask], od_results[2][~zero_mask])
+        assert np.all(np.isposinf(od_results[2][zero_mask]))
+
+
+def test_saving_output(tmp_path):
+    """Test behavior of Tracker.save_results."""
+
+    path = os.path.join(tmp_path, "temp.npz")
```
rocco8773, line 605:
```suggestion
```

namurphy:
You have my continued thanks, @pheuer!

rocco8773:
I just have to make sure all of the cross-ref repairs to the change log entries work and then I'll merge it.

rocco8773 (approved):

PR #1073, opened 2021-03-13
Improve consistency and ReST formatting in documentation

namurphy:
Some minor touch-ups and updates to the documentation, including to the stability matrix.   The changes should be pretty straightforward.

PR #1136, opened 2021-05-11
Rename integer charge to charge number

### LINKED ISSUE #945: Create @deprecated decorator based off of Astropy's
Astropy has a decorator, `astropy.utils.decorators.deprecated`, that is used to mark deprecated functions.  We should create a similar decorator based off of Astropy's and put it in `plasmapy.utils.decorators`.  It should use `PlasmaPyDeprecationWarning` in `plasmapy.utils.exceptions`, which I'm creating in #920.  I'm thinking our function could wrap Astropy's decorator while keeping the same positional and keyword arguments, while changing the default warning to `PlasmaPyDeprecationWarning`.  

This is potentially a good first issue for someone who is new to the package but has some experience with with decorators.

### LINKED ISSUE #1126: Deprecate `integer_charge` in favor of `charge_number`
In `plasmapy.particles` and elsewhere, we use the name `integer_charge` to represent the ratio of the charge of a particle to the elementary positive charge.  However, there are a few problems with the name "integer charge":

 - "Integer charge" is not the standard name for [charge number](https://en.wikipedia.org/wiki/Charge_number).
 - It is not always going to be an integer!  For example in the following situations...
   - Quarks!   (which would probably be worth adding to `plasmapy.particles` for quark-gluon plasma and really high energy density plasmas)
   - `CustomParticle` instances, especially when representing a composite particle
   - Mean charges (for example in the ionization state classes)
   - Particle-in-cell simulations where the electron-to-ion mass ratio is greater than physical (i.e. electrons are assumed to be heavy)
 - The first result in a search for "integer charge" leads to an article on "integer charge transfer" in semiconductors, not a resource that refers to charge number.  

I propose that we deprecate the name `integer_charge` in favor of `charge_number`, and use the name "charge number" to refer to this concept in our documentation.  This would be a significant change to the `plasmapy.particles` API so we would want to not remove `integer_charge` until a couple of releases have passed.  This would definitely require discussion before implementing it.

namurphy:
This PR replaces usage of the term "integer charge" and code elements named like `integer_charge` with the term "charge number" and `charge_number`.  This is because [charge number](https://en.wikipedia.org/wiki/Charge_number) is the technical name for this concept (see #1126).  The things named after "integer charge" are going to be kept for a few releases and then removed.  

As part of this PR, I'm likely going to use a variant of the `astropy.utils.decorators.deprecated` decorator.

Closes #1126.  Closes #945.


StanczakDominik (changes requested):
LGTM in general; test failures seem to point to a mis-import probably since 5d2d129; the doc failure is unrelated ([see Element](https://matrix.to/#/!hkWCiyhQyxiYJlUtKF:matrix.org/$162080590746337jdVlJ:matrix.org?via=matrix.org&via=openastronomy.org&via=gitter.im)). A couple of docstring mishaps in there.

docs/particles/particle_class.rst
```diff
@@ -59,7 +59,727 @@ the corresponding `~plasmapy.particles.Particle` object.
...
+>>> electron.charge_number
+-1
+>>> triton.mass_number
+3
+
+Some of these properties are returned as a
+
+>>> proton.atomic_number
+1
+>>> electron.charge_number
```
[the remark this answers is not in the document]
namurphy: 🙀  Weird!  No idea how this happened.  🙀  Weird!  No idea how this happened.  🙀  Weird!  No idea how this happened.  🙀  Weird!  No idea how this happened.  🙀  Weird!  No idea how this happened.  🙀  Weird!  No idea how this happened. 🙀  Weird!  No idea how this happened. 🙀  Weird!  No idea how this happened. 🙀  Weird!  No idea how this happened.

namurphy:
I'm not sure why sections of docstrings ended up getting repeated.  It could have been a mishap with auto-renaming via PyCharm, though it could have been a side effect of coding while listening to a BBC dramatization of my favorite book series. 🐉  In any case, I'll need to double check for this problem before this gets merged, and I have a few other things I need to finish up too.  Thanks for finding that!

namurphy:
My to do list for this PR and beyond.  It's mostly tedious work at this point.

 - [ ] Add back stand-alone `integer_charge` function and deprecate it
 - [ ] Test the `integer_charge` function
 - [ ] Add back the `IonizationState.integer_charges` and deprecate it
 - [ ] Test `IonizationState.integer_charges`
 - [ ] Add back `ParticleList.integer_charges` and deprecate it
 - [ ] Test `ParticleList.integer_charges`
 - [ ] Check the files changed to make sure I didn't add in excessive extra redundant redundancies again.
 - [ ] Add a change log entry
 - [ ] Create an issue to remove deprecated features from `plasmapy.particles` and label it with the milestone for, say, 0.9.0
 - [ ] Create an issue in PlasmaPy-NEI repo to update all of this

StanczakDominik:
Probably don't need to have exhaustive tests for integer_charge if it's just going to be a deprecated alias for charge_number!

namurphy:
> Probably don't need to have exhaustive tests for integer_charge if it's just going to be a deprecated alias for charge_number!

Very true.  I was planning on having maybe one or two tests for each function/method...just enough to make sure it works.

docs/particles/decorators.rst
```diff
@@ -8,20 +8,21 @@ Decorators
...
-When calculating plasma parameters, we very frequently need to access
+When calculating plasma parameters, we frequently need to access
 the properties of the particles that make up that plasma. The
-`~plasmapy.particles.decorators.particle_input` decorator allows functions
-and methods to easily access properties of different particles.
+`~plasmapy.particles.decorators.particle_input` decorator allows
+functions and methods to easily access properties of different particles.
 
-The `~plasmapy.particles.particle_input` decorator takes valid
+The `~plasmapy.particles.decorators.particle_input` decorator takes valid
```
namurphy, line 16: I expanded a few ReST links in `particle_class.py` because I was getting an error with the docs that did not happen to say which file the error was in.  So, a little bit of scope creep here, but these should be quick to review.  (Turns out the error was because I had half of a link in my incomplete change log entry.)

plasmapy/particles/parsing.py
```diff
@@ -441,12 +441,12 @@ def _reconstruct_ion_symbol(
...
         if Z > _elements[element]["atomic number"]:
             raise InvalidParticleError(
-                f"The integer charge Z = {Z} cannot exceed the atomic number "
+                f"The charge number Z = {Z} cannot exceed the atomic number "
                 f"of {element}, which is {_elements[element]['atomic number']}."
             )
         elif Z <= -3:
             warnings.warn(
-                f"Particle '{argument}' has an integer charge "
+                f"Particle '{argument}' has an charge number "
```
namurphy, line 449:
```suggestion
                f"Particle '{argument}' has a charge number "
```

plasmapy/particles/atomic.py
```diff
@@ -378,16 +395,16 @@ def electric_charge(particle: Particle) -> u.C:
         If charge information for the particle is not available.
 
     `~plasmapy.particles.exceptions.ParticleWarning`
-        If the input represents an ion with an integer charge that is
+        If the input represents an ion with an charge number that is
```
namurphy, line 398:
```suggestion
        If the input represents an ion with a charge number that is
```

plasmapy/particles/parsing.py
namurphy:
```suggestion
                f"Particle '{argument}' has a charge number "
```

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -152,15 +160,28 @@ def test_instantiation(self, test_name):
...
+        expected_charge_numbers = np.arange(instance.atomic_number + 1)
+        assert np.allclose(instance.charge_numbers, expected_charge_numbers)
+
     @pytest.mark.parametrize("test_name", test_names)
     def test_integer_charges(self, test_name):
         """
-        Test that an `IonizationState` instance has the correct integer
-        charges.
+        Test that `IonizationState.integer_charges` instance has the
+        correct charge numbers and issues a future warning.
```
namurphy, line 177:
```suggestion
        Test that `IonizationState.integer_charges` has the
        correct charge numbers and issues a future warning.
```

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -402,8 +423,8 @@ def test_getitem(self, test_name):
         instance = self.instances[test_name]
         particle_name = instance.base_particle
 
-        integer_charges = np.arange(instance.atomic_number + 1)
-        symbols = [particle_symbol(particle_name, Z=Z) for Z in integer_charges]
+        charge_numnbers = np.arange(instance.atomic_number + 1)
+        symbols = [particle_symbol(particle_name, Z=Z) for Z in charge_numnbers]
```
namurphy, line 427:
```suggestion
        charge_numbers = np.arange(instance.atomic_number + 1)
        symbols = [particle_symbol(particle_name, Z=Z) for Z in charge_numbers]
```

plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -414,7 +435,7 @@ def test_getitem(self, test_name):
         # these objects, we compare the string representations instead
         # (see Astropy issue #7901 on GitHub).
 
-        for keys in zip(integer_charges, symbols, particles):
+        for keys in zip(charge_numnbers, symbols, particles):
```
namurphy, line 438:
```suggestion
        for keys in zip(charge_numbers, symbols, particles):
```

plasmapy/particles/atomic.py
```diff
@@ -322,34 +325,48 @@ def integer_charge(particle: Particle) -> Integral:
         If charge information for the particle is not available.
 
     `~plasmapy.particles.exceptions.AtomicWarning`
-        If the input represents an ion with an integer charge that is
+        If the input represents an ion with an charge number that is
```
namurphy, line 328:
```suggestion
        If the input represents an ion with a charge number that is
```

rocco8773 (dismissed):
Add a stub file in `docs/api_static` for `plasmapy.utils.decorators.deprecation`.

plasmapy/utils/decorators/deprecation.py
```diff
@@ -0,0 +1,16 @@
...
+from plasmapy.utils.exceptions import PlasmaPyDeprecationWarning
+
+deprecated = partial(
+    astropy.utils.decorators.deprecated,
+    warning_type=PlasmaPyDeprecationWarning,
+)
+"""
+A wrapper of `astropy.utils.decorators.deprecated` that by default assumes
+a warning type of `~plasmapy.utils.exceptions.PlasmaPyDeprecationWarning`.
+"""
```
rocco8773, line 16: So...While I like `functools.partial`, it does cause some documentation issues here.  If you go look at the RTD builds, `deprecated` is showing up under "Variables & Attributes."  This is because `deprecated` is missing [all the attributes that would make it look like a function](https://github.com/python/cpython/blob/eaede0ded72e67cee4a91c086847d54cb64ca74c/Lib/inspect.py#L277-L288).  This makes me side on that `functools.partial` should be restricted to being used in code and not for outward functionality/API.

AS a work around, here's my suggested replacement.  It's not as elegant, but it does what it should.

```suggestion
__all__ = ["deprecated"]

import inspect

from astropy.utils.decorators import deprecated as astropy_deprecated

from plasmapy.utils.exceptions import PlasmaPyDeprecationWarning


def deprecated(*args, warning_type=PlasmaPyDeprecationWarning, **kwargs):
    """
    A wrapper of `astropy.utils.decorators.deprecated` that by default assumes
    a warning type of `~plasmapy.utils.exceptions.PlasmaPyDeprecationWarning`.
    """
    return astropy_deprecated(*args, warning_type=warning_type, **kwargs)


# override deprecated's signature
asig = inspect.signature(astropy_deprecated)
new_default = asig.parameters["warning_type"].replace(
    default=PlasmaPyDeprecationWarning
)
new_sig = dict(asig.parameters)
new_sig["warning_type"] = new_default
new_sig = inspect.Signature(
    parameters=tuple(new_sig.values()), return_annotation=asig.return_annotation,
)
deprecated.__signature__ = new_sig
"""
```

namurphy: Thanks — I think we can do that.  

> This makes me side on that `functools.partial` should be restricted to being used in code and not for outward functionality/API.

This seems quite sensible to me.  If we end up using `functools.partial` a lot then we could put the signature override into a function that implements your suggestion, but that should wait until we have a demonstrated need for it.

plasmapy/particles/tests/test_ionization_collection.py
```diff
@@ -650,7 +650,7 @@ def test_getitem_two_indices(self, indices):
         integer_charge = indices[1]
```
rocco8773, line 650: To be consistent, it would be a good idea to rename cases like this one to `charge_number`.

plasmapy/particles/tests/test_atomic.py
```diff
@@ -781,3 +783,8 @@ def test_missing_atomic_data(self):
 @pytest.mark.parametrize("particle, electron", str_electron_table)
 def test_is_electron(particle, electron):
     assert _is_electron(particle) == electron
+
+
+def test_integer_charge():
+    with pytest.warns(PlasmaPyFutureWarning):
+        assert integer_charge("Fe 20+") == 20
```
rocco8773, line 790: If you want to be thorough, then it would be worthwhile to do a `mock` test here to ensure `integer_charge` calls `charge_number`.   It's a bit overkill, but it does enforce the connection between the two.

plasmapy/particles/tests/test_ionization_collection.py
namurphy: Good catch!  I had thought I had found all of those...

docs/particles/decorators.rst
rocco8773: I thought I got all of these with PR #1105.  I guess not 😬. If you haven't done so yet, it would be good to merge master into this PR because #1105 did touch a lot of these files.

rocco8773:
Do you want to merge #1073 first because it sure looks like you branched this PR off of that one?

plasmapy/particles/tests/test_atomic.py
namurphy: It's probably more effort than is worth it for a deprecated feature, though this idea is worth keeping in mind for future features that we intend to keep around for a while.

docs/particles/decorators.rst
namurphy: Yep, I did that earlier.  And with changes like that which are hard to search for and automate fixes, it's only natural to miss some.

namurphy:
> Do you want to merge #1073 first because it sure looks like you branched this PR off of that one? If so, that would avoid the scope creep of this PR.

I'd rather merge this one first and then go back to #1073 and enjoy the wonders of resolving git conflicts.  Thanks for the quick review!

(Also: it's amazing how productive I can be when I'm procrastinating writing emails that I don't want to write.)

StanczakDominik (approved):

PR #1146, opened 2021-05-20
Remove deprecated Particle.particle

namurphy:
This PR removes `Particle.particle`, which had previously been deprecated in favor of `Particle.symbol` in #984.  This wasn't a widely used feature so it should be fine to have it emit a `FutureWarning` in 0.6.0, and then to remove it in the next release (0.7.0).


StanczakDominik (approved):
How about a magic trick? I'm going to make this property disappear...

PR #1147, opened 2021-05-20
Set up ReST substitutions for docs

namurphy:
This PR follows the example of Astropy in setting up substitutions in docs (see their `docs/common_links.txt` and `docs/conf.py` files).  Instead of having `~plasmapy.particles.particle_class.Particle`, we can use `|Particle|`.  

This PR adapts code from Astropy but we can do that since we have a copy of Astropy's license.


namurphy:
- The substantive things to review are `docs/conf.py` and `docs/common_links.py`.
 - The other changes are mostly:
   - Doing things like ``` `~plasmapy.particles.particle_class.Particle` ```→ `|Particle|`
   - Adjusting paragraphs so that the lines are ≲72 characters as per PEP8
 - @rocco8773 and I had discussed about whether or not to include notation like `|Particle|` in docstrings and didn't reach consensus, except that we both agreed that it would be reasonable to include that notation in the docstrings within `|Particle|`.  So, I limited changes within `plasmapy.particles` docstrings to cases like this.  
 - I changed the flake8 settings so that it doesn't check for undefined substitutions in docstrings since it didn't seem to find the substitutions defined in the separate file.  
 - I mostly did this for `plasmapy.particles` and decided to stop because I'm finding this to be kind of fun and was risking some scope creep.

namurphy:
Stuff I still need to do:
 - [ ] Add a discussion of this into the documentation guide in the dev guide.

docs/common_links.txt
```diff
@@ -0,0 +1,60 @@
...
+.. ------------------
+
+.. |CustomParticle| replace:: :class:`~plasmapy.particles.particle_class.CustomParticle`
+.. |DimensionlessParticle| replace:: :class:`~plasmapy.particles.particle_class.DimensionlessParticle`
+.. |IonicLevel| replace:: :class:`~plasmapy.particles.ionization_state.IonicLevel`
+.. |IonizationState| replace:: :class:`~plasmapy.particles.ionization_state.IonizationState`
+.. |IonizationStateCollection| replace:: :class:`~plasmapy.particles.ionization_state_collection.IonizationStateCollection`
+.. |Particle| replace:: :class:`~plasmapy.particles.particle_class.Particle`
+.. |particle_input| replace:: :func:`~plasmapy.particles.decorators.particle_input`
+.. |ParticleList| replace:: :class:`~plasmapy.particles.particle_collection.ParticleList`
```
rocco8773, line 15:
```suggestion
.. |ParticleList| replace:: :class:`~plasmapy.particles.particle_collections.ParticleList`
```

docs/common_links.txt
```diff
@@ -0,0 +1,60 @@
+.. These are ReST substitutions and links that can be used throughout the docs
```
rocco8773, line 1: I think this should be a `.rst` file and not a `.txt` file.  It's already formatted as a RST file and my IDE is going mad with RST commands in a normal txt file.

namurphy: I just tried changing that and got an error that the substitution for `CustomParticle` is being defined twice.  I think what's happening is that this is being treated as a source file now when it shouldn't be.  A little less likely is that `rst_epilog`...which contains the contents of this file...is being appended to this file.  I'm attempting to add this file name to `exclude_patterns` in `conf.py`...

docs/particles/particle_class.rst
```diff
@@ -1,25 +1,25 @@
 .. _particle-class:
 
-`~plasmapy.particles.particle_class.Particle` Class
-***************************************************
+|Particle| Class
+****************
```
rocco8773, line 4: I'm surprised substitutions in headings don't produce build warnings due to the length of underline being shorter than the heading after the substitution.  I'm really curious what's going on behind the scenes here, but I'm going to resist going down the rabbit hole.

namurphy: Given my experience with rabbit holes, we should probably rename them into dragon holes.

docs/particles/functional.rst
```diff
@@ -71,7 +72,7 @@ accessed using `~plasmapy.particles.standard_atomic_weight`.
 <Quantity 207.2 u>
```
rocco8773, line 72: I can't directly comment on the line, but go up 3 lines from here and up the xref needs `.atomic.` added to it...

`~plasmapy.particles.atomic.standard_atomic_weight`

namurphy: Looks like there are a few others in that file. I'll learn my lesson and start a separate PR with a well-defined scope.

docs/particles/decorators.rst
```diff
@@ -8,21 +8,19 @@ Decorators
 Passing `~plasmapy.particles.Particle` objects to functions and methods
 =======================================================================
```
rocco8773, line 9:
```suggestion
Passing |Particle| objects to functions and methods
===================================================
```

docs/install.rst
```diff
@@ -30,22 +30,21 @@ PlasmaPy also depends on the following packages for optional features:
...
 
-
 .. _install-pip:
 
 Installation with pip
 ---------------------
 
 To install the most recent release of PlasmaPy on `PyPI`_
-with `pip <https://pip.pypa.io/en/stable/>`_ into an existing Python environment
+with `pip`_ into an existing Python environment
```
rocco8773, line 47: If we add `pip` to the `intersphinx_mappings` dict, then `` `pip` `` will likely automatically cross-reference.

docs/development/doc_guide.rst
```diff
@@ -61,3 +59,17 @@ Narrative Documentation
 
 * Each subpackage must have narrative documentation describing its
   use.
+
+Substitutions
+=============
+
+The `common_links.txt <https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/common_links.txt>`_
```
rocco8773, line 66:
```suggestion
The `common_links.rst
<https://github.com/PlasmaPy/PlasmaPy/blob/master/docs/common_links.rst>`_
```

You can actually split the hyperlink like this on two or more lines and it will still work.  This was really nice when I discovered it.

changelog/1147.doc.rst
```diff
@@ -0,0 +1,2 @@
+Added common substitutions for classes and links that can be used in
+docs.
```
rocco8773, line 2:
```suggestion
Implement a framework to define and use common `Sphinx substitutions
<https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
#substitutions>`_ across the narrative documentation and docstrings.
```

rocco8773 (changes requested):
This is looking good.  Definitely makes the narrative documentation more readable.

Just address the following comments and then this gets my approval.

* https://github.com/PlasmaPy/PlasmaPy/pull/1147#discussion_r641862432
* https://github.com/PlasmaPy/PlasmaPy/pull/1147#discussion_r641864373
* (optional) https://github.com/PlasmaPy/PlasmaPy/pull/1147#discussion_r641864940

rocco8773 (approved):
🚀🥳

docs/install.rst
namurphy: I tried adding `pip` to `intersphinx_mappings` but the link didn't work, so I reverted the change rather than go down a dragon hole.  I'd be okay with this change at some future time.

Issue #1055, opened 2021-03-04
Add glossary that defines common and potentially ambiguous terms


namurphy:
With respect to choosing variable names, it's often suggested that each word or term corresponds to a single concept, and that each concept corresponds to a single word or term.  In practice, this can be difficult when science or programming terminology is used in multiple different but somewhat related ways.  I'm wondering if it would be helpful to have a glossary or set of definitions in our documentation to help reduce potential ambiguity.  A possibility would be to put this in the development guide.

_Example 1:_ "ionization state" can refer to (1) a single ionic level, (2) all ionic levels of a single element or isotope, or (3) all ionic levels of all elements or isotopes (#1053).  

_Example 2:_  Another difficulty for me has been coming up with a word or term that encompasses both "element" and "isotope", especially in the context of ions.  I'm wondering if we could use "base atom" for this.

namurphy:
This issue has the potential for scope creep...since it has the potential to be repurposed as a glossary for plasma science as a whole.  We'd want to carefully define the scope on the page, and probably only include terms that are relevant to code development (including some commonly used programming terms like "instance" and "instantiate" that are not widely known in the plasma community).

namurphy:
@rocco8773 brought up a point earlier that a glossary is related to the idea of having a lexicon for different variable names (i.e. making sure that `n_e` is consistently used as the argument for electron number density, `T_i` for ion temperature, etc.).  We could possibly put both together in the same part of the docs.

rocco8773:
Yeah, I could potentially see a combined lexicon/glossary.  I still feel a lexicon would be more beneficial to package development, because it would give some guidance to developers/contributors. 

Things I could see put into a lexicon entry....

* physical/plasma parameter name (e.g. "electron temperature")
* description of physical parameter
* primary parameter name (e.g. `T_e`)
* alternative parameter names (e.g. `Te`, `Te_arr`, etc.)
    * these name could only be used when there conflicts with the  primary 
* primary unit `K`
* `validate_quantities` settings `can_be_negative=False`
* primary parameter type `Quantity`
* alternative parameter type
* primary docstring description
* alternative docstring description

As you can see, this can runaway quickly.  I don't think every entry to the lexicon needs all this information. Some may just be a physical parameter name and description.

If we could utilize `sphinx-copybotton` for the docstrings, then a contributor could directly copy-paste a docstring.  That would be  nice!

namurphy:
[...]

PR #1149, opened 2021-05-21
Begin a project glossary

namurphy:
This PR begins a glossary, following the lead of Astropy.  This is part of #1055, but might not close it if this PR does not define potentially ambiguous terms.


namurphy:
I've been attempting to get `particle-like` to link to the term in the glossary in the docstring for `Particle`.  Haven't figured it out yet, but still waiting for the docs to build again.

namurphy:
This one is ready for review, I think. It's not intended to be complete but it's a start.  I included a few things in there that I remember being confused about, and there are undoubtedly a bunch of things that should still be added.  

The thing I like most about this is that if we do ``` :term:`particle-like` ``` somewhere in the docs, it'll link to the term in the glossary.  [In #1158, I've been trying to set it up so that the linking is done automatically (i.e., if `particle-like` is in the description for a parameter, then it'll link to the appropriate term even if it does not include the `:term:` directive).  But, I have not been able to get that to work yet.]

This could potentially go into 0.7 but it's fine to wait until 0.8.

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+      that type.  For example, ``"He 2+"`` is :term:`particle-like`
+      because it can be used to instantiate |Particle|.
+
+   charge number
+      The charge of a particle in units of elementary charge.
+
+   keyword-only
+      An argument that must be provided along with the name of that
+      argument. If ``z`` is a keyword-only argument to ``f(z)``, then
+      it can be provided as ``f(z=2)`` but not ``f(2)``.
```
StanczakDominik, line 23: Could use a link to the python docs on this.

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+
+   kwargs
+      An abbreviation for `keyword arguments`_.
+
+   particle-like
+      An `object` is *particle-like* if it is a |Particle| or
+      |CustomParticle|, or can be cast into one.
```
StanczakDominik, line 30: So it's just an instance of an abstract particle or a string that cleanly initializes one. I did just wake up, haven't had tea yet, excuse my confusion!

StanczakDominik (approved):
Two minor comments!

docs/glossary.rst
namurphy: Good idea.  I linked to the terms in the Python glossary for argument and parameter, which include the descriptions for keyword-only, etc.

docs/glossary.rst
namurphy: Pretty much!  It could also be an `int` representing an atomic number.  I also added the line from the docstring for `ParticleLike` that describes why a `DimesionlessParticle` isn't particle-like.

StanczakDominik (approved):

namurphy:
@StanczakDominik — thank you for the reviews!

PR #1151, opened 2021-05-22
Added notebook to environment.yml

FinMacDov:
Added notebook to be installed by pip so jupyter notebook is there to be launched to run example cases with extensions ipynb without further steps.

      docstrings.


StanczakDominik:
Well, given that the conda environment.yml is supposed to be a full dev environment, I don't see why not!

... except for the minor detail that we're not testing that env for whether tests run with it right now, but that's something I can get to later on. In the meantime, thanks, let's get this in (given tests pass)!

StanczakDominik (approved):

PR #1154, opened 2021-05-25
Create IonizationState attribute for a ParticleList of the ionic levels

namurphy:
This PR adds an attribute to `IonizationState` that provides a `ParticleList` containing the different ionic levels from the neutral to fully ionized.  For example, `IonizationState("He-4").as_particle_list` would provide `ParticleList(["He-4 0+", "He-4 1+", "α"])` (where I unnecessarily used the unicode character simply for the enjoyment of @rocco8773 😬 and also because I am pure evil).  This is a prequel to #1028 but separate enough change to warrant its own change log entry.


plasmapy/particles/tests/test_ionization_state.py
```diff
@@ -408,7 +408,7 @@ def test_getitem(self, test_name):
 
         charge_numbers = np.arange(instance.atomic_number + 1)
         symbols = [particle_symbol(particle_name, Z=Z) for Z in charge_numbers]
-        particles = instance._particle_instances
+        particles = instance.as_particle_list
```
namurphy, line 411: Fortunately this was pretty much already tested...except that I had previously written a test of a private attribute because I am pure evil 😬, when it's actually a better practice to only test the public API.  Testing the private API or implementation details can lead to brittle tests when the implementation changes but not the public API...I've since learned.

StanczakDominik: I would respectfully disagree with the general statement here!

Without even thinking of the public/private split, the "test everything" strategy as opposed to "test only the public API" gets you more certainty of knowing the precise movements of each gear in the multi-component machine that's our public API.

If you test your private API as well as the private one, and change an implementation detail (say, fixing a numerical stability bug, for concreteness), you know the tests for that private part will have their results changed; but you also know your results propagate to the public API. Otherwise, you only get the indirect result.

Of course, there is also such a thing as over-testing, overengineering tests, etc, and that can definitely be bad.

plasmapy/particles/ionization_state.py
```diff
@@ -307,6 +308,9 @@ def __init__(
                 )
 
         self._particle = particle
+        self._particle_list = ParticleList(
+            [Particle(particle, Z=i) for i in range(0, particle.atomic_number + 1)]
```
StanczakDominik, line 312:
```suggestion
            [Particle(particle, Z=i) for i in range(0, self._number_of_particles)]
```

This got added recently. No reason not to use it! :)

Also, I kind of wonder whether you might not be able to use a generator here, just like the first of the following:

```python
[ins] In [1]: list(x**2 for x in range(10))
Out[1]: [0, 1, 4, 9, 16, 25, 36, 49, 64, 81]

[ins] In [2]: list([x**2 for x in range(10)])
Out[2]: [0, 1, 4, 9, 16, 25, 36, 49, 64, 81]
````

But there doesn't seem to be any actual performance difference between the two, so, meh.

File: plasmapy/particles/ionization_state.py [hunk 3]
```diff
@@ -710,6 +714,14 @@ def base_particle(self) -> str:
         """The symbol of the element or isotope."""
         return self.isotope if self.isotope else self.element
 
+    @property
+    def as_particle_list(self) -> ParticleList:
+        """
+        A `~plasmapy.particles.particle_collections.ParticleList` of
+        the ionic levels.  The index corresponds to the charge number.
+        """
+        return self._particle_list
```
StanczakDominik, line 723: This is why I was harping on `functools.cached_property` a while back; since you're not using `self._particle_list` anywhere else, using that decorator here and moving the definition of `self._particle_list` here would get us more "spatially concentrated" code, if you catch my cross-field drift.

StanczakDominik (approved):
Short and sweet! Looks good, two comments!

plasmapy/particles/tests/test_ionization_state.py
namurphy: True, those are good points.  Perhaps the best testing strategy here is to balance all of the above factors.

The strategy I brought up came from [this book](https://www.manning.com/books/unit-testing) which I started listening to on my morning pandemic era walks.  There's some really good stuff in there, as well as some stuff that was a lot more relevant to software for business applications than for scientific applications.

plasmapy/particles/ionization_state.py
namurphy: Oh good point!  I like the idea for using a generator here.

And now that you mention it, I do remember making a code review comment on that very attribute!

plasmapy/particles/tests/test_ionization_state.py
StanczakDominik: Yeah, that definitely felt business-y! 😅

namurphy:
This PR now needs to be merged after #1223 is merged.

StanczakDominik:
How about `.to_list()`? Making this a method is also closer to precedent for this kind of behavior in numpy arrays, and helps signify that this creates a whole new object without any "persistence" when attached to the main object. E.g. I can't do `ionizationstate.ions[3].charge_number = 7`.

plasmapy/particles/ionization_state.py
```diff
@@ -718,25 +726,15 @@ def atomic_number(self) -> int:
...
-        """
-        return [
-            Particle(self._particle.symbol, Z=i) for i in range(self.atomic_number + 1)
-        ]
-
     @property
     def ionic_symbols(self) -> List[str]:
         """The ionic symbols for all charge states."""
-        return [particle.ionic_symbol for particle in self._particle_instances]
+        return self.ions.symbols
```
StanczakDominik, line 732:
```suggestion
        return self.to_list().symbols
```

plasmapy/particles/ionization_state.py
```diff
@@ -718,25 +726,15 @@ def atomic_number(self) -> int:
...
 
     @property
     def charge_numbers(self) -> np.ndarray:
         """An array of the charge numbers."""
-        return np.arange(0, self.atomic_number + 1, dtype=int)
+        return self.ions.charge_number
```
StanczakDominik, line 737:
```suggestion
        return self.to_list().charge_number
```

StanczakDominik (changes requested):
The changelog needs editing, and I'm making two minor follow-up usage renames that might solve the tests failing:

StanczakDominik (approved):
LGTM! :)

PR #1157, opened 2021-05-28
Enable math in documentation to be enclosed by dollar signs

namurphy:
The [sphinx-math-dollar](https://www.sympy.org/sphinx-math-dollar/) extension to Sphinx allows us to write math in docs as `$x$` instead of ``` :math:`x` ```.  This is the way in-line math is represented in LaTeX, so this is familiar notation to many plasma physicists.  I attempted to change `Alfven_speed` to see if this works.


namurphy:
I'm not sure why but I keep getting the following error, which appears to indicate an incompatibility between `nbsphinx.GalleryToc` and `sphinx-math-dollar`.  I attempted the instructions on the [sphinx-math-dollar docs](https://www.sympy.org/sphinx-math-dollar/) to exclude this but haven't been able to get it to work.  Hm...

```
Exception occurred:
  File "/home/runner/work/PlasmaPy/PlasmaPy/.tox/build_docs/lib/python3.8/site-packages/docutils/nodes.py", line 2018, in unknown_visit
    raise NotImplementedError(
NotImplementedError: <class 'sphinx_math_dollar.extension.MathDollarReplacer'> visiting unknown node type: GalleryToc
The full traceback has been saved in /tmp/sphinx-err-kl_ttji_.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

StanczakDominik:
Oh yeah, I recall the nbsphinx main dev being quite vocally against $math$. That's still weird though...

namurphy:
I'm going to close this for the time being.  We could fix this upstream by adding a `try`/`except` clause in `sphinx-math-dollar` to keep the build going in case of a failure like this, but then the extension would (I think?) not be able to be used everywhere in our documentation (like presumably our notebooks).  

I had previously tried to use the `sphinx-math-dollar` option to skip this particular class, but I wasn't able to get it to work and couldn't figure out why.  

In any case, we could revisit this in the future, but it's a low priority.

PR #1159, opened 2021-05-29
Fixing binder bug

FinMacDov:
I have implemented a bug fix by creating a new yml file in the binder dir, as Binder couldn't run example Notebooks. There might be better fixes than the one I proposed, I am Binder noob, so if there less clunky ways of fixing this I am happy to implement it. 

More info:
The example notebooks currently do not run on Binder as it can't find automated-documentation-test.txt. It looks like this file has been removed by accident [(link to missing file)](https://github.com/PlasmaPy/PlasmaPy/blob/7972832a17a79e959f0a712f1794862f4362ecc9/requirements/automated-documentation-tests.txt) or the requirements text need to be pointed somewhere else [link to requirements.txt](https://github.com/PlasmaPy/PlasmaPy/blob/master/binder/requirements.txt).

      docstrings.


namurphy:
Thank you, @FinMacDov!  We've been in the process of consolidating requirements in the `requirements/` directory so that they're not defined in a bunch of different places, so I updated this pull request to pull the requirements from the files from there.  That way, when the requirements for PlasmaPy are updated, we won't have to remember to update the binder requirements separately.

PR #1161, opened 2021-05-29
Fix binder links and change branch name to main in stable

namurphy:
This is to change the branch name in stable in an attempt to make binder work.  This is different than #1159.

One thing I'm wondering about is why there was an attempt to access `master` when launching a binder notebook from the `stable` version of the docs.  So the next thing to do is...

 - [x] Verify what version of PlasmaPy should get launched in binder


namurphy:
Closing this since we're going to do a release soon, which will make this obsolete.

PR #1171, opened 2021-06-21
Input ion velocity in thomson notebook

bryancfoo:
The example dicts used in the thomson notebook contain an "ion_vel" index, but this is not used when calling the spectral_density function at the end. Added a line with the relevant input parameter to correct this.


pheuer (approved):
Looks good to me! Thanks @bryancfoo for catching this omitted line.

PR #1173, opened 2021-06-23
Fix Volume Weighted Interpolator in CartesianGrid class (grids.py)

pheuer:
I realized the current volume weighted interpolator in grids.CartesianGrid was incorrect. This PR fixes that. 

Here's an example of the old interpolator: the linear interpolation between data points is backwards! 
image

And here it is with the corrected algorithm
image

I haven't yet found a good implementation of this algorithm to copy, so my code probably isn't the most efficient. The diagram below shows the idea in 2D

image

The particle position "p" is surrounded by a box of size dx * dy. The four surrounding grid points are similarly surrounded. The weight for the field from each grid point is then the area (volume in 3D) of the overlap between the particle box and the box for each grid point.


pheuer:
For some reason the readthedocs check is just never completing? I've tried restarting the checks.

rocco8773:
It is completing, but for some reason the checks are not updating or not getting an update from RTD.  But they are passing on RTD.  It might clear itself in due time, otherwise we'll have to investigate our repo connection between RTD and GitHub.

pheuer:
Got it - thanks.

rocco8773:
@pheuer I figured out the issue. It was related to the `master` -> `main` renaming.

rocco8773:
This PR did not touch these lines, but there's a case not covered by your try-except statement on lines 1129-1134 in `grids.py`.

image


If `del self._interp_quantities` raises `AttributeError`, then `del self._interp_units` will not be executed.

rocco8773 (changes requested):
The `CartesianGrid.volume_averaged_interpolator` could really benefit from a detailed docstring explaining how the interpolation works.  I know it's currently inhertiting the docstring from `AbstractGrid` and if you don't want to have to rewrite that portion, then you can use the [`plasmapy.utils.decorators.helpers.modify_docstring` decorator](https://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.helpers.modify_docstring.html#plasmapy.utils.decorators.helpers.modify_docstring) to append the original docstring with a robust "Notes" section.

rocco8773:
So I have a speed improvement for your `nan_mask` on line 1111 of `gris.py`

```python
# what you have
nan_mask = np.where(np.isnan(np.sum(arr, axis=1)), 0, 1)

# a faster solution
nan_mask = np.isnan(i).any(axis=1)  # this creates a boolean array you can use for indexing later on
```

image

Then, down on line 1200 you want to

```python
# replace this
weight *= nan_mask

# with this
weight[nan_mask] = 0
```

pheuer:
> If `del self._interp_quantities` raises `AttributeError`, then `del self._interp_units` will not be executed.

True: this shouldn't ever happen but I'll break it into two separate try statements

> Actually, I really don't understand the concept behind these `del`s. Both objects you're trying to delete are methods of the class you are working in and, thus, I don't think can be deleted. Wouldn't it make more sense to zero out the values or set them to `None`?

Those properties are both cached properties: the `del` keyword deletes the cached value so it has to re-execute. This is the desired behavior if a non-persistent interpolator is used AFTER a persistent interpolator is used on the same grid, since in that case we need to clear the persistent values from the previous usage.

pheuer:
> The `CartesianGrid.volume_averaged_interpolator` could really benefit from a detailed docstring explaining how the interpolation works. I know it's currently inhertiting the docstring from `AbstractGrid` and if you don't want to have to rewrite that portion, then you can use the [`plasmapy.utils.decorators.helpers.modify_docstring` decorator](https://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.helpers.modify_docstring.html#plasmapy.utils.decorators.helpers.modify_docstring) to append the original docstring with a robust "Notes" section.

That's an awesome decorator, and perfect for this application! Unfortunately it doesn't seem to work in the case of an inherited docstring, I get the following traceback:

```

  File "c:\users\pvheu\github\plasmapy\plasmapy\utils\decorators\helpers.py", line 66, in decorator
    doclines = inspect.cleandoc(wrapper.__doc__).splitlines()

  File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\inspect.py", line 619, in cleandoc
    lines = doc.expandtabs().split('\n')

AttributeError: 'NoneType' object has no attribute 'expandtabs'
```

The decorator works if I manually provide a docstring, but then that defeats the point.

rocco8773:
> Those properties are both cached properties: the `del` keyword deletes the cached value so it has to re-execute. This is the desired behavior if a non-persistent interpolator is used AFTER a persistent interpolator is used on the same grid, since in that case we need to clear the persistent values from the previous usage.

Ah, ok that make more sense.

> > The `CartesianGrid.volume_averaged_interpolator` could really benefit from a detailed docstring explaining how the interpolation works. I know it's currently inhertiting the docstring from `AbstractGrid` and if you don't want to have to rewrite that portion, then you can use the [`plasmapy.utils.decorators.helpers.modify_docstring` decorator](https://docs.plasmapy.org/en/latest/api/plasmapy.utils.decorators.helpers.modify_docstring.html#plasmapy.utils.decorators.helpers.modify_docstring) to append the original docstring with a robust "Notes" section.
> 
> That's an awesome decorator, and perfect for this application! Unfortunately it doesn't seem to work in the case of an inherited docstring, I get the following traceback:
> 
> ```
> 
>   File "c:\users\pvheu\github\plasmapy\plasmapy\utils\decorators\helpers.py", line 66, in decorator
>     doclines = inspect.cleandoc(wrapper.__doc__).splitlines()
> 
>   File "C:\Users\pvheu\Anaconda3\envs\plasmapy\lib\inspect.py", line 619, in cleandoc
>     lines = doc.expandtabs().split('\n')
> 
> AttributeError: 'NoneType' object has no attribute 'expandtabs'
> ```
> 
> The decorator works if I manually provide a docstring, but then that defeats the point.

You want to set it up like...

```python
@modify_docstring(prepend=AbstractGrid.volume_averaged_interpolator.__doc__)
def volume_averaged_interpolator(...):
    """
    Notes
    -----

    This is the detailed notes.
    """
```

plasmapy/plasma/grids.py
```diff
@@ -1105,16 +1108,16 @@ def volume_averaged_interpolator(
...
-        # position are NaN, and 0 otherwise.
-        nan_mask = np.where(np.isnan(np.sum(i, axis=1)), 0, 1)
+        nan_mask = np.isnan(i).any(axis=1)
         # Replace all NaNs temporarily with 0
         i = np.where(np.isnan(i), 0, i)
         i = i.astype(np.int32)  # Cast as integers
 
         # Load grid attributes (so this isn't repeated)
         ax0, ax1, ax2 = self.ax0.si.value, self.ax1.si.value, self.ax2.si.value
+        dx, dy, dz = self.dax0.si.value, self.dax1.si.value, self.dax2.si.value
```
rocco8773, line 1118: Unless you're trying to enforce/check some kind of units here, I'd stick with using `self.ds.coords["ax0"].values` since the adding and stripping of units incurs a non-negligible time penalty.

pheuer: Unfortunately the units of the axes aren't known and we need to make sure they match the provided position units, so I think this is necessary. 

I could use .to(u.m), given that this is Cartesian coordinates: I think that's faster.

rocco8773: Ok. If I remember correctly, that method is faster.

It might be faster to do something more like...

```python

if isinstance(pos, u.Qantity)
    pos_unit = pos.unit
    pos = pos.value
else:
    pos_unit = self.unit

....

if pos_unit == self.unit:
    # assuming you're enforcing units[0] == units[1] == units[2]
    ax0 = self.ds.coords["ax0"]
    ...
else:
    ax0 = self.ax0.to(u.m).value
    ...
```

It might even be worthwhile to have a unitless version of the "ax0" properties...

```python
@property
def ax0_unitless(self):
    if self.is_uniform:
        return self.ds.coords["ax0"].values
    else:
        raise ValueError(...)

@property
def ax0(self):
    return self.ax0_unitless * self.unit0
```

I only suggest this so you don't have to keep adding and stripping unis for calculations, and the functionality is wrapped into an interface so it will be easier to make modifications (if needed) without breaking a bunch of code.

plasmapy/plasma/grids.py
```diff
@@ -1065,9 +1070,28 @@ def _validate(self):
...
         self, pos: Union[np.ndarray, u.Quantity], *args, persistent=False
     ):
+        """
+        Notes
+        -----
+
+        This interpolator approximates the value of a quantity at a given
+        interpolation point using a weighted sum of the values at the eight grid
+        verticies that surround the point. The weighting factors are calculated by
+        defining a volume :math:`dx \times dy \times dz`
```
rocco8773, line 1084:
```suggestion
        This interpolator approximates the value of a quantity at a given
        interpolation point using a weighted sum of the values at the eight grid
        vertices that surround the point. The weighting factors are calculated by
        defining a volume :math:`dx \\times dy \\times dz`
```

You need to use double backslashes since the docstring is not a raw string `r"""`.

plasmapy/plasma/grids.py
pheuer: I was thinking about making private methods without units: I will try that. 

What do you think about making many of these axes cached properties? They always return the same thing unless someone manually messes with the coordinates in `self.ds` which is not a supported way of interacting with the object.

rocco8773: ... I'm reluctant to that since they are public attributes and thus we just create the potential for miss use and unintended behavior.  If they were private methods, then I wouldn't be as concerned.

rocco8773: I did run into an issue w/ the implementation of the first code block because `self.interpolate_indices(pos)` is called on line 1134 and it's needing units to return the correct value.

plasmapy/plasma/grids.py
```diff
@@ -1105,16 +1129,24 @@ def volume_averaged_interpolator(
 
         # Get the indices that are equal to nan (fill values), then set
         # their values to 0. They will be over-written after the interpolation
-
-        # Nan array is shape [n] and is 1 if none of the indices for a
-        # position are NaN, and 0 otherwise.
-        nan_mask = np.where(np.isnan(np.sum(i, axis=1)), 0, 1)
+        nan_mask = np.isnan(i).any(axis=1)
         # Replace all NaNs temporarily with 0
         i = np.where(np.isnan(i), 0, i)
```
rocco8773, line 1134:
```suggestion
        i[np.isnan(i)] = 0
```

Speed boost since you're setting element values instead of rebuilding and redefining an array.

plasmapy/plasma/grids.py
pheuer: Yeah, this is a little complicated because we need to reconcile units on the grid and in user-provided positions. 

I think this is non-trivial and should be a separate PR in the future if we want to get into it. I'm also not as concerned about the speed enhancements here, because the vast majority of time is spent in the scipy interpolator.

rocco8773:
Ok, so I think I understand how all this is working.  I have a few things...

1. As a common rule, you should use more explicit naming for your variables.  Using `i` for anything other than an index in a for loop makes the code very difficult to read and follow.  Because of the non-explicit naming it took me a very long time to decipher what all the variables were.  Using something like `nearest_neighbor_indices` instead of `i` will go along ways in improving readability.
2. In an effort to understand the routine, I did end up developing my own version along side yours, see [`vai2`](https://github.com/rocco8773/PlasmaPy/blob/9c7a0e40ac0c1ebc0241a76e00582a253d901af9/plasmapy/plasma/grids.py#L1250-L1416).  I did confirm this version passes your tests.  It elminates all the for loops and relys heavily on just `numpy` functionality.  As a result, the execution time is reduced by about 0.75x.  The biggest time suck of the whole routing is the call to `self.interpolate_indices`, so my version bypasses that.
image

    I don't want to just suggest you replace your version with mine, so take a look at what I did and pick and choose the elements you like.  Your version appears to be working as intended but if you add some of my elements I think you can increase performance and readability.
3. I do think the last bit of performance to squeeze out of this functionality will come from being able to retrieve the coordinates without units.  The process of adding and stripping units is surprisingly slow.

pheuer:
> 2. In an effort to understand the routine, I did end up developing my own version along side yours, see [`vai2`](https://github.com/rocco8773/PlasmaPy/blob/9c7a0e40ac0c1ebc0241a76e00582a253d901af9/plasmapy/plasma/grids.py#L1250-L1416).  I did confirm this version passes your tests.  It elminates all the for loops and relys heavily on just `numpy` functionality.  As a result, the execution time is reduced by about 0.75x.  

Awesome, thank you!! As far as I can see your version passes the tests, so I will accept all of it. I think this complicated `numpy` array math actually probably reduces readability slightly, but for this function I think the tradeoff for performance is well worth it.

> The biggest time suck of the whole routing is the call to `self.interpolate_indices`, so my version bypasses that.

This is frustrating: the Scipy interpolator functions are surprisingly slow. I didn't realize they were slower than just doing it in numpy, though. Unfortunately `self.interpolate_indices` is also used in the `nearest_neighbor` interpolator. In the case of uniform Cartesian grids, there is probably a speed improvement to be gained there too, then. I was trying to avoid writing separate interpolators for every subclass of Grid, but maybe that is unavoidable. 

> 3. I do think the last bit of performance to squeeze out of this functionality will come from being able to retrieve the coordinates without units.  The process of adding and stripping units is surprisingly slow.

I've added some private attributes to do this. The wrinkle is that the variables stored in the `ds` are not necessarily SI units, and converting them every time re-introduces the unit time penalty. I've solved this for now by calculating the unit conversion factors once (when the dataset is created), then using those for conversions on each call.

plasmapy/plasma/grids.py
```diff
@@ -513,6 +539,14 @@ def _load_grid(
         self.ds = xr.Dataset()
 
         self.ds.attrs["axis_units"] = [pts0.unit, pts1.unit, pts2.unit]
+
+        # Store the conversion factors for each axis to SI
+        self.si_factors = [
+            (1 * pts0.unit).si.value,
+            (1 * pts1.unit).si.value,
+            (1 * pts2.unit).si.value,
```
rocco8773, line 547:
```suggestion
            pts0.unit.si.scale,
            pts1.unit.si.scale,
            pts2.unit.si.scale,
```

Since we know converting to units and stripping units are painfully slow, we should always be looking to avoid/minimize these actions.  The current implementation does both of these slow processes, my suggestion does neither.

image

pheuer: Oh nice! I didn't know about this scale attribute.

plasmapy/plasma/grids.py
```diff
@@ -1098,96 +1153,131 @@ def volume_averaged_interpolator(
...
+            self._ax1_si,
+            self._ax2_si,
+        )
+
+        dx, dy, dz = (
+            self._dax0_si,
+            self._dax1_si,
+            self._dax2_si,
+        )
         n0, n1, n2 = self.shape
```
rocco8773, line 1185:
```suggestion
        ax0, ax1, ax2 = self._ax0_si, self._ax1_si, self._ax2_si
        dx, dy, dz = self._dax0_si, self._dax1_si, self._dax2_si
        n0, n1, n2 = self.shape
```

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,20 @@ def unit(self):
     # *************************************************************************
     # 1D axes and step sizes (valid only for uniform grids)
     # *************************************************************************
+    @property
+    def si_scale_factors(self):
+        """
+        Scale factors that convert the units of each axis to SI units.
+
+        """
```
rocco8773, line 365:
```suggestion
        """
        Scale factors that convert the units of each axis to SI units.
        """
```

rocco8773:
```suggestion
    def si_scale_factors(self) -> List[float]:
        """
        3-element list containing unitless scale factors for converting
        the corresponding axis from its stored units to SI.
        """
```

You'll also need to import `List` from `typing`.

pheuer:
I've split up the volume integrated interpolator tests. While doing so I realized that the 3D test is sensitive to the number of points in the `example_grid`. For example, it passes with `num=20` but fails with `num=25`. This merits some investigation/maybe a better test...

plasmapy/plasma/grids.py
rocco8773: Darn! I was too slow in removing the old comment and making this one.  I'll redo this comment.

rocco8773: nevermind

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,20 @@ def unit(self):
...
+        the corresponding axis from its stored units to SI.
+        """
+        return self._si_factors
+
+    @property
+    def _ax0_si(self):
+        """
+        The ax0 axis in SI units.
+        """
+        return self.ds.coords["ax0"].values * self._si_factors[0]
```
rocco8773, line 373:
```suggestion
        return self.ds.coords["ax0"].values * self.si_scale_factors[0]
```

plasmapy/plasma/grids.py
```diff
@@ -374,6 +390,13 @@ def ax0(self):
                 "The axis properties are only valid on " "uniformly spaced grids."
             )
 
+    @property
+    def _ax1_si(self):
+        """
+        The ax1 axis in SI units.
+        """
+        return self.ds.coords["ax1"].values * self._si_factors[1]
```
rocco8773, line 398:
```suggestion
        return self.ds.coords["ax1"].values * self.si_scale_factors[1]
```

plasmapy/plasma/grids.py
```diff
@@ -391,6 +414,13 @@ def ax1(self):
                 "The axis properties are only valid on " "uniformly spaced grids."
             )
 
+    @property
+    def _ax2_si(self):
+        """
+        The ax2 axis in SI units.
+        """
+        return self.ds.coords["ax2"].values * self._si_factors[2]
```
rocco8773, line 422:
```suggestion
        return self.ds.coords["ax2"].values * self.si_scale_factors[2]
```

rocco8773 (changes requested):
Once you've made the `property` it's better to use that property than the private attribute, unless there's a good reason to do so.  This approach make maintenance easier.

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,20 @@ def unit(self):
...
+        3-element list containing unitless scale factors for converting
+        the corresponding axis from its stored units to SI.
+        """
+        return self._si_factors
+
+    @property
+    def _ax0_si(self):
+        """
+        The ax0 axis in SI units.
+        """
```
rocco8773, line 372:
```suggestion
        """
        The ax0 axis without units, but scaled such that its values
        correspond to SI values.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -374,6 +390,13 @@ def ax0(self):
                 "The axis properties are only valid on " "uniformly spaced grids."
             )
 
+    @property
+    def _ax1_si(self):
+        """
+        The ax1 axis in SI units.
+        """
```
rocco8773, line 397:
```suggestion
        """
        The ax1 axis without units, but scaled such that its values
        correspond to SI values.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -391,6 +414,13 @@ def ax1(self):
                 "The axis properties are only valid on " "uniformly spaced grids."
             )
 
+    @property
+    def _ax2_si(self):
+        """
+        The ax2 axis in SI units.
+        """
```
rocco8773, line 421:
```suggestion
        """
        The ax2 axis without units, but scaled such that its values
        correspond to SI values.
        """
```

rocco8773 (changes requested):
We need to be careful here...These properties are returning an array that does NOT contain units, but there are other properties that do return arrays with units.  We have to choose our wording carefully so the user does not get confused to what is what, or at least mitigate that chance of confusion.

plasmapy/plasma/grids.py
```diff
@@ -408,6 +441,13 @@ def ax2(self):
                 "The axis properties are only valid on " "uniformly spaced grids."
             )
 
+    @property
+    def _dax0_si(self):
+        """
+        The grid spacing along the ax0 axis in SI units.
+        """
```
rocco8773, line 448:
```suggestion
        """
        Grid step size along axis :attr:`ax0` without units and scaled such
        that its values correspond to SI values.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -426,6 +466,13 @@ def dax0(self):
                 "uniformly spaced grids."
             )
 
+    @property
+    def _dax1_si(self):
+        """
+        The grid spacing along the ax1 axis in SI units.
+        """
```
rocco8773, line 473:
```suggestion
        """
        Grid step size along axis :attr:`ax1` without units and scaled such
        that its values correspond to SI values.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -444,6 +491,13 @@ def dax1(self):
                 "uniformly spaced grids."
             )
 
+    @property
+    def _dax2_si(self):
+        """
+        The grid spacing along the ax2 axis in SI units.
+        """
```
rocco8773, line 498:
```suggestion
        """
        Grid step size along axis :attr:`ax2` without units and scaled such
        that its values correspond to SI values.
        """
```

rocco8773 (changes requested):

rocco8773:
In order to generalize the `ax` and `dax` properties and reduce code duplication, why not take an approach like...

```python
def _get_ax(self, *, axis: int, si: bool = False):
    ax_name = f"ax{axis}"

    if not self.is_uniform:
        raise ValueError(
            "The axis properties are only valid on uniformly spaced grids."
        )

    vals = self.ds.coords[ax_name].values
    if si:
        return vals * self.si_scale_factors[axis]
    else:
        return vals * self.units[axis]

def _get_dax(self, *, axis: int, si: bool = False):
    ax = self._get_ax(axis=axis, si=si)
    return np.mean(np.gradient(ax))

@property
def _ax0_si(self):
    return self._get_ax(axis=0, si=True)

@property
def ax0(self):
    return self._get_ax(axis=0)

@property
def _dax0_si(self):
    return self._get_dax(axis=0, si=True)

@property
def _dax0(self):
    return self._get_dax(axis=0)
```

A nice benefit of this is that we only have to implement the `is_uniform` checking in one spot.  In your current implementation this check is missing in all the "`_si`" flavor properties.

plasmapy/plasma/grids.py
```diff
@@ -1065,23 +1130,41 @@ def _validate(self):
...
+        introduces a linear interpolation between grid vertices.
+
+        This implementation of this algorithm assumes that the grid is uniformly
+        spaced and Cartesian.
+        """
+
+        if isinstance(pos, u.Quantity):
+            pos = pos.to(u.m).value
+        elif self.unit != u.m:
+            pos *= self.unit.to(u.m).value
```
rocco8773, line 1159:
```suggestion
            pos *= self.unit.si.scale
```

pheuer:
> In order to generalize the `ax` and `dax` properties and reduce code duplication

Sure, we could do that (implemented)

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,45 @@ def unit(self):
...
+        """
+        3-element list containing unitless scale factors for converting
+        the corresponding axis from its stored units to SI.
+        """
+        return self._si_factors
+
+    def _get_ax(self, *, axis: int, si: bool = False):
+        """
+        Helper function for generating the axis properties
+        """
```
rocco8773, line 371:
```suggestion
        """
        Helper function for retrieving axis values.

        Parameters
        ----------
        axis: int
            Axis index for axis of interest (e.g. ``0`` for :attr:`ax0`).
        si: bool
            (Default: `False`) If `True` then convert the axis values to
            their SI equivalent.

        Returns
        -------
        ~numpy.ndarray or ~astropy.units.Quantity
            If ``si==True`` then return a unitless `~numpy.ndarray`.
            If ``si==False`` then return a `~astropy.units.Quantity`
            array with the original units.

        Raises
        ------
        ValueError
            If the grid is not uniform.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,45 @@ def unit(self):
...
+        vals = self.ds.coords[ax_name].values
+        if si:
+            return vals * self.si_scale_factors[axis]
+        else:
+            return vals * self.units[axis]
+
+    def _get_dax(self, *, axis: int, si: bool = False):
+        """
+        Helper function for generating the axis spacing properties
+        """
```
rocco8773, line 388:
```suggestion
        """
        Helper function for calculating grid spacing.

        See Also
        --------
        plasmapy.plasma.grids.AbstractGrid._get_ax
        """
```

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,45 @@ def unit(self):
...
+        ax = self._get_ax(axis=axis, si=si)
+        return np.mean(np.gradient(ax))
+
+    @property
+    def _ax0_si(self):
+        """
+        The ax0 axis without units, but scaled such that its values
+        correspond to SI values.
+        """
```
rocco8773, line 397:
```suggestion
        """
        The :attr:`ax0` axis without units, but scaled such that its values
        correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -355,6 +357,45 @@ def unit(self):
...
+        return self._get_ax(axis=0, si=True)
 
     @property
     def ax0(self):
```
rocco8773, line 401:
```python
    def ax0(self):
        """
        First axis of the grid.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -366,13 +407,15 @@ def ax0(self):
...
-        else:
-            raise ValueError(
-                "The axis properties are only valid on " "uniformly spaced grids."
-            )
+    @property
+    def _ax1_si(self):
+        """
+        The ax1 axis without units, but scaled such that its values
+        correspond to SI values.
+        """
```
rocco8773, line 417:
```suggestion
        """
        The :attr:`ax1` axis without units, but scaled such that its values
        correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -366,13 +407,15 @@ def ax0(self):
...
+        return self._get_ax(axis=1, si=True)
 
     @property
     def ax1(self):
```
rocco8773, line 421:
```python
    def ax1(self):
        """
        Second axis of the grid.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -384,12 +427,15 @@ def ax1(self):
...
-                "The axis properties are only valid on " "uniformly spaced grids."
-            )
+        return self._get_ax(axis=1)
+
+    @property
+    def _ax2_si(self):
+        """
+        The ax2 axis without units, but scaled such that its values
+        correspond to SI values.
+        """
```
rocco8773, line 437:
```suggestion
        """
        The :attr:`ax2` axis without units, but scaled such that its values
        correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -384,12 +427,15 @@ def ax1(self):
...
+        return self._get_ax(axis=2, si=True)
 
     @property
     def ax2(self):
```
rocco8773, line 441:
```python
    def ax2(self):
        """
        Third axis of the grid.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -401,12 +447,15 @@ def ax2(self):
...
-                "The axis properties are only valid on " "uniformly spaced grids."
-            )
+        return self._get_ax(axis=2)
+
+    @property
+    def _dax0_si(self):
+        """
+        Grid step size along axis :attr:`ax0` without units and scaled such
+        that its values correspond to SI values.
+        """
```
rocco8773, line 457:
```suggestion
        """
        Grid step size along axis :attr:`ax0` without units and scaled such
        that its values correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -401,12 +447,15 @@ def ax2(self):
...
+        return self._get_dax(axis=0, si=True)
 
     @property
     def dax0(self):
```
rocco8773, line 461:
```python
    def dax0(self):
        """
        Grid step size along axis :attr:`ax0`.

        Only valid for uniform grids.
        """```

File: plasmapy/plasma/grids.py [hunk 27]
```diff
@@ -418,13 +467,15 @@ def dax0(self):
...
-                "uniformly spaced grids."
-            )
+        return self._get_dax(axis=0)
+
+    @property
+    def _dax1_si(self):
+        """
+        Grid step size along axis :attr:`ax1` without units and scaled such
+        that its values correspond to SI values.
+        """
```
rocco8773, line 477:
```suggestion
        """
        Grid step size along axis :attr:`ax1` without units and scaled such
        that its values correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -418,13 +467,15 @@ def dax0(self):
...
+        return self._get_dax(axis=1, si=True)
 
     @property
     def dax1(self):
```
rocco8773, line 481:
```python
    def dax1(self):
        """
        Grid step size along axis :attr:`ax1`.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -436,13 +487,15 @@ def dax1(self):
...
-                "uniformly spaced grids."
-            )
+        return self._get_dax(axis=1)
+
+    @property
+    def _dax2_si(self):
+        """
+        Grid step size along axis :attr:`ax2` without units and scaled such
+        that its values correspond to SI values.
+        """
```
rocco8773, line 497:
```suggestion
        """
        Grid step size along axis :attr:`ax2` without units and scaled such
        that its values correspond to SI values.

        Only valid for uniform grids.
        """
```

plasmapy/plasma/grids.py
```diff
@@ -436,13 +487,15 @@ def dax1(self):
...
+        return self._get_dax(axis=2, si=True)
 
     @property
     def dax2(self):
```
rocco8773, line 501:
```python
    def dax2(self):
        """
        Grid step size along axis :attr:`ax2`.

        Only valid for uniform grids.
        """
```

rocco8773 (changes requested):
This is to make the docstrings more consistent.

plasmapy/plasma/grids.py
```diff
@@ -1065,23 +1141,41 @@ def _validate(self):
...
+        (where :math:`dx`, :math:`dy`, and :math:`dz` are the grid
+        spacings in each direction) around each grid vertex and around the
+        interpolation point. The contribution of each grid vertex is then
+        weighted by the fraction of the volume surrounding the interpolation
+        point that overlaps the volume surrounding that vertex. This effectively
+        introduces a linear interpolation between grid vertices.
+
+        This implementation of this algorithm assumes that the grid is uniformly
+        spaced and Cartesian.
+        """
```
rocco8773, line 1165:
```suggestion
        """
        Notes
        -----

        This interpolator approximates the value of a quantity for all
        the desired interpolation points in ``pos``.  The value at a given
        point is approximated by taking a weighted average of values of
        the surrounding grid points.  The number of grid points used in
        the average depends on where the interpolated point is: 1 if
        the point is in a corner of the grid, 2 if the point is on an
        edge of the grid, 4 if the point is on a face of the grid, and
        8 if the point is in the volume of the grid.

        It is assumed the grid is regularly spaced along all axes, but the
        spacing along each axis (:math:`dx`, :math:`dy`, and :math:`dz`)
        need not be equal.  This grid spacing defines the imaginary volume
        :math:`dV = dx \\times dy \\times dy` centered on each grid point.

        **How the algorithm works**

        1. The 8, 4, 2, or 1 grid point(s) surrounding the point of interest
           is determined.
        2. The weights for each grid point is calculated by taking the
           imaginary volume :math:`dV` centered on the point of interested
           and determining the fraction of that volume that intersects with
           the volume of the surround grid point volumes.
        3. Use the weights determined in (2) and the associated grid quantities
           to calculate the weighted average.
        """
```

I felt this docstring was missing a few key points, so here's my stab at expanding it.  I feel it's a step in the right direction, but it is definitely not perfect.

rocco8773 (changes requested):
At the end of `test_grids.py` remove the `if __name__ == "__main__"` statement.  This statement is only used if you need to run a file as script and an importable module.  I don't think we have any scenario like that in `plasmapy`.

plasmapy/plasma/grids.py
pheuer: I think of the algorithm as always averaging over eight points, where the extra points outside of the grid just continue the grid spacing and have a value of zero?

pheuer:
> At the end of `test_grids.py` remove the `if __name__ == "__main__"` statement. This statement is only used if you need to run a file as script and an importable module. I don't think we have any scenario like that in `plasmapy`.

Sure, I can remove that. I use that for debugging - if pytest has a way of running the tests and plotting matplotlib output, I haven't found it yet?

pheuer:
@rocco8773 I have a question about these lines in your version of the interpolator: 

```python
norms = np.sum(bounding_cell_weights, axis=1)
mask_norm_zero = norms == 0.0
bounding_cell_weights[~mask_norm_zero] = (
    bounding_cell_weights[~mask_norm_zero, ...] / norms[~mask_norm_zero, None]
)
```

I think the correct behavior should be to always normalize to `dx * dy * dz`. As the weights become zero as the particle nears the edge, this corresponds to still including the part of the particle volume that is now off the grid. The quantities should then gradually fall off to zero as the particle crosses the grid boundary.

rocco8773:
> 
> 
> > At the end of `test_grids.py` remove the `if __name__ == "__main__"` statement. This statement is only used if you need to run a file as script and an importable module. I don't think we have any scenario like that in `plasmapy`.
> 
> Sure, I can remove that. I use that for debugging - if pytest has a way of running the tests and plotting matplotlib output, I haven't found it yet?

It's bad practice to leave unused, commented out code like this in distributed work.  I was actually going to have you place the code into the test function's docstring as an example, or the like.

Another approach...You can create a directory in `plasmapy`'s root directly called `notebooks_for_proto`.  This directory is already added to the `.gitignore` so it won't be under version control.  You can place your plotting script here without it affecting everything else.

rocco8773:
> 
> 
> @rocco8773 I have a question about these lines in your version of the interpolator:
> 
> ```python
> norms = np.sum(bounding_cell_weights, axis=1)
> mask_norm_zero = norms == 0.0
> bounding_cell_weights[~mask_norm_zero] = (
>     bounding_cell_weights[~mask_norm_zero, ...] / norms[~mask_norm_zero, None]
> )
> ```
> 
> I think the correct behavior should be to always normalize to `dx * dy * dz`. As the weights become zero as the particle nears the edge, this corresponds to still including the part of the particle volume that is now off the grid. The quantities should then gradually fall off to zero as the particle crosses the grid boundary. Eg. these lines should be replaced by
> 
> ```python
> bounding_cell_weights *= 1/(dx * dy * dz)
> ```

It all depends on how you want to treat the boundary.  Your approach would be fine if you forced all grid quantities to go to zero by the time it reaches the boundary, but would produce erroneous behavior if you had something like a DC magnetic field everywhere.  Your approach would essentially create a non-existent step down the magnetic field at the boundary.

In my approach I take the view that I don't know what the quantities are outside the grid, so I just use the nearest points on the grid boundary to estimate the quantities until the particle moves off the grid.  So, when the particle is on the boundary I'm only using 4, 2, or 1 grid points to estimate quantity opposed to the typical 8 grid points.

While my approach will still have errors associated with the estimated quantities at the boundary, I think it will better handle at least the two scenarios I mentioned above.

This could be an issue for future PRs where you can control boundary behaviors: Impose a hard boundary, a mirrored boundary, a soft boundary, etc.

plasmapy/plasma/grids.py
rocco8773: As I mentioned in https://github.com/PlasmaPy/PlasmaPy/pull/1173#issuecomment-892015118, this would be problematic for something like a DC magnetic field.  I think my suggested approach is a little safer, but I'm open to arguments against it.  Simulations are not my forté so I'm sure there's a better approach.

Also as mentioned in that comment, this could be a future PR where the user can defined how to treat the boundary.

pheuer:
>Another approach...You can create a directory in plasmapy's root directly called notebooks_for_proto. This directory is >already added to the .gitignore so it won't be under version control. You can place your plotting script here without it >affecting everything else.

This would be a good solution, except I want those scripts to be accessible to anyone else trying to debug the code in the future... 

>It all depends on how you want to treat the boundary. Your approach would be fine if you forced all grid quantities to >go to zero by the time it reaches the boundary, but would produce erroneous behavior if you had something like a DC >magnetic field everywhere. Your approach would essentially create a non-existent step down the magnetic field at the >boundary.

In the case of proton radiography, we do require that the grids go to zero at the edges (otherwise a warning will pop up). In this case the way I described is better, since the quantities will smoothly drop off at the edges rather than abruptly, which is the desired behavior. 

I'll change this for now - we can add the other functionality in the future if another application starts using `grids.py`

rocco8773:
> > It all depends on how you want to treat the boundary. Your approach would be fine if you forced all grid quantities to go to zero by the time it reaches the boundary, but would produce erroneous behavior if you had something like a DC magnetic field everywhere. Your approach would essentially create a non-existent step down the magnetic field at the >boundary.
> 
> In the case of proton radiography, we do require that the grids go to zero at the edges (otherwise a warning will pop up). In this case the way I described is better, since the quantities will smoothly drop off at the edges rather than abruptly, which is the desired behavior.

NO!  That is a restriction of `SyntheticProtonRadiography` and is currently only enforced by that class.  This volume averaging as no affect on the [code throwing the warning you mentioned](https://github.com/PlasmaPy/PlasmaPy/blob/f77c538147a2bc79a54f81836f188a7255b6536c/plasmapy/diagnostics/proton_radiography.py#L214-L237).  It is completely up to the user to generate a grid quantity that smoothly transitions to zero at the boundary.  If the user successfully does that then your approach and my approach will yield essentially the same results.  On the other-hand, your approach will produce erroneous values for other (non-radiograph) uses.

This zero bound should NOT be a condition of the grid functionality since the grid classes ought to be generalized for multiple uses.

pheuer:
I'm not suggesting that the validation in `SyntheticProtonRadiograph` covers this function, only that since that's the only use case currently planned it makes sense to use that class to guide this design decision. I agree that in most cases the difference for this function is trivial, so maybe it's not worth arguing about.

**To be clear, there are no "erroneous" values because there are no universally "correct" values outside the grid (without additional assumptions).** A common assumption is that everything is zero outside the domain, in which case I believe the implementation I've proposed is the most natural interpretation of "volume averaging" over that boundary (grid value slowly transitions to zero). 

If I'm understanding it right, your algorithm also assumes that values outside the domain are zero, except it also extends the grid 1 grid cell in every direction (equal to a constant), then drops sharply to zero (sketched below). I don't think this is "volume averaging" over anything, and I think it's much less natural than the other version. Let me know if I'm misunderstanding your algorithm. 

image

rocco8773:
> I'm not suggesting that the validation in `SyntheticProtonRadiograph` covers this function, only that since that's the only use case currently planned it makes sense to use that class to guide this design decision. I agree that in most cases the difference for this function is trivial, so maybe it's not worth arguing about.

But radiography is not going to be the only use case for grids.  The purpose of putting grids into `plasmapy.plasma` is that it would be used to help define plasma objects with physical domains.  In that case a DC magnetic field everywhere will be a common occurrence.  I know this is not implemented yet, but it's not a far off evolution.

> **To be clear, there are no "erroneous" values because there are no universally "correct" values outside the grid (without additional assumptions).** A common assumption is that everything is zero outside the domain, in which case I believe the implementation I've proposed is the most natural interpretation of "volume averaging" over that boundary (grid value slowly transitions to zero).

I disagree this is the most natural interpretation.  There are many ways to treat boundary conditions.  Zeroing the out of domain points makes a lot of sense for radiography, but that won't be the logical choice for all use cases of grids.  For cases were it's not the logical choice, then it will lead to erroneous values.

> If I'm understanding it right, your algorithm also assumes that values outside the domain are zero, except it also extends the grid 1 grid cell in every direction (equal to a constant), then drops sharply to zero (sketched below). I don't think this is "volume averaging" over anything, and I think it's much less natural than the other version. Let me know if I'm misunderstanding your algorithm.

My approach assumes nothing about the values outside the domain and does not try to extend the grid.  It goes, "I don't know the values beyond this point, so I'm just going to use the values I do know."  It's true then that a point on the face of the grid is equivalent to an area weighted average, a point on an edge of the grid is equivalent to a line weighted average, and a point on the corner is just the corner value.

In fact, I think it would be better for `volume_average_interpolation` to return a `np.nan` for positions (`pos` values that are) off the grid (`pos < ax[0] - 0.5 * dx or pos > ax[-1] + 0.5 * dx`) than the current behavior of returning zero.  In this case, `SyntheticProtonRadiograph` would go "oh, this is a `np.nan` I'm just going to zero it."

Here's more how I see how the two approaches work.  Case 2 would be the radiograph case and, yes, the original approach is more accurate but since values are close to zero it would be negligible compare to the "Erik" approach.  Case 1 would be a quantity that does not approach zero at the boundary.  Depending on how high above zero the last point is, the current approach could lead to highly erroneous values.  Now, the "Erik" approach is not good in all cases either but would be the better approximation in this scenario.  This is why I think some boundary keyword for `volume_weighted_average` would be a good thing.  For now, I still think the "Erik" (I hate using my name like this...it sounds so egotistical...arg) is the safer of the two approaches.

image

rocco8773:
> 
> 
> As a compromise solution, what if we included a `fill_value = 0` keyword for the function, then set the values outside the grid to this value (in the units of the quantity)? Then the interpolation would linearly transition from the last grid value to this fill value, and that would be more transparent in the documentation.

I'd probably go with something like `boundary_condition="zeroed"` or `"mirrored"` to switch between the two conditions.  Obviously zeroed would be your approach and mirrored would be my approach.  There's probably a better name than mirrored though.

rocco8773:
Another option that would replace both our current views...if the point of interest is out side the volume bounded by the grid points (i.e. it's not surrounded by 8 grid points), then it's considered off the grid.  In this case there will always be 8 points (8 volumes) to do the weighted average with.

pheuer:
I agree: we should return `np.nan` if the position is entirely off the grid - that would be consistent with how the scipy interpolators are structured. 

The tricky situation is positions that are still on the grid but within `dx/2` of the edge, such that some of the volume crosses out. Maybe we should consider these `np.nan` as well, effectively shrinking the grid dimension by `dx`? That would be the simplest to document, so maybe it's the right choice?

rocco8773:
It at least feels like the right choice until we have proper mechanisms for treating the boundary interpolation as per the user's choice.

rocco8773 (changes requested):
I just noticed in [`AbstractGrid._load_grid` online 605](https://github.com/PlasmaPy/PlasmaPy/blob/84c6f325837c16a142b5fdbe2c6fc122317f527d/plasmapy/plasma/grids.py#L532) you used `add_quantity` instead of `add_quantities`.  I'm surprised there's no failed tests from this and I'm guessing that's because tests never reached this line.

https://github.com/PlasmaPy/PlasmaPy/blob/84c6f325837c16a142b5fdbe2c6fc122317f527d/plasmapy/plasma/grids.py#L532

rocco8773:
> The tricky situation is positions that are still on the grid but within `dx/2` of the edge, such that some of the volume crosses out. Maybe we should consider these `np.nan` as well, effectively shrinking the grid dimension by `dx`? That would be the simplest to document, so maybe it's the right choice?

Just occurred to me we'd run into some issues for grids that are effectively a surface (i.e. `shape = (Nx, Ny, 1)`) or a line (i.e. `shape=(Nx, 1, 1)`.  However, I think it would be relatively easy to catch these scenarios and handle them appropriately.

Additionally, if you take a look at [`numpy.interp`](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) the default behavior for points outside the array is essentially my approach.  However, they include keywords `left` and `right` so the user can specify a return value for points outside the array.  An approach similar to this would probably be good for `volume_averaged_interpolation`.

For now, I still think just returning `np.nan` for points outside the grid is the safest approach.

rocco8773:
I think the currents tests a good start, but need to be a little more exhaustive.  Besides `test_volume_average_interpolator_at several_positions`, there's no exhaustive testing of the averaging algorithm.  I'd suggest setting up a grid of `shape=(2, 2, 2)` with know values at each grid point.  Then pick several points in that volume where you can hand calculate the weighted average and compare it to the algorithm.

In the end, I'd suggest setting up a test structure that looks a bit like the following code. This is the test structure I commonly use for my code because it allows me to better compartmentalize aspects of the code I need to test.  It also lends itself well to parametrization.  I also find it much more readable if/when I have to come back to it.  Now, this is just a sample, so the docstrings and parametrized test need to be expanded on.

```python
class TestCartesianGridVolumeAveragedInterpolation:

    def make_grid(self, *, shape):
        if isinstance(shape, int):
            shape = (shape, ) * 3

        # Create grid
        _grid = grids.CartesianGrid(-1 * u.cm, 1 * u.cm, num=shape)

        # Add some data to the grid
        ones = np.ones(shape, dtype=np.float64)
        with warnings.catch_warnings():
            warnings.simplefilter("ignore")
            _grid.add_quantities(
                mass=np.arange(0, ones.size, dtype=np.float64).reshape(shape) * u.g,
                ones=ones * u.m,
            )
        return _grid

    @pytest.fixture(scope="class")
    def grid_222(self):
        return self.make_grid(shape=(2, 2, 2))

    @pytest.mark.parametrize(
        "pos",
        [
            np.array([0.0, 0.0, 0.0]) * u.cm,
            np.array([0.5, 0.0, 0.0]) * u.cm,
            np.array([0.0, 0.5, 0.0]) * u.cm,
            np.array([0.0, 0.0, 0.5]) * u.cm,
            np.array(
                [[-0.5, 0.0, 0.0], [0.0, -0.5, 0.0], [0.0, 0.0, -0.5]]
            ) * u.cm,
        ],
    )
    def test_vai_ones(self, pos, grid_222):
        """
        Test interpolation of points in bounds of a (2, 2, 2) shaped
        grid of ones.
        """
        vals = grid_222.volume_averaged_interpolator(pos, "ones")
        if pos.ndim == 1:
            expected = np.array([1.]) * u.m
        else:
            expected = np.ones(pos.shape[0], dtype=np.float64) * u.m

        assert np.allclose(vals, expected)
        assert vals.unit == expected.unit

    @pytest.mark.parametrize(
        "pos, expected",
        [(np.array([0.0, 0.0, 0.0]) * u.cm, np.array([3.5]) * u.g), ],
    )
    def test_vai(self, pos, expected, grid_222):
        """
        Test interpolation of points in bounds of a (2, 2, 2) shaped
        grid of known values.
        """
        vals = grid_222.volume_averaged_interpolator(pos, "mass")
        assert np.allclose(vals, expected)
        assert vals.unit == expected.unit

    def test_vai_off_grid(self):
        """Test that out of bound points are returned as `numpy.nan`."""
        # or out of bounds
        ...

    def test_vai_persistence(self):
        """Test functionality of the `persistence` keyword."""
        ...

    def test_vai_1D(self):
        """Test interpolation when the grid is effectively linear."""
        # test interpolation for grids of shape (Nx, 1, 1), or the like
        ...

    def test_vai_2D(self):
        """Test interpolation when the grid is effectively planar."""
        # test interpolation for grids of shape (Nx, Ny, 1) or the like
        ...

    def test_vai_raises(self):
        """Test scenarios that should raise Exceptions."""
        # e.g. missing keys
        ...

    def test_vai_warnings(self):
        """Test scenarios that should issue warnings."""
        ...
```

I'm a big proponent of organizing tests into classes if it does not make sense to split them into their dedicated file.  It helps me frame my mindset when I'm reading/writing tests, and it helps de-clutter the file.  However this is highly subjective, so it's not an approach you have to take.

Now, the tests where you compare the nearest neighbor interpolation to the volume average interpolation is nice, but I don't think it's necessary.  It doesn't actually test that the functionality is working, just that it's better or worse than some other functionality.  I'll leave it up to you if you want to trim those tests or leave them.

pheuer:
What is the timeline on v0.7.0? I don't have time to rework all the tests right now, but I want to make sure the original bug that started this PR is fixed prior to that release.

rocco8773:
This PR is basically a requirement for the release.  But I'm guessing the release won't happen for at least 2 more weeks.

If you like, we can work in tandem to get the tests complete.

pheuer:
@rocco8773 I have been unsuccessful at returning `np.nan` for positions outside the grid: it is possible I am misunderstanding how your re-write of the interpolator works. You will have to do this.

plasmapy/plasma/grids.py
```diff
@@ -1098,96 +1192,130 @@ def volume_averaged_interpolator(
...
+        # Create a mask for positions that are off the grid. The values at
+        # these points will be set to zero later.
+        mask_particle_off = (
+            (pos[:, 0] < ax0.min() + 0.5 * dx)
+            | (pos[:, 0] > ax0.max() - 0.5 * dx)
+            | (pos[:, 1] < ax1.min() + 0.5 * dy)
+            | (pos[:, 1] > ax1.max() - 0.5 * dy)
+            | (pos[:, 2] < ax2.min() + 0.5 * dz)
+            | (pos[:, 2] > ax2.max() - 0.5 * dz)
+        )
```
rocco8773, line 1232:
```suggestion
        mask_particle_off = (
            (pos[:, 0] < ax0.min())
            | (pos[:, 0] > ax0.max())
            | (pos[:, 1] < ax1.min())
            | (pos[:, 1] > ax1.max())
            | (pos[:, 2] < ax2.min())
            | (pos[:, 2] > ax2.max())
        )
```

plasmapy/plasma/grids.py
```diff
@@ -1098,96 +1192,130 @@ def volume_averaged_interpolator(
...
+        bounding_cell_indices[mask_cell_off, :] = 0
+
+        # Calculate the volume of the overlap between the point volume
+        # and the volume of each of the surrounding vertices
+        lx = dx - np.abs(pos[:, None, 0] - ax0[bounding_cell_indices[..., 0]])
+        ly = dy - np.abs(pos[:, None, 1] - ax1[bounding_cell_indices[..., 1]])
+        lz = dz - np.abs(pos[:, None, 2] - ax2[bounding_cell_indices[..., 2]])
+        bounding_cell_weights = lx * ly * lz
+
+        # Set the weight for any off-grid vertices to zero
```
rocco8773, line 1296:
```suggestion
        # Set the weight for any off-grid vertices (cell or particle) to zero
```

plasmapy/plasma/grids.py
```diff
@@ -1098,96 +1192,130 @@ def volume_averaged_interpolator(
...
+        vals = self._interp_quantities[
+            bounding_cell_indices[..., 0],
+            bounding_cell_indices[..., 1],
+            bounding_cell_indices[..., 2],
+            :,
+        ]
+        # Construct a weighted average of the interpolated quantities
+        weighted_ave = np.sum(bounding_cell_weights[..., None] * vals, axis=1)
+
+        weighted_ave[mask_particle_off, :] = np.nan
```
rocco8773, line 1312:
```suggestion
        weighted_ave = np.sum(bounding_cell_weights[..., None] * vals, axis=1)
        weighted_ave[mask_particle_off, :] = np.nan
```

plasmapy/plasma/grids.py
```diff
@@ -1098,96 +1192,130 @@ def volume_averaged_interpolator(
...
+        # and the volume of each of the surrounding vertices
+        lx = dx - np.abs(pos[:, None, 0] - ax0[bounding_cell_indices[..., 0]])
+        ly = dy - np.abs(pos[:, None, 1] - ax1[bounding_cell_indices[..., 1]])
+        lz = dz - np.abs(pos[:, None, 2] - ax2[bounding_cell_indices[..., 2]])
+        bounding_cell_weights = lx * ly * lz
+
+        # Set the weight for any off-grid vertices to zero
+        bounding_cell_weights[mask_cell_off] = 0.0
+        bounding_cell_weights[mask_particle_off, ...] = 0.0
+        bounding_cell_weights *= 1 / (dx * dy * dz)
```
rocco8773, line 1299:
```suggestion
        norms = np.sum(bounding_cell_weights, axis=1)
        mask_norm_zero = norms == 0.0
        bounding_cell_weights[~mask_norm_zero] = (
            bounding_cell_weights[~mask_norm_zero, ...] / norms[~mask_norm_zero, None]
        )
```

I still feel this is the safer algorithm, but at this point they're essentially doing the same thing.   What actually might be good here is to issue a warning or exception if 

```python
not np.allclose(np.norm(bound_cell_weights[~mask_norm_zero]) / (dx * dy * dz), 1.0)
```

If this condition is not satisfied then the weight was not calculated correctly.

rocco8773 (changes requested):
These changes will properly add the `np.nan` values.

plasmapy/plasma/grids.py
pheuer: I think the current version of this is correct: as soon as a particle crosses `ax0.max() - dx/2` part of the particle volume is off the grid and therefore undefined.

pheuer:
> I just noticed in [`AbstractGrid._load_grid` online 605](https://github.com/PlasmaPy/PlasmaPy/blob/84c6f325837c16a142b5fdbe2c6fc122317f527d/plasmapy/plasma/grids.py#L532) you used `add_quantity` instead of `add_quantities`. I'm surprised there's no failed tests from this and I'm guessing that's because tests never reached this line.

This was never hit because it isn't actually supported to add quantities this way anymore: I've just removed this line. Quantities should be added after creating the grid, using `add_quantities()`. 

I changed the way the proton radiography code was handling NaN values and all the tests are now passing.

rocco8773 (approved):

pheuer:
@rocco8773 Awesome, thanks for cleaning this up and getting it merged!

PR #1187, opened 2021-07-06
change behaviour of gyroradius when arguments are nan

### LINKED ISSUE #1022: Computation of gyroradius from Plasmapy, in case of missing temperature should return NaNs instead of ValueError
## Feature request

At present if I do the following, a ```ValueError``` is raised:
```python
>>> from plasmapy.formulary.parameters import gyroradius
>>> gyroradius(5 * u.nT, particle='p+', T_i=numpy.nan * u.K)
ValueError: Must give Vperp or T_i, but not both, as arguments to gyroradius
```
which is the wrong kind of error being raised. Error arises because temperature (```T_i```) data is NaN and not because the user is providing both ```T_i``` and ```Vperp```., as the text accompanying the error claims.

Another problem with this is that if we are trying to calculate gyroradius from a spacecraft data and at at any specific time, if both the magnetic field and (```B```) and ```T_i``` or ```Vperp``` data are missing, it raises the ```ValueError``` and stops the code from executing.

If instead of doing that, the code has the option to just set such values of gyroradius to ```numpy.nan``` (as is already happening when ```B``` is ```numpy.nan```), it increases its functionality and doesn't force the user to feed clean data.

I have also attached a screenshot of the error!
gyroradius_error

alfkoehn:
This PR closes #1022 . The general behaviour of `gyroradius` is changed such that if one of the arguments (`T_i`, `Vperp`) is nan, and the other is not supplied, no error is raised, but nan is returned.

      docstrings.

### DIFF
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -1198,7 +1198,7 @@ def gyroradius(
 
     # check 1: ensure either Vperp or T_i invalid, keeping in mind that
     # the underlying values of the astropy quantity may be numpy arrays
-    if np.any(np.logical_not(np.logical_xor(isfinite_Vperp, isfinite_Ti))):
+    if np.any(np.logical_and(isfinite_Vperp, isfinite_Ti)):
         raise ValueError(
             "Must give Vperp or T_i, but not both, as arguments to gyroradius"
         )
--- a/changelog/1187.feature.rst
+++ b/changelog/1187.feature.rst
@@ -1,0 +1,3 @@
+The behaviour of the function `~plasmapy.formulary.parameters.gyroradius` has
+been changed: if NaN-values are provided for T_i or Vperp, no longer is a 
+slightly misleading error raised, but simply NaN is returned.
--- a/changelog/1187.feature.rst
+++ b/changelog/1187.feature.rst
@@ -1,3 +1,3 @@
 The behaviour of the function `~plasmapy.formulary.parameters.gyroradius` has
-been changed: if NaN-values are provided for T_i or Vperp, no longer is a 
+been changed: if NaN-values are provided for T_i or Vperp, no longer is a
 slightly misleading error raised, but simply NaN is returned.
--- a/plasmapy/formulary/tests/test_parameters.py
+++ b/plasmapy/formulary/tests/test_parameters.py
@@ -802,6 +802,11 @@ def test_gyroradius():
 
     assert gyroradius(B, "e-", Vperp=25 * u.m / u.s).unit.is_equivalent(u.m)
 
+    # test for possiblity to allow nan for input values
+    assert np.isnan(gyroradius(np.nan*u.T, particle="e-", T_i=T_i))
+    assert np.isnan(gyroradius(1*u.T, particle="e-", T_i=np.nan*u.K))
+    assert np.isnan(gyroradius(1*u.T, particle="e-", Vperp=np.nan*u.m/u.s))
+
     Vperp = 1e6 * u.m / u.s
     Bmag = 1 * u.T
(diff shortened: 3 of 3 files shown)


plasmapy/formulary/parameters.py
```diff
@@ -1198,7 +1198,7 @@ def gyroradius(
 
     # check 1: ensure either Vperp or T_i invalid, keeping in mind that
     # the underlying values of the astropy quantity may be numpy arrays
-    if np.any(np.logical_not(np.logical_xor(isfinite_Vperp, isfinite_Ti))):
+    if np.any(np.logical_and(isfinite_Vperp, isfinite_Ti)):
```
StanczakDominik, line 1201: Do you think we could test this to ensure it works as expected? The tests for gyroradius are located around https://github.com/PlasmaPy/PlasmaPy/blob/main/plasmapy/formulary/tests/test_parameters.py#L798 here :)

StanczakDominik (changes requested):

plasmapy/formulary/parameters.py
alfkoehn: Ahh, didn't knew about that proper test section, sorry - had only made tests on my own. Will do the proper testing later.

alfkoehn: Does _testing_ here simply means to run `test_gyroradius()`, or also to modify `test_gyroradius()` and add some new tests? (The former one should be sufficient in my opinion but I might overlook something)

StanczakDominik: The tests are ran automatically anyway (see the Checks tab!), so that's already being done :) What I meant instead was to add (or ensure there are already! haven't checked 😅) a few checks to `test_gyroradius`:

1. gyroradius(T_i = nan) == nan, no exception;
2. gyroradius(Vperp = nan) == nan, no exception;
3. gyroradius(T_i = nan, Vperp = nan) raises exception (there should be a `pytest.raises` clause around that general area we could adapt :) )

(`np.isnan` would be helpful for the comparison, since anything `== np.nan` is `False`, even if that anything is `np.nan`!)

alfkoehn: Those are exactly the checks I have done manually :)

Now I will try to understand the syntax in the `test_gyroradius` function and include them there.

alfkoehn: Note that this might not be (well, I'm pretty sure it is not) the most elegant or "pythonic" way of testing it, but all the three new cases are included via `assert` tests. Not sure though if this qualifies the official test criteria....?

alfkoehn:
@StanczakDominik I don't understand why the `pre-commit.ci` fails, it says

`black....................................................................Failed`
`- hook id: black`
`- files were modified by this hook`
`reformatted plasmapy/formulary/tests/test_parameters.py`

I have no idea what was actually reformatted, so I don't know what to change to let this check pass... Any hints are greatly appreciated

StanczakDominik:
Iirc the magic incantation for this is...

pre-commit.ci autofix

alfkoehn:
No idea about magic incantation, sorry :D

plasmapy/formulary/parameters.py
StanczakDominik: This is literally what I'm here for :)

They're great, exactly what I was hoping for! Since the test suite isn't exploding, we can probably merge it now :)

StanczakDominik:
Hah, sorry! This part is a bit new, and we need to document it better.

StanczakDominik (approved):
Looks great, let's get it in! Thanks! :)

alfkoehn:
Awesome :)

qudsiramiz:
Yaayyy! Thank you!

PR #1156, opened 2021-05-28
Update the documentation guide within the development guide

namurphy:
The goal of this PR is to expand the documentation within our development guide on how to write documentation in order to make it more complete and accessible to new contributors.  This is in preparation for [Plasma Hack Week](https://hack.plasmapy.org).

Here's an incomplete list of things I'm planning.  I'll probably edit this list as we go.

 - [x] Introduce ReST, including examples
 - [x] Briefly introduce Markdown
 - [x] Introduce Sphinx
 - [x] Discuss the numpydoc standard
 - [x] Add an example docstring with links, references, math, etc.  
 - [x] Discuss ReST substitutions (should merge #1147 first, and then incorporate it into here)
 - [x] Mention `plasmapy_sphinx` and create an issue on adding more information on it to the documentation guide
 - [x] Update the description of how to build documentation

PR #1205, opened 2021-07-23
Update release guide including about binder requirements

namurphy:
This PR includes some updates to the release guide prior to 0.7.0, including some minor formatting and reST changes.

The main purpose is to follow up on #1159 by changing `binder/requirements.txt` such that binder installs the version of PlasmaPy that was just released. Some points:
 - I haven't tested this step yet, so we might need to update it after the release.
 - There might be a better place to put this step in the release guide.
 - This step could be automated, but even if so, we would want to keep a step where the binder notebooks are tested on both ``stable`` and the ``<major>.<minor>.x`` branch.

### DIFF
--- a/docs/development/release_guide.rst
+++ b/docs/development/release_guide.rst
@@ -31,14 +31,14 @@ Release
 
 * Reserve a digital object identifier (DOI) on Zenodo_ for version ``0.6.0``.
 
-* Update ``docs/about/citation.rst`` with the DOI for version ``0.6.0``.
+* Update :file:`docs/about/citation.rst` with the DOI for version ``0.6.0``.
 
-* Update version metadata in ``codemeta.json``.  In particular, update the
+* Update version metadata in :file:`codemeta.json`.  In particular, update the
   ``"identifier"`` tag with the DOI for version ``0.6.0``.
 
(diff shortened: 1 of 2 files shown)


StanczakDominik:
> This step could be automated, but even if so, we would want to keep a step where the binder notebooks are tested on both stable and the <major>.<minor>.x branch.

*groans in test complexity*

docs/development/release_guide.rst
```diff
@@ -29,16 +29,16 @@ Pre-release
 Release
 -------
 
-* Reserve a digital object identifier (DOI) on `Zenodo`_ for version ``0.6.0``.
+* Reserve a digital object identifier (DOI) on Zenodo_ for version ``0.6.0``.
```
StanczakDominik, line 32: And this works? Without the `s?

I've been catching up on Star Trek TOS recently and let me just say, fascinating.

StanczakDominik (approved):
LGTM!

docs/development/release_guide.rst
namurphy: Yep!  One of the many things I learned but never wanted to while working on #1156.

PR #1206, opened 2021-07-23
Update tox configuration for building docs

namurphy:
This makes/proposes a few changes:

 - I added a new test, `build_docs_nitpicky`, that issues warnings for missing links.  
 - I added the `--keep-going` flag when we have the `-W` flag.  This way the test doc build will keep going after a warning, which will let us fix multiple warnings at the same time instead of having to wait for the doc build run multiple times.
 - I changed all doc builds to use `--parallel=auto`.  This is experimental, and does not ~~make the change in GitHub Actions~~ alter our GitHub Actions setup, so as far as I can tell, the doc builds there will still only be done on one core rather than in parallel.  (edited as a clarification)

PR #1207, opened 2021-07-23
Fix broken reST links in old changelogs

namurphy:
This PR is a spin-off of #1206 in order to fix the warnings associated with broken reST links. I'm limiting the scope of this PR to fixing links, without making any substantive changes to the changelog otherwise.  Thus far I've been able to get us from 1466 to 908 warnings.

My main motivation for this pull request is that we **really** ought to enable the nitpicky (`-n`) option in `sphinx-build` in the documentation tests for each pull request.  That way, we would be able to catch broken links before they make their way into the repository.  Before we can enable that option...however...we have to fix the broken links that are currently in the documentation, which were largely concentrated in old changelogs as references to stuff that doesn't exist anymore.

I removed the `|PhysicalType|` substitution since it's in Astropy 4.3+ only, and because it's not used anywhere in PlasmaPy (yet).

I also avoided using substitutions for code within PlasmaPy because the locations of objects sometimes changes with time, and the changelog should reflect the state of the repository at that time.


StanczakDominik:
> My main motivation for this pull request is that we really ought to enable the nitpicky (-n) option in sphinx-build in the documentation tests for each pull request. That way, we would be able to catch broken links before they make their way into the repository. Before we can enable that option...however...we have to fix the broken links that are currently in the documentation, which were largely concentrated in old changelogs as references to stuff that doesn't exist anymore.

Good call! It's probably best to make a one liner PR with that change and just periodically merge main into it.

StanczakDominik (approved):

PR #1158, opened 2021-05-29
Add reST substitutions and use them in narrative docs

namurphy:
This is a spin-off from #1147 to expand some xrefs in the docs for particles...in particular `functional.rst`.  I'll come back to this after #1147 is merged.

PR #1211, opened 2021-07-25
Fix broken reST links and use substitutions in `plasmapy.particles`

namurphy:
Another PR to fix broken reST links using the nitpicky option in `sphinx-build`, this time in `plasmapy.particles`.  This follows up on #1206 and #1207.


docs/common_links.rst
```diff
@@ -5,40 +5,41 @@
...
+.. |ParticleError| replace:: :py:exception:`~plasmapy.particles.exceptions.ParticleError`
+.. |ParticleWarning| replace:: :py:exception:`~plasmapy.particles.exceptions.ParticleWarning`
+.. |UnexpectedParticleError| replace:: :py:exception:`~plasmapy.particles.exceptions.UnexpectedParticleError`
 
 .. ------------------
 .. NumPy replacements
 .. ------------------
 
-.. |ndarray| replace:: :class:`numpy.ndarray`
+.. |ndarray| replace:: :py:class:`numpy.ndarray`
```
rocco8773, line 33: Note, you do not have to include `:py:` here because it is the default domain for sphinx.  It is assumed unless a different domain is given.

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-primary_domain

rocco8773 (changes requested):
As we've debated at nauseam, I really do not think we should be using substitutions outside of the narrative documentation!  Using something like `|Quantity|` and `|ndarray|` and `|Particle|` in these places, while won't be an issue for the experienced user, it will be a real problem for someone just taking up the package.  I will not be able to approve this PR as is for this reason/

StanczakDominik:
With my own ipython centric development workflow, I tend to do most of my docstring lookups from the shell first and switching to the browser only when necessary. If we were designing for browser only, then sure, there's no difference here.

With the way these docstrings worked before, if I see something like `~plasmapy.particles.Particle`, I instantly know where to go for further reading. Seeing `|Particle|` instead gives me no jumping off point.

So I'd tend to agree with Erik on this one!

namurphy:
Pre-emptively superseded by #1158.

PR #1225, opened 2021-08-02
Changed behavour of Particle.__eq__ to return false when used with other types

Quettle:
This PR changes the behavior oh the `__eq__` method of |Particle| in instances where an error would be raised if the compared object isn't a |Particle| or a valid str. This make the use of both |Particle| and |CustomParticle| more convenient when mixed in a list or as dict keys.
The only case where an error is raised is when the comparison is maid with an other |Particle| with se same symbol but different attributes.

      docstrings.

### DIFF
--- a/changelog/1225.breaking.rst
+++ b/changelog/1225.breaking.rst
@@ -1,1 +1,1 @@
-Changed behaviour of `~plasmapy.particles.Particle.__eq__` to return False instead of an error when used to compare with an other type or a string that doesn't represents a particle.
+Changed behaviour of `~plasmapy.particles.particle_class.Particle.__eq__` to return `False` instead of an error when used to compare with another type or a string that does not represent a particle.
--- a/changelog/1225.breaking.rst
+++ b/changelog/1225.breaking.rst
@@ -1,0 +1,1 @@
+Changed behaviour of `~plasmapy.particles.Particle.__eq__` to return False instead of an error when used to compare with an other type or a string that doesn't represents a particle.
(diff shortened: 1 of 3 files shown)


changelog/1225.breaking.rst
```diff
@@ -0,0 +1 @@
+Changed behaviour of `~plasmapy.particles.Particle.__eq__` to return False instead of an error when used to compare with an other type or a string that doesn't represents a particle.
```
namurphy, line 1:
```suggestion
Changed behaviour of `~plasmapy.particles.particle_class.Particle.__eq__` to return `False` instead of an error when used to compare with another type or a string that does not represent a particle.
```

plasmapy/particles/particle_class.py
```diff
@@ -604,13 +604,8 @@ def __eq__(self, other) -> bool:
...
         |Particle| instance or a `str` representing the same particle,
-        and return `False` if ``other`` is a different |Particle| or a
-        `str` representing a different particle.
-
-        If ``other`` is not a `str` or |Particle| instance, then this
-        method will raise a `TypeError`.  If ``other.symbol`` equals
-        ``self.symbol`` but the attributes differ, then this method
-        will raise a `~plasmapy.particles.exceptions.ParticleError`.
+        and return `False` if ``other`` is a different |Particle|, a
+        `str` representing a different particle or an other type.
```
namurphy, line 608:
```suggestion
        `str` representing a different particle or another type.
```

namurphy (approved):
I have two suggested wording changes, and I think we can get this change in quickly.  The test failures we've been having with xarray-dev and numpy-dev are unrelated so they're fine to ignore.  Thank you for making this change so quickly!

namurphy (approved):

PR #1223, opened 2021-08-02
Add function that returns a `ParticleList` of ionic levels of an atom

namurphy:
This pull request adds `plasmapy.particles.particle_collections.ion_list` as a way to create a `ParticleList` that contains the `Particle` instances for different ionic levels of an element or isotope.  For example...
```Python
>>> from plasmapy.particles import ion_list
>>> ion_list("H-1")
ParticleList(['H-1 0+', 'p+'])
```
I've had a need for this functionality come up fairly frequently in non-equilibrium ionization stuff, so it seems quite worthwhile to add since I'm sure others will have this need too.

PR #1226, opened 2021-08-04
Have Giles verify PR number of changelog entries

namurphy:
On #1223 I accidentally had the wrong pull request number for the changelog files, and noticed that the Giles towncrier check for the changelog entry still passed.  I looked up the documentation for [baldrick](https://baldrick.readthedocs.io/en/latest/plugins.html#towncrier-changelog-checker) and noticed that we were missing `verify_pr_number = true`, which I just added.

### DIFF
--- a/changelog/999999.trivial.rst
+++ b/changelog/999999.trivial.rst
@@ -1,0 +1,1 @@
+Changelog entry with the wrong pull request number.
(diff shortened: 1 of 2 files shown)


namurphy:
Huh.  I added the `verify_pr_number` option, but when I added a changelog entry with the wrong PR number, it still passed.  I wonder if it's necessary for this option to be in the `pyproject.toml` in the main repo in order for it to work, which means that we'd need to merge this in order to be able to test it.

PR #1233, opened 2021-08-12
Remove G function

StanczakDominik:
This should shut the numba-dev and xarray-dev tests up.


StanczakDominik:
Since I'm not ready with a numerically stable implementation of this function, I don't think this function is currently ready for PlasmaPy proper. I'll just keep the implementation in the side project.

changelog/1233.bugfix.rst
```diff
@@ -0,0 +1 @@
+Remove nmba-scipy from dependencies.
```
rocco8773, line 1:
```suggestion
Remove `numba-scipy` from dependencies.
```

I'd call this breaking instead of bugfix since you're deleting functionality from the package.

plasmapy/formulary/mathematics.py
```diff
@@ -94,73 +91,6 @@ def Fermi_integral(
         raise TypeError(f"Improper type {type(x)} given for argument x.")
 
 
-@vectorize(
-    [
-        float64(float64),
-    ]
-)
```
rocco8773, line 101: I was digging into this `_UFunNoLoopError`, which is causing all the test failures including the `hypothesis` tests, and I'm almost 100% certain it's caused by `numba.vectorize` and it's incompatibility with the current development branch of `numpy`.  From what I can tell, `numba.vectorize` is creating a `ufunc` signature that looks like `float64 -> None` while `numpy` requires `float64 -> float64` and, thus, `numpy` is complaining.

So, I'll leave the solution up to you, but you can do either...

1. what you're doing here and delteing `Chandrasekhar_G`
2. remove `numpy-dev` tests and set dependencies to `numpy < 1.22`

Personally, I don't like deleting functionality that "isn't broken" so I'd go with placing the dependency restriction, unless we really want what is in `numpy==1.22`.

changelog/1233.bugfix.rst
namurphy: I expect that this won't get intersphinxed so I'd suggest ``` ``numba-scipy`` ``` instead.  (It took all of ten back ticks to write that code snippet.)

rocco8773: That, or keep it as is so if `numba-scipy` is added to the intersphinx mappings then it will start working.  ... Doesn't matter in the long run.

namurphy (approved):
Looks good to me!  The only change would be to fix the typo in `nmba-scipy` that @rocco8773 mentioned.  After that it looks ready to merge.

PR #1238, opened 2021-08-19
Use full namespaces in docstrings for formulary aliases

namurphy:
Some quick changes to make sure that the reST links show up correctly in the docs for aliases.


namurphy:
This one should be quick enough of a review to be able to get it into 0.7.0 without any problems.  It's almost entirely just fixing a bunch of reST links so that they work correctly in the docs.

rocco8773 (approved):
LGTN! Thanks for the quick patch on this!

Issue #1243, opened 2021-08-21
Create a documentation-wide bibliography page


namurphy:
One possibility for PlasmaPy's online documentation would be to create a single page that would act as the bibliography for the documentation and package.  Here's a first attempt at a tradeoff analysis:

**Advantages**
* References would have a consistent reference code throughout the documentation.
* Anyone adding a reference would be able to find examples nearby rather than having to go to the doc guide or use their own style.
* Newly added references would be more likely to follow the pattern and style of existing references.
* References would be easier to maintain.
* Putting references in one place would make it more practical to include them in Zenodo records, which in turn would connect PlasmaPy records to other relevant records in portals like NASA ADS.

**Disadvantages**
* References would not show up in the docstring when using `help`
  * A naming convention that includes the author and year would reduce this disadvantage, i.e.  [[ZagerEvans2525]](https://www.youtube.com/watch?v=izQB2-Kmiic)
  * Someone would have to go online anyway to access the resource.
* Would be less convenient to find references that are only needed once.

namurphy:
I attempted having a `bibliography.rst` page in the docs in #1163, but decided to keep that PR separate from this issue.  It was really straightforward to get it to work, but I decided that adding a project-wide bibliography page should happen at the same time as the corresponding updates to the doc guide.  I don't want the doc guide to become out-of-date after just a few weeks!

StanczakDominik:
I'll leave the one liner that greps for the references I came up with during the telecon here, as it seems relevant:

```bash
rg -A 6 References
```

`rg` being https://github.com/BurntSushi/ripgrep; `-A 6` tells it to print out 6 lines after the `References` line.

namurphy:
Some of the relevant documentation is from the reStructuredText standard:
 - [Citation References](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#citation-references)
 - [Citations](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#citations)

namurphy:
Here's also [a reference from the Sphinx docs on citations](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?highlight=citations#citations).

namurphy:
[...]

PR #1245, opened 2021-08-24
Consolidate references in a single bibliography page

namurphy:
Here's a pull request that would create a bibliography page.  The way it's currently implemented does not require any additional Sphinx extensions (though an alternative would be to use [sphinxcontrib-bibtex](https://sphinxcontrib-bibtex.readthedocs.io) which has its own set of advantages & disadvantages).  A trade-off analysis is given in #1243.

Here's what would need to be done before this would be able to be merged, if we decide to proceed with this approach:

 - [ ] Move citations from other docstrings to the bibliography page
 - [ ] Move citations from the narrative documentation to the bibliography page
 - [ ] Update documentation guide to describe how to add references
 - [ ] Link to relevant reST & Sphinx documentation from documentation guide


namurphy:
Tentatively putting this on hold in favor of #1263.

namurphy:
Superseded by #1263.

PR #1249, opened 2021-08-26
added property based tests using hypothesis

### LINKED ISSUE #1242: Refactor plasma dispersion function tests to use hypothesis
Right now there are a few [property-based tests](https://hypothesis.works/articles/what-is-hypothesis/) of the plasma dispersion function and its derivative that have been implemented manually.  These are mostly properties related to symmetry, I believe.  It would be helpful to refactor these using [hypothesis](https://hypothesis.readthedocs.io), which is a Python package for property-based testing.

It's low priority since the implementation plasma dispersion function is fairly stable at this point, but it would be an opportunity to learn how to use hypothesis for someone who is unfamiliar with it.  There is also a chance that more thorough property-based testing might uncover some bugs that the current tests do not catch. Moreover, it would provide some plasma physics examples in the code that we could point to in the testing guide.  This would be a good first contribution for someone who has intermediate knowledge of Python.

14tstinson:
Closes #1242

      docstrings.


StanczakDominik:
The test failure disappeared after a subsequent re-run. I suspect it was something broken upstream. Oh well, let's merge it! Thanks once again, @14tstinson !

PR #1178, opened 2021-06-29
Start a documentation page for workflow documentation

namurphy:
This pull request is to create the outline of a page in our code development guide that describes the development workflow that we use when contributing to PlasmaPy

- [ ] I have added a changelog entry for this pull request.

- [ ] If adding new functionality, I have added tests and
      docstrings.

- [ ] I have fixed any newly failing tests.

PR #1255, opened 2021-08-28
Add a page on the contribution workflow in the contributor guide

### LINKED ISSUE #760: Describe the development workflow in docs
Right now our development guide does not quite have an equivalent to Astropy's page on [how to make a code contribution](https://docs.astropy.org/en/stable/development/workflow/development_workflow.html).  We should add one to our own development guide.  

We can adapt Astropy's guide pretty freely, since it's under their BSD 2.5 +/- 0.5 clause license which we already have in our `licenses` directory.  Astropy's guide goes into a bit more detail than would be necessary for a quickstart guide, so we might want to make it more concise so that it's easier for newcomers to get the most important information.  Extra information that is important but not strictly necessary for getting started should go into the later parts of the development guide.

namurphy:
During a live tutorial during Plasma Hack Week, I started an outline for a page in the contributor guide on the workflow for contributing to PlasmaPy (#1178).  That only included section titles.  Given insufficient time before the 0.7 release, this page got hidden temporarily in #1217.  This pull request is intended to put content on the workflow page.  Some of this could be adapted from [Astropy's dev workflow page](https://docs.astropy.org/en/latest/development/workflow/development_workflow.html) and other resources.  We'll also need to list some resources that would be helpful for new contributors to look at while getting started.

Closes #760.


namurphy:
Superseded by #1656, largely because I forgot that I already had this pull request open.

PR #1258, opened 2021-09-08
Test development version of Astropy with Python 3.8

namurphy:
We started getting a test failure with Python 3.7 for astropy-dev because they bumped their minimum version of Python to 3.8.  This PR updates our CI tests for astropy-dev to be consistent with the requirements that Astropy will have for their next release.  

If we keep the test at Python 3.8 rather than Python 3.9, we'll get a helpful test failure when Astropy bumps their minimum version again.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -28,10 +28,10 @@ jobs:
             python: 3.7
             toxenv: py37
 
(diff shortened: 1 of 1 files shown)


StanczakDominik (approved):
good stuff!

PR #1264, opened 2021-09-17
Describe conventions for putting formulae in docstrings

namurphy:
Following up on #935 and #1156, this PR adds a description of where formulae and extended mathematical stuff should be put in docstrings.  I'll likely make a few other small changes here too.

### DIFF
--- a/docs/development/doc_guide.rst
+++ b/docs/development/doc_guide.rst
@@ -750,6 +750,9 @@ Docstring guidelines
   be ≲ 4 sentences long. Any additional information should included in
   the "Notes" section.
 
+* Put any necessary highly technical information in the "Notes" section
+  of a docstring.
+
 * The short summary should start on the line immediately following the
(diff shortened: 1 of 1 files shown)


pheuer:
LGTM: My one observation is that, as a new-ish python developer, I don't know what it means for `:term:` to be a "role". That itself may need some defining.

namurphy:
@pheuer — might you have a chance to look at this again?  Thanks!

pheuer (approved):
Looks good! 

General question: if `|role|` links back to the Sphinx glossary, what happens if we re-define `role` in the plasmapy glossary?

namurphy:
> General question: if `|role|` links back to the Sphinx glossary, what happens if we re-define `role` in the plasmapy glossary?

Good point!  I updated this to say that the intersphinx terms are used if they haven't already been defined in PlasmaPy's glossary.  And thanks!

Issue #1266, opened 2021-09-17
Set up GitHub Actions to test hyperlinks in docs


namurphy:
If we're in `docs`, we can run `make linkcheck` to make sure that our links are working.  It will also check for redirects (minus those defined in `linkcheck_allowed_redirects` in `docs/conf.py`) and anchors (minus those defined in `linkcheck_anchors_ignore` in `docs/conf.py`).  This seems well-suited for a cron job that could be done weekly or even monthly. An alternative would be to have it be done following a successful doc build, though that seems like it would be much more frequent than is necessary.

PR #972, opened 2021-01-02
Added diamagnetic drift 

anthony-vo:
Addresses issue #966. User inputs the pressure gradient, background field, density, and charge. I included descriptions and examples in the docstring. Also tests (similar to ExB and force drift) were made, with an addition of the case in Fig. 3.4 of Chen. I assumed the density scale n'/n were 1 / m and verified the magnitude of the drift.

Issue #966, opened 2020-12-18
Diamagnetic Drift


rocco8773:
Add to `plasmapy.formulary.drifts` functionality for the diamagnetic drift.

![image](https://user-images.githubusercontent.com/29869348/102567260-fc2c5280-4095-11eb-9191-8041cb571cf4.png)

See eq. 3-65 in Chen, Francis F. *Introduction to Plasma Physics and Controlled Fusion.* 2nd ed. Springer, https://doi.org/10.1007/978-1-4757-5595-4.

StanczakDominik:
Closed by #972 :)

PR #1267, opened 2021-09-18
Fix hyperlinks and define redirects & anchors to avoid checking

namurphy:
To get ready for a time when we can have `make linkcheck` be part of our test suite (#1266), this PR fixes hyperlinks, defines allowable redirects (like for `doi.org`, to `latest` or `stable` or `v3.9.3 versions of docs, and login pages for GitHub and Google), and specifies anchors to avoid checking (like for specific lines in a page showing source code or for our Matrix room).

This should be a fairly quick review, so if it looks good to you, feel free to merge it so it can go in 0.7.0.


docs/whatsnew/0.6.0.rst
```diff
@@ -91,7 +91,7 @@ Features
 - For `plasmapy.plasma.grids` functionality, improve interpolation performance on
   non-uniform grids. (`#963 <https://github.com/plasmapy/plasmapy/pull/963>`__)
 - Added the `~plasmapy.formulary.drifts.diamagnetic_drift` function to
-  `~plasmapy.formulary.drifts`. (`#966 <https://github.com/plasmapy/plasmapy/pull/966>`__)
+  `~plasmapy.formulary.drifts`. (`#972 <https://github.com/plasmapy/plasmapy/pull/972>`__)
```
namurphy, line 94: It turns out that #966 was the issue being addressed, and #972 was the pull request that addressed it.

docs/conf.py
```diff
@@ -154,12 +151,25 @@
...
 todo_include_todos = False
 
 default_role = "obj"
 
+# Customizations for make linkcheck using regular expressions
+linkcheck_allowed_redirects = {
+    r"https://.+\.(org|io|com)": r".+(org|io|com)/en/.+",  # to different doc branches
+    r"https://doi\.org/.+": r"https://.+",  # DOI links are more persistent
+    r"https://.+": r".+(google|github).+[lL]ogin.+",  # some links require logins
+}
```
namurphy, line 164: Since I didn't know this beforehand, `.+` is effectively a wildcard marker, `(...|...)` is either/or, and `[ ]` means matching any one of these characters.

docs/conf.py
```diff
@@ -154,12 +151,25 @@
...
 
-
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 
 default_role = "obj"
 
+# Customizations for make linkcheck using regular expressions
+linkcheck_allowed_redirects = {
+    r"https://.+\.(org|io|com)": r".+(org|io|com)/en/.+",  # to different doc branches
```
StanczakDominik, line 161: Wouldn't this allow redirects to any website on the top level `.org`, `.io` or `.com` domains, rather than just different doc branches?

StanczakDominik (approved):
PDD label much? :D

One quick question, otherwise LGTM!

docs/conf.py
namurphy: Yeah, I had been wondering how to handle this. I couldn't find a way in the linkcheck configuration options to specify that the domain name be the same between the link and the redirect without giving every domain name.  My thinking has been that with the `/en/` in the second part, chances are the redirect is going to be for a docs page or otherwise somehow related to internationalization.  I'll make this a bit more specific...we'll catch more things links that need to be updated, with the caveat that we'll need to add some special cases in there.

namurphy:
> PDD label much? :D

Me?  No, never!  😶

PR #1268, opened 2021-09-19
Adopt a code highlighting style with higher contrast

namurphy:
According to the results posted in https://github.com/pygments/pygments/issues/1718, `"xcode"` is the only [pygments](https://pygments.org/) style that meets the [WCAG AA contrast standard](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) for web accessibility, apart from `"bw"` for black & white.  This PR changes our code formatting style to meet this accessibility standard.  

This should fix a bunch of the contrast errors that we get when [checking our docs against the WAVE accessibility tool](https://wave.webaim.org/report#/https://docs.plasmapy.org/en/latest/), though there are other style elements that are giving contrast errors too.

### DIFF
--- a/changelog/1268.doc.rst
+++ b/changelog/1268.doc.rst
@@ -1,2 +1,3 @@
-Adopted the ``"xcode"`` code highlighting style for `pygments `__ to
+Adopted the ``"xcode"`` code highlighting style for
+`pygments <https://pygments.org/>`__ to
 increase color contrast and improve web accessibility.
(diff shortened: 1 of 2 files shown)


namurphy:
Here's the [output for using an accessibility tool on one of our pages](https://wave.webaim.org/report#/https://plasmapy--1268.org.readthedocs.build/en/1268/particles/ionization_states.html#the-ionization-state-of-a-single-element).  As hoped, there are no longer any contrast warnings for the code blocks.  There are still a few contrast issues with the color of the links and the red in-line code snippets., plus some other things here & there.

StanczakDominik (approved):
LGTM!

Issue #1270, opened 2021-09-19
Make `plasmapy_sphinx` compatible with type aliases in parameter descriptions


namurphy:
Using `sphinx.ext.napoleon` as a Sphinx extension should allow us to use type aliases.  If we add the following lines to `docs/conf.py`:

```Python
napoleon_type_aliases = {"keyword-only": ":term:`keyword-only`")
napoleon_use_param = True
```

...and then have a docstring section like:
```Python
"""
Parameters
----------
arg : particle-like
    Description
"""
```
It should render as something like:

* **arg** ([particle-like](https://docs.plasmapy.org/en/latest/glossary.html#term-particle-like))  Description

This functionality by itself will be really helpful for a lot of our functions that allow different types of inputs.  However, there has also been work in Astropy to put physical types in parameter descriptions (like in many of the arguments in [FlatLambdaCDM](https://docs.astropy.org/en/stable/api/astropy.cosmology.FlatLambdaCDM.html#astropy.cosmology.FlatLambdaCDM)), which would be incredibly helpful for the documentation for `plasmapy.formulary`.

However, when `"plasmapy_sphinx"` ∈ `extensions` in `docs/conf.py`, the alias for keyword-only does not get applied.  However, when `"plasmapy_sphinx"` ∉ `extensions`, the link for keyword-only does show up correctly.  I have almost no understanding of why this is happening, but for some reason the type aliases are not (yet) compatible with `plasmapy_sphinx`.

PR #1271, opened 2021-09-19
Enable type aliases in documentation

namurphy:
This is an attempt following #1270 to get type aliases to work in our documentation. At the very least, this PR is intended to set up the infrastructure for the type aliases, even if they don't get displayed yet.

To check if it's working, there should be a link for [particle-like](https://docs.plasmapy.org/en/latest/glossary.html#term-particle-like) in the docstring for `Particle`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -154,6 +154,14 @@ exclude_patterns = [
     "common_links.rst",
 ]
 
+napoleon_google_docstring = False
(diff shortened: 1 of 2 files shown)


namurphy:
Closing this for the time being since I wasn't able to get this to work and probably won't be able to focus on it for a while.

PR #1273, opened 2021-09-21
Create GitHub Action for tests that run on a schedule

### LINKED ISSUE #1162: Test documentation build against sphinx-dev as cron job
We currently run our test suite against numpy-dev, xarray-dev, and astropy-dev.  I was thinking that it would be helpful to test our documentation build against sphinx-dev.  This could help us find and address the occasional breaking changes in Sphinx ahead of time.  This seems a good candidate for a weekly test and wouldn't be necessary for most pull requests.

namurphy:
This is my attempt to set up cron jobs for running with numpy-dev, xarray-dev, and astropy-dev.  This also closes #1162, since I'm adding a test of sphinx-dev too.

This might be a PR that would need to get merged before we figure out if it works or not.  I'm attempting to use the `workflow_dispatch` option, which means that we should be able to request that any or all of these jobs be run.

I'm not really comfortable with the GitHub Actions schema yet...so wish me luck!

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
       - v0.*.x
     tags:
       - "v*"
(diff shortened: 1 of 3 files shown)


StanczakDominik (approved):
LGTM, I'd be happy to try it out!

namurphy:
> LGTM, I'd be happy to try it out!

Thanks!  We might need a follow up pull request in case there are any problems with the configuration, but it would be good to give it a try.

Issue #1230, opened 2021-08-09
Bullet point lists not displaying correctly on several documentation pages


namurphy:
The behavior is essentially that for code like:

```rst

* First bullet point
* Second bullet point
```

The rendered docs show:

> First bullet point
> 
> Second bullet point

A couple of the affected sections are:
 * [Documentation guide](https://docs.plasmapy.org/en/latest/development/doc_guide.html#docstring-guidelines)
 * [PlasmaPy Contributors](https://docs.plasmapy.org/en/latest/about/credits.html#plasmapy-contributors)

PR #1107, opened 2021-04-12
Set docutils < 0.17 requirement

rocco8773:
The recent release (2021-04-03) of `docutils` `v0.17.0` is causing issues with rendering on RTD, https://github.com/readthedocs/sphinx_rtd_theme/issues/1115.  We are specifically suffering from improper rendering of the sidebar, like https://github.com/rdswift/picard-docs/issues/96.  This PR sets the requirement `docutils < 0.17`.

PR #1275, opened 2021-09-22
Specify minimum version of sphinx_rtd_theme to allow most recent release of docutils

namurphy:
Earlier this year we had some difficulties with docutils 0.17 and higher, which appeared to be incompatibilities with sphinx_rtd_theme.  I checked the changelog for sphinx_rtd_theme, and the recently released version 1.0.0 is apparently now compatible with more recent versions of docutils.  This PR updates the requirements for both docutils and sphinx_rtd_theme to test that this fix actually does work.


namurphy:
Fortunately, this seems to be working!  I'm not sure about which minimum version of docutils should be allowed, so I'll need to:

 - [ ] Determine what the minimum version of docutils should be

StanczakDominik:
> There is a minimum version of docutils required by Sphinx, so I don't think we need to specify one ourselves.

I would agree if we were only depending on functionality provided by Sphinx itself; but `plasmapy-sphinx` makes me less sure of that!

namurphy:
The two problems with docutils ≥ 0.17 that I knew of were...
 - Odd padding of the sidebar #docutils (#1107)
 - Bullet points not showing up in lists (#1230)
 
Both of those seem to no longer be problems in the docs preview with sphinx_rtd_theme ≥ 1.0.0 (looking at doc previews on Firefox from Ubuntu).  @rocco8773 — is there any reason why we'd need to set a minimum version for docutils?

changelog/1275.doc.rst
```diff
@@ -0,0 +1,2 @@
+Updated the requirements for the documentation build to be docutils ≥ 0.17.1
+and sphinx_rtd_them ≥ 1.0.0.
```
rocco8773, line 2:
```suggestion
Updated the requirements for the documentation build to be no restrictions on `docutils`
and `sphinx_rtd_them >= 1.0.0`.  `docutils == 0.17` is not compatible with
`sphinx_rtd_theme < 1.0` (see :pr:`1107`) and `docutils == 0.17` is not compatible with
`sphinx_rtd_theme < 1.0` (see :issue:`1230`).
```

namurphy:
- [ ] Check that `:issue:` and `:pr:` work in changelog entry

rocco8773 (approved):

PR #1278, opened 2021-09-28
Fix inconsistency in xarray cron tests and update README badges

namurphy:
A quick follow-up on #1273.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -9,6 +9,7 @@ on:
     tags:
       - "v*"
   pull_request:
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -9,6 +9,7 @@ on:
     tags:
       - "v*"
   pull_request:
+  workflow_dispatch:
```
namurphy, line 12: I just learned about `workflow_dispatch` in #1273.  It lets us trigger a workflow manually.

StanczakDominik (approved):

PR #1280, opened 2021-09-28
Add security analysis linter

namurphy:
This PR adds [dlint](https://github.com/dlint-py/dlint) as one of our static analysis tools run in the linters environment for tox.  

It came up with one "stop using `eval`" error when running locally, so we should get that here too if the tests are running.

An alternative would be to use [bandit](https://bandit.readthedocs.io/en/latest/), but for the time being either one would be helpful.


StanczakDominik (approved):
Hokay then!

PR #1282, opened 2021-09-29
Apply changes from static analysis tool that looks for likely bugs and design problems

namurphy:
~~This PR adds [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear) to the `linters` environment for tox, and makes some changes to get this check to pass.  We'd probably want to go through the [list of warnings](https://github.com/PyCQA/flake8-bugbear#list-of-warnings) in case there are some that we don't want to apply.~~

~~There are a couple of minor issues in `thomson.py` and `proton_radiography.py` that I'm ignoring for now to avoid conflicts between PRs.~~

This PR applies some of the changes that were recommended by [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear).  I decided not to implement this extension since there are a number of changes that I didn't implement yet, and probably won't get back to for a while yet.

I *think* that some of the errors from this extension are not ones that can be automatically fixed by pre-commit, which is another reason why I don't want to enable this by default (at least not yet). I'd need to check on this again later, though.  In the meantime, I'll probably use this extension locally.


plasmapy/diagnostics/thomson.py
```diff
@@ -279,7 +279,7 @@ def spectral_density(
     # Calculate the susceptibilities
 
     chiE = np.zeros([efract.size, w.size], dtype=np.complex128)
-    for i, fract in enumerate(efract):
+    for i in len(efract):
```
StanczakDominik, line 282:
```suggestion
    for i in range(len(efract)):
```

StanczakDominik: Going to commit it to see whether the tests pass...

StanczakDominik (approved):
All in all looks like a helpful linter. It's obviously still screaming at us for initializing numpy arrays as default arguments (could be replaced with `argument=None`, then `if argument is None: argument = np.array(...)`) - but I like it, overall.

PR #1283, opened 2021-09-29
Use flake8 checker for absolute imports and remove unused imports

namurphy:
- Used [`unimport`](https://unimport.hakancelik.dev/) to remove unused import statements.  
   - Modified a docstring example
 - Added [flake8-absolute-import](https://github.com/bskinn/flake8-absolute-import) to the tox environment for linters
 - Changed imports from relative to absolute

Something I don't understand yet is if these are compatible with pre-commit because they're done through flake8, or if we need to do something else to make it work with pre-commit.

### DIFF
--- a/changelog/1283.trivial.1.rst
+++ b/changelog/1283.trivial.1.rst
@@ -1,0 +1,2 @@
+Added `flake8-absolute-import <https://github.com/bskinn/flake8-absolute-import>`_ to the
+``linters`` tox_ environment.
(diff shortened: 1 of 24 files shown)


StanczakDominik:
> Something I don't understand yet is if these are compatible with pre-commit because they're done through flake8, or if we need to do something else to make it work with pre-commit.

AFAIK there are hooks for pre-commit that run flake8, but unlike `black` or `isort`, they don't autofix stuff, so I dislike keeping them there.

StanczakDominik (approved):

PR #1284, opened 2021-09-29
Add and apply pre-commit hooks for `.ini`, `.toml`, and `.yaml` files

namurphy:
This PR adds some pre-commit autoformatters for `.ini`, `.toml`, and `.yaml` files, and applies changes from those autoformatters ∀ files ∈PlasmaPy.

There don't seem to be many customization options, but autoformatters for all of these file types were available together so I'm guessing that they should all produce consistent styles.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,18 +1,18 @@
-name: 'Close stale issues and PR'
+name: Close stale issues and PR
 on:
   schedule:
-    - cron: '30 1 * * *'
+  - cron: 30 1 * * *
 
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v4
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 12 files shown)


.github/workflows/stale.yml
```diff
@@ -1,18 +1,18 @@
-name: 'Close stale issues and PR'
+name: Close stale issues and PR
 on:
   schedule:
-    - cron: '30 1 * * *'
+  - cron: 30 1 * * *
```
namurphy, line 4:
```suggestion
  - cron: '30 1 * * *'
```
Un-apply this fix since `*` is a wildcard character.

namurphy:
One worry I had with some of the changes is that many strings that included `*` are no longer in quotes.  I *think* that this is fine, given this description of [YAML syntax](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) talks about strings that *start* with `*`.  There's a chance I'm wrong about this, so if this PR causes problems, this would be a good place to look.

rocco8773 (approved):

PR #1288, opened 2021-10-02
Add a pre-commit hook to auto-format docstrings and partially apply it

namurphy:
This PR installs a pre-commit hook for [docformatter](https://github.com/myint/docformatter), which we can use to auto-format docstrings.  

Applying the changes all at once would probably lead to merge conflicts, so for now my intention is to make these changes only for files that are not being touched by any pull requests.  The rest of the formatting changes will get applied gradually and we touch files.

The only substantive change is in `.pre-commit-config.yaml`.  All of the other changes are being automatically applied.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,3 +54,9 @@ repos:
     args: [--autofix]
   - id: pretty-format-yaml
     args: [--autofix, --indent, '2']
+
+- repo: https://github.com/myint/docformatter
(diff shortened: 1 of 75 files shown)


namurphy:
Closing this for the time being because there are a few problems with how docformatter is formatting our docs, so we shouldn't add this as a pre-commit hook...yet.  I had really been wanting to add a docstring formatter too.  😿

PR #1295, opened 2021-10-06
Faster interpolation on 'grid' objects

pheuer:
Interpolation on `grid` objects is a significant bottleneck. For example, ~90% of the execution time of the charged particle radiography `Tracker._push()` method (the main simulation loop) is spent on the interpolation operation. 

<details>
  <summary>Current cpr.Tracker _push() method line profile</summary>

```
Timer unit: 1e-07 s

Total time: 76.6656 s
File: c:\users\pvheu\github\plasmapy\plasmapy\diagnostics\charged_particle_radiography.py
Function: _push at line 860

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   860                                               @profile
   861                                               def _push(self):
   862                                                   r"""
   863                                                   Advance particles using an implementation of the time-centered
   864                                                   Boris algorithm
   865                                                   """
   866                                                   # Get a list of positions (input for interpolator)
   867       416    4394059.0  10562.6      0.6          pos = self.x[self.grid_ind, :] * u.m
   868                                           
   869                                                   # Update the list of particles on and off the grid
   870       416   11811912.0  28394.0      1.5          self.on_grid = self.grid.on_grid(pos)
   871                                                   # entered_grid is zero at the end if a particle has never
   872                                                   # entered the grid
   873       416     346047.0    831.8      0.0          self.entered_grid += self.on_grid
   874                                           
   875                                                   # Estimate the E and B fields for each particle
   876                                                   # Note that this interpolation step is BY FAR the slowest part of the push
   877                                                   # loop. Any speed improvements will have to come from here.
   878       416       5193.0     12.5      0.0          if self.field_weighting == "volume averaged":
   879       208       3463.0     16.6      0.0              Ex, Ey, Ez, Bx, By, Bz = self.grid.volume_averaged_interpolator(
   880       208       1213.0      5.8      0.0                  pos,
   881       208       1201.0      5.8      0.0                  "E_x",
   882       208       1194.0      5.7      0.0                  "E_y",
   883       208       1188.0      5.7      0.0                  "E_z",
   884       208       1178.0      5.7      0.0                  "B_x",
   885       208       1168.0      5.6      0.0                  "B_y",
   886       208       1176.0      5.7      0.0                  "B_z",
   887       208  675598413.0 3248069.3     88.1                  persistent=True,
   888                                                       )
   889       208       1474.0      7.1      0.0          elif self.field_weighting == "nearest neighbor":
   890       208       2339.0     11.2      0.0              Ex, Ey, Ez, Bx, By, Bz = self.grid.nearest_neighbor_interpolator(
   891       208       1191.0      5.7      0.0                  pos,
   892       208       1323.0      6.4      0.0                  "E_x",
   893       208       1219.0      5.9      0.0                  "E_y",
   894       208       1340.0      6.4      0.0                  "E_z",
   895       208       1194.0      5.7      0.0                  "B_x",
   896       208       1181.0      5.7      0.0                  "B_y",
   897       208       1176.0      5.7      0.0                  "B_z",
   898       208    5657958.0  27201.7      0.7                  persistent=True,
   899                                                       )
   900                                           
   901                                                   # Interpret any NaN values (points off the grid) as zero
   902       416    3651889.0   8778.6      0.5          Ex = np.nan_to_num(Ex, nan=0.0 * u.V / u.m)
   903       416    2852127.0   6856.1      0.4          Ey = np.nan_to_num(Ey, nan=0.0 * u.V / u.m)
   904       416    2808941.0   6752.3      0.4          Ez = np.nan_to_num(Ez, nan=0.0 * u.V / u.m)
   905       416    2145823.0   5158.2      0.3          Bx = np.nan_to_num(Bx, nan=0.0 * u.T)
   906       416    2185203.0   5252.9      0.3          By = np.nan_to_num(By, nan=0.0 * u.T)
   907       416    2134640.0   5131.3      0.3          Bz = np.nan_to_num(Bz, nan=0.0 * u.T)
   908                                           
   909                                                   # Create arrays of E and B as required by push algorithm
   910       416       5024.0     12.1      0.0          E = np.array(
   911       416    3836587.0   9222.6      0.5              [Ex.to(u.V / u.m).value, Ey.to(u.V / u.m).value, Ez.to(u.V / u.m).value]
   912                                                   )
   913       416     146390.0    351.9      0.0          E = np.moveaxis(E, 0, -1)
   914       416    1719429.0   4133.2      0.2          B = np.array([Bx.to(u.T).value, By.to(u.T).value, Bz.to(u.T).value])
   915       416     106452.0    255.9      0.0          B = np.moveaxis(B, 0, -1)
   916                                           
   917                                                   # Calculate the adaptive timestep from the fields currently experienced
   918                                                   # by the particles
   919                                                   # If user sets dt explicitly, that's handled in _adpative_dt
   920       416    8653705.0  20802.2      1.1          dt = self._adaptive_dt(Ex, Ey, Ez, Bx, By, Bz)
   921                                           
   922                                                   # TODO: Test v/c and implement relativistic Boris push when required
   923                                                   # vc = np.max(v)/_c
   924                                           
   925       416    2989250.0   7185.7      0.4          x = self.x[self.grid_ind, :]
   926       416    2913081.0   7002.6      0.4          v = self.v[self.grid_ind, :]
   927       416   27832421.0  66904.9      3.6          boris_push(x, v, B, E, self.q, self.m, dt)
   928       416    2442019.0   5870.2      0.3          self.x[self.grid_ind, :] = x
   929       416    2394850.0   5756.9      0.3          self.v[self.grid_ind, :] = v
```
</details>

Within the volume averaged interpolator, the most expensive step (60% of runtime) is a simple nearest-neighbor interpolation. 

```python
nearest_neighbor_index[..., 0] = np.abs(pos[:, 0, None] - ax0).argmin(axis=1)
```

<details>
  <summary>Current Volume Averaged Interpolator Line Profile</summary>

```
Timer unit: 1e-07 s

Total time: 69.3817 s
File: c:\users\pvheu\github\plasmapy\plasmapy\plasma\grids.py
Function: volume_averaged_interpolator at line 1172

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1172                                               @profile
  1173                                               def volume_averaged_interpolator(
  1174                                                   self, pos: Union[np.ndarray, u.Quantity], *args, persistent=False
  1175                                               ):
  1176                                                   """
  1177                                                   Notes
  1178                                                   -----
  1179                                           
  1180                                                   This interpolator approximates the value of a quantity at a given
  1181                                                   interpolation point using a weighted sum of the values at the eight grid
  1182                                                   vertices that surround the point. The weighting factors are calculated by
  1183                                                   defining a volume :math:`dx \\times dy \\times dz`
  1184                                                   (where :math:`dx`, :math:`dy`, and :math:`dz` are the grid
  1185                                                   spacings in each direction) around each grid vertex and around the
  1186                                                   interpolation point. The contribution of each grid vertex is then
  1187                                                   weighted by the fraction of the volume surrounding the interpolation
  1188                                                   point that overlaps the volume surrounding that vertex. This effectively
  1189                                                   introduces a linear interpolation between grid vertices.
  1190                                           
  1191                                                   This implementation of this algorithm assumes that the grid is uniformly
  1192                                                   spaced and Cartesian.
  1193                                                   """
  1194                                           
  1195       208       8953.0     43.0      0.0          if isinstance(pos, u.Quantity):
  1196       208    1088628.0   5233.8      0.2              pos = pos.to(u.m).value
  1197                                                   elif self.unit != u.m:
  1198                                                       pos *= self.unit.si.scale
  1199                                           
  1200                                                   # If a single point was given, add empty dimension
  1201       208       4394.0     21.1      0.0          if pos.ndim == 1:
  1202                                                       pos = np.reshape(pos, [1, 3])
  1203                                           
  1204                                                   # -- Validate args --
  1205                                                   # must be np.ndarray or u.Quantity arrays of same shape as grid
  1206      1456      17849.0     12.3      0.0          for arg in args:
  1207      1248     241101.0    193.2      0.0              if arg not in self.quantities:
  1208                                                           raise KeyError(
  1209                                                               "Quantity arguments must correspond to "
  1210                                                               "DataArrays in the DataSet. "
  1211                                                               f"{arg} was not found. "
  1212                                                               f"Existing keys are: {self.quantities}"
  1213                                                           )
  1214                                           
  1215                                                   # If persistent, double check the arguments list hasn't changed
  1216                                                   # If they have, run as non-persistent this time
  1217       208       3942.0     19.0      0.0          if persistent and args != self._interp_args:
  1218         1         17.0     17.0      0.0              persistent = False
  1219                                           
  1220                                                   # Update _interp_args variable
  1221       208       3133.0     15.1      0.0          self._interp_args = args
  1222                                           
  1223                                                   # If not persistent, clear the cached properties so they are re-created
  1224                                                   # when called below
  1225       208       2373.0     11.4      0.0          if not persistent:
  1226         1         17.0     17.0      0.0              try:
  1227         1         43.0     43.0      0.0                  del self._interp_quantities
  1228         1         20.0     20.0      0.0              except AttributeError:
  1229         1         19.0     19.0      0.0                  pass
  1230         1         16.0     16.0      0.0              try:
  1231         1         24.0     24.0      0.0                  del self._interp_units
  1232         1         18.0     18.0      0.0              except AttributeError:
  1233         1         17.0     17.0      0.0                  pass
  1234                                           
  1235                                                   # -- begin averaging --
  1236                                           
  1237       208       4018.0     19.3      0.0          nparticles = pos.shape[0]
  1238       208       2868.0     13.8      0.0          nargs = len(args)
  1239                                           
  1240                                                   # Load grid attributes (so this isn't repeated)
  1241       208     848529.0   4079.5      0.1          ax0, ax1, ax2 = self._ax0_si, self._ax1_si, self._ax2_si
  1242       208    1546030.0   7432.8      0.2          dx, dy, dz = self._dax0_si, self._dax1_si, self._dax2_si
  1243       208     975244.0   4688.7      0.1          n0, n1, n2 = self.shape
  1244                                           
  1245                                                   # find cell nearest to each position
  1246       208     586326.0   2818.9      0.1          nearest_neighbor_index = np.zeros((nparticles, 3), dtype=np.int32)
  1247       208  125926741.0 605417.0     18.1          nearest_neighbor_index[..., 0] = np.abs(pos[:, 0, None] - ax0).argmin(axis=1)
  1248       208  123074811.0 591705.8     17.7          nearest_neighbor_index[..., 1] = np.abs(pos[:, 1, None] - ax1).argmin(axis=1)
  1249       208  125973774.0 605643.1     18.2          nearest_neighbor_index[..., 2] = np.abs(pos[:, 2, None] - ax2).argmin(axis=1)
  1250                                           
  1251                                                   # Create a mask for positions that are off the grid. The values at
  1252                                                   # these points will be set to zero later.
  1253                                                   mask_particle_off = (
  1254                                                       (pos[:, 0] < ax0.min())
  1255                                                       | (pos[:, 0] > ax0.max())
  1256                                                       | (pos[:, 1] < ax1.min())
  1257                                                       | (pos[:, 1] > ax1.max())
  1258                                                       | (pos[:, 2] < ax2.min())
  1259       208    1804231.0   8674.2      0.3              | (pos[:, 2] > ax2.max())
  1260                                                   )
  1261                                           
  1262                                                   # Get the physical positions for the nearest neighbor cell
  1263                                                   # for the each particle
  1264       208     624631.0   3003.0      0.1          xpos = ax0[nearest_neighbor_index[:, 0]]
  1265       208     866149.0   4164.2      0.1          ypos = ax1[nearest_neighbor_index[:, 1]]
  1266       208     852757.0   4099.8      0.1          zpos = ax2[nearest_neighbor_index[:, 2]]
  1267       208     917167.0   4409.5      0.1          nearest_neighbor_pos = np.array([xpos, ypos, zpos]).swapaxes(0, 1)
  1268                                           
  1269                                                   # Determine the indices for the grid cells bounding the particle
  1270                                                   # - The imaginary cell centered on the particle will overlap with
  1271                                                   #   1 to 8 surrounding cells
  1272                                                   # - typically this is 8 but will be 4, 2, or 1 when the particle is
  1273                                                   #   near the boundary of the grid
  1274       208      37593.0    180.7      0.0          bounding_cell_indices = np.empty((nparticles, 8, 3), dtype=np.int32)
  1275       208       3485.0     16.8      0.0          lower_indices = np.where(
  1276       208    1395249.0   6707.9      0.2              pos >= nearest_neighbor_pos,
  1277       208       3407.0     16.4      0.0              nearest_neighbor_index,
  1278       208    1806273.0   8684.0      0.3              nearest_neighbor_index - 1,
  1279                                                   )
  1280                                           
  1281                                                   # populate x indices
  1282       208       6321.0     30.4      0.0          bounding_cell_indices[:, 0:4, 0] = np.tile(
  1283       208     839218.0   4034.7      0.1              lower_indices[:, 0], (4, 1)
  1284       208    3430921.0  16494.8      0.5          ).swapaxes(0, 1)
  1285       208    6409795.0  30816.3      0.9          bounding_cell_indices[:, 4:, 0] = bounding_cell_indices[:, 0:4, 0] + 1
  1286                                           
  1287                                                   # populate y indices
  1288       208       8006.0     38.5      0.0          bounding_cell_indices[:, [0, 1, 4, 5], 1] = np.tile(
  1289       208     865314.0   4160.2      0.1              lower_indices[:, 1], (4, 1)
  1290       208    4404428.0  21175.1      0.6          ).swapaxes(0, 1)
  1291                                                   bounding_cell_indices[:, [2, 3, 6, 7], 1] = (
  1292       208    8709377.0  41872.0      1.3              bounding_cell_indices[:, [0, 1, 4, 5], 1] + 1
  1293                                                   )
  1294                                           
  1295                                                   # populate z indices
  1296       208      13132.0     63.1      0.0          bounding_cell_indices[:, 0::2, 2] = np.tile(
  1297       208     971725.0   4671.8      0.1              lower_indices[:, 2], (4, 1)
  1298       208    1586399.0   7626.9      0.2          ).swapaxes(0, 1)
  1299       208    3023317.0  14535.2      0.4          bounding_cell_indices[:, 1::2, 2] = bounding_cell_indices[:, 0::2, 2] + 1
  1300                                           
  1301                                                   # Create a mask for cells whose locations would be off the grid,
  1302                                                   # which occurs when the interpolation point is near the edge of the
  1303                                                   # grid. These values will be weighted as zero later.
  1304                                                   mask_cell_off = (
  1305                                                       (bounding_cell_indices < 0).any(axis=2)
  1306                                                       | (bounding_cell_indices[:, :, 0] >= n0)
  1307                                                       | (bounding_cell_indices[:, :, 1] >= n1)
  1308       208   30308654.0 145714.7      4.4              | (bounding_cell_indices[:, :, 2] >= n2)
  1309                                                   )
  1310                                           
  1311                                                   # Zero any out of bounds indices so IndexError is not raised
  1312                                                   # during indexing.  This means an incorrect value will be retrieved
  1313                                                   # but will not be used because of the zero weighting and the
  1314                                                   # off the grid mask
  1315       208   15590151.0  74952.6      2.2          bounding_cell_indices[mask_cell_off, :] = 0
  1316                                           
  1317                                                   # Calculate the volume of the overlap between the point volume
  1318                                                   # and the volume of each of the surrounding vertices
  1319       208   18584476.0  89348.4      2.7          lx = dx - np.abs(pos[:, None, 0] - ax0[bounding_cell_indices[..., 0]])
  1320       208   18462579.0  88762.4      2.7          ly = dy - np.abs(pos[:, None, 1] - ax1[bounding_cell_indices[..., 1]])
  1321       208   18622794.0  89532.7      2.7          lz = dz - np.abs(pos[:, None, 2] - ax2[bounding_cell_indices[..., 2]])
  1322       208    7189681.0  34565.8      1.0          bounding_cell_weights = lx * ly * lz
  1323                                           
  1324                                                   # Set the weight for any off-grid vertices (cell or particle) to zero
  1325       208    5303491.0  25497.6      0.8          bounding_cell_weights[mask_cell_off] = 0.0
  1326       208    3045214.0  14640.5      0.4          bounding_cell_weights[mask_particle_off, ...] = 0.0
  1327       208    2732543.0  13137.2      0.4          norms = np.sum(bounding_cell_weights, axis=1)
  1328       208      85219.0    409.7      0.0          mask_norm_zero = norms == 0.0
  1329                                                   bounding_cell_weights[~mask_norm_zero] = (
  1330       208   10767744.0  51768.0      1.6              bounding_cell_weights[~mask_norm_zero, ...] / norms[~mask_norm_zero, None]
  1331                                                   )
  1332                                           
  1333                                                   # Get the values of each of the interpolated quantities at each
  1334                                                   # of the bounding vertices
  1335       208     351548.0   1690.1      0.1          vals = self._interp_quantities[
  1336       208       7805.0     37.5      0.0              bounding_cell_indices[..., 0],
  1337       208       3327.0     16.0      0.0              bounding_cell_indices[..., 1],
  1338       208   51655461.0 248343.6      7.4              bounding_cell_indices[..., 2],
  1339                                                       :,
  1340                                                   ]
  1341                                                   # Construct a weighted average of the interpolated quantities
  1342       208   84508743.0 406292.0     12.2          weighted_ave = np.sum(bounding_cell_weights[..., None] * vals, axis=1)
  1343       208    3144540.0  15118.0      0.5          weighted_ave[mask_particle_off, :] = np.nan
  1344                                           
  1345                                                   # Split output array into arrays with units
  1346                                                   # Apply units to output arrays
  1347       208       5167.0     24.8      0.0          output = []
  1348      1456      35433.0     24.3      0.0          for arg in range(nargs):
  1349      1248    4515946.0   3618.5      0.7              output.append(weighted_ave[..., arg] * self._interp_units[arg])
  1350                                           
  1351       208       4418.0     21.2      0.0          if len(output) == 1:
  1352                                                       return output[0]
  1353                                                   else:
  1354       208       4188.0     20.1      0.0              return tuple(output)
```
</details>

In this PR I will try to get rid of any low hanging fruit to speed up these functions while maintaining the existing functionality. I will try to restrict the scope of modifications here to `grids.py`, leaving improvements to `cpr.Tracker` for later.

...I also ended up moving a couple grid attributes from AbstractGrid to subclasses to avoid future errors, and cleaning up the tests for AbstractGrid.

TODO

- [x] Introduce a nearest-neighbor interpolator that assumes the axes are sorted (a valid assumption for uniform grids). 
- [x] Apply faster nearest-neighbor interpolator to volume averaging interpolator for Cartesian grids.
- [x] Eliminate units within the interpolator functions as much as possible.


pheuer:
Big speedup by assuming that the axis is sorted! These two code blocks are equivalent 

Execution time (1e5 particles, axis=(200,): 60 ms
```python
indices = np.abs(pos[:,None] - ax).argmin(axis=1)

```

Execution time (1e5 particles, axis=(200,): 5 ms
```python
  # Find the index where each position would be inserted into the axis.
  # This is equivalent to a nearest neighbor interpolation but always
  # rounding to the next highest axis point.
  indices = np.searchsorted(ax, pos)

  # searchsorted can return an out-of-bound index on the top end, so replace
  # that point with the closest in bound index
  indices = np.where(indices == ax.size, ax.size - 1, indices)

  # For any points that are closer to the point below than the point above,
  # correct the index
  indices = np.where(
      np.abs(ax[indices] - pos) > np.abs(ax[indices - 1] - pos), indices - 1, indices
  )
```

It should be possible to use the later whenever axes are sorted, which should always be the case (even for non-uniform grids). We should probably add a check that makes sure this is the case when the grid is created?

This change alone drops the runtime of a test `cpr.Tracker()` run from 70 s to 29 s.

StanczakDominik:
Sorry for the delay; tackling this one this week!

plasmapy/plasma/tests/test_grids.py
```diff
@@ -556,5 +970,22 @@ def debug_volume_avg_interpolator():
...
 
-if __name__ == "__main__":
-    test_volume_averaged_interpolator_known_solutions()
+def test_fast_nearest_neighbor_interpolate():
+    """
+    Confirms that the fast linear interpolator function is equivalent to
+    the np.argmin(np.abs(x-y)) search method for ordered arrays
+    """
+    ax = 100 * np.linspace(0, 1, num=100)
+    pos = np.random.random([300])
```
namurphy, line 979: Could you add a random seed to this and to other places where random number generators are used?  In general it's better to have tests that are deterministic, since it's hard to tell when an intermittently failing test is actually fixed.  Thanks!

pheuer: Oh, yes, good idea. Done.

plasmapy/plasma/grids.py
```diff
@@ -921,65 +943,43 @@ def _interp_units(self):
...
-            self.ax0.si.value,
-            self.ax1.si.value,
-            self.ax2.si.value,
-        )
-
-        self._interpolator = interp.RegularGridInterpolator(
-            pts, indgrid, method="nearest", bounds_error=False, fill_value=np.nan
+        raise NotImplementedError(
+            "Nearest-neighbor interpolator is not yet "
+            "implemented for this grid type."
```
StanczakDominik, line 974: Could/should this `nearest_neighbor_interpolator`, then, be an [abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod)? In other words - what minimal set of promises do we make about each actual implementation of the AbstractGrid?

Just food for thought for now, I'm okay with this as it is :)

plasmapy/plasma/grids.py
```diff
@@ -557,23 +557,6 @@ def dax2(self):
...
-        For non-uniform grids, it is the closest spacing between any two points.
-        """
-
-        if self.is_uniform:
-            return min(self.dax0, self.dax1, self.dax2)
-        else:
-            distances = distance.cdist(self.grid, self.grid)
-            np.fill_diagonal(distances, np.inf)
-            return np.min(distances)
-
```
StanczakDominik, line 576: Seems like this, given it's implemented in the subclasses, could also be an abstract method ([like this](https://docs.python.org/3/library/abc.html#abc.abstractproperty)).

plasmapy/plasma/grids.py
```diff
@@ -989,54 +989,53 @@ def interpolate_indices(self, pos: Union[np.ndarray, u.Quantity]):
...
-        # currently the nonuniform interpolator can't tell when a value
-        # is out of bounds...
-
-        # Note: i contains nan values which must be replaced with 0's with
-        # appropriate units in the second layer interpolator functions.
 
-        return i
+        raise NotImplementedError(
+            "Volume-averaged interpolator is not yet " "implemented for this grid type."
+        )
```
StanczakDominik, line 1006: If this is a special case in the CartesianGrid, do we want this in the base class, or do we just want to move it into that subclass? I guess you also expect other eventual subclasses to implement this one...?

plasmapy/plasma/grids.py
```diff
@@ -1101,57 +1094,32 @@ def nearest_neighbor_interpolator(
...
-        )
+    """
+    # Find the index where each position would be inserted into the axis.
+    # This is equivalent to a nearest neighbor interpolation but always
+    # rounding to the next highest axis point.
+    indices = np.searchsorted(ax, pos)
+
+    # searchsorted can return an out-of-bound index on the top end, so replace
+    # that point with the closest in bound index
+    indices = np.where(indices == ax.size, ax.size - 1, indices)
```
StanczakDominik, line 1115: Good call on `np.where` here, I'd have thought to use `np.clip` and that'd be 6 times slower. 👏🏻

StanczakDominik (approved):
*technically within a week!* 😅 

All right! Some minor comments/questions to talk about, a few typos, rewrote your seeding a little, but all in all I wouldn't dare to hold this down for any longer!

StanczakDominik:
I can help out with the tests (failing linter?) tomorrow after work.

plasmapy/plasma/grids.py
pheuer: Yes, I think probably it should (not the volume_averaged_interpolator, though: those aren't defined for all grids)

plasmapy/plasma/grids.py
pheuer: Yes, although probably not all of them (very computationally intensive on a non-uniform grid, I think). These algorithms have to be custom-written for each grid since they rely on the specific geometry.

plasmapy/plasma/grids.py
```diff
@@ -1101,57 +1094,32 @@ def nearest_neighbor_interpolator(
...
-            `~numpy.ndarray` is provided, units will be assumed to match
-            those of the grid.
+def _fast_nearest_neighbor_interpolate(pos, ax):
+    """
+    This function finds the indices in the axis 'ax' that are closet to the
+    values in the array 'pos'
 
-        *args : `str`
-            Strings that correspond to DataArrays in the dataset
+    Assumes the axis 'ax' is sorted in accending order.
```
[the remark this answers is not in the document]
pheuer: Ack, sorry for my avant garde spelling...

plasmapy/plasma/grids.py
pheuer: I rewrote this function a few times with different alternatives before arriving at this one! Might be room to speed it up even further by rewriting it with Numba...

plasmapy/plasma/tests/test_grids.py
```diff
@@ -9,278 +9,599 @@
...
-    # Test adding a quantity of incompatible size
-    q = np.random.randn(5, 20, 5) * u.T
-    with pytest.raises(ValueError):
-        grid.add_quantities(B_x=q)
+abstract_attrs = [
+    ("uniform", "is_uniform", bool, None, True),
+    ("nonuniform", "is_uniform", bool, None, False),
+    ("uniform", "shape", tuple, int, (21, 21, 21)),
+    ("nonuniform", "shape", tuple, int, (125,)),
+    ("uniform", "grid", u.Quantity, None, None),
```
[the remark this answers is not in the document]
pheuer: Eh, I thought about it more and agree with you: changed.

plasmapy/plasma/grids.py
pheuer: Ok, so I made `nearest_neighbor_interpolator` an abstract method. I also did the same with `vector_intersects` because the current implementation is inherently Cartesian.

As a result, one can no longer instantiate instances of AbstractGrid. So, in the tests I changed all the AbstractGrid tests to us CartesianGrid, which is a subclass.

plasmapy/plasma/tests/test_grids.py
pheuer: Thanks!

plasmapy/plasma/grids.py
StanczakDominik: Seems to fit the Abstract in AbstractGrid, anyway!

plasmapy/plasma/grids.py
StanczakDominik: Yeah, that's tougher. In the cylindrical case I'd expect this to mean "smallest of the three grid resolutions, provided the angular resolution is..."  Oh. Yeah, because the angular resolution aka arc length is going to be a function of distance from the Z axis, this would also be a function of the position... Yeah, let's skip this one. Or we could just decide to do min(Res_r, Res_z). But that seems arbitrary until we encounter a use case.

plasmapy/plasma/grids.py
StanczakDominik: Probably not, because I expect numba would act on each element... Hmm. Come to think of it, my timings didn't include defining `indices` for the case of `where`, so that's cheating a little. But yeah, this looks great.

StanczakDominik:
Okay, the linter failure is unrelated. I'm just going to go ahead and merge this :)

Issue #1056, opened 2021-03-04
Convert from energy to velocity in relativistic formulary


sfeister:
## Feature request
I can't find a good function in either plasmapy or astropy to convert from kinetic energy into relativistic velocity for a single particle. If it doesn't exist, I'm happy to add it to the relativistic formulary.

pheuer:
If this function doesn't exist somewhere I've overlooked I'm happy to help @phyzicist add this. This function would be a convenient refactor of these lines in `diagnostics.proton_radiography.SyntheticProtonRadiograph`:

```
# Calculate the velocity corresponding to the proton energy
ER = self.proton_energy * 1.6e-19 / (self.m * self._c ** 2)
v0 = self._c * np.sqrt(1 - 1 / (ER + 1) ** 2)
```

phyzicist:
I propose adding a function:

relativistic_velocity(m, E) | Calculate the velocity (in km/s) of an object of mass m and relativistic energy E.

phyzicist:
I could also see this being articulated as an equivalency

StanczakDominik:
Relativistic_velocity with an energy input sounds great and definitely more obvious than than any name with "equivalency" :)

One possible variant: speed rather than velocity? With an energy input we can't get the direction, and the way it's taught here in Poland speed implies the scalar magnitude of velocity. I don't actually know whether that distinction is common elsewhere.

pheuer:
Agreed, `relativistic_speed` would be better. It should probably take `energy` and a `particle` as required inputs?

rocco8773:
> I propose adding a function:
>
> relativistic_velocity(m, E, kinetic=False) | (default) Calculate the velocity (in meters/second) of an object with rest mass m and total relativistic energy E.
> relativistic_velocity(m, E, kinetic=True) | Calculate the velocity (in meters/second) of an object with rest mass m and kinetic energy E.

My issue with just using `relativistic_speed` is that the speed could be calculated from total energy, kinetic energy, and momentum.  `relativistic_speed` is not descriptive enough to indicate which one it is.  We could possibly sidestep this if we queue off off the input units, something like...

```python
def relativistic_speed(value: [u.J, u.kg * u.m / u.s], mass):
    # if value.unit equivalent to u.J, convert from energy
    ...

    # if value.unit equivalent to u.kg * u.m / u.s, convert from momentum
    ...

    return speed
```

---

[...]

PR #1297, opened 2021-10-06
Convert from energy to velocity in relativistic formulary (#1056)

Tlord18:
* Add function to convert kinetic energy and total energy to relativistic speed

      docstrings.

### DIFF
--- a/plasmapy/formulary/relativity.py
+++ b/plasmapy/formulary/relativity.py
@@ -155,3 +155,25 @@ def relativistic_energy(m: u.kg, v: u.m / u.s) -> u.Joule:
     γ = Lorentz_factor(v)
     E = γ * m * c ** 2
     return E
+
+
+@validate_quantities(value={"can_be_negative": False})
+def relativistic_speed(
+    value: [u.J, u.kg * u.m / u.s],
+    particle: ParticleLike,
+    is_kinetic_energy: bool = False,
+) -> u.m / u.s:
+    if value.units.is_equivalent(u.J):
+        if is_kinetic_energy:
+            p = (1 / c) * sqrt(
+                ((value + particle.mass * c ** 2) ** 2)
+                - ((particle.mass ** 2) * (c ** 4))
+            )
+        else:
+            p = (1 / c) * sqrt((value ** 2) - ((particle.mass ** 2) * (c ** 4)))
+
+    p = value
+    speed = p / (γ * particle.mass)
+
+
+return speed


plasmapy/formulary/relativity.py
```diff
@@ -155,3 +155,25 @@ def relativistic_energy(m: u.kg, v: u.m / u.s) -> u.Joule:
...
     E = γ * m * c ** 2
     return E
+
+
+@validate_quantities(value={"can_be_negative": False})
+def relativistic_speed(
+    value: [u.J, u.kg * u.m / u.s],
+    particle: ParticleLike,
+    is_kinetic_energy: bool = False,
+) -> u.m / u.s:
```
pheuer, line 165: Add a docstring. Look at the other formulary functions to see how to format it, and what kind of information should be included. Make sure to explicitly explain the three separate ways this can be used 

1) Momentum -> velocity
2) KE -> velocity
3) Total energy -> velocity

plasmapy/formulary/relativity.py
```diff
@@ -155,3 +155,25 @@ def relativistic_energy(m: u.kg, v: u.m / u.s) -> u.Joule:
...
+    if value.units.is_equivalent(u.J):
+        if is_kinetic_energy:
+            p = (1 / c) * sqrt(
+                ((value + particle.mass * c ** 2) ** 2)
+                - ((particle.mass ** 2) * (c ** 4))
+            )
+        else:
+            p = (1 / c) * sqrt((value ** 2) - ((particle.mass ** 2) * (c ** 4)))
```
pheuer, line 173: Better to use `np.sqrt` everywhere, rather than `sqrt`?

pheuer (commented):
This is a good start! In addition to the comments I left, you will also need to write test functions that test this new function. For examples, look in the file formulary/tests/test_relativity.py. Any function in these files with a name starting in 'test" will be automatically run by the unit test module every time a new change to plasmapy is made, to make sure this function is still performing as expected.

(You may already know this, but you can make changes to this PR by pushing new commits to your branch. In the future, it's usually better to make a new branch for a PR rather than using your main branch (Tlord18:main). But, this should be ok for now.)

plasmapy/formulary/relativity.py
```diff
@@ -155,3 +155,25 @@ def relativistic_energy(m: u.kg, v: u.m / u.s) -> u.Joule:
...
+
+
+@validate_quantities(value={"can_be_negative": False})
+def relativistic_speed(
+    value: [u.J, u.kg * u.m / u.s],
+    particle: ParticleLike,
+    is_kinetic_energy: bool = False,
+) -> u.m / u.s:
+    if value.units.is_equivalent(u.J):
+        if is_kinetic_energy:
```
namurphy, line 167: I'm wondering if here we could use the `physical_type` attribute of units so that we wouldn't need the `is_kinetic_energy` keyword.

```Python
if value.unit.physical_type == "energy":
```

namurphy (commented):
Thank you for making this contribution!  This will be an important thing to have in `plasmapy.formulary`.  A few thoughts...

- It'd also be helpful to use the [`plasmapy.particles.particle_input`](https://docs.plasmapy.org/en/stable/api/plasmapy.particles.particle_input.html) decorator here, which auto-converts a representation of a particle into a `Particle`.  (The annotation would have to be changed from `ParticleLike` to `Particle` for the time being, but that will change in the refactoring of `particle_input` that I'm hoping to get to by our 0.8 release.)

- We just updated our [doc guide](https://docs.plasmapy.org/en/latest/development/doc_guide.html) and [testing guide](https://docs.plasmapy.org/en/latest/development/testing_guide.html) which might be helpful for some of @pheuer's comments.

Thank you again!

Tlord18:
Thanks a lot for these comments, I'll implement them and commit them soon. I must've forgotten the else, so I'll change that. I'll also get started on the tests and the documentation

> (You may already know this, but you can make changes to this PR by pushing new commits to your branch. In the future, it's usually better to make a new branch for a PR rather than using your main branch (Tlord18:main). But, this should be ok for now.)

Ah yes, I had initially created a branch, where I had made a few sample commits initially. In an attempt to squash them before making a PR, I kinda messed it up somewhere. So I ended up making a PR through the main branch. I'll make a new branch again. Thanks a lot!

pheuer:
@namurphy  #1056 is closed - should we close this PR?

PR #1308, opened 2021-10-15
Lite-Functions Ep. 2 | plasma_frequency_lite

rocco8773:
PR #1145 needs to be merged first

This PR adds the plasma frequency to the lite-function framework.  Creates `plasma_frequency_lite`.

PR #1145, opened 2021-05-19
Lite-Functions Ep. 1 | Jump to Hyperspace

rocco8773:
This is part 1 of incorporating the Lite-Function framework into `plasmapy`.   This PR focuses on the functionality to define and setup lite-functions, whereas part 2 will focus on the documentation side (i.e. custom functionality for `plasmapy_sphinx.autodoc` that documents function attributes).

What's added here:

1. setup of the `__lite_func__` group so `plasampy_sphinx.autodoc.automodapi` generates summary tables for the specific functionality
2. define decorator `@mark_has_lite_func` to do all the necessary attribute binding and registry of lite-functions
3. Implement the two-function scheme for normal and lite functions (i.e. `thermal_speed` is the normal function and `thermal_speed_lite` is the lite-function)

This should be merged after PR #1144.

PR #1312, opened 2021-10-16
[WIP] Decompose `plasmapy/formulary/parameters.py`

rocco8773:
The `parameters.py` has become increasingly monolithic, making it burdensome to maintain and expand.  This PR is my attempt at decomposing `parameters.py` to make it's functionality much more manageable.

---

I'd like to get this refactoring done before the release of `v0.8.0`, but definitely after the release of `v0.7.0`.  I'd also like to merge this after incorporating the lite-function functionality (#1145, #1308).  If there are any other open PRs touching code in `parameters.py`, then I'd wait on their merge as well and I'd handle the merge conflicts in this PR.

---

I potential see two options for decomposing `parameters.py`.  This PR is currently incorporating **option 1**, but I'm slowly thinking **option 2** is better.  @namurphy @StanczakDominik What are your thoughts on this?

**Option 1**

Convert `parameters.py` into a directory with the following structure....

```bash
plasmapy
+-- formulary
|   +-- parameters
|   |   +-- __init__.py
|   |   +-- dimensionless.py
|   |   +-- frequencies.py
|   |   +-- lengths.py
|   |   +-- misc.py
|   |   +-- speeds.py
```

If you did the imports correctly in the `__init__.py` then this wouldn't cause anything to break.  The reason I'm falling out of favor with this option is that it creates some ambiguity on where to place some functionality.  For example, there would be a `dimensionless.py` file in both the `formulary` directory and the `parameters` directory.

**Option 2**

Decompose the functionality in `parameters.py` into separate files like option 1, but keep these files in the `formulary` directory.  Additionally, keep a `parameters.py` with proper imports to maintain backwards compatibility, but `parameters.py` would not contain any defined functionality.

```bash
plasmapy
+-- formulary
|   +-- __init__.py
|   +-- ...
|   +-- dimensionless.py
|   +-- frequencies.py
|   +-- lengths.py
|   +-- misc.py
|   +-- parameters.py
|   +-- speeds.py
|   +-- ...
```

I feel this approach is more clear on where to place future functionality.  I'd just place some kind of `FutureWarning` on imports from `parameters.py` so users have time to move away from it.


StanczakDominik:
Gah, looks like I'm late to the party... I'm definitely also in favor of option 2. Zen of Python, line 5: *flat is better than nested*. Seems applicable here.

It also looks like this PR got mixed up with #1145 at least, so I guess it's stalled until we get that one through?

rocco8773:
@StanczakDominik @namurphy Ok, I'm going to proceed with option 2.  I always knew this was going to fall behind quickly, so the refactor is likely to be a pain.  But the pkg will be better for it.

rocco8773:
@StanczakDominik @namurphy This PR has gotten far enough behind that it will be difficult to properly update.  Additionally, a lot of code needs to be refactored which will make a single PR rather large.  In an attempt to make this process less painful, I think we should do each decomposition (lengths, frequencies, speeds, etc.) in separate PRs.  I've outlined this process in issue #1433.

PR #1316, opened 2021-10-19
Discuss parametrization with argument unpacking in testing guide

namurphy:
This PR expands the testing guide to include an example where we want to test a function both with and without a keyword argument so that we use argument unpacking. 

I still need to briefly mention what `*args` and `**kwargs` do in this example.


StanczakDominik (approved):
LGTM! 👍🏻

PR #1334, opened 2021-11-18
Bump minimum version of matplotlib and enable matplotlib 3.5.0

namurphy:
The `nonposy` keyword was removed from matplotlib 3.5.0, and replaced with `nonpositive` which was added in matplotlib 3.3.0.  The easiest fix was to replace `nonposy` with `nonpositive` and bump the minimum version of matplotlib from 3.2.0 to 3.3.0.  

If we wanted to stay compatible with matplotlib 3.2.0, an alternative would be to put a `try`/`except` clause to catch the `TypeError`.  I don't prefer that approach because that fix is less permanent.


StanczakDominik (approved):
Makes sense!

Issue #1341, opened 2021-11-23
Scipy nightly wheel in cron job fails to find compatible numpy version?


StanczakDominik:
(...) I'm having trouble with scipy due to a version conflict:

<details>
<summary>Details</summary>

```
🕙[ 19:20:44 ] ➜ tox -e py39-scipydev --notest
py39-scipydev create: /home/dominik/Code/github/PlasmaPy/PlasmaPy/.tox/py39-scipydev
py39-scipydev installdeps: :NIGHTLY:scipy, pytest-xdist, pytest-github-actions-annotate-failures
ERROR: invocation failed (exit code 1), logfile: /home/dominik/Code/github/PlasmaPy/PlasmaPy/.tox/py39-scipydev/log/py39-scipydev-1.log
===================================== log start ======================================
Looking in indexes: https://pypi.anaconda.org/scipy-wheels-nightly/simple
Collecting scipy
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/scipy/1.8.0.dev0%2B2083.8dd161d/scipy-1.8.0.dev0%2B2083.8dd161d-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (102.6 MB)
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/scipy/1.8.0.dev0%2B2082.296b550/scipy-1.8.0.dev0%2B2082.296b550-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (102.6 MB)
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/scipy/1.8.0.dev0%2B2060.9484b9d/scipy-1.8.0.dev0%2B2060.9484b9d-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (104.8 MB)
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/scipy/1.8.0.dev0%2B2047.048bac7/scipy-1.8.0.dev0%2B2047.048bac7-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (104.8 MB)
[...]

PR #1347, opened 2021-12-08
Drop scipy-dev tests for now

StanczakDominik:
Drops the failing scipy-dev cron job test from #1341.


namurphy (approved):
This looks good to me!  My current thinking is that we shouldn't put too much effort into setting up tests with scipy-dev since it's unlikely we'll uncover any problems with it, especially since we use SciPy less ubiquitously than other several of the other packages we're doing this for.  Thank you!

PR #1292, opened 2021-10-04
Add Python 3.10 tests to GitHub Actions

namurphy:
This pull request is to add Python 3.10 to our test suite.  Python 3.10 is due to be released approximately today 🥳 🎈 📈 but this PR will probably not work yet so we should wait to merge it.

- [x] I have added a changelog entry for this pull request.

- [x] If adding new functionality, I have added tests and
      docstrings.

- [x] I have fixed any newly failing tests.

Closes #1250.

PR #1350, opened 2021-12-08
Two stages of tests

StanczakDominik:
What this does can be best seen in the [Checks tab](https://github.com/PlasmaPy/PlasmaPy/pull/1350/checks). I split tests into two groups, one happening after the other, and moved code coverage to a single run in the latter. If I got it right, the second stage should run conditionally on the first one (linters and one quick run of tests, except the slow ones), and codecov should only trigger once.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -105,10 +105,48 @@ jobs:
       uses: codecov/codecov-action@v2.1.0
       with:
         file: ./coverage.xml
+  documentation:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    needs: initial-tests
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+
+        - name: Documentation
+          os: ubuntu-latest
+          python: 3.8
+          toxenv: build_docs
+          toxposargs: -q
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2.4.0
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2.3.0
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install Python dependencies
+      run: python -m pip install --progress-bar off --upgrade tox codecov
+    - name: Install language-pack-fr and tzdata
+      if: startsWith(matrix.name, 'Documentation')
+      run: sudo apt-get install graphviz pandoc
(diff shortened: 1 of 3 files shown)


StanczakDominik:
Also, seems like the minimal dependencies test was not actually running correctly: only now it started running in a single process context, raising the same issue as the recent https://github.com/conda-forge/plasmapy-feedstock/pull/16#issuecomment-983764940 . I have hopes for the rework in https://github.com/PlasmaPy/PlasmaPy/pull/1295 to fix that test.

Meanwhile `Python 3.8, slow tests with code coverage` is expected, but not coming, as it got renamed. `CI / Python 3.8, all tests, code coverage (pull_request)` is sort of doing the Faramir routine of not being exactly what the Steward expected, but still the next best thing.

namurphy:
I'm liking this!  Some quick thoughts...

* It'd be helpful to have the `CI / Documentation (pull_request)` be one of the initial tests, especially for PRs that change the docs only.  
* Eventually I'd prefer the initial round of tests to be done with Python 3.10 because of [improvements with error messages](https://docs.python.org/3/whatsnew/3.10.html#better-error-messages).  But that's for another day.  

Many thanks!

StanczakDominik:
> I'm liking this! Some quick thoughts...
> 
>     * It'd be helpful to have the `CI / Documentation (pull_request)` be one of the initial tests, especially for PRs that change the docs only.

I was hesitant to put those in there, because - if you take a look at the runtimes - the current Python3.9 tests and the linters run in one minute. The full test suites take 2-4 minutes, and the doc build takes five. Having it stall the full test suites seems meh.

Perhaps instead we could run it independently, completely in parallel, and together with linters? Does that sound better?

>     * Eventually I'd prefer the initial round of tests to be done with Python 3.10 because of [improvements with error messages](https://docs.python.org/3/whatsnew/3.10.html#better-error-messages).  But that's for another day.

I updated https://github.com/PlasmaPy/PlasmaPy/pull/1292/ to re-check whether it's still blowing up on us.

namurphy:
> Perhaps instead we could run it independently, completely in parallel, and together with linters? Does that sound better?

Yes!  That's a great idea.

StanczakDominik:
@namurphy does https://github.com/PlasmaPy/PlasmaPy/actions/runs/1579737561 look better, now? :)

namurphy (approved):
Yep, looks good to me!

PR #1351, opened 2021-12-08
Limit CODEOWNERS to pinging plasmapy-reviewers group

StanczakDominik:
As discussed in https://github.com/PlasmaPy/PlasmaPy/discussions/1348

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,3 +2,8 @@
 
 # any file, anywhere
 *                                                        @PlasmaPy/plasmapy-reviewers
+
+# any conftest.py, anywhere
+conftest.py                                              @StanczakDominik
+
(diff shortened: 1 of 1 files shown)


namurphy (approved):
LGTM!  It'd probably make sense to wait to merge until @rocco8773 and @pheuer have a chance to weigh in.

pheuer:
From #1348 
> In theory I like being auto-requested for any changes to the proton radiography module (actually, I'd like to add Thomson to that too). No one other than me submits anything there, but if they did...
> 
> Maybe the solution is to keep these, but only for specific cases where a specific "code owner" can be identified (like proton radiography), but otherwise just ping plasmapy-reviewers for everything else?

PR #1353, opened 2021-12-09
Enable sphinx-hoverxref for documentation

namurphy:
This PR enables [sphinx-hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/) for our docs.  From their documentation:

> sphinx-hoverxref is a Sphinx extension to show a floating window (tooltips or modal dialogues) on the cross references of the documentation embedding the content of the linked section on them. With sphinx-hoverxref, you don’t need to click a link to see what’s in there.

Many thanks to @StanczakDominik for bringing this up in the chat!

### DIFF
--- a/changelog/1353.doc.rst
+++ b/changelog/1353.doc.rst
@@ -1,0 +1,2 @@
+Enabled the `sphinx-hoverxref <https://sphinx-hoverxref.readthedocs.io>`_
+extension to Sphinx.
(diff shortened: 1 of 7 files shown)


StanczakDominik (approved):
This builds now, but I think it might not work yet, because...

> This extension requires a backend server to retrieve the tooltip content. Currently, only Read the Docs is supported as backend server. Take into account that your documentation has to be hosted on Read the Docs for this extension to work.

as per https://sphinx-hoverxref.readthedocs.io/en/latest/installation.html .

So it's probably best to try merging it and seeing what happens.

PR #1360, opened 2021-12-17
Create helper function to sort an iterable into a `dict` with `PhysicalType` objects as keys

namurphy:
This PR adds the module named `plasmapy.utils.units_helpers` (though we could change this) which contains a function `get_physical_type_dict`.  This function takes a `list` or `tuple` that is expected to contain `Quantity` instances and possibly other objects, and then create a `dict` that has `astropy.units.PhysicalType` instances as keys and the corresponding `Quantity` instances (or other objects with a physical type) as keys.  

I'm planning to use this for a `Normalizations` class, and it could also be used for perhaps a `UniformPlasma` class or our `PlasmaBlob` class.  An application could look vaguely like this:

```Python
class Normalizations:
    def __init__(self, *args):  # allow users to input a bunch of quantities of different physical types in arbitrary order
        physical_type_dict = get_physical_type_dict(args)
        # store normalizations coming from args here
```

We could possibly make this a private function, at least temporarily.  The implementation and interface for this will probably change whilst working on the normalizations classes in future PRs.

### DIFF
--- a/changelog/1360.trivial.rst
+++ b/changelog/1360.trivial.rst
@@ -1,0 +1,3 @@
+Added a helper function that takes an iterable and creates a `dict` with
+physical types as keys and the corresponding objects from that iterable
+as values.
--- a/changelog/1360.trivial.rst
+++ b/changelog/1360.trivial.rst
@@ -1,3 +1,4 @@
 Added a helper function that takes an iterable and creates a `dict` with
 physical types as keys and the corresponding objects from that iterable
-as values.
+as values. This change updates the minimum required version of Astropy
+to 4.3.1.
--- a/plasmapy/utils/tests/test_units_helpers.py
+++ b/plasmapy/utils/tests/test_units_helpers.py
@@ -1,3 +1,5 @@
+"""Tests of `plasmapy.utils.units_helpers`."""
+
 import astropy.units as u
 import pytest
 
--- a/plasmapy/utils/tests/test_units_helpers.py
+++ b/plasmapy/utils/tests/test_units_helpers.py
@@ -1,17 +1,72 @@
 import astropy.units as u
-import astropy.units.physical as physical
 import pytest
+
+from astropy.constants import c, m_e
+from collections import namedtuple
 
 from plasmapy.utils.units_helpers import get_physical_type_dict
 
 
-@pytest.fixture
-def pre_physical_type_dict():
-    units = (u.m, u.m ** -3, u.m * u.s)
-    return {unit.physical_type: 5 * unit for unit in units}
+def test_get_physical_type_dict_specific_example():
+    units = [u.m, u.m ** -3, u.m * u.s]
+    quantities = [5 * unit for unit in units]
+    expected = {quantity.unit.physical_type: quantity for quantity in quantities}
+    new_physical_type_dict = get_physical_type_dict(quantities)
+    assert new_physical_type_dict == expected
 
 
-def test_get_physical_type(pre_physical_type_dict):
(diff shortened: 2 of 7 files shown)


namurphy:
`PhysicalType` was added in Astropy 4.3.0, so this PR might necessitate a bump in the minimum supported version of Astropy.  I also expect to use `PhysicalType` in the normalizations classes also...probably.

plasmapy/utils/units_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+
+__all__ = ["get_physical_type_dict"]
+
+import astropy.units as u
+
+from numbers import Number
+from typing import Dict, Iterable
+
+
+def get_physical_type_dict(
```
StanczakDominik, line 13: I think I'd indeed prefer for it to be private, at least for now. It's always easier to turn things from private to public than the other way around.

plasmapy/utils/units_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+            continue
+
+        try:
+            physical_type = u.get_physical_type(obj)
+        except TypeError:
+            pass
+        else:
+            if physical_type in physical_types:
+                raise ValueError(f"Duplicate physical type: {physical_type}")
+            physical_types[physical_type] = obj
```
StanczakDominik, line 73: Glad this case is handled 👍🏻

plasmapy/utils/units_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+        if isinstance(obj, Number) and numbers_become_quantities:
+            obj = u.Quantity(obj, u.dimensionless_unscaled)
+
+        if only_quantities and not isinstance(obj, u.Quantity):
+            continue
+
+        try:
+            physical_type = u.get_physical_type(obj)
+        except TypeError:
+            pass
```
StanczakDominik, line 69: Can we add a test for this case? I would expect `"banana"` would fit, unless [`astropy.units.BED`](https://en.wikipedia.org/wiki/Banana_equivalent_dose) is a thing?

plasmapy/utils/tests/test_units_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+    expected = {quantity.unit.physical_type: quantity for quantity in quantities}
+    new_physical_type_dict = get_physical_type_dict(quantities)
+    assert new_physical_type_dict == expected
+
+
+velocity = u.get_physical_type("velocity")
+mass = u.get_physical_type("mass")
+dimensionless = u.get_physical_type(1)
+
+test_case = namedtuple("case", ["collection", "kwargs", "expected"])
```
StanczakDominik, line 22: Sure looks `dict`y, but I see what you're going for. 👍🏻

StanczakDominik (commented):
Go ahead and bump astropy right up! :)

changelog/1360.trivial.rst
```diff
@@ -0,0 +1,3 @@
+Added a helper function that takes an iterable and creates a `dict` with
+physical types as keys and the corresponding objects from that iterable
+as values.
```
namurphy, line 3:
```suggestion
Added a helper function that takes an iterable and creates a `dict` with
physical types as keys and the corresponding objects from that iterable
as values. This change updates the minimum required version of Astropy
to 4.3.1.
```

StanczakDominik (approved):
Given changelog fix to mention astropy 4.3.1 bump, LGTM!

PR #1364, opened 2021-12-27
Provide more flexibility in `CustomParticle` arguments

namurphy:
Since it's pretty hard to remember which of `charge` or `mass` comes first in the arguments to `CustomParticle`, this PR allows them to be placed in reversed order.

Now all of the following will work:
```Python
>>> from plasmapy.particles import CustomParticle
>>> mass = 1e-27 * u.kg
>>> charge =  1e-19 * u.C
>>> CustomParticle(charge, mass)
>>> CustomParticle(mass, charge)
>>> CustomParticle(mass)
>>> CustomParticle(charge)
```


plasmapy/particles/particle_class.py
```diff
@@ -2045,6 +2049,13 @@ class CustomParticle(AbstractPhysicalParticle):
     """
 
     def __init__(self, mass: u.kg = None, charge: (u.C, Real) = None, symbol=None):
+
+        if (
+            isinstance(mass, u.Quantity)
+            and u.get_physical_type(mass) == "electrical charge"
+        ):
+            charge, mass = mass, charge
+
```
rocco8773, line 2058: This seems like an odd approach to me...  Why not utilize `@validate_quantities` and enforce keyword only?  That seems like a more robust option and would add unit validation that does not currently exist here.

```python
    @validate_quantities(
        mass={"none_shall_pass": True},
        charge={"none_shall_pass": True},
    )
    def __init__(self, *, mass: u.kg = None, charge: u.C = None, symbol=None):
        ...
```

rocco8773: If you want to help with depreciation, then you could do something like....

```python
    @validate_quantities(
        mass={"none_shall_pass": True},
        charge={"none_shall_pass": True},
    )
    def __init__(self, *args, mass: u.kg = None, charge: u.C = None, symbol=None):

        if len(*args) != 0:
             # say you depreciation waning and take mass = args[0] and charge = args[1]
            ...
        ...
```

StanczakDominik:
I'll take a look tomorrow!

plasmapy/particles/particle_class.py
StanczakDominik: Have you tried Erik's suggested approach here, Nick? 

Suppose we live in an ideal world and we have MyPy for astropy Quantities. From that perspective, if we go with the current version, from the outside it looks like you'd be forcing a `charge` block into a `mass`-shaped hole and vice versa. That ideal-world unit-type-checker would rightfully complain - and anyone reading any code using this particular feature would likely be a smidge confused!

I kind of like the first version proposed by Erik here, but obviously deprecation period etc etc, so we'd have to go with the latter for now.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -1081,6 +1081,21 @@ def test_custom_particle_is_category(
 ]
 
 
+@pytest.mark.parametrize(
+    "charge, mass",
+    [
+        (1 * u.C, None),
+        (1 * u.C, 1 * u.kg),
```
StanczakDominik, line 1088: I think we could also add these cases:
```suggestion
        (1 * u.C, None),      # initialize a massless CustomParticle
        (None, 1 * u.C,),     # by symmetry
        (1 * u.C, 1 * u.kg),
        (1 * u.kg, 1 * u.C),
        (None, 1 * u.kg),    # initialize a chargeless and chonky CustomParticleBoi
        (1 * u.kg, None),    # by symmetry
```

StanczakDominik:
Just pointing out to scroll up and look at the comment branch from January ;)

namurphy:
Closing this since I did a workaround in the particle_input refactoring.

PR #1365, opened 2021-12-27
Add an interface function that creates physical particle objects

namurphy:
This PR adds an interface function that takes positional and keyword arguments, and creates a `Particle`, `CustomParticle`, or `ParticleList` instance as appropriate.

When writing this, I tried going through the logic of deciding which of `Particle`, `CustomParticle`, and `ParticleList` to instantiate. However, programming that logic quickly became messy and hard to read.  I settled on using a `try`/`except` block instead for simplicity because (as far as I can tell) there is only one overlap in types of inputs that can be passed to more than one of these classes.  

The one overlap is when no positional or keyword arguments are provided, which could either be accepted by `CustomParticle` to create a particle with `nan` mass & charge, or by `ParticleList` to create an empty instance.  I resolved this by having this cause a `TypeError`, but we might want to revisit this.

My motivation for creating this function is to concentrate particle creation logic in it when refactoring `particle_input`.  I'm planning to make this function private after making sure that the docstring builds correctly.  I think that using `Particle`, `CustomParticle`, or `ParticleList` explicitly would result in cleaner code in most (but not all) situations.


namurphy:
- [ ] Make sure documentation builds correctly
 - [ ] Make function private and remove it from `__init__.py` and `__all__`

plasmapy/particles/tests/test_factory.py
```diff
@@ -0,0 +1,55 @@
...
+    ([custom_particle], {}, custom_particle),
+    ([ParticleList(["p+", "e-"])], {}, ParticleList(["p+", "e-"])),
+]
+
+
+@pytest.mark.parametrize("args, kwargs, expected", test_cases)
+def test_physical_particle_factory(args, kwargs, expected):
+    result = _physical_particle_factory(*args, **kwargs)
+    assert result == expected
+    assert type(result) == type(expected)
```
StanczakDominik, line 42:
```suggestion
    assert isinstance(result, type(expected))
```
unless you want "hard" type equality?

StanczakDominik (approved):

namurphy:
Keeping this a private function for now, but we might want to revisit in the future whether or not this should be the main interface for creating particle objects.  If so, we'd probably want to rename it to something shorter like `particle`, though that could lead to confusion with `Particle`.  We'd probably also need to have more specific descriptions of parameters in the docstring.  

Also, thanks for the review!

PR #1366, opened 2021-12-27
Fix bug with proton equality while continuing `Particle.__init__` refactoring

### LINKED ISSUE #1363: Bug in `Particle.__eq__` for protons
I'm finding differences between protons created by `Particle("p+")` and `Particle("H", Z=1, mass_numb=1)`, as follows.

```Python
>>> from plasmapy.particles import Particle
>>> proton = Particle("p+")
>>> should_be_proton = Particle("H", Z=1, mass_numb=1)
  File "/home/namurphy/.miniconda/lib/python3.9/site-packages/plasmapy-0.0.0-py3.9.egg/plasmapy/particles/particle_class.py", line 665, in __eq__
    raise ParticleError(
plasmapy.particles.exceptions.ParticleError: p+ and p+ should be the same Particle, but have differing attributes.

The attributes of p+ are:

defaultdict(<class 'NoneType'>, {'symbol': 'p+', 'element': 'H', 'atomic number': 1, 'mass number': 1, 'element name': 'hydrogen', 'isotope': 'H-1', 'ion': 'p+', 'mass': <<class 'astropy.constants.codata2018.CODATA2018'> name='Proton mass' value=1.67262192369e-27 uncertainty=5.1e-37 unit='kg' reference='CODATA 2018'>, 'charge number': 1, 'periodic table': _PeriodicTable(group=1, period=1, block='s', category='nonmetal'), 'name': 'proton', 'spin': 0.5, 'class': 'baryon', 'lepton number': 0, 'baryon number': 1, 'half-life': <Quantity inf s>, 'antimatter': False, 'charge': <<class 'astropy.constants.codata2018.EMCODATA2018'> name='Electron charge' value=1.602176634e-19 uncertainty=0.0 unit='C' reference='CODATA 2018'>, 'isotope mass': None, 'isotopic abundance': None})

The attributes of p+ are:

defaultdict(<class 'NoneType'>, {'symbol': 'p+', 'element': 'H', 'isotope': 'H-1', 'ion': 'p+', 'mass number': 1, 'charge number': 1, 'atomic number': 1, 'element name': 'hydrogen', 'lepton number': 0, 'baryon number': 1, 'isotope mass': <Quantity 1.00782503 u>, 'isotopic abundance': 0.999885, 'half-life': <Quantity inf s>, 'mass': <<class 'astropy.constants.codata2018.CODATA2018'> name='Proton mass' value=1.67262192369e-27 uncertainty=5.1e-37 unit='kg' reference='CODATA 2018'>, 'periodic table': _PeriodicTable(group=1, period=1, block='s', category='nonmetal'), 'charge': <<class 'astropy.constants.codata2018.EMCODATA2018'> name='Electron charge' value=1.602176634e-19 uncertainty=0.0 unit='C' reference='CODATA 2018'>, 'name': None, 'spin': None, 'class': None, 'antimatter': None})
>>> proton._attributes == should_be_proton._attributes
False
>>> for key, val in proton._attributes.items():
...     new_val = should_be_proton._attributes[key]
...     if val != new_val:
...         print(f"{key}: {val}, {new_val}")
name: proton, None
spin: 0.5, None
class: baryon, None
antimatter: False, None
isotope mass: None, 1.00782503223 u
isotopic abundance: None, 0.999885
```
So far I've only found this discrepancy for protons, but it's good practice to test exhaustively for bugs, so the fix should include similar tests for protons, deuterons, tritons, and alphas.  Yay for special cases...

namurphy:
The purpose of this PR is to fix a bug that: `Particle("p+") == Particle("H", Z=1, mass_numb=1)` results in a `ParticleError` that the particles should be the same but have differing attributes.  Closes #1363.


namurphy:
The issue with the proton equality bug was that protons need to use the initialization for both special particles and for atoms.  When it was getting input like `Particle("H", Z=1, mass_numb=1)`, it was only doing the atom initialization.

I ended up doing more refactoring than I expected in this.  I think the only real change in functionality was the fixing of that bug.  The main changes are in `particle_class.py`, with changes in most other `plasmapy.particles` files being renaming of various things.

I also dropped the creation of `Particle.__name__` which didn't seem like it was needed.

plasmapy/particles/tests/test_atomic.py
```diff
@@ -626,10 +626,10 @@ def test_half_life():
 def test_half_life_unstable_isotopes():
     """Test that `half_life` returns `None` and raises an exception for
     all isotopes that do not yet have half-life data."""
-    for isotope in _isotopes.keys():
+    for isotope in _data_about_isotopes.keys():
```
namurphy, line 629:
```suggestion
    for isotope in _data_about_isotopes:
```
I think the `.keys()` is unnecessary?  Need to check...

plasmapy/particles/atomic.py
```diff
@@ -1122,7 +1124,7 @@ def periodic_table_category(argument: Union[str, Integral]) -> str:
             "symbol, or an integer representing its atomic number."
         )
     symbol = atomic_symbol(argument)
-    category = _elements[symbol]["category"]
+    category = _data_about_elements[symbol]["category"]
```
namurphy, line 1127: Perhaps rename to `_element_data_dict`, `_isotope_data_dict`, `_special_particle_data_dict`?

plasmapy/particles/particle_class.py
```diff
@@ -437,37 +469,70 @@ def _validate_arguments(argument, Z, mass_numb):
...
+            raise InvalidParticleError(errmsg) from exc
 
-    def _initialize_special_particle(self, particle_symbol, Z, mass_numb):
+        self._attributes["symbol"] = information_about_atom["symbol"]
+
+        for key in information_about_atom:
+            self._attributes[key] = information_about_atom[key]
+
+    def _assign_particle_attributes(self):
+        """Assign the attributes and"""
```
namurphy, line 505: Finish docstring...

plasmapy/particles/particle_class.py
```diff
@@ -421,9 +421,41 @@ class Particle(AbstractPhysicalParticle):
...
+
+        self._initialize_attributes_and_categories()
+        self._store_particle_identity()
+        self._assign_particle_attributes()
+        self._add_charge_information()
+        self._add_half_life_information()
+
+        delattr(self, "_inputs")
+
+        self.__name__ = self.__repr__()
```
namurphy, line 448:
```suggestion
```
I think I can get rid of this without adverse consequences?

plasmapy/particles/particle_class.py
```diff
@@ -421,9 +421,41 @@ class Particle(AbstractPhysicalParticle):
...
+        if isinstance(argument, Particle):
+            argument = argument.symbol
+
+        self._inputs = argument, mass_numb, Z
+
+        self._initialize_attributes_and_categories()
+        self._store_particle_identity()
+        self._assign_particle_attributes()
+        self._add_charge_information()
+        self._add_half_life_information()
```
namurphy, line 444: Make double underscores for names

namurphy:
- [ ] Look into `Particle.__name__` getting removed, but it is still (I think) being used in tests.  Need to make sure that the tests don't depend on `Particle.__name__`

plasmapy/particles/particle_class.py
```diff
@@ -421,9 +421,38 @@ class Particle(AbstractPhysicalParticle):
...
```
namurphy, line 443: A question that came up during live code review was about whether or not these methods should take arguments.  My motivation for not having arguments comes mostly from pages 40–41 of in Chapter 3 of _Clean Code_.  Also, while I was working on this, I found that the `Particle.__init__` was harder to refactor when there were arguments to these methods.  A worry about this approach is that not having arguments could sweep some of the story under the rug.  We essentially decided to try out this approach for the time being.  If this ends up leading to confusion later on, then we can change this later on.

plasmapy/particles/particle_class.py
namurphy: @rocco8773 — It turns out that double leading underscores causes the interpreter to engage in [name mangling](https://docs.python.org/3/tutorial/classes.html#private-variables).  

> Any identifier of the form `__spam` (at least two leading underscores, at most one trailing underscore) is textually replaced with `_classname__spam`, where `classname` is the current class name with leading underscore(s) stripped.

I had no idea about this before, but that does explain a few things I've noticed in the past that seemed weird at the time.  Because of that, I'm going to leave these with single underscores, which follows the convention to mark these attributes as not being part of the "public" API.

plasmapy/particles/parsing.py
```diff
@@ -42,7 +45,7 @@ def _create_alias_dicts(Particles: dict) -> (Dict[str, str], Dict[str, str]):
     case_sensitive_aliases = {}
     case_insensitive_aliases = {}
 
-    for symbol in Particles.keys():
+    for symbol in Particles:
```
rocco8773, line 48:
```suggestion
    for symbol in particles:
```

This should be refactored to `particles` to meet PEP8 standards, and to avoid easy confusion with `Particle`.

plasmapy/particles/tests/test_ionization_collection.py
```diff
@@ -35,16 +35,15 @@ def check_abundances_consistency(
         f"log_abundances.keys(): {log_abundances.keys()}"
     )
 
-    for element in abundances.keys():
-        abundance_from_abundances = abundances[element]
+    for element, abundance_from_abundances in abundances.items():
```
rocco8773, line 38: 👍 I love see these refactors!

plasmapy/particles/particle_class.py
```diff
@@ -423,9 +423,42 @@ class Particle(AbstractPhysicalParticle):
...
+        # defined in plasmapy/particles/__init__.py.
+
+        self.__name__ = self.__repr__()
+
+    def _initialize_attributes_and_categories(self):
+        """Create empty collections for attributes and categories."""
+        self._attributes = defaultdict(type(None))
+        self._categories = set()
+
+    def _validate_arguments(self):
```
rocco8773, line 458:
```suggestion
    def _validate_inputs(self):
```

I feel this is a more appropriate name since you're validating `__inputs`.

plasmapy/particles/particle_class.py
```diff
@@ -439,37 +472,72 @@ def _validate_arguments(argument, Z, mass_numb):
         if Z is not None and not isinstance(Z, Integral):
             raise TypeError("Z is not an integer.")
 
-    def _initialize_attrs_categories(self):
-        """Create empty collections for attributes & categories."""
+    def _store_particle_identity(self):
+        """Store the particle's symbol and identifying information."""
+        argument, mass_numb, Z = self.__inputs
+        self._validate_arguments()
```
rocco8773, line 478:
```suggestion
        self._validate_arguments()

        argument, mass_numb, Z = self.__inputs
```

Just flipping it around so you're not unpacking `__inputs` twice before raising errors.

plasmapy/particles/particle_class.py
```diff
@@ -439,37 +472,72 @@ def _validate_arguments(argument, Z, mass_numb):
...
+        symbol = _dealias_particle_aliases(argument)
```
rocco8773, line 479:
```suggestion
        symbol = parsing._dealias_particle_aliases(argument)
```

I'm curious if we can improve readability here by calling the parsing functionality directly from `parsing`.  I know you and I have different views on chunking up functionality, but when it's "too much" I feel like I'm trying to find a grain of sand in a buck of sand instead of pebble in a bucket of pebbles.  By using `parsing` directly, I feel we now at least have a clump of sand.

rocco8773 (approved):
LGTM! Thanks for the patience!

PR #1369, opened 2021-12-28
Clean up tests in particles

### LINKED ISSUE #750: Make test order deterministic so we can use `pytest-xdist` in `plasmapy.particles`
The order that tests are run in (what will soon be) `plasmapy.particles` is not deterministic.  Some of the functionality in that subpackage uses [set](https://docs.python.org/3/tutorial/datastructures.html#sets) operations, which do not preserve order.  Since the order of our tests sometimes depends on set operations, the order of tests changes.

Nominally, unit tests should be able to be run in any order.  However, I ran into a problem when trying to use [`pytest-xdist`](https://docs.pytest.org/en/3.0.1/xdist.html) to run tests in parallel, as this package currently requires test order to be deterministic when figuring out which tests to send to which processor.  Since our test order will only get bigger with time, it would be helpful to make our test order deterministic so that we have the capability of running tests in parallel.

The two possibilities that I can think of are:
 - Use [OrderedSets](https://pypi.org/project/ordered-set/) instead of regular sets (with the disadvantage that this would require adding another dependency to PlasmaPy)
 - Use sorting in the operations in which tests get selected (probably using `sorted`).  

The files that are affected include:

 - [ ] `test_ionization_state.py`
 - [ ] `test_ionization_states.py`
 - [ ] `test_parsing.py`
 - [ ] `test_particle_class.py`
 - [ ] `test_special_particles.py`

The quick way to test this is to install `pytest-xdist`, go to the tests directory, and then try running in the command prompt: 

```pytest -n 4 test_parsing.py```  

Thanks!

### LINKED ISSUE #1398: Make `IonizationState` and `IonizationStateCollection` tests independent of each other
I tried running our tests with the pytest extension pytest-randomly, and it turns out that many of the tests in `plasmapy/particles/test_ionization_state.py` and `plasmapy/particles/test_ionization_collection.py` fail when they are run in random order.  This is because the tests depend on each other, which is generally something to avoid, which I only learned recently.  We should modify these tests so that they don't depend on the order of test execution.  That is to say, we should make these tests completely independent of each other.  

I found this out by running these lines in the command line:

```bash
pip install pytest-randomly
pytest --randomly-seed=1235
```

These tests might be a good place to use pytest fixtures.

StanczakDominik:
Closes #750. 

- [ ] double check that it closes #1398.


plasmapy/particles/tests/test_exceptions.py
```diff
@@ -0,0 +1,1118 @@
...
+    tests_for_exceptions,
+)
+def test_named_tests_for_exceptions(
+    test_name, tested_object, args, kwargs, expected_exception
+):
+    """
+    Test that appropriate exceptions are raised for inappropriate inputs
+    to `IonizationState` or `IonizationStateCollection`
+    """
+    print(test_name)  # TODO find better ways for this
```
namurphy, line 250:
```suggestion
```
Maybe drop this line?

namurphy: Or perhaps an explicit `pytest.fail(f"problem with {test_name}")` at the end of the `with pytest.raises` block?

plasmapy/particles/tests/test_exceptions.py
```diff
@@ -0,0 +1,1118 @@
...
+        tested_object(*args, **kwargs)
+
+    assert expected_exception == exc_info.type
+
+    # TODO tbh given how ugly this is I don't think we should even be doing this check
+    if hasattr(exc_info, "expected_warning"):
+        for expected_warning, recorded_warning in zip(
+            exc_info.expected_warning, exc_info.list
+        ):
+            assert expected_warning == recorded_warning.category
```
StanczakDominik, line 261:
```suggestion
```

changelog/1369.trivial.rst
```diff
@@ -0,0 +1 @@
+Refactor particle tests.
```
namurphy, line 1:
```suggestion
Refactored tests in `plasmapy.particles`.
```
Minor optional rewording.

namurphy (approved):
LGTM!  Lots of lines in the diff due to reformatting, but the changes were pretty straightforward.

StanczakDominik:
This should now be good to go! The docs build failure is unrelated (same as https://github.com/pydata/xarray/issues/6250).

PR #1371, opened 2021-12-29
Create generic decorator that can be adapted to process arguments based on annotations

### LINKED ISSUE #1079: Create a factory for decorators that process arguments based on annotations
So...refactoring `@particle_input` in #1057 is turning out to be a pain, which is an indication that the approach I'm taking is probably less than optimal.  The biggest source of headaches for me is that I keep getting confused due to the decorator logic becoming intertwined with the particle input logic.  It would be helpful to have a way to separate the decorator logic from the "business" logic.  

One possibility would be to create a decorator factory for the generic pattern where arguments and outputs are processed based on annotations and keyword arguments to be supplied to the decorator.  The decorator factory would accept a callable that accepts `(args_to_function, kwargs_to_function, function_annotations, args_to_decorator, kwargs_to_decorator)` and returns (`new_args_to_function`, `new_kwargs_to_function`, and `new_function_annotations`).  There are a couple of possibilities for how to design this, so it'd be worth thinking more about that ahead of time (a habit I should probably get into...).

namurphy:
This PR is to create a generic class-based decorator that can be subclassed into different decorators that process arguments based on annotations.  My plan is to subclass this in the refactoring of `@particle_input` in #1057, and actually much of the content within it has been extracted from #1057.  There was also inspiration from `@astropy.units.quantity_input`.  I'd still like to compare this to the implementation of our other decorators. I haven't been able to find an equivalent to this in the pythoniverse, but I'm still wondering if there's something out there.

So far I'm working on this as a prototype.  I'm not sure if this should ultimately go in PlasmaPy, but if this is implemented cleanly and documented well, then this could simplify the creation of new decorators and perhaps the implementation of some of our existing decorators.  

My hope is that this and #1365 will make refactoring `@particle_input` a lot more straightforward.  It'd be great if there was something out there in the pythoniverse already like this, but haven't been able to find what I was looking for.

I'm also wondering about an implementation like...

```Python
def process_arguments_based_on_annotations(args, kwargs, annotations):
    return new_args, new_kwargs, new_annotations

def create_decorator(process_arguments):
    ...
```
Closes #1079.

### DIFF
--- a/plasmapy/utils/decorators/__init__.py
+++ b/plasmapy/utils/decorators/__init__.py
@@ -14,6 +14,7 @@ __all__ = [
     "CheckUnits",
     "CheckValues",
     "ValidateQuantities",
+    "GenericDecorator",
 ]
 
 from plasmapy.utils.decorators.checks import (
@@ -26,5 +27,6 @@ from plasmapy.utils.decorators.checks import (
 )
 from plasmapy.utils.decorators.converter import angular_freq_to_hz
 from plasmapy.utils.decorators.deprecation import deprecated
+from plasmapy.utils.decorators.generic import GenericDecorator
 from plasmapy.utils.decorators.helpers import modify_docstring, preserve_signature
 from plasmapy.utils.decorators.validators import validate_quantities, ValidateQuantities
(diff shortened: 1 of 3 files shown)


namurphy:
Closing this (at least for the time being) because what I was thinking might be in one of these two packages.  I just need to take some time to read and understand them to make sure.
 - [wrapt](https://wrapt.readthedocs.io/en/latest/index.html)
 - [decorate](https://github.com/micheles/decorator)

PR #1372, opened 2021-12-30
Add a fortnightly test for a nitpicky documentation build

namurphy:
This adds a cron test that uses the [nitpicky](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpicky) option for our documentation build.  This will make Sphinx warn about all references where the target cannot be found.  The warnings will be turned into failures because  the `-W` flag for `sphinx-build` is being used too.  

This test currently fails because we have numerous references that are missing targets, for example in-line code snippets that have single back ticks but should have double back ticks. Once we do get this to pass, I'd like to turn on this option for our regular CI to catch missing targets right away.  

I think we might need to merge this before we can check that it's set up correctly.

### DIFF
--- a/.github/workflows/fortnightly.yml
+++ b/.github/workflows/fortnightly.yml
@@ -41,6 +41,11 @@ jobs:
           python: 3.8
           toxenv: py38-xarraydev
 
+        - name: Documentation build (nitpicky)
+          os: ubunti-latest
+          python: 3.9
+          toxenv: build_docs_nitpicky
+
     steps:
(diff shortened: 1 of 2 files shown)


namurphy:
Oh yeah...I forgot that there were a bunch of errors associated with words to the right of parameter names.  There are some settings that need to be set before it makes sense to merge this.

```Python
    """
    Parameters
    ----------
    n_e : words here are generally treated as missing references
    """
```

PR #1373, opened 2021-12-30
Describe more `make` environments for doc build in doc guide

namurphy:
I was checking out `docs/Makefile` as one often does to reminisce about the past year (?), and I noticed that there were a few `make` environment things that I hadn't included in the documentation guide previously.  This adds a sentence each on `make html-nonb` and `make clean-api` in the documentation guide.


StanczakDominik (approved):
:+1:

PR #1375, opened 2021-12-31
Minor docstring fixes in `thomson.py` and `langmuir.py`

namurphy:
More reST fixes that I started after doing a `make nitpicky` doc build and looking at the errors, and then taking a few minutes to make a few other docstring edits.  I don't think that there are any substantive changes here at all.

PR #1376, opened 2021-12-31
Minor docstring fixes in `plasmapy.formulary`

namurphy:
This follows up on #1375 with some more minor reST and other docstring fixes in a few files in `plasmapy.formulary`.  Again, I don't think that there's anything substantive here.

 - [ ] Check the modified docstrings to see that they render okay

### DIFF
--- a/plasmapy/formulary/mathematics.py
+++ b/plasmapy/formulary/mathematics.py
@@ -59,8 +59,7 @@ def Fermi_integral(
     Warnings
     --------
     At present this function is limited to relatively small arguments
-    due to limitations in the `mpmath` package's implementation of
-    `~mpmath.polylog`.
+    due to limitations in `mpmath` implementation of `~mpmath.polylog`.
 
     Examples
     --------
--- a/plasmapy/formulary/mathematics.py
+++ b/plasmapy/formulary/mathematics.py
@@ -36,7 +36,7 @@ def Fermi_integral(
         If the argument is a `~astropy.units.Quantity` but is not
         dimensionless.
 
-    ValueError
+    `ValueError`
         If the argument is not entirely finite.
 
     Notes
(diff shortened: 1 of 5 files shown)


plasmapy/formulary/parameters.py
```diff
@@ -131,10 +131,10 @@ def mass_density(
         equivalent).  If ``density`` is a mass density, then it will be passed
         through and returned without modification.
 
-    particle : `~plasmapy.particles.Particle`
+    particle : `~plasmapy.particles.particle_class.Particle`
```
StanczakDominik, line 134: Doesn't the previous version resolve through `__init__.py`?

StanczakDominik (approved):
Looks good! One question. Could go without a changelog here.

plasmapy/formulary/parameters.py
namurphy: The way we have it set up, it needs the full namespace to resolve.  I'm not actually sure why that is, but was what motivated me to set up the reST substitutions.  Thanks for the review!

plasmapy/formulary/mathematics.py
```diff
@@ -56,9 +56,10 @@ def Fermi_integral(
     .. math::
         F_j (x) = -Li_{j+1}\left(-e^{x}\right)
 
-    Warning: at present this function is limited to relatively small
-    arguments due to limitations in the `~mpmath` package's
-    implementation of `~mpmath.polylog`.
+    Warnings
+    --------
+    At present this function is limited to relatively small arguments
+    due to limitations in `mpmath` implementation of `~mpmath.polylog`.
```
rocco8773, line 62: This is weird.  I would have used a warnings admonition here `.. warning::``, but the `numpydoc` "Warnings" section is rendering just like that admonition.

plasmapy/formulary/parameters.py
rocco8773: You need the full path path because, even though `Particle` is exposed to the `plasmapy.particles` namespace, only one stub file is created for `Particle` at the full path `plasmapy.particles.particle_class.Particle`.  The original behavior of `automodapi` created two stub files for each location, but our version does not.

If you look the `plasmapy.particles` documentation page, then you'll notice the `Particle` link is actually point to `plasmapy.particles.particle_class.Particles`.

PR #1378, opened 2022-01-04
Add `ParticleList.is_category`

namurphy:
This PR adds the `ParticleList.is_category` method that is analogous to `Particle.is_category`.  I realized the need for this while figuring out how to make `@particle_input` work with `ParticleList` instances.

Currently the docstring for the method refers to the `Particle.is_category` docstring.  I did this to make the docstring for `Particle.is_category` the definitive one, and to avoid the possibility that the docstrings for the two methods could diverge.  I'm open to including the parameters list in the docstring of the new method if preferred.


StanczakDominik (approved):
Looks good, at a glance!

PR #1387, opened 2022-01-12
Add temperature to glossary

namurphy:
Because of the nuances with how we accept temperature as an argument to formulary functions, I was thinking that it'd be helpful to have a description of temperature that we could easily link to in docstrings by using ``` :term:`temperature` ```.  I chose the glossary since it seems the most sensible place to define a term, but I'm open to putting this elsewhere as long as there's an easy way to link to the information.  I want to think about the wording a little more carefully since using eV as a unit of temperature is a point of confusion for newcomers to the field (at least it was for me back in the day).

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -86,6 +86,18 @@ Glossary
       dimensional |Quantity| and a real number is provided, then the
       real number is often assumed to have the appropriate SI units.
 
+   temperature
+      Most functions in PlasmaPy accept temperature :math:`T` as a
+      `~astropy.units.Quantity` with units of temperature (e.g., kelvin)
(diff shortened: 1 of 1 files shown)


namurphy:
Some possibilities @rocco8773 mentioned...

 - We could have a separate glossary for things like temperature where there are nuances.  This could potentially include angular frequencies.
 - We could even have a separate glossary for variable names (like ``B``, ``T_e``) and how they should be configured.

namurphy:
Or maybe...``` :term:`temperature-like` ``` could be added to the glossary to represent `Quantity` objects that can be converted into temperature with the `u.temperature_energy()` equivalency.

StanczakDominik (approved):

PR #1309, opened 2021-10-15
Add a function to create CustomParticles from molecule symbols.

Quettle:
This new functions is useful to build objects representing molecules from user input or when parsing input files The mass of the created `CustomParticle` is simply the sum of the masses of its components. If the input is a simple particle, then a `Particle` object is created.

- [x] I have added a changelog entry for this pull request.

- [x] If adding new functionality, I have added tests and
      docstrings.

- [x] I have fixed any newly failing tests.

PR #1397, opened 2022-01-21
Changed CustomParticle __repr__ to display the symbol if present

Quettle:

      docstrings.

This PR is adding the display of the symbol in `CustomParticle.__repr__`. This make it more convenient to work with the new molecule functionality from PR  #1309

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -2072,8 +2072,18 @@ class CustomParticle(AbstractPhysicalParticle):
         >>> custom_particle = CustomParticle(mass=1.2e-26 * u.kg, charge=9.2e-19 * u.C)
         >>> repr(custom_particle)
         'CustomParticle(mass=1.2...e-26 kg, charge=9.2...e-19 C)'
+
+        If present, the symbol is displayed as well.
+
+        >>> custom_particle = CustomParticle(mass=4.21e-25 * u.kg, charge=1.6e-19 * u.C, symbol="I2+")
+        >>> repr(custom_particle)
+        'CustomParticle(mass=4.21e-25 kg, charge=1.6e-19 C, symbol=I2+)'
         """
-        return f"CustomParticle(mass={self.mass}, charge={self.charge})"
+        return (
+            f"CustomParticle(mass={self.mass}, charge={self.charge})"
+            if self._symbol is None
+            else f"CustomParticle(mass={self.mass}, charge={self.charge}, symbol={self.symbol})"
+        )
 
     @property
     def json_dict(self) -> dict:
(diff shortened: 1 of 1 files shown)


plasmapy/particles/particle_class.py
```diff
@@ -2210,12 +2220,12 @@ def symbol(self) -> str:
 
         If no symbol was defined, then return the value given by `repr`.
         """
-        return self._symbol
+        return repr(self) if self._symbol is None else self._symbol
```
Quettle, line 2223: This changes also guarantees that the repr displayed when symbol is accessed  while no symbol was provided before is still valid even if the charge and mass have been modified.

plasmapy/particles/particle_class.py
```diff
@@ -2244,7 +2254,7 @@ def __hash__(self) -> int:
         Allow use of `hash` so that a |CustomParticle| instance may be used
         as a key in a `dict`.
         """
-        return hash((self.__repr__(), self.symbol))
+        return hash(self.__repr__())
```
namurphy, line 2257:
```suggestion
        return hash(self.symbol)
```
I'm wondering if it would make sense to have the hash be the same as the symbol.  That way, if a `CustomParticle` with a symbol were used as a `dict` key, then it should be possible for the symbol to also be used to access the dict item.  What would your opinion on this be?

```pycon
>>> dilithium = CustomParticle(symbol="dilithium")
>>> d = {dilithium: "..."}
>>> d[dilithium]
>>> d["dilithium"]  # access dict item with string
```

plasmapy/particles/particle_class.py
namurphy: Yeah, I think this is a good idea.

namurphy (commented):
Thank you for doing this!  I have a few quick suggestions, including one about `hash` that's more of a question.  Besides that, this looks pretty straightforward, so we should be able to merge this quickly.

plasmapy/particles/particle_class.py
namurphy: That's a really good point.  If we created two different `CustomParticle` instances with the symbol `"I2"`, and put them both in the same`dict`, then using `"I2"` as a key would be ambiguous.  Let's skip this suggestion.  Thanks!

namurphy (approved):
Looks good to me!  Were there any more changes that you wanted to make?  If not, I'll merge this in the next day or two.  Thank you again!

Quettle:
> Looks good to me! Were there any more changes that you wanted to make? If not, I'll merge this in the next day or two. Thank you again!

I don't see anything else. Thank you again!

PR #1399, opened 2022-01-22
Remove warnings from `CustomParticle` and `DimensionlessParticle` for `nan` attributes

namurphy:
This PR changes `CustomParticle` and `DimensionlessParticle` so that they no longer emit a warning when they are instantiated without either a charge or a mass.  The warnings had stated that that particular attribute was not provided and thus was given a value of `numpy.nan` in the appropriate units.

I'm removing this since it's not uncommon for me to only need to define either a mass or a charge, and the warnings are getting kind of annoying.  If the warnings do end up being helpful to others, though, I'm open to keeping them.  

I'm not quite sure what sort of changelog entry makes the most sense, but I'm going to go with `trivial` for now.

### DIFF
--- a/changelog/1399.trivial.rst
+++ b/changelog/1399.trivial.rst
@@ -1,0 +1,3 @@
+|CustomParticle| and |DimensionlessParticle| no longer emit a warning
+when the charge and/or mass is not provided and got assigned a value of
+|nan| in the appropriate units.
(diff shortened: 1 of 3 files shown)


StanczakDominik (approved):
I remember having doubts about these warnings in the first place. Glad they're going away! :)

PR #1410, opened 2022-01-28
[Docs] adds sphinx-codeautolink to link code samples to API docs

jeandet:
This PR adds sphinx-codeautolink which basically scan all code samples and tries to adds references to known API docs.
I also added matplotlib intersphinx entry since it seems it was missing.

This is a small change which I believe makes browsing code sample even more user friendly.


StanczakDominik (dismissed):
Awesome, I was gonna add it myself! :D I can confirm that this works (given, say, the first block in https://plasmapy--1410.org.readthedocs.build/en/1410/particles/functional.html#symbols-and-names). I just wonder why https://github.com/PlasmaPy/PlasmaPy/runs/4978380187?check_suite_focus=true fails. Looks like `tox` didn't pull... oh yeah, it uses `setup.cfg` because we have duplication (that we should probably get to, anyway...) in there. Could you add the package in there as well, and we can get this a speedy merge?

A changelog entry would be neat as well :)

jeandet:
@StanczakDominik sure, that should be ok now.

jeandet:
Weird doc still fails on CI, not sure why.

namurphy:
Awesome, thank you for doing this!  I'm fully in support of making it easier for people to access and read the source code.

I took a quick look at the doc failure and didn't immediately understand what was causing the problem, but I'll try to take a look at it later.

namurphy (approved):

PR #1416, opened 2022-02-03
[PROTO] Adding Cython framework for lite-functions (and more)

rocco8773:
This is currently a prototype to see how we can incorporate `Cython` into `plasmapy`.  I'm particularly curious in how...

1. To write `cython` code. (Using lite-functions as the prototype)
2. How to package and distribute with `cython` code.
3. How to develop with `cython` code.

----

Using `cython` in `plasmapy` means developers/contributors have an extra step to get off the ground, because, like c, `cython` code needs to be compiled.  `Cython` looks very similar to Python and can be written in either `.py` files or `.pyx` files.  `.pyx` files are specifically `Cython` files.  Once this is done, the code needs to be cythonized (i.e. built/compile).  In this process c code is generated from the source code and then the c code is compiled into share object files.  Python can then import directly from these shared object files.  However, these shared object files are platform specific and this is why we can not package with pre-compiled code.  Luckily, these shared object files are automatically generate with `pip` installs, so there shouldn't be any extra steps for end users.  For developers/contributors, we don't typically `pip` install so we have manually create the shared object files.  This is done by...

1. Navigating to the root directory for `plasmapy`.
2. Executing the following command
    `python setup.py build_ext --inplace`

This will create the shared object files in your repository.  However, these are not under version control (i.e. excluded by `.gitignore`), so whenever `cython` code is modified or you switch to a new branch you'll likely need to re-compile.

To assist in cleaning up the compiled code I did modify the `setup.py` so that following commands will remove files generated by `python setup.py build_ext --inplace`.

    python setup.py clean
    python setup.py clean --all

If you want to see what file are deleted before cleaning, then you can use...

    python setup.py clean --dry-run


setup.py
```diff
@@ -1,13 +1,73 @@
...
+        zip_safe=False,
+        include_dirs=[np.get_include()],
+        ext_modules=cythonize(build_extensions(), language_level=3),
+    )
+
+    # allow for cleaning of --inplace c and shared object files generated
+    # by:
+    #     python setup.py build_ext --inplace
+    #
+    # this get executed with the command
```
namurphy, line 52:
```suggestion
    # this gets executed with the command
```

namurphy (commented):
Looking good!  My suggested changes are all minor, and you're free to ignore them if you so choose.

The main thing I'm still wondering about, which I think we talked about yesterday, is on whether or not it specifically needs to be a numpy array as input, or if it could be a real or int.  I think I remember you saying that yes, it does need to be a numpy array for the input.

If you're looking to eventually merge this prototype, we'd want to include the additional required steps in the contributor guide section on installing PlasmaPy for development.  

Thank you for doing this!

setup.py
```diff
@@ -1,13 +1,73 @@
...
+    #       python setup.py clean
+    #
+    if "clean" in sys.argv:
```
StanczakDominik, line 55: `clean` is already a `setup.py` command and I would be wary of reimplementing it. Have you tried playing around with https://stackoverflow.com/questions/30359216/compile-a-cython-project-and-clean and see whether the default `python setup.py clean --all` does the trick?

People are generally moving away from `setup.py` in the python space in general... but that's probably an issue for another day.

.github/workflows/testing.yml
```diff
@@ -144,7 +144,8 @@ jobs:
     - name: Install requirements
       run: |
         pip install --progress-bar off --upgrade pip
-        pip install --progress-bar off setuptools numpy wheel setuptools_scm twine
+        pip install --progress-bar off twine
+        pip install --progress-bar off -r requirements/build.txt
```
StanczakDominik, line 148: good cleanup! 👍🏻

StanczakDominik (approved):
LGTM, one potential idea to make this introduce less code!

setup.py
rocco8773: I'm not overriding `python setup.py clean`, this still runs as expected.  I'm adding an additional setup after the `clean` command is executed.  The `clean` command does not know about, and thus can not remove, `--inplace` shared object files.  All this is doing is seeing if the `clean` command is specified, letting the normal `clean` run, and then our `setup.py` removes the `--inplace` files.

StanczakDominik: Oooooh okay then, gotcha. 👍🏻

PR #1418, opened 2022-02-07
Add numba.(n)jit decorators to coverage exclude

### LINKED ISSUE #1386: Numba jitted functions are tracted by `coverage`
As stated in [comment (PR1145)](https://github.com/PlasmaPy/PlasmaPy/pull/1145/files#r782341157), when testing the c-complied code generated by `numba` is executed instead of the original python function.  This leads to a gap in the coverage report, so we should work out scheme to handle this.

> Note: Since `thermal_speed_lite` is decorated by `numba.njit`, the code that is actually covered by tests is the c-compiled version and thus `coverage` does not recognize if the python code gets hit.
> 
> For now I'm updating the tests such that the following is essentially done...
> 
> ```python
> lite = thermal_speed_lite(**kwargs)
> pylite = thermal_speed_lite.py_func(**kwargs)
> assert lite == pylite
> ```
> 
> The call to `thermal_speed_lite.py_func` calls the original python function and, thus, is calculated in the coverage report. An alternative to this is to add a scheme to the `excluded_lines` in the `[coverage:report]` settings of `setup.cfg` such that anything decorated with `jit` or `njit` is ignored.

StanczakDominik:
Closes #1386.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -239,3 +239,7 @@ exclude_lines =
     ModuleNotFoundError
     @vectorize
     @numba.vectorize
+    @numba.jit
+    @jit
+    @numba.njit
+    @njit


namurphy (approved):
LGTM!  Happy to get this merged quickly.

PR #1427, opened 2022-02-17
Improving code coverage for null point finder by adding unit test for the case when vector values are entered by the user

haman80:

      docstrings.

### DIFF
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -591,16 +591,6 @@ def _trilinear_analysis(vspace, cell):
         that a grid cell may contain more than one nullpoint.
     """
 
-    # Helper Function
-    def is_close(a, b):
-        arr = np.isclose(a, b, atol=_ATOL)
-        if type(arr) == np.bool_:
-            return arr
-        res = True
-        for b in arr:
-            res = res and b
-        return res
-
     # Critical Cell Corners
     f000 = cell
     f111 = [cell[0] + 1, cell[1] + 1, cell[2] + 1]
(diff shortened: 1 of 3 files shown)


plasmapy/analysis/nullpoint.py
```diff
@@ -978,16 +925,16 @@ def bound(epoint):
...
+    # # Check on the Surfaces
+    # for p in BxByEndpoints:
+    #     if np.linalg.norm(tlApprox(p[0], p[1], p[2])) < _EQUALITY_ATOL:
+    #         return True
+    # for p in BxBzEndpoints:
+    #     if np.linalg.norm(tlApprox(p[0], p[1], p[2])) < _EQUALITY_ATOL:
+    #         return True
+    # for p in ByBzEndpoints:
+    #     if np.linalg.norm(tlApprox(p[0], p[1], p[2])) < _EQUALITY_ATOL:
+    #         return True
```
namurphy, line 937: It seems reasonable to me to leave this as a comment for the moment because we think we don't need it, but aren't sure.

namurphy (commented):
It's looking good!  I suggested a few minor suggestions, and I think after that we can merge it.  

One thing I'm wondering about is if we could [parametrize](https://docs.plasmapy.org/en/latest/development/testing_guide.html#test-independence-and-parametrization) the `test_null_point_find1` through `test_null_point_find8` functions.  That could be either addressed here, or we could create a follow-up issue or pull request and address it later.  My usual inclination is to have multiple smaller/focused pull requests since that makes them easier to review.  

Thank you again for doing this!

namurphy (approved):

PR #1430, opened 2022-02-21
Update `gyroradius` FutureWarning and tests

### LINKED ISSUE #1420: Numpy isnan bug in plasmapy.formulary.parameters.gyroradius()
I recently experienced an error during an ion gyroradius calculation: `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`. The problem seems to stem from the line:

`if not np.isnan(T_i):`
 
in

`plasmapy.formulary.parameters.gyroradius()`

It looks like you're expecting `T_i` to be a scalar but in my case it was an array. 

To reproduce, you can run `plasmapy.formulary.parameters.gyroradius(B, 'p', T_i=T_i)` with a `T_i` that's an array.

rocco8773:
This PR focuses on...

1. Updating the conditional for generating the `PlasmaPyFutureWarning` to resolve issue #1420.
2. Refactors the tests for `gyroradius` by combining the existing test function and test class into one test class, as well as parametrizing all tests.

closes #1420

### DIFF
--- a/changelog/1430.bugfix.rst
+++ b/changelog/1430.bugfix.rst
@@ -1,0 +1,5 @@
+For `~plasmapy.formulary.parameters.gyroradius`, updated the default
+keyword arguments and conditional for issuing the
+`~plasmapy.utils.exceptions.PlasmaPyFutureWarning`.  This addresses the
+incorrect behavior where a `ValueError` is raised if an array is passed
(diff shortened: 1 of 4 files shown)


namurphy (approved):
LGTM!  I only had a few slight suggestions, but otherwise it looks like we can get this in quickly.

Issue #1433, opened 2022-02-22
[EPIC] Decompose `plasmapy.formulary.parameters`


rocco8773:
As mentioned in PR #1312, `plasmapy.formulary.parameters` has become extremely monolithic and thus it is difficult to maintain and developed.  This also applies to its associated tests.  This issue will serve as an epic to monitor the progress of the decomposition process.

Option 2 outlined in PR #1312 will be adopted for this process.

> **Option 2**
> 
> Decompose the functionality in `parameters.py` into separate files like option 1, but keep these files in the `formulary` directory. Additionally, keep a `parameters.py` with proper imports to maintain backwards compatibility, but `parameters.py` would not contain any defined functionality.
> 
> ```shell
> plasmapy
> +-- formulary
> |   +-- __init__.py
> |   +-- ...
> |   +-- dimensionless.py
> |   +-- frequencies.py
> |   +-- lengths.py
> |   +-- misc.py
> |   +-- parameters.py
> |   +-- speeds.py
> |   +-- ...
> ```
> 
> I feel this approach is more clear on where to place future functionality. I'd just place some kind of `FutureWarning` on imports from `parameters.py` so users have time to move away from it.

----

## Items to complete:

- [x] Create and populated `plasmapy.formulary.dimensionless`.
      PR #1444
- [x] Create and populated `plasmapy.formulary.frequencies`.
      PR #1439
- [x] Create and populated `plasmapy.formulary.lengths`.
      PR #1434 
- [x] Create and populated `plasmapy.formulary.misc`.
      PR #1453  
- [x] Create and populated `plasmapy.formulary.speeds`.
      PR #1448 

For each of the above steps the associated tests need to be decomposed/migrated, `plasmapy/formulary/parameters.py` needs to be updated to issue a `PlasamPyFutureWarning`, and `plasmapy/formulary/__init__.py` needs to be updated so functionality is populated from the decomposed modules.

----

I think the best way to create the `PlasmaPyFutureWarning` in `parameters.py` is to add a bit of code the looks like...

```python
from plasmapy.formulary.lengths import inertial_length as _inertial_length
from plasmapy.utils.decorators import deprecated
from plasmapy.utils.exceptions PlasmaPyFutureWarning

inertial_length = deprecated(
    since="0.7.0",
    warning_type=PlasmaPyFutureWarning,
    message=(
        "The inertial_length() function has been moved to plasmapy.formulary.lengths."
        "  Update your import to get rid of this warning."
[...]

PR #1434, opened 2022-02-22
Create and populate `plasmapy.formulary.lengths`

rocco8773:
This PR moves length plasma parameters from `plasmapy.formulary.parameters` to `plasmapy.formulary.lengths` in accordance with issue #1433.


plasmapy/formulary/lengths.py
```diff
@@ -0,0 +1,366 @@
+"""Functions to calculated fundamental plasma length parameters."""
```
namurphy, line 1:
```suggestion
"""Functions to calculate fundamental plasma length parameters."""
```

docs/notebooks/dispersion/two_fluid_dispersion.ipynb
```diff
@@ -132,7 +133,7 @@
     "    ),\n",
     "    \"wci\": pfp.gyrofrequency(inputs[\"B\"], inputs[\"ion\"]),\n",
     "}\n",
-    "params[\"lpe\"] = pfp.inertial_length(params[\"n_e\"], \"e-\")\n",
+    "params[\"lpe\"] = inertial_length(params[\"n_e\"], \"e-\")\n",
     "params[\"wpe\"] = pfp.plasma_frequency(params[\"n_e\"], \"e-\")"
```
namurphy, line 137: It's a little inconsistent here in that there's a namespace prefix for `plasma_frequency` but not `inertial_length`.  But, if this is going to be changed when `frequencies.py` gets extracted from `parameters.py`, then there's no need to change that in this PR.

namurphy (approved):
LGTM!  One minor typo fix, and one comment that can probably wait until a subsequent PR.  I also didn't look into the details of the different functions, since I think you copied them over without changes.  Thank you for doing this!

docs/notebooks/dispersion/two_fluid_dispersion.ipynb
rocco8773: Yes, as I do the decomposition I plan to update this notebook with direct imports.

PR #1437, opened 2022-02-23
Configure `sphinx-hoverxref`

### LINKED ISSUE #1421: Fully define configuration values for sphinx-hoverxref in `docs/conf.py`
Previously we had added the [sphinx-hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/index.html) extension, but we did not fully define the configuration variables that go into `docs/conf.py`.  The only one we defined so far was `hoverxref_auto_ref` which only affects the `:ref:` role. We still need to add in the relevant [sphinx-hoverxref configuration variables](https://sphinx-hoverxref.readthedocs.io/en/latest/configuration.html).

rocco8773:
Fully configure the sphinx extension package `sphinx-hoverxref` in our `conf.py`.

Closes #1421

### DIFF
--- a/changelog/1437.doc.rst
+++ b/changelog/1437.doc.rst
@@ -1,0 +1,1 @@
+Configured the |Sphinx| extension `sphinx-hoverxref`.
--- a/changelog/1437.doc.rst
+++ b/changelog/1437.doc.rst
@@ -1,1 +1,1 @@
-Configured the |Sphinx| extension `sphinx-hoverxref`.
+Configured the Sphinx_ extension `sphinx-hoverxref`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -240,19 +240,19 @@ hoverxref_role_types = {
     #
     # roles with py domain
     "attr": "tooltip",
-    "class": "modal",
+    "class": "tooltip",
     "const": "tooltip",
     "data": "tooltip",
-    "exc": "modal",
+    "exc": "tooltip",
     "func": "tooltip",
     "meth": "tooltip",
-    "mod": "modal",
+    "mod": "tooltip",
     "obj": "tooltip",
     #
     # roles with std domain
     "confval": "tooltip",
     "hoverxref": "tooltip",
-    "ref": "modal",
+    "ref": "tooltip",
     "term": "tooltip",
 }
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -133,13 +133,6 @@ hoverxref_intersphinx = [
     "sphinx",
     "numba",
 ]
-hoverxref_intersphinx_types = {
-    "astropy": {
-        "ref": "tooltip",
-        "class": "tooltip",
-        "obj": "tooltip",
-    },
-}
 
 autoclass_content = "both"
 
@@ -206,7 +199,7 @@ exclude_patterns = [
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 
-default_role = "py:obj"
+default_role = "obj"
 
 # Customizations for make linkcheck using regular expressions
 linkcheck_allowed_redirects = {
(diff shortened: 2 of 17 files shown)


changelog/1437.doc.rst
```diff
@@ -0,0 +1 @@
+Configured the |Sphinx| extension `sphinx-hoverxref`.
```
namurphy, line 1:
```suggestion
Configured the Sphinx_ extension `sphinx-hoverxref`.
```

docs/conf.py
```diff
@@ -214,6 +225,33 @@
 pygments_style = "default"
 
 hoverxref_auto_ref = True
+hoverxref_mathjax = True
+hoverxref_sphinxtabs = True
+
+# hoverxref has to applied to these
```
namurphy, line 231:
```suggestion
# hoverxref has to be applied to these
```

rocco8773:
@namurphy I've discovered that `sphinx-hoverxref` will not generate a tooltip if you reference an external ojbect like `` `astropy.units.Quantity` ``, but will generate the tooltip if you use the correct role like `` :class:`astropy.units.Quantity` ``.  Since we predominately use the former, most external reference will not get a tooltip, and I'm not about to go do that refactoring.

I have opened an issue at `sphinx-hoverxref` to see what they say. https://github.com/readthedocs/sphinx-hoverxref/issues/170

docs/conf.py
```diff
@@ -213,7 +226,45 @@
 # Use a code highlighting style that meets the WCAG AA contrast standard
 pygments_style = "default"
 
+if os.environ.get("READTHEDOCS"):
+    # Building on Read the Docs
+    hoverxref_api_host = "https://readthedocs.org"
+
+    if os.environ.get("PROXIED_API_ENDPOINT"):
+        # Use the proxied API endpoint
+        hoverxref_api_host = "/_"
```
namurphy, line 235: [optional] I'm wondering if it would be worth updating the comment to describe what a proxied API endpoint is, since this might be unfamiliar teminology for many of us (including me), and the sphinx-hoverxref source code and docs doesn't really describe what this is doing.

docs/conf.py
```diff
@@ -107,19 +107,32 @@
 # in other packages. When mappings are removed or added, please update
 # the section in docs/doc_guide.rst on references to other packages.
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3", None),
+    "readthedocs": ("https://docs.readthedocs.io/en/stable/", None),
```
namurphy, line 110: Huh, I didn't know that we could connect to readthedocs via intersphinx.  Neat!

namurphy (approved):
LGTM!  I agree with using the tooltip option for everything rather than modal.  I have one minor optional suggestion, but nothing to prevent me from approving it.  Many thanks!

docs/conf.py
```diff
@@ -213,7 +226,45 @@
 # Use a code highlighting style that meets the WCAG AA contrast standard
 pygments_style = "default"
 
+if os.environ.get("READTHEDOCS"):
```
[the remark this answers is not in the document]
rocco8773: Yeah, readthedocs has several build variables...  They're good to know about even if I don't completely understand their use.

https://docs.readthedocs.io/en/stable/builds.html#id1

docs/conf.py
rocco8773: I honestly don't understand it myself, I just copied it over from `sphinx-hoverxref`'s `conf.py`.

https://github.com/readthedocs/sphinx-hoverxref/blob/003b84fee48262f1a969c8143e63c177bd98aa26/docs/conf.py#L81-L95

humitos: This is a RTD thing to avoid a CSRF block. Without this, the request is done to readthedocs.org. When using "_/" the request is done to the same domain where the documentation is hosted.

Under RTD "_/" has a special meaning. It means to hit the Django app instead of the "hosting app".

Not sure if that clarifies. I'm on the phone, but I can expand later if needed.

humitos:
Hi all! I opened this PR https://github.com/readthedocs/sphinx-hoverxref/pull/171 that should solve the problem that you reported. If you have some time, please give it a try installing that version and let me know if it works as you expected! 🙏🏼

PR #1440, opened 2022-02-24
Privatize some `plasmapy.particles` modules

namurphy:
This PR makes the following changes to module names:
 - `plasmapy.particles.parsing` → `plasmapy.particles._parsing`
 - `plasmapy.particles.elements` → `plasmapy.particles._elements`
 - `plasmapy.particles.isotopes` → `plasmapy.particles._isotopes`
 - `plasmapy.particles.special_particles` → `plasmapy.particles._special_particles`

I also made the following changes:
 - Dropped the initial underscore of objects within these modules, as suggested by @rocco8773
 - Changed the names of some variables to match PEP 8 conventions
 - Imported these modules directly rather than the contents of these modules
 

      docstrings.


plasmapy/particles/_elements.py
```diff
@@ -24,7 +26,7 @@ class _PeriodicTable:
     category: str
 
 
-def _element_obj_hook(obj):
+def element_obj_hook(obj):
     if "unit" in obj:
         return obj["value"] * u.Unit(obj["unit"])
     return obj
```
namurphy, line 32: - [ ] Create an issue about adding a docstring

plasmapy/particles/isotopes.py
```diff
@@ -24,15 +25,15 @@
 #     json.dump(_Isotopes, f, default=plasma_default, indent=2)
 
 
-def _isotope_obj_hook(obj):
+def isotope_obj_hook(obj):
     """Provide an `object_hook` designed for `json.load` and `json.loads`."""
     if "unit" in obj:
         return obj["value"] * u.Unit(obj["unit"])
     return obj
```
namurphy, line 32: - [ ] Create an issue about adding a docstring

plasmapy/particles/_special_particles.py
```diff
@@ -293,9 +291,3 @@ def _create_Particles_dict() -> Dict[str, dict]:
     "anti_nu_mu": "nu_mu",
     "anti_nu_tau": "nu_tau",
 }
-
-if __name__ == "__main__":  # coverage: ignore
-    from pprint import pprint
-
-    print("Particles:")
-    pprint(_data_about_special_particles)
```
namurphy, line 301: - [ ] Add a changelog entry for this

plasmapy/particles/tests/test_atomic.py
```diff
@@ -17,6 +17,7 @@
 from plasmapy.utils.exceptions import PlasmaPyFutureWarning
 from plasmapy.utils.pytest_helpers import run_test
 
+from .._isotopes import data_about_isotopes
 from ..atomic import (
```
namurphy, line 21: Change to absolute imports

namurphy:
This PR changes a lot of lines, but nearly all of the changes are renaming modules, variables, & imports.  I think the only substantive change is the removal of an `if __name__ == "__main__"` block at the end of one file that printed out special particle properties.

namurphy:
In this PR, I ended up removing the leading underscores from the objects within the newly private modules (i.e., `_qwbvdc` →  `qwbvdc`).  After that I put these objects in `__all__`, but ended up getting a bunch of documentation build errors after doing that.  I ended up changing `__all__` to an empty list for each of these modules, which ended up removing the doc build errors.

plasmapy/particles/__init__.py
```diff
@@ -5,6 +5,7 @@
 # __all__ will be auto populated below
 __all__ = []
```
rocco8773, line 6: I"m a little perplexed at the moment, since the `autodoc` is still trying to generate documentation for `_elements`, `_isotopes`, `_parsing`, and `_special_particles`.  (see https://plasmapy--1440.org.readthedocs.build/en/1440/particles/index.html#variables-attributes)  I'm confused on two fronts: (1) those modules, except for `_special_particles`, aren't exposed to the `plasmapy.particles` namespace and (2) the `autodoc` is interpreting them as attributes instead of modules.  ... I'll need to do some investigation.

rocco8773: I think this would be a working solution...  You'd need to replace the code block for auto-populating `__all__` with

```python
# auto populate __all__
for name, obj in list(globals().items()):
    if inspect.ismodule(obj):
        continue

    if not (name.startswith("__") or name.endswith("__")):
        __all__.append(name)
__all__.sort()

del name, obj
```

This will just ensure modules (dirs and files) are not added to `__all__`.

plasmapy/particles/__init__.py
```diff
@@ -5,6 +5,7 @@
...
 
+from plasmapy.particles._special_particles import particle_zoo
```
rocco8773, line 8: The `particle_zoo` documentation is a bit odd, since it inherits the docstring from `ParticleZoo`.  Basically it's a variable with a class like documentation, minus the the documentation for the methods.  I would consider exposing `ParticleZoo` to the `plasmapy.particles` namespace and defining `particle_zoo` in `plasmapy.particles` with an appropriate new docstring.  I say define `particle_zoo` in `plasmapy.particles` because if you define it in `_special_particles` the docstring won't (at least in my experience) carry over to the instance in `plasmapy.particles`.

I'd see this looking something like the following in `plasampy/particles/__init__.py`

```python
from plasmapy.particles._special_particles import ParticleZoo

particle_zoo = ParticleZoo()
"""
Data container for representations of "special" particles, like leptons, baryons, etc.

An instance of `~plasmapy.particles._special_particles.ParticleZoo`.
"""
```

I do think for completeness, if `particle_zoo` is in our documentation then it must be accompanied with `ParticleZoo`.

---

I would also consider having `ParticleZoo` inherit from `collections.UserDict` so `particle_zoo` would become the `taxonomy_dict`.

plasmapy/particles/_special_particles.py
```diff
@@ -28,14 +28,13 @@ class _ParticleZooClass:
...
     Examples
     --------
-    >>> ParticleZoo = _ParticleZooClass()
+    >>> ParticleZoo = ParticleZoo()
     >>> 'e-' in ParticleZoo.leptons
     True
     >>> 'nu_e' in ParticleZoo.antineutrinos
     False
     >>> 'mu+' in ParticleZoo.antiparticles
     True
```
rocco8773, line 37:
```suggestion
    >>> par_zoo = ParticleZoo()
    >>> 'e-' in par_zoo.leptons
    True
    >>> 'nu_e' in par_zoo.antineutrinos
    False
    >>> 'mu+' in par_zoo.antiparticles
    True
```

I only chose `par_zoo` so there wouldn't be any name shadowing with `particle_zoo` defined later in the file.

rocco8773:
> In this PR, I ended up removing the leading underscores from the objects within the newly private modules (i.e., `_qwbvdc` → `qwbvdc`). After that I put these objects in `__all__`, but ended up getting a bunch of documentation build errors after doing that. I ended up changing `__all__` to an empty list for each of these modules, which ended up removing the doc build errors.

Did you ever push a build with `__all__` populated?  I tested this out locally and I didn't get any build errors, and I wouldn't expect to get build errors.  If there were build errors, then it is likely due to some other structural issue than having the names in `__all__`.

plasmapy/particles/__init__.py
namurphy: Thanks, this looks helpful!  I'll try this out hopefully later today.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -10,6 +10,8 @@
 from astropy.constants import c, e, m_e, m_n, m_p
 
 from plasmapy.particles import json_load_particle, json_loads_particle, molecule
+from plasmapy.particles._isotopes import data_about_isotopes
+from plasmapy.particles._special_particles import particle_zoo
```
rocco8773, line 14: If we adopt what I mentioned earlier about defining `particle_zoo` in the particles `__init__.py` then this import would need to reflect that.

plasmapy/particles/__init__.py
rocco8773: Since `particle_zoo` is an instance of a class and thus it has a `__doc__` attribute (unlike most variables), think we can still define it in `_special_particles`.  I think it would then have to look like...

**In `__init__.py`**

```python
from plasmapy.particles._special_particles import particle_zoo, ParticleZoo
```

**In `_special_particles.py`**

```python
particle_zoo = ParticleZoo()
particle_zoo.__doc__ = """
Data container for representations of "special" particles, like leptons, baryons, etc.

An instance of `~plasmapy.particles._special_particles.ParticleZoo`.
"""
```

rocco8773 (approved):
LGTM!!

PR #1444, opened 2022-02-25
Migrate dimensionless functionality from `plasmapy.formulary.parameters` to `dimensionless`

rocco8773:
This PR moves dimensionless plasma parameters from `plasmapy.formulary.parameters` to `plasmapy.formulary.dimensionless` in accordance with issue #1433.

PR #1446, opened 2022-02-26
Patch for #1444

rocco8773:
After merging PR #1444 we started getting an `ImportError` from `plasampy.formulary`.  This was coming from the import of `from plasmapy.formulary.parameters import *`.  The reason for this is two fold: (1) the `plasmapy.formulary.dimensionless` pkg existed before the functionality migration done in PR #1444 and thus it contained functionality not originally contained in `plasmapy.formulary.parameters`; and (2) the auto-population scheme of `__all__`, `__aliases__`, and `__lite_funcs__` was adding the names of these "non-parameters" functions to the `parameters` dunders.  This resulted in `from plasmapy.formulary.parameters import *` trying to import functionality not contained in the namespace.

image

### DIFF
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -73,7 +73,7 @@ for name in (
     "Re_",
     "Reynolds_number",
     "Rm_",
-):
+):  # coverage: ignore
     try:
         __aliases__.remove(name)
     except ValueError:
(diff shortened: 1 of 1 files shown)


namurphy (commented):
Thanks for catching this!  I have some minor optional suggestions that I'll trust your judgement on.  I'm happy to approve this even without these changes.

rocco8773:
Thank for the quick review! I wasn't expecting anyone to be around to do this over the weekend.

PR #1449, opened 2022-02-27
[PROTO] Syncing CI dependencies with pkg dependencies

rocco8773:
will write later...


rocco8773:
Way out of date and no longer applicable.

PR #1455, opened 2022-02-28
Update docstring for `molecule`

namurphy:
This PR updates the docstrings for `plasmapy.particles.particle_class.molecule` and associated private functions.  In particular, this expands/edits some parameter descriptions, and makes a few reST edits.  

I also adopted a minor refactoring suggested by Sourcery, which was to explicitly raise an exception from another exception.

### DIFF
--- a/plasmapy/particles/_parsing.py
+++ b/plasmapy/particles/_parsing.py
@@ -477,34 +477,33 @@ def parse_and_check_molecule_input(argument: str, Z: Integral = None):
 
     Parameters
     ----------
-    argument : 'str'
+    argument : `str`
         The molecule symbol to be parsed.
 
-    Z : 'Integral', optional
+    Z : integer, optional
         The provided charge number.
 
     Returns
     -------
-    elements_dict : 'dict'
-        A dictionary with identified element symbols as keys and amount of each as values.
-        The molecule symbol stripped of the charge.
-        The integer charge.
+    elements_dict : `dict`
+        A dictionary with identified element symbols as keys and amount
+        of each as values.
 
-    molecule_info : 'str'
+    molecule_info : `str`
         The molecule symbol stripped of its charge.
 
-    Z : 'int'
-        The molecule charge.
+    Z : `int`
+        The charge number of the molecule.
 
     Raises
     ------
-    'InvalidParticleError'
-        If the Symbol couldn't be parsed.
+    `InvalidParticleError`
+        If the symbol couldn't be parsed.
 
     Warns
     -----
-    `ParticleWarning`
-        If The charge is given both as an argument and in the symbol.
+    : `ParticleWarning`
+        If the charge is given both as an argument and in the symbol.
     """
     molecule_info, z_from_arg = extract_charge(argument)
     if not re.fullmatch(r"(?:[A-Z][a-z]?\d*)+", molecule_info):
--- a/plasmapy/particles/_parsing.py
+++ b/plasmapy/particles/_parsing.py
@@ -498,7 +498,7 @@ def parse_and_check_molecule_input(argument: str, Z: Integral = None):
     Raises
     ------
     `InvalidParticleError`
-        If the symbol couldn't be parsed.
+        If ``argument`` could not be parsed as a molecule.
 
     Warns
     -----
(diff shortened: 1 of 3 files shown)


plasmapy/particles/_parsing.py
```diff
@@ -477,34 +477,33 @@ def parse_and_check_molecule_input(argument: str, Z: Integral = None):
 
     Parameters
     ----------
-    argument : 'str'
+    argument : `str`
         The molecule symbol to be parsed.
 
-    Z : 'Integral', optional
+    Z : integer, optional
```
rocco8773, line 483:
```suggestion
    Z : `int`, optional
```

Why not this?  I'm sure you purposefully used integer over int, but I'm missing the reason.

plasmapy/particles/_parsing.py
```diff
@@ -477,34 +477,33 @@ def parse_and_check_molecule_input(argument: str, Z: Integral = None):
...
 
     Returns
     -------
-    elements_dict : 'dict'
-        A dictionary with identified element symbols as keys and amount of each as values.
-        The molecule symbol stripped of the charge.
-        The integer charge.
+    elements_dict : `dict`
+        A dictionary with identified element symbols as keys and amount
+        of each as values.
```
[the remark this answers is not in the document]
namurphy: Yes, I've definitely been having trouble with the wording on this one!

rocco8773 (changes requested):
For the most part it looks good.  I just had a few minor suggestions.

plasmapy/particles/particle_class.py
```diff
@@ -2317,22 +2323,23 @@ def molecule(
...
         mass = 0 * u.kg
         for element_symbol, amount in element_dict.items():
             try:
                 element = Particle(element_symbol)
-            except ParticleError as e:
+            except ParticleError as exc:
                 raise InvalidParticleError(
                     f"Could not identify {element_symbol}."
-                ) from e
+                ) from exc
```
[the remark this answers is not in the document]
namurphy: Good catch!

plasmapy/particles/_parsing.py
rocco8773: Use `numpy.integer` then.  If I remember correctly, that represents all integer dtypes and the Python integer.

rocco8773: I was wrong.  It only represents all numpy integer dtypes.  You'd need to do something like `Union[int, numpy.integer]`.

namurphy:
```Python
>>> isinstance(1, np.integer)
False
```
Hm...doesn't include the Python integer, unfortunately...

rocco8773: Yep.... :( It would have to be this... I always forget this...

```Python
>>> isinstance(5, (int, numpy.integer))
True
>>> isinstance(numpy.int32(5), (int, numpy.integer))
True
```

namurphy: I'll go with `int` for now since that's our standard practice.  I'm wondering if we could do a ``` :term:`integer` ``` here and link to a glossary entry, but I'm pretending to have learned my lesson about scope creep.

plasmapy/particles/particle_class.py
```diff
@@ -2317,22 +2323,23 @@ def molecule(
...
         mass = 0 * u.kg
         for element_symbol, amount in element_dict.items():
             try:
                 element = Particle(element_symbol)
-            except ParticleError as e:
+            except ParticleError as exc2:
                 raise InvalidParticleError(
                     f"Could not identify {element_symbol}."
-                ) from e
+                ) from exc
```
rocco8773, line 2337:
```suggestion
                ) from exc2
```

plasmapy/particles/particle_class.py
```diff
@@ -2317,22 +2323,23 @@ def molecule(
...
             if not element.is_category("element"):
                 raise InvalidParticleError(
                     f"Molecule symbol contains a particle that is not an element: {element.symbol}"
-                )
+                ) from exc2
```
rocco8773, line 2341:
```suggestion
                ) from exc
```

rocco8773 (changes requested):

rocco8773 (approved):

PR #1456, opened 2022-02-28
Make `Z` and `mass_numb` parameters for `Particle` keyword-only

namurphy:
In a lot of functions in `plasmapy.particles`, we've made the `Z` and `mass_numb` parameters keyword-only because of the potential for errors.  This PR makes these parameters keyword-only for `Particle` also.  This is a breaking change.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -413,10 +413,16 @@ class Particle(AbstractPhysicalParticle):
     def __init__(
         self,
         argument: ParticleLike,
-        *,
+        *args,
         mass_numb: Integral = None,
         Z: Integral = None,
     ):
+
+        if args:
+            raise TypeError(
+                "The parameters mass_numb and Z to Particle are now "
+                "keyword-only [e.g., Particle('H', mass_numb=2, Z=1)]."
+            )
 
         # If argument is a Particle instance, then construct a new
         # Particle instance for the same particle.
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -413,12 +413,14 @@ class Particle(AbstractPhysicalParticle):
     def __init__(
         self,
         argument: ParticleLike,
-        *args,
+        *_,
         mass_numb: Integral = None,
         Z: Integral = None,
     ):
 
-        if args:
+        # TODO: Remove the following block during or after the 0.9.0 release
+
+        if _:  # coverage: ignore
             raise TypeError(
                 "The parameters mass_numb and Z to Particle are now "
                 "keyword-only [e.g., Particle('H', mass_numb=2, Z=1)]."
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -420,7 +420,7 @@ class Particle(AbstractPhysicalParticle):
 
         # TODO: Remove the following block during or after the 0.9.0 release
 
-        if _:  # coverage: ignore
+        if _:
             raise TypeError(
                 "The parameters mass_numb and Z to Particle are now "
                 "keyword-only [e.g., Particle('H', mass_numb=2, Z=1)]."
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -234,10 +234,10 @@ class Particle(AbstractPhysicalParticle):
         integer representing the atomic number of an element; or a
         |Particle| instance.
 
-    mass_numb : `int`, optional
+    mass_numb : `int`, optional, keyword-only
         The mass number of an isotope or nuclide.
 
-    Z : `int`, optional
+    Z : `int`, optional, keyword-only
         The charge number of the particle.
 
     Raises
@@ -413,6 +413,7 @@ class Particle(AbstractPhysicalParticle):
     def __init__(
         self,
         argument: ParticleLike,
+        *,
         mass_numb: Integral = None,
         Z: Integral = None,
     ):
--- a/plasmapy/particles/tests/test_particle_class.py
+++ b/plasmapy/particles/tests/test_particle_class.py
@@ -620,6 +620,8 @@ test_Particle_error_table = [
     (["Fe 25+"], {}, ".recombine(8.2)", TypeError),
     (["e-"], {}, ".ionize()", InvalidElementError),
     (["e+"], {}, ".recombine()", InvalidElementError),
+    (["H", 1], {}, "", TypeError),
+    (["H", 1, 1], {}, "", TypeError),
 ]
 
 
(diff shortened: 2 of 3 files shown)


namurphy:
A possibility that I'm considering is to have an explicit error message that says that `Z` and `mass_numb` have been made keyword-only.  Something like...
```Python
def __init__(argument, *args, Z, mass_numb):
    if args:
        raise TypeError("...")
```

rocco8773:
If you do... (note the use of `*` instead of `*args`)

```python
def __init__(argument, *, Z, mass_numb):
    ...
```

then you can rely on Python to throw the TypeError.  This looks like...

image

Not sure you need something more specific than that.

namurphy:
Yep, that matches what I was originally thinking.  I was considering the alternative because it would give a more detailed error message that specifically mentions the API change and the names of the parameters that were changed.  I think I'll sleep on it and bring it up at the community meeting tomorrow.

rocco8773:
I get that.  ... When I can I like to rely on Python's internals since it's less work for me, less pkg maintenance long term, and lower risk for functionality getting out of wack.  ... At the moment I'd support either option though.

namurphy:
Yep, those are all good points.  Since this is an API change, perhaps the best middle ground would be to include the explicit error message in the `0.8.0` release, and then remove it in the `0.9.0` or `0.10.0`...

plasmapy/particles/particle_class.py
```diff
@@ -413,10 +413,19 @@ class Particle(AbstractPhysicalParticle):
...
         self,
         argument: ParticleLike,
+        *_,
         mass_numb: Integral = None,
         Z: Integral = None,
     ):
 
+        # TODO: Remove the following block during or after the 0.9.0 release
+
+        if _:  # coverage: ignore
```
rocco8773, line 423: I don't think this should be ignored for tests.  Since we're "mimicking" Python's default behavior for `foo(x, *, bar=None)`, we should test that it is behaving that way.  ... I can't really solidify my arguments for this, but my gut is telling me it should be tested.

namurphy: Sure thing!  My conscience was telling me I probably ought to have tested that block anyway...

 - [ ] Do this ↑

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -584,56 +584,60 @@ def test_Particle_equivalent_cases(equivalent_particles):
...
+    (["D"], {}, ".recombine()", ChargeError),
+    (["Fe 26+"], {}, ".ionize()", InvalidIonError),
+    (["Fe 6+"], {}, ".ionize(-1)", ValueError),
+    (["Fe 25+"], {}, ".recombine(0)", ValueError),
+    (["Fe 6+"], {}, ".ionize(4.6)", TypeError),
+    (["Fe 25+"], {}, ".recombine(8.2)", TypeError),
+    (["e-"], {}, ".ionize()", InvalidElementError),
+    (["e+"], {}, ".recombine()", InvalidElementError),
+    (["H", 1], {}, "", TypeError),
+    (["H", 1, 1], {}, "", TypeError),
```
namurphy, line 624: These are the two new lines

plasmapy/particles/particle_class.py
```diff
@@ -234,10 +234,10 @@ class Particle(AbstractPhysicalParticle):
         integer representing the atomic number of an element; or a
         |Particle| instance.
 
-    mass_numb : `int`, optional
+    mass_numb : `int`, optional, keyword-only
         The mass number of an isotope or nuclide.
 
-    Z : `int`, optional
+    Z : `int`, optional, keyword-only
```
rocco8773, line 240: I'm not sure how I feel about using `keyword-only` here.  I have never seen any documentation use this term, or practice.  From Python to `numpy` to `sphinx` and beyond, no one calls out keyword-only arguments.

I accept the use of `optional` only because it is [documented by `numpydoc`](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).  Additionally, `optional` implies `keyword-only`.

I'm not convinced this is a practice we should pickup, but I'm open to hearing arguments for it.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -584,56 +584,60 @@ def test_Particle_equivalent_cases(equivalent_particles):
...
+    "args, kwargs, attribute, exception", test_Particle_error_table
+)
+def test_Particle_errors(args, kwargs, attribute, exception):
     """
     Test that the appropriate exceptions are raised during the creation
     and use of a `~plasmapy.particles.Particle` object.
     """
     with pytest.raises(exception):
-        exec(f"Particle(arg, **kwargs){attribute}")
+        exec(f"Particle(*args, **kwargs){attribute}")
```
rocco8773, line 637:
```suggestion
        par = Particle(*args, **kwargs)
        par_attr = getattr(par, attr_name)
```

As discussed in today's community meeting, this would be a more secure implementation.

plasmapy/particles/tests/test_particle_class.py
rocco8773, line 624: Even though using a list for `args` works here, positional arguments are tuples behind the scenes.  I think it just looks better on us to keep that standard.  That is,

```python
test_Particle_error_table = [
    (("a",), {}, "", InvalidParticleError),
    ...,
]
```

rocco8773 (approved):

plasmapy/particles/tests/test_particle_class.py
namurphy: I'll work on this in a follow-up pull request.  This would require more than two lines because some of the things being tested are properties (`".standard_atomic_weight"`), and some are methods that have their own arguments (`".ionize()"`).

plasmapy/particles/particle_class.py
namurphy: I may have originally learned this from Astropy's documentation (e.g., [`Supergalactic`](https://docs.astropy.org/en/stable/api/astropy.coordinates.builtin_frames.Supergalactic.html)).  Also, an argument can be keyword-only and not optional, or optional and not keyword-only.
```Python
def keyword_only_and_required(x, *, y): ...
def optional_and_not_keyword_only(x, y=None): ...
```
I think it's useful to specify that it's keyword-only for the reasons that explicit is better than implicit.  Eventually it might be useful to have "keyword-only" automatically link to the corresponding term in our glossary.  I also noticed that there was a discussion on https://github.com/numpy/numpydoc/issues/330 about this.

plasmapy/particles/tests/test_particle_class.py
namurphy: I'm going to be restructuring this table in my follow-up PR, so I'll make this change then.

PR #1468, opened 2022-03-05
Merge `v0.7.x` into `stable`, which I forgot to do while releasing it

StanczakDominik:
There's a bunch of merge conflicts in here that I think I might need a hand with...

This helps with https://docs.plasmapy.org/en/stable/ still pointing to 0.6.


StanczakDominik:
Funnily enough the merge conflicts were simpler on GitHub than locally.

namurphy (approved):
Thank you for doing this!

I'm wondering if the linter test failures are because of checks that we started doing after our 0.7.0 release.  If that's the case, then they shouldn't be applied to 0.7.0 code, and we could ignore them.  

The read the docs failure so far looks like a concurrency problem rather than a doc build failure.

Issue #1402, opened 2022-01-25
[EPIC] Swept Langmuir Overhaul


rocco8773:
The issues outlines and summarizes the work being done on the swept Langmuir overhaul.  This overhaul will happen in several phase: Phase 1 will be the development of tools in `plasmapy.analysis` and Phase 2 will be the incorporation of the `plasmapy.analysis` tools into an `xarray` framework replacing functionality in `plasmapy.diagnostics`.

GitHub Project: https://github.com/PlasmaPy/PlasmaPy/projects/19 (not well maintained)

## Functionality:

- [x] Create `plasmapy.analysis.swept_langmuir`
    * PR #889
- [x] Floating potential
    * PR #889
    * `plasmapy.analysis.swept_langmuir.find_floating_potential`
    * Via `I = 0` crossing point
- [ ] Ion-Saturation Current (`Isat`)
    * PR #1469 
    * `plasmapy.analysis.swept_langmuir.find_ion_saturation_current`
    * curve fit ion-saturation region of the I-V curve
- [ ] Plasma Potential via `dI/dV`
    * PR #897
    * `plasmapy.analysis.swept_langmuir.find_plasma_potential_didv`
    * Estimates the floating potential from the peak of the first-derivative of the I-V curve (i.e. dI/dV)
- [ ] Electron Temperature from left side fit of `dI/dV`
- [ ] Electron Temperature from linear fit of `log(I)` vs `V`

## Notebooks

- [x] [`find_floating_potential`](https://docs.plasmapy.org/en/latest/notebooks/analysis/swept_langmuir/find_floating_potential.html)
    * PR #889
- [ ] `find_ion_saturation_current`
- [ ] `find_plasma_potential_didv`

PR #1469, opened 2022-03-07
Swept Langmuir | Ion-Saturation Current

rocco8773:
This is part of the Swept Langmuir analysis overhaul, issue #1402 .  This PR introduces `plasmapy.analysis.swept_langmuir.find_ion_saturation_current` for fitting the ion-saturation current region of an IV-curve.  The linear portion of the fit, representing the ion-saturation current, is returned to the user.

The associated notebook will be created in a followup PR.


plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,228 @@
...
+    fit_type: str = "exp_plus_linear",
+    current_bound: float = None,
+    voltage_bound: float = None,
+) -> Tuple[ffuncs.Linear, ISatExtras]:
+    """
+    Determines the ion-saturation current (:math:`I_{sat}`) for a given
+    current-voltage (IV) curve obtained from a swept Langmuir probe.
+    The current collected by a Langmuir probe reaches ion-saturation
+    when the probe is sufficiently biased so the influx of electrons is
+    completely repelled leading to only the collection of ions.  (For
```
namurphy, line 55:
```suggestion
    The current collected by a Langmuir probe reaches ion-saturation
    when the probe is sufficiently biased so the influx of electrons is
    completely repelled, which leads to only the collection of ions.  (For
```
[optional]

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,228 @@
...
+        (should be in volts).
+
+    current: `numpy.ndarray`
+        1-D numpy array of probe current (should be in amperes)
+        corresponding to the ``voltage`` array.
+
+    fit_type: `str`
+        The type of curve (:term:`fit-function`) to be fitted to the
+        Langmuir trace, valid options are listed below.
+        (DEFAULT ``"exp_plus_linear"``)
```
namurphy, line 74:
```suggestion
        Defaults to ``"exp_plus_linear"``.
```
One thing I learned somewhat recently is that using [all caps](https://en.wikipedia.org/wiki/All_caps) has generally been found to reduce legibility.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,228 @@
...
+        +-----------------------+--------------------------------------+
+        | ``"exponential"``     | 1.0                                  |
+        +-----------------------+--------------------------------------+
+        | ``"exp_plus_linear"`` | 1.0                                  |
+        +-----------------------+--------------------------------------+
+
+    voltage_bound: `float`
+        A bias voltage (in volts) that specifies an upper bound used to
+        collect the points for the curve fit.  That is, points that
+        satisfy ``voltage <= voltage_bound`` are used in the fit.
```
namurphy, line 107: Sooooooo tempted to suggest a unicode character here even though it would be a bad idea!

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,228 @@
...
+
+    # check voltage and current arrays
+    voltage, current = check_sweep(voltage, current, strip_units=True)
+
+    # condition kwargs voltage_bound and current_bound
+    if voltage_bound is None and current_bound is None:
+        current_bound = default_current_bound
+    elif voltage_bound is not None and current_bound is not None:
+        raise ValueError(
+            "Both keywords 'current_bound' and `voltage_bound' are specified, "
```
namurphy, line 179:
```suggestion
            "Both keywords 'current_bound' and 'voltage_bound' are specified, "
```

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,228 @@
...
+def find_ion_saturation_current(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    *,
+    fit_type: str = "exp_plus_linear",
+    current_bound: float = None,
+    voltage_bound: float = None,
+) -> Tuple[ffuncs.Linear, ISatExtras]:
+    """
+    Determines the ion-saturation current (:math:`I_{sat}`) for a given
```
namurphy, line 51: [optional] I'm wondering if ```:math:`I_{sat}` ``` is used often enough to warrant a reST substitution in `docs/common_links.rst` (possibly with the `\` at the end to help with spacing).

namurphy (commented):
It's looking good!  I have only minor/unimportant suggestions, most of which you are free to ignore or adopt as you see fit.  I could probably approve this pretty quickly after it's reviewed by someone who has a much better understanding of Langmuir stuff than I do.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
rocco8773: At this point I don't thinks so, but might I might change my tune as the `swept_langmuir` module expands.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
rocco8773: While I generally agree with the statement about the readability of all caps, I don't think it applies here.  The all caps statement applies to general writings (emails, books, journals, etc.), but we have something a little more specialized here.  One of my biggest issues with documenting functionality is that default values are buried within the documentation, making it rather difficult to find.  This is my issue with the suggested replacement.  Now, while I'm not particularly thrilled with my own implementation, I do think it provides better transparency.

My current take is to leave it how it is, and change it when we finally decided on a standard to implement throughout the package.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+        | ``"linear"``          | 0.4                                  |
+        +-----------------------+--------------------------------------+
+        | ``"exponential"``     | 1.0                                  |
+        +-----------------------+--------------------------------------+
+        | ``"exp_plus_linear"`` | 1.0                                  |
+        +-----------------------+--------------------------------------+
+
+        |
+
+        Can not be used with keyword ``voltage_bound``.
```
namurphy, line 107:
```suggestion
        Cannot be used with keyword ``voltage_bound``.
```

namurphy (commented):
Looks good!  I had wondered about using the extract function refactoring pattern, but I don't think it'd make any real difference here.  I didn't look into the Langmuir analysis stuff, so after @svincena gives a review of that, I'd be happy to merge it soon after.  Thank you for doing this!

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+    2. Fitting the selected points with the :term:`fit-function`
+       specified by ``fit_type``.
+    3. Extracting the linear component of the fit and returning that as
+       the ion-saturation current.
+
+    This routine opts to return a function representing a linear
+    ion-saturation current, since, while ideal planar Langmuir probes
+    reach a steady-state ion-saturation current, real world Langmuir
+    probes "suffer" from expanding sheaths as the bias voltage
+    increases.  This sheath expansion results the ion-saturation
```
svincena, line 151: becomes increasingly negative

svincena (approved):
Good to go, following our meeting today!

svincena (approved):

plasmapy/analysis/swept_langmuir/floating_potential.py
```diff
@@ -38,8 +38,7 @@ def find_floating_potential(
     ----------
 
     voltage: `numpy.ndarray`
-        1-D numpy array of monotonically ascending/descending probe biases
-        (should be in volts)
+        1-D numpy array of monotonically ascending probe biases (should be in volts)
 
     current: `numpy.ndarray`
         1-D numpy array of probe current (should be in amperes) corresponding
```
namurphy, line 44:
```suggestion
        1-D numpy array of monotonically ascending probe biases (must be in volts)

    current: `numpy.ndarray`
        1-D numpy array of probe current (must be in amperes) corresponding
```
[optional] If the function requires volts and amperes here, then we could be more specific with the language.  If not, feel free to ignore.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+    The current collected by a Langmuir probe reaches ion-saturation
+    when the probe is sufficiently biased so the influx of electrons is
+    completely repelled, which leads to only the collection of ions.
+    (For additional details see the **Notes** section below.)
+
+    **Aliases:** `find_isat_`
+
+    Parameters
+    ----------
+
```
namurphy, line 63:
```suggestion
```
[optional]

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+    voltage: `numpy.ndarray`
+        1-D numpy array of monotonically increasing probe biases
+        (should be in volts).
+
+    current: `numpy.ndarray`
+        1-D numpy array of probe current (should be in amperes)
```
namurphy, line 69:
```suggestion
        (must be in volts).

    current: `numpy.ndarray`
        1-D numpy array of probe current (must be in amperes)
```
Same optional comment as above.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+    2. Fitting the selected points with the :term:`fit-function`
+       specified by ``fit_type``.
+    3. Extracting the linear component of the fit and returning that as
+       the ion-saturation current.
+
+    This routine opts to return a function representing a linear
+    ion-saturation current, since, while ideal planar Langmuir probes
+    reach a steady-state ion-saturation current, real world Langmuir
+    probes "suffer" from expanding sheaths as the bias voltage
+    becomes increasingly negative.  This sheath expansion results the
```
namurphy, line 151:
```suggestion
    becomes increasingly negative.  This sheath expansion results in the
```
This my only quasi-required comment, so if you fix this I can approve it.  All of my other suggestions are either optional or hilarious.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+       ``voltage_bound`` or ``current_bound``.
+    2. Fitting the selected points with the :term:`fit-function`
+       specified by ``fit_type``.
+    3. Extracting the linear component of the fit and returning that as
+       the ion-saturation current.
+
+    This routine opts to return a function representing a linear
+    ion-saturation current, since, while ideal planar Langmuir probes
+    reach a steady-state ion-saturation current, real world Langmuir
+    probes "suffer" from expanding sheaths as the bias voltage
```
namurphy, line 150: [optional] Perhaps break this up into two sentences?

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+    becomes increasingly negative.  This sheath expansion results the
+    ion-saturation current also increasing.
+    """
+    rtn_extras = ISatExtras(rsq=None, fitted_func=None, fitted_indices=None)._asdict()
```
namurphy, line 154:
```suggestion
    rtn_extras = ISatExtras(rsq=None, fitted_func=None, fitted_indices=None)._asgard()
```
😬 ⚡ 🔨 

This was a really (t)horrible pun, if I do say so myself.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+            "current_bound": 1.0,
+        },
+    }  # type: Dict[str, Dict[str, Any]]
+    try:
+        default_current_bound = _settings[fit_type]["current_bound"]
+        fit_func = _settings[fit_type]["func"]()
+        rtn_extras["fitted_func"] = fit_func
+    except KeyError:
+        raise ValueError(
+            f"Requested fit '{fit_type}' is not a valid option.  Valid options "
```
namurphy, line 176:
```suggestion
            f"Requested fit {fit_type!r} is not a valid option.  Valid options "
```
[optional]

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+
+    if current_bound is not None:
+        if not isinstance(current_bound, numbers.Real):
+            raise TypeError(
+                f"Keyword 'current_bound' is of type {type(current_bound)}, "
+                f"expected an int or float."
+            )
+
+        current_min = current.min()
+        current_bound = (1.0 - current_bound) * current_min
```
namurphy, line 200: These variable names would also be appropriate in sea shanties.

> And in the mighty mighty storm 🎶 
> We sailed the current min. 🎵

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
```diff
@@ -0,0 +1,240 @@
...
+        mask = np.where(current <= current_bound)[0]
+    else:  # voltage_bound is not None
+        if not isinstance(voltage_bound, numbers.Real):
+            raise TypeError(
+                f"Keyword 'voltage_bound' is of type {type(voltage_bound)}, "
+                f"expected an int or float."
+            )
+
+        mask = np.where(voltage <= voltage_bound)[0]
```
namurphy, line 209: I wondered if [masked arrays from NumPy](https://numpy.org/doc/stable/reference/maskedarray.html) would be useful here, but it doesn't seem like changing the implementation would improve maintainability much.

namurphy (commented):
I have one quasi-required suggestion after which I'm happy to approve it.  All my other suggestions/comments are optional.

Thank you again for doing this, and I'm excited to be able to merge this soon!

namurphy (approved):
Actually, since we made the change to the settings today, I'll disable auto-merge and then approve it so that you can merge it when you're ready.  Thank you again!

plasmapy/analysis/swept_langmuir/floating_potential.py
rocco8773: I have to think about this. `voltage` and `current` should be in volts and amperes, respectively, but it is not needed for this function.  The Isat can be determined regardless of the units; it's just the returned Isat will share the same equivalent units as `voltage` and `current`.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
rocco8773: 🤦‍♂️

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
rocco8773: I'll have to read more about masked arrays, but I don't think it's quite applicable here. Because, `mask` is being used to filter two independent arrays, `voltage` and `current`.  I'm also suspecting, without definitive proof, that my current approach is probably less resource heavy.

plasmapy/analysis/swept_langmuir/floating_potential.py
rocco8773: I wrote this thinking about $I_{sat}$, but the statement is still true for $V_f$.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
rocco8773: Interesting little trick.  Not sure how I feel about it.  As a coder I like that it could help with consistency, but I also feel like that comes with a penalty of less explicitness.  I don't think this is the original intent behind `!r` and there might be other behaviors hidden behind the veil.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
namurphy: Cool!  I'm perfectly content keeping it as is.

plasmapy/analysis/swept_langmuir/ion_saturation_current.py
StanczakDominik: Oh wow, yeah, that's the first time I'm seeing it as well! Definitely a nice trick, but the previous implementation is way more obvious.

PR #1474, opened 2022-03-08
[PROTO] Replace nbsphinx with `myst_nb` for example notebooks

### LINKED ISSUE #1472: Enable Myst Markdown in example Jupyter notebooks
One of the problems with our current setup for our example notebooks is that we cannot use Sphinx roles or references in the Markdown cells of our notebooks.  For example, we can't do something like ``` `plasmapy.particles` ``` in a Markdown cell and get the link to be automagically generated.

[MyST](https://myst-parser.readthedocs.io/en/latest/) is a "rich and extensible flavor of Markdown meant for technical documentation and publishing."  My understanding is that its goals are to enable Markdown to be used in Sphinx docs.  I believe that the intention is for us to be able to do everything in MyST that you can in reStructuredText.  Using MyST in our notebooks would probably be really helpful.   

[MyST-NB](https://myst-nb.readthedocs.io/en/latest/index.html) is a Sphinx extension for MyST Markdown Notebooks.  I haven't looked into it much, but this might be the tool we're looking for.  It'd be worth at least trying out the steps on their [getting started page](https://myst-nb.readthedocs.io/en/latest/use/start.html) with our setup.  I'm not sure if this would replace or cause conflicts with [nbsphinx](https://nbsphinx.readthedocs.io).

If we do decide to adopt this extension, we could create some follow-up issues to convert the particular notebooks to use MyST/MyST-NB.

namurphy:
This PR is testing out using [myst_nb](https://myst-nb.readthedocs.io/en/latest/) instead of using [nbsphinx](https://nbsphinx.readthedocs.io) for generating the documentation associated with our example notebooks.  This is so that we can use the [MyST](https://myst-parser.readthedocs.io/en/latest/) flavor of Markdown in our notebooks.  More details are in #1472.  

Below are (probably) the steps that are needed to make this change.  I'll probably revise these steps since I'm only just starting to figure this out.

 - [x] Replace myst_nb in `docs/conf.py`, requirements files, and `setup.cfg`
 - [x] Replace `nbgallery` with Sphinx toctree for notebook pages
 - [ ] Replace `nbsphinx_thumbnails` in `docs/conf.py`
 - [ ] Replace `nbsphinx_prolog` in `docs/conf.py`
 - [ ] Update documentation guide to reflect changes

Closes #1472.


namurphy:
So, I was able to get this to work (at least locally; I'm still waiting for the RTD build).  However, as far as I can tell, the current implementation of `myst_nb` does not have a way to generate a thumbnail gallery.  In particular, the documentation states that we need to put the notebooks in a `:toctree:`, whereas `nbsphinx` has the `:nbgallery:` directive.  As such, the notebooks show up as a plain table of contents.  However, `myst_nb` is a pretty new package and I'd expect that they would add that functionality in the not-too-distant future.  

My inclination at the moment is to hold off on switching to `myst_nb` until there's a supported way of generating a thumbnail gallery of notebooks.  With that said, being able to use intersphinx and all of MyST's capabilities in our example notebooks would be a significant advantage, which would probably make it less work to write and maintain notebooks.

rocco8773:
One thing to note, it looks like `myst-nb` is executing the radiography notebooks.  At least that's what's happening on my local build and the RTD builds have jumped by 3+ mins.

rocco8773:
> So, I was able to get this to pretty much work (here's the [examples page for this PR](https://plasmapy--1474.org.readthedocs.build/en/1474/examples.html)). However, as far as I can tell, the current implementation of `myst_nb` does not have a way to generate a thumbnail gallery. In particular, the documentation states that we need to put the notebooks in a `:toctree:`, whereas `nbsphinx` has the `:nbgallery:` directive. As such, the notebooks show up as a plain table of contents. However, `myst_nb` is a pretty new package and I'd expect that they would add that functionality in the not-too-distant future.


Yeah, this really bums me out.  I much prefer the look and feel of the thumbnail gallery over the bullet list.

rocco8773:
I do have a bit of a hack that gets the notebook gallery working again.  The downside is that it doesn't know where to find the thumbnail image unless you hard code it in the configuration variable `nbsphinx_thumbnails`.  By updating `conf.py` with the following code block I got the galleries to work on my local build...

```python
def hack_nbsphinx(app: Sphinx) -> None:
    from nbsphinx import (
        NbGallery,
        doctree_resolved,
        patched_toctree_resolve,
        GalleryNode,
        depart_gallery_html,
    )
    from sphinx.environment.adapters import toctree

    def builder_inited(app: Sphinx):
        if not hasattr(app.env, 'nbsphinx_thumbnails'):
            app.env.nbsphinx_thumbnails = {}

    def do_nothing(*node):
        pass

    app.add_config_value("nbsphinx_thumbnails", {}, rebuild="html")
    app.add_directive("nbgallery", NbGallery)
    app.add_node(
        GalleryNode,
        html=(do_nothing, depart_gallery_html),
        latex=(do_nothing, do_nothing),
        text=(do_nothing, do_nothing),
    )
    app.connect('builder-inited', builder_inited)
    app.connect("doctree-resolved", doctree_resolved)

    # Monkey-patch Sphinx TocTree adapter
    toctree.TocTree.resolve = patched_toctree_resolve


    hack_nbsphinx(app)

    app.add_config_value("revision", "", True)
    app.add_css_file("css/admonition_color_contrast.css")
    app.add_css_file("css/plasmapy.css", priority=600)
```

I basically pulled just the gallery parts from `nbsphinx` and [duplicated the extension setup they had in their repo](https://github.com/spatialaudio/nbsphinx/blob/d7241b111178fb850f09b52574c8d1d720c84099/src/nbsphinx.py#L2312).

If we did this, then we'd likely want to freeze the `nbsphinx` version.  I don't know if we can rely on this functionality not changing.

namurphy:
> I do have a bit of a hack that gets the notebook gallery working again. The downside is that it doesn't know where to find the thumbnail image unless you hard code it in the configuration variable `nbsphinx_thumbnails`.

Nice!  That downside is one I could live with, especially given everything else in MyST-NB.  I'm wondering if it would be worth checking with the developers of MyST-NB to ask if/when they're planning to add gallery functionality, since we might not need to go through the effort of hardcoding the thumbnail images.

namurphy:
> One thing to note, it looks like `myst-nb` is executing the radiography notebooks. At least that's what's happening on my local build and the RTD builds have jumped by 3+ mins.

It looks like they have [instructions on executing and caching content](https://myst-nb.readthedocs.io/en/latest/use/execute.html).  I'll need to go through them.

namurphy:
So far I'm pretty excited about MyST-NB!  There are still several things that we'd need to do before we would be able to switch over, but we at least have a working prototype after implementing the suggested hack above for the gallery.  Overall I'm thinking that it'd be best to wait a while to switch over since there are several features that we are using from nbsphinx that are not yet in MyST-NB, like the gallery feature.  MyST-NB is also still in beta, so the package is still evolving, and it might be better to wait until it stabilizes a bit more.  

I converted the units notebook from `.ipynb` → `.md` to try this out.  I was able to do the initial conversion with `jupytext units.ipynb --to myst`, but I still needed to do a lot of manual changes (like adding links to code objects).  I was able to link to code objects by doing things like ``` {py:mod}`~plasmapy.particles` ```, though there appear to be some ways to do substitutions/replacements (akin to what we do in `docs/common_links.rst`) that I'd eventually like to look into.  I'm thinking that it's likely that keeping (at least some of) our notebooks in markdown format will make them a lot easier to maintain...though I don't know if that'll be true for more complicated things like the charged particle radiography notebooks.  There might be some ways to [use markdown/MyST with nbsphinx](https://nbsphinx.readthedocs.io/en/0.8.8/a-markdown-file.html) too, though I haven't looked into it much.

The remaining tasks that I can think of are...

 - [ ] Decide on which extensions from `myst_enable_extensions` to enable in `docs/conf.py`
 - [ ] Figure out a way to do substitutions (akin to how we do `|Particle|` in `docs/common_links.rst`) 
 - [ ] Update the other notebooks that have warnings for (mostly) multiply defined references.
 - [ ] Enable notebooks to be downloaded directly (not sure if there's a way to do this yet)
 - [ ] Store the thumbnail images for the notebooks that generate their own thumbnail image (currently done in `nbsphinx_thumbnails` in `docs/conf.py`)
 - [ ] Figure out a way to replace what's being done in `nbsphinx_prolog` in `docs/conf.py`
 - [ ] Figure out a way to open the notebooks in binder or another site that can open Jupyter notebooks
 - [ ] Figure out how to cache computationally intensive notebooks
 - [ ] Make sure that the links in the table of contents for the units notebook are working

rocco8773:
@namurphy Why are we converting notebooks to markdown files!?  I don't like this idea!!!  A big reason for incorporating notebooks is so users can download and play with them.  Converting notebooks defeats that point and will make future modification/maintenance a headache.

rocco8773:
My biggest issue with `myst_nb` at the moment is not the gallery, but it seems to want to execute notebooks on every build, regardless of the notebook already being executed.  According to their documentation, this is not suppose to happen, but it is currently happening with the radiography notebook on both the RTD and my local builds.  I can tells this because the build times for the `myst-nb` have shot up by 3 min.  Since long build times was a major issue in the past, I don't think the `myst_nb` version should be merged at all until this is resolved!!

namurphy:
> @namurphy Why are we converting notebooks to markdown files!?

For the purposes of this PR, I'm mostly doing this as a prototype so we can see what it's like.  As with pretty much all of our choices, there are tradeoffs, and I'm thinking it's worthwhile to try this out to understand what the tradeoffs are.  I tend to think of prototype PRs as learning experiences where we can try out new stuff.

There are tools like [Jupytext](https://jupytext.readthedocs.io/en/latest/) that can convert between `.md` and `.ipynb` files, so there should be a way to enable the notebooks to be generated during the documentation build so that users would be able to download them.  The way we have it set up, though, we don't have links to download any of the notebooks regardless of whether the source is `.md` or `.ipynb`, or links that load the notebooks in binder.  I agree fully that there needs to be the capability for end users to download the notebooks.

> My biggest issue with `myst_nb` at the moment is not the gallery, but it seems to want to execute notebooks on every build, regardless of the notebook already being executed.

Agreed — we'd need to figure this out.  I started looking into the configuration variables for that but haven't figured it out yet.

One of my goals with this PR is to figure out what capabilities we'd need to switch over the MyST-NB, and then raise issues in their repository.  I'm still excited about this, while simultaneously thinking that it's too early to switch over in the near future.

namurphy:
There was a major update to `myst_nb` done in https://github.com/executablebooks/MyST-NB/pull/380, so the way it's set up in this PR in `docs/conf.py` might need to be adjusted when the changes are included in a release.

twiecki:
This PR was super helpful in creating a gallery a PyMC: https://github.com/pymc-devs/pymc-examples/pull/328

I'm now trying (but failing) to get thumbnails to work. I tried using metadata tags or using the `nbsphinx_thumbnails` config option but neither work. Did you also run into this problem or are thumbnails working correctly for you?

namurphy:
> This PR was super helpful in creating a gallery a PyMC: [pymc-devs/pymc-examples#328](https://github.com/pymc-devs/pymc-examples/pull/328)

Happy to hear that it was helpful!

> I'm now trying (but failing) to get thumbnails to work. I tried using metadata tags or using the `nbsphinx_thumbnails` config option but neither work. Did you also run into this problem or are thumbnails working correctly for you?

I took at look at our [examples gallery for this PR](https://plasmapy--1474.org.readthedocs.build/en/1474/examples.html) and noticed that most of the thumbnails weren't working.  There are a few that are showing up correctly, which are the ones defined in [`nbsphinx_thumbnails`](https://github.com/PlasmaPy/PlasmaPy/blob/1b9238439c1e4a7c1d35b8f812b4805872f0aae6/docs/conf.py#L452) in our `docs/conf.py`.  I'm not sure what's happening here.

twiecki:
@namurphy Thanks for the quick reply. I now also got it working with the conf.py approach, although I have to copy the images into the build output dir manually currently. But I guess that aspect is working for you?

namurphy:
Closing this for the time being, though it would be great to come back to this in a year or two.

One lesson from this prototype was that working on example notebooks in the form of an `.md` file was significantly more pleasant and less frustrating than working on it as an `.ipynb` file, and code review & maintenance would be a lot easier too.  We'd need to convert them to `.ipynb` files when building the docs though, but there are tools (like `nbconvert`, I think) that can handle this really well.

PR #1473, opened 2022-03-08
Lite-Functions Ep. 3 | plasma dispersion functions

pheuer:
Creates lite versions of the plasma dispersion function and its derivative.

This just eliminates input validation: future work necessary to cythonize these functions.

Necessary for #974

PR #974, opened 2021-01-11
Fitting Thomson Scattering Spectra

pheuer:
The use of Thomson scattering as a diagnostic depends on the ability to extract plasma parameters (eg. density, temperature, flow velocity) from scattered light spectra. One of the design uses for the `diagnostics.thomson.spectral_density` function would be as part of a fit function to curve fit experimental data to extract these parameters.  As such, I've received requests to develop a function for fitting TS data. 

This function provides a `lmfit.Model` function (`thomson.thomson_model`) with which the user can apply any of the `lmfit` fitting algorithms and other features to Thomson scattering data. The example notebook exhibits in detail how this is done.

** EDIT **

This PR was starting to get big, so I've restricted its scope thus:

**This PR**
- Basic Thomson fitting `lmfit` model function.
- Changes to `thomson.spectral_density` necessary to make the fitting work.
- One example notebook that demonstrates fitting
- Minor easy speed improvements.

**Future PRs**
- Optimize `thomson.spectral_density` and sub-functions to make the fitting algorithm fast enough for research (#1085)
- Create a function that estimates the errors in a Thomson fit (or maybe more generally in an `lmfit` result?). PR to come

PR #1476, opened 2022-03-10
Lite-Functions Ep. 4 | permittivity_1D_Maxwellian

pheuer:
Creates a lite function version of the `permittivity_1D_Maxwellian` function, which is required for #974 

This PR depends on #1473 and should be merged after that PR is complete. 

This function internally calls `thermal_speed` and `plasma_frequency`. So, in the lite function we have a choice: 

- Take all the arguments for both of those functions, plus new ones for `permittivity_1D_Maxwellian`, and call the other lite functions inside `permittivity_1D_Maxwellian_lite`.

- Take `vTh` and `wp` as inputs, and require the lite function user to calculate them using their lite functions separately. 

I chose the second for a number of reasons 

- Minimizes the number of inputs and duplicated documentation. Otherwise we have to re-document every argument on this new lite function.
- Potentially more efficient: if the user has already calculated vTh and wp (as is indeed the case in the Thomson module) repeating the computation is wasteful. 
- This way the regular `permittivity_1D_Maxwellian` function still calls the regular `thermal_speed` and `plasma_frequency` functions, which means all of their input validation is being done as normal. 

The only downside is that, if a lite function user only wants `permittivity_1D_Maxwellian_lite`, they will have to explicitly call the functions it depends upon. But I don't think that's a big deal: they're a lite function user, so they should be willing to go through this extra bit of trouble for the sake of efficiency.


pheuer:
This PR (and #1473 ) are ready to review. The failing changelog test here is an error, because I based this branch off of #1473 and so it's seeing that changelog entry.

rocco8773:
> This PR (and #1473 ) are ready to review. The failing changelog test here is an error, because I based this branch off of #1473 and so it's seeing that changelog entry.
> 
> Edit: Does @namurphy or @StanczakDominik know a way to avoid this? Should I do something differently with the PR, or is this a bug with the changelog CI?

You don't need to worry about it right now.  It should resolve itself once PR #1473 is merged.  It's showing up in this PR right now because it is a change over the current `main` branch.

rocco8773:
> This PR depends on #1473 and should be merged after that PR is complete.
> 
> This function internally calls `thermal_speed` and `plasma_frequency`. So, in the lite function we have a choice:
> 
>     * Take all the arguments for both of those functions, plus new ones for `permittivity_1D_Maxwellian`, and call the other lite functions inside `permittivity_1D_Maxwellian_lite`.
> 
>     * Take `vTh` and `wp` as inputs, and require the lite function user to calculate them using their lite functions separately.
> 
> 
> I chose the second for a number of reasons
> 
>     * Minimizes the number of inputs and duplicated documentation. Otherwise we have to re-document every argument on this new lite function.
> 
>     * Potentially more efficient: if the user has already calculated vTh and wp (as is indeed the case in the Thomson module) repeating the computation is wasteful.
> 
>     * This way the regular `permittivity_1D_Maxwellian` function still calls the regular `thermal_speed` and `plasma_frequency` functions, which means all of their input validation is being done as normal.

@namurphy Can I get your opinion on this?  I completely agree with @pheuer's points, but I can pull myself away from the position that the `lite` signature should be the same as the non-lite version and the arguments should be in the base parameters because that's the rule for the whole formulary.

rocco8773 (approved):
For expeditious reason I went ahead and made the changes I would have suggested.  All my comments would have been related to typos, PEP8 fixes, or the same type of suggestions I made in #1473.  Nothing too crazy.

rocco8773 (approved):

pheuer:
Awesome  - thanks @rocco8773 !

PR #1478, opened 2022-03-12
Describe doctests in documentation guide

### LINKED ISSUE #834: Document use of ELLIPSIS for floating-point doctests in case of Astropy 3.1 tests with old values of physical constants in the Testing Guide
> I noticed that the error is a mismatch in the floating value of the charge.
> [#833 (checks)](https://github.com/PlasmaPy/PlasmaPy/pull/833/checks?check_run_id=740174437)
> 
> Is it possible that this may change depending on the system and platform being used? I have heard somewhere that floating points are not represented the same way in all machines and platforms. Any suggestions on how we can potentially achieve value consistency across devices?

Good intuition! But I've seen this bug before (and should definitely have documented it) - it's actually a case of new Astropy versions having newer and more accurate values of physical constants (most likely, here - the mass of a proton and electron)! 

Take a look at how doctests in `parameters` use ellipsis to handle this and skip checking in significant digits:

https://github.com/PlasmaPy/PlasmaPy/blob/c03d8c999c35ca55adce8ac577a668f8ae8870da/plasmapy/formulary/parameters.py#L108-L114

<details>
<summary>If you care about how to set up pytest to do this kind of skipping:</summary>

https://github.com/PlasmaPy/PlasmaPy/blob/c03d8c999c35ca55adce8ac577a668f8ae8870da/setup.cfg#L76-L79

</details>

_Originally posted by @StanczakDominik in https://github.com/PlasmaPy/PlasmaPy/pull/833#issuecomment-639288741_

---

That should do it. I'm trying to be more explicit about issue requirements nowadays. :3

namurphy:
This PR adds a short section on doctests in the documentation guide.  Importantly, it documents the use of `# doctest: +SKIP` for skipping lines and the use of `...` for avoiding spurious doctest failures when doing floating point comparisons or when fundamental constants are updated.  Closes #834.


StanczakDominik (approved):
LGTM!

PR #1465, opened 2022-03-04
Bump minimum required version of Python from 3.7 to 3.8

namurphy:
This PR changes PlasmaPy's minimum required version of Python from 3.7 to 3.8.  It seems timely to do this since a lot of our upstream packages (Astropy, NumPy, SciPy, xarray) have dropped support for Python 3.7 in accordance with the deprecation schedule proposed in [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) for the broader scientific python community.  NEP 29 is really helpful in that it sets a community standard for a drop schedule.

There have also been a number of times in the last few months where I had an opportunity to use a Python 3.8 feature (usually either [assignment expressions](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) or [f-string support for self-documenting expressions](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)) but didn't.  Since I'm attempting some automated refactorings with Sourcery (#1463, #1464), I have an opportunity to make a lot of these changes automatically.

Closes #1262.

PR #1482, opened 2022-03-15
Update requirements and installation instructions

### LINKED ISSUE #1481: Hasten CI tests for Python 3.8 with minimum dependencies
I've noticed that at least since #1465, the tests for Python 3.8 with minimum dependencies have been much slower to run than the other tests.  Looking through the output of some of the tests, it appears that the test suite is being run twice per CI check which is probably the problem.  We should figure this out and speed these tests up.  

On the chance that there is some other reason why these tests are slow, we could run them during the initial round of tests, but not require that they pass before the second round of tests is begun.

namurphy:
The original goal of this PR is to speed up the tests of Python 3.8 with minimal dependencies, which I noticed have been slow since at least #1465 (possibly before).  I'm not sure why this is happening, so it may take some trial and error.  Closes #1481.

Edit: The expanded goal is to also update/reorganize our requirements, update the installation instructions, and deprecate the `[all]` option for installing `plasmapy` with `pip`.  These changes are somewhat coupled to each other so I decided to keep it in one PR despite the scope creep and monotonically increasing number of changlog entries.


tox.ini
```diff
@@ -103,8 +103,6 @@ deps =
   hypothesis
   pytest-regressions
   numba
-setenv =
-  PYTEST_COMMAND = pytest --pyargs plasmapy --durations=25
```
namurphy, line 107: Removing these lines broke it...

rocco8773:
When I reviewed the CI log, PR #1465 was the first PR to experience the slow minimal tests.  (https://github.com/PlasmaPy/PlasmaPy/actions/runs/1970484613)

requirements/install.txt
```diff
@@ -4,11 +4,11 @@ astropy >= 4.3.1
 cached-property >= 1.5.2
 ipywidgets >= 7.6.5
 matplotlib >= 3.3.0
-numpy >= 1.18.1
+numpy >= 1.19.0
```
rocco8773, line 7: FYI, `numpy` introduced typing in `v1.20`, [`numpy.typing`](https://numpy.org/devdocs/reference/typing.html).  I'm just saying...

namurphy: WHOA!  That's awesome!  We'd need to keep supporting NumPy 1.19 for the moment to be consistent with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), but I'd support requiring 1.20+ starting just after our 0.8.0 release.

requirements/environment.yml
```diff
@@ -1,34 +1,55 @@
...
-  - hypothesis
-  - docutils
-  - numpydoc
-  - pytest
-  - lmfit
-  - pygments >= 2.11.0
-  - sphinx >= 3.2.0
-  - sphinx-gallery
+  - flake8-absolute-import
+  - sphinx-changelog
```
rocco8773, line 54: I think we can replace all of this with just... (https://stackoverflow.com/a/59056234)
```YAML
dependencies
  - python=3.10
  - pip
  - pip:
    - -r requirements.txt
```

namurphy: You read my mind!  I was just about to look up if there was a way to do this without having to unnecessarily repeat everything.

namurphy: Looks like it's working!  Reducing the number of places where we specify requirements is a huge advantage...especially since it's easy for the `environment.yml` file to become out-of-date.

One consideration is that it's generally bad to mix installation between pip and conda.  However, this change will make all of the package installations be done by pip, so I think we're good.

changelog/1482.trivial.1.rst
```diff
@@ -0,0 +1,3 @@
+Bumped the minimum versions of `mpmath` to ``1.2.1``, `numpy` to
+``1.19.0``, `pandas` to ``1.0.0``, `pytest` to ``5.4.0``, `scipy` to
+``1.5.0``, and `xarray` to ``0.15.0``.
```
namurphy, line 3: - [ ] Fix mpmath & xarray links

docs/install.rst
```diff
@@ -10,29 +10,6 @@ Installing PlasmaPy
...
-- `tqdm <https://tqdm.github.io>`_ — 4.41 or newer
-- `cached_property <https://pypi.org/project/cached-property>`_ — 1.5.2 or newer
-
-PlasmaPy also depends on the following packages for optional features:
-
-- `h5py <https://www.h5py.org>`_ — 2.8 or newer
-- `lmfit <https://lmfit.github.io/lmfit-py>`_ — 1.0.1 or newer
-- matplotlib_ — 2.0 or newer
-- `mpmath <https://mpmath.org>`_ — 1.0 or newer
-
```
namurphy, line 35: I am still debating about this change.  Here's my tradeoff analysis so far.

Reasons to keep this:
 - There may be some situations where users need or want to know the requirements in detail.

Reasons to delete this:
 - This information isn't necessary for users when installing PlasmaPy, and may actually be distracting.
 - This information is likely to become out-of-date when we forget to update it.
 - This is one more location where requirements are specified.
 - When advanced users need to know the requirements, they are likely to know where to look in the source code.

Alternatives
 - Update this without making other changes
 - Move this lower in the installation page (or elsewhere in the documentation)
 - Automatically get the requirements from the `.txt` files in `requirements/`

namurphy:
Why hello, my old friend scope creep!  It's been a minute. 😹

docs/install.rst
rocco8773: I'd be in support of removing it, but it is scope creep.  So, I'd also support leaving it for now and taking care of it in a follow up PR.

namurphy: I'm inclined to keep it in this PR because we'd need to edit this section anyway to update the requirements even if we didn't remove it.  Thanks!

rocco8773: However, I still think at a minimum we should keep the statement about Python 3.8 or newer.  It makes sense that that is stated at the beginning of the installation page.

changelog/1482.trivial.2.rst
```diff
@@ -0,0 +1,3 @@
+Added dependencies on ``dlint``, flake8_, ``flake8-absolute-import``,
+``flake8-rst-docstrings``, ``flake8-use-fstring``, pre-commit_,
+pydocstyle_, and pygments_.
```
rocco8773, line 3:
```suggestion
Added ``dlint``, flake8_, ``flake8-absolute-import``,
``flake8-rst-docstrings``, ``flake8-use-fstring``, pre-commit_,
pydocstyle_, and pygments_ to test dependencies, but not
install dependencies.
```

changelog/1482.trivial.3.rst
```diff
@@ -0,0 +1,3 @@
+Updated :file:`docs/environment.yml` to use `pip` to install all
```
rocco8773, line 1: Note that `pip` does not get linked by intersphinx.  I think this is because we don't have `pip` in the mappings, which I think is fine.  You could do one of your fancy replacements to point to the [`pip` documentation](https://pip.pypa.io/en/stable/), or just double back-tick it.

rocco8773 (changes requested):
Looks straightforward to me.  There's just a few minor things to address, main that I think we should still state the minimum Python dependency at the top of the install page.

It just occurred to me, but the install page could also benefit from a `.. contents::` table of contents at the top of the page.  That is scope creep though.

namurphy:
- [ ] Create issue about `plasmapy/optional_deps.py`

namurphy:
I made one more change to delete `optional_deps.py`.  We will no longer need it because this PR makes it so that we do not have optional dependencies.  (Later on if we decided to go with optional dependencies again, we can add it back in.)

docs/install.rst
```diff
@@ -10,145 +10,216 @@ Installing PlasmaPy
...
 
-Installation with pip
----------------------
+   python -m pip install plasmapy
 
-To install the most recent release of PlasmaPy on `PyPI`_
-with `pip`_ into an existing Python environment
-with both required and optional dependencies, run
+On some systems, it might be necessary to specify the Python_ version
+number by using ``python3`` or ``python3.10`` instead of ``python``.
```
namurphy, line 45:
```suggestion
number, for example by using ``python3`` or ``python3.10`` instead of ``python``.
```

namurphy:
pre-commit.ci autofix

namurphy:
I made one more change to delete `optional_deps.py` and its usages.  We will no longer need it because this PR makes it so that we do not have optional dependencies.  (Later on if we decide to go with optional dependencies again, we can add it back in.)

rocco8773 (approved):

PR #1484, opened 2022-03-16
Fix broken hyperlinks

namurphy:
I tried `make linkcheck` for the docs because waiting for beans to boil for 87 minutes is boring, and found a few broken hyperlinks.  This PR fixes those hyperlinks, and the beans are tasty.


StanczakDominik (approved):
No questions except maybe for further details on the beans!

PR #1492, opened 2022-03-19
Use different action to generate PR size labels

RAJAGOPALAN-GANGADHARAN:
Overwrites the work of PR #1489 , looks like the previous used action was faulty hence replacing it. Tested the changes on my fork it works fine. Since this is a trivial change not adding a changelog entry. 

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -12,7 +12,21 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         sync-labels: true
     - name: Add Size Label to PR
-      uses: WizardOhio24/label-pr-size-action@master
+      uses: codelytv/pr-size-labeler@v1
       with:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        size-label-color: 10, very small, ffaf65 | 50, small, ff9e58 | 200, medium, ff9654 | 400, large, ff905d | 100000, massive, ff905d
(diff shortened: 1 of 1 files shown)


RAJAGOPALAN-GANGADHARAN:
Requesting review from @namurphy.

.github/workflows/labeler.yml
```diff
@@ -12,7 +12,21 @@ jobs:
...
+        m_label: size/medium
+        m_max_size: '200'
+        l_label: size/large
+        l_max_size: '400'
+        xl_label: size/extra-large
+        fail_if_xl: 'false'
+        message_if_xl: >
+          'This PR exceeds the recommended size of less than 400 lines.
+          Please make sure you are NOT addressing multiple issues with one PR.
+          Note this PR might be rejected due to its size.'
```
namurphy, line 31:
```suggestion
        message_if_xl: ''
```
At the moment, my personal preference would be to not post a message if the PR is extra large, though we could probably revisit this later.  Looking at [this line](https://github.com/CodelyTV/pr-size-labeler/blob/6601a2d04f0fd3f581cd7b175bbb3dce8dcfd6bf/src/labeler.sh#L24) of the action...I *think* that if this variable is set to an empty string (which is checked with the `-n` flag), then the message won't get added in a comment.

namurphy (dismissed):
Thank you again for doing this!  I have one quick suggestion, and then I think we can merge this.

namurphy (approved):

PR #1493, opened 2022-03-21
Use `codespell` for spellchecking

namurphy:
For this PR, I used [`codespell`](https://github.com/codespell-project/codespell) to perform some spellchecking on the package.  Because `codespell` results in false positives (for us, this includes `Te` and some atomic symbols), it should be run interactively, so we wouldn't need to be able to automate this.  Because it is only looking for common misspellings, this will not catch every spelling mistake.

The command I used was:

```bash
pip install codespell
codespell --skip ".tox,.git" -i 3 -w
```


StanczakDominik:
Hmm, apparently you can use `codespell -I FILE, --ignore-words=FILE` to set a list of words to exclude. Given that, would we want to automate this?

StanczakDominik (approved):
LGTM!

PR #1494, opened 2022-03-22
Use `contextlib.suppress` to suppress exceptions instead of `try` & `except`

namurphy:
This PR implements a recommended refactoring that I learned recently.  Instead of:

```Python
try:
    ...
except ValueError:
    pass
```

we can do:

```Python
    with contextlib.suppress(ValueError):
        ...
```

However, the main purpose of this is to test the pull request labeler in #1492.

### DIFF
--- a/changelog/1494.trivial.rst
+++ b/changelog/1494.trivial.rst
@@ -1,0 +1,2 @@
+Used `contextlib.suppress` to suppress exceptions, instead of `try` &
+`except` blocks.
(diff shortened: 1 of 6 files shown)


RAJAGOPALAN-GANGADHARAN:
Seems to be failing at Add Labels PR step. Seems like a transient issue, can we re run the workflow?

namurphy:
Hm...the output from the workflow includes:
```
Labeling pull request with size/small
```
but it didn't seem to work.  It might have been because I created the label _after_ the PR was made.  I also noticed that GitHub was having some problems an hour or two ago, so it could be related to that.  Might need to see what happens on the next PR...

RAJAGOPALAN-GANGADHARAN:
https://github.com/RAJAGOPALAN-GANGADHARAN/PlasmaPy/pull/4 - Tested on my local fork it was able to label it properly. Not sure if it's missing some permissions to add labels?

rocco8773 (approved):

PR #1498, opened 2022-03-24
[TEST] skipping GA checks with labels

rocco8773:
This PR puts a framework into the CI Test workflow that allows parts of the tests to be skipped based on labels added to the PR.  

The motivation for this prototype came about when the GA Documentation Test was failing due to a recent release of `jinja2` and not related to what the contributor was doing.  In this scenario we were completely block from merging the PR until the test was resolved, luckily in that case it was an easy fix.

Another upside to this approach, we can turn off comprehensive tests until we reach the end of a PR development.  This should allows to both expand PR tests while not overloading our resources.

I did build a job into the workflow that does fail if other jobs are skipped, unless the "GA Skip Merge Allowed" label is added to the PR.  This should add some redundancy against us unintentionally merging a PR with skipped tests.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -34,16 +34,14 @@ jobs:
           echo "Label is '${{ env.SKIP_DOCUMENTATION_LABEL }}'"
           SKIP=${{ contains(github.event.pull_request.labels.*.name, env.SKIP_DOCUMENTATION_LABEL) }}
           echo "SKIP_DOCUMENTATION=${SKIP}" >> $GITHUB_ENV
-          if [[ $SKIP ]]; then
-            echo "1"
-            echo "dSKIP=1" >> $GITHUB_ENV 
-          else
-            echo "2"
-            echo "dSKIP=2" >> $GITHUB_ENV
-          fi
-      - name: Skip Documentation?
+          SKIP=${{ contains(github.event.pull_request.labels.*.name, env.SKIP_COMPREHENSIVE_LABEL) }}
+          echo "SKIP_COMPREHENSIVE=${SKIP}" >> $GITHUB_ENV
+      - name: Skip Documentation? ${{ env.SKIP_DOCUMENTATION }}
         run: |
           echo "Documentation can be skipped: ${{ env.SKIP_DOCUMENTATION }}"
+      - name: Skip Comprehensive Tests? ${{ env.SKIP_COMPREHENSIVE }}
+        run: |
+          echo "Comprehensive tests can be skipped: ${{ env.SKIP_COMPREHENSIVE }}"
 
 #      - name: Were tests skipped?
 #        run: echo ${{ env.SKIPS_DONE }}
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Skip Tests Allowed? ${{ env.SKIPS_ALLOWED }}
         run: |
           echo "Is skipping tests allowed: ${{ env.SKIPS_ALLOWED }}"
-          FAIL=${{  !env.SKIPS_ALLOWED }}
+          FAIL=${{ env.SKIPS_ALLOWED }}
           if [[ FAIL ]]; then
             echo "Some tests are skipped, but skipping needs to be removed before merging!!"
             exit 1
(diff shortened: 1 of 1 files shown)


rocco8773:
@namurphy @StanczakDominik I'm curious what both of your takes are on this setup.  I have my own takes on it, but I want to hear yours before I give my thoughts.

rocco8773:
Oh, and feel free to play with the labels to see how they work.

StanczakDominik:
I was briefly confused about what checks are running for General Atomics... I'll give this a look through today.

.github/workflows/testing.yml
```diff
@@ -46,7 +119,10 @@ jobs:
   comprehensive-tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
-    needs: initial-tests
+
+    if: ${{ needs.skips-allowed.outputs.SKIP_COMPREHENSIVE == 'false' }}
+    needs: [initial-tests, skips-allowed]
```
StanczakDominik, line 124: So if `SKIP_COMPREHENSIVE` is false, this needs both of these... oh, no, right, this runs conditionally only if that var is false, correct?

.github/workflows/testing.yml
```diff
@@ -101,6 +176,8 @@ jobs:
       with:
         file: ./coverage.xml
   documentation:
+    if: ${{ needs.skips-allowed.outputs.SKIP_DOCUMENTATION == 'false' }}
+    needs: skips-allowed
```
StanczakDominik, line 180: What I don't understand is... does this only run if you're enabling `skips-allowed` via `GA Skip Merge Allowed`? Suppose I don't skip anything, does this still run?

StanczakDominik (changes requested):
The overall idea looks great! A couple questions for the implementation, and I think we could also give the labels descriptions in https://github.com/PlasmaPy/PlasmaPy/labels.

.github/workflows/testing.yml
```diff
@@ -9,9 +9,82 @@ on:
...
+      SKIP_COMPREHENSIVE: false
+      SKIPS_ALLOWED: false
+
+      SKIPS_DONE: false
+
+    outputs:
+      SKIP_DOCUMENTATION: ${{ env.SKIP_DOCUMENTATION }}
+      SKIP_COMPREHENSIVE: ${{ env.SKIP_COMPREHENSIVE }}
+      SKIPS_ALLOWED: ${{ env.SKIPS_ALLOWED }}
+      SKIPS_DONE: ${{ env.SKIPS_DONE }}
```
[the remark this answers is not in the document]
rocco8773: Yep.  I originally tries this by defining `env` on the workflow level, but it turns out the workflow level `env` just acts as an initialization for the job level `env`.  So, if I modify an environment variable in `job1` then that doesn't carry over to `job2`.  The only way I've found that I can pass this information between jobs is using `outputs` for the job.

.github/workflows/testing.yml
```diff
@@ -9,9 +9,82 @@ on:
...
+
+    steps:
+      - name: Determine Skips
+        id: step1
+        if: ${{ env.IS_PR }}
```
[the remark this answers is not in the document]
rocco8773: I could.  I think I was originally thinking this would need to be used in multiple spots.

.github/workflows/testing.yml
```diff
@@ -9,9 +9,82 @@ on:
...
+    env:
+      SKIPS_ALLOWED: ${{ needs.skips-allowed.outputs.SKIPS_ALLOWED }}
+      SKIPS_DONE: ${{ needs.skips-allowed.outputs.SKIPS_DONE }}
+
+    steps:
+      - name: Skip Tests Allowed? ${{ env.SKIPS_ALLOWED }}
+        run: |
+          echo "Is skipping tests allowed: ${{ env.SKIPS_ALLOWED }}"
+          if [[ "$SKIPS_ALLOWED" == "false" && "$SKIPS_DONE" == "true" ]]; then
+            echo "Some tests are skipped, but skipping needs to be removed before merging!!"
```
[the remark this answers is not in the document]
rocco8773: I believe this would work.  I originally didn't write something like this because I only wanted us to know about the "skips allows" functionality.  Only we should be adding that label and not the common contributor.

.github/workflows/testing.yml
rocco8773: Correct

.github/workflows/testing.yml
rocco8773: The `skips-allowed` job always runs regardless of labeling, and will never fail.  I breakout that "merge-able failure" into a separate job since if it fails any dependent jobs would automatically skip.

Since the `outputs` are defined using the job's `env`, the `outputs` are automatically initialized with the `env` values.

Issue #1447, opened 2022-02-26
Do we need a `import plasmapy` GitHub Action check?


rocco8773:
I'm a little puzzled how all the checks for PR #1444 passed when an `ImportError` was raised when importing `plasmapy.formulary.parameters` (see PR #1446).  I'm thinking we should add to our CI checks a check on importing `plasmapy` to ensure it imports without error.

@namurphy @StanczakDominik Thoughts?

namurphy:
I like it.  Given my history of directly inspiring the Astropy 4.3.1 bugfix release due to accidentally deleting the `astropy.units.imperial` import in `__init__.py`, I'd say this check (and perhaps even checks that our subpackages are in the top-level namespace) would be worth having.  With that said, I've *probably* learned my lesson.  Probably.

namurphy:
I'm wondering if there's a pytest extension for this...

namurphy:
Closed by #1501.

PR #1500, opened 2022-03-29
Test the consistency of requirements defined in different locations

### LINKED ISSUE #1277: Test the consistency of requirements defined in multiple places
The problem described in #1276 is that our requirements are defined in multiple places.  While #1276 proposes to consolidate requirements in a single place, it's not clear if there is a reasonable way to do so.

An alternative to #1276 would be to keep the structure of the requirements as they are now, and create a test to make sure that the different places where requirements are defined are consistent with each other.  This would probably be more straightforward than consolidating the requirements into a single place, but there's a possibility that the tests that do these checks could end up being brittle.  This is probably one of those situations where we ought to do a tradeoff analysis, since I'm not sure what the best way to approach this would be.

namurphy:
This PR adds tests to check that the requirements defined in the `{build,docs,extras,install,tests}.txt` files are consistent with the requirements defined in `setup.cfg` and `pyproject.toml`.  Turns out that we can use `setuptools.config.read_configuration` to read in `setup.cfg`, and `toml.loads` to read in `pyproject.toml`.  Closes #1277.

One thing I needed to handle was that the `tests` and `docs` requirements in `setup.cfg` also include the `extras` requirements.

This PR also includes an explicit `import plasmapy` statement, so I'm wondering if #1447 could possibly be closed by this as well.  I checked and didn't find any `import plasmapy` lines in any of our other tests, so this would be helpful to add.


plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,140 @@
...
+        "setup.cfg",
+    ),
+    (
+        "tests",
+        requirements_from_txt["tests"],
+        "requirements/tests.txt",
+        requirements_from_setup_cfg["tests"] - requirements_from_txt["extras"],
+        "setup.cfg",
+    ),
+]
```
namurphy, line 110: I was thinking about refactoring this to reduce the number of parameters used in the test parametrization, but it'd be a lot of effort for not much gain in readability/maintainability.  In particular, this would be harder because of the special cases that the `tests` and `docs` requirements include the `extras` requirement in `setup.cfg`.

namurphy:
- [ ] Finish updating usage of `toml` to `tomli`, since they seem to have a different API.

namurphy:
@rocco8773's idea regarding the pre-commit test failure — try putting a restriction on which version of `click` can be used

plasmapy/tests/test_requirements.py
StanczakDominik: What about just compressing the first three lines to, for the last test case, just `"tests"`, and getting the other two lines in the function body?

plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,141 @@
...
+import pytest
+import setuptools
+import tomli
+
+from typing import Dict, Set
+
+import plasmapy
+
+base_directory = os.path.realpath(f"{plasmapy.__path__[0]}/..")
+requirements_directory = f"{base_directory}/requirements"
```
StanczakDominik, line 13: Just use pathlib, like this:

```suggestion
import pathlib
base_directory = pathlib.Path(__file__).parent.parent
requirements_directory = base_directory / "requirements"
```

this will then also work on Windows. I used `__file__` to dodge weirdness with pip installed versions.

plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,141 @@
...
+    Returns
+    -------
+    list of str
+        A `list` containing the lines of the requirements file,
+        excluding lines that do not start with an alphabetic character
+        (like comments).
+    """
+    filename = f"{requirements_directory}/{prefix}.txt"
+    with open(filename) as file:
+        lines_of_file = file.readlines()
```
StanczakDominik, line 38:
```suggestion
    file = (requirements_directory / prefix).with_suffix(".txt")
    lines_of_file = file.read_text().splitlines()
```

plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,141 @@
...
+
+def get_requirements_from_txt() -> Dict[str, Set[str]]:
+    """
+    Get the requirements from the ``.txt`` files in the ``requirements``
+    directory.
+    """
+    return {
+        prefix: read_requirements_txt_file(prefix, requirements_directory)
+        for prefix in requirements_prefixes
+    }
```
StanczakDominik, line 50: A point on design. I've been reading https://web.stanford.edu/~ouster/cgi-bin/book.php lately, and I can't help but notice that these few functions you've got here are rather shallow - they don't hide much complexity, if any. I get the fear of long functions, but I think these oneliners could be, well, inlined.

plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,141 @@
...
+
+def get_requirements_from_setup_cfg() -> Dict[str, Set[str]]:
+    """Get the requirements that are contained in ``setup.cfg``."""
+    configuration = setuptools.config.read_configuration(f"{base_directory}/setup.cfg")
+    return {
+        "docs": set(configuration["options"]["extras_require"]["docs"]),
+        "extras": set(configuration["options"]["extras_require"]["extras"]),
+        "install": set(configuration["options"]["install_requires"]),
+        "tests": set(configuration["options"]["extras_require"]["tests"]),
+    }
```
StanczakDominik, line 61:
```suggestion
   d = {
        key: set(configuration["options"]["extras_require"][key]) for key in "docs extras tests".split()
        }
    d["install"] = set(configuration["options"]["install_requires"])
    return d
```
Spares a line (modulo `black`, I guess!), and more clearly shows that `install` is a special case :)

plasmapy/tests/test_requirements.py
```diff
@@ -0,0 +1,141 @@
...
+import pytest
+import setuptools
+import tomli
+
+from typing import Dict, Set
+
+import plasmapy
+
+import pathlib
+base_directory = pathlib.Path(__file__).parent.parent
```
StanczakDominik, line 13: `__file__` is `test_requirements.py`, whose parent is `tests`, whose parent is `plasmapy`, which means we're missing one more parent, or, more succinctly:
```suggestion
base_directory = pathlib.Path(__file__).parents[2]
```

StanczakDominik (approved):
I have one single suggestion that fixes the CI problem, then it's a speedy merge :) I've tested that it does work nicely on adding a surplus package to `setup.cfg`.

StanczakDominik:
pre-commit.ci autofix

namurphy:
pre-commit.ci autofix

PR #1501, opened 2022-04-05
Test that ``plasmapy`` can be imported

namurphy:
This PR adds a simple test in our test suite that runs ``import plasmapy`` to make sure that it can be imported.

I decided to include this as a test rather than as a GitHub Action so that:
 - We are able to test this on platforms besides GitHub, including locally
 - We are able to straightforwardly test this on multiple operating systems and versions of Python.

We also currently have a [tox environment](https://github.com/PlasmaPy/PlasmaPy/blob/9e4cc617e56ce01fea1a5e356d25c659627b208c/tox.ini#L122) that can be used to perform this test.

### DIFF
--- a/.github/workflows/fortnightly.yml
+++ b/.github/workflows/fortnightly.yml
@@ -41,6 +41,21 @@ jobs:
           python: 3.8
           toxenv: py38-xarraydev
 
+        - name: Import PlasmaPy (Windows)
+          os: windows-latest
+          python: 3.8
+          toxenv: py38-minimal-pypi-import
+
+        - name: Import PlasmaPy (macOS X)
+          os: macos-latest
+          python: 3.8
+          toxenv: py38-minimal-pypi-import
+
+        - name: Import PlasmaPy (Ubuntu)
+          os: ubuntu-latest
+          python: 3.8
+          toxenv: py38-minimal-pypi-import
+
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


StanczakDominik:
IIRC the tox environment tries to import PlasmaPy from the command line, via something like 

```
python -c `import plasmapy`
```

And I think that's simpler, gives us the same core ("was PlasmaPy imported?") information with less additional conditions ("... and pytest was installed and didn't crash?") and less overhead.

So, what's the advantage here?

StanczakDominik (commented):
Meant to submit the above as a comment 😅

namurphy:
> IIRC the tox environment tries to import PlasmaPy from the command line, via something like
> 
> ```
> python -c `import plasmapy`
> ```
> 
> And I think that's simpler, gives us the same core ("was PlasmaPy imported?") information with less additional conditions ("... and pytest was installed and didn't crash?") and less overhead.
> 
> So, what's the advantage here?

Thanks for the feedback!  My motivations were to be able to do this locally without having to think about it, and have it done on different operating systems...though these aren't major motivators.  @rocco8773 also had some apprehensions about putting this in our test suite, and had preferred it to be a GitHub Action.  At the moment, I noticed that we don't do the `py38-minimal-pypi-import` `tox` environment in GitHub Actions.

Perhaps as a middle ground, I could change this by adding the `py38-minimal-pypi-import` environment for `tox` to GitHub Actions for a single OS for each PR, and then doing a grid for it for multiple OSes (or whatever the plural for OS is).  Would that work?

StanczakDominik:
Yeah, that sounds good! Also, pretty sure the plural for OS is just Docker.

StanczakDominik (approved):
👍🏻

namurphy:
@StanczakDominik — many thanks for the review!

Issue #1503, opened 2022-04-05
Update pre-commit to use the most recent version of `black`


namurphy:
One of the most recent releases of `black` made a change so that there will be no spaces surrounding the exponential operator `**` if the operands are simple.  Unfortunately this change would affect ∼70 files [...more power(s) to us?], which would need to be updated at the same time as this change.  It would still probably be worth making this change since `black` is now out of beta, and their FAQ now says that the formatting [will remain stable for releases in a calendar year](https://black.readthedocs.io/en/stable/faq.html#how-stable-is-black-s-style) starting in 2022.

This would probably lead to a some merge conflicts, but if we do this just before the 0.8.0 release, then we'd probably have a simpler time backporting any changes.

I did look to see if there was a configuration option that would let us disable this change, but wasn't able to find one.

PR #1504, opened 2022-04-06
Set maximum requirement for `click` in `pre-commit` configuration

namurphy:
The version of `black` that we currently use in `pre-commit` has started having an issue with the newest release of `click`.  This PR temporarily sets a maximum requirement for `click`, since updating to the next version of `black` is going to be really involved (#1503).  When we update to the next version of `black` in `pre-commit`, we should be able to drop this maximum requirement.

### DIFF
--- a/changelog/1504.trivial.rst
+++ b/changelog/1504.trivial.rst
@@ -1,3 +1,0 @@
-Set the maximum version of ``click`` to be less than ``8.1`` in the
-pre-commit_ configuration for compatibility with black_ version
-``21.12b0``.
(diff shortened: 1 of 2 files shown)


changelog/1504.trivial.rst
```diff
@@ -0,0 +1,3 @@
+Set the maximum version of ``click`` to be less than ``8.1`` in the
+pre-commit_ configuration for compatibility with black_ version
+``21.12b0``.
```
rocco8773, line 3: Since `click` isn't a package dependency and we're only changing the CI here (no changes to the `plasmapy` pkg), I'd argue we we don't need (and probably shouldn't have) a change log for this.

namurphy: I was on the fence about that, so I'm okay with deleting the changelog entry (which I just did).

rocco8773 (approved):

Issue #1153, opened 2021-05-25
Make `Particle` instances for elements/isotopes explicitly have a charge number of 0


namurphy:
One more `Particle` design decision that I regret!  😹  

The way `Particle` is set up now, there is a difference between `Particle("He-4")` and `Particle("He-4 0+)`.  The former lacks charge information, while the latter explicitly has zero charge.  
```Python
>>> Particle("He-4").charge
ChargeError: The charge of particle He has not been specified.
>>> Particle("He-4", Z=0).charge   # or Particle("He 0+")
<Quantity 0. C>
```
However, when we input an element, there is usually an implicit assumption that the charge number is zero.  I propose that we make `Particle("He-4")` assume a charge number of zero, and thus make it equivalent to `Particle("He-4 0+")`.

Before implementing this it'd be worth thinking about any adverse consequences, just in case.  For example, in nuclear reactions, it's often implicitly assumed that an isotope is a nuclide with a charge number equal to the atomic number.  This would be a breaking change, but I don't think that it would be a significant change.

namurphy:
I'm also wondering if it'd be worth creating a GitHub label for `Particle` design decisions that Nick regrets... 😅

StanczakDominik:
We need to generalize that latter part to <code> design decisions that <person> regrets. I know I've got mine.

PR #1514, opened 2022-04-14
Allow `Particle("Li").ionize()`, assuming meant as neutral atom

StanczakDominik:
Currently, if you have a `p = Particle("Li")` and try to `.ionize()` it, you get a `ChargeError` complaining about the charge not being specified.

However, I think a sane default here is to assume `p` is meant as a neutral atom, in which case  we can safely assume the charge number is zero. Whatcha think?

      docstrings.

### DIFF
--- a/changelog/1514.feature.rst
+++ b/changelog/1514.feature.rst
@@ -1,0 +1,1 @@
+`Particle("Li").ionize()` no longer results in a `ChargeError`. Instead, ionization of a neutral atom is assumed.
--- a/changelog/1514.feature.rst
+++ b/changelog/1514.feature.rst
@@ -1,1 +1,1 @@
-`Particle("Li").ionize()` no longer results in a `ChargeError`. Instead, ionization of a neutral atom is assumed.
+``Particle("Li").ionize()`` no longer results in a `~plasmapy.particles.exceptions.ChargeError`. Instead, ionization of a neutral atom is assumed.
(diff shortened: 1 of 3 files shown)


rocco8773:
I agree.  If you're trying to `.ionize()` a neutral particle without a arguments, then I think it's safe to assume an ionization of `+1` (i.e.  `Particle("Li").ionize() -> Particle("Li+")`).  In fact, the [documentation signature ](https://docs.plasmapy.org/en/latest/api/plasmapy.particles.particle_class.Particle.html#plasmapy.particles.particle_class.Particle.ionize)suggests that this would be the behavior.

namurphy:
> I agree. If you're trying to `.ionize()` a neutral particle without a arguments, then I think it's safe to assume an ionization of `+1` (i.e. `Particle("Li").ionize() -> Particle("Li+")`). In fact, the [documentation signature ](https://docs.plasmapy.org/en/latest/api/plasmapy.particles.particle_class.Particle.html#plasmapy.particles.particle_class.Particle.ionize)suggests that this would be the behavior.

Agreed, this would be the more sensible behavior.  This would be a subset of #1153, which might not happen before 0.8.0.  My current thinking is that we could merge this shortly, and then try to get to 0.8.0 soon.  Does that sound reasonable?

changelog/1514.feature.rst
```diff
@@ -0,0 +1 @@
+`Particle("Li").ionize()` no longer results in a `ChargeError`. Instead, ionization of a neutral atom is assumed.
```
namurphy, line 1:
```suggestion
``Particle("Li").ionize()`` no longer results in a `~plasmapy.particles.exceptions.ChargeError`. Instead, ionization of a neutral atom is assumed.
```

namurphy (approved):

namurphy (approved):

PR #1515, opened 2022-04-15
Make `Particle` instances for an atom without charge info default to neutral

### LINKED ISSUE #1153: Make `Particle` instances for elements/isotopes explicitly have a charge number of 0

namurphy:
Prior to this PR, doing something like `Particle("H")` led to the charge number being undefined.  For example `Particle("H").charge_number` would raise a `ChargeError`. The goal of this PR is to make  things like `Particle("H")` instead have an explicit charge number of 0.  If the charge number `Z` is not provided, then it will be assumed that `Z=0`.  

Unfortunately this change in behavior will break a bunch of things that will need to be fixed.  So far I've gotten the tests in `test_particle_class.py` to pass locally, though I ran into test collection errors when testing the whole package locally.  This will take more effort than I thought. 😅  It'll also be worth thinking through the different changes that arise.

Inspired by #1514. Closes #1153.

      docstrings.


namurphy:
Oh good...only 224 tests failed!  This one is gonna be some work 😅

PR #1509, opened 2022-04-11
Reduce number of warnings for nitpicky doc builds

namurphy:
This PR greatly reduces the number of warnings that arise during a "nitpicky" documentation build (i.e. `tox -e build_docs_nitpicky`).  On my computer, the number went from 4151 →  21 (i.e. a reduction of 99.5%).  My hope is that we'll soon be able to use this setting to check that we don't end up with broken reST links in pull requests so that they don't end up piling up over time again.

A bunch of the changes were related to `plasmapy.parameters` being disintegrated, the workaround reST substitutions defined at the end of `docs/common_links.rst` that show up in `rst_epilog` (which inflated the count), changelog entries for code objects that no longer exist, namespaces that were not fully spelled out, and typos.  The remaining ones on my computer are for the `whatsnew` page for 0.8.0, plus some files that are being worked on in other PRs, and a handful of others.

The changes I made were...
 - A whole bunch of reST formatting fixes
 - Defining `nitpick_ignore_regex` in `docs/conf.py` to specify role patterns and target patterns for things that the nitpicky doc build should ignore.  
 - Changing up `formulary.__init__` to avoid `contextlib` showing up in `__all__` (and adding another `flake8` ignore configuration for that file)
 - Adding in some comments 
 - Some minor edits & bibliography entries

PR #1518, opened 2022-04-15
Update docs and the use of `kwargs` in `fit_functions.Linear.curve_fit`

namurphy:
While going through #1509, I noticed that `kwargs` was accepted for `fit_functions.Linear.curve_fit`, but it wasn't used in the method when calling `linregress`.  This PR makes it so that the `kwargs` are passed to `linregress`. I also updated the reST link to be `scipy.stats.linregress`.

@rocco8773 — could you take a look at this to make sure the change is consistent with the intended behavior?  Thank you!


namurphy:
Awesome, thanks!

PR #1519, opened 2022-04-18
Specify allowed reST/Sphinx roles & directives for flake8-rst-docstrings

namurphy:
The [flake8-rst-docstrings](https://github.com/peterjc/flake8-rst-docstrings) extension to flake8 checks a bunch of things regarding reStructuredText (reST) formatting in our docstrings, including about whether or not there are any undefined [roles](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html) or [directives](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html).

Unfortunately, it appears to only know the roles/directives from basic reST, and not the roles/directives defined by Sphinx which are a superset of those from basic reST.  This means that we have to specify which reST role/directives are allowed in docstrings via the appropriate variables in `setup.cfg` (`rst-roles` and `rst-directives`).  This PR adds most of the remaining Sphinx roles/directives to these two variables so that they can be used in docstrings. 

This extension does not check our narrative docs.  Some of the added roles/directives might never end up being needed in docstrings, but I put them in since they're valid.  Roles/directives from Sphinx extensions may still need to be added here in the future.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -186,9 +186,11 @@ exclude =
     setup.py,
     .jupyter
 # Use rst-roles and rst-directives to list roles and directives from
-# Sphinx extensions so that they don't get flagged when using
+# Sphinx and its extensions so that they don't get flagged when using
 # flake8-rst-docstrings.
 rst-roles =
+    abbr
+    any
     attr
     cite
     cite:ct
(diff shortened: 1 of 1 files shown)


namurphy:
@pheuer — would you be able to give this a review?  If it looks good, feel free to merge it, since that would fix the problem with linters that you mentioned in #974.  Thanks!

pheuer (approved):
@namurphy This looks straightforward - thanks

PR #1527, opened 2022-04-20
Base bibliography html anchors on citation key

namurphy:
Prior to this PR, the html anchors used for each reference in our bibliography was of a form like `id20`, where the number corresponds to the number in the bibliography.  While links created using roles like `:cite:p:` and `:cite:t:` will be updated automatically, external links that point to one bibliography entry at one point in time are not guaranteed to point to that same entry at a later time (for example, if more bibliography entries got added).  This is also true for hard links to bibliography entries that were included in Markdown cells in our example notebooks.

This PR sets the html anchors to the citation key, with colons replaced by dashes.  This provides an extra level of stability, as the anchors won't change as long as the citation key doesn't change.  If a citation key does end up changing, then we would be more likely to catch a broken anchor from using `make linkcheck`.

### DIFF
--- a/docs/development/install_dev.rst
+++ b/docs/development/install_dev.rst
@@ -59,7 +59,7 @@ Conda
 -----
 
 To set up a development environment for PlasmaPy, we strongly recommend
-the `Anaconda distribution <https://www.anaconda.com/products/individual>`_.
+the `Anaconda distribution <https://www.anaconda.com/products/distribution>`_.
 
 Activate Anaconda
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


namurphy:
Here's how the link with anchor looks for Alvén (1942) for the doc build for this pull request:

https://plasmapy--1527.org.readthedocs.build/en/1527/bibliography.html#alfven-1942

docs/development/install_dev.rst
```diff
@@ -59,7 +59,7 @@ Conda
 -----
 
 To set up a development environment for PlasmaPy, we strongly recommend
-the `Anaconda distribution <https://www.anaconda.com/products/individual>`_.
+the `Anaconda distribution <https://www.anaconda.com/products/distribution>`_.
```
namurphy, line 62: Fixing a permanent redirect from doing a `make linkcheck`.

rocco8773 (approved):

PR #1528, opened 2022-04-23
Create `ParticleListLike` typing construct and "particle-list-like" term

namurphy:
This pull request is to create `plasmapy.particles.particle_collections.ParticleListLike` as a typing construct that could be used in `particle_input`.  This PR follows up on #985.

Eventually I'm wondering if it would make sense to create `plasmapy.particles.typing` to contain things like this, but not in this PR.

### DIFF
--- a/changelog/1528.feature.rst
+++ b/changelog/1528.feature.rst
@@ -1,0 +1,2 @@
+Created the |ParticleListLike| typing construct and added
+:term:`particle-list-like` to the |glossary|.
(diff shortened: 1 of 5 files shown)


StanczakDominik (approved):
There is a parallel world where this gets merged into a feature branch for the `particle_input` update, rather than into main, where it won't yet see much use.

So long as it keeps chugging along, though, I'm fine with it!

PR #1529, opened 2022-04-25
Hollweg Dispersion Solver Bug Fix

Sjbrownian:

      docstrings.


plasmapy/dispersion/numerical/hollweg_.py
```diff
@@ -310,12 +308,14 @@ def hollweg(
...
+    coefficients = np.array([c3, c2, c1, c0], ndmin=3)
     nroots = coefficients.shape[0] - 1  # 3
     nks = coefficients.shape[1]
-    roots = np.empty((nroots, nks), dtype=np.complex128)
+    nthetas = coefficients.shape[2]
+    roots = np.empty((nroots, nks, nthetas), dtype=np.complex128)
     for ii in range(nks):
-        roots[:, ii] = np.roots(coefficients[:, ii])[:]
+        for jj in range(nthetas):
+            roots[:, ii, jj] = np.roots(coefficients[:, ii, jj])[:]
```
StanczakDominik, line 318: Do we actually need the `[:]` at the end?

StanczakDominik (commented):
Looks like a pretty simple fix. Could you add a test, I guess in `plasmapy/dispersion/tests/test_dispersion.py`, to ensure the previous bug is fixed?

plasmapy/dispersion/numerical/hollweg_.py
Sjbrownian: I see no difference in the output when the `[:]` is removed. Do you remember why this was added @rocco8773 ?

rocco8773:
```suggestion
            roots[:, ii, jj] = np.roots(coefficients[:, ii, jj])
```

Yes, the `[:]` at the end is not required in this case.

plasmapy/dispersion/numerical/tests/test_hollweg_.py
```diff
@@ -256,6 +256,14 @@ def test_z_mean_override(self, kwargs, expected):
...
                 {"shape": (3,)},
             ),
+            (
+                {
+                    **_kwargs_single_valued,
+                    "k": [1, 2, 3] * u.rad / u.m,
+                    "theta": [50, 77] * u.deg,
+                },
+                {"shape": (3, 2)},
+            ),
```
rocco8773, line 266: Should also add a test case where `k` is single valued and `theta` is an array.

Sjbrownian:
@StanczakDominik I've added a test in `plasmapy/dispersion/numerical/tests/test_hollweg_.py` to account for multiple `k` and `theta` values and I also made an adjustment to a PhysicsWarning in `plasmapy.dispersion/numerical/hollweg_.py` which was causing issues. Would you be able to do another quick review when you have the chance? No rush :).

plasmapy/dispersion/numerical/hollweg_.py
```diff
@@ -262,7 +262,7 @@ def hollweg(
     if not (theta.ndim == 0 or theta.ndim == 1):
```
rocco8773, line 262:
```suggestion
    if theta.ndim not in (0, 1):
```

This suggestion is a little out of scope, and nothing is wrong with the original code, but this is the more Pythonic way.

changelog/1529.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix bug in the :file:`hollweg_.py` module that didn't allow for theta array inputs.
```
rocco8773, line 1:
```suggestion
Fixed bug in :func:`~plasmapy.dispersion.numerical.hollweg_.hollweg`
that did not allow for arguments ``theta`` and ``k`` to simultaneously
be arrays.
```

rocco8773 (commented):
Would it be possible to at cases to `test_hollweg1999_vals` that give arrays for `k` and `theta`?  You don't need to come up with new values to test, just use the already tested values but in array form.

plasmapy/dispersion/numerical/tests/test_hollweg_.py
Sjbrownian: Added 👍

rocco8773 (approved):

PR #1530, opened 2022-04-25
Add GitHub Action for codespell

namurphy:
This PR follows up on #1493 by adding a GitHub Action for [codespell](https://github.com/codespell-project/codespell).  This is very similar to the GA added in https://github.com/HinodeXRT/xrtpy/pull/37.  

Codespell works by looking for common misspellings rather than by looking for words that aren't in a dictionary, so it has a low (but nonzero) number of false positives.  I was hesitant about including codespell in our pre-commit configuration because of the occasional false positives and the need for human intervention, but I think a GA would be appropriate so long as any failure messages describe what to do to fix any false positives.  

 - [x] Make tox environment print out how to fix false positives and how to run codespell locally to write changes
 - [x] Update testing guide to reflect these changes

### DIFF
--- a/changelog/1530.doc.rst
+++ b/changelog/1530.doc.rst
@@ -1,0 +1,3 @@
+Described the `GitHub Action`_ for
+`codespell <https://github.com/codespell-project/codespell>`__
+in the |testing guide|.
--- a/changelog/1530.doc.rst
+++ b/changelog/1530.doc.rst
@@ -1,3 +1,2 @@
-Described the `GitHub Action`_ for
-`codespell <https://github.com/codespell-project/codespell>`__
+Described the GitHub Action for `codespell <https://github.com/codespell-project/codespell>`__
 in the |testing guide|.
--- a/changelog/1530.trivial.rst
+++ b/changelog/1530.trivial.rst
@@ -1,3 +1,3 @@
-Added a `GitHub Action`_ for `codespell
+Added a GitHub Action for `codespell
 <https://github.com/codespell-project/codespell>`__, and updated the
 corresponding tox_ environment to print out contextual information.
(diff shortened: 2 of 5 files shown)


StanczakDominik:
> I was hesitant about including codespell in our pre-commit configuration because of the occasional false positives and the need for human intervention, but I think a GA would be appropriate so long as any failure messages describe what to do to fix any false positives.

That's also what I've converged to over time 👍🏻 

Can you add a typo somewhere so we can see how this works?

namurphy:
Here's an example of how it looks when it finds a misspelling, with some extra context supplied before and after...
```
Codespell finds typos in source code. Rather than checking if each word
matches a dictionary entry, it looks for a set of common misspellings
in order to reduce the number of false postives.

./tox.ini:139: postives ==> positives
ERROR: InvocationError for command /home/runner/work/PlasmaPy/PlasmaPy/.tox/codespell/bin/codespell . (exited with code 65)

After codespell has been installed locally (`pip install codespell`),
running the command `codespell -i 2 -w` will interactively go through
misspellings and suggest one or more replacements.
```

changelog/1530.doc.rst
```diff
@@ -0,0 +1,3 @@
+Described the `GitHub Action`_ for
+`codespell <https://github.com/codespell-project/codespell>`__
+in the |testing guide|.
```
namurphy, line 3:
```suggestion
Described the GitHub Action for `codespell <https://github.com/codespell-project/codespell>`__
in the |testing guide|.
```

changelog/1530.trivial.rst
```diff
@@ -0,0 +1,3 @@
+Added a `GitHub Action`_ for `codespell
```
namurphy, line 1:
```suggestion
Added a GitHub Action for `codespell
```

StanczakDominik (approved):
👍🏻

namurphy:

PR #1531, opened 2022-04-26
Faster distributions

userr2232:
Partially takes care of speed issues mentioned in PR #816  
- [X] Reduced execution time by avoiding calculations with units while keeping unit validation. I also added a file `plasmapy/utils/units_definitions.py` because precomputing units also helps to speedup the program (see: https://docs.astropy.org/en/stable/units/index.html#performance-tips). Using `.si.value` creates a copy of a Quantity and should be avoided if a copy is not needed `to_value()` should be used instead. 
Speedup demo: https://colab.research.google.com/drive/1KLRp7jc2_TicsjUz0PROqGmTBaGcOllm#scrollTo=E85v9bWw4qUv


StanczakDominik (dismissed):
Hey, awesome! Great work, I had no idea about that `<<` operator and I still don't quite get the difference between `.to` and `.to_value`, but the changes look logically consistent and the speed up is lovely!

Some merge conflicts in there now, but I'm sure I can fix them and merge this soon.

Thanks a lot for doing this!

userr2232:
Hi @StanczakDominik, thanks for the kind comments! `to` and `.to_value` make conversions (and validate units) but `.to_value` removes the units (and returns a scalar or an ndarray instead of a Quantity).

PR #1546, opened 2022-05-05
Fix to Z dependence in `fundamental_collision_freq`. 

### LINKED ISSUE #1545: Missing factor of Z in fundamental collision frequency
Hi, I think the function `formulary.collisions.fundamental_electron_collision_freq` is calling `collision_frequency` incorrectly by using the electron density `n_e` rather than `n_i`. In the input description for `collision_frequency` it specifies that for electron-ion collisions `n` should be the ion density, and in `fundamental_electron_collision_freq` it explicitly sets the species as `[ion, "e-"]`.

I noticed this problem while checking the Spitzer thermal conductivity and spotting that the Z dependence comes out wrong. The original Spitzer-Harm paper PR 89, 5 (1953) gives the Z dependence in the conductivity K as delta_T * epsilon / Z. In the current implementation, however, the Z dependence in K comes from kappa_hat (giving delta_T * epsilon) and 1/collision_frequency, where it's proportional to n*b_perp^2, which is proportional to n_e*Z^2. That gives a Z dependence of delta_T * epsilon/Z^2, which will go very wrong for high Z. If instead we call `collision_frequency` using n_i=n_e/Z, it should give n_i*Z^2=n_e*Z, giving a final Z dependence of delta_T * epsilon/Z, as required.

https://github.com/PlasmaPy/PlasmaPy/blob/2a138014b5789a816dc89c87859b5107b62ab176/plasmapy/formulary/collisions.py#L1195

ChrisArran:
Changed from calling collision_frequency using n_e to calling with n_i, to reflect calculating for electron-ion collisions. This should fix the Z dependence to match e.g. Spitzer thermal conductivity.

Closes #1545

### DIFF
--- a/changelog/1546.bugfix.rst
+++ b/changelog/1546.bugfix.rst
@@ -1,4 +1,4 @@
 Fixed the ``Z`` dependence in
-`~plasmapy.formulary.collisions.fundamental_collision_freq`, by replacing
-``n_e`` with ``n_i`` while calling
+`~plasmapy.formulary.collisions.fundamental_electron_collision_freq`,
+by replacing ``n_e`` with ``n_i`` while calling
 `~plasmapy.formulary.collisions.collision_frequency`.
--- a/changelog/1546.bugfix.rst
+++ b/changelog/1546.bugfix.rst
@@ -1,1 +1,4 @@
-Fixed the Z dependence in formulary.collisions.fundamental_collision_freq, by replacing n_e with n_i while calling collision_frequency. Closes Issue #1545.
+Fixed the ``Z`` dependence in
+`~plasmapy.formulary.collisions.fundamental_collision_freq`, by replacing
+``n_e`` with ``n_i`` while calling
+`~plasmapy.formulary.collisions.collision_frequency`.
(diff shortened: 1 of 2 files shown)


rocco8773:
@namurphy Can you do a double check on this to confirm that our assessment is correct?

@ChrisArran I agree with you an this. In the `collision_rate` formulation `n` is the density of the field particles, and in this case the field particles are the ions and the test particles are the electrons.  I also double check the formulation with the NRL formulary and [Callen](https://docs.plasmapy.org/en/latest/bibliography.html#callen-unpublished), which are referenced by `fundamental_electron_collision_rate`, and they both use `n_i` instead of `n_e`.

@ChrisArran Could you please also add a [bugfix changelog](https://docs.plasmapy.org/en/latest/development/changelog_guide.html) entry for this PR.

ChrisArran:
Sure, thanks for getting on this so quickly! - I've added the changelog in a second commit, hope it's okay. :)

All the best,
Chris

rocco8773 (approved):

PR #1548, opened 2022-05-09
Undefined Particle mass now returns np.nan instead of raising exception

### LINKED ISSUE #1016: Make undefined Particle attributes return NaNs instead of raising exceptions
Right now, if we try to access an undefined `Particle` attribute, an exception is raised.
```Python
>>> tau_neutrino = Particle("tau neutrino")
>>> tau_neutrino.mass
MissingAtomicDataError: the mass of nu_tau is not available.
```
The problem with this is that if we're looping over a bunch of particles, we could end up having to deal with the exception as a special case.  I propose that instead of raising an exception for situations like this, we return `numpy.nan`, multiplied by the expected units when appropriate.  We can then let `numpy.nan` take care of keeping track of whether it is defined or not.

Fixing this issue would make #948 a little bit easier to address.  

Returning `numpy.nan` in the appropriate units would also make the behavior of `Particle` and `CustomParticle` more consistent.

ChrisMoriarty:
For particles where the mass is unavailable (e.g., for neutrinos or elements with no standard atomic weight), we now return numpy.nan multiplied by the expected units.  I also added a test called "test_undefined_mass".

This PR is being used to help me get acquainted with PlasmaPy, in order to start helping with code reviews (as part of the Engage an Engineer effort in Central Engineering). 

Closes #1016


changelog/1548.feature.rst
```diff
@@ -0,0 +1,3 @@
+Updated the return value of the ``mass`` method of `~plasmapy.particles.particle_class.Particle`
+to return np.nan, rather than raising a MissingParticleDataError, when the mass is unavailable
+(e.g. for neutrinos or elements with no standard atomic weight).
```
namurphy, line 3:
```suggestion
Updated the return value of the `~plasmapy.particles.particle_class.Particle.mass`
attribute of |Particle| to return |nan| kg, rather than a |MissingParticleDataError|, when 
the mass is unavailable (e.g. for neutrinos or elements with no standard atomic weight).
```
This should make it link to the docstring for `Particle.mass`.  I also applied some reST substitutions that we've defined in `docs/common_links.rst`.

docs/about/credits.rst
```diff
@@ -103,6 +103,7 @@ in parentheses are `ORCID author identifiers <https://orcid.org>`__.
 * :user:`Anna Lanteri <alanteriBW>`
 * :user:`Leo Murphy <LeoMurphyWM24>`
 * :user:`Reynaldo Rojas Zelaya <userr2232>`
+* :user:`Chris Moriarty <ChrisMoriarty>` (:orcid:`0000-0002-1757-7573`)
```
namurphy, line 106: Always nice to see a new person added to this list!

plasmapy/particles/particle_class.py
```diff
@@ -1044,11 +1044,9 @@ def mass(self) -> u.kg:
...
         value for the particle's mass.
 
-        Raises
-        ------
-        `~plasmapy.particles.exceptions.MissingParticleDataError`.
-            If the mass is unavailable (e.g., for neutrinos or elements
-            with no standard atomic weight.
+        For particles where the mass is unavailable (e.g., for neutrinos or
+        elements with no standard atomic weight), this attribute will return
+        numpy.nan multiplied by the expected units.
```
namurphy, line 1049:
```suggestion
        |nan| kg.
```

plasmapy/particles/particle_class.py
```diff
@@ -1146,8 +1144,9 @@ def mass_energy(self) -> u.J:
         If the particle is an isotope or nuclide, return the mass energy
         of the nucleus only.
 
-        If the mass of the particle is not known, then raise a
-        `~plasmapy.particles.exceptions.MissingParticleDataError`.
+        For particles where the mass is unavailable (e.g., for neutrinos or
+        elements with no standard atomic weight), this attribute will return
+        numpy.nan multiplied by the expected units.
```
namurphy, line 1149:
```suggestion
        |nan| kg.
```

plasmapy/particles/tests/test_atomic.py
```diff
@@ -515,8 +515,7 @@ def test_incorrect_units(self):
             reduced_mass("N", 6e-26 * u.l)
 
     def test_missing_atomic_data(self):
-        with pytest.raises(MissingParticleDataError):
-            reduced_mass("Og", "H")
+        assert reduced_mass("Og", "H") == np.nan
```
namurphy, line 518:
```suggestion
        assert u.isclose(reduced_mass("Og", "H"), np.nan * u.kg, equal_nan=True)
```
It turns out that `np.nan == np.nan` returns `False` (sigh!), so we have to use the `equal_nan` keyword in `u.isclose` or `np.isclose`.

namurphy (changes requested):
Thank you for doing this!  It looks really close.  I had a few reST formatting suggestions and one other suggestion for a test, but we could probably get this merged soon.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -465,7 +465,7 @@
             "element": None,
             "isotope": None,
             "isotope_name": InvalidElementError,
-            "mass": MissingParticleDataError,
+            "mass": np.nan * u.kg,
```
namurphy, line 468: I think we'll need to use `equal_nan=True` in line 551 below too.

namurphy:
...should also update the comment around line 530 in `test_particle_class.py`.

Issue #1542, opened 2022-05-02
Create value test for `hollweg` using arrays for `k` and `theta`.


rocco8773:
Modify `test_hollweg1999_vals` or create a new `test_vals` to test `hollweg` values when given arrays for both `k` and `theta`.  Follow up to PR #1529.

PR #1549, opened 2022-05-09
Hollweg values test using arrays for k and theta

Sjbrownian:
- [ ] If adding new functionality, I have added tests and docstrings.

Closes #1542


Sjbrownian:
@rocco8773 Hi Erik, I've added some tests (`test_vals`) for dealing with array arguments for k and theta in hollweg but I'm getting an error when using `assert`. I've spent some time messing around with it to see if I could figure out why the assert was failing but I'm unsure what's causing it. Would you be able to take a look sometime?

I have finals to focus on for the next week or two so I may not be at the next meeting. If you feel the issue isn't an easy fix and we should discuss in person in two weeks or so I'm fine with that as well.

plasmapy/dispersion/numerical/tests/test_hollweg_.py
```diff
@@ -116,6 +116,81 @@ def test_warning(self, kwargs, _warning):
...
+                    "fast_mode": [3406.43522969 + 0.0j, 2278.76967883 + 0.0j],
+                    "alfven_mode": [2144.81200575 + 0.0j, 2138.00403666 + 0.0j],
+                    "acoustic_mode": [0.11044097 + 0.0j, 0.07358991 + 0.0j],
+                },
+            ),
+            (
+                {
+                    **_kwargs_single_valued,
+                    "k": np.logspace(-7, -2, 2),
+                    "theta": [86, 87, 88],
```
rocco8773, line 145:
```suggestion
                    "theta": [86, 87, 88] * u.deg,
```

plasmapy/dispersion/numerical/tests/test_hollweg_.py
```diff
@@ -116,6 +116,81 @@ def test_warning(self, kwargs, _warning):
...
+                    "k": np.logspace(-7, -2, 2) * u.rad / u.m,
+                },
+                {
+                    "fast_mode": [2.62911663e-02 + 0.0j, 2.27876968e03 + 0.0j],
+                    "alfven_mode": [7.48765909e-04 + 0.0j, 2.13800404e03 + 0.0j],
+                    "acoustic_mode": [0.00043295 + 0.0j, 0.07358991 + 0.0j],
+                },
+            ),
+            (
+                {**_kwargs_single_valued, "theta": [87, 88]},
```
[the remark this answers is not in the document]
Sjbrownian: Ah ok that makes sense, thank you. I have all the checks passing now and I think I have coverage for each type of case that can arise with k and theta array arguments. Does this look good to merge?

plasmapy/dispersion/numerical/tests/test_hollweg_.py
```diff
@@ -116,6 +116,81 @@ def test_warning(self, kwargs, _warning):
         with pytest.warns(_warning):
             hollweg(**kwargs)
 
+    @pytest.mark.parametrize(
+        "kwargs, expected",
+        [
+            (  # array input for k, single value for theta
+                {
+                    **_kwargs_single_valued,
+                    "k": np.logspace(-7, -2, 2) * u.rad / u.m,
```
rocco8773, line 125: Why are you using negative values for `k` here?  Since we're using `k` and `theta` as input arguments, I would think we would require `k > 0`.

Sjbrownian: It's `np.logspace` so it should range from [10^-7,10^-2] and be positive right?

rocco8773: ah...I see...my brain automatically read `np.linspace`

rocco8773 (approved):
This looks good!  Thank you for doing this.

I just made some minor tweaks to make a few things more explicit.  Once I confirm the docs build correctly, I'll go ahead and merge the PR.

rocco8773 (approved):

PR #1556, opened 2022-05-15
added examples in boris_push docstring

MKastek:

      docstrings.

pre-commit.ci autofix


plasmapy/simulation/particle_integrators.py
```diff
@@ -36,6 +36,29 @@ def boris_push(x, v, B, E, q, m, dt):
...
+    >>> for _ in range(150):
+    >>>     boris_push(x = x_0, v = v_0, B=B, E=0, q=1.0, m=1.0, dt=0.01)
+    >>>     x.append(x_0[0,0])
+    >>>     y.append(x_0[0,1])
+    >>> plt.plot(x,y)
+    >>> plt.xlabel('x [m]')
+    >>> plt.ylabel('y [m]')
+    >>> plt.title(r"Single particle motion without $\vec{E}$ field")
+    >>> plt.grid()
+    >>> plt.gca().set_aspect('equal')
```
StanczakDominik, line 60: Rather than plotting, which is not going to display in rendered docs at https://docs.plasmapy.org/en/stable/api/plasmapy.simulation.particle_integrators.boris_push.html#plasmapy.simulation.particle_integrators.boris_push and is best kept in jupyter notebooks such as https://docs.plasmapy.org/en/stable/notebooks/ExB_drift.html, perhaps we can instead display a few examples? This chains into something I also just realized...

plasmapy/simulation/particle_integrators.py
```diff
@@ -36,6 +36,29 @@ def boris_push(x, v, B, E, q, m, dt):
...
+    >>> import matplotlib.pyplot as plt
+    >>>
+    >>> x_0 = np.array([0.0, 0.0, 0.0])
+    >>> v_0 = np.array([5.0, 0.0, 0.0])
+    >>> B = np.array([0.0, 0.0, -5.0])
+    >>> x = []
+    >>> y = []
+    >>>
+    >>> for _ in range(150):
+    >>>     boris_push(x = x_0, v = v_0, B=B, E=0, q=1.0, m=1.0, dt=0.01)
```
StanczakDominik, line 52: So I just realized that the `boris_push` implementation we've got here only modifies the input `x, v` arrays. This is good for performance but is dreadful for quickly displaying how it works. I think adding an `inplace = False` keyword argument would be okay both for docs and for future developments here (for example I have a couple ideas on doing this through jax or some such optimizer...), if you'd be up for it!

And then, rather than having an embedded minimal simulation here, I think it'd be better to show how how to use the boris pusher:

1. how it returns the x, v arrays in the next time step (the essence of what it does) 
2. how it modifies the x, v arrays by default (show them before/after)
3. how, for B parallel to V, it does nothing;
4. what it does for B perpendicular to V;
5. how for nonzero E and zero B it reduces to the leapfrog algorithm.

Also, to easily do this from the IPython shell, try the `%doctest_mode` magic command: 

```
[ins] In [1]: %doctest_mode
Exception reporting mode: Plain
Doctest mode is: ON
>>> 2 + 2
4
```

StanczakDominik (changes requested):

StanczakDominik:
So, uh, I guess this one issue was a bit under-specified 😅

plasmapy/simulation/particle_integrators.py
MKastek: Ok I will work on that and prepare new version in accordance with your suggestions.

MKastek:
pre-commit.ci autofix

plasmapy/simulation/particle_integrators.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 
 
-def boris_push(x, v, B, E, q, m, dt):
+def boris_push(x, v, B, E, q, m, dt, inplace=False):
```
StanczakDominik, line 12: On further thought, so as to not make this a breaking change and keep the current behavior the same, let's keep the default as it is. Those jax ideas etc will still work, just with a different default, which is fine, I think.

```suggestion
def boris_push(x, v, B, E, q, m, dt, inplace=True):
```

I'll circle back and modify the docstrings accordingly...

plasmapy/simulation/particle_integrators.py
```diff
@@ -36,6 +36,69 @@ def boris_push(x, v, B, E, q, m, dt):
     dt : `float`
         Timestep, in SI (second) units.
 
+    inplace : 'bool'
+        If True, v and x array are changed with boris_push, default False.
```
StanczakDominik, line 40: Expanded on this a little:
```suggestion
        If set, data in the v and x arrays is changed during execution.
        Defaults to True for performance reasons.
        If False, this function returns `None`.
```

plasmapy/simulation/particle_integrators.py
```diff
@@ -36,6 +36,69 @@ def boris_push(x, v, B, E, q, m, dt):
...
+    v : `~numpy.ndarray`
+        Particle velocity after Boris push algorithm, in SI (meter/second) units.
+
+    Examples
+    --------
+    >>> B = np.array([[0.0, 0.0, 5.0]])
+    >>> E = np.array([[0.0, 0.0, 0.0]])
+    >>> x_t0 = np.array([[0.0, 0.0, 0.0]])
+    >>> v_t0 = np.array([[5.0, 0.0, 0.0]])
+    >>> x_t1, v_t1 = boris_push(x = x_t0, v = v_t0, B = B, E = E, q = 1.0, m = 1.0, dt = 0.01)
```
StanczakDominik, line 56:
```suggestion
    >>> x_t1, v_t1 = boris_push(x = x_t0, v = v_t0, B = B, E = E, q = 1.0, m = 1.0, dt = 0.01, inplace=False)
```

we don't technically need to remove `inplace` from the rest.

StanczakDominik (changes requested):
Hey, finally circled back to looking into this! I'm sorry for the long delay, a lot has been happening, as you likely well know :) Looks great, I do have one minor suggestion though - changing the default back to the previous one.

If you agree with the proposed changes, we can merge this pretty quickly. It would still need a changelog entry, but besides that, it looks pretty much done.

StanczakDominik:
I also think merging the suggestions and re-triggering CI might get rid of some of the noise from the Documentation test 😅

MKastek:
> I also think merging the suggestions (or doing anything to the branch, really) and re-triggering CI might get rid of some of the noise from the Documentation test 😅

Unfortunately, I continue to get errors related to the documentation :(

StanczakDominik:
Let's see if that fixes things. The doc build seems to have been an unrelated issue fixed elsewhere (that's distributed project development for ya, sorry I missed it! 😅).

I also adjusted the doctest values for a few changes where they started differing from what we actually got.

StanczakDominik (approved):
NOW this should do it :)

PR #1557, opened 2022-05-16
Add `flake8-implicit-str-concat` and `flake8-mutable` extensions to linters

namurphy:
First, this PR adds [flake8-implicit-str-concat](https://github.com/flake8-implicit-str-concat/flake8-implicit-str-concat), which checks for situations like the following (which are sometimes created when black reformats a file):
```Python
    raise ValueError(
        "original first line" "previously was the second line"
    )
```

Second, this PR adds [flake8-mutable](https://github.com/ebeweber/flake8-mutable), which checks that functions do not have mutable default arguments, since that can lead to unexpected behavior.

### DIFF
--- a/plasmapy/diagnostics/charged_particle_radiography.py
+++ b/plasmapy/diagnostics/charged_particle_radiography.py
@@ -1094,7 +1094,7 @@ class Tracker:
 
         self._log("Run completed")
 
-        self._log(f"Fraction of particles tracked: {self.fract_tracked*100:.1f}%")
+        self._log(f"Fraction of particles tracked: {self.fract_tracked:.1%}")
 
         self._log(
             "Fraction of tracked particles that entered the grid: "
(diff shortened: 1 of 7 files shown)


plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -1094,7 +1094,7 @@ def run(
 
         self._log("Run completed")
 
-        self._log("Fraction of particles tracked: " f"{self.fract_tracked*100:.1f}%")
+        self._log(f"Fraction of particles tracked: {self.fract_tracked*100:.1f}%")
```
rocco8773, line 1097:
```suggestion
        self._log(f"Fraction of particles tracked: {self.fract_tracked:.1%}")
```

namurphy: Whoa!  I like f-strings even more!

rocco8773 (approved):
Looks perfectly reasonable to me!!! :)

PR #1558, opened 2022-05-17
Add `flake8-simplify` to linters

namurphy:
This PR adds [`flake8-simplify`](https://github.com/MartinThoma/flake8-simplify) as a flake8 extension, and implements most of the suggested changes.

### DIFF
--- a/plasmapy/formulary/parameters.py
+++ b/plasmapy/formulary/parameters.py
@@ -19,6 +19,8 @@ Functions to calculate fundamental plasma parameters.
 
 __all__ = []
 
+import contextlib
+
 from plasmapy.utils.decorators import deprecated
 from plasmapy.utils.exceptions import PlasmaPyFutureWarning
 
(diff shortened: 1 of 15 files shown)


plasmapy/formulary/parameters.py
```diff
@@ -54,17 +56,11 @@
...
-    try:
+    with contextlib.suppress(ValueError):  # ValueError if name was not in __aliases__
         __aliases__.remove(name)
-    except ValueError:
-        # name was not in __aliases__
-        pass
 
-    try:
+    with contextlib.suppress(ValueError):  # ValueError if name was not in __all__
         __all__.remove(name)
```
StanczakDominik, line 63: These could be the same block.

StanczakDominik (approved):
LGTM!

plasmapy/formulary/parameters.py
namurphy: I had wondered the same thing, but if there's an exception raised in the first line of the [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) block, the second line won't be run.  That would have been nice, though.

namurphy:
Many thanks for the quick review!

plasmapy/formulary/parameters.py
StanczakDominik: Oh, huh, didn't know that! Then there really is no difference between the try block and the suppress call.

PR #1578, opened 2022-05-31
Created magnetic_skeleton file

haman80:

      docstrings.


plasmapy/analysis/magnetic_skeleton.py
```diff
@@ -0,0 +1,1056 @@
...
+    Returns
+    -------
+    array_like
+        A 1 by 3 numpy array which represents the trilinearly
+        approximated vector field value calculated at the given
+        location.
+
+    """
+    # Find the cell
+    # Maybe use numpy.argwhere or numpy.where
```
StanczakDominik, line 201: Good catch that you could use those! I'll fire up a debugger by Sunday (a bit tired atm) and try to optimize that for you.

plasmapy/analysis/magnetic_skeleton.py
```diff
@@ -0,0 +1,1056 @@
...
+        if (y[0][j][0] <= loc[1]) and (loc[1] <= y[0][j + 1][0]):
+            cell[1] = j
+            break
+    for k in range(len(z) - 1):
+        if (z[0][0][k] <= loc[2]) and (loc[2] <= z[0][0][k + 1]):
+            cell[2] = k
+            break
+    # Estimate
+    tlApprox = trilinear_approx(vspace, cell)
+    return tlApprox(loc[0], loc[1], loc[2])
```
StanczakDominik, line 218: Note: I haven't yet read past this point.

StanczakDominik (changes requested):
Hey, got a partial review going! This looks pretty good. I do have some questions, mostly because without the context (and I've been out of context here for a few months now, so you can basically treat me  as a new user 😄 ) it's pretty hard to guess what the classes themselves do. There's also a bunch of minor refactors I'd suggest and the most important question is, are you really returning 1x3 numpy arrays and whether we can't replace them with simple tuples. Let's discuss that!

haman80:
NonZeroDivergence() error has to be modified to a warning in nullpoint.py in order for magnetic_skeleton.py to run without an error. The reason is due to the fact that nullpoint.py will estimate the divergence of magnetic fields at the location of the null, and sometimes even though, analytically, the divergence is zero, the estimation will not be close enough and, therefore, we will get an error.

namurphy:
Thank you, @StanczakDominik!  It turns out that today is the last day for @haman80's internship, and our plan was to merge this into the `magnetic-topology` feature branch so that I can continue to work on it (through probably at a slower pace!).  I'll create an issue to address the comments that you raised here, which I think were very helpful.  

And @haman80 — thank you for all of your hard work this past year!  I'm incredibly grateful for having had the chance to work with you which I really enjoyed, and I'm super excited to be able to use these tools in the future.  Thank you again!

namurphy:
pre-commit.ci autofix

PR #1582, opened 2022-06-01
Update to newest version of `black` for `pre-commit`

### LINKED ISSUE #1503: Update pre-commit to use the most recent version of `black`

namurphy:
This PR upgrades `black` to version 22.3.0 in our `pre-commit` configuration, and updates our files to meet this new formatting.  The reason why this pull request affects as many lines as it does is that there was a formatting change for `black` that removes spaces around exponentiation operators in straightforward enough cases: `a ** b` → `a**b`.  I believe their reasoning was because the exponentiation operator is usually first in the order of operations.

Closes #1503.


namurphy:
I scrolled through this and the only formatting change I saw besides the power operator was for string prefixes, either `fr"..."` to `rf"..."` or vice versa.

This will probably cause a few merge conflicts with other pull requests (sigh) but this will make it so that we're using a production release of `black` rather than a beta release, and `black` should be stable now for all releases in a given year.  In any case, it's better to get changes like this implemented sooner rather than later...

.pre-commit-config.yaml
```diff
@@ -46,7 +46,7 @@ repos:
   hooks:
   - id: nbqa-black
     additional_dependencies:
-    - black==21.12b0
+    - black==22.3.0
     - click<8.1  # remove this when black is next updated
```
namurphy, line 50:
```suggestion
```
Remove click upper limit dependency

rocco8773 (approved):

PR #1585, opened 2022-06-15
[TEST] Remove `graphviz` from GitHub Actions & doc builds

namurphy:
Currently we install `graphviz` in GitHub Actions, and have `sphinx.ext.graphviz` as a Sphinx extension for our doc builds.  However, I haven't been able to figure out where we use `graphviz` in our package.  This PR is an experiment to remove `graphviz` and see what happens.  If it turns out that we do need it, I'm planning to reimagine this PR as an update to the doc guide to mention what we use `graphviz` for.

### DIFF
--- a/.github/workflows/fortnightly.yml
+++ b/.github/workflows/fortnightly.yml
@@ -69,7 +69,7 @@ jobs:
       run: python -m pip install --upgrade tox codecov
     - name: Install language-pack-fr and tzdata
       if: startsWith(matrix.name, 'Documentation')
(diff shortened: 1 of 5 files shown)


namurphy:
We got the warning message...
```
WARNING: dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
```
...so we do need `graphviz`, but I haven't figured out how it gets used.  For inheritance diagrams, maybe?

In any case, closing this since it's appropriate to have `graphviz` installed for doc builds.

PR #1569, opened 2022-05-23
Create a notebook on Coulomb collisions

namurphy:
This PR adds an example notebook that discusses Coulomb collisions, including `impact_parameter` and `Coulomb_logarithm`.  I'm also planning on including this content in the [Python in Heliophysics Community summer school](https://heliopython.org/summer-school) that's next week.

PR #1586, opened 2022-06-16
Update error messages in Coulomb logarithm calculations

namurphy:
This PR makes some updates to the error messages in `plasmapy.formulary.collisions`.  For example, instead of printing out the entire result from `Coulomb_logarithm` (which could be a long array), it prints out the minimum value.  

I'm spinning this off from #1569 since there's a bit more work to do here (like updating tests), but I might delay this until after Plasma Hack Week.  It'd be worth checking how the `docs/notebooks/formulary/coulomb.ipynb` notebook renders in the docs during code review, since that was my original motivation for editing these error messages.


StanczakDominik (approved):
These look like improvements! I'd be happy to merge them whenever you'd like.

changelog/1586.trivial.rst
```diff
@@ -0,0 +1 @@
+Updated the error messages in `~plasmapy.formulary.collisions.Coulomb_logarithm`.
```
namurphy, line 1:
```suggestion
Updated warning messages in `~plasmapy.formulary.collisions.coulomb.Coulomb_logarithm`.
```

StanczakDominik:
CI is failing because the warning tests check the actual text of these messages. Pretty simple to update them though.

PR #1257, opened 2021-09-07
Fix some broken and redirected links

namurphy:
This PR makes a few quick corrections to web links that were either broken or were redirects.  One of the changes is for an old changelog entry that had linked to the issue it closed rather than the pull request number.  I found these out doing `make linkcheck` in `docs/` which was pretty handy.  (I made a reST update here and there while I was the area too.)

PR #1587, opened 2022-06-17
Increase the strictness of the `tox` environment for building documentation

namurphy:
Prior to this pull request, our documentation builds would pass when there were typos like `plsmapy.formulary`.  Because of this, we ended up with a bunch of broken reST links that built up over time.  I ended up doing some large PRs to fix these (e.g., #1207, #1257, and #1509).  

This PR enables the `-n` option for `sphinx-build` in the `build_docs` environment for `tox`.  This will cause warnings to be emitted when there are broken reST links.  The `-W` flag (which was already present) made it so that the build would fail in case of any warnings.  Having the `-n` option enabled will prevent the situation where broken reST links get passed through silently.

This is a follow-up on #1206.


namurphy:
Oh, good.  Only 959 lines of warnings this time. 😐 

(Fortunately, though, most of them are the same warning but just repeated.)

namurphy:
I think this is the first time I've gotten a nitpicky doc build to work without errors!  🥲

docs/api_static/plasmapy.particles.particle_collections.rst
```diff
@@ -0,0 +1,8 @@
+:orphan:
+
+`plasmapy.particles.particle_collections`
+=========================================
+
+.. currentmodule:: plasmapy.particles.particle_collections
+
+.. automodapi::  plasmapy.particles.particle_collections
```
namurphy, line 8: This had only been missing for approximately 4.2 × 10<sup>7</sup> seconds!

CHANGELOG.rst
```diff
@@ -59,7 +59,7 @@ Deprecations and Removals
   `~plasmapy.particles.ionization_state.IonizationState.charge_numbers`. (`#1136 <https://github.com/plasmapy/plasmapy/pull/1136>`__)
```
rocco8773, line 59: For a future thought...  Do we want to keep this `CHANGELOG.rst`?  It feels like we're just duplicating the information.  To me it makes more sense to remove this file and rename `docs/whatsnew` -> `docs/changelog`, and only have the changelog in one location.  The only reason I see to keep `CHANGELOG.rst` is if `towncrier` forces us to keep it.

rocco8773 (commented):
I'm going to review this as one of the last PRs for the `0.8.0` release.  I think a lot of these changes would also be a part of #1597.

CHANGELOG.rst
namurphy: Good point!  I'm wondering about having the separate whatsnew pages for each of the releases, and then having a main changelog page that doesn't duplicate the information, but rather does a `.. include::` directive for each of the whatsnew pages.  I find it useful to have a full changelog page as well as the separate whatsnew pages (which we can link to in release announcements), though we have a couple of options that would work well.  In any case, we'll gain some experience with `towncrier` in the next few days.

rocco8773: I've used `.. include::` in the past, but I've been moving away from it because is [presents a potential security hole](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment).  I don't think it's a big deal in our usage, but still...

namurphy: Yep, it's best to avoid potential security holes even if it doesn't seem like it'll affect us.  I'll link to this comment in #1595 as a post-release thing to think about or create an issue on.

plasmapy/utils/tests/test_units_helpers.py
```diff
@@ -6,7 +6,7 @@
 from astropy.constants import c, m_e
```
rocco8773, line 6: The module docstring still uses `plasmapy.utils.units_helpers`, instead of `plasmapy.utils._units_helpers`.

changelog/1531.trivial.rst
```diff
@@ -1,2 +1,3 @@
-Reduced execution time by avoiding calculations with units while keeping unit validation.
-Added `plasmapy/utils/units_definitions.py` to precompute units.
+Reduced execution time by avoiding calculations with units while keeping
+unit validation. Added :file:`plasmapy/utils/units_definitions.py` to
+precompute units.
```
rocco8773, line 3: This entry does not make sense on it's own.  There is no context.  Reduced execution timing of what?  What is `units_definitions.py`?  Also, this seems like two separate changelog entries.

changelog/1587.doc.rst
```diff
@@ -0,0 +1,4 @@
+Increased the strictness of the ``build_docs`` tox_ environment so that
+broken reST_ links now emit warnings which are then treated as errors,
+removed the ``build_docs_nitpicky`` tox_ environment, and updated the
+|documentation guide| accordingly.
```
rocco8773, line 4: Possibly add something about fixing all links that were failing under `nitpicky`.  This could be added to this entry or as its own.

rocco8773 (changes requested):
This looks nearly there.  Just need to address those few comments I had and then we're all good.

plasmapy/utils/tests/test_units_helpers.py
namurphy: Done.

rocco8773 (approved):

PR #1598, opened 2022-06-28
Update DOI in citation instructions for 0.8.0 release

namurphy:

      docstrings.


docs/about/citation.rst
```diff
@@ -9,15 +9,19 @@ specific version of PlasmaPy used in your project.  Citing a software
...
+.. |version_to_cite| replace:: 0.8.0
+.. |doi_for_version| replace:: 10.5281/zenodo.6774350
 
-   PlasmaPy Community et al. (2021). *PlasmaPy*, version 0.6.0, Zenodo,
-   https://doi.org/10.5281/zenodo.4602818
+Version |version_to_cite| of PlasmaPy may be cited with the following
+reference:
+
+   PlasmaPy Community et al. (2021). *PlasmaPy*, version |version_to_cite|,
+   Zenodo, https://doi.org/\ |version_to_cite|
```
namurphy, line 19: - [ ] Double check that this works

rocco8773 (changes requested):
Currently this does not work as intended.  Specifically the doi link mashing is not working.  I'm going to let you play with this, so ping me when it's ready for review.

Also, the replacements should either be moved to the top of file or put into `common_links.rst`.  The release guide would also need to be updated to reflect that these items would need updating during a release.

docs/about/citation.rst
```diff
@@ -3,36 +3,43 @@
 Acknowledging and Citing
 ========================
 
-If you use PlasmaPy for a project that results in a research publication,
-we ask that you cite the `Zenodo <https://zenodo.org>`__ record for the
-specific version of PlasmaPy used in your project.  Citing a software
-package promotes scientific reproducibility, gives credit to its developers,
-and highlights the importance of software as a vital research product.
+.. |version_to_cite| replace:: 0.8.0
+.. |doi_hyperlink| replace:: https://doi.org/10.5281/zenodo.6774350
```
rocco8773, line 7: When I go to this link I get a "DOI Not Found"?  Has this doi only been reserved and not yet "created"?

namurphy: Correct: the link to the DOI does not exist yet, so this link won't work until after we've created the Zenodo record.  This does make it annoying to test, though...

rocco8773: Ok, just wanted to double check before approving.

rocco8773 (approved):

PR #1604, opened 2022-06-29
Modify impact_parameter() to work with N-D arrays

pheuer:
The `impact_parameter` function in `collisions.py` can currently handle a 1D array of densities or temperatures, but not arrays of higher dimension. This PR changes line that handles 1D arrays to work for any shape array. 

`impact_parameter` is called by many of the other functions in the collisions module (for example, `collision_frequency'), so this change will also enable support for higher dimension arrays in some of those functions.

### DIFF
--- a/changelog/1604.feature.rst
+++ b/changelog/1604.feature.rst
@@ -1,1 +1,1 @@
-Added support for arbitrarily shaped input arrays to the function `formulary.collisions.impact_parameter`.
+Added support for arbitrarily shaped input arrays to the function `plasmapy.formulary.collisions.impact_parameter`.
--- a/changelog/1604.feature.rst
+++ b/changelog/1604.feature.rst
@@ -1,0 +1,1 @@
+Added support for arbitrarily shaped input arrays to the function `formulary.collisions.impact_parameter`.
--- a/plasmapy/formulary/tests/test_collisions.py
+++ b/plasmapy/formulary/tests/test_collisions.py
@@ -1016,6 +1016,36 @@ class Test_impact_parameter:
         self.V = 1e4 * u.km / u.s
         self.True1 = np.array([7.200146594293746e-10, 2.3507660003984624e-08])
 
+    @pytest.mark.parametrize(
+        "n_e_shape,T_shape",
+        # Scalar T
+        [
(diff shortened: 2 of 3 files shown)


changelog/1604.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for arbitrarily shaped input arrays to the function `formulary.collisions.impact_parameter`.
```
namurphy, line 1:
```suggestion
Added support for arbitrarily shaped input arrays to
`~plasmapy.formulary.collisions.impact_parameter`.
```

namurphy (commented):
Thank you, @pheuer!  This looks really helpful.

Would you be able to add some tests of this capability?  If this function gets revised in the future, I can imagine that this change could get accidentally/unknowingly reverted.

Thank you again!

namurphy (commented):
It's looking close!  I have a couple of minor suggestions, but we can probably merge this today.

pheuer:
Those suggestions look good to me - thanks @namurphy !

namurphy:
Cool!  Assuming everything passes, we can merge this just in time for the 0.8.0 release which may happen Friday we hope.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1094,13 +1094,40 @@ def test_handle_nparrays(self, insert_some_nans, insert_all_nans, kwargs):
...
+        bmin and bmax have the correct shapes.
+
+        This is necessary in addition to test_handle_nparrays to ensure that
+        the output arrays are extended correctly.
+
         """
-        (bmin, bmax) = impact_parameter(1 * u.eV, self.n_e_arr, self.particles)
-        assert len(bmin) == len(bmax)
+
+        output_shape = max(map(len, [T_shape, n_e_shape]))
```
namurphy, line 1120:
```suggestion
        output_shape = T_shape if len(T_shape) >= len(n_e_shape) else n_e_shape
```
Oops!  Let's try the other option...

namurphy (approved):
I'll enable auto-merge since as long as tests pass then this will be ready.  Thanks again for doing this!

PR #1605, opened 2022-06-29
Change directory of contributor guide to better match its name (with redirects)

namurphy:
The directory name `docs/development` came from the name "Development Guide", which was later changed to "Contributor Guide".  This PR renames the directory to `docs/contributing` to better match the new name.

I also used the [sphinx-reredirect](https://documatt.gitlab.io/sphinx-reredirects/) extension to set up redirects.  This looks like a really helpful extension since we'll probably end up changing the names of pages again in the future too.


rocco8773 (changes requested):
Should we just delete `docs/contributing/workflow.rst`?  If I remember correctly, it was only created as part of a demo for the 2021 hack week.  I don't see why we should retain it.  Especially since it's not referenced anywhere in the rest of the documentation.

namurphy:
> Should we just delete `docs/contributing/workflow.rst`? If I remember correctly, it was only created as part of a demo for the 2021 hack week. I don't see why we should retain it. Especially since it's not referenced anywhere in the rest of the documentation.

Agreed.  I keep meaning to work on this but too many other things keep coming up...sigh!

rocco8773 (approved):

Issue #1602, opened 2022-06-29
Generate author list from `codemeta.json`


namurphy:
Right now we have authorship recorded in two locations:

 - `docs/about/credits.rst`
 - `codemeta.json`

We might also potentially create `.zenodo.json` again (#1340) for a third place.  The [CodeMeta](https://codemeta.github.io/) file has become more out-of-date than the author list in the docs.  

It's a lot of work to maintain multiple author lists, which is often work done around the time of releases.  Automating this (in conjunction with #1601) would help streamline the release process and reduce the amount of work that needs to be done manually.

A possibility would be to automagically generate the author list in `docs/credits/about.rst` from the information contained within `codemeta.json`.  I think there might be crosswalk tools out there between `codemeta.json` and `.zenodo.json` files too, and eventually (hopefully) Zenodo will start loading in data from CodeMeta files too.  If we do this, then `codemeta.json` would become the primary location for where authors are defined (which would mean changing #1601).

PR #1606, opened 2022-06-29
Update `codemeta.json`

namurphy:
This PR updates our [CodeMeta](https://codemeta.github.io/) file to be consistent with our requirements and the updated authors list in #1599.


namurphy:
I'm wondering if it would be easiest to do this by addressing #1602 directly, since the information is already in `credits.rst`.  Hm...

docs/about/credits.rst
```diff
@@ -41,7 +41,7 @@ in parentheses are `ORCID author identifiers <https://orcid.org>`__.
 * :user:`BH4`
 * :user:`Brian Goodall <goodab>`
 * :user:`Brigitta Sipőcz <bsipocz>` (:orcid:`0000-0002-3713-6337`)
-* :user:`Bryan Foo <bryancfoo>` (:orcid:`0000-0001-5308-6870`)
+* :user:`Bryan Foo <bryancfoo>`
```
namurphy, line 44: I added this ORCID in #1599, but I don't know if this is the correct ORCID.

namurphy:
After getting this pretty much done, I looked up Zenodo's FAQ and found that they haven't yet implemented their integration with CodeMeta.  Zenodo does have an integration with Citation File Format (CFF), but I haven't found a way to convert CodeMeta to CFF that worked.  But, if I upload the metadata to Zenodo manually, then there are tools that can take the Zenodo metadata and make them a CFF file, and once we have a CFF file, then we can convert that back to CodeMeta with existing tools.  At the very least, this will make librarians happy.  📚

change_authors_to_dict.ipynb
```diff
@@ -0,0 +1,843 @@
+{
```
rocco8773: Are you wanting this to be in the root directory?  I can see these being useful tools for the future, but having the functionality in the notebook in the root directory does not seem right.  I would actually propose making `plasmapy_helpers` package in the root directory that is ignored by `find_packages()` so it's not distributed with `plamsapy`.  We can stuff `plasmapy_helpers` with all the useful pkg management tools we can imagine.

For now, I suggest throwing this notebook into a root directory called `notebooks_for_proto` since that directory is already listed as ignored by `.gitignore`.

namurphy: Oops!  I meant to commit it temporarily so that it was in version control history for future me, but I had thought that I had deleted it.  

But with that said, I do very much like the idea of having something akin to a `plasmapy_helpers` directory that could include things like this.  In the past I had wondered about a separate repo for things like this.

namurphy: Correction: it looks like I already did delete that notebook.

codemeta.json
```diff
@@ -454,80 +811,147 @@
...
+            "familyName": "Xu"
+        },
+        {
+            "@type": "Person",
+            "@id": "https://github.com/syip1",
+            "givenName": "Chun Hei",
+            "familyName": "Yip"
+        },
+
+        {
```
namurphy, line 950:
```suggestion
        },
        {
```

codemeta.json
```diff
@@ -454,80 +811,147 @@
...
+            "givenName": "Sixue",
+            "familyName": "Xu"
+        },
+        {
+            "@type": "Person",
+            "@id": "https://github.com/syip1",
+            "givenName": "Chun Hei",
+            "familyName": "Yip"
+        },
+
```
namurphy, line 949:
```suggestion
```

changelog/1606.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Updated the :file:`codemeta.json` file with metadata for the
+version 0.8.0 release.
```
rocco8773, line 2:
```suggestion
version ``0.8.0`` release.
```

optional

rocco8773 (approved):

namurphy:
@rocco8773 — I committed your suggestion, which reset the review. Could you approve it again?  Thanks!

rocco8773 (approved):

PR #1607, opened 2022-06-29
Add `robots.txt` to tell web crawlers for search engines to ignore all but `stable` & `latest` doc builds

namurphy:
In a discussion about *Stars in My Pocket Like Grains of Sand* by Samuel R. Delaney, it came up that in the mid-1990s there emerged a [standard for robot exclusion](http://www.robotstxt.org/orig.html) via a `robots.txt` file.  This file is used to tell web crawlers (spiders!) to ignore or exclude certain files from being indexed.  I occasionally get search results for version 0.1.0, and was hoping to try this out for us.  This PR copies the `robots.txt` file from Astropy, and modifies `docs/conf.py` so that it appears in the online documentation.

### DIFF
--- a/docs/robots.txt
+++ b/docs/robots.txt
@@ -1,0 +1,6 @@
+User-agent: *
+Allow: /*/latest/
+Allow: /en/latest/   # Fallback for bots that don't understand wildcards
+Allow: /*/stable/
+Allow: /en/stable/   # Fallback for bots that don't understand wildcards
+Disallow: /
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -209,6 +209,8 @@ exclude_patterns = [
     "**Untitled*",
 ]
 
+html_extra_path = ["robots.txt"]
+
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 


docs/robots.txt
```diff
@@ -0,0 +1,6 @@
+User-agent: *
+Allow: /*/latest/
+Allow: /en/latest/   # Fallback for bots that don't understand wildcards
+Allow: /*/stable/
+Allow: /en/stable/   # Fallback for bots that don't understand wildcards
+Disallow: /
```
namurphy, line 6: Verbatim from [Astropy's `robots.txt` file](https://github.com/astropy/astropy/blob/2f4b3d2e51e22d2b4309b9cd74aa723a49cfff99/docs/robots.txt).

rocco8773 (commented):
Any idea what happens if someone searches for documentation for a specific PlasmaPy version?

---

On a parallel note, I was think we should put an admonition at the top of the documentation landing page that might say something like...

"You are on the `latest` version of the documentation which corresponds to the ongoing development on the `plamsapy` `main` branch.  If you're looking for the documentation related to the latest release of `plasmapy`, then please go to the `stable` version of the documented.  You can also find version specific documentation in the version selector in the bottom left corner of you screen."

Thoughts?

namurphy:
Just tried a search for "astropy 4.3.1 documentation", and the Astropy 4.3.1 doc page is not included in those results.  😿 

On the other hand, if I do a search for "plasmapy alfven_speed", then the first link that pops up is for version 0.1.1, which is not good.  👀 

Hm...

namurphy:
> On a parallel note, I was think we should put an admonition at the top of the documentation landing page that might say something like...
> 
> "You are on the `latest` version of the documentation which corresponds to the ongoing development on the `plamsapy` `main` branch. If you're looking for the documentation related to the latest release of `plasmapy`, then please go to the `stable` version of the documented. You can also find version specific documentation in the version selector in the bottom left corner of you screen."
> 
> Thoughts?

This sounds like a great idea to me!  Probably for 0.9.0 or 0.10.0, unless you have the capacity to do this by tomorrow.

rocco8773:
Definitely!

Do you want this PR a part of `v0.8.0`? ...Anyways, I'll approve it and you can decide what you want to do with it.

rocco8773 (approved):

PR #1613, opened 2022-07-01
Use `pathlib` to get path to `plasma-calculator` notebook

### LINKED ISSUE #1612: `plasma-calculator` does not launch on a Windows 10 system
Original comment...https://github.com/PlasmaPy/PlasmaPy/pull/1610#discussion_r912220976

The `plasma-calculator` does not launch when using a Windows 10 system...

> Also, I get an error when trying to run the calculator on my windows machine...
> 
> image
>
> My suspicion is whatever functionality is generating the file path is ignorant of a windows file separator, since the path and file name are mashed into one without a separator.

namurphy:
This PR uses `pathlib` instead of `os` to create the path to the plasma calculator notebook.  Closes #1612.  I hope.

### DIFF
--- a/plasmapy/utils/calculator/__init__.py
+++ b/plasmapy/utils/calculator/__init__.py
@@ -43,12 +43,13 @@ def main():
 
     command = [
         "voila",
-        no_browser,
-        f"--port={args.port}" f"--{theme=}",
         notebook_path,
-        "--VoilaConfiguration.file_whitelist",
-        favicon_path,
+        f"--port={args.port}",
+        f"--theme={theme}",
+        f"--VoilaConfiguration.file_whitelist={favicon_path}",
     ]
+    if no_browser:
+        command.append(no_browser)
 
     try:
         subprocess.call(command)
--- a/plasmapy/utils/calculator/__init__.py
+++ b/plasmapy/utils/calculator/__init__.py
@@ -33,15 +33,24 @@ def main():
         "--no-browser", action="store_true", help="Do not open the browser"
     )
 
-    notebook_path = pathlib.Path(__file__).parent.absolute() / "plasma_calculator.ipynb"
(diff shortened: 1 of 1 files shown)


rocco8773:
See my comment...https://github.com/PlasmaPy/PlasmaPy/issues/1612#issuecomment-1172767853

I think the easiest solution here is to just define `command` as a list of strings, and not try to split the string after creation.

namurphy:
> I think the easiest solution here is to just define `command` as a list of strings, and not try to split the string after creation.

Interesting!  I attempted to do that.

plasmapy/utils/calculator/__init__.py
```diff
@@ -33,17 +33,24 @@ def main():
...
-    command = f"voila {no_browser} --port={args.port} --theme={theme} {computed_calculator_path} \
-        --VoilaConfiguration.file_whitelist favicon.ico"
+    command = [
+        "voila",
+        no_browser,
+        f"--port={args.port}" f"--{theme=}",
+        notebook_path,
+        "--VoilaConfiguration.file_whitelist",
+        favicon_path,
+    ]
```
rocco8773, line 51:
```suggestion
    command = [
        "voila",
        notebook_path,
        f"--port={args.port}",
        f"--theme={theme}",
        f"--VoilaConfiguration.file_whitelist={favicon_path}",
    ]
    if no_browser:
        command.append(no_browser)

```

This is what got it to work on my machine.  Adding `no_browser` to the list when it's an empty string just makes voila think you're specifying two notebooks, and it gets confused.

namurphy: This was eerily similar to what I ended up getting to work on my machine, except yours is a little bit cleaner.

rocco8773 (approved):

PR #1615, opened 2022-07-02
Update release guide after 0.8.1 release

namurphy:
I'm planning to update this while I'm working on the release to fill in any bits of it that I get confused about.  This should not be merged until after the release is finished.


namurphy:
- [ ] Mention the GitHub search status for counting the number of pull requests merged and issues created.

      ```
      is:pr merged:>=2021-11-19
      is:issue closed:>=2021-11-19
      ```

docs/contributing/release_guide.rst
```diff
@@ -5,170 +5,349 @@ Release Guide
...
+  link in :file:`docs/about/citation.rst` will not work yet since the
+  Zenodo_ record will not be published until after the official release.
+
+  .. tip::
+
+     Use ``linkcheck_allowed_redirects`` in :file:`docs/conf.py` to
+     specify allowed redirects. For example, DOI links are more
+     persistent than most hyperlinks, but are always redirects.
+
+* Open a pull request to update the changelog.
```
namurphy, line 96: @rocco8773's suggestion is to perform the towncrier magic on the release branch and then to merge it back in to main.

For the release, we have to delete `dev` from toctree for the changelog index page, but the `dev` should be added back in the PR back to `main`.

(though can we do an `if` thing for this so we don't have to do this step?)

namurphy: See https://github.com/BaPSF/bapsflib/pull/92 and https://github.com/BaPSF/bapsflib/issue/80

StanczakDominik:
@rocco8773 had the idea of grouping stuff in the release guide as in https://github.com/BaPSF/bapsflib/issues/80 into 4-5 categories (possibly adding a metadata step after the first part). We could add a few headers in this PR, probably? :)

docs/contributing/release_guide.rst
```diff
@@ -26,42 +19,27 @@ Throughout this guide, ``0.9.0`` denotes the version you're releasing.
...
+* Three weeks before a minor or major release, announce that a feature
+  freeze will occur one week before the anticipated release date. Only
+  pull requests with a limited scope that do not significantly change
+  functionality should be merged during the feature freeze.
 
-Pre-release
------------
+* Three weekdays before a minor or major release, announce a code
+  freeze. Only bugfixes and pull requests that are directly related to
+  the release should be merged during the code freeze.
```
StanczakDominik, line 38: Not to hold up the PR, but this still feels like something we could avoid with sufficiently advanced ~technology~ ~magic~ Git usage.

docs/contributing/release_guide.rst
```diff
@@ -70,17 +48,57 @@ Pre-release
...
+* Open a pull request to ``main`` to update black_ and other pre-commit_
+  hooks to their most recent versions in :file:`.pre-commit-config.yaml`
+  for major and minor releases. Apply all changes by running:
+
+  .. code-block:: bash
+
+     pre-commit run --all-files
+
+* Open a pull request to re-execute pre-executed notebooks, such as
+  those for charged particle radiography.
```
StanczakDominik, line 90: Sure feels like doing all of these via PRs would extend the duration of every release. If we wanna release more often, we need to be going in the opposite direction. A single PR should be good enough, no?

docs/contributing/release_guide.rst
```diff
@@ -204,8 +234,8 @@ Release
   to make sure that the new version installs correctly.
 
   - Open Python and run ``import plasmapy`` and ``dir(plasmapy)``.
-  - Run ``plasma-calculator`` to make sure that the plasma calculator is
-    behaving correctly.
+  - Run ``plasma-calculator`` from the terminal to make sure that the
+    plasma calculator is behaving correctly.
```
StanczakDominik, line 226: In other words, that part isn't tested... welp. :(

docs/contributing/release_guide.rst
namurphy: I'm all for automating as much of the release process that we reasonably can!

docs/contributing/release_guide.rst
namurphy: My motivation for separate PRs is that the dominant timescale is often code reviews, and code review time often scales more than linearly with respect to PR size.  If someone else is available for rapid code reviews, I'd expect that multiple focused PRs would be faster so long as they're not interdependent.

namurphy:
- [ ] Add that `CITATION.cff` needs to be updated along with `codemeta.json` (or we might want to delete `codemeta.json` at least temporarily since CFF has better compatibility with GitHub & Zenodo at this point)

StanczakDominik:
Can we merge this PR asap and break the remaining todos into separate issues?

namurphy:

I was thinking that for subsequent updates to the release guide, it would be worth looking at the version of the release guide from just before this PR.  I deleted a few things that I wasn't able to get to work on my end (like the `xonsh` script), and there's some content that we might want to bring back.

I'll also usually be happy to take on the responsibility of doing things like checking hyperlinks and updating metadata & author lists, since I actually find it kind of fun and relaxing.

PR #1616, opened 2022-07-02
Merge `v0.8.x` branch back into `main`

namurphy:
This PR contains the changes to `v0.8.x` that were made during the release that need to be merged back into `main`.


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Upload Python Package
```
rocco8773, line 9: Note, by breaking this out into its own workflow no tests will be run before the publish happens.  I don't know if this is a good or bad thing.  Just stating it is a thing.

namurphy: Correct.  The tests won't run before the release happens, but the tests will be run when pushing to a `v*.*.*` branch and for pull requests.  I think there is still...some room for improvement.

PR #1633, opened 2022-07-07
Remove `integer_charge` attributes

namurphy:
In #1136, we renamed `integer_charge` to `charge_number` and deprecated `integer_charge`.  This PR removes `integer_charge` for the 0.9.0 release.


plasmapy/particles/atomic.py
```diff
@@ -6,7 +6,6 @@
     "common_isotopes",
     "electric_charge",
     "half_life",
-    "integer_charge",
```
namurphy, line 9: I'm wondering if it'd be worth it to have a `charge_number` function here too.  Maybe a follow-up PR...

namurphy: Oh wait, we already have one.  Nevermind.

changelog/1633.removal.1.rst
```diff
@@ -0,0 +1,3 @@
+The ``integer_charge`` attributes of |Particle| and |IonicLevel| have
+been removed after being deprecated in ``v0.7.0``. Use the
+``charge_number`` attribute instead.
```
StanczakDominik, line 3: I think I see your logic behind the split of these changelogs into three different notes, but just for future reference, if anyone's removing related, deprecated functionality, would they be expected to also do this, well, dry and formal split, or is a single note fine?

StanczakDominik (approved):
I'd say that this is a solid addition to PlasmaPy if it wasn't mostly a subtraction. A solid one, though!

changelog/1633.removal.1.rst
namurphy: Good question!  I'd say it would be fine to do it either way, so long as the changelog entry doesn't end up too bulky.  My usual strategy is to have each changelog describe one change, though what "one change" means can be a bit nebulous.  For this PR, I split them up because it was easier to write that way since there were a couple of small differences between the different changes.

namurphy:
Awesome, thank you for the review!

changelog/1633.removal.1.rst
StanczakDominik: Makes sense! 👍🏻

PR #1634, opened 2022-07-07
Fix assertions in tests that were always `True`

namurphy:
There were a few tests in `test_dielectric.py` which were of the form
```Python
assert (
    np.isclose(...), "message"
)
```
This was doing an `assert` on a non-empty `tuple`, which consequently always evaluated to `True`.  This PR adjusts the parentheses on the tests.

### DIFF
--- a/plasmapy/formulary/tests/test_dielectric.py
+++ b/plasmapy/formulary/tests/test_dielectric.py
@@ -172,8 +172,7 @@ class Test_permittivity_1D_Maxwellian:
         kwargs["kWave"] = kwargs["omega"] / vth
 
         val = permittivity_1D_Maxwellian(**kwargs)
-        assert (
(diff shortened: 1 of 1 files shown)


StanczakDominik (approved):
Oh darn. Gotta admit that assert syntax was always a little magical to me.

So without the parenthesis, it passes in an unzipped tuple (like `*args`), and with them, just a tuple that's always true since it's non empty?

Good catch, all in all!

namurphy:
> So without the parenthesis, it passes in an unzipped tuple (like `*args`), and with them, just a tuple that's always true since it's non empty?

That's my understanding of it!  I think it was an extension to PyCharm (or maybe `flake8`?) that caught this...otherwise I don't think I'd have noticed it!

Issue #1635, opened 2022-07-07
Fix problems with opening binder links


namurphy:
I tried opening some example notebooks in binder (one of the steps in #1595) but got a problem where it says...
```
ERROR: Could not find a version that satisfies the requirement plasmapy==0.8 (from versions: 0.1.dev1, 0.1.dev2, 0.1.dev835, 0.1.dev836, 0.1.dev841, 0.1.dev968, 0.1.0, 0.1.1, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 0.5.0, 0.6.0, 0.7.0)
ERROR: No matching distribution found for plasmapy==0.8
```
This was what I mistakenly put in a slightly older binder requirements file, which I have since updated to approximately `plasmapy >= 0.8, < 0.9` on `v0.8.x`.  We should fix this so that our binder environments work for us to open our notebooks in.  

To test that the binder links are working, we can try opening it from the following pages:

 - [ ] [Particles notebook on `latest`](https://docs.plasmapy.org/en/latest/notebooks/getting_started/particles.html)
 - [ ] [Particles notebook on `stable`](https://docs.plasmapy.org/en/stable/notebooks/getting_started/particles.html)
 - [ ] [Particles notebook on `v0.8.x`](https://docs.plasmapy.org/en/v0.8.x/notebooks/getting_started/particles.html)

Once we fix this, we should also...

 - [ ] Update the release guide to reduce chances of this issue coming up again

namurphy:
Closed by #2672!

PR #1636, opened 2022-07-07
Install PlasmaPy from repo when launching notebooks with binder for `latest` docs 

namurphy:
During the 0.8.1 release, I mistakenly included a change to `binder/requirements.txt` in #1621 that should not have gone back into `main`.  This PR will revert that change for `main`, so that binder will install PlasmaPy from the `main` branch of the repository and not `pip`.

This will fix part of #1635, but not all.

 - [ ] Test that binder links from the doc build for this PR are able successfully spin up an environment

### DIFF
--- a/binder/requirements.txt
+++ b/binder/requirements.txt
@@ -1,5 +1,5 @@
 -r ../requirements/build.txt
 -r ../requirements/install.txt
 -r ../requirements/extras.txt
+.
 jupytext
-plasmapy == 0.8


namurphy:
This appears to be one of those infrequent cases where the PR has to be merged before we know whether or not it will work.  When I tried opening binder from the particles notebook from this PR's doc build, it looked like it was unable to recognize the hash of the commit...which makes sense since the commit is not yet on `main` for the PlasmaPy repo, but rather on my fork.  Despite that, this is reverting it to what it was before `0.8.1`, so it should be fine.

binder/requirements.txt
```diff
@@ -1,5 +1,5 @@
 -r ../requirements/build.txt
 -r ../requirements/install.txt
 -r ../requirements/extras.txt
+.
```
rocco8773, line 4:
```suggestion

```

What is this?  It doesn't look right to me.

namurphy: That's a good question!  The period is to say that binder should install the version of PlasmaPy in the repository, and not the official release.  This means we'll open the notebooks with the version of PlasmaPy in the `main` branch, and not the latest release.

StanczakDominik (approved):
Makes sense to me. We can always iterate on it afterwards.

PR #1644, opened 2022-07-25
Magnetics branch

aminaahmed1803:
Description: I have added a new module named **magnetics.py** in plasmapy/diagnostics. 

The module implements a class **Magnetics** which has three functions: 

1. bdot_field: computes the magnetic field for bots
2. mean_sub: preforms mean subtraction on a data array
3. band_pass_filter: applies a band pass filter on a data array
 
Each function has docstrings that contain its description.

There is an additional scratch file called **testfile.py** which I used for running my functions and testing them.
It doesn't contain anything significant and can be removed.


namurphy:
@aminaahmed1803 — thank you for making this contribution! We really appreciate it!

It looks like the issue with the **CI / Python 3.10 (pull_request)** check is related to `testfile.py`, which you said will be removed at some point anyway.  If you delete that file then the tests should run.  We could work with you to put the tests into a new file, `plasmapy/formulary/diagnostics/tests/test_magnetics.py`.  

The **pre-commit.ci - pr** check and the **CI / Linters (pull_request)** checks are both related to formatting, and it's completely fine to ignore both of those at this point.  If you'd like, you can comment on this pull request with `pre-commit.ci autofix` to run the automatic formatting tools, and then pull the changes back to your own computer (e.g., with `git pull`).

I'm not sure what's happening with the **CI / Documentation** check yet.  @rocco8773 — was this related to the Sphinx 5.1 thing that came up while I was away?

Thank you again!

PR #1647, opened 2022-07-26
Make docs build off sphinxdev actually build docs

### LINKED ISSUE #1646: Fix the setup for the fortnightly test doc build with the development version of Sphinx
We currently have a fortnightly test for [build_docs-sphinxdev](https://github.com/PlasmaPy/PlasmaPy/blob/b0654f6815855c46bbc3a354d03b0221228620b7/.github/workflows/fortnightly.yml#L33).  However, instead of doing a test documentation build, it is running the test suite (as in [this run](https://github.com/PlasmaPy/PlasmaPy/runs/7353748770?check_suite_focus=true)).  We should fix it so that it's actually running a test documentation build.

StanczakDominik:
07ee524db54416a8afc2c44123562238f79641d7 fixes #1646, but I'd still like to play around with this a little to see whether the three `build_docs` tox envs we have now couldn't be compressed into one.

The issue here is that `sphinxdev` is defined in `testenv` rather than in `testenv:build_docs`, and I think there's logic in `tox` that pulls the `commands` field from `all: {env:PYTEST_COMMAND} {posargs}` defined in  `testenv`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,6 +49,17 @@ setenv =
     HOME = {envtmpdir}
 commands = sphinx-build docs docs{/}_build{/}html -W -n --keep-going -b html {posargs}
 
+[testenv:build_docs-sphinxdev]
+changedir = {toxinidir}
+extras = docs
+setenv =
+    HOME = {envtmpdir}
+commands = sphinx-build docs docs{/}_build{/}html -W -n --keep-going -b html {posargs}
+deps =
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Gah, I've started a `nox` based rewrite and wasn't exactly happy with that either. This is an improvement, let's get it in.

namurphy (approved):
LGTM!  And thank you for doing this!

Issue #1648, opened 2022-07-28
[EPIC] Introducing Magnetic Flux Probes


rocco8773:
1. Functionality for `plasmapy.signal`
2. Functionality for `plasmapy.analysis`

aminaahmed1803:
I completed part one and opened a pull request _(signal.py and bdot_probe.py #1652)_ for it

StanczakDominik:
For the uninitiated, what does `new_time` represent??

aminaahmed1803:
After integration of the bdot array, the resultant bfield array is one index shorter because of the integration. new_time is the array containing time values corresponding to bfield.

svincena:
Definition : cumulative_trapezoid(y, x=None, dx=1.0, axis=-1, initial=None)

Typically, I like to specify the initial value of the trapezoidal integration scheme (usually zero), in which case the time arrays are the same and some overhead of passing in and out entire time arrays that are the same and as large (or off by one) as the data for many subsequent integrations is reduced.

aminaahmed1803:
I created diagnostics/magnetics.py as formatted above. I haven't documented it yet. Just wanted to make sure it was correct first!

PR #1652, opened 2022-07-29
signal.py and bdot_probe.py

aminaahmed1803:
- Added functionality for plasmapy.analysis
      - Created bdot_probe.py module:
                1. calc_bfield() function. 
      - Created sginal.py module:
                2.  remove_offset() function.
                3. band_pass_filter() function. 

All functions do not use units.
All functions strictly use 1D numpy arrays.

### DIFF
--- a/plasmapy/analysis/bdot_probe.py
+++ b/plasmapy/analysis/bdot_probe.py
@@ -1,0 +1,29 @@
+"""_summary_
+   
+"""
+
+import numpy as np
+import scipy.io as spio
+import scipy.integrate as sp
+from warnings import warn
+
+
+def compute_bfield(bdot_data: np.ndarray, times: np.ndarray, tloop_area: float) -> Tuple(np.ndarray, np.ndarray):
+   """_summary_ : Function to compute magnetic field for bdots
+        Args:
+            Bdot (array): - volts  
+            tloop_area (float): - meters squared 
+            times_s (array): - seconds   
+        Returns:
+            field_arr : array of magnetic field values for the bdots
+            new_time: corrsponding time array  
+
+   """
+   if len(bdot_data) != len(times):
+     raise Exception("length of time and voltage arrays in not equal\n")
+
+   bdot_data[:] = [x / tloop_area for x in bdot_data]
+   field_arr = sp.cumtrapz(bdot_data, times)  # Tesla
+   new_time = times[1:]
+
+   return field_arr, new_time
--- a/plasmapy/analysis/signal.py
+++ b/plasmapy/analysis/signal.py
@@ -1,0 +1,45 @@
+"""_summary_
+
+"""
+
+import numpy as np
+import scipy.io as spio
+import scipy.integrate as sp
+from scipy import signal
+from warnings import warn
+
+
+def remove_offset(data_array: np.ndarray, start_idx: int, end_idx: int) ->  np.ndarray:
+   """_summary_ : Function which preforms mean subtraction
+   
+   Args:     
+      data_array (array): array containing values from which to mean sub
+      start_idx (int): idx of where to start obtaining mean from
+      end_idx (int): idx of where to stop obtaining mean from
+   Returns:
+      data_array (array): array after mean_sub is preformed 
+   """
+   data_array = np.array(data_array)
+   meansub = np.mean(data_array[start_idx:end_idx])
+   data_array[:] = [x - meansub for x in data_array]
+   return data_array
+
+def band_pass_filter(data_array: np.ndarray, cutoff: float, sampling_frequency: float, order: int) ->  np.ndarray:
+   """_summary_ : Function which applies a band_pass_filter
+   
+   Args:
+      data_array (array): _description_
+      cuttoff (_type_): _description_
+      sampling_frequency (_type_): _description_
+      order (_type_): _description_
+   Returns:
+      filtered_arr : array with filter applied 
+   """
+   if cutoff > sampling_frequency:
+      raise ValueError("sampling frequency must be greater than cutoff value\n")
+   
+   nyq = 0.5 * sampling_frequency
+   normal_cutoff = cutoff/nyq
+   b, a = signal.butter(order, normal_cutoff, btype='highpass', analog=False)
+   filt_arr = signal.filtfilt(b, a, data_array)
+   return filt_arr


plasmapy/analysis/signal.py
```diff
@@ -0,0 +1,45 @@
...
+      end_idx (int): idx of where to stop obtaining mean from
+   Returns:
+      data_array (array): array after mean_sub is preformed 
+   """
+   data_array = np.array(data_array)
+   meansub = np.mean(data_array[start_idx:end_idx])
+   data_array[:] = [x - meansub for x in data_array]
+   return data_array
+
+def band_pass_filter(data_array: np.ndarray, cutoff: float, sampling_frequency: float, order: int) ->  np.ndarray:
```
svincena, line 27: This function implements a high-pass filter, not a band-pass filter. Consider renaming this function "high_pass_filter" or implement the high and low cutoffs

plasmapy/analysis/bdot_probe.py
```diff
@@ -0,0 +1,31 @@
+"""_summary_
```
rocco8773, line 1: Remove this file from this pull request (PR).  We'll include this in its own PR.

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,46 @@
+"""_summary_
```
rocco8773, line 1: Rename this file `signal.py`.

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,46 @@
...
+def remove_offset(data_array: np.ndarray, start_idx: int, end_idx: int) -> np.ndarray:
+    """_summary_ : Function which preforms mean subtraction
+
+    Args:     
+       data_array (array): array containing values from which to mean sub
+       start_idx (int): idx of where to start obtaining mean from
+       end_idx (int): idx of where to stop obtaining mean from
+    Returns:
+       data_array (array): array after mean_sub is preformed 
+    """
```
rocco8773, line 20: We use the [`numpydoc` style](https://numpydoc.readthedocs.io/en/latest/format.html) for our documentation.  Can you please look into converting your docstrings?

You can also refer to our [documentation guide](https://docs.plasmapy.org/en/stable/contributing/doc_guide.html).

rocco8773 (changes requested):
Thank you for working on this.  I've left a few comments for changes.  

How comfortable are you with writing tests?  We'll need to cover the functionality being introduced here.  @namurphy Has put together a rather robust [testing guide](https://docs.plasmapy.org/en/stable/contributing/testing_guide.html) that you can use for reference.

aminaahmed1803:
Hi,

I changed the docstrings to the numpydoc style. If I
missed something, please let me know.

I do not know how to write tests but I can always learn. I will read up on
the testing guide!

Additonally, I can't change the name to signal.py because that causes a
circular import issue. Numpy already has a module called signal.py

Best,
Amina

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,75 @@
+"""_summary_
+"""
+
+import numpy as np
+import scipy.io as spio
+import scipy.integrate as sp
+from scipy import signal
```
StanczakDominik, line 7: > Additonally, I can't change the name to signal.py because that causes a circular import issue. Numpy already has a module called signal.py

That's not technically a circular import issue. You can, since it's sort of necessary here, do

```suggestion
import scipy.signal
```

And then you can refer to this later on as
```python
    b, a = scipy.signal.butter(order, normal_cutoff, btype='highpass', analog=False)
```

plasmapy/analysis/magnetics.py
```diff
@@ -0,0 +1,43 @@
...
+    field_arr: `numpy.ndarray`
+        The array containing the magnetic field fluctuations. 
+
+    new_time: `numpy.ndarray`
+        The corresponding time series.   
+    """
+    if len(bdot_data) != len(times):
+        raise Exception("length of time and voltage arrays in not equal\n")
+
+    bdot_data[:] = [x / loop_area for x in bdot_data]
```
StanczakDominik, line 39:
```suggestion
    bdot_data /= loop_area
```
As you wrote it, this would run the division via Python rather than numpy (slow!) and modify the array in-place, which is fine for performance reasons, but really, really, really needs to be explicitly stated in the documentation.

plasmapy/analysis/magnetics.py
```diff
@@ -0,0 +1,43 @@
...
+    field_arr = sp.cumtrapz(bdot_data, times)  # Tesla
+    new_time = times[1:]
```
StanczakDominik, line 41: I definitely agree with Steve here: https://github.com/PlasmaPy/PlasmaPy/issues/1648#issuecomment-1200451551 that if we do `initial = 0` in `cumtrapz`, we don't need to modify the `times` array and could skip the whole `new_time` thing.

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,75 @@
...
+    Parameters
+    ----------
+    data_array: `numpy.ndarray`
+       An array containing data from which offset should be removed.
+
+    start_idx: int
+       Index of value in array from where to start calulating offset.
+
+    end_idx: int
+       Index of value in array from where to end calulating offset.
```
StanczakDominik, line 24: Typos:
```suggestion
    start_idx: int
       Index of value in array from where to start calculating offset.

    end_idx: int
       Index of value in array from where to end calculating offset.
```

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Returns
+    -------
+    data_array: `numpy.ndarray`
+       the original data array with the offset removed.
+
+    """
+    data_array = np.array(data_array)
+    meansub = np.mean(data_array[start_idx:end_idx])
+    data_array[:] = [x - meansub for x in data_array]
```
StanczakDominik, line 34: Same caveat as above, don't use loops for this!

plasmapy/analysis/sig_analysis.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+
+import numpy as np
+import scipy.io as spio
+import scipy.integrate as sp
+from scipy import signal
+from warnings import warn
+
+
+def remove_offset(data_array: np.ndarray, start_idx: int, end_idx: int) -> np.ndarray:
```
StanczakDominik, line 11: This whole function seems overly complicated for a two-liner:

```python
offset_region = data_array[start_idx, end_idx]
data_array -= offset_region.mean()
```

Perhaps we could skip it?

StanczakDominik (changes requested):

aminaahmed1803:
Hi,

Thank you for the feedback.
I made the suggested changes except for signal.py
I tried what you recommended: changing *from scipy import signal *to *import
scipy.signal *and using *scipy.signal.butter* instead of *signal.butter*
However, it gave me the following error:

  File
"/Users/aminaahmed/Documents/Project_Plasma/PlasmaPy/plasmapy/analysis/signal.py",
line 4, in <module>

    import numpy as np

  File
"/Users/aminaahmed/opt/anaconda3/lib/python3.9/site-packages/numpy/__init__.py",
line 140, in <module>

    from . import core

  File
"/Users/aminaahmed/opt/anaconda3/lib/python3.9/site-packages/numpy/core/__init__.py",
line 100, in <module>

    from . import _add_newdocs_scalars

  File
"/Users/aminaahmed/opt/anaconda3/lib/python3.9/site-packages/numpy/core/_add_newdocs_scalars.py",
line 9, in <module>

    import platform

  File "/Users/aminaahmed/opt/anaconda3/lib/python3.9/platform.py", line
119, in <module>

    import subprocess

  File "/Users/aminaahmed/opt/anaconda3/lib/python3.9/subprocess.py", line
49, in <module>

    import signal

  File
"/Users/aminaahmed/Documents/Project_Plasma/PlasmaPy/plasmapy/analysis/signal.py",
line 5, in <module>

    import scipy.io as spio

  File
"/Users/aminaahmed/opt/anaconda3/lib/python3.9/site-packages/scipy/__init__.py",
line 71, in <module>

    from numpy import __version__ as __numpy_version__

ImportError: cannot import name '__version__' from partially initialized
module 'numpy' (most likely due to a circular import)
(/Users/aminaahmed/opt/anaconda3/lib/python3.9/site-packages/numpy/__init__.py)

I didn't know what to do with the error so I discarded the changes. Please
let me know what you think.

Additionally, for the last comment:

*This whole function seems overly complicated for a two-liner:*

*offset_region = data_array[start_idx, end_idx]

data_array -= offset_region.mean()*

*Perhaps we could skip it?*

I didn't quite understand what you are saying. Should I change the code or
just completely remove the function?
It would be great if you could rephrase it so I can understand it better.

Best,
Amina

StanczakDominik:
Hey, I've made a couple changes to your branch, please pull them via `git pull` :)

> I didn't quite understand what you are saying. Should I change the code or
just completely remove the function? It would be great if you could rephrase it so I can understand it better.

Yeah, sorry, that's on me - communication, especially clear communication, can be difficult 😅  I do think that since this function can be expressed as a numpy one-liner, it isn't necessary and can be removed - which I did in [6e645ba](https://github.com/PlasmaPy/PlasmaPy/pull/1652/commits/6e645ba3e2b98d198a97771ae41ea53089f2093d). 

I also could not reproduce your `signal`-related error. Can you tell me whether you're still getting it after pulling in the changes?

aminaahmed1803:
Hi,

Sorry for the late reply. I am in a different country right now.
I still got the error with signal.py when I pulled the changes. However,
since you are not getting the error, I think the problem is with my
environment. I have had issues with it before so I don't think this one is
any different.
I'll try to figure out what is wrong at my end but since you are not
getting the error, I think the code is probably fine.

Best,
Amina

StanczakDominik:
It does look like an env problem! Since from the previous error logs it looks like you're using anaconda, I'll point out the instructions over at https://docs.plasmapy.org/en/stable/contributing/install_dev.html#conda, ending at and including the `pip install -e .` part. If you still have issues, I'm sure we can figure them out :)

PR #1656, opened 2022-08-01
Describe code contribution workflow in contributor guide

### LINKED ISSUE #760: Describe the development workflow in docs

namurphy:
The goal of this PR is to add a page that describes the code contribution workflow to PlasmaPy using `git` and GitHub.  It'll be based on content that we have in the code guide already, as well as [Astropy's workflow guide](https://docs.astropy.org/en/latest/development/workflow/development_workflow.html).  I'll need to figure out the overlap with `docs/contributing/install_dev.rst`.

Closes #760.  Related to #1178


docs/contributing/getting_ready.rst
```diff
@@ -0,0 +1,300 @@
...
+
+      git config --global core.editor emacs
+      git config --global core.editor notepad
+
+   For more editor options, see this page on `git commands for setup and
+   config`_.
+
+#. `Add a new SSH key to your GitHub account`_.
+
+Initial setup
```
mohawk811, line 124: A cool idea could be to integrate short recorded gifs of the steps you reference to visually show the step for further clarity. It could also make the page feel more alive if there are more visuals.

StanczakDominik: If we're going in that direction, I'll point out the existence of asciinema; and then we could use https://github.com/asciinema/agg to turn that into gifs, even!

docs/contributing/doc_guide.rst
```diff
@@ -1,5 +1,8 @@
 .. _documentation guide:
```
ejohnson-96, line 1: L515: Undefined substitution referenced "docs/api_static/"
L848: Undefined substitution referenced "|docs/_static/|_"
L1301: Same as L848

docs/contributing/coding_guide.rst
```diff
@@ -861,97 +861,6 @@ Exceptions to these guidelines should only be made when there are major
 improvements or fixes to upstream functionality or when other required
 packages have stricter requirements.
 
-.. The following section will be moved off of this page in #1656
-
```
ejohnson-96, line 865: L25: undefined label: 'code-contribution'

ejohnson-96 (commented):
Have read over the coding guide etc in changing the docs, seems to be a good guide. Though I was not actively coding while doing the review. Overall pull request appears ready, updates are only needed to some of the doc strings as listed in the respective docs.

namurphy:
Thank you for the review, @ejohnson-96!  I fixed the broken references, and then broke a few more references, and then fixed them too.  If this pull request looks good to you, could you [approve the pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/approving-a-pull-request-with-required-reviews) when you have a chance so this can be merged?  Thanks again!

ejohnson-96 (commented):
👍

ejohnson-96 (approved):

PR #1658, opened 2022-08-04
Discuss lite-functions and aliases in the coding guide

namurphy:
This PR adds sections to the coding guide on lite functions and aliases. This is another excerpt from #1305.


docs/contributing/code_guide.rst
```diff
@@ -418,6 +418,114 @@ by an angular frequency to get a length scale:
...
+
+* Lite-functions are bound to their normal version as the ``lite``
+  attribute using the
+  `~plasmapy.utils.decorators.lite_func.bind_lite_func` decorator.
+
+* Each lite-function should be decorated with
+  `~plasmapy.utils.decorators.helpers.preserve_signature`.
+
+* A lite-function should usually be decorated with ``@numba.njit`` (or
+  the like) as a just-in-time compiler. If a decorator from `numba` is
```
StanczakDominik, line 526:
```suggestion
* A lite-function would usually be decorated with ``@numba.njit`` (or
  the like) for just-in-time compilation. If a decorator from `numba` is
```

That rings better to me, but up to you on whether you like it!

StanczakDominik (approved):

docs/common_links.rst
```diff
@@ -175,6 +182,9 @@
    nitpick_ignore_regex in docs/conf.py so that it doesn't get counted
    as an error in a nitpicky doc build (e.g., tox -e doc_build_nitpicky).
 
+.. _`astropy.units`: https://docs.astropy.org/en/stable/units/index.html
+.. |astropy.units| replace:: `astropy.units`
+
```
rocco8773, line 187:
```suggestion
```

Since this replacement isn't used anywhere, it should be deferred to a PR that utilizes it.  Are we sure `` `astropy.units` `` does not already link to the appropriate documentation?  That is, do we actually need a replacement to accomplish the linking?

rocco8773 (changes requested):
I have a little bit of a dilemma with these two sections.  We're basically repeating/duplicating ourselves in three different locations for both aliases and lite-functions: (1) in the glossary, (2) in the [summary for the aliases and lite-functions sections](https://github.com/PlasmaPy/PlasmaPy/blob/5a51592a91a88e288f3ec3fa86518fabd0b02d67/docs/conf.py#L33-L64), and (3) here.  It's going to be SUPER easy getting these three locations out of sync.

I don't think I have a "best solution" for this at the moment.  My first thought would be to update locations 1 and 2 to have a very top-level summary and then refer to location 3 for additional details.  An even better idea, if possible, would be to have locations 1 and 2 automatically repeat the first paragraph of location 3.  So, if the location 3 paragraph is updated then the other locations will also updated.  If we did that, then there would need to be a note indicating the relationship.  ... Like I said, I don't think I have a "best solution" yet.

namurphy:
> I have a little bit of a dilemma with these two sections. We're basically repeating/duplicating ourselves in three different locations for both aliases and lite-functions: (1) in the glossary, (2) in the [summary for the aliases and lite-functions sections](https://github.com/PlasmaPy/PlasmaPy/blob/5a51592a91a88e288f3ec3fa86518fabd0b02d67/docs/conf.py#L33-L64), and (3) here. It's going to be SUPER easy getting these three locations out of sync.
> 
> I don't think I have a "best solution" for this at the moment. My first thought would be to update locations 1 and 2 to have a very top-level summary and then refer to location 3 for additional details. 

Yep, I'm worried about this too.  My inclination for (1) and (2) has been to include only the information that is unlikely to change.  The implementation details are the most likely to change, and those would be included in (3).  This seems in alignment with your first thought above.

> An even better idea, if possible, would be to have locations 1 and 2 automatically repeat the first paragraph of location 3. 

We could potentially do this via a reST substitution, though we'd want to do ``` :term:`alias` ``` in (2) and (3) and not in (1).  I'm hesitating since we could end up overengineering it at this point...though it could end up being useful to have this infrastructure in place (e.g., for parameter definitions for ``B``, etc. in functions).

I think I'll go with the first thought for the time being.

docs/glossary.rst
```diff
@@ -53,29 +60,20 @@ Glossary
...
+      numbers and |array_like| inputs that are implicitly assumed to be
+      in SI units. The name of a lite-function ends with ``_lite``. A
+      lite-function can be accessed as the ``lite`` attribute of the
+      corresponding regular function.
+
+      .. caution::
+
+         Unlike most `~plasmapy.formulary` functions, no validations are
+         performed on the arguments provided to a lite-function for the
+         sake of computational efficiency.
```
namurphy, line 76: For this revised glossary entry, I attempted to include only the characteristics of lite-functions that are unlikely to change.

StanczakDominik (approved):

docs/conf.py
```diff
@@ -34,29 +34,28 @@
...
-            <https://docs.astropy.org/en/stable/units/>`_ for user convenience and
-            to mitigate calculation errors from inappropriate units, but this
-            comes at the sacrifice of speed.  While this penalty is not significant
-            for typical use, it can become substantial during intensive numerical
-            calculations. **Lite-functions** are introduced for the specific case
-            where speed matters, but **[USER NOTICE]** this comes with the
-            reduction of safeguards so a user needs to know what they are doing!
-            For additional details look to the glossary entry for
-            :term:`lite-function`.
+            :term:`Lite-functions` are lightweight versions of existing
```
rocco8773, line 49:
```suggestion
            :term:`Lite-functions` are lite-weight versions of existing
```

[optional] Just to stay consistent with the lexicon.

docs/conf.py
```diff
@@ -34,29 +34,28 @@
...
+            `plasmapy` functions that are intended for applications where
+            computational efficiency matters most. Lite-functions accept
+            numbers and NumPy arrays that are implicitly assumed to be
+            in SI units, and do not accept |Quantity| objects as inputs.
+
+            .. caution::
+
+               Lite-functions do not include the safeguards that are
+               included in most `plasmapy.formulary` functions.
```
rocco8773, line 58:
```suggestion
               Lite-functions do not include the safeguards that are
               included in most `plasmapy.formulary` functions.  So,
               make sure to double-check and triple-check your
               implementation!!!
```

I'm just looking for some kind of statement that iterates it's the user's responsibility to to ensure the implementation is correct.

docs/conf.py
```diff
@@ -34,29 +34,28 @@
...
-            "PlasmaPy provides short-named (alias) versions of the most "
-            "common plasma functionality.  These aliases are only given to "
-            "functionality where there is a common lexicon in the community, "
-            "for example `~plasmapy.formulary.frequencies.plasma_frequency` "
-            " has the alias `~plasmapy.formulary.frequencies.wp_`.  All aliases "
-            "in PlasmaPy are denoted with a trailing underscore ``_``."
+            """
+            PlasmaPy provides :term:`aliases` of the most common plasma
+            functionality for user convenience. Aliases in PlasmaPy are
+            denoted with a trailing underscore.
```
rocco8773, line 40:
```suggestion
            denoted with a trailing underscore, e.g. ``alias_``.
```

Also needs a link to the "Aliases" section in the doc guide.

docs/conf.py
rocco8773, line 49: Needs a link to the Also needs a link to the "Lite-Functions" section in the doc guide.

docs/conf.py
namurphy: Added something similar.

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,118 @@ by an angular frequency to get a length scale:
...
 
+.. _aliases:
+
+Aliases
+=======
+
+An :term:`alias` is an abbreviated version of a commonly used function
+that is intended for interactive use. For example,
+`~plasmapy.formulary.speeds.va_` is an alias to
+`~plasmapy.formulary.speeds.Alfven_speed`.
```
rocco8773, line 492:
```suggestion
An :term:`alias` is an abbreviated version of a commonly used function.
For example, `~plasmapy.formulary.speeds.va_` is an alias to
`~plasmapy.formulary.speeds.Alfven_speed`.  It is intended to give
users the option for shortening their code while maintaining
readability and explicit meaning.  As such, :term:`aliases` are given
to functionality that already has a widely used symbol in the
community's literature.
```

[semi-optional]  The phrase "...intended for interactive use." threw me off because it does not seem to have a real meaning.  Trying to come up with something else, sent me down this rabbit hole.

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,118 @@ by an angular frequency to get a length scale:
...
+  for :math:`c/ω_p`\ .
+
+* The name of an alias should end with a trailing underscore.
+
+* Each alias should have a one-line docstring that refers users to the
+  original function.
+
+* The name of the main function should be included in ``__all__`` near
+  the top of each module, and the name of the alias should be included
+  in ``__aliases__``, which will then get appended to ``__all__``.
```
rocco8773, line 526:
```suggestion
* The name of the main function should be included in ``__all__`` near
  the top of each module, and the name of the alias should be included
  in ``__aliases__``, which will then get appended to ``__all__``.
  This is done so both the :term:`alias` and its parent function get
  properly documented.
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,118 @@ by an angular frequency to get a length scale:
...
+==============
+
+Most functions in `plasmapy.formulary` accept |Quantity| instances as
+arguments and use |validate_quantities| to verify that |Quantity|
+arguments are valid. The use of |Quantity| operations and validations do
+not noticeably impact performance during typical interactive use, but
+the performance penalty can become significant for numerically intensive
+applications.
+
+A :term:`lite-function` is a lightweight version of another `plasmapy`
```
rocco8773, line 540:
```suggestion
A :term:`lite-function` is a lite-weight version of another `plasmapy`
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,125 @@ by an angular frequency to get a length scale:
...
+
+* Lite-functions assume SI units for all of arguments that represent
+  physical quantities.
+
+* Lite-functions should be defined immediately before the normal version
+  of the function.
+
+* Lite-functions are bound to their normal version as the ``lite``
+  attribute using the
+  `~plasmapy.utils.decorators.lite_func.bind_lite_func` decorator.
```
rocco8773, line 597:
```suggestion
* Lite-functions should be defined immediately before the normal version
  of the function.

* Lite-functions should be used by their associate non-lite counterpart,
  except for well reasoned exceptions.  This is done to reduce code
  duplication.

* Lite-functions are bound to their normal version as the ``lite``
  attribute using the
  `~plasmapy.utils.decorators.lite_func.bind_lite_func` decorator.  This
  allows the lite function to also be accessed like
  ``thermal_speed.lite()``.
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,125 @@ by an angular frequency to get a length scale:
...
+
+* Each lite-function should be decorated with
+  `~plasmapy.utils.decorators.helpers.preserve_signature`.
```
rocco8773, line 600:
```suggestion
* If a lite-function is decorated with something like ``@njit``,
  then it should also be decorated with
  `~plasmapy.utils.decorators.helpers.preserve_signature`.  This
  preserves the function signature so interpreters can still
  give hints about function arguments.
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,125 @@ by an angular frequency to get a length scale:
...
+
+* A lite-function should usually be decorated with ``@numba.njit`` (or
+  the like) for just-in-time compilation. If a decorator from `numba` is
```
rocco8773, line 603:
```suggestion
  the like) for just-in-time compilation. If a decorator from |numba| is
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,128 @@ by an angular frequency to get a length scale:
...
+  not able to be used, then an alternative is Cython_.
```
rocco8773, line 604:
```suggestion
* When possible, a lite-function should incorporate `numba's
  just-in-time compilation 
  <https://numba.pydata.org/numba-doc/latest/reference/jit-compilation.html>`__
  or utilize Cython_.  At a minimum any "extra" code beyond the raw
  calculation should be removed.
```

namurphy:
@rocco8773 — I worked in most of the suggestions so please take another look!

I also noticed that codespell starting failing, presumably because of a new release.  This PR includes an update on which words codespell should skip, so it will fix the failing tests.

rocco8773:
Just waiting for the docs to finish building so I can do a final look over.

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,146 @@ by an angular frequency to get a length scale:
...
+  of the function.
+
+* Lite-functions should be used by their associate non-lite counterpart,
+  except for well reasoned exceptions.  This is done to reduce code
+  duplication.
+
+* Lite-functions are bound to their normal version as the ``lite``
+  attribute using the
+  `~plasmapy.utils.decorators.lite_func.bind_lite_func` decorator.  This
+  allows the lite function to also be accessed like
```
rocco8773, line 609:
```suggestion
  allows the lite-function to also be accessed like
```

docs/contributing/coding_guide.rst
```diff
@@ -481,6 +481,146 @@ by an angular frequency to get a length scale:
...
+  `~plasmapy.utils.decorators.helpers.preserve_signature`.  This
+  preserves the function signature so interpreters can still
+  give hints about function arguments.
+
+* When possible, a lite-function should incorporate `numba's
+  just-in-time compilation
+  <https://numba.pydata.org/numba-doc/latest/reference/jit-compilation.html>`__
+  or utilize Cython_.  At a minimum any "extra" code beyond the raw
+  calculation should be removed.
+
```
rocco8773, line 623:
```suggestion
* When possible, a lite-function should incorporate `numba's
  just-in-time compilation
  <https://numba.pydata.org/numba-doc/latest/reference/jit-compilation.html>`__
  or utilize Cython_.  At a minimum any "extra" code beyond the raw
  calculation should be removed.

* The name of the original function should be included in ``__all__``
  near the top of each module, and the name of the lite-function
  should be included in ``__lite_funcs__``, which will then get
  appended to ``__all__``. This is done so both the :term:`lite-function`
  and the original function get properly documented.
```

docs/glossary.rst
```diff
@@ -53,29 +62,24 @@ Glossary
...
-        SI units.
-      - **UNITS ARE NOT VALIDATED BY THE FUNCTIONS.**  It is solely up to
-        the user to pass in parameters with the correct units.
-      - When possible, lite-functions are decorated with `numba.jit`
-        (or the like) for an additional speed boost.
+      A lightweight version of an existing `plasmapy` function intended
+      for applications where computational efficiency is most important.
+      While most `~plasmapy.formulary` functions accept |Quantity|
+      objects created using `astropy.units`, lite-functions accept
+      numbers and |array_like| inputs that are implicitly assumed to be
```
rocco8773, line 69: The `|array_like|` is not working correctly.  Not sure what the reason is at the moment.

docs/glossary.rst
```diff
@@ -9,6 +9,15 @@ Glossary
...
    :sorted:
 
+   alias
+   aliases
+      An abbreviated version of a commonly used function that is
+      intended primarily for interactive use. Aliases are named with a
+      trailing underscore. For example, `~plasmapy.formulary.speeds.va_`
+      is an alias for `~plasmapy.formulary.speeds.Alfven_speed`). For
+      further details, please refer to the :ref:`contributor guide's
+      section on aliases <aliases>`.
```
rocco8773, line 19:
```suggestion
      An abbreviated version of a commonly used function.  For
      example, `~plasmapy.formulary.speeds.va_` is an alias for
      `~plasmapy.formulary.speeds.Alfven_speed`.  Aliases are
      named with a trailing underscore. 
      
      For further details, please refer to the :ref:`contributor guide's
      section on aliases <aliases>`.
```

Just trying to match this up with the rewording we did to the Aliases section.

docs/glossary.rst
rocco8773: It's because your branch is not up-to-date with `main`, so `|array_like|` is not defined in `common_links.rst`.

namurphy: Just updated it — thanks!

namurphy:
pre-commit.ci autofix

docs/glossary.rst
namurphy: Since it's a glossary entry and fairly short, I'm inclined to keep it a single paragraph.

rocco8773: I'm fine with that. I just wanted to get that confusing "interactive use" phrase out of there and replace it with the same language that we put into the aliases section.

rocco8773 (approved):
Looks good! Thanks for doing this.  These really needed to be summarized in a better way than they originally were.

Issue #1276, opened 2021-09-23
Consolidate requirements in a single place


namurphy:
Right now, when we change requirements, we need to change them in three places:

 - `setup.cfg`
 - `requirements/*.txt`
 - `requirements/environment.yml`

As I have empirically demonstrated 😬, there's a good chance that we'll end up forgetting to update them in all three places and end up with inconsistencies, even with the comments that state the other places where things need to be changed.  

We should consolidate requirements in a single place, if there is a reasonable way to go about it.  The problem is that I don't know if there currently is a reasonable way to do this.  For example, there is information in `environment.yml` that is not contained in the other two places (i.e. which channels to use and whether packages get installed via those channels or via `pip`).  

The previous major changes to the requirements in #789 were a huge and much-appreciated improvement over what was there before, and I really like the way that the requirements are split up in each of the `setup.cfg` and the `requirements/*.txt` files, so it'd be really helpful to keep a structure like if there is a way to consolidate them.

VladimirSiv:
Hello all,

I was just looking for an open source project to contribute to, when I stumbled upon PlasmaPy.
Awesome stuff! Reminds me of my studies of theoretical physics.

Anyway, I see that you have a problem with `setup.py` and `requirements.txt`.
They serve different purposes but can be pain to manage both independently.

As far as I know there is a way to combine them as stated here: [requirements.txt vs setup.py](https://stackoverflow.com/a/49684835/10947997)

Copying here for the sake of reference:


----
> Quoting user `init_js`:

There is one trick you may consider to avoid duplicating your list of dependencies between `requirements.txt` and `setup.py`. If you have written a fully working `setup.py` for your package already, and your dependencies are mostly external, you could consider having a simple `requirements.txt` with only the following:

```
 # requirements.txt
 #
 # installs dependencies from ./setup.py, and the package itself,
 # in editable mode
 -e .

 # (the -e above is optional). you could also just install the package
 # normally with just the line below (after uncommenting)
 # .
```

The -e is a special pip install option which installs the given package in editable mode. When pip -r requirements.txt is run on this file, pip will install your dependencies via the list in ./setup.py. The editable option will place a symlink in your install directory (instead of an egg or archived copy). It allows developers to edit code in place from the repository without reinstalling.

You can also take advantage of what's called "setuptools extras" when you have both files in your package repository. You can define optional packages in setup.py under a custom category, and install those packages from just that category with pip:

```python
# setup.py
from setuptools import setup
[...]

PR #1659, opened 2022-08-04
Add sections to coding guide on imports and requirements

namurphy:
This is another spinoff of #1305 that describes guidelines related to imports and requirements.  

When we are done with #1276, we should make sure to update the requirements section being added here.


StanczakDominik (approved):
👍🏻

PR #1660, opened 2022-08-04
Add sections on comments and error messages in coding guide

namurphy:
This PR adds sections to the coding guide on best practices for writing comments and error messages. This PR is possibly the last spinoff of #1305.

~~It might be worthwhile to briefly comment on LBYL vs. EAFP here.  The LBYL approach allows us to produce better error messages, which are really important for a library...even though EAFP usually leads to cleaner code.~~

This PR has been brought to you by my desire to procrastinate washing dishes. 🍽️


docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
 PlasmaPy fork on GitHub.  After a pull request is merged into the
```
rocco8773, line 279: The coding guide is getting rather long.  I'm thinking it might be time to add a ``.. contents:` to the top of the page.  Thoughts?

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
+  .. code-block:: python
+
+     # collision frequency
+     nu = 1e6 * u.s ** -1
+
+  could be changed to:
+
+  .. code-block:: python
+
+     collision_frequency = 1e6 * u.s ** -1  # no comment needed!
```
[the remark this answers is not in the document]
namurphy: Oh good point!

docs/contributing/code_guide.rst
namurphy: Yes!

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
 
-Commit Messages
----------------
-Good commit messages communicate context and intention to other
-developers and to our future selves.  They provide insight into why we
-chose a particular implementation, and help us avoid past mistakes.
 
-Suggestions on `how to write a git commit message
-<https://cbea.ms/git-commit>`_:
+Comments
```
rocco8773, line 283: I wonder if establishing a "golden rule" would be useful here.  Maybe even put it into an admonition.  I personally get lost in a list of guide points, so having a golden rule to check against is what I find most useful.  The golden rule I tend to tell people goes something like:

> Your code should read like a book, that is the full meaning of your functionality should be attainable by reading the code.  A comment is only used when code itself can not communicate the full detailed meaning.

I'm sure there's a better way of phrasing this, but I think it covers all the points you say below.  I think pairing the "golden rule" with your points as examples would be overall more effective.

Thoughts?

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
+* Use comments to communicate information that you wish you knew before
+  starting to work on a particular section of code, including
+  information that took some time to learn.
+
+* Use comments to communicate information that the code cannot,
+  such as why an alternative approach was *not* taken.
+
+* When possible, include enough contextual information that the comment
+  will make sense even if it is displaced from the code it was
+  originally describing.
```
rocco8773, line 320: I'm not sure how to visualize this point.  Is there a good example.  I think it's a little confusing without more context. (Sorry, I couldn't resist saying that last bit.)

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
+* Use comments to include references to books or articles that describe
+  the equation, algorithm, or software design pattern that is being
+  implemented. Even better, include these references in docstrings.
+
+* Provide enough contextual information in the comment for a new user
+  to be able to understand it.
+
+* Remove commented out code before merging a pull request.
+
+* When updating code, be sure to update the comments too!
```
rocco8773, line 331:
```suggestion
* When updating code, be sure to review and update, if necessary, associated comments too!
```

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
+
+* When a comment is used as the header for a section of code, consider
```
rocco8773, line 333: You're going to make me rage quit with this point. 😂 I'd like to see some kind of comment in here that there's a "subjective" balancing act with the `extract function refactoring pattern`.  Overall it's good, but when used on a too fine of a scale it can significantly reduce code readability.

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
 
-* Separate subject from body with a blank line
+Error messages
+==============
 
-* Limit the subject line to 50 characters
+Error messages are a vital but underappreciated form of documentation.
+A good error message can help someone pinpoint the source of a problem
+in seconds, while a cryptic or missing error message can lead to hours
+of frustration.
```
rocco8773, line 385: I had a `sphinx` error like this yesterday.  It didn't even tell me what file the error was in. Ahhhh... 😤

docs/contributing/code_guide.rst
namurphy: Addressing it in #1662 for not only the coding guide, but also the other pages in the contributor guide.

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
 
-* Capitalize the subject line
+* Use error messages to indicate the source of the problem while
+  providing enough information for the user to fix it. When possible,
+  make it clear what the user should do next.
```
rocco8773, line 389:
```suggestion
* Use error messages to indicate the source of the problem while
  providing enough information for the user to troubleshoot it. When
  possible, make it clear what the user should do next.
```

Only because errors can be the side effect of another issue.

docs/contributing/code_guide.rst
namurphy: YEP.

docs/contributing/code_guide.rst
namurphy: That's a good point!  This is a good place to put in a tradeoff analysis and how there are different factors we need to balance.

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
 
-* Do not end the subject line with a period
+* Include diagnostic information when appropriate.  If an occurred at
+  a single point in an array operation, including the coordinates of
+  where the error happened can help the user better understand the cause
+  of the error.
```
rocco8773, line 394:
```suggestion
* Include diagnostic information when appropriate.  For example, if an
  error occurred at a single index in an array operation, then including
  the index where the error happened can help the user better understand
  the cause of the error.
```

docs/contributing/code_guide.rst
namurphy: Good idea!  I'll add something like this in, likely after I get back from visiting family.

docs/contributing/code_guide.rst
namurphy: Thanks!  I was having trouble wording this.

docs/contributing/code_guide.rst
```diff
@@ -279,39 +279,138 @@ to edit the pull request by changing the corresponding branch on your
...
+  - Alternatively, an error message may reference a docstring or a page
+    in the narrative documentation.
 
-* "Errors should never pass silently."  Users should be notified when
-  problems arise by either issuing a warning or raising an exception.
+* Write error messages that are friendly, supportive, and helpful. Error
+  message should never be condescending or blame the user.
 
 * The exceptions raised by a method should be described in the
   method's docstring.  Documenting exceptions makes it easier for
```
rocco8773, line 416: Do we have a doc guide rule on this?  This is one of the most difficult parts of writing a docstring.  And, I often find this part of the docstring is like a comment, always out of date.

namurphy: Hm...perhaps we could emulate the [numpydoc section](https://numpydoc.readthedocs.io/en/latest/format.html#raises) on this, and say that the `Raises` section should be used judiciously and only include non-obvious exceptions.  

- [ ] Do ↑
- [ ] Update the doc guide section on this too

> And, I often find this part of the docstring is like a comment, always out of date.

All too true...

namurphy: Updated this, and moved it to the doc guide as a means of procrastinating packing.

docs/contributing/code_guide.rst
namurphy: Added in a discussion on this below.

namurphy:
So that I have it handy when I get back from vacation, the things from code review that I still need to address are...

 - [ ] Add in the golden rule
 - [ ] Provide more context on providing more context

docs/contributing/code_guide.rst
namurphy: I decided to delete this point since there is already another one that says to provide context.

rocco8773 (approved):

PR #1661, opened 2022-08-04
Make unit validators compatible with `u.Quantity[u.T]`-style annotations

### LINKED ISSUE #1413: Adopt ``Quantity`` type annotations that preserve units
For many of our functions, we use annotations for units:

```Python
def func(x: u.m, t: u.s) -> u.m / u.s:
    return (x / t).to(u.m / u.s)
```

These are not *type hint* annotations, since the expected value is a `Quantity` in most situations.  Instead the annotations are used by decorators like our validators and `astropy.units.quantity_input`.  Unfortunately, this means that we can't use static type checkers like [mypy](http://mypy-lang.org/) to check PlasmaPy code. 😿  Using mypy has the potential to help us find bugs, as well as errors in existing type hint annotations (#268).

Astropy now allows for [Quantity type annotations that preserve units](https://docs.astropy.org/en/stable/units/type_hints.html#preserving-units).  Both `typing.Union` and `typing.Optional` are supported.

```Python
def func(x: Quantity[u.kpc]) -> Quantity[u.m]:
    return x << u.m
```

We should migrate towards using `Quantity` annotations that preserve units in PlasmaPy.  However, before we do this, we would need to change our validators to make them work with this style of annotation.

### LINKED ISSUE #1414: Make validators compatible with `Quantity` type annotations that preserve units
In #1413, I raised an issue to adopt `astropy.units.Quantity` [Quantity type annotations that preserve units](https://docs.astropy.org/en/stable/units/type_hints.html#preserving-units).  

Before we can resolve #1413, we would need to make the decorators in `plasmapy.utils.decorators.validators` compatible with annotations like:
```Python
import astropy.units as u
from typing import Optional, Union

def f(x: u.Optional[u.Quantity[u.m]]) -> u.Quantity[u.m]):
    ...

def g(x: Union[u.Quantity[u.K], u.Quantity[u.eV]]) -> Optional[u.Quantity[u.K]]:
    ...
```

This might be a good first issue for someone who is familiar with Python but not PlasmaPy.

StanczakDominik:
Possibly closes #1413 and also closes #1414, maybe? Let's see if this blows up.

Essentially what happens when you do, say, `def f(B: u.Quantity[u.T])` is that the annotation becomes a `typing._AnnotatedAlias`. Apparently the canonical way of reading type hints for a function is [`typing.get_type_hints`](https://docs.python.org/3/library/typing.html#typing.get_type_hints), but since we didn't go that route for the existing code, I didn't want to start rewriting the whole thing just yet. Instead, the annotation has a `__metadata__` field, an iterable of things within the `[]` parenthesis. So I just picked the first item in that for now.

I've also added this kind of annotation in `frequencies` for now, to see whether they'd explode.

      docstrings.


StanczakDominik:
@rocco8773 wanted to ping you for short feedback on what you think about this approach :)

StanczakDominik:
I wonder whether we shouldn't deprecate the old, `u.m` style syntax.

StanczakDominik:
Oh yeah, here's the crux of the issue. [`typing._AnnotatedAlias`, which is somehow related to `typing.Annotated` as I understand it, requires python 3.9](https://docs.python.org/3/library/typing.html#typing.Annotated).

StanczakDominik:
To answer potential questions about `__metadata__`, here's Guido:

> [It looks like __metadata__ was *meant* to be a public attribute, but somehow overseen when the PEP and docs were written. :-(](https://bugs.python.org/issue45380#msg412485)

StanczakDominik:
- [ ] no need to bump to 3.9 since we can do https://github.com/astropy/astropy/blob/58d45f2d4bc2147ea70c51fbf6abcf2f2cf7134a/astropy/units/_typing.py#L1

StanczakDominik:
Looks like this won't work without a proper rewrite of `check_units` :(

PR #1663, opened 2022-08-06
[SLA]: Plotter function for Swept Langmuir floating potential 

rocco8773:
This is a plotter function for graphing the results of `find_floating_potential()` in the swept Langmuir analysis module.


rocco8773:
@namurphy I'm getting this PR backup and running.  **I'm curious if you have any thoughts on writing tests for this functionality.**  Since it is just plotting functionality, I'm not sure how to test it and I feel that we probably do not need to write tests for it.

namurphy:
> **I'm curious if you have any thoughts on writing tests for this functionality.**

I'm wondering if a regression test would make sure that the plot doesn't change over time from a known state.  There are also packages like `pytest-mpl` that might help with this sort of thing. 🤔

rocco8773:
> I'm wondering if a regression test would make sure that the plot doesn't change over time from a known state. There are also packages like `pytest-mpl` that might help with this sort of thing. 🤔

Thanks for the suggestion.  I'll precariously crawl down the rabbit hole...

rocco8773:
pre-commit.ci autofix

PR #1664, opened 2022-08-09
Add `critical_density`

JaydenR2305:
Resolves PlasmaPy/hack-week#26

      docstrings.

<!--

### DIFF
--- a/changelog/1664.feature.rst
+++ b/changelog/1664.feature.rst
@@ -1,2 +1,2 @@
-Added the `~plasmapy.formulary.frequencies.plasma_critical_density`
-function to the `~plasmapy.formulary.frequencies` module. This function calculates the critical density of the plasma for a given plasma frequency.
+Add the `~plasmapy.formulary.densities` submodule
+Add the `~plasmapy.formulary.densities.plasma_critical_density` function to the `~plasmapy.formulary.densities` module. This function calculates the critical density of a plasma for a given frequency of radiation.
--- a/changelog/1664.feature.rst
+++ b/changelog/1664.feature.rst
@@ -1,0 +1,2 @@
+Added the `~plasmapy.formulary.frequencies.plasma_critical_density`
+function to the `~plasmapy.formulary.frequencies` module. This function calculates the critical density of the plasma for a given plasma frequency.
--- a/changelog/1664.feature.rst
+++ b/changelog/1664.feature.rst
@@ -1,2 +1,2 @@
 Add the `~plasmapy.formulary.densities` submodule
-Add the `~plasmapy.formulary.densities.plasma_critical_density` function to the `~plasmapy.formulary.densities` module. This function calculates the critical density of a plasma for a given frequency of radiation.
+Add the `~plasmapy.formulary.densities.critical_density` function to the `~plasmapy.formulary.densities` module. This function calculates the critical density of a plasma for a given frequency of radiation.
--- a/docs/formulary/densities.rst
+++ b/docs/formulary/densities.rst
@@ -1,7 +1,7 @@
 .. _densities:
 
 ******************************************************************
-Density related functionality (`plasmapy.formulary.densities`)
+Density Plasma Parameters (`plasmapy.formulary.densities`)
 ******************************************************************
 
 .. currentmodule:: plasmapy.formulary.densities
--- a/docs/formulary/densities.rst
+++ b/docs/formulary/densities.rst
@@ -1,0 +1,9 @@
+.. _densities:
+
+******************************************************************
+Density related functionality (`plasmapy.formulary.densities`)
+******************************************************************
+
+.. currentmodule:: plasmapy.formulary.densities
+
+.. automodapi:: plasmapy.formulary.densities
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -14,9 +14,6 @@ from plasmapy.utils.decorators import validate_quantities
     omega_p={"can_be_negative": False},
     validations_on_return={
         "units": [u.m**-3],
-        "equivalencies": [
-            (u.F * u.kg * u.rad**2 / (u.C**2 * u.m * u.s**2), u.m**-3)
-        ],
     },
 )
 def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
@@ -53,4 +50,4 @@ def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
 
     n_c = m_e * eps0 * omega_p**2 / (e**2)
 
-    return n_c
+    return (n_c/u.rad**2).to(u.m**-3)
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -32,25 +32,16 @@ def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
     n_c : `~astropy.units.Quantity`
         The plasma critical density.
 
-    Raises
-    ------
-    `TypeError`
-        If ``omega_p`` is not a `~astropy.units.Quantity`.
-
-    `~astropy.units.UnitTypeError`
-        If ``omega_p`` is not in correct units.
-
-    `ValueError`
-        If ``omega_p`` is an invalid value.
-
-    Warns
-    -----
-    : `~astropy.units.UnitsWarning`
-        If units are not provided, SI units are assumed.
-
     Notes
     -----
+    The plasma critical density is given by the formula
 
+    .. math::
+        n_{c}=\frac{m_{e}\varepsilon_0\omega_{p}^{2}}{e^{2}}
+
+    where :math:`m_{e}` is the mass of an electron,
+    :math:`\varepsilon_0` is the permittivity of free space, :math:`\omega_{p}`
+    is the plasma frequency, and :math:`e` is the elementary charge.
 
     Examples
     --------
@@ -60,4 +51,6 @@ def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
 
     """
 
-    return m_e * eps0 * omega_p**2 / (e**2)
+    n_c = m_e * eps0 * omega_p**2 / (e**2)
+
+    return n_c
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -50,4 +50,4 @@ def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
 
     n_c = m_e * eps0 * omega_p**2 / (e**2)
 
-    return (n_c/u.rad**2).to(u.m**-3)
+    return (n_c / u.rad**2).to(u.m**-3)
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -54,4 +54,4 @@ def plasma_critical_density(omega: u.rad / u.s) -> u.m**-3:
 
     n_c = m_e * eps0 * omega**2 / (e**2)
 
-    return (n_c / u.rad**2).to(u.m**-3)
+    return n_c.to(u.m**-3, equivalencies=u.dimensionless_angles())
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -48,7 +48,7 @@ def plasma_critical_density(omega: u.rad / u.s) -> u.m**-3:
     --------
     >>> from astropy import units as u
     >>> plasma_critical_density(5e15 * u.rad/u.s)
-    <Quantity 7.8551 1 / m3>
+    <Quantity 7.85519457e+27 1 / m3>
 
     """
 
--- a/plasmapy/formulary/densities.py
+++ b/plasmapy/formulary/densities.py
@@ -1,0 +1,63 @@
+"""Functions to calculate plasma density parameters."""
+__all__ = [
+    "plasma_critical_density",
+]
+
+import astropy.units as u
+
+from astropy.constants.si import e, eps0, m_e
+
+from plasmapy.utils.decorators import validate_quantities
+
+
+@validate_quantities(
+    omega_p={"can_be_negative": False},
+    validations_on_return={
+        "units": [u.m**-3],
+        "equivalencies": [
+            (u.F * u.kg * u.rad**2 / (u.C**2 * u.m * u.s**2), u.m**-3)
+        ],
+    },
+)
+def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
+    r"""Calculate the plasma critical density.
+
+    Parameters
+    ----------
+    omega_p : `~astropy.units.Quantity`
+        The plasma frequency in units of angular frequency.
+
+    Returns
+    -------
+    n_c : `~astropy.units.Quantity`
+        The plasma critical density.
+
+    Raises
+    ------
+    `TypeError`
+        If ``omega_p`` is not a `~astropy.units.Quantity`.
+
+    `~astropy.units.UnitTypeError`
+        If ``omega_p`` is not in correct units.
+
+    `ValueError`
+        If ``omega_p`` is an invalid value.
+
+    Warns
+    -----
+    : `~astropy.units.UnitsWarning`
+        If units are not provided, SI units are assumed.
+
+    Notes
+    -----
+
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> plasma_critical_density(1 * u.rad/u.s)
+    <Quantity 0.00031421 1 / m3>
+
+    """
+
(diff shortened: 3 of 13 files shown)


plasmapy/formulary/parameters.py
```diff
@@ -77,6 +77,7 @@
     ("frequencies", "wlh_"),
     ("frequencies", "upper_hybrid_frequency"),
     ("frequencies", "wuh_"),
+    ("frequencies", "plasma_critical_density"),
```
JaydenR2305, line 80: Is this necessary?

changelog/1664.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added the `~plasmapy.formulary.frequencies.plasma_critical_density`
+function to the `~plasmapy.formulary.frequencies` module. This function calculates the critical density of the plasma for a given plasma frequency.
```
pheuer, line 2: The derivation uses the plasma frequency, but I think it's more accurate to say "calculates the critical density of a plasma for a given frequency of radiation" or something similar.

plasmapy/formulary/frequencies.py
```diff
@@ -4,6 +4,7 @@
     "lower_hybrid_frequency",
     "plasma_frequency",
     "upper_hybrid_frequency",
+    "plasma_critical_density",
```
pheuer, line 7: The way we are organizing the formulary files now (by dimensions) I guess we should actually move this function to a new file called 'densities.py'. You can just copy structure of this file, e.g. include a __all__ dunder with the name of the function and add it to this modules formulary/__init__.py

Don't worry about aliases or lite functions: that doesn't include what you're doing here.

plasmapy/formulary/frequencies.py
```diff
@@ -519,3 +520,56 @@ def upper_hybrid_frequency(B: u.T, n_e: u.m**-3) -> u.rad / u.s:
...
 wuh_ = upper_hybrid_frequency
 """Alias to `~plasmapy.formulary.frequencies.upper_hybrid_frequency`."""
+
+
+@validate_quantities(
+    omega_p={"can_be_negative": False},
+    validations_on_return={
+        "units": [u.m**-3],
+        "equivalencies": [
+            (u.F * u.kg * u.rad**2 / (u.C**2 * u.m * u.s**2), u.m**-3)
```
pheuer, line 530: Why do you need this equivalencies statement?

plasmapy/formulary/parameters.py
pheuer: No I don't think so - this is just an artifact of the way we use to have this organized: all these functions were in `parameters.py`. This code is just for backward compatibility for a few versions. Since your function is new, it doesn't need to go here.

plasmapy/formulary/frequencies.py
```diff
@@ -519,3 +520,56 @@ def upper_hybrid_frequency(B: u.T, n_e: u.m**-3) -> u.rad / u.s:
...
+    `TypeError`
+        If ``omega_p`` is not a `~astropy.units.Quantity`.
+
+    `~astropy.units.UnitConversionError`
+        If ``omega_p`` is not in correct units.
+
+    `ValueError`
+        If ``omega_p`` is an invalid value.
+
+    Warns
```
pheuer, line 558: Is this true for your code, or left over from another docstring?

plasmapy/formulary/frequencies.py
```diff
@@ -519,3 +520,56 @@ def upper_hybrid_frequency(B: u.T, n_e: u.m**-3) -> u.rad / u.s:
...
+    ----------
+    omega_p : `~astropy.units.Quantity`
+        The plasma frequency in radians per second.
+
+    Returns
+    -------
+    n_c : `~astropy.units.Quantity`
+        The plasma critical density.
+
+    Raises
```
pheuer, line 547: Does your code actually raises these exceptions? If so, I guess it must be the validator decorator doing it?

plasmapy/formulary/frequencies.py
```diff
@@ -519,3 +520,56 @@ def upper_hybrid_frequency(B: u.T, n_e: u.m**-3) -> u.rad / u.s:
...
+        ],
+    },
+)
+def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
+    r"""Calculate the plasma critical density.
+
+    Parameters
+    ----------
+    omega_p : `~astropy.units.Quantity`
+        The plasma frequency in radians per second.
```
pheuer, line 540: I would say 'in units of angular frequency' since the input can be any valid Quantity, e.g. deg/hour would work too.

pheuer (commented):
Nice start! I've left some comments on the code. The next step (after creating the densities.py file) will be to create tests using Pytest. You should create a /formulary/tests/test_densities.py file (look at test_frequenices or test_speeds for ideas). Some ideas for tests would be: 

- Use pytest.raises() and pytest.warns() to check that invalid inputs raise the correct exceptions and warnings. 
-  Manually compute 1-2 values of n_c and write an assert test that checks that your function returns the correct value. That would look something like: 
``assert np.isclose(n_c, true_value)``

plasmapy/formulary/frequencies.py
JaydenR2305: Astropy uses those units as the return of the calculation for critical density, I assumed it would be cleaner to return the critical density as 1/m^3

plasmapy/formulary/frequencies.py
JaydenR2305: Yep this is correct, a negative frequency will raise a ValueError

plasmapy/formulary/frequencies.py
JaydenR2305: Yeah all the exceptions come from the decorator, should I keep them ?

plasmapy/formulary/frequencies.py
pheuer: @namurphy just opened a new PR to address this in the guidelines (#1667 ): the docstring should only document errors that are explicitly raised by the function itself, not the validator. So, you should remove these errors + warnings here.

plasmapy/formulary/frequencies.py
pheuer: Oh, yes I guess this accomplishes that. You could also just do that inside this function: 

```python 
n_c = ....
return n_c.to(u.m**-3)
```

I think that would be easier to read, so that's probably my recommendation.

pheuer:
So you're getting a CI error for the new module: 

```
/home/runner/work/PlasmaPy/PlasmaPy/docs/formulary/densities.rst: WARNING: document isn't included in any toctree
```

(you can see that by clicking 'details' next to the failing test)

I think you need to also add a link to this file in \docs\formulary\index.rst. That tells autodoc where to include this file's information in the overall structure of the documentation.

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,63 @@
...
+    ----------
+    omega_p : `~astropy.units.Quantity`
+        The plasma frequency in units of angular frequency.
+
+    Returns
+    -------
+    n_c : `~astropy.units.Quantity`
+        The plasma critical density.
+
+    Raises
```
pheuer, line 35: These exceptions + warnings should be left out of the docstring per #1667

plasmapy/formulary/tests/test_densities.py
```diff
@@ -0,0 +1,33 @@
+"""Tests for functionality contained in `plasmapy.formulary.frequencies`."""
+import astropy.units as u
+import numpy as np
+import pytest
+
+from plasmapy.formulary.densities import plasma_critical_density
+from plasmapy.formulary.frequencies import plasma_frequency
+
+n_i = 5e19 * u.m**-3
```
pheuer, line 9: I would move these definitions into the test, since (at least for now) they are not being used by multiple tests

plasmapy/formulary/tests/test_densities.py
```diff
@@ -0,0 +1,33 @@
...
+ion = "e-"
+
+
+def test_plasma_critical_density():
+    r"""Test the plasma_critical_density function in frequencies.py."""
+
+    omega_p = plasma_frequency(n=n_i, particle=ion)
+    n_c = plasma_critical_density(omega_p)
+
+    assert n_c.unit.is_equivalent(u.m**-3)
```
pheuer, line 19: These two tests are good! However you should split them into two separate test functions, so that if the first fails the other still gets tested. If you want to calculate n_c once for both tests, look into pytest Fixtures

plasmapy/formulary/tests/test_densities.py
```diff
@@ -0,0 +1,33 @@
...
+
+    assert np.isclose(n_c, 5e19 * u.m**-3)
+
+    with pytest.raises(TypeError):
```
pheuer, line 23: I've asked a question about this on #1667 , but I don't think these need to be tested explicitly? If you do keep them, they should also be separate test functions. You can keep that organized by using pytest fixtures and parameterize (e.g. make one function to test errors, parameterized with a list of inputs and expected error types)

plasmapy/formulary/tests/test_densities.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_plasma_critical_density():
+    r"""Test the plasma_critical_density function in frequencies.py."""
+
+    omega_p = plasma_frequency(n=n_i, particle=ion)
+    n_c = plasma_critical_density(omega_p)
+
+    assert n_c.unit.is_equivalent(u.m**-3)
+
+    assert np.isclose(n_c, 5e19 * u.m**-3)
```
pheuer, line 21: Note that this could also be written: 

```python
assert np.isclose(n_c.to(u.m**-3).value, 5e19)
```
which might be slightly better since it doesn't depend on how `np.isclose` decides to handle comparing `u. Quantity` objects?

docs/formulary/densities.rst
```diff
@@ -0,0 +1,9 @@
+.. _densities:
+
+******************************************************************
+Density related functionality (`plasmapy.formulary.densities`)
```
pheuer, line 4: To match the others, I think this should say 
"Density Plasma Parameters" 

e.g. look at frequencies.rst

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,63 @@
...
+    `ValueError`
+        If ``omega_p`` is an invalid value.
+
+    Warns
+    -----
+    : `~astropy.units.UnitsWarning`
+        If units are not provided, SI units are assumed.
+
+    Notes
+    -----
```
pheuer, line 52: You don't need to include empty section headers like this

However, it would actually be good to include the equation for the critical density here! Look at the `plasma_frequency` function for an example of how to format this. You can use LaTeX character codes. 

```
    .. math::
        ω_{p} = Z |e| \sqrt{\frac{n}{\epsilon_0 m}}
```

the notes section could also include a brief (few sentence) explanation of what the critical density is.

plasmapy/formulary/frequencies.py
JaydenR2305: astropy doesn't seem to like that
`astropy.units.core.UnitConversionError: 'F kg rad2 / (C2 m s2)' and '1 / m3' (number density) are not convertible`

plasmapy/formulary/tests/test_densities.py
JaydenR2305: I'll probably get rid of them

plasmapy/formulary/frequencies.py
pheuer: Oh, the problem is the units of radians (since rad isn't a 'real' unit, [it's dimensionless](https://math.stackexchange.com/questions/803955/why-radian-is-dimensionless)). I wish astropy handled this better - it clearly knows they are equivalent because the previous way you had this worked. You could do: 

```python
return (n_c/u.rad**2).to(u.m**-3)
```

Or I guess do it the way you were doing it before, since this isn't great

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,56 @@
...
+        n_{c}=\frac{m_{e}\varepsilon_0\omega_{p}^{2}}{e^{2}}
+
+    where :math:`m_{e}` is the mass of an electron,
+    :math:`\varepsilon_0` is the permittivity of free space, :math:`\omega_{p}`
+    is the plasma frequency, and :math:`e` is the elementary charge.
+
+    Examples
+    --------
+    >>> from astropy import units as u
+    >>> plasma_critical_density(1 * u.rad/u.s)
```
pheuer, line 46: Not that it really matters, but 1 rad/s is a very, very, very low frequency that isn't really relevant in plasma physics. Maybe replace it with ~ 5e15 rad/s, which is about the frequency of the OMEGA laser beams (just as a more relevant number).

plasmapy/formulary/frequencies.py
JaydenR2305: That's definitely a lot cleaner than what I was doing. I'll give that a try

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,56 @@
...
+        The plasma frequency in units of angular frequency.
+
+    Returns
+    -------
+    n_c : `~astropy.units.Quantity`
+        The plasma critical density.
+
+    Notes
+    -----
+    The plasma critical density is given by the formula
```
pheuer, line 34: How about adding that sentence of explanation you added to the test here to explain what this is physically?

"The plasma critical density is defined as the value at which the plasma frequency
        equals the frequency of radiation."

Although I would maybe revise slightly (in both places) 


"The plasma critical density _for a given frequency of radiation_ is defined as the value at which the _electron_ plasma frequency
        equals the frequency of _the_ radiation."

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,56 @@
...
+        ],
+    },
+)
+def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
+    r"""Calculate the plasma critical density.
+
+    Parameters
+    ----------
+    omega_p : `~astropy.units.Quantity`
+        The plasma frequency in units of angular frequency.
```
pheuer, line 25: Actually, this should be the radiation frequency. They are equal at the critical density, but conceptually usually you know the radiation frequency and then what to find out what density it will be absorbed at. I would go through and change this and the variable name throughout. Just 'omega' is fine.

plasmapy/formulary/densities.py
```diff
@@ -0,0 +1,56 @@
...
+    omega_p={"can_be_negative": False},
+    validations_on_return={
+        "units": [u.m**-3],
+        "equivalencies": [
+            (u.F * u.kg * u.rad**2 / (u.C**2 * u.m * u.s**2), u.m**-3)
+        ],
+    },
+)
+def plasma_critical_density(omega_p: u.rad / u.s) -> u.m**-3:
+    r"""Calculate the plasma critical density.
```
pheuer, line 20: "...for a radiation of a given frequency"

pheuer (changes requested):
Very nice! Just a couple more minor comments and I think this is ready to merge.

plasmapy/formulary/frequencies.py
namurphy: Another possibility would be to use the [`dimensionless_angles`](https://docs.plasmapy.org/en/latest/contributing/coding_guide.html#angular-frequencies) equivalency here.

JaydenR2305: Yeah let me give that a try, I thought I saw something about that somewhere

Thanks for the suggestion!

pheuer (changes requested):
Thanks @JaydenR2305 ! This is really close - just a couple more notes and I think it's ready to go.

pheuer (approved):
Looks good to me: thanks for putting this together!

I've approved the changes - you can now squash and merge your commits onto the main branch below.

JaydenR2305:
> Looks good to me: thanks for putting this together!
> 
> I've approved the changes - you can now squash and merge your commits onto the main branch below.

Hey Peter, I need someone with write access to merge my pull request. Thank you for your patience throughout this process!

pheuer:
@JaydenR2305 oops, yes I guess so. Ok, merged!

pheuer:
And no problem at all - this was an excellent first contribution and I'm happy to help you with any more you wish to work on in the future!

plasmapy/formulary/tests/test_densities.py
```diff
@@ -0,0 +1,35 @@
...
+import astropy.units as u
+import numpy as np
+import pytest
+
+from plasmapy.formulary.densities import critical_density
+from plasmapy.formulary.frequencies import plasma_frequency
+
+
+class TestCriticalDensity:
+    """Test the plasma_critical_density function in densities.py."""
```
JaydenR2305, line 11: _missed a spot_

PR #1667, opened 2022-08-11
Update guidelines for "Parameters", "Raises", and "Warns" sections of docstrings

namurphy:
This PR updates the documentation guide by saying that the "Raises" section of a docstring should only include the exceptions raised by the function itself, and not exceptions that are raised by decorators.  The motivations are:

 - Including the exceptions raised by decorators leads to a lot of repeated text, which should instead be uniquely defined in the docstring of the decorator.  
 - The "Raises" section is one of the least useful sections of a docstring, so it should be correspondingly short.
 - The "Raises" section of docstrings is one of the most likely to become outdated, at least in my experience.
 - The numpydoc guidelines themselves say that this section should be used judiciously.

### DIFF
--- a/docs/contributing/doc_guide.rst
+++ b/docs/contributing/doc_guide.rst
@@ -946,12 +946,16 @@ Docstring guidelines
   derivations and extensive discussions of mathematics in the "Notes"
   section.
 
-* Docstrings may include a Raises_ section that describes which
+* Docstrings may include a "Raises_" section that describes which
   exceptions get raised and under what conditions. This section should
   be used only for errors that are non-obvious or have a large chance of
   getting raised. Documenting exceptions makes it easier for future
-  developers to plan exception handling. However, in practice this
-  section often becomes outdated.
+  developers to plan exception handling.
+
+  * The "Raises_" section of a docstring should include exceptions that
+    are raised by the function itself, but not exceptions that are
+    raised by decorators such as |validate_quantities| and
+    |particle_input|.
 
 * Private code objects (e.g., code objects that begin with a single
   underscore, like ``_private_object``) should have docstrings. A
--- a/docs/contributing/doc_guide.rst
+++ b/docs/contributing/doc_guide.rst
@@ -970,8 +970,8 @@ Docstring guidelines
 
   * If the shapes and sizes of the parameters are interrelated, then
     include that information in parentheses immediately before the type
-    information. A trailing comma should be included inside the
-    parentheses when the parameter is 1D::
+    information. Include a trailing comma inside the parentheses when
+    the parameter is 1D.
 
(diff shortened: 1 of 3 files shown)


pheuer:
Question: should the docstring then contain some information about what validation is applied to the arguments? How does someone know the input is being validated without looking at the code?

pheuer:
Also, can you clarify whether tests for functions should include testing validate quantities? Or do we assume the validator works because it its being tested itself elsewhere.

pheuer:
Note that some of existing documentation does seem to contradict this: see `plasma_frequency`

namurphy:
> Question: should the docstring then contain some information about what validation is applied to the arguments? How does someone know the input is being validated without looking at the code?

Good point!  I updated the discussion of the parameters section to state that any preconditions should be specified there.

> Also, can you clarify whether tests for functions should include testing validate quantities? Or do we assume the validator works because it its being tested itself elsewhere.

It's still usually a good idea to test function behavior that is controlled by `validate_quantities` and `particle_input`, in particular for errors that are more likely for users to encounter.  These tests would not need to be as thorough, but they would be helpful if we ended up doing some refactoring of the function and/or decorators.  We could potentially add this to the testing guide in a separate PR.

> Note that some of existing documentation does seem to contradict this: see `plasma_frequency`

Yep!  I was planning on raising an issue afterwards about changing the Raises and Warns sections in docstrings in the formulary and elsewhere.  That's going to be a fair bit of work, but might be appropriate for a good first contribution.

docs/contributing/doc_guide.rst
```diff
@@ -946,12 +946,67 @@ Docstring guidelines
...
+
+       b : (N,) array_like
+           Second input vector.
+
+       out : (M, N) ndarray, optional
+           A location where the result is stored.
+
+    Use an asterisk (``*``) for a dimension that can be of arbitrary
+    size, and an ellipsis (``...``) to represent an arbitrary number of
+    dimensions that can each be of arbitrary size.
```
namurphy, line 991: This is adapted from https://github.com/numpy/numpydoc/pull/423, which includes a discussion of the motivation behind this syntax.  It's also done this way for a lot of the functions in NumPy (like [`numpy.outer`](https://numpy.org/doc/stable/reference/generated/numpy.outer)).  That PR doesn't include the `*` and `...` parts yet though.

namurphy: Also this is more consistent with the [matplotlib doc guide](https://matplotlib.org/stable/devel/documenting_mpl.html#parameter-type-descriptions).

StanczakDominik (approved):
Looks great! Codifying the shape output is a great idea, I'm really glad the numpydoc folks trailblazed that for us!

PR #1670, opened 2022-08-15
Add pygrep-hooks to pre-commit

namurphy:
This PR adds two pre-commit pygrep-hooks for reStructuredText files.  The two checks are:
 - Making sure that reST directives end with two colons
 - Checking for when inline code is next to normal text

I think errors like these would normally lead to doc build failures, but this will catch the issues sooner without having to wait for the doc build to complete.


StanczakDominik (approved):
👍🏻

Issue #1672, opened 2022-08-17
Make `Lorentz_factor` accept & return NaN values


namurphy:
Right now, `plasmapy.formulary.relativity.Lorentz_factor` does not accept a velocity with a magnitude of [`np.nan`](https://numpy.org/doc/stable/reference/constants.html#numpy.nan) (i.e. if we set `V = np.nan * u.m / u.s` or have `nan` in a `Quantity` array with the appropriate units).  The desired behavior would be for `Lorentz_factor` to accept `nan * u.m / u.s` value and then return `nan`, including when `nan` shows up in an array.  The reason is that we sometimes might want to perform a calculation on an array with bad data or with some elements undefined.

StanczakDominik:
For bonus points, ensure that the following work (in other words, that Lorentz_factor behaves like a numpy ufunc):
```pycon
>>> Lorentz_factor(1 * u.m/u.s)
... (numerical)
>>> Lorentz_factor(np.nan * u.m/u.s)
np.nan
>>> Lorentz_factor(1)
... (numerical with warning that m/s is assumed)
>>> Lorentz_factor(np.array([np.nan, np.nan]) * u.m/u.s)
np.array([np.nan, np.nan])
>>> Lorentz_factor(np.array([1, np.nan]) * u.m/u.s)
np.array([... (same as before), np.nan])
```

PR #1673, opened 2022-08-17
Add coding guideline that functions should accept and return `numpy.nan` values

namurphy:
Following up on #1672, this PR adds a guideline that says that functions should accept and return `nan` (not a number) values.  The reason is that sometimes we will work with arrays that have bad data that are flagged as `nan`, and still need to perform calculations on them.

### DIFF
--- a/docs/contributing/coding_guide.rst
+++ b/docs/contributing/coding_guide.rst
@@ -119,6 +119,17 @@ Coding guidelines
   >>> print(f"{package_name!r}")  # shortcut for f"{repr(package_name)}"
   'PlasmaPy'
 
+* Functions that accept |array_like| or |Quantity| inputs should accept
+  and return |nan| values. This guideline applies when |nan| is the
+  input as well as when |nan| is included in an array.
+
+  .. tip::
+
+     Normally, ``numpy.nan == numpy.nan`` evaluates to `False`, which
+     complicates testing |nan| behavior. Functions such as
+     `numpy.allclose` often have an ``equal_nan`` keyword such that
+     ``allclose(nan, nan, equal_nan=True)`` will return `True` instead.
+
 * Do not use :term:`mutable` objects as default values in the function
   or method declaration. This can lead to unexpected behavior.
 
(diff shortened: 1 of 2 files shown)


StanczakDominik:
There's an IEEE standard for dealing with nan/inf that should be attached here. I'm out walking the dog ATM, but it shouldn't be difficult to find.

namurphy:
> There's an IEEE standard for dealing with nan/inf that should be attached here. I'm out walking the dog ATM, but it shouldn't be difficult to find.

I think that's [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754). I was originally thinking of linking to this, but thinking about it always makes my brain start to melt.

namurphy:
I decided to link to the [Wikipedia article on NaNs](https://en.wikipedia.org/wiki/NaN) since it appears that the [IEEE 754](https://standards.ieee.org/ieee/754/6210/) standard is not openly available.  The NaN article seems to be the one that's most immediately useful, and that links to IEEE 754 for anyone who is really interested.

docs/contributing/coding_guide.rst
```diff
@@ -119,6 +119,18 @@ Coding guidelines
...
+* Functions that accept |array_like| or |Quantity| inputs should accept
+  and return |nan| (`not a number`_) values. This guideline applies when
+  |nan| is the input as well as when |nan| values are included in an
+  array.
+
+  .. tip::
+
+     Normally, ``numpy.nan == numpy.nan`` evaluates to `False`, which
+     complicates testing |nan| behavior. Functions such as
+     `numpy.allclose` often have an ``equal_nan`` keyword such that
```
StanczakDominik, line 131: Once more for everyone in the back!

```
numpy.
testing.
assert_
allclose
```

StanczakDominik (approved):

docs/contributing/coding_guide.rst
namurphy: Good point! I mentioned that as well.

PR #1674, opened 2022-08-17
Add tip to changelog guide on updating broken reST links

namurphy:
When a code object is deleted, any reST links to the old namespace will break.  While reST links in most of the documentation should be updated, the namespace in old changelog entries should be preserved.  This PR states how to change a reST link into an inline literal in old changelog entries.

An alternative would be to change all reST links in changelog entries into inline literals, which is what some other packages do (like Astropy, I think).  This would be less maintenance, but would make it more difficult for someone reading the changelog entry to access newly created features.

### DIFF
--- a/changelog/1674.doc.rst
+++ b/changelog/1674.doc.rst
@@ -1,0 +1,3 @@
+Added an admonition to the |changelog guide| that describes how to
+change reST_ links for removed code objects into inline literals in old
+changelog entries.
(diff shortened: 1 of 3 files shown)


StanczakDominik (approved):
That one looks familiar! ;)

It really does feel like our docs are growing to encompass the entire experience of being a package developer. Which, of course, is worthwhile for future generations, but I don't want to overdo it.

But sure, this one's :+1:

Issue #1668, opened 2022-08-12
Implement an autoformatter for docstrings


namurphy:
We use the numpydoc standard for our docstrings, there is still some inconsistency in the formatting (e.g., whether or not there's a blank line after a section header or if there are blank lines between parameters).  Consistent formatting makes documentation easier to read or skim, but it would be very tedious and error-prone to consistentify the formatting manually.  It would be helpful to apply an autoformatter for our docstrings, in particular as a pre-commit hook.  

I tried a few docstring autoformatters a few months ago but had some issues getting them to work, like if the one-liner was more than one line, or if numpydoc wasn't supported.  I'm not sure if there's a docstring autoformatter that fits our needs yet and is well-maintained.

namurphy:
Some of the ones I've looked into before but haven't tried recently:
 - [vélin](https://github.com/Carreau/velin).
 - [docformatter](https://github.com/PyCQA/docformatter) (which I think had limited numpydoc support?)
 - [darglint](https://github.com/terrencepreilly/darglint) (which I think is mostly a linter?)

Issue #1458, opened 2022-02-28
Intermittent test failure in `TestSyntheticRadiograph.test_optical_density_histogram`


namurphy:
In #1456, I got a test failure for the Python 3.8 (Windows) check for `TestSyntheticRadiograph.test_optical_density_histogram`.  The test failure only showed up in one check, and did not show up in any of the other checks.  

I'm wondering if specifying a random seed for this test would enable us to resolve the problem, and also make the test deterministic.

Here's the test failure message:

```
==================================================================================== FAILURES =====================================================================================
_____________________________________________________________ TestSyntheticRadiograph.test_optical_density_histogram ______________________________________________________________
[gw0] win32 -- Python 3.8.10 D:\a\PlasmaPy\PlasmaPy\.tox\py38-all\Scripts\python.EXE

self = <plasmapy.diagnostics.tests.test_charged_particle_radiography.TestSyntheticRadiograph object at 0x0000023CE26F7AF0>

    def test_optical_density_histogram(self):
        """
        Test the optical density calculation is correct and stuffed
        with numpy.inf when the intensity is zero.
        """
        bins = (200, 60)
        size = np.array([[-1, 1], [-1, 1]]) * 30 * u.cm
    
        intensity_results = cpr.synthetic_radiograph(
            self.sim_results, size=size, bins=bins
        )
        od_results = cpr.synthetic_radiograph(
            self.sim_results, size=size, bins=bins, optical_density=True
        )
    
        assert np.allclose(intensity_results[0], od_results[0])
        assert np.allclose(intensity_results[1], od_results[1])
    
        intensity = intensity_results[2]
        zero_mask = intensity == 0
        i0 = np.mean(intensity[~zero_mask])
        od = -np.log10(intensity / i0)
    
>       assert np.allclose(od[~zero_mask], od_results[2][~zero_mask])
E       assert False
[...]

PR #1676, opened 2022-08-18
Add `CollisionFrequencies`

JaydenR2305:
Resolves PlasmaPy/hack-week#23

This pull request implements the four types of relaxation described on page 31 of the NRL formulary (see issue for screenshot)

For now, the class relies on a parameter called `coulomb_log`, which is the value of the Coulomb logarithm evaluated for the two interacting plasmas. This is because the current implementation of the Coulomb logarithm is limited to electron-electron interactions... In a future pull request I will be revisiting the Coulomb logarithm to include for a broader array of interactions

      docstrings.

### DIFF
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2145,6 +2145,11 @@ class CollisionFrequencies:
         self.test_particle = test_particle
         self.field_particle = field_particle
         self.v_a = v_a
+        self.T_a = (
+            T_a
+            if T_a is not None
+            else (0.5 * test_particle.mass.to(u.g) * v_a**2 / k_B).to(u.K)
+        )
         self.n_b = n_b
         self.T_b = T_b
         self.Coulomb_log = Coulomb_log
@@ -2155,13 +2160,29 @@ class CollisionFrequencies:
 
         mass_ratio = test_particle.mass / field_particle.mass
 
-        self.v_0 = self.v_0()
+        self.v_0 = self.Lorentz_collision_frequency()
         self.slowing_down = (1 + mass_ratio) * phi * self.v_0
         self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * self.v_0
         self.parallel_diffusion = (phi / x) * self.v_0
         self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self.v_0
 
-    def v_0(self):
+    def Lorentz_collision_frequency(self):
+        r"""
+        The Lorentz collision frequency (see Ch. 5 of :cite:t:`chen:2016`) is given
+        by
+
+        .. math::
+
+            ν = n σ v \ln{Λ}
+
+        where :math:`n` is the particle density, :math:`σ` is the
+        collisional cross-section, :math:`v` is the inter-particle velocity
+        (typically taken as the thermal velocity), and :math:`\ln{Λ}` is the
+        Coulomb logarithm accounting for small angle collisions.
+
+        See Equation (2.14) in :cite:t:`callen:unpublished`.
+        """
+
         return (
             4
             * np.pi
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2173,21 +2173,22 @@ class CollisionFrequencies:
         self.T_b = T_b
         self.Coulomb_log = Coulomb_log
 
-        # These attributes are used in testing
-        self._x = self.x
-        self._v_0 = self.Lorentz_collision_frequency
-
-        phi = self._phi(self.x)
-        phi_prime = self._phi_prime(self.x)
-
         mass_ratio = test_particle.mass / field_particle.mass
 
-        self.momentum_loss = (1 + mass_ratio) * phi * self._v_0
-        self.transverse_diffusion = (
-            2 * ((1 - 1 / (2 * self.x)) * phi + phi_prime) * self._v_0
+        self.momentum_loss = (
+            (1 + mass_ratio) * self.phi * self.Lorentz_collision_frequency
         )
-        self.parallel_diffusion = (phi / self.x) * self._v_0
-        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self._v_0
+        self.transverse_diffusion = (
+            2
+            * ((1 - 1 / (2 * self.x)) * self.phi + self._phi_prime)
+            * self.Lorentz_collision_frequency
+        )
+        self.parallel_diffusion = (self.phi / self.x) * self.Lorentz_collision_frequency
+        self.energy_loss = (
+            2
+            * (mass_ratio * self.phi - self._phi_prime)
+            * self.Lorentz_collision_frequency
+        )
 
     @property
     def Lorentz_collision_frequency(self):
@@ -2247,18 +2248,22 @@ class CollisionFrequencies:
 
         return integral
 
-    def _phi(self, x: u.dimensionless_unscaled):
+    @property
+    def phi(self):
         """
         The parameter phi used in calculating collision frequencies
-        For more information refer to page 19 of the NRL Formulary
+        calculated using the default error tolerances of`~scipy.integrate.quad`
+
+        For more information refer to page 31 of the NRL Formulary.
         """
         vectorized_integral = np.vectorize(self._phi_explicit)
 
-        return 2 / np.pi**0.5 * vectorized_integral(x.value)
+        return 2 / np.pi**0.5 * vectorized_integral(self.x.value)
 
-    def _phi_prime(self, x: u.dimensionless_unscaled):
+    @property
+    def _phi_prime(self):
         """
         The derivative of phi evaluated at x
         """
 
-        return 2 / np.pi**0.5 * self._phi_integrand(x)
+        return 2 / np.pi**0.5 * self._phi_integrand(self.x)
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2055,21 +2055,21 @@ class CollisionFrequencies:
         test_particle: particles.Particle,
         field_particle: particles.Particle,
         *,
-        v_a: u.cm / u.s = None,
-        T_a: u.K = None,
+        v_a: u.cm / u.s = np.nan,
+        T_a: u.K = np.nan,
         n_b: u.cm**-3,
         T_b: u.K,
         coulomb_log: u.dimensionless_unscaled,
         dx: float = 1e-20,
     ):
-        if v_a is None:
-            if T_a is None:
-                raise ValueError("Please specify either v_a or T_a.")
 
-            v_a = thermal_speed(T_a, test_particle)
-        else:
-            if T_a is not None:
-                raise ValueError("Please specify either v_a or T_a, not both.")
+        if np.any(np.logical_and(np.isnan(v_a), np.isnan(T_a))):
+            raise ValueError("Please specify either v_a or T_a.")
+
+        if np.any(np.logical_and(np.isfinite(v_a), np.isfinite(T_a))):
+            raise ValueError("Please specify either v_a or T_a, not both.")
+
+        v_a = _replace_nan_velocity_with_thermal_velocity(v_a, T_a, test_particle.mass)
 
         self.test_particle = test_particle
         self.field_particle = field_particle
@@ -2110,9 +2110,15 @@ class CollisionFrequencies:
     def _phi_integrand(t: u.dimensionless_unscaled):
         return t**0.5 * math.exp(-t)
 
-    def _phi(self, x: u.dimensionless_unscaled):
+    def _phi_explicit(self, x: float) -> float:
         integral, _ = scipy.integrate.quad(self._phi_integrand, 0, x)
-        return 2 / math.pi**0.5 * integral
+
+        return integral
+
+    def _phi(self, x: u.dimensionless_unscaled):
+        vectorized_integral = np.vectorize(self._phi_explicit)
+
+        return 2 / math.pi**0.5 * vectorized_integral(x.value)
 
     def _phi_prime(self, x: u.dimensionless_unscaled):
         return scipy.misc.derivative(self._phi, x, dx=self.dx)
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2039,10 +2039,10 @@ def coupling_parameter(
 
 
 class RelaxationRates:
-	@validate_quantities(
-	    n_b={"can_be_negative": False},
-	    T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
-	)
+    @validate_quantities(
+        n_b={"can_be_negative": False},
+        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+    )
     def __init__(
         self,
         species: (particles.Particle, particles.Particle),
@@ -2050,12 +2050,14 @@ class RelaxationRates:
         n_b: u.m**-3,
         T_b: u.K,
         coulomb_log: u.dimensionless_unscaled,
+        dx: float = 1e-20,
     ):
         self.species = species
         self.v_a = v_a
         self.n_b = n_b
         self.T_b = T_b
         self.coulomb_log = coulomb_log
+        self.dx = dx
 
         v_0 = self._v_0()
         x = self._x()
@@ -2093,4 +2095,4 @@ class RelaxationRates:
         return 2 / math.pi**0.5 * integral
 
     def _phi_prime(self, x: u.dimensionless_unscaled):
-        return scipy.misc.derivative(self._phi, x)
+        return scipy.misc.derivative(self._phi, x, dx=self.dx)
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -59,6 +59,7 @@ import numpy as np
 import warnings
 
 from astropy.constants.si import e, eps0, hbar, k_B, m_e
+import math
 from numbers import Real
 from numpy import pi
 
@@ -2034,6 +2035,7 @@ def coupling_parameter(
 
     return coulomb_energy / kinetic_energy
 
+
 @validate_quantities(
     n_b={"can_be_negative": False},
     T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
@@ -2061,9 +2063,7 @@ class RelaxationRates:
         mu = species[0].mass / species[1].mass
 
         self.slowing_down = (1 + mu) * phi * v_0
-        self.transverse_diffusion = (
-            2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
-        )
+        self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
         self.parallel_diffusion = (phi / x) * v_0
         self.energy_loss = 2 * (mu * phi - phi_prime) * v_0
 
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2222,14 +2222,14 @@ class CollisionFrequencies:
     @cached_property
     def momentum_loss(self):
         """
-        The momentum loss for the specified particles.
+        The momentum loss rate due to collisions.
         """
         return (1 + self._mass_ratio) * self.phi * self.Lorentz_collision_frequency
 
     @cached_property
     def transverse_diffusion(self):
         """
-        The transverse diffusion for the specified particles.
+        The rate of transverse diffusion due to collisions.
         """
         return (
             2
@@ -2240,14 +2240,14 @@ class CollisionFrequencies:
     @cached_property
     def parallel_diffusion(self):
         """
-        The parallel diffusion for the specified particles.
+        The rate of parallel diffusion due to collisions.
         """
         return (self.phi / self.x) * self.Lorentz_collision_frequency
 
     @cached_property
     def energy_loss(self):
         """
-        The energy loss for the specified particles.
+        The energy loss rate due to collisions.
         """
         return (
             2
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2111,6 +2111,8 @@ class CollisionFrequencies:
 
         For general solutions and limiting cases we encourage the curious reader to refer to p. 31 of :cite:t:`nrlformulary:2019`
 
+        This function uses CGS units under the hood to coincide with our references, but passing MKS units is acceptable.
+
         Examples
         --------
         >>> import astropy.units as u
@@ -2123,11 +2125,18 @@ class CollisionFrequencies:
         ... )
         >>> frequencies.slowing_down
         <Quantity 1.83701432e+11 Hz>
+
+        See Also
+        --------
+        ~plasmapy.formulary.collisions.Coulomb_log : Evaluates the Coulomb
+            logarithm for two interacting electron species.
+
+        ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
         """
 
         if v_a is None:
             if T_a is not None:
-                v_a = thermal_speed(T_a, test_particle).to(u.cm / u.s)
+                v_a = thermal_speed(T_a, test_particle)  # .to(u.cm / u.s)
             else:
                 raise ValueError("Please specify either v_a or T_a.")
         elif T_a is not None:
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2064,7 +2064,7 @@ class CollisionFrequencies:
         Coulomb_log: u.dimensionless_unscaled,
     ):
         r"""
-        Compute collision frequencies
+        Compute collision frequencies between test particles (labeled 'a') and field particles (labeled 'b')
 
         Parameters
         ----------
@@ -2075,7 +2075,7 @@ class CollisionFrequencies:
             The background particle being interacted with.
 
         v_a : `~astropy.units.Quantity`, optional
-            The relative velocity between particles. If not provided,
+            The relative velocity between particles. If not provided, T_a must be specified and
             thermal velocity is assumed: :math:`μ v_a^2 \sim 2 k_B T_a` where
             :math:`μ` is the reduced mass.
 
@@ -2128,7 +2128,7 @@ class CollisionFrequencies:
 
         See Also
         --------
-        ~plasmapy.formulary.collisions.Coulomb_log : Evaluates the Coulomb
+        ~plasmapy.formulary.collisions.Coulomb_logarithm : Evaluates the Coulomb
             logarithm for two interacting electron species.
 
         ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
@@ -2136,7 +2136,7 @@ class CollisionFrequencies:
 
         if v_a is None:
             if T_a is not None:
-                v_a = thermal_speed(T_a, test_particle)  # .to(u.cm / u.s)
+                v_a = thermal_speed(T_a, test_particle).to(u.cm / u.s)
             else:
                 raise ValueError("Please specify either v_a or T_a.")
         elif T_a is not None:
@@ -2155,13 +2155,13 @@ class CollisionFrequencies:
 
         mass_ratio = test_particle.mass / field_particle.mass
 
-        self.v_0 = self._v_0()
+        self.v_0 = self.v_0()
         self.slowing_down = (1 + mass_ratio) * phi * self.v_0
         self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * self.v_0
         self.parallel_diffusion = (phi / x) * self.v_0
         self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self.v_0
 
-    def _v_0(self):
+    def v_0(self):
         return (
             4
             * np.pi
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2317,7 +2317,7 @@ class CollisionFrequencies:
     def phi(self):
         """
         The parameter phi used in calculating collision frequencies
-        calculated using the default error tolerances of `~scipy.integrate.quad`
+        calculated using the default error tolerances of `~scipy.integrate.quad`.
 
         For more information refer to page 31 of the NRL Formulary.
         """
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2033,3 +2033,62 @@ def coupling_parameter(
         )
 
     return coulomb_energy / kinetic_energy
+
+@validate_quantities(
+    n_b={"can_be_negative": False},
+    T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+)
+class RelaxationRates:
+    def __init__(
+        self,
+        species: (particles.Particle, particles.Particle),
+        v_a: u.m / u.s,
+        n_b: u.m**-3,
+        T_b: u.K,
+        coulomb_log: u.dimensionless_unscaled,
+    ):
+        self.species = species
+        self.v_a = v_a
+        self.n_b = n_b
+        self.T_b = T_b
+        self.coulomb_log = coulomb_log
+
+        v_0 = self._v_0()
+        x = self._x()
+        phi = self._phi(x)
+        phi_prime = self._phi_prime(x)
+
+        mu = species[0].mass / species[1].mass
+
+        self.slowing_down = (1 + mu) * phi * v_0
+        self.transverse_diffusion = (
+            2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
+        )
+        self.parallel_diffusion = (phi / x) * v_0
+        self.energy_loss = 2 * (mu * phi - phi_prime) * v_0
+
+    def _v_0(self):
+        return (
+            4
+            * math.pi
+            * self.species[0].charge ** 2
+            * self.species[1].charge ** 2
+            * self.coulomb_log
+            * self.n_b
+            / (self.species[0].mass ** 2 * self.v_a**3)
+        )
+
+    def _x(self) -> u.dimensionless_unscaled:
+        x = self.species[1].mass * self.v_a**2 / (2 * k_B * self.T_b)
+        return x.to(u.dimensionless_unscaled)
+
+    @staticmethod
+    def _phi_integrand(t: u.dimensionless_unscaled):
+        return t**0.5 * math.exp(-t)
+
+    def _phi(self, x: u.dimensionless_unscaled):
+        integral, _ = scipy.integrate.quad(self._phi_integrand, 0, x)
+        return 2 / math.pi**0.5 * integral
+
+    def _phi_prime(self, x: u.dimensionless_unscaled):
+        return scipy.misc.derivative(self._phi, x)
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2150,6 +2150,9 @@ class CollisionFrequencies:
             logarithm for two interacting electron species.
         """
 
+        # Note: This function uses CGS units internally to coincide with our references.
+        # Input is taken in MKS units and then converted as necessary. Output is in MKS units.
+
         if v_a is None:
             if T_a is not None:
                 v_a = thermal_speed(T_a, test_particle).to(u.cm / u.s)
@@ -2170,17 +2173,21 @@ class CollisionFrequencies:
         self.T_b = T_b
         self.Coulomb_log = Coulomb_log
 
-        x = self._x()
-        v_0 = self.Lorentz_collision_frequency
-        phi = self._phi(x)
-        phi_prime = self._phi_prime(x)
+        # These attributes are used in testing
+        self._x = self.x
+        self._v_0 = self.Lorentz_collision_frequency
+
+        phi = self._phi(self.x)
+        phi_prime = self._phi_prime(self.x)
 
         mass_ratio = test_particle.mass / field_particle.mass
 
-        self.momentum_loss = (1 + mass_ratio) * phi * v_0
-        self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
-        self.parallel_diffusion = (phi / x) * v_0
-        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * v_0
+        self.momentum_loss = (1 + mass_ratio) * phi * self._v_0
+        self.transverse_diffusion = (
+            2 * ((1 - 1 / (2 * self.x)) * phi + phi_prime) * self._v_0
+        )
+        self.parallel_diffusion = (phi / self.x) * self._v_0
+        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self._v_0
 
     @property
     def Lorentz_collision_frequency(self):
@@ -2210,26 +2217,48 @@ class CollisionFrequencies:
             * (self.field_particle.charge_number * e.esu) ** 2
             * self.Coulomb_log
             * self.n_b
-            / (self.test_particle.mass.to(u.g) ** 2 * self.v_a**3)
+            / (self.test_particle.mass**2 * self.v_a**3)
         ).to(u.Hz)
 
-    def _x(self) -> u.dimensionless_unscaled:
-        x = self.field_particle.mass.to(u.g) * self.v_a**2 / (2 * k_B.cgs * self.T_b)
+    @property
+    def x(self) -> u.dimensionless_unscaled:
+        """
+        The ratio of kinetic energy in the test particle to the thermal energy of the field particle.
+        This parameter determines the regime in which the collision falls (fast or slow).
+
+        """
+
+        x = self.field_particle.mass * self.v_a**2 / (2 * k_B.cgs * self.T_b)
         return x.to(u.dimensionless_unscaled)
 
     @staticmethod
     def _phi_integrand(t: u.dimensionless_unscaled):
+        """
+        The phi integrand used in calculating phi
+        """
+
         return t**0.5 * np.exp(-t)
 
     def _phi_explicit(self, x: float) -> float:
+        """
+        The non-vectorized method for evaluating the integral for phi
+        """
         integral, _ = scipy.integrate.quad(self._phi_integrand, 0, x)
 
         return integral
 
     def _phi(self, x: u.dimensionless_unscaled):
+        """
+        The parameter phi used in calculating collision frequencies
+        For more information refer to page 19 of the NRL Formulary
+        """
         vectorized_integral = np.vectorize(self._phi_explicit)
 
         return 2 / np.pi**0.5 * vectorized_integral(x.value)
 
     def _phi_prime(self, x: u.dimensionless_unscaled):
+        """
+        The derivative of phi evaluated at x
+        """
+
         return 2 / np.pi**0.5 * self._phi_integrand(x)
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2045,14 +2045,16 @@ class RelaxationRates:
     )
     def __init__(
         self,
-        species: (particles.Particle, particles.Particle),
+        test_particle: particles.Particle,
+        field_particle: particles.Particle,
         v_a: u.cm / u.s,
         n_b: u.cm**-3,
         T_b: u.K,
         coulomb_log: u.dimensionless_unscaled,
         dx: float = 1e-20,
     ):
-        self.species = species
+        self.test_particle = test_particle
+        self.field_particle = field_particle
         self.v_a = v_a
         self.n_b = n_b
         self.T_b = T_b
@@ -2064,26 +2066,26 @@ class RelaxationRates:
         phi = self._phi(x)
         phi_prime = self._phi_prime(x)
 
-        mu = species[0].mass / species[1].mass
+        particle_mass_ratio = test_particle.mass / field_particle.mass
 
-        self.slowing_down = (1 + mu) * phi * v_0
+        self.slowing_down = (1 + particle_mass_ratio) * phi * v_0
         self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
         self.parallel_diffusion = (phi / x) * v_0
-        self.energy_loss = 2 * (mu * phi - phi_prime) * v_0
+        self.energy_loss = 2 * (particle_mass_ratio * phi - phi_prime) * v_0
 
     def _v_0(self):
         return (
             4
             * math.pi
-            * (self.species[0].charge_number * e.esu) ** 2
-            * (self.species[1].charge_number * e.esu) ** 2
+            * (self.test_particle.charge_number * e.esu) ** 2
+            * (self.field_particle.charge_number * e.esu) ** 2
             * self.coulomb_log
             * self.n_b
-            / (self.species[0].mass.to(u.g) ** 2 * self.v_a**3)
+            / (self.test_particle.mass.to(u.g) ** 2 * self.v_a**3)
         ).to(u.Hz)
 
     def _x(self) -> u.dimensionless_unscaled:
-        x = self.species[1].mass.to(u.g) * self.v_a**2 / (2 * k_B.cgs * self.T_b)
+        x = self.field_particle.mass.to(u.g) * self.v_a**2 / (2 * k_B.cgs * self.T_b)
         return x.to(u.dimensionless_unscaled)
 
     @staticmethod
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2055,7 +2055,7 @@ def coupling_parameter(
 
 class CollisionFrequencies:
     @validate_quantities(
-        v_a={"none_shall_pass": True},
+        v_a={"none_shall_pass": True, "can_be_negative": False},
         T_a={
             "none_shall_pass": True,
             "can_be_negative": False,
@@ -2070,9 +2070,9 @@ class CollisionFrequencies:
         test_particle: particles.Particle,
         field_particle: particles.Particle,
         *,
-        v_a: u.cm / u.s = None,
+        v_a: u.m / u.s = None,
         T_a: u.K = None,
-        n_b: u.cm**-3,
+        n_b: u.m**-3,
         T_b: u.K,
         Coulomb_log: u.dimensionless_unscaled,
     ):
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -1002,6 +1002,11 @@ def collision_frequency(
     >>> species = ('e', 'p')
     >>> collision_frequency(T, n, species)
     <Quantity 70249... Hz>
+
+    See Also
+    --------
+    ~plasmapy.formulary.collisions.CollisionFrequencies.Lorentz_collision_frequency
+
     """
 
     deprecated(
@@ -2088,18 +2093,18 @@ class CollisionFrequencies:
             The background particle being interacted with.
 
         v_a : `~astropy.units.Quantity`, optional
-            The relative velocity between particles. If not provided, T_a must be specified and
+            The relative speed between particles. If not provided, T_a must be specified and
             thermal velocity is assumed: :math:`μ v_a^2 \sim 2 k_B T_a` where
-            :math:`μ` is the reduced mass.
+            :math:`μ` is the reduced mass. Cannot be negative.
 
         T_a : `~astropy.units.Quantity`, optional
             The temperature of the test species. Only necessary if :math:`v_a` is not provided.
 
         n_b : `~astropy.units.Quantity`
-            The number density of the background field particles.
+            The number density of the background field particles in units convertable to :math:`\frac{1}{m^{3}}`
 
         T_b : `~astropy.units.Quantity`
-            The temperature of the background field particles.
+            The temperature of the background field particles in units convertable to degrees Kelvin.
 
         Coulomb_log : `~astropy.units.Quantity`
             The value of the Coulomb logarithm evaluated for the two interacting particles.
@@ -2114,7 +2119,7 @@ class CollisionFrequencies:
         -----
         The collision frequencies between two interacting particles are given by four differential equations:
 
-            slowing down - :math:`\frac{d\textbf{v}_{α}}{dt}=-ν_{s}^{α\backslashβ}\textbf{v}_{α}`
+            momentum loss - :math:`\frac{d\textbf{v}_{α}}{dt}=-ν_{s}^{α\backslashβ}\textbf{v}_{α}`
 
             transverse diffusion - :math:`\frac{d}{dt}\left(\overline{\textbf{v}}_{α}-\overline{\textbf{v}}_{α}\right)_{⊥}^{2}=ν_{⊥}^{α\backslashβ}v_{α}^{2}`
 
@@ -2136,15 +2141,13 @@ class CollisionFrequencies:
         >>> frequencies = CollisionFrequencies(
         ...     "e-", "e-", T_a=T_a, n_b=n_b, T_b=T_b, Coulomb_log=Coulomb_log
         ... )
-        >>> frequencies.slowing_down
+        >>> frequencies.momentum_loss
         <Quantity 1.83701432e+11 Hz>
 
         See Also
         --------
         ~plasmapy.formulary.collisions.Coulomb_logarithm : Evaluates the Coulomb
             logarithm for two interacting electron species.
-
-        ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
         """
 
         if v_a is None:
@@ -2161,24 +2164,25 @@ class CollisionFrequencies:
         self.T_a = (
             T_a
             if T_a is not None
-            else (0.5 * test_particle.mass.to(u.g) * v_a**2 / k_B).to(u.K)
+            else (0.5 * test_particle.mass * v_a**2 / k_B).to(u.K)
         )
         self.n_b = n_b
         self.T_b = T_b
         self.Coulomb_log = Coulomb_log
 
         x = self._x()
+        v_0 = self.Lorentz_collision_frequency
         phi = self._phi(x)
         phi_prime = self._phi_prime(x)
 
         mass_ratio = test_particle.mass / field_particle.mass
 
-        self.v_0 = self.Lorentz_collision_frequency()
-        self.slowing_down = (1 + mass_ratio) * phi * self.v_0
-        self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * self.v_0
-        self.parallel_diffusion = (phi / x) * self.v_0
-        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self.v_0
+        self.momentum_loss = (1 + mass_ratio) * phi * v_0
+        self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * v_0
+        self.parallel_diffusion = (phi / x) * v_0
+        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * v_0
 
+    @property
     def Lorentz_collision_frequency(self):
         r"""
         The Lorentz collision frequency (see Ch. 5 of :cite:t:`chen:2016`) is given
@@ -2194,6 +2198,9 @@ class CollisionFrequencies:
         Coulomb logarithm accounting for small angle collisions.
 
         See Equation (2.14) in :cite:t:`callen:unpublished`.
+
+        The Lorentz collision frequency is equivalent to the variable
+        :math:`\nu_0^{\alpha/\beta}` on p. 31 of :cite:t:`nrlformulary:2019`.
         """
 
         return (
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -40,6 +40,7 @@ These include:
 * `~plasmapy.formulary.collisions.coupling_parameter`
 """
 __all__ = [
+    "CollisionFrequencies",
     "Coulomb_logarithm",
     "impact_parameter_perp",
     "impact_parameter",
@@ -52,7 +53,6 @@ __all__ = [
     "mobility",
     "Knudsen_number",
     "coupling_parameter",
-    "CollisionFrequencies",
 ]
 
 import astropy.units as u
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -559,7 +559,9 @@ def _process_inputs(T: u.K, species: (particles.Particle, particles.Particle), V
     return T, masses, charges, reduced_mass, V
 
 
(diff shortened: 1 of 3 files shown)


plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,61 @@ def coupling_parameter(
...
 
     return coulomb_energy / kinetic_energy
+
+
+@validate_quantities(
+    n_b={"can_be_negative": False},
+    T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+)
+class RelaxationRates:
+    def __init__(
```
StanczakDominik, line 2046: Whoa, if this actually works then @rocco8773 is an even greater wizard than I thought! AFAIK you have to decorate `__init__` instead of the main class declaration (move the decorator one line down 😉)
```suggestion
class RelaxationRates:
	@validate_quantities(
	    n_b={"can_be_negative": False},
	    T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
	)
    def __init__(
```

JaydenR2305: 🤦 This is what happens when I try to program too early in the morning

StanczakDominik: Hey, some of my best code was written before getting fully dressed in the morning! Some of my worst, too, as it happens. 😅

plasmapy/formulary/collisions.py
```diff
@@ -2075,15 +2075,15 @@ def _v_0(self):
         return (
             4
             * math.pi
-            * self.species[0].charge ** 2
-            * self.species[1].charge ** 2
+            * (self.species[0].charge_number * e.esu) ** 2
```
JaydenR2305, line 2078: Tips on how to make this less ugly would be appreciated. Shouldn't functions only take MKS units?

pheuer: @namurphy 

To elaborate, the problem here is that the NRL formulary (from which we are adapting these functions) is in CGS, but we want to take MKS input for PlasmaPy.

pheuer: @JaydenR2305 I don't think you want the charge number here, I think this should just be `self.species[0].charge.to(u.statC)`

Looks like u.esu is equivalent to u.statC

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+
+class RelaxationRates:
+    @validate_quantities(
+        n_b={"can_be_negative": False},
+        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+    )
+    def __init__(
+        self,
+        species: (particles.Particle, particles.Particle),
+        v_a: u.cm / u.s,
```
pheuer, line 2049: I think v_a should be optional, and if not provided should be calculated using `thermal_speed` from T_a

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+        n_b: u.cm**-3,
+        T_b: u.K,
+        coulomb_log: u.dimensionless_unscaled,
+        dx: float = 1e-20,
```
pheuer, line 2053: Make sure to explain this parameter carefully when doing the documentation. 

Also important to carefully explain the a and b notation.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+            * self.n_b
+            / (self.species[0].mass.to(u.g) ** 2 * self.v_a**3)
+        ).to(u.Hz)
+
+    def _x(self) -> u.dimensionless_unscaled:
+        x = self.species[1].mass.to(u.g) * self.v_a**2 / (2 * k_B.cgs * self.T_b)
+        return x.to(u.dimensionless_unscaled)
+
+    @staticmethod
+    def _phi_integrand(t: u.dimensionless_unscaled):
```
pheuer, line 2090: Probably not necessary to have this as a separate method, since you never use it other than within _phi

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+        return t**0.5 * math.exp(-t)
+
+    def _phi(self, x: u.dimensionless_unscaled):
```
pheuer, line 2093: Look into using cached_property so this is only calculated once (for example if you want to calculate multiple frequencies)

https://pypi.org/project/cached-property/

Python 3.9+ has these built in, but I think we still support 3.8 for now so we are using this package to mimic that functionality for now.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+
+
+class RelaxationRates:
+    @validate_quantities(
+        n_b={"can_be_negative": False},
+        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+    )
+    def __init__(
+        self,
+        species: (particles.Particle, particles.Particle),
```
pheuer, line 2048: Suggest having 'test_particle' and 'background_particle' or something as separate arguments for clarity, since the distinction between the a and b species is subtle but very important.

plasmapy/formulary/collisions.py
pheuer: I see you're effectively doing this already by just creating the self.phi variable, but leaving this comment just FYI

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
...
+        self.T_b = T_b
+        self.coulomb_log = coulomb_log
+        self.dx = dx
+
+        v_0 = self._v_0()
+        x = self._x()
+        phi = self._phi(x)
+        phi_prime = self._phi_prime(x)
+
+        mu = species[0].mass / species[1].mass
```
pheuer, line 2067: Confusing to call this mu, since mu is species[x].mass/ proton mass

plasmapy/formulary/collisions.py
JaydenR2305: Would this be one of the cases where using a lambda function is acceptable?

pheuer (commented):
When you add the documentation, add a citation of these pages in the NRL formulary. 

In the code, add comments wherever you convert to CGS noting that the NRL formulas are in cgs and so we convert to preserve those formulas.

plasmapy/formulary/collisions.py
pheuer: Oh I see, I misunderstood what you were doing. I usually calculate integrate like 

```python

t = np.linspace(0, x, num=d/dx)
integrand = _integrand(t)
phi = np.cumsum(integrand)
```

But the way you're doing is fine

plasmapy/formulary/collisions.py
JaydenR2305: Hmm, do you have any suggestions for a name? It's a quantity that pops up in a few of the expressions

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2036,63 @@ def coupling_parameter(
         )
 
     return coulomb_energy / kinetic_energy
+
+
+class RelaxationRates:
```
pheuer, line 2041: Personally I would change this to 'CollisionFrequencies': I hear that term more than relaxation rates.

JaydenR2305: Is calling them "relaxation parameters" in the changelog still acceptable in addition to referencing the function?

pheuer: I'd call them collision frequencies there too

plasmapy/formulary/collisions.py
pheuer: maybe mass_ratio

pheuer:
This function exists, so you should look at the code and make sure we are generalizing it. This PR could potentially replace it, if it can provide the same functionality but in a more general way

https://docs.plasmapy.org/en/stable/api/plasmapy.formulary.collisions.collision_frequency.html

JaydenR2305:
> This function exists, so you should look at the code and make sure we are generalizing it. This PR could potentially replace it, if it can provide the same functionality but in a more general way
> 
> https://docs.plasmapy.org/en/stable/api/plasmapy.formulary.collisions.collision_frequency.html

Okay so that function seems to use a different expression for finding the collision frequency. The only thing I think we could potentially transfer is a relativity warning for relativistic plasmas and maybe an argument to specify the method to calculate the Coulomb logarithm

pheuer:
>Okay so that function seems to use a different expression for finding the collision frequency. 

Slightly different in that it's less general (assumes ion-electron collisions I think). Look at the equation it references in Callan (Eq. 2.14): https://drive.google.com/file/d/1mSpES1BDTbrD0L124pwH5s0c7t41L6g5/view 

That's basically the expression for v0 but in MKS units I think, right? 

>The only thing I think we could potentially transfer is a relativity warning for relativistic plasmas and maybe an argument to specify the method to calculate the Coulomb logarithm

These are good ideas - the Coulomb log method can be added once we add that functionality. The relativistic warning is a good idea.

JaydenR2305:
> > Okay so that function seems to use a different expression for finding the collision frequency.
> 
> Slightly different in that it's less general (assumes ion-electron collisions I think). Look at the equation it references in Callan (Eq. 2.14): https://drive.google.com/file/d/1mSpES1BDTbrD0L124pwH5s0c7t41L6g5/view
> 
> That's basically the expression for v0 but in MKS units I think, right?
> 
> > The only thing I think we could potentially transfer is a relativity warning for relativistic plasmas and maybe an argument to specify the method to calculate the Coulomb logarithm
> 
> These are good ideas - the Coulomb log method can be added once we add that functionality. The relativistic warning is a good idea.

Oh, awesome! I concur- it is the same expression but in MKS units. I'll keep CGS version since it's cleaner without Coulomb's constant on the bottom

plasmapy/formulary/collisions.py
JaydenR2305: What about v_b?

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2039,91 @@ def coupling_parameter(
...
+            "equivalencies": u.temperature_energy(),
+        },
+        n_b={"can_be_negative": False},
+        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+    )
+    @particles.particle_input
+    def __init__(
+        self,
+        test_particle: particles.Particle,
+        field_particle: particles.Particle,
```
JaydenR2305, line 2059: Are `test_particle` and `field_particle` better left as individual parameters or should I do this as a tuple??

JaydenR2305: Actually, I could probably use the GenericPlasma class

pheuer: Hm, not sure if GenericPlasma is a good fit here, since both test and field particles could be ions or electrons? 

I like having them as separate parameters to make it really explicit which is which (since that's very important)

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,91 @@ def coupling_parameter(
...
+        *,
+        v_a: u.cm / u.s = np.nan * u.cm / u.s,
```
pheuer, line 2060: I think it would be easier to just set the default values for v_a and T_a to be None

JaydenR2305: Hmm, the only issue with that would the replace_nan_velocities... Unless isnan() returns true for NaN

pheuer: Is there a problem with just doing something like: 

```python
if v_a is None and T_a is not None:
    v_a = thermal_speed(T_a, particle_a)
elif v_a is None and T_a is None:
    raise ValueError("One of v_a or T_a must be provided")
else:
   # Test to make sure v_a is an acceptable type input
```

JaydenR2305: Yeah so currently the class is vectorized. In the case of a numpy array, if there is a mix of specified and unspecified v_a (unspecified v_a being represented by NaN) then v_a will be calculated from the corresponding T_a using the _replace_nan_velocities. By setting the default value to NaN, we mimic an entry for which the velocity is unspecified, and needs to be calculated using T_a

plasmapy/formulary/collisions.py
JaydenR2305: Yeah there seem to have been some efforts for some sort of species class awhile ago that would extend functionality of the particle class to include number density and temperature, though I don't think it was ever implemented.

That would probably be our best bet. But yeah, I agree, probably best to keep them separate

plasmapy/formulary/collisions.py
pheuer: Ah, I see, I don't think you need to support the use case where T_a is partially supplied and v_a is partially supplied. So, it could just be one or the other at a time.

JaydenR2305: Ok that definitely simplifies things then! In that case we can use your suggested version for checking everything

JaydenR2305:
@pheuer Any clue what's going on with `test_optical_density_histogram` ?

pheuer:
> @pheuer Any clue what's going on with `test_optical_density_histogram` ?

Hm, no, but that's one of mine. If it's failing intermittently that's a problem - I'll try to fix it

pheuer:
Error text: doesn't occur on my machine so maybe it's a python 3.8 issue. But then not sure why it's happening now...

Error messsage:
<details>
=================================== FAILURES ===================================
____________ TestSyntheticRadiograph.test_optical_density_histogram ____________

self = <plasmapy.diagnostics.tests.test_charged_particle_radiography.TestSyntheticRadiograph object at 0x7f971c4a1520>

        """
        """
        bins = (200, 60)
    
        )
        )
    
        zero_mask = intensity == 0
        od = -np.log10(intensity / i0)
    
>       assert np.allclose(od[~zero_mask], od_results[2][~zero_mask])
E       assert False
E        +  where False = <function allclose at 0x7f97470a25e0>(array([ 3.81742655e-01,  9.16855857e-01,  4.00226061e-01, -9.53785996e-02,\n       -2.33759381e-02,  4.60923901e-01,  1...0,  3.64013888e-01,\n       -1.24536828e-01, -1.46531122e-01,  3.81742655e-01,  5.57833914e-01,\n        5.31504975e-01]), array([ 0.37923951,  0.91435271,  0.397[722](https://github.com/PlasmaPy/PlasmaPy/runs/7942041305?check_suite_focus=true#step:6:723)92, -0.09788174, -0.02587908,\n        0.45842076,  1.45842076, -0.01140126, ...\n       -0.0187005 ,  1.45842076,  0.36151074, -0.12703997, -0.14903427,\n        0.37923951,  0.55533077,  0.52900183]))
E        +    where <function allclose at 0x7f97470a25e0> = np.allclose

plasmapy/diagnostics/tests/test_charged_particle_radiography.py:597: AssertionError
</details>

JaydenR2305:
It's working fine for me locally as well. There was an issue created awhile ago #1458

JaydenR2305:
All that is left to do now is add tests for x >> 0 !

plasmapy/formulary/collisions.py
```diff
@@ -565,21 +570,21 @@ def _replace_nan_velocity_with_thermal_velocity(V, T, m):
         raise utils.PhysicsError("Collisions are not possible with a zero velocity.")
 
     if V is None:
-        return thermal_speed(T, "e-", mass=m)
+        return thermal_speed(T, species, mass=m)
 
     if not np.any(np.isnan(V)):
         return V
 
     if not (np.isscalar(V.value) or np.isscalar(T.value)):
```
pheuer, line 578: I think the better way to handle the scalar/vector issue is to just force everything to be a vector at the beginning. E.g. 

```python
if np.isscalar(V.value):
     V = np.array([V.value, ]) * V.unit
```

Then you can treat it like a vector from then on

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,157 @@ def coupling_parameter(
...
+        Parameters
+        ----------
+        test_particle : |Particle|
+            The test particle streaming through a background of field particles.
+
+        field_particle : |Particle|
+            The background particle being interacted with.
+
+        v_a : `~astropy.units.Quantity`, optional
+            The relative velocity between particles. If not provided,
```
pheuer, line 2078: Note that if v_a is not provided, T_a is required.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,157 @@ def coupling_parameter(
...
+        ... )
+        >>> frequencies.slowing_down
+        <Quantity 1.83701432e+11 Hz>
+
+        See Also
+        --------
+        ~plasmapy.formulary.collisions.Coulomb_log : Evaluates the Coulomb
+            logarithm for two interacting electron species.
+
+        ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
```
pheuer, line 2134: I would explain how this differs from this function. Would this be eventually replacing `collision_frequency`?

plasmapy/formulary/collisions.py
```diff
@@ -555,7 +558,9 @@ def _process_inputs(T: u.K, species: (particles.Particle, particles.Particle), V
     return T, masses, charges, reduced_mass, V
 
 
-def _replace_nan_velocity_with_thermal_velocity(V, T, m):
+def _replace_nan_velocity_with_thermal_velocity(
```
pheuer, line 561: Are you using this function anywhere anymore?

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,157 @@ def coupling_parameter(
...
+        field_particle: particles.Particle,
+        *,
+        v_a: u.cm / u.s = None,
+        T_a: u.K = None,
+        n_b: u.cm**-3,
+        T_b: u.K,
+        Coulomb_log: u.dimensionless_unscaled,
+    ):
+        r"""
+        Compute collision frequencies
```
pheuer, line 2067: "Compute collision frequencies between test particles (labeled 'a') and field particles (labeled 'b')"

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,157 @@ def coupling_parameter(
...
+
+        mass_ratio = test_particle.mass / field_particle.mass
+
+        self.v_0 = self._v_0()
+        self.slowing_down = (1 + mass_ratio) * phi * self.v_0
+        self.transverse_diffusion = 2 * ((1 - 1 / (2 * x)) * phi + phi_prime) * self.v_0
+        self.parallel_diffusion = (phi / x) * self.v_0
+        self.energy_loss = 2 * (mass_ratio * phi - phi_prime) * self.v_0
+
+    def _v_0(self):
```
pheuer, line 2164: Is this part equivalent to the Lorentz collision frequency in collision_frequency? If so, I recommend making this public, exposing it, and then planning to depreciate the `collision_frequency` function in a future release.

plasmapy/formulary/collisions.py
JaydenR2305: I wasn't able to get the reference for that working... I was getting a reference error so I have temporarily removed the see also section. Are you able to see what I did wrong?

plasmapy/formulary/collisions.py
JaydenR2305: This function is used in namurphy's functions for processing input. I just extended it to include particles other than electrons

plasmapy/formulary/collisions.py
JaydenR2305: I didn't write this function but I agree there are some modifications that would make it a little cleaner

plasmapy/formulary/collisions.py
pheuer: Ah, got it

plasmapy/formulary/collisions.py
pheuer: Ah ok, you can also just leave it

plasmapy/formulary/collisions.py
pheuer: Hm, the collision_frequency one is working for me. The other one is just a typo: "Coulomb_logarithm"

plasmapy/formulary/collisions.py
JaydenR2305: How would I go about exposing it, besides just making it public? Just making sure to document that it's available within the class?

plasmapy/formulary/collisions.py
JaydenR2305: Oh that might be why then

plasmapy/formulary/collisions.py
pheuer: Yeah, just remove the underscore so it's public and add documentation. I would give it a more descriptive name (`Lorentz_collision_frequency`?) and add a citation to Eq. 2.14 in  Callan to the Notes section for this method (as on the docs for the current collision_freq function)

Look through the existing `collision_freq` function and make sure tehre are no use-cases that aren't currently covered by this new implementation. If  you can cover all of them, I'd suggest we add a Depreciation warning to that function and open an issue to remove it a couple releases after this PR is merged.

@namurphy let me know if you have any objection to phasing out the older form of this function in favor of this new, more generalized class.

namurphy: Nope, no problem from me!

Andrew3577:
Nick, 

How do I disconnect myself from 50 messages that I receive from PlasmaPY every day? 
Sorry, but it must have been my error, but I do not know how to fix it. 

On a good side, I am already trying to run ExH fields and particles around, and I already discovered problems with the PlasmaPy subroutines. 

Thank you for your help again. 

Andrew 
***@***.***

JaydenR2305:
@pheuer I can't seem to find what's wrong with the `slowing_down` fast limit for ion-ion interactions...

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+        "energy_loss",
+        "v_0",
+    ]
+
+    return_values_to_test = [
+        "slowing_down",
+        "transverse_diffusion",
+        "parallel_diffusion",
+        "energy_loss",
+    ]
```
pheuer, line 1730: Possibly more efficient?

```suggestion
    attributes_to_test = [
        "slowing_down",
        "transverse_diffusion",
        "parallel_diffusion",
        "energy_loss",
    ]
    return_values_to_test = attributes_to_test + ["v_0",]
```

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+    @pytest.mark.parametrize("attribute_to_test", attributes_to_test)
+    def test_units(self, attribute_to_test):
+        """Test the return units"""
+
+        assert getattr(self.attribute_test_case, attribute_to_test).unit.is_equivalent(
+            u.Hz
+        )
+
+    @staticmethod
+    def get_limit_value(interaction_type, limit_type, frequencies):
```
pheuer, line 1743: I find the name 'frequencies' here confusing since it contains stuff other than just the frequencies. What about 'cases' or something?

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+        T_a = (frequencies.T_a * k_B).to(u.eV).value
+        T_b = (frequencies.T_b * k_B).to(u.eV).value
+
+        limit_values = []
+
+        if interaction_type == "e|e":
+            if limit_type == "slow":
+                limit_values.extend(
+                    [
+                        5.8 * 10**-6 * T_b ** (-3 / 2),
```
pheuer, line 1759: FYI you can write scientific notation more concisely as `5.8e-6`. Also, -3/2 = -1.5 and etc. is also a little more compact since you eliminate the need for parentheses

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+                        1.8 * 10**-7 * mu ** (-1 / 2) * T_a ** (-3 / 2),
+                        9
+                        * 10**-8
+                        * mu ** (1 / 2)
+                        * mu_prime ** (-1)
+                        * T_b
+                        * T_a ** (-5 / 2),
+                    ]
+                )
+
```
pheuer, line 1855:
```suggestion
# The expected energy loss collision frequency should always equal this
```

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+                    "n_b": 1e20 * u.cm**-3,
+                    "Coulomb_log": 10 * u.dimensionless_unscaled,
+                },
+            ),
+            # Fast limit (x >> 1)
+            (
+                "e|e",
+                "fast",
+                (Particle("e-"), Particle("e-")),
+                {
```
pheuer, line 1917: In the fast limit cases you should supply v_a not T_a. If the plasma isn't thermal than it doesn't really have a Maxwellian 'temperature' so it's better to avoid using one.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+    ):
+        """Test the return values"""
+
+        value_test_case = CollisionFrequencies(
+            *constructor_arguments, **constructor_keyword_arguments
+        )
+
+        coulomb_density_constant = (
+            constructor_keyword_arguments["Coulomb_log"].value
+            * constructor_keyword_arguments["n_b"].value
```
pheuer, line 1974: Careful with the units - if you want this to not be a quantity then I would do .to(u.cm**-3).value just to be sure the units are what you expect. Otherwise .value will just strip whatever is there which can hide an issue.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+        constructor_arguments,
+        constructor_keyword_arguments,
+    ):
+        """Test the return values"""
+
+        value_test_case = CollisionFrequencies(
+            *constructor_arguments, **constructor_keyword_arguments
+        )
+
+        coulomb_density_constant = (
```
pheuer, line 1972: Those limits also have a factor of Z^2 on the LHS for all the cases except for e-e, but I don't see that anywhere - should that also be included in this constant? That would definitely change the collision frequency.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+
+        expected_limit_values = self.get_limit_value(
+            interaction_type, limit_type, value_test_case
+        )
+
+        for i, (attribute_name, expected_limit_value) in enumerate(
+            zip(self.return_values_to_test, expected_limit_values)
+        ):
+            calculated_limit_value = getattr(value_test_case, attribute_name).value
+
```
pheuer, line 1985: Explain why this is being done - otherwise it looks a little odd why this one is in different units than the others.

```suggestion
# Energy loss limit value is already in units of frequencies because of the way it is calculated
```

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+            assert np.allclose(
+                calculated_limit_value, expected_limit_value, rtol=0.1, atol=0
+            )
+
+    @pytest.mark.parametrize(
+        "expected_error, constructor_arguments, constructor_keyword_arguments",
+        [
+            (
+                ValueError,
```
pheuer, line 1999: Add comments above each of these test cases that explain what each one is testing (e.g. why does it raise that error). That also helps delineate the cases and makes this mess easier to read haha.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+                "test_particle": Particle("e-"),
+                "field_particle": Particle("e-"),
+                "T_a": ones_array * u.eV,
+                "T_b": ones_array * u.eV,
+                "n_b": ones_array * u.cm**-3,
+                "Coulomb_log": ones_array * u.dimensionless_unscaled,
+            },
+        ],
+    )
+    def test_handle_nparrays(self, constructor_keyword_arguments):
```
pheuer, line 2043: @namurphy Don't we have some testing utilities for making sure functions handle ndarrays? I can't find them...

pheuer, line 2043: nparray -> ndarray ([technically what the datatype is called](https://numpy.org/doc/stable/reference/arrays.ndarray.html#:~:text=An%20ndarray%20is%20a%20(usually,the%20sizes%20of%20each%20dimension.))

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+                limit_values.extend(
+                    [
+                        5.8 * 10**-6 * T_b ** (-3 / 2),
+                        5.8 * 10**-6 * T_b ** (-1 / 2) * T_a ** (-1),
+                        2.9 * 10**-6 * T_b ** (-1 / 2) * T_a ** (-1),
+                    ]
+                )
+            elif limit_type == "fast":
+                limit_values.extend(
+                    [
```
pheuer, line 1766: T_a is always actually v_a wherever it appears in this section: temperature and velocity have the same units but they aren't the same physical quantity in this context (thermal motion vs. a uniform drift) so I would rename T_a -> v_a to avoid confusion.

pheuer:
> @pheuer I can't seem to find what's wrong with the slow limit for `slowing_down` for ion-ion interactions...

Check out my most recent comments: I think you might be missing a factor of Z^2 on the e-i and i-e limiting cases, and a factor of Z_a^2 Z_b^2 on the i-i one. It could be it's only failing the i-i one because there the factor is bigger and is pushing it out of your error margins. 

If that turns out to be the case, I'd tighten the error on those assertions to 5% or 1% to make sure they fail if something else is broken.

JaydenR2305:
> > @pheuer I can't seem to find what's wrong with the slow limit for `slowing_down` for ion-ion interactions...
> 
> 
> 
> Check out my most recent comments: I think you might be missing a factor of Z^2 on the e-i and i-e limiting cases, and a factor of Z_a^2 Z_b^2 on the i-i one. It could be it's only failing the i-i one because there the factor is bigger and is pushing it out of your error margins. 
> 
> 
> 
> If that turns out to be the case, I'd tighten the error on those assertions to 5% or 1% to make sure they fail if something else is broken.

Z is equal to one, I used Na+ and Cl- for the test ions. I should definitely add support for Z =/= 1 though

JaydenR2305:
> > @pheuer I can't seem to find what's wrong with the slow limit for `slowing_down` for ion-ion interactions...
> 
> 
> 
> Check out my most recent comments: I think you might be missing a factor of Z^2 on the e-i and i-e limiting cases, and a factor of Z_a^2 Z_b^2 on the i-i one. It could be it's only failing the i-i one because there the factor is bigger and is pushing it out of your error margins. 
> 
> 
> 
> If that turns out to be the case, I'd tighten the error on those assertions to 5% or 1% to make sure they fail if something else is broken.

Z is equal to one, I used Na+ and Cl- for the test ions. I should definitely add support for Z =/= 1 though

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,347 @@ def test_kwarg_method_error(self):
...
+            mu_prime = (frequencies.field_particle.mass / m_p).value
+
+            if limit_type == "slow":
+                limit_values.extend(
+                    [
+                        6.8
+                        * 10**-8
+                        * mu_prime ** (1 / 2)
+                        * mu ** (-1)
+                        * (1 + mu_prime / mu) ** (-1 / 2)
```
pheuer, line 1822: I think this is a typo? 
```suggestion
                        * (1 + mu_prime / mu) 
```

pheuer:
> Z is equal to one, I used Na+ and Cl- for the test ions. I should definitely add support for Z =/= 1 though

Ah, I see, ok that would be good to include a note on and provide at least a couple test case for Z_a !=1, Z_b != 1, and Z_a != Z_b !=  1 to make sure the charge scaling is working correctly in the `CollisionFrequency` methods!

plasmapy/formulary/tests/test_collisions.py
JaydenR2305: Great catch ! MIT has a typo in their formulary 👀
image

plasmapy/formulary/tests/test_collisions.py
JaydenR2305: Yep. It just doesn't work with classes (since they don't return values)

plasmapy/formulary/tests/test_collisions.py
pheuer: Yikes - that's the 2013 formulary so I guess that got fixed in later versions! It's right in my 2016 one. 

If you want a digital copy here's a more up-to-date link I should have searched for when opening that issue...
https://library.psfc.mit.edu/catalog/online_pubs/NRL_FORMULARY_18.pdf

JaydenR2305:
> > @pheuer I can't seem to find what's wrong with the slow limit for `slowing_down` for ion-ion interactions...
> 
> Check out my most recent comments: I think you might be missing a factor of Z^2 on the e-i and i-e limiting cases, and a factor of Z_a^2 Z_b^2 on the i-i one. It could be it's only failing the i-i one because there the factor is bigger and is pushing it out of your error margins.
> 
> If that turns out to be the case, I'd tighten the error on those assertions to 5% or 1% to make sure they fail if something else is broken.

I double checked on this and the current margin seems to do quite well for preventing this sort of error with charge 👍

pheuer (commented):
Add a depreciation warning in collision_frequency

@namurphy am I doing this approximately right?

```python
from plasmapy.utils.decorators import deprecated
from plasmapy.utils.exceptions import PlasmaPyFutureWarning

...

deprecated(
since="0.8.0", 
warning_type=PlasmaPyFutureWarning,
message= ("The collision_frequency function has been replaced by the more general "
                   "CollisionFrequency class. To replicate the functionality of collision_frequency, "
                   "create a CollisionFrequency class and access the `Lorentz_collision_frequency` "
                   "attribute.")
)
                    

```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+    )
+    @particles.particle_input
+    def __init__(
+        self,
+        test_particle: particles.Particle,
+        field_particle: particles.Particle,
+        *,
+        v_a: u.cm / u.s = None,
+        T_a: u.K = None,
+        n_b: u.cm**-3,
```
pheuer, line 2075: same MKS

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        T_b: u.K,
+        Coulomb_log: u.dimensionless_unscaled,
+    ):
+        r"""
+        Compute collision frequencies between test particles (labeled 'a') and field particles (labeled 'b')
+
+        Parameters
+        ----------
+        test_particle : |Particle|
```
pheuer, line 2084:
```suggestion
        test_particle : `~plasmapy.particles.particle_class.Particle`
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+            The test particle streaming through a background of field particles.
+
+        field_particle : |Particle|
```
pheuer, line 2087:
```suggestion
        field_particle : `~plasmapy.particles.particle_class.Particle`
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+            The background particle being interacted with.
+
+        v_a : `~astropy.units.Quantity`, optional
+            The relative velocity between particles. If not provided, T_a must be specified and
+            thermal velocity is assumed: :math:`μ v_a^2 \sim 2 k_B T_a` where
+            :math:`μ` is the reduced mass.
```
pheuer, line 2093: Add a note about cannot be negative. Also remember to add that this is required if T_a is not provided.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+
+        T_a : `~astropy.units.Quantity`, optional
+            The temperature of the test species. Only necessary if :math:`v_a` is not provided.
```
pheuer, line 2096:
```suggestion
            The temperature of the test species - used to calculate the thermal velocity. Ignored if :math:`v_a` is provided.
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+
+        n_b : `~astropy.units.Quantity`
+            The number density of the background field particles.
+
+        T_b : `~astropy.units.Quantity`
+            The temperature of the background field particles.
```
pheuer, line 2102: Specify in units of temperature (or 'convertible to degrees Kelvin')

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        v_a : `~astropy.units.Quantity`, optional
+            The relative velocity between particles. If not provided, T_a must be specified and
+            thermal velocity is assumed: :math:`μ v_a^2 \sim 2 k_B T_a` where
+            :math:`μ` is the reduced mass.
+
+        T_a : `~astropy.units.Quantity`, optional
+            The temperature of the test species. Only necessary if :math:`v_a` is not provided.
+
+        n_b : `~astropy.units.Quantity`
+            The number density of the background field particles.
```
pheuer, line 2099: Specify in units convertible to m**-3

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+
+            transverse diffusion - :math:`\frac{d}{dt}\left(\overline{\textbf{v}}_{α}-\overline{\textbf{v}}_{α}\right)_{⊥}^{2}=ν_{⊥}^{α\backslashβ}v_{α}^{2}`
+
+            parallel diffusion - :math:`\frac{d}{dt}\left(\overline{\textbf{v}}_{α}-\overline{\textbf{v}}_{α}\right)_{∥}^{2}=ν_{∥}^{α\backslashβ}v_{α}^{2}`
+
+            energy loss - :math:`\frac{d}{dt}v_{α}^{2}=-ν_{ϵ}^{α\backslashβ}v_{α}^{2}`
+
+        For general solutions and limiting cases we encourage the curious reader to refer to p. 31 of :cite:t:`nrlformulary:2019`
+
+        This function uses CGS units under the hood to coincide with our references, but passing MKS units is acceptable.
```
pheuer, line 2127: I would delete this and make it a comment later - it's only relevant to people reading the code
```suggestion
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        <Quantity 1.83701432e+11 Hz>
+
+        See Also
+        --------
+        ~plasmapy.formulary.collisions.Coulomb_logarithm : Evaluates the Coulomb
+            logarithm for two interacting electron species.
+
+        ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
+        """
+
```
pheuer, line 2149:
```suggestion
# Note: This function uses CGS units internally to coincide with our references. Input is taken in MKS units and then converted as necessary. Output is in MKS units.
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        ------
+        `ValueError`
+            If both :math:`v_a` and :math:`T_a` are specified, or neither
+            :math:`v_a` nor :math:`T_a` are specified.
+
+        Notes
+        -----
+        The collision frequencies between two interacting particles are given by four differential equations:
+
+            slowing down - :math:`\frac{d\textbf{v}_{α}}{dt}=-ν_{s}^{α\backslashβ}\textbf{v}_{α}`
```
pheuer, line 2117: I think we should replace 'slowing down' everywhere with 'momentum loss' since that's really what this is (I think). Anyway, same thing as 'slowing down' just more precise. And it mirrors energy loss.

So I would change that in the attribute name too.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        ... )
+        >>> frequencies.slowing_down
+        <Quantity 1.83701432e+11 Hz>
+
+        See Also
+        --------
+        ~plasmapy.formulary.collisions.Coulomb_logarithm : Evaluates the Coulomb
+            logarithm for two interacting electron species.
+
+        ~plasmapy.formulary.collisions.collision_frequency : The Lorentz collision frequency.
```
pheuer, line 2147: Remove this from here (since we're deprecating that function) but put a See Also reference to this new class in the collision_frequency docstring.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        elif T_a is not None:
+            raise ValueError("Please specify either v_a or T_a, not both.")
+
+        self.test_particle = test_particle
+        self.field_particle = field_particle
+        self.v_a = v_a
+        self.T_a = (
+            T_a
+            if T_a is not None
+            else (0.5 * test_particle.mass.to(u.g) * v_a**2 / k_B).to(u.K)
```
pheuer, line 2164: Do you need the .to(u.g) here, if you're then going to convert to u.K anyway?

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        self.T_b = T_b
+        self.Coulomb_log = Coulomb_log
+
+        x = self._x()
+        phi = self._phi(x)
+        phi_prime = self._phi_prime(x)
+
+        mass_ratio = test_particle.mass / field_particle.mass
+
+        self.v_0 = self.Lorentz_collision_frequency()
```
pheuer, line 2176: If you want to have this as a shorter alias then you should make this private, e.g. `self._v0`. I don't see a reason to have it around after you calculate the other frequencies though, so maybe just make it a variable rather than an attribute?

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        The collision frequencies between two interacting particles are given by four differential equations:
+
+            slowing down - :math:`\frac{d\textbf{v}_{α}}{dt}=-ν_{s}^{α\backslashβ}\textbf{v}_{α}`
+
+            transverse diffusion - :math:`\frac{d}{dt}\left(\overline{\textbf{v}}_{α}-\overline{\textbf{v}}_{α}\right)_{⊥}^{2}=ν_{⊥}^{α\backslashβ}v_{α}^{2}`
+
+            parallel diffusion - :math:`\frac{d}{dt}\left(\overline{\textbf{v}}_{α}-\overline{\textbf{v}}_{α}\right)_{∥}^{2}=ν_{∥}^{α\backslashβ}v_{α}^{2}`
+
+            energy loss - :math:`\frac{d}{dt}v_{α}^{2}=-ν_{ϵ}^{α\backslashβ}v_{α}^{2}`
+
```
pheuer, line 2124: For completeness I would write out the formulas for the collision frequencies and the definitions of the Lorentz collision frequency, x, and phi. It is important to explain what x is, I think, to understand what is happening conceptually here. 

For a future PR you could write a nice example notebook using this function in which you plot psi as a function of x and illustrate where the limiting approximations work! I made a plot like this in my thesis: 

image

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        .. math::
+
+            ν = n σ v \ln{Λ}
+
+        where :math:`n` is the particle density, :math:`σ` is the
+        collisional cross-section, :math:`v` is the inter-particle velocity
+        (typically taken as the thermal velocity), and :math:`\ln{Λ}` is the
+        Coulomb logarithm accounting for small angle collisions.
+
+        See Equation (2.14) in :cite:t:`callen:unpublished`.
```
pheuer, line 2196:
```suggestion
        See Equation (2.14) in :cite:t:`callen:unpublished`.
        
        The Lorentz collision frequency is equivalent to the variable :math:`\nu_0^{\alpha/\beta}` on p. 31 of :cite:t:`nrlformulary:2019`
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+            4
+            * np.pi
+            * (self.test_particle.charge_number * e.esu) ** 2
+            * (self.field_particle.charge_number * e.esu) ** 2
+            * self.Coulomb_log
+            * self.n_b
+            / (self.test_particle.mass.to(u.g) ** 2 * self.v_a**3)
+        ).to(u.Hz)
+
+    def _x(self) -> u.dimensionless_unscaled:
```
pheuer, line 2209: Include a brief 1-sentence docstring on these private methods, just for future coders reference. Also, x could be a property and might actually be good to expose to the users and document fully: that way they can see which regime the are in.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,402 @@ def test_kwarg_method_error(self):
...
+
+    ones_array = np.ones(5)
+
+    @pytest.mark.parametrize("attribute_to_test", attributes_to_test)
+    def test_units(self, attribute_to_test):
+        """Test the return units"""
+
+        assert getattr(self.attribute_test_case, attribute_to_test).unit.is_equivalent(
+            u.Hz
+        )
```
pheuer, line 1735: Once you add x as a property, add a test here that it is dimensionless

plasmapy/formulary/tests/test_collisions.py
pheuer: This notation still needs to be updated.

pheuer (commented):
Looking good! Here's another batch of comments

plasmapy/formulary/collisions.py
```diff
@@ -555,7 +558,9 @@ def _process_inputs(T: u.K, species: (particles.Particle, particles.Particle), V
     return T, masses, charges, reduced_mass, V
 
 
-def _replace_nan_velocity_with_thermal_velocity(V, T, m):
+def _replace_nan_velocity_with_thermal_velocity(
+    V, T, m, species=particles.Particle("e-")
```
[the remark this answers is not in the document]
JaydenR2305: We could get rid of the m though it would mean I have to change all of the functions that use it- which might be necessary in this case since we're expanding the generality of the function

plasmapy/formulary/collisions.py
JaydenR2305: nope

plasmapy/formulary/collisions.py
JaydenR2305: The current behavior is to error if both temperature and velocity are provided. Did you intend for it to behave as you say here?

plasmapy/formulary/collisions.py
JaydenR2305: This should also say relative speed instead of velocity

plasmapy/formulary/collisions.py
JaydenR2305: Is there a reason for typing out the whole reference as opposed to using the substitution?

plasmapy/formulary/tests/test_collisions.py
JaydenR2305: Oh my bad I misread this

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,402 @@ def test_kwarg_method_error(self):
...
+        ):
+            calculated_limit_value = getattr(value_test_case, attribute_name).value
+            # Energy loss limit value is already in units of frequencies because of the way it is calculated
+            if attribute_name != "energy_loss":
+                calculated_limit_value = calculated_limit_value / (
+                    coulomb_density_constant * charge_constant
+                )
+
+            assert np.allclose(
+                calculated_limit_value, expected_limit_value, rtol=0.1, atol=0
```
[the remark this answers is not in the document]
JaydenR2305: How does 5% sound?

plasmapy/formulary/collisions.py
pheuer: Oh, is that what that does? I've never seen that before.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2038,178 @@ def coupling_parameter(
...
+        n_b={"can_be_negative": False},
+        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
+    )
+    @particles.particle_input
+    def __init__(
+        self,
+        test_particle: particles.Particle,
+        field_particle: particles.Particle,
+        *,
+        v_a: u.cm / u.s = None,
```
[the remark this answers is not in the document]
pheuer: Ok, then I think its fine to leave it like this and just make sure the documentation says it takes MKS units for consistency with the rest of PlasmaPy

plasmapy/formulary/collisions.py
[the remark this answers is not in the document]
pheuer: Hm, I see what you mean, ok maybe just leave a # TODO comment here saying that that should be done in the future

plasmapy/formulary/collisions.py
pheuer: Oh, I guess either is ok but your way is probably more failsafe - let's stick with that.

plasmapy/formulary/collisions.py
pheuer: Great - those unit conversions are surprisingly expensive if you do them in high performance code so I'm used to trying to minimize them haha

plasmapy/formulary/tests/test_collisions.py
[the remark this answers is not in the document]
pheuer: I mean I'd like 1% better - what is x for the test examples? If it's <0.01 or >100 I think the limits should be quite accurate?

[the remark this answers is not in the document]
JaydenR2305: X is usually within those bounds you gave for the limit cases.. Getting them to about 3% agreement is doable but 1% might take some more extreme values. I'll do some research when I get a second

JaydenR2305:
@pheuer Any clue what's going on here? I'm able to build the docs fine

plasmapy/formulary/tests/test_collisions.py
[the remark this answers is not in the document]
pheuer: Ok, 1% is actually probably pushing the precision of the constants in the formulary out front of the limit forms, so 5% is probably fine.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2056,209 @@ def coupling_parameter(
...
+            T_a
+            if T_a is not None
+            else (0.5 * test_particle.mass * v_a**2 / k_B).to(u.K)
+        )
+        self.n_b = n_b
+        self.T_b = T_b
+        self.Coulomb_log = Coulomb_log
+
+        # These attributes are used in testing
+        self._x = self.x
```
pheuer, line 2177: _x and _v0 don't really need to be attributes - they could just be variables here (just keeping the class simple)

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2056,209 @@ def coupling_parameter(
...
+        self._v_0 = self.Lorentz_collision_frequency
+
+        phi = self._phi(self.x)
```
pheuer, line 2180: since x is an attribute of self, you don't need to pass it, you can just access it within these functions as `self.x`

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2056,209 @@ def coupling_parameter(
...
+        The phi integrand used in calculating phi
+        """
+
+        return t**0.5 * np.exp(-t)
+
+    def _phi_explicit(self, x: float) -> float:
+        """
+        The non-vectorized method for evaluating the integral for phi
+        """
+        integral, _ = scipy.integrate.quad(self._phi_integrand, 0, x)
```
pheuer, line 2246: Maybe include a note in the documentation that this function uses `scipy.integrate.quad` with the default error tolerances to do the psi integral (which you should be adding to the documentation too). 

I hadn't seen this function before but it's clever: reduces the sensitivity to choosing an integration step size!

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2056,209 @@ def coupling_parameter(
...
+
+        return integral
+
+    def _phi(self, x: u.dimensionless_unscaled):
+        """
+        The parameter phi used in calculating collision frequencies
+        For more information refer to page 19 of the NRL Formulary
```
pheuer, line 2253: pg. 31?

plasmapy/formulary/collisions.py
```diff
@@ -996,7 +1002,24 @@ def collision_frequency(
     >>> species = ('e', 'p')
     >>> collision_frequency(T, n, species)
     <Quantity 70249... Hz>
+
```
pheuer, line 1005: Add a note to the collision_frequency docstring saying that it is being deprecated and linking to the new class. 

@namurphy is there a way to get this to appear as a banner warning or something above the docs on this page?

plasmapy/formulary/collisions.py
JaydenR2305: 🤦‍♂️

plasmapy/formulary/collisions.py
JaydenR2305: Yeah the "vectorization" for calculating phi is just a for loop under the hood to ensure that we minimize the error as much as possible... I read somewhere that attempting to vectorize integration using other methods can cause unwanted error

plasmapy/formulary/collisions.py
JaydenR2305: Oh weird, for some reason I thought I had tried that and `getattr` had an issue. I guess it does work though

plasmapy/formulary/collisions.py
JaydenR2305: image

Now just to figure out why it doesn't like the `CollisionFrequencies` reference

pheuer: Must have  been - it's working for me now

JaydenR2305:
@pheuer Ready for another batch of comments

JaydenR2305:
One thing I think we're missing is documentation how to actually access the value of the rates. Should I make them each into properties so that they show up in the attributes documentation?

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2063,234 @@ def coupling_parameter(
...
+            if T_a is not None
+            else (0.5 * test_particle.mass * v_a**2 / k_B).to(u.K)
+        )
+        self.n_b = n_b
+        self.T_b = T_b
+        self.Coulomb_log = Coulomb_log
+
+        mass_ratio = test_particle.mass / field_particle.mass
+
+        self.momentum_loss = (
```
pheuer, line 2205: So the way you have it written now, it will re-compute the properties phi, phi_prime, and Lorentz_collision_frequency every time they are called. You could prevent this two ways: 

1) Use the [cached_property package](https://pypi.org/project/cached-property/) so that the value is stored after the first execution and then just returned. 
2) Do something like you were doing before with a variable, e.g. 

```python
phi = self.phi

slowing_down = .... * phi
transverse_diffusion = .... * phi
...
```

My preference would be towards the cached property option, since you are exposing `Lorentz_collision_frequency` and making it cached would mean that users of that would also get this benefit. You could also make `x` cached too for that reason.

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,408 @@ def test_kwarg_method_error(self):
...
+    )
+
+    return_values_to_test = [
+        "momentum_loss",
+        "transverse_diffusion",
+        "parallel_diffusion",
+        "energy_loss",
+    ]
+
+    ones_array = np.ones(5)
```
pheuer, line 1724: Can you add a test where the inputs are 2D arrays? If I want to make a plot of a parameter space, I'd like to feed in a 2D array of density (varying in one direction, constant in the other) and velocity (varying and constant in the opposite directions). Just checking if it works with `ones_array2d = np.ones([5,5])` should be sufficient

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1698,3 +1700,408 @@ def test_kwarg_method_error(self):
...
+
+    @pytest.mark.parametrize(
+        "attribute_to_test, expected_attribute_units",
+        [
+            ("momentum_loss", u.Hz),
```
pheuer, line 1729: Add `Lorentz_collision_frequency` to this test

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2063,234 @@ def coupling_parameter(
...
+            (1 + mass_ratio) * self.phi * self.Lorentz_collision_frequency
+        )
+        self.transverse_diffusion = (
+            2
+            * ((1 - 1 / (2 * self.x)) * self.phi + self._phi_prime)
+            * self.Lorentz_collision_frequency
+        )
+        self.parallel_diffusion = (self.phi / self.x) * self.Lorentz_collision_frequency
```
pheuer, line 2213: Good catch - we DEFINITELY need these to be documented! Making them properties is the best way to do that. Use cached_property on those too.

JaydenR2305:
@pheuer Ready for another batch

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2064,271 @@ def coupling_parameter(
...
+        self.Coulomb_log = Coulomb_log
+
+    @cached_property
+    def _mass_ratio(self):
+        return self.test_particle.mass / self.field_particle.mass
+
+    @cached_property
+    def momentum_loss(self):
+        """
+        The momentum loss for the specified particles.
```
pheuer, line 2225: "momentum loss rate due to collisions" 

This is just the rate at which momentum is lost, not the amount of momentum lost, which would be the solution to this differential equation (which would be exponential energy loss, e.g.   v = e^(- \nu_ab t) )

Same for the others: e.g. rate of transverse velocity diffusion due to collisions , rate of energy loss.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2064,271 @@ def coupling_parameter(
...
+
+        For values of x<<1 (the 'slow' or 'thermal' limit) or x>>1 (the 'fast' or 'beam' limit),
+        :math:`\psi` asymptotes to zero or one respectively. For simplified expressions in these limits
+        we encourage the curious reader to refer to p. 31 of :cite:t:`nrlformulary:2019`
+
+        Examples
+        --------
+        >>> import astropy.units as u
+        >>> T_a = 1 * u.eV
+        >>> n_b = 1e20 * u.cm**-3
```
pheuer, line 2173: You should use MKS units in the examples for consistency with the rest of PlasmaPy

pheuer (commented):
Only one minor change - I'll look over the completed  CI tests + docs one more time before merging, but otherwise I think this looks great! Thanks so much for doing this!!

plasmapy/formulary/collisions.py
JaydenR2305: Is expressing the temperature in electronvolts acceptable?

pheuer: That's fine (that's not really MKS or CGS, just plasma physicists being weird)

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1714,6 +1714,35 @@ class TestCollisionFrequencies:
         Coulomb_log=1 * u.dimensionless_unscaled,
     )
 
+    MKS_unit_conversion_test_constructor_arguments = {
```
JaydenR2305, line 1717: @pheuer If you could think of a better name for this that would be appreciated, I think this is probably fine though

pheuer: I think this is fine! A bit long, but it fits in the lines fine and is  very descriptive.

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2064,271 @@ def coupling_parameter(
...
+        return (
+            2
+            * (self._mass_ratio * self.phi - self._phi_prime)
+            * self.Lorentz_collision_frequency
+        )
+
+    @cached_property
+    def Lorentz_collision_frequency(self):
+        r"""
+        The Lorentz collision frequency (see Ch. 5 of :cite:t:`chen:2016`) is given
```
pheuer, line 2261: This line is rendering wrong for some reason: 

image

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2064,271 @@ def coupling_parameter(
...
+        """
+        integral, _ = scipy.integrate.quad(self._phi_integrand, 0, x)
+
+        return integral
+
+    @cached_property
+    def phi(self):
+        """
+        The parameter phi used in calculating collision frequencies
+        calculated using the default error tolerances of `~scipy.integrate.quad`
```
pheuer, line 2320: Add period at end 
```suggestion
        calculated using the default error tolerances of `~scipy.integrate.quad`.
```

plasmapy/formulary/collisions.py
```diff
@@ -2033,3 +2064,271 @@ def coupling_parameter(
...
+        .. math::
+
+            ν = n σ v \ln{Λ}
+
+        where :math:`n` is the particle density, :math:`σ` is the
+        collisional cross-section, :math:`v` is the inter-particle velocity
+        (typically taken as the thermal velocity), and :math:`\ln{Λ}` is the
+        Coulomb logarithm accounting for small angle collisions.
+
+        See Equation (2.14) in :cite:t:`callen:unpublished`.
```
pheuer, line 2273: I think this is rendering like this because the Callan book doesn't have a date attached to it?

image

pheuer (changes requested):
@JaydenR2305 Found a couple minor typos, but then I think it's done!

plasmapy/formulary/collisions.py
pheuer: This appears this way elsewhere in the docs, so I guess we can just leave it.

plasmapy/formulary/collisions.py
JaydenR2305: It could also just cut off after a certain number of words

pheuer, line 2261: I think I figured this out - docstrings are supposed to start with a simple 1-sentence 'summary' which I think is what the attribute summary is, which is getting cut off here. If we separate a simple summary out, that should fix it.



```suggestion
        The Lorentz collision frequency.
        
        The Lorentz collision frequency (see Ch. 5 of :cite:t:`chen:2016`) is given
```

JaydenR2305: Ohhh that makes sense. This is why we need to implement #1668 !

pheuer (approved):
Awesome work @JaydenR2305 !

PR #1678, opened 2022-08-18
Resurrect `chemical_potential` and associated functions

### LINKED ISSUE #726: `plasmapy.formulary.quantum.chemical_potential` bug with powers

JaydenR2305:
Resolves #726 

Floating point errors associated with the `chemical_potential` function seem to have silently been corrected

      docstrings.

### DIFF
--- a/changelog/1678.feature.1.rst
+++ b/changelog/1678.feature.1.rst
@@ -1,0 +1,1 @@
+Reimplemented `~plasmapy.formulary.quantum.chemical_potential`
--- a/changelog/1678.feature.2.rst
+++ b/changelog/1678.feature.2.rst
@@ -1,0 +1,1 @@
+Reimplemented `~plasmapy.formulary.quantum._chemical_potential_interp`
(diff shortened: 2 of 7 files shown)


StanczakDominik (approved):
Huh! Fascinating that this works again. I don't exactly have anything to validate this against, but the changes look reasonable and I think we can try it out in main :) Thanks, @JaydenR2305 !

Issue #1539, opened 2022-04-29
Create a `RelativisticBody` class


namurphy:
Right now it's pretty inconvenient to have a particle and go between velocity, Lorentz factor, energy, and momentum.  It'd be helpful to have a `RelativisticParticle` class that we could use to take one such quantity and make it straightforward to calculate the rest.  I was thinking that this would be best in `plasmapy.formulary.relativistic`.

Some of the input parameters could be `particle`, `V`, `gamma` (for Lorentz factor), `kinetic_energy`, and `momentum`. Some of the attributes could be `rest_mass`, `relativistic_mass`, `total_energy`, `kinetic_energy`, and `mass_energy`.  There could also be an attribute like `v_over_c`, which is often referred to as `beta`, but that could potentially be confused with plasma beta.  

I'm wondering if we could even define the dunder methods for `__add__` and `__subtract__` so that we could do things like...
```Python
>>> cosmic_ray = RelativisticParticle('p+', energy = 100 * u.GeV)
>>> cosmic_ray += 50 * u.GeV  # would increase the energy of the particle by 50 GeV
```

Doing this would make #1538 a lot easier, and it's related to #1537 too.

namurphy:
I'm wondering about the name for this, since `RelativisticParticle` and `Particle` would have very different purposes.  It could be `RelativisticObject` or `RelativisticBody`, which would probably be better than `RelativisticThing`, `RelativisiticStuff`, or even `RelativisticCorporeality`.

StanczakDominik:
I do like `RelativisticBody` for this.

PR #1681, opened 2022-08-19
Add NaN check to prevent raising a false relativity error

TrestanSimon:
This PR adds an additional check in `plasmapy.formulary.relativity.Lorentz_factor` to prevent input of `np.nan` from raising `utils.RelativityError`. Closes https://github.com/PlasmaPy/PlasmaPy/issues/1672.

      docstrings.


TrestanSimon:
pre-commit.ci autofix

StanczakDominik (changes requested):
At a first, early-morning and insufficiently-caffeinated glance (so I might be way off!), it doesn't look like this change will do the trick... Which is why it'd be great to add some tests here, to convince even the most sleepy maintainer ;) ! You can find some inspiration in the ones I listed at https://github.com/PlasmaPy/PlasmaPy/issues/1672#issuecomment-1218086218 :)

TrestanSimon:
> At a first, early-morning and insufficiently-caffeinated glance (so I might be way off!), it doesn't look like this change will do the trick... Which is why it'd be great to add some tests here, to convince even the most sleepy maintainer ;) ! You can find some inspiration in the ones I listed at [#1672 (comment)](https://github.com/PlasmaPy/PlasmaPy/issues/1672#issuecomment-1218086218) :)

It works on my machine and passes the suggested tests (51ee531acb5fac43a50427cfb8c4a4c05307f060). Maybe I am missing something...

plasmapy/formulary/tests/test_relativity.py
```diff
@@ -29,6 +29,14 @@ def test_Lorentz_factor():
...
     ).unit == u.dimensionless_unscaled
 
+    assert np.isnan(Lorentz_factor(np.nan * u.m / u.s))
+
+    assert np.all(np.isnan(Lorentz_factor(np.array([np.nan, np.nan]) * u.m / u.s)))
+
+    numerical_result, nan_result = Lorentz_factor(np.array([1, np.nan]) * u.m / u.s)
+    assert np.isnan(nan_result)
+    assert not np.isnan(numerical_result)
+
```
namurphy, line 39:
```suggestion

    def test_lorentz_factor_nan_input():
        assert np.isnan(Lorentz_factor(np.nan * u.m / u.s))


    def test_lorentz_factor_array_of_nans():
        assert np.all(np.isnan(Lorentz_factor(np.array([np.nan, np.nan]) * u.m / u.s)))


    def test_lorentz_factor_nan_in_array():
        numerical_result, nan_result = Lorentz_factor(np.array([1, np.nan]) * u.m / u.s)
        assert np.isnan(nan_result)
        assert not np.isnan(numerical_result)


    def test_lorentz_factor_exceptions():
```
The tests do look like they're testing the behavior that I was hoping would be added to `Lorentz_factor`, so thank you!

This suggestion breaks up the tests that you added into different functions. A while after we originally wrote the tests for `Lorentz_factor`, we learned that it's a better practice to have each test verify a single unit of behavior.  If there are multiple unrelated `assert` statements in a test (like how the tests of `Lorentz_factor` were written before this pull request), then if one of them fails, the subsequent tests won't get run which could potentially hide errors.

namurphy (commented):
Thank you for submitting this pull request!  This is definitely going to make my life easier in #1539, so I very much appreciate it!

I added some minor suggestions related to formatting and test organization.  Overall it looks good and we should be able to merge this pretty soon.  

Thank you again!

TrestanSimon:
pre-commit.ci autofix

TrestanSimon:
pre-commit.ci autofix

TrestanSimon (commented):
> At a first, early-morning and insufficiently-caffeinated glance (so I might be way off!), it doesn't look like this change will do the trick... Which is why it'd be great to add some tests here, to convince even the most sleepy maintainer ;) ! You can find some inspiration in the ones I listed at [#1672 (comment)](https://github.com/PlasmaPy/PlasmaPy/issues/1672#issuecomment-1218086218) :)

@StanczakDominik it appears to work properly and passes all tests. No changes necessary.

plasmapy/formulary/relativity.py
```diff
@@ -63,7 +63,7 @@ def Lorentz_factor(V: u.m / u.s):
     inf
     """
 
-    if not np.all(np.abs(V) <= c):
+    if not np.all((np.abs(V) <= c) + (np.isnan(V))):
```
StanczakDominik, line 66: This is purely optional, but I think this should be equivalent and might make it a little cleaner to read:
```suggestion
    if not np.all((np.abs(V) <= c) || np.isnan(V)):
```

StanczakDominik (approved):
Yeah, since it passes the tests, I'm a-okay with it! Sorry for the confusion earlier!

plasmapy/formulary/relativity.py
namurphy: I hadn't realized that we could use a logical "or" operator with numpy arrays!  I only remember trying them for lists (e.g., `[True, False] | [False, False]`), which results in a `TypeError`.  The `||` is Java syntax, I think, so we'd want to use `|` here instead of `||`.

TrestanSimon: Resolved by https://github.com/PlasmaPy/PlasmaPy/pull/1681/commits/bf92298c8d24a813890b6acbad5b7ad4f820b7de.

StanczakDominik: Huh! I have no idea where I pulled the double pipe from. It sure wasn't Java! Thanks for spotting that, though!

PR #1683, opened 2022-08-19
921remove direct imports from utils

### LINKED ISSUE #921: Discontinue importing functions, classes, & exceptions directly into plasmapy.utils
Right now `plasmapy/utils/__init__.py` includes a bunch of direct imports of exceptions, etc. from, e.g., `exceptions.py`.  Right now we're able to do things like this:
```Python
>>> from plasmapy.utils import PhysicsError, call_string
```
Because the contents of `plasmapy.utils` are highly varied, I think it would ultimately be cleaner if we didn't import functions, classes, & exceptions directly into `plasmapy.utils` but rather we imported the modules and sub-sub-packages, so that we would have to do things like
```Python
>>> from plasmapy.utils.exceptions import PhysicsError
>>> from plasmapy.utils.formatting import call_string  # after #920 is merged
```
This would prevent namespace pollution in `plasmapy.utils` (which is not a problem yet) and keep things a bit better organized.

hwiks:
Closes #921 

This PR is for now just a test to see if I can set up everything needed to contribute to this project

      docstrings.


StanczakDominik:
!pre-commit.ci autofix

PR #1686, opened 2022-08-22
Separate the patent clause from the license into its own file

namurphy:
[our current license]: https://github.com/PlasmaPy/PlasmaPy/blob/301660f1ac35411f42e2defba7e53698b6e31acc/LICENSE.md
[BSD+Patent license]: https://opensource.org/licenses/BSDplusPatent

Way back in the antiquity of PlasmaPy (#114), we added a patent grant to [our current license] that came from the OSI-approved [BSD+Patent license].  The BSD+Patent license is the BSD 2-clause license plus language that protects against software patents (which we adopted because there are many commercial applications to plasma science).  One complication was that we already had the BSD 3-clause license, so we ended up going with the BSD 3-clause license + patent instead of the canonical BSD 2-clause license + patent, and the BSD-3-clause-patent license doesn't show up on lists.

This pull request splits off the patent grant into its own file.  PlasmaPy will effectively be under the BSD 3-clause license again, but the patent grant will still be in effect.

One additional thing to note is that the BSD+Patent, BSD 2-clause, and BSD 3-clause licenses all say that the package can be redistributed and/or modified so long as redistributions retain "the above copyright notice, this list of conditions, and the following disclaimer" where the list refers to the enumerated list and the disclaimer is the final paragraph of each of the licenses that talks about no implied warranties.  In the BSD+Patent license, the patent grant is not explicitly listed as one of the things that is required to be included in redistributed versions of the code.


StanczakDominik (commented):
Looks okay to me, though of course my only law experience comes from playing Ace Attorney (so if you need someone to shout *OBJECTION!* really loudly...) and this is not even a murder case!

My only worry here is that now the license "key" or "enum value" in setup.cfg and citation, sorry, *CITATION!cff*, describes our license as BSD-3-clause, but that's not the extent of the license because of the separated out patent clause. So the license files are not interchangeable or uniquely identified by the "key".

Does that matter? No idea, just pointing out a possible issue.

StanczakDominik (approved):

rocco8773 (approved):

PR #1690, opened 2022-08-24
change os.path to pathlib

### LINKED ISSUE #1684: Switch to using `pathlib` instead of `os.path`
In newer Python code, it's generally recommended to use [`pathlib`](https://docs.python.org/3/library/pathlib.html) instead of `os.path`.  There are [a few places where we use `os.path`](https://github.com/PlasmaPy/PlasmaPy/search?q=os.path+-plasmapy_sphinx) where we should switch to `pathlib`. 

There are also [places where we are using `pathlib`](https://github.com/PlasmaPy/PlasmaPy/search?q=pathlib&type=code) which could be used as examples.  

We shouldn't make this change in `docs/plasmapy_sphinx` since that's being moved to another package (#1643).

Thanks!

MKastek:

      docstrings.

Closes #1684


MKastek:
pre-commit.ci autofix

MKastek:
pre-commit.ci autofix

plasmapy/diagnostics/tests/test_charged_particle_radiography.py
```diff
@@ -612,7 +611,7 @@ def test_saving_output(tmp_path):
     results_1 = sim.results_dict
 
     # Save result
-    path = os.path.join(tmp_path, "temp.npz")
+    path = str(tmp_path.joinpath("temp.npz"))
```
StanczakDominik, line 614: I always forget if it is, but if tmp path is the pathlib.Path variant, you can do


```suggestion
    path = str(tmp_path / "temp.npz")
```
Which calls joinpath under the hood!

plasmapy/particles/data/test/__init__.py
```diff
@@ -3,7 +3,8 @@
...
 """
 import glob
-import os
 
-data_dir = os.path.dirname(__file__)
-data_files = glob.glob(os.path.join(data_dir, "*.[!p]*"))
+from pathlib import Path
+
+data_dir = Path(__file__).parents[0]
+data_files = glob.glob(str(data_dir.joinpath("*.[!p]*")))
```
StanczakDominik, line 10: Don't Path objects provide their own glob method? So you wouldn't need the extension glob import nor the str conversion?

StanczakDominik (changes requested):
Some minor possible changes to make this more idiomatic for pathlib, also coverage seems to have gone down a little but I'm not investigating THAT from my phone :D if I don't get to it in time, can you take a look and see which lines are newly untested and why? If it's just that there's fewer total lines in the files, we can live with that

MKastek:
pre-commit.ci autofix

plasmapy/particles/data/test/__init__.py
MKastek: Yes Path objects provide glob method, it returns generator so I turned it to list.

plasmapy/plasma/sources/openpmd_hdf5.py
```diff
@@ -148,8 +148,7 @@ def is_datasource_for(cls, **kwargs):
         hdf5 = kwargs.get("hdf5")
         openPMD = kwargs.get("openPMD")
 
-        isfile = os.path.isfile(hdf5)
-        if not isfile:
+        if not Path(hdf5).is_file():
```
MKastek, line 151: This line of code is not covered by tests.

StanczakDominik (approved):
All right! I'll just change the changelog a little for nicer display in the docs. Thanks!

plasmapy/plasma/sources/openpmd_hdf5.py
StanczakDominik: Ah, missed that comment. Let me still investigate this before merging .

StanczakDominik:
Ah, yeah, that looks fine. We'll be rewriting the plasma classes anyway (as in #1687).

PR #1698, opened 2022-08-26
Enable inline code highlighting with `:py:` role

### LINKED ISSUE #1576: Enable inline code highlighting via Sphinx 5.0.0
Sphinx 5.0.0 was released today, and it turns out that it's now possible to have [inline code syntax highlighting](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#inline-code-highlighting) (!!!!!) just like in `code-block` directives.  This change would require making a new role for reStructuredText, so it wouldn't automagically get applied to code snippets with double back ticks. This change would probably make us require `sphinx >= 5.0.0`, and since this release is only a few hours old, and my inclination would be to make this change after our 0.8.0 release.

 - [ ] Add new role (`:py:` or `:python:`?) to [`rst_prolog`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog) in `docs/conf.py` (which is similar to [`rst_epilog`](https://github.com/PlasmaPy/PlasmaPy/blob/1b9238439c1e4a7c1d35b8f812b4805872f0aae6/docs/conf.py#L187) but would need to be defined).
 - [ ] Describe this capability in the documentation guide
 - [ ] Update the requirements to `sphinx >= 5.0.0`

namurphy:
A new capability of Sphinx ≥ 5.0.0 is [inline code highlighting](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html?highlight=inline%20code%20highlighting#inline-code-highlighting).  This PR adds `rst_prolog` to `docs/conf.py` and defines the `:python:` role there, and also bumps the minimum version of Sphinx to 5.0.0.

Closes #1576.


namurphy:
Screenshot from 2022-08-25 23-09-58

From a test I did in the changelog entry.  It works!

StanczakDominik (approved):

PR #1703, opened 2022-08-27
Move fundamental collision frequencies to `MaxwellianCollisionFrequencies`

### LINKED ISSUE #1699: Replace fundamental collision frequencies with CollisionFrequencies
I think the functions `plasmapy.formulary.collisons.fundamental_ion_collision_freq` and `plasmapy.formulary.collisons.fundamental_electron_collision_freq` are closely related to the `collision_frequency` function, which is set to be deprecated in favor of the new `CollisionFrequencies` class introduced by #1676.  When that happens, these other functions will also need to be rewritten in terms of the new `CollisionFrequencies` class.

JaydenR2305:
Resolves #1699 

This pull request moves `plasmapy.formulary.collisons.fundamental_ion_collision_freq` and `plasmapy.formulary.collisons.fundamental_electron_collision_freq` to `CollisionFrequencies`.

      docstrings.


plasmapy/formulary/collisions.py
```diff
@@ -2288,6 +2294,27 @@ def Lorentz_collision_frequency(self):
...
+    def v_i(self):
+        mu = self.test_particle.mass / m_p
+
+        return (
+            4.8e-8
+            * self.n_a.to(u.cm**-3)
+            * self.Coulomb_log
+            * (self.T_a * k_B).to(u.eV) ** -1.5
+            * mu**-0.5
+        ).value * u.Hz
```
JaydenR2305, line 2316: Kids, don't try this at home

Is there any other way to do this?

JaydenR2305:
Also, should the fundamental frequencies be calculated using the NRL equivalent or using the current method?

plasmapy/formulary/collisions.py
pheuer: Ah, I see, is this from the NRL formulary that has the constant out front with the units? In that case I would do 

```python
4.8e-8
            * self.n_a.to(u.cm**-3).value
            * self.Coulomb_log.value
            * (self.T_a * k_B).to(u.eV).value ** -1.5
            * mu**-0.5
        )* u.Hz
```
Which should be equivalent to what you're doing although I guess I prefer stripping the units on each step?

JaydenR2305: Yeah we'll save this fun part for the tests 👍

JaydenR2305:
There seems to have been some discrepancy between the documentation for v_i and how it was actually calculated. The old implementation seems to have sqrt(8 / pi) whereas the documentation would suggest 8 * sqrt(1/pi).

I believe I recall reading somewhere that the formulary has some separate definition for v_i ? Which would explain why the value test I wrote for it is now failing

JaydenR2305:
> There seems to have been some discrepancy between the documentation for v_i and how it was actually calculated. The old implementation seems to have sqrt(8 / pi) whereas the documentation would suggest 8 * sqrt(1/pi).
> 
> 
> 
> I believe I recall reading somewhere that the NRL formulary has some separate definition for v_i ? Which would explain why the value test I wrote for it is now failing

@pheuer any clue what's going on here?

pheuer:
From Braginskii (expressions in CGS units, note tau = 1/freq): 

image

From Callan (expressions in MKS units)
image
^ This expression agrees with the one in Braginskii 
image
^ v0 appears to be the Lorentz collision frequency (see Eq. 2.86 in Callan, and the following pages which closely correspond to the collisions sections of the NRL formulary). So, I would expect this frequency to be 4/(3 sqrt(pi) * the Lorentz frequency. 
image
^ This is also consistent with Braginskii (once you account for the 4piepsilon0 factor from CGS->MKS)

The expressions in the NRL formulary are
image

The one for nu_e is identical to Callan when you take into account that Callan is MKS units (so n is in m^-3 rather than cm^-3 in NRL formulary, so you get a factor of 1E6)

I agree that the NRL formulary for nu_i seems wrong...I think the constant should be 6.8E-8, not 4.8E-8? Callan says that it should be nu_e / sqrt(m_p/m_e) = nu_e / sqrt(1836) so I think this formula in the NRL formulary may be incorrect? Try the test with 6.8e-8.

pheuer:
More generally, I object to calling this the 'fundamental collision frequency'. This is a very specific case for Maxwellian populations with Ti <~ Te. That may be a common case in fusion plasmas,  but it's by no means universally 'fundamental'. Users may be misguided into using this collision frequency in other contexts. I suggest: 

1) Renaming these "maxwellian_avg_ei_collision_freq" and "maxwellian_avg_ii_collision_freq"
2) Making the assumptions much more clear in the docstrings. Namely, 
a) Both populations are Maxwellian
b) Ti <~ Te
c) Any others I'm missing?

@namurphy what's your opinion.

JaydenR2305:
> I agree that the NRL formulary for nu_i seems wrong...I think the constant should be 6.8E-8, not 4.8E-8? Callan says that it should be nu_e / sqrt(m_p/m_e) = nu_e / sqrt(1836) so I think this formula in the NRL formulary may be incorrect? Try the test with 6.8e-8.
>

Oh nice, this makes the tests pass. I seem to have bad luck in programming these special cases from the formulary

> More generally, I object to calling this the 'fundamental collision frequency'. This is a very specific case for Maxwellian populations with Ti <~ Te. That may be a common case in fusion plasmas, but it's by no means universally 'fundamental'. Users may be misguided into using this collision frequency in other contexts. I suggest:
> 

Yeah that makes sense. The current documentation also specifies that the species are "slowly flowing"

JaydenR2305:
> So maybe you were just missing a factor of sqrt(2)?
That seems to get it working as well 👍

pheuer:
> That seems to get it working as well 👍

That's because 6.8/1.4 = 4.8 ;)

JaydenR2305:
> > That seems to get it working as well 👍
> 
> That's because 6.8/sqrt(2) = 4.8 ;)

🤯

changelog/1703.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Moved ``plasmapy.formulary.collisions.fundamental_electron_collision_freq`` to
```
pheuer, line 1: I would say "deprecates .... and replaces it with ...." since the function isn't just being directly moved but rather replaced.  Also, reference the specific name of the attribute that is replacing it since we will probably be changing the name.

plasmapy/formulary/collisions.py
```diff
@@ -1123,6 +1123,13 @@ def fundamental_electron_collision_freq(
     Average momentum relaxation rate for a slowly flowing Maxwellian
     distribution of electrons.
 
+    .. note::
```
pheuer, line 1126: v_e is definitely too simple of a name for this. I think `Maxwellian_averaged_ei_collision_freq` is the shortest accurate name I think of

plasmapy/formulary/collisions.py
```diff
@@ -2300,6 +2321,93 @@ def Lorentz_collision_frequency(self):
...
 
+    @cached_property
+    def v_e(self):
+        r"""Average momentum relaxation rate for a slowly flowing Maxwellian
+        distribution of electrons.
+
+        In order to use this attribute, ``test_particle`` must be an electron
+        and ``field_particle`` must be an ion.
+
+        :cite:t:`braginskii:1965` provides a derivation of this as an
```
pheuer, line 2332: I'd cite Callen here - Braginskii doesn't really derive this I don't think (I can send you that paper if you really feel like diving into theoretical plasma physics from the 1960's...)

plasmapy/formulary/collisions.py
```diff
@@ -2300,6 +2321,93 @@ def Lorentz_collision_frequency(self):
             / (self.test_particle.mass**2 * self.v_a**3)
         ).to(u.Hz)
 
+    @cached_property
+    def v_e(self):
+        r"""Average momentum relaxation rate for a slowly flowing Maxwellian
```
pheuer, line 2326: I think it's important to state that "flowing" is here relative to another population of stationary ions (and therefore this is electron-ion collisions, which are dominant usually). 

Also, we should enforce the meeting of 'slowly' here with a physics error. Callen defines a parameter V_{Tss'} (Eq. 2.113) that  |Vs - Vs'| << V_{Tss'}, so we should raise a physics error if that is not true. 

Also, this whole model might only make sense when you have a drifting Maxwellian, e.g. you use `v_a` as input not `T_a`. I wonder if we should store T_a and v_a (as entered, so one as None) so that in this case we can raise an error if T_a was used. Elsewhere we can make _T_a  and _v_a private properties that do the conversion where allowable (like in the main collision frequencies)?

plasmapy/formulary/collisions.py
```diff
@@ -2300,6 +2321,93 @@ def Lorentz_collision_frequency(self):
...
+                "Please specify an electron-ion interaction to use the v_e attribute"
+            )
+
+        coeff = 4 / (np.sqrt(np.pi) * 3)
+
+        return coeff * self.Lorentz_collision_frequency
+
+    @cached_property
+    def v_i(self):
+        r"""Average momentum relaxation rate for a slowly flowing Maxwellian
```
pheuer, line 2371: Same comments as above, except here these are ion-ion collisions (collisions with ions generally dominate because they are more massive).

pheuer (commented):
Just a few comments - for such simple functions the physics here definitely requires us to tread carefully!

plasmapy/formulary/collisions.py
JaydenR2305: Sure, it's definitely a good idea to enforce that definition

For calculating v_a, is this acceptable? 
v_a = (2 * k_B * self.T_a / self.test_particle.mass)**0.5

pheuer: So, in the general collision frequencies yes, but that only really makes sense when the two populations have no bulk motion relative to one another (so their relative velocity is just the thermal velocity). In this case they are explicitly flowing relative to one another, so I don't think it makes sense to calculate v_a from T_a, because the bulk flow velocity is totally different from the thermal velocity. It's like the thermal velocity is the movement in the plasma CM frame, but then it can also be drifting. Does that make sense?

JaydenR2305: Don't we need T_a for calculating V_Tss' ?

JaydenR2305:
Why doesn't it like my documentation

pheuer:
@StanczakDominik Any what's up with this documentation error? 

```
/home/runner/work/PlasmaPy/PlasmaPy/docs/whatsnew/index.rst:11: WARNING: toctree contains reference to document 'whatsnew/dev' that doesn't have a title: no link will be generated
```

@JaydenR2305 try merging main to make sure it isn't a bug that's been fixed already.

JaydenR2305:
@pheuer I think it has more to do with
`/home/runner/work/PlasmaPy/PlasmaPy/docs/whatsnew/dev.rst:3: CRITICAL: render_fragments() got multiple values for argument 'top_underline'`
I tried removing dev.rst from the index.rst in whatsnew but it still fails

pheuer:
>The issue started popping up ~20 hours ago, which correlates with a new towncrier release. Coincidence?

Nice detective work! If that's the case, try changing the line in requirements/docs.txt 

```
towncrier >= 19.2.0
```
to 
```
towncrier == 21.9.0
```
Since it looks like 22.8.0 is the new version that dropped last night.

Or (not sure if this works...)
```
towncrier >= 19.2.0, <= 21.9.0
```

plasmapy/formulary/collisions.py
pheuer: Yes, we do, so we would need v_a and T_a in this case (because they're not the same thing if v_a is not just the thermal velocity associated with T_a)

I think we should allow the user to set either or both?

plasmapy/formulary/collisions.py
```diff
@@ -1343,6 +1369,18 @@ def fundamental_ion_collision_freq(
     ~plasmapy.formulary.collisions.collision_frequency
     ~plasmapy.formulary.collisions.fundamental_electron_collision_freq
     """
+
+    deprecated(
+        since="0.8.0",
```
pheuer, line 1374: same

plasmapy/formulary/collisions.py
```diff
@@ -2106,14 +2146,18 @@ def __init__(
             :math:`μ` is the reduced mass. Cannot be negative.
 
         T_a : `~astropy.units.Quantity`, optional
-            The temperature of the test species. Only necessary if :math:`v_a` is not provided.
+            The temperature of the test particles. Only necessary if :math:`v_a` is not provided.
 
-        n_b : `~astropy.units.Quantity`
-            The number density of the background field particles in units convertible to :math:`\frac{1}{m^{3}}`
+        n_a : `~astropy.units.Quantity`
```
pheuer, line 2151: Oh, good catch, it does look like that's the test particle density doesn't it...

I have a feeling that, as we add more cases, explaining which of these parameters are required will get cumbersome in the docs. So, I suggest dropping it here, adding the required parameters to the docstring of each attribute, and then using errors to ensure that the right parameters are set whenever an attribute is called.

plasmapy/formulary/collisions.py
```diff
@@ -2196,10 +2240,11 @@ def __init__(
...
+        # _v_a and _T_a denote attributes only to be used in cases in which a conversion between
+        # temperature and velocity is acceptable based on the assumptions of a collision frequency
+
         if v_a is None:
-            if T_a is not None:
-                v_a = thermal_speed(T_a, test_particle)
-            else:
+            if T_a is None:
                 raise ValueError("Please specify either v_a or T_a.")
         elif T_a is not None:
```
pheuer, line 2249: I think we should change it so both can be specified -as you said we'll need that for v_ss' and more generally for any case of a drifting Maxwellian

plasmapy/formulary/collisions.py
```diff
@@ -2214,13 +2259,16 @@ def __init__(
         self.test_particle = test_particle
         self.field_particle = field_particle
         self.v_a = v_a
-        self.T_a = (
+        self._v_a = v_a if v_a is not None else thermal_speed(T_a, test_particle)
+        self.T_a = T_a
```
pheuer, line 2263: Add a comment about what _v_a and _T_a are. 

It seems T_a and v_a are the input directly as the user gave it, and the private versions are converted using thermal speed? To avoid confusion maybe the private ones should just be variables and re-computed wherever they are needed. 

Actually, maybe the solution is to distinguish between v_a and v_T_a everywhere, as you have done for v_T_b in the mean thermal velocity function! There are some cases where they are the same, but generally the distinction is important.

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+    @cached_property
+    def _mean_thermal_velocity(self):
+        """
+        Parameter used in enforcing the definition of "slowly flowing" Maxwellian
+        particles. See Callen (Eq. 2.113).
+        """
+
+        v_T_b = thermal_speed(self.T_b, self.field_particle)
+
+        return (self.v_a**2 + v_T_b**2) ** 0.5
```
pheuer, line 2360: Update so this is v_T_a and v_T_b, since both are thermal velocities

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+        and ``field_particle`` must be an ion.
+
+        This function assumes that both populations are Maxwellian, and :math:`T_{i} \lesssim T_{e}`
+
+        :cite:t:`callen:unpublished` provides a derivation of this as an
+        average collision frequency between electrons and ions for a
+        Maxwellian distribution. It is thus a special case of the collision
+        frequency with an averaging factor, and is on many occasions
+        in transport theory the most relevant collision frequency that has
+        to be considered. It is heavily related to diffusion and resistivity
```
pheuer, line 2386:
```suggestion
        to be considered. It commonly occurs in relation to diffusion and resistivity
```

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+        in plasmas.
+
+        Raises
+        ------
+        `~plasmapy.utils.exceptions.PhysicsError`
+            The specified species aren't slowly flowing
```
pheuer, line 2392:
```suggestion
            The test particles are not 'slowly flowing' relative to the field particles (see notes)
```

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+
+        `ValueError`
+            If the specified interaction isn't electron-ion, or ``n_a`` isn't specified.
```
pheuer, line 2395:
```suggestion
            If the specified interaction isn't electron-ion, or the parameter ``n_a`` isn't provided.
```

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+        >>> T_b = 1e3 * u.eV
+        >>> Coulomb_log = 10 * u.dimensionless_unscaled
+        >>> electron_ion_collisions = CollisionFrequencies(
+        ...     "e-", "Na+", n_a=n_a, T_a=T_a, n_b=n_b, T_b=T_b, Coulomb_log=Coulomb_log
+        ... )
+        >>> electron_ion_collisions.Maxwellian_avg_ei_collision_freq
+        <Quantity 2906316911556553.5 Hz>
+        """
+
+        if self.v_a is None:
```
pheuer, line 2412: Personal opinion - it gets annoying when a function tells you it needs something, you add it, and then a new error pops up. Instead, I would check for each of these, add the error messages to a list if the conditions fail, and then if the list is not empty raise an error and print the whole list.

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+            )
+
+        if not self.n_a:
+            raise ValueError(
+                "Please specify the test particle number density to use the Maxwellian_avg_ei_collision_freq attribute"
+            )
+
+        if not self._is_slowly_flowing:
+            raise PhysicsError(
+                "This frequency is only defined for slowly flowing species."
```
pheuer, line 2429: Maybe say "(see documentation for further details)"

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+
+        :cite:t:`callen:unpublished` provides a derivation of this as an
+        average collision frequency between ions and ions for a Maxwellian
+        distribution. It is thus a special case of the collision frequency
+        with an averaging factor.
+
+        Raises
+        ------
+        `~plasmapy.utils.exceptions.PhysicsError`
+            The specified species aren't slowly flowing
```
pheuer, line 2454: Same comments as above

plasmapy/formulary/collisions.py
```diff
@@ -2297,9 +2345,156 @@ def Lorentz_collision_frequency(self):
...
+        >>> T_b = 1e3 * u.eV
+        >>> Coulomb_log = 10 * u.dimensionless_unscaled
+        >>> ion_ion_collisions = CollisionFrequencies(
+        ...     "Na+", "Na+", n_a=n_a, T_a=T_a, n_b=n_b, T_b=T_b, Coulomb_log=Coulomb_log
+        ... )
+        >>> ion_ion_collisions.Maxwellian_avg_ii_collision_freq
+        <Quantity 79364412.21510696 Hz>
+        """
+
+        if self.v_a is None:
```
pheuer, line 2474: Same comment re this structure

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -1893,6 +1893,20 @@ def get_limit_value(interaction_type, limit_type, cases):
 
         return limit_values
 
+    @staticmethod
+    def get_fundamental_frequency(species, n, v_a, Coulomb_log):
+        # This special case for computing the fundamental frequencies comes from page 33 of the NRL Formulary
```
pheuer, line 1898: Make this a docstring 

Also add that that formula in the NRL formulary is a special case of the Maxwellian_avg_##_collision_freq (I think in the limit that T_a ~ T_b?)

plasmapy/formulary/tests/test_collisions.py
```diff
@@ -2102,6 +2116,50 @@ def test_limit_values(
                 calculated_limit_value, expected_limit_value, rtol=0.05, atol=0
             )
 
+    @pytest.mark.parametrize(
+        "frequency_to_test, constructor_keyword_arguments",
+        [
+            (
+                "Maxwellian_avg_ei_collision_freq",
+                {
+                    "test_particle": Particle("e-"),
```
pheuer, line 2125: Add a test with Z>1 ions

pheuer (commented):
Here's some comments!

plasmapy/formulary/collisions.py
JaydenR2305: Good call on converting _v_a to v_T_a, that definitely clarifies things. Do you have any suggestions for a new name for _T_a?

plasmapy/formulary/tests/test_collisions.py
JaydenR2305: The formulary specifies that the limiting cases are for Z=1, is there some factor of Z I can multiply the expression by to get an expression for species Z > 1?

plasmapy/formulary/collisions.py
JaydenR2305: Good idea! I was able to kill two birds with one stone by using the new `_validate_arguments` decorator, it'll give you a list of the parameters that you didn't specify and also removes the need for a lot of the `ValueError` if statements

plasmapy/formulary/collisions.py
```diff
@@ -2104,6 +2104,56 @@ def coupling_parameter(
...
+                if getattr(self, mutually_exclusive_argument) is not None:
+                    number_of_both_or_either_provided += 1
+
+            if number_of_both_or_either_provided == 0:
+                arguments_not_provided.append("/".join(required_argument))
+
+    return arguments_not_provided
+
+
+def _validate_arguments(expected_arguments):
```
JaydenR2305, line 2132: Should I move this to somewhere in the utilities modules?

plasmapy/formulary/tests/test_collisions.py
pheuer: Oh oops forgot about that, yep ok.

plasmapy/formulary/collisions.py
pheuer: Definitely more general than this file on collisions - is there nothing similar in the decorators file already? If not, my vote is to add them there. Not sure if they need separate tests there?

JaydenR2305: The current validation decorators only work for functions, so I figured it was probably more simple to write a new decorator for attributes provided during class instantiation as opposed to trying to expand the functionality of the existing ones

pheuer: If it does something similar to an existing one for functions, maybe this one should have the same name but with '_class' at the end?

plasmapy/formulary/collisions.py
pheuer: Where do you use _T_a? I don't see that anymore. 

I like what you have now with _v_a and v_T_a. All I would do is add a switch in the following functions: 

Lorentz_collision_frequency
x

that uses v_a if provided but otherwise uses v_T_a (maybe add a warning that prints out saying you are assuming the relative velocity between the particles is the test particle thermal speed)? 


Not in `mean_thermal_velocity` because there you explicitly want v_T_a even if v_a != 0. So in that function, if T_a is not set then an error should be raised.

plasmapy/formulary/collisions.py
```diff
@@ -2214,34 +2305,51 @@ def __init__(
...
             else Coulomb_log * u.dimensionless_unscaled
         )
 
+    @cached_property
+    def _v_T_a(self):
+        """
+        The thermal velocity of the test particle species.
+        """
+
+        return (
```
pheuer, line 2324: I don't think you should have this if/else statement, because there can be a v_T_a even if a separate v_a is not set. 

If the test particles have a drift velocity, then their velocity is 

v_a = v_drift + v_T_a

if there is no drift, then 

v_a = v_T_a

What is confusing is that if the thermal motion is small compared to the drift motion (e.g. v_T_a << v_drift) for example in the fast limit of the original four frequencies we have 

v_a = v_drift 

So I think our input v_a from the user should really be interpreted as v_drift. If no T_a is set, then we should assume it's small and v_a = v_drift. If both v_drift and T_a are given, we compute v_T_a and add the velocities together to get v_a. And if only T_a is given, we should set v_a = v_T_a.  Does that make sense?

plasmapy/formulary/collisions.py
```diff
@@ -2214,34 +2305,51 @@ def __init__(
...
         return self.test_particle.mass / self.field_particle.mass
 
     @cached_property
+    @_validate_arguments(expected_arguments=["n_b", ("T_a", "v_a")])
     def momentum_loss(self):
         """
         The momentum loss rate due to collisions.
+
+        In order to access this attribute, ``n_b`` must be specified in addition
+        to either ``T_a`` or ``v_a``.
```
pheuer, line 2341: These are all enforced by your validate_arguments decorator, right? Should all be tested too.

plasmapy/formulary/collisions.py
```diff
@@ -2297,18 +2417,159 @@ def Lorentz_collision_frequency(self):
...
+
+        if not self.test_particle.is_electron or not self.field_particle.is_ion:
+            raise ValueError(
+                "Please specify an electron-ion interaction to use the Maxwellian_avg_ei_collision_freq attribute."
+            )
+
+        if not self._is_slowly_flowing:
+            raise PhysicsError(
+                "This frequency is only defined for slowly flowing species."
+                "(see documentation for further details)"
```
pheuer, line 2496:
```suggestion
                "(see CollisionFrequencies class documentation for further details)"
```

plasmapy/formulary/collisions.py
```diff
@@ -2297,18 +2417,159 @@ def Lorentz_collision_frequency(self):
...
         """
         The ratio of kinetic energy in the test particle to the thermal energy of the field particle.
-        This parameter determines the regime in which the collision falls
-        (see documentation for the `~plasmapy.formulary.collisions.CollisionFrequencies` class for details).
+        This parameter determines the regime in which the collision falls.
+
+        In order to use this attribute, ``T_b`` must be specified in addition to
+        either ``T_a`` or ``v_a``.
+
+        (see documentation for the `~plasmapy.formulary.collisions.CollisionFrequencies` class for details)
```
pheuer, line 2569: This is good, just replicate this line in the previous function

plasmapy/formulary/collisions.py
```diff
@@ -2297,18 +2417,159 @@ def Lorentz_collision_frequency(self):
...
         """
 
-        x = self.field_particle.mass * self.v_a**2 / (2 * k_B.cgs * self.T_b)
+        x = self.field_particle.mass * self._v_T_a**2 / (2 * k_B.cgs * self.T_b)
```
pheuer, line 2572: So in this case this v should be the total relative velocity between the particles, so 

v_a = v_drift + v_T_a

so in this case this should still be v_a

pheuer:
Here's a more general comment repeating the review comment before (about v_a). In summary, I think we should replace "v_a" in the arguments with "v_drift" and then make the changes below 

If the test particles have no drift

v_a = v_T_a

If the thermal motion is small compared to the drift motion (e.g. v_T_a << v_drift) for example in the >fast limit of the original four frequencies we have

v_a = v_drift

But in the case where there is both thermal motion AND a drift velocity, we'd have to be careful. Some particles (in the drifting frame) are moving with the drift velocity, and some against it. I think this is (or is related to) the Maxwellian_avg frequencies. But, for the four main frequencies I'm not sure how to handle this so we should just raise an error if both v_drift and T_a are set.


Breaking it down case by case:
Below, v_T_a is always calculated from T_a

- momentum_loss, transverse_diffusion, parallel_diffusion, energy_loss, x, Lorentz collision frequency: 
If v_drift is set, then v_a = v_drift. Otherwise v_a = v_T_a. I would just raise an error for now if both are specified for this attribute? 

- mean_thermal_velocity: This always has to be v_T_a and v_T_b (since it's the thermal velocities specifically) so raise an error if T_a is not provided. 

- _is_slowly_flowing : This should be v_drift / _mean_thermal_velocity, since we want to measure the 'flow' relative to the thermal motion. 

- Maxwellian_avg_**_collision_freq: No changes other than those in the functions these call.

JaydenR2305:
- [ ] Add new decorator to changelog
- [ ] Distinguish between v_drift, v_a, v_T_a, and where it's acceptable to convert the two former to the latter
- [ ] Finish up attribute error testing

JaydenR2305:
>momentum_loss, transverse_diffusion, parallel_diffusion, energy_loss, x, Lorentz collision frequency:
>If v_drift is set, then v_a = v_drift. Otherwise v_a = v_T_a. I would just raise an error for now if both are specified for this attribute?
>

The issue with this is that the Maxwellian_avg functions use the Lorentz collision frequency method. Maybe we'll make an exception for that method be sure to enforce the requirement of one of v_drift or T_a for x and the other relaxation rate attributes?

JaydenR2305:
I should also probably provide better documentation for `validate_class_attributes`

pheuer:
Ok, so in the Lorentz collision frequency I think `v` is actually either v_drift of v_T_b. The whole derivation is specifically for a single test particle a, so it doesn't make sense to talk about a 'T_a' there. If v_drift is very small, then the relative velocity that dominates is v_T_b and that is the slow limit. 

The Maxwellian rates are derived explicitly assuming a population of species a and a population of species b, both with thermal velocities and maybe a relative drift v_drift (as long as its not too fast). In this version of the Lorentz collision frequency, it looks like v -> v_ss' (see Callen 2.118). That's just fundamentally different than what we had before. 

I wonder if the problem here is that we're actually putting two things together that shouldn't be. Maybe we need two classes, 
"SingleParticleCollisionFrequencies" and "MaxwellianCollisionFrequencies"?  Then those would have arguments:  

"SingleParticleCollisionFrequencies" 
v_a (the test particle velocity)
T_b
...


"MaxwellianCollisionFrequencies"
T_a
T_b
v_drift 
... 

Does that make sense? Sorry for the confusion - I've clearly never delved this deeply into the collision frequencies before!

JaydenR2305:
>I wonder if the problem here is that we're actually putting two things together that shouldn't be. Maybe we need two classes,
>"SingleParticleCollisionFrequencies" and "MaxwellianCollisionFrequencies"?
>

Yeah this is probably the way to go

JaydenR2305:
Actually, I don't think we need a new class. We just need to be more careful about how we calculate the value of v_a

plasmapy/formulary/collisions.py
```diff
@@ -2196,52 +2248,74 @@ def __init__(
...
+    @validate_class_attributes(both_or_either_attributes=[("T_a", "v_drift")])
+    def _v_a(self):
+        """
+        The net velocity of the test particle species due to both drift and thermal motions.
+        If T_a is not specified, then the net velocity is assumed to be dominated by drift.
+        If v_drift is not specified, then the net velocity is assumed to be dominated by thermal motion.
+        """
+
+        if self.T_a is not None and self.v_drift is not None:
+            return self._v_T_a + self.v_drift
```
JaydenR2305, line 2290: This should be V_T_ss' as opposed to just a straight-up addition, right?

pheuer: In the Maxwellian ones it looks like yes, but in the original ones it should be either v_drift or v_T_b (or maybe we should be taking the mean of those too?))

JaydenR2305:
Gotcha, it'll probably simplify things a bit too since we don't have all the confusing attribute checking. I'll keep the decorator there though just in case someone needs it later 👍 (it also seems to improve coverage quite a bit??)

JaydenR2305:
looks like it's that time of day when `chemical_potential` wants to randomly break

changelog/1676.feature.rst
```diff
@@ -1,3 +1,3 @@
-Added the `~plasmapy.formulary.collisions.CollisionFrequencies` class. This class can be used to calculate collision frequencies for two interacting plasmas.
+Added the `~plasmapy.formulary.collisions.SingleParticleCollisionFrequencies` class. This class can be used to calculate collision frequencies for two interacting plasmas.
```
pheuer, line 1: I don't think we should edit the existing changelog - just add lines to the changelog for this PR saying that we renamed the function. We want the changelogs to correspond to what was actually done in that PR for bookkeeping purposes.

plasmapy/formulary/collisions.py
```diff
@@ -2066,31 +2109,29 @@ def coupling_parameter(
...
 
 
-class CollisionFrequencies:
+class SingleParticleCollisionFrequencies:
     @validate_quantities(
-        v_a={"none_shall_pass": True, "can_be_negative": False},
-        T_a={
-            "none_shall_pass": True,
+        v_a={"can_be_negative": False},
+        n_a={"can_be_negative": False},
```
pheuer, line 2115: Since we're removing n_a, remove it from the validator too.

plasmapy/formulary/collisions.py
```diff
@@ -2100,29 +2141,24 @@ def __init__(
...
-
-        T_b : `~astropy.units.Quantity`
+        T_b : `~astropy.units.Quantity`, optional
             The temperature of the background field particles in units convertible to degrees Kelvin.
 
+        n_b : `~astropy.units.Quantity`, optional
+            The number density of the background field particles in units convertible to :math:`\frac{1}{m^{3}}`.
+
         Coulomb_log : `~astropy.units.Quantity`
             The value of the Coulomb logarithm evaluated for the two interacting particles.
```
pheuer, line 2156: "two interacting particles" seems wrong since it's a test particle interacting with a _population_ of field particles (more than 1)

```suggestion
            The value of the Coulomb logarithm for the interaction.
```

plasmapy/formulary/collisions.py
```diff
@@ -2170,21 +2206,14 @@ def __init__(
         Examples
         --------
         >>> import astropy.units as u
-        >>> T_a = 1 * u.eV
+        >>> v_a = 1e5 * u.m / u.s
```
pheuer, line 2209: Make sure to update v_a -> v_drift in these examples

plasmapy/formulary/collisions.py
```diff
@@ -2346,3 +2363,210 @@ def _phi_prime(self):
...
+    def _is_slowly_flowing(self):
+        """
+        Criteria used in determining whether `Maxwellian_avg_ei_collision_freq` and
+        `Maxwellian_avg_ii_collision_freq` can be applied to the specified species.
+        """
+
+        return self.v_drift / self._mean_thermal_velocity < 0.1
+
+    @cached_property
+    def Lorentz_collision_frequency(self):
```
pheuer, line 2437: Potentially confusing that both this and the SingleParticle class have this attribute. Can you modify the docstring to reference Eq. 2.86 in Callen and explain the difference (here we are using the mean thermal velocity instead of the test particle velocity as in the other class)

pheuer (commented):
I know you're still working on this (take your time!) but here are a few comments. Overall I feel much better with these two separated classes - much cleaner and more explicit!

plasmapy/formulary/collisions.py
```diff
@@ -2100,29 +2141,24 @@ def __init__(
...
         Coulomb_log : `~astropy.units.Quantity`
             The value of the Coulomb logarithm evaluated for the two interacting particles.
 
         Raises
         ------
         `ValueError`
-            If both :math:`v_a` and :math:`T_a` are specified, neither
-            :math:`v_a` nor :math:`T_a` are specified, or specified arrays
-            don't have equal size.
+            If specified arrays don't have equal size.
```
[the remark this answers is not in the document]
JaydenR2305: Whoops. The docstrings are a mess at this point since I have changed everything around so many times. I believe only v_drift can be specified for that function

changelog/1676.feature.rst
JaydenR2305: What should I do about the failing documentation build?

JaydenR2305:
@pheuer Is it fine if I just convert references of `CollisionFrequencies` to just the plain text reference in changelogs?

pheuer:
> @pheuer Is it fine if I just convert references of `CollisionFrequencies` to just the plain text reference in changelogs?

You mean since the class no longer exists? Yeah I think thats ok in this case.

plasmapy/formulary/collisions.py
```diff
@@ -2390,6 +2390,17 @@ def __init__(
         r"""Compute collision frequencies between two slowly flowing
         Maxwellian populations.
 
+        The condition of "slowly flowing", outlined by Eq. 2.133 in :cite:t:`callen:unpublished`
```
JaydenR2305, line 2393: Does this sound good ? @pheuer

pheuer: I think so!

pheuer:
@JaydenR2305 I see most of the tests are passing now (except for the changelog, since you're modifying that old one). Let me know whenever you want me to do another review.

JaydenR2305:
Ok. I think I'm ready at this point for another review

pheuer:
> Ok. I think I'm ready for another review at this point

Sounds good - I'll try to take another pass in the next couple days :)

StanczakDominik:
@pheuer if you can sign off on the theory, I can review this from the pure python/design side if you'd like :)

pheuer:
@StanczakDominik It would be great to have your review! I think the theory should be right now.

StanczakDominik:
All right, then I'll take over :)

plasmapy/formulary/collisions.py
```diff
@@ -2066,31 +2105,28 @@ def coupling_parameter(
...
         T_b: u.K,
+        n_b: u.m**-3,
         Coulomb_log: u.dimensionless_unscaled,
     ):
         r"""
-        Compute collision frequencies between test particles (labeled 'a') and field particles (labeled 'b')
+        Compute collision frequencies between test particles (labeled 'a') and field particles (labeled 'b').
 
         Parameters
         ----------
```
pheuer, line 2132: Modify these to `ParticleLike` too

plasmapy/formulary/collisions.py
```diff
@@ -2100,29 +2136,24 @@ def __init__(
         field_particle : |Particle|
             The background particle being interacted with.
 
-        v_a : `~astropy.units.Quantity`, optional
-            The relative speed between particles. If not provided, T_a must be specified and
-            thermal velocity is assumed: :math:`μ v_a^2 \sim 2 k_B T_a` where
+        v_drift : `~astropy.units.Quantity`, optional
+            The relative drift between the test and field particles. If not provided, T_a must be specified and
```
pheuer, line 2140: We got rid of T_a entirely in this function, right?

plasmapy/formulary/collisions.py
```diff
@@ -2284,7 +2295,7 @@ def Lorentz_collision_frequency(self):
         (typically taken as the thermal velocity), and :math:`\ln{Λ}` is the
```
pheuer, line 2295:
```suggestion
         and :math:`\ln{Λ}` is the
```

I think we should drop this part about the thermal velocity here, since for many uses of this function v_drift is going to be set much higher than the thermal velocity.

plasmapy/formulary/collisions.py
```diff
@@ -2346,3 +2358,258 @@ def _phi_prime(self):
...
+        n_a={"can_be_negative": False},
+        T_b={
+            "can_be_negative": False,
+            "equivalencies": u.temperature_energy(),
+        },
+        n_b={"can_be_negative": False},
+    )
+    def __init__(
+        self,
+        test_particle: particles.Particle,
```
pheuer, line 2380: Same "ParticleLike" update here

plasmapy/formulary/collisions.py
```diff
@@ -2346,3 +2358,258 @@ def _phi_prime(self):
...
+        test_particle : |Particle|
+            The test particle streaming through a background of field particles.
+
+        field_particle : |Particle|
+            The background particle being interacted with.
+
+        v_drift : `~astropy.units.Quantity`, optional
+            The relative drift between the test and field particles.
+
+        T_a : `~astropy.units.Quantity`, optional
```
pheuer, line 2415: I don't think any of these should be marked optional, except v_drift?

plasmapy/formulary/collisions.py
pheuer: This still needs to be addressed - I'd add a note to both this method and the one in SingleParticleCollisionFrequencies pointing out that they are different equations! The equation numbers are already different, so careful readers should notice that, but I want to make it more obvious to avoid confusion.

plasmapy/utils/decorators/tests/test_validators.py
```diff
@@ -555,3 +560,76 @@ def test_decorator_func_def(self, mock_vq_class):
                 # reset
                 mock_vq_class.reset_mock()
                 mock_foo.reset_mock()
+
+
+class TestValidateClassAttributes:
```
pheuer, line 565: I don't feel qualified to review this decorator - I suggest moving it to a separate PR and then maybe @namurphy can take a look at it?

pheuer (commented):
As far as I can see, almost done! Here's a few more comments.

pheuer:
@StanczakDominik Do you still want to give this a brief look-over before we merge it?

StanczakDominik:
Yes, sorry, I wanted to get it done over the weekend and didn't predict I'd be feeling off throughout it :( I'll sit down to it now.

pheuer:
> Yes, sorry, I wanted to get it done over the weekend and didn't predict I'd be feeling off throughout it :( I'll sit down to it now.

No problem - take your time and hope you feel better!

plasmapy/utils/decorators/tests/test_validators.py
StanczakDominik: I can definitely sign off on it 😉 The test looks reasonable at least, still digging into the implementation.

plasmapy/utils/decorators/validators.py
```diff
@@ -536,3 +536,82 @@ def foo(arg1):
...
         # `validate_quantities` called as a decorator "sugar-syntax"
         return ValidateQuantities(**validations)
+
+
+def get_attributes_not_provided(
+    self,
+    expected_attributes=None,
+    both_or_either_attributes=None,
+    mutually_exclusive_attributes=None,
+):
```
StanczakDominik, line 546: I'd love to see type hinting here, but it's okay as it is.

plasmapy/utils/decorators/validators.py
```diff
@@ -536,3 +536,82 @@ def foo(arg1):
...
+                raise ValueError(
+                    f"{attribute.__name__} expected the following "
+                    f"additional arguments: {', '.join(arguments_not_provided)}"
+                )
+
+            return attribute(self, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
```
StanczakDominik, line 617: You know, I expected this to be longer!

Honestly looks great; could use some type hints to increase readability, but it's nice as it is.

StanczakDominik (commented):
LGTM 👍🏻! I agree with Peter's questions where they're still unresolved, but on the decorator and the general layout of things, I can definitely sign off.

Consider this a pre-approval 😄

plasmapy/formulary/collisions.py
StanczakDominik: Just to be sure - I think it's fine for the two to have the attribute share the name, but a note making things explicit here would definitely be great.

plasmapy/formulary/collisions.py
```diff
@@ -2066,31 +2105,28 @@ def coupling_parameter(
...
+        T_b={
             "can_be_negative": False,
             "equivalencies": u.temperature_energy(),
         },
         n_b={"can_be_negative": False},
-        T_b={"can_be_negative": False, "equivalencies": u.temperature_energy()},
     )
     def __init__(
         self,
         test_particle: particles.Particle,
```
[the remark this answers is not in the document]
JaydenR2305: Can I still use the `particle_input` decorator? I'm getting issues with the .is_electron component of the tests

plasmapy/utils/decorators/validators.py
JaydenR2305: I'll take that as a compliment haha 

I have added type hints. Good call!

plasmapy/formulary/collisions.py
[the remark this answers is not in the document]
StanczakDominik: Yeah, you should be able to - if it doesn't work, leave it as is and we'll fix it elsewhere (just make sure to note it down as a bug!).

plasmapy/utils/decorators/validators.py
```diff
@@ -540,9 +540,9 @@ def foo(arg1):
 
 def get_attributes_not_provided(
     self,
-    expected_attributes=None,
-    both_or_either_attributes=None,
-    mutually_exclusive_attributes=None,
+    expected_attributes: List[str] = None,
+    both_or_either_attributes: List[Iterable[str]] = None,
+    mutually_exclusive_attributes: List[Iterable[str]] = None,
```
StanczakDominik, line 545: Technically - and purely technically since we aren't `mypy` compliant yet but I'm hoping to improve that soon - these should all be `Optional[List[str]]`, because `None` sure don't look like no list of strings to me 😜. But I'll be fixing those soon, anyway.

plasmapy/formulary/collisions.py
pheuer: @JaydenR2305 Should we make v_drift optional with default value of 0? That's a physically reasonable default.

plasmapy/formulary/collisions.py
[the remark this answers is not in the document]
JaydenR2305: Yeah I seem to be getting an error when using the decorator there...

plasmapy/utils/decorators/validators.py
JaydenR2305: Oh duh. Good catch

plasmapy/formulary/collisions.py
JaydenR2305: I don't see why not

plasmapy/formulary/collisions.py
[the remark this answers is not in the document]
pheuer: Ok, that's strange - let's leave it like this for now, then. But if you could reproduce the bug and open an issue with the error message (tagging @namurphy ) that would be great!

pheuer:
@PlasmaPy/peer-reviewers Git help please! I pushed a few minor changes to @JaydenR2305's fork prior to merging this, but now I'm getting a readthedocs error. It looks like it can't check out the latest commit, although I can see it just fine on the fork? Is there something special you need to do when pushing to someone elses's branch?

StanczakDominik:
Oh wow, that one's new...

https://readthedocs.org/projects/plasmapy/builds/18143144/ :
```
fatal: reference is not a tree: 2c08848c7a29239036703eb67ffee0bbbc47a967
```

StanczakDominik:
@pheuer out of curiosity, how did you do this part?

> I pushed a few minor changes to @JaydenR2305's fork prior to merging this

pheuer:
> @pheuer out of curiosity, how did you do this part?
> 
> > I pushed a few minor changes to @JaydenR2305's fork prior to merging this

I just checked out the fork (by selecting this PR in GitHub Desktop) and then pushed to that.

StanczakDominik:
Okay, sorry, I'm going to have to revert this to [8af3866](https://github.com/PlasmaPy/PlasmaPy/pull/1703/commits/8af38662aca22e4e0af42f5fe010f40ae24be068), I have no idea what's going on here.

StanczakDominik:
This was doubly weird because I couldn't see anything past the merge commit with `gh pr checkout -f 1703`, but when I cloned Jayden's repo separately, I could see those three commits. I have no idea what happened there.

StanczakDominik:
@JaydenR2305 if you pulled Peter's changes while they were alive on those repo, I mean these three:
```
commit 8190d36ae2b10ef30e81259565066b09306f65df (HEAD -> fundamental-collision-frequencies, origin/fundamental-collision-frequencies)
Author: Peter Heuer <...>
Date:   Thu Sep 22 10:32:40 2022 -0400

    Minor changelog updates

commit 2c08848c7a29239036703eb67ffee0bbbc47a967
Author: Peter Heuer <...>
Date:   Thu Sep 22 10:10:56 2022 -0400

    Update Particle -> ParticleLike

commit 1ae35457e5a5cb36be5f70195146aaa732ce3bda
Merge: 8af38662 8ec56daf
Author: Peter Heuer <...>
Date:   Thu Sep 22 10:07:25 2022 -0400

    Merge branch 'main' into pr/1703
```

Then please do `git reset 8af38662`.

Peter, kudos, I think you've just pounded git into submission. How? No idea, but, kudos.

pheuer:
> Okay, sorry, I'm going to have to revert this to [8af3866](https://github.com/PlasmaPy/PlasmaPy/pull/1703/commits/8af38662aca22e4e0af42f5fe010f40ae24be068), I have no idea what's going on here.

No problem - I'm sure I was just using git wrong! 

How do you usually go about pushing to someone else's PR branch like this?

StanczakDominik:
I usually use https://cli.github.com/, then it's just a matter of checking out the branch with `gh pr checkout 1703`, commiting whatever's needed, then `git push` does the rest - nothing else necessary.

I think the merge commit you had in there was *somehow* the culprit, but how... no clue! :(

changelog/1703.breaking.1.rst
```diff
@@ -0,0 +1,5 @@
+Deprecated ``plasmapy.formulary.collisions.fundamental_electron_collision_freq`` and replaced it
```
pheuer, line 1: Update name to SingleParticleCollisionFrequencies

changelog/1705.trivial.rst
```diff
@@ -1 +1 @@
-Minor improvements to `~plasmapy.formulary.collisions.CollisionFrequencies`
```
pheuer, line 1: Add another changelog feature entry for creating the MaxwellianCollisionFrequencies class!

plasmapy/formulary/collisions.py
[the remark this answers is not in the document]
pheuer: This worked for me locally - can you update this branch from main and then try this again with ParticleLike? It's possible you just haven't merged main for a while, since the changes to particle_input were like a week ago.

pheuer (changes requested):
I tried to make these changes myself but apparently I broke git :/ 

If merging in main doesn't solve the issue with ParticleLike, then we can drop it and leave it for later.

PR #1705, opened 2022-08-27
Add check for unequal array lengths to `CollisionFrequencies`

### LINKED ISSUE #1704: Minor changes in CollisionFrequencies
@JaydenR2305 I noticed a couple minor changes I'd like to make for the new `CollisionFrequencies` class

1) Right now the `Coulomb_log` attribute must be an astropy quantity object (u.dimensionless_unscaled). However, it's natural to just enter this as a float. Can you add a line like this: 

```python
if not isinstance(Coulomb_log, u.Quantity):
      Coulomb_log *= u.dimensionless_unscaled

```

2) Right now an  unhandled error is raised if the n_b and v_a/T_a arrays are different sizes. I think there should be a validation check at the beginning of the function making sure they have the same shape. That requirement should also be added to the docstring.

If you open a new PR for that, it should be easy to merge

JaydenR2305:
Resolves #1704

### DIFF
--- a/changelog/1705.trivial.rst
+++ b/changelog/1705.trivial.rst
@@ -1,0 +1,1 @@
+Minor improvements to `~plasmapy.formulary.collisions.CollisionFrequencies`
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -2120,8 +2120,9 @@ class CollisionFrequencies:
         Raises
         ------
         `ValueError`
-            If both :math:`v_a` and :math:`T_a` are specified, or neither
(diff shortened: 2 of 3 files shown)


JaydenR2305:
@pheuer This probably doesn't need a changelog entry

pheuer:
@JaydenR2305 I think it should probably still get a changelog just for recording purposes, but you can use the trivial tag e.g. 

"1705.trivial.rst" 

And just say "Minor improvements to plasmapy.formulary.collisions.CollisionFrequencies"

StanczakDominik (approved):
Good changes, let's get them in! 👍🏻 Thanks!

PR #1706, opened 2022-08-28
Apply `|keyword-only|` substitution in docstrings and reST files

namurphy:
This PR does a quick search and replace in `.py` files for `, keyword-only` → `, |keyword-only|`, where `|keyword-only|` is a substitution defined in `docs/common_links.rst` for ``` :term:`keyword-only` ```.  This PR also applies this substitution in a few `.rst` files. Linking this to the glossary entry will probably be mildly helpful for new users.


StanczakDominik (approved):
Looks mildly good to me! 😉

PR #1713, opened 2022-08-30
Update docstrings in `ParticleList`

### LINKED ISSUE #1712: Add examples for `ParticleList.sort` (e.g., sorting by charge number)
The docstring for `ParticleList.sort` doesn't contain any examples, which makes it harder to use.  It would be helpful to expand this docstring by adding an examples section.

A suggestion from @MKastek:
```Python
sorted(element_list, key=lambda x: atomic_number(x))
```

namurphy:
* Expanded the docstring for `ParticleList.sort`, in particular by adding sections for "Parameters", "Raises", "See Also", and "Examples".  This follows a suggestion from @MKastek.
* Change`ParticleList.sort` to use `super`, as suggested by @rocco8773.
* Update the docstrings of `ParticleList` and many of its methods/attributes.

Closes #1712.

### DIFF
--- a/plasmapy/particles/particle_collections.py
+++ b/plasmapy/particles/particle_collections.py
@@ -20,7 +20,7 @@ from plasmapy.particles.particle_class import (
 
 class ParticleList(collections.UserList):
     """
-    A `list` like collection of particle objects.
+    A `list` like collection of |Particle| and |CustomParticle| objects.
 
     Parameters
     ----------
--- a/plasmapy/particles/particle_collections.py
+++ b/plasmapy/particles/particle_collections.py
@@ -374,16 +374,16 @@ class ParticleList(collections.UserList):
         >>> print(elements)
         ParticleList(['H', 'He', 'Li'])
 
-        We can also create a function that we then pass to ``key``. In
-        this example, we sort by atomic number, mass_number, and
-        charge number using different attributes of |Particle|.
+        We can also create a function to pass to ``key``. In this
+        example, we sort by atomic number and mass number using
+        different attributes of |Particle|.
 
-        >>> def sort_key(ion):
-        ...     return ion.atomic_number, ion.mass_number, ion.charge_number
-        >>> ions = ParticleList(["α", "p+", "He-3 1+", "He-3 0+", "H-1 0+"])
-        >>> ions.sort(key=sort_key)
-        >>> print(ions)
-        ParticleList(['H-1 0+', 'p+', 'He-3 0+', 'He-3 1+', 'He-4 2+'])
+        >>> def sort_key(isotope):
+        ...     return isotope.atomic_number, isotope.mass_number
+        >>> isotopes = ParticleList(["He-3", "T", "H-1", "He-4", "D"])
+        >>> isotopes.sort(key=sort_key)
+        >>> print(isotopes)
+        ParticleList(['H-1', 'D', 'T', 'He-3', 'He-4'])
         """
         if key is None:
             raise TypeError("Unable to sort a ParticleList without a key.")
--- a/plasmapy/particles/particle_collections.py
+++ b/plasmapy/particles/particle_collections.py
@@ -20,15 +20,15 @@ from plasmapy.particles.particle_class import (
 
 class ParticleList(collections.UserList):
     """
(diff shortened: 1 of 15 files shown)


plasmapy/particles/particle_collections.py
```diff
@@ -22,35 +20,34 @@
 
 class ParticleList(collections.UserList):
     """
-    A `list` like collection of
-    `~plasmapy.particles.particle_class.Particle` and/or
-    `~plasmapy.particles.particle_class.CustomParticle` objects.
+    A `list` like collection of particle objects.
```
pheuer, line 23: Why don't you want the links here?

plasmapy/particles/particle_collections.py
```diff
@@ -217,16 +211,24 @@ def data(self) -> List[Union[Particle, CustomParticle]]:
         A `list` containing the particles contained in the
         |ParticleList| instance.
 
-        The `~plasmapy.particles.particle_collections.ParticleList.data`
-        attribute should not be modified directly.
+        .. important::
```
pheuer, line 214: You already sort of enforce this by using a private attribute. You could do one better by adding a __setattr__ method for data that raises an exception?

plasmapy/particles/particle_collections.py
```diff
@@ -292,24 +301,37 @@ def is_category(
     @property
     def charge_number(self) -> np.array:
         """
-        An array of the quantized charges of the particles, as
-        multiples of the elementary charge.
+        The charges of the particles in units of the elementary
+        positive charge.
```
pheuer, line 305: Isn't the elementary charge e defined to be positive?

```suggestion
         charge.
```

plasmapy/particles/particle_collections.py
```diff
@@ -321,16 +343,61 @@ def sort(self, key: Callable = None, reverse: bool = False):
...
+        To sort a |ParticleList| by atomic number, set ``key`` to
+        |atomic_number|.
+
+        >>> from plasmapy.particles import ParticleList, atomic_number
+        >>> elements = ParticleList(["Li", "H", "He"])
+        >>> elements.sort(key=atomic_number)
+        >>> print(elements)
+        ParticleList(['H', 'He', 'Li'])
+
+        We can also create a function that we then pass to ``key``. In
```
pheuer, line 377: I don't quite understand this example yet - how do the three quantities coming from this function define the sorting? Does it sort based off of the first one and then break ties with the next one? I think that could be said a tad more explicitly.

plasmapy/particles/particle_collections.py
```diff
@@ -321,16 +343,61 @@ def sort(self, key: Callable = None, reverse: bool = False):
         """
         Sort the |ParticleList| in-place.
 
-        For more information, refer to the documentation for `list.sort`.
+        Parameters
+        ----------
+        key : callable
+            A function of one argument that is used to extract a
```
pheuer, line 349:
```suggestion
            A function that accepts one |Particle| argument that is used to extract a
```

plasmapy/particles/particle_collections.py
namurphy: Good question. In retrospect, it does seem a better idea to link to `Particle` and `CustomParticle`, doesn't it?

plasmapy/particles/particle_collections.py
namurphy: This one is kind of complicated.  `ParticleList` inherits `data` from `collections.UserList`, so we would need to keep `data` itself public to be consistent with the [Liskov Substitution Principle](https://reflectoring.io/lsp-explained/).  Even if we make `data` a `property`, it is still a `list` so we can do operations like:
```Python
>>> particle_list = ParticleList([])
>>> particle_list.data.append("not a particle")
```
The admonition is to warn against operations like that.

pheuer: Ok, got it. 

I think I've run into this problem before - would be nice if there were an abstract datatype like a list that isn't modifiable except by subclass methods. Like a 'secret list'

pheuer (approved):
Great, otherwise looks like straightforward and good improvements!

namurphy:
Thank you for the review, @pheuer!

PR #1718, opened 2022-08-31
Improve consistency of docstrings, annotations, and signatures in `plasmapy.particles`

namurphy:
This PR is an effort to make the user experience for `plasmapy.particles` more consistent and in-line with our more recent practices.

 - Introduce the term "atom-like"
 - Apply `particle-like` and `atom-like` to type lines in parameters sections of docstrings
 - Make `Z` and `mass_numb` keyword-only in `ionic_symbol` and `particle_symbol`
 - Apply consistent descriptions for `Z` and `mass_numb`
 - Remove the names of the variables that get returned (since they're usually quite redundant with the function name)
 - Update annotations from eons past when it was recommended that we didn't use `typing.Optional` when the default argument is `None`
 - Change ``` `int` ``` to `integer` in type lines (since a dimensionless `Quantity` can be used too)
 - Apply some reST formatting or substitutions
 - Specify defaults in a way that is consistent with the numpydoc standard
 - Apply `ParticleLike` as annotations; this may overlap with #1057 
 - Apply more consistent formatting of docstrings
 - Used `-> typing.NoReturn` instead of `-> typing.None` for return annotations
 - Make docstring formatting a little more consistent
 - Remove `TypeError` from "Raises" sections following recent updates to doc guide in #1667


changelog/1718.breaking.rst
```diff
@@ -0,0 +1,3 @@
+The parameters ``Z`` and ``mass_numb`` are now |keyword-only| for
+`~plasmapy.particles.symbols.ionic_symbol` and
+`~plasmapy.particles.symbols.particle_symbol`.
```
StanczakDominik, line 3: A lot of breaking changes in these changelogs, recently. I'm actually kind of happy to see it, but from a semver standpoint, it's... 😅

docs/glossary.rst
```diff
@@ -22,13 +22,24 @@ Glossary
...
       An abbreviation for `positional arguments`_.
 
+   atom-like
+      A |particle-like| `object` is **atom-like** if it is or could be
+      cast into:
+
+      * A |Particle| representing an element, isotope, or ion; or
+      * A |ParticleList| including only elements, isotopes, or ions.
+
+      For full details, please refer to |particle-like|.
```
StanczakDominik, line 32: When is a particle *not* `atom-like`?

plasmapy/particles/ionization_state.py
```diff
@@ -34,14 +34,14 @@ class IonicLevel:
...
     ----------
-    ion: `~plasmapy.particles.particle_class.ParticleLike`
+    ion : |atom-like|
         The ion for the corresponding ionic fraction.
 
-    ionic_fraction: real number between 0 and 1, optional
+    ionic_fraction : real number, optional
         The fraction of an element or isotope that is at this ionization
-        level.
+        level. Must be between 0 and 1.
```
StanczakDominik, line 42: Non-inclusive, I suspect?

plasmapy/particles/ionization_state.py
```diff
@@ -123,7 +123,7 @@ def ionic_fraction(self) -> Real:
         An ionic fraction must be in the interval :math:`[0, 1]`.
 
         If no ionic fraction is specified, then this attribute will be
-        assigned the value of `~numpy.nan`.
+        assigned the value of |nan|.
```
StanczakDominik, line 126: There's room for [a delicious glossary easter egg](https://en.wikipedia.org/wiki/Naan) here.

StanczakDominik (commented):
Partial review for now, because I managed to trick myself into wanting to eat breakfast earlier.

- [ ] atomic
- [ ] nuclear
- [ ] particle_class
- [ ] symbols

changelog/1718.breaking.rst
namurphy: Agreed!

plasmapy/particles/ionization_state.py
namurphy: Good catch — I added a clarification.

docs/glossary.rst
```diff
@@ -29,7 +29,8 @@ Glossary
       * A |Particle| representing an element, isotope, or ion; or
       * A |ParticleList| including only elements, isotopes, or ions.
 
-      For full details, please refer to |particle-like|.
+      Fundamental particles are not atom-like, except for protons. For
+      full details, please refer to |particle-like|.
```
StanczakDominik, line 33: I still think explicit examples would be great here...

plasmapy/particles/ionization_state.py
StanczakDominik: Oh right, it would be inclusive because you want to be able to represent a pure state. Makes sense 👍🏻

StanczakDominik (approved):
Looks good all in all, except that AtomLike term is still a tad vague to me.

namurphy:
I just updated the description for [atom-like](https://plasmapy--1718.org.readthedocs.build/en/1718/glossary.html#term-atom-like) with some examples.  For information on what constitutes a valid representation of a particle, I wanted to make sure to refer users to the definition of [particle-like](https://plasmapy--1718.org.readthedocs.build/en/1718/glossary.html#term-particle-like) so that we don't end up repeating the same information in multiple places.

StanczakDominik:
Okay, one last follow-up question. Is a single `He-4`, not in a list with an electron, atom-like?

docs/glossary.rst
StanczakDominik: My bad, should have communicated that more clearly earlier!

StanczakDominik:
All right, let's get it in! Sorry for the, er, iterative approach in this review 😅

PR #1721, opened 2022-09-01
Return `False` when comparing an ionization object to an object of a different type

namurphy:
This PR changes `IonicLevel.__eq__`, `IonizationState.__eq__`, and `IonizationStateCollection.__eq__` so that an equality comparison with another type of object doesn't raise an exception but rather returns `False`.  This change is to better match the typical/expected behavior of Python objects.


StanczakDominik (approved):

StanczakDominik:
I vaguely recall having had issues with the original behavior before, so I'm happy to see this!

PR #1726, opened 2022-09-03
Resolve numerical instability in `chemical_potential`

JaydenR2305:
Currently, for reasonable electron densities (~1e10 cm^-3 - 1e20 cm^-3), `chemical_potential` returns a number extremely close to zero (~1e-16). This behavior is reasonable and expected. However, there are some occasions where the current least squares regression method will return an abnormally large value, demonstrating numerical instability. This instability can sometimes cause the docstring test to fail

I propose we use some alternative to a least squares method. The method tested in the figures is the [Broyden–Fletcher–Goldfarb–Shanno algorithm](https://en.wikipedia.org/wiki/Broyden%E2%80%93Fletcher%E2%80%93Goldfarb%E2%80%93Shanno_algorithm). The figures were generated at 11000 Kelvin using the two different methods to calculate the chemical potential, and the resulting residual.

For densities greater than 1e20 cm^-3,  the graphs produce identical results

The first regime is where the chemical potential is ~0. The algorithms generally agree on zero but sometimes the least squares method produces unexpected results
Figure_1

The residuals for both functions are the same suggesting they have an approximately equal error
Figure_2

Both algorithms produce the same results for electron densities greater than 1e20
Figure_3

      docstrings.


rocco8773:
My suspicion is that you're running into the same issue discussed in https://github.com/astropy/astropy/issues/9721.

I'd suggest exploring something like...

```python
@validate_quantities(
    n_e={"can_be_negative": False, "pass_equivalent_units": True},
)
def chemical_potential(n_e: u.m**-3, T: u.K) -> u.dimensionless_unscaled:
    # deBroglie wavelength
    lambdaDB = thermal_deBroglie_wavelength(T)
    # degeneracy parameter
    degen = (np.cbrt(n_e) * lambdaDB).to(u.dimensionless_unscaled)**3

    ...
```

Using `"pass_equivalent_units"` will allow `n_e` to NOT be converted to `u.m`, `n_e` will stay in the units it originally was passed with.  This means, once you get to the `degen` line and multiply `np.cbrt(n_e) * lambdaDB` the `.to()` method only has to convert one ratio of `u.cm/u.m` instead of a power of three.  This should reduce the floating point arithmetic error and make the calculation more stable.

rocco8773:
This needs to be merged before #1727

JaydenR2305:
Looks like 5.0 still uses `asscalar` somewhere..

JaydenR2305:
> My suspicion is that you're running into the same issue discussed in [astropy/astropy#9721](https://github.com/astropy/astropy/issues/9721).
> 
> I'd suggest exploring something like...
> 
> ```python
> @validate_quantities(
>     n_e={"can_be_negative": False, "pass_equivalent_units": True},
>     T={"can_be_negative": False, "equivalencies": u.temperature_energy()},
> )
> def chemical_potential(n_e: u.m**-3, T: u.K) -> u.dimensionless_unscaled:
>     # deBroglie wavelength
>     lambdaDB = thermal_deBroglie_wavelength(T)
>     # degeneracy parameter
>     degen = (np.cbrt(n_e) * lambdaDB).to(u.dimensionless_unscaled)**3
> 
>     ...
> ```
> 
> Using `"pass_equivalent_units"` will allow `n_e` to NOT be converted to `u.m`, `n_e` will stay in the units it originally was passed with. This means, once you get to the `degen` line and multiply `np.cbrt(n_e) * lambdaDB` the `.to()` method only has to convert one ratio of `u.cm/u.m` instead of a power of three. This should reduce the floating point arithmetic error and make the calculation more stable.

This seems to be producing identical results
Figure_4

rocco8773:
> Looks like 5.0 still uses `asscalar` somewhere..

You are correct, see my comment https://github.com/PlasmaPy/PlasmaPy/pull/1727#issuecomment-1238766243.

rocco8773 (changes requested):
In the docstring, can you make the executive decision to either consistently use the subscript $_a$ or drop it all together?  I think either approach is valid.  I just want it to be consistent across the docstring, since switching back-and-forth and lead to confusion for users.  If you choose to keep the subscript, then can you update the sentence explaining the equation to read something like:

"... where $\chi_a$ is the degeneracy parameter **of particle 'a'**..."

If we keep the subscript, then we at least need to explain what it means.

rocco8773 (changes requested):
Do you know which section in Bontiz discusses the chemical potential?  If so, can you included in the first line of the docstring Notes section?

JaydenR2305:
> Do you know which section in Bontiz discusses the chemical potential? If so, can you included in the first line of the docstring Notes section?

Good call! I added the specific equation number from Bonitz and improved the overall explanation. The documentation was somewhat vague on how the calculation is performed so I tried to make things more explicit

changelog/1726.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Change `~plasmapy.formulary.quantum.chemical_potential` to use the Broyden Fletcher Goldfarb Shanno algorithm to
+implicitly solve for beta.
```
namurphy, line 2:
```suggestion
Changed `~plasmapy.formulary.quantum.chemical_potential` to use the 
Broyden-Fletcher-Goldfarb-Shanno algorithm to implicitly solve for beta.
```

namurphy (approved):
LGTM!  It looks like you addressed all of @rocco8773's comments, and I only have a very minor suggestion for the changelog entry.  I'm happy to get this merged soon.  Thank you for doing this!

lemmatum:
Cool, I've never heard of the BFGS algorithm. I have some reading up to do!
My only comment is it'd be good to add a test where the chemical potential is non-zero. Otherwise I think the changes look good.

Thanks for digging into this!

plasmapy/formulary/tests/test_quantum.py
```diff
@@ -156,7 +156,7 @@ def setup_class(self):
         self.n_e = 1e20 * u.cm**-3
         self.n_e_fail = 1e23 * u.cm**-3
         self.T = 11604 * u.K
-        self.True1 = 6.133671348607095e-11
+        self.True1 = 0
```
lemmatum, line 159: Add another test where the result is non-zero.

lemmatum (approved):

changelog/1726.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Changed `~plasmapy.formulary.quantum.chemical_potential` to use the
+Broyden-Fletcher-Goldfarb-Shanno algorithm to implicitly solve for beta.
```
rocco8773, line 2:
```suggestion
Changed `~plasmapy.formulary.quantum.chemical_potential` to use the
Broyden-Fletcher-Goldfarb-Shanno algorithm to implicitly solve for 
the β(\ :math:`= 1/(k_B T)`\ ) parameter.
```

Only so beta is not mistaken for the plasma beta, which would likely be common in this community.

plasmapy/formulary/quantum.py
```diff
@@ -437,56 +437,55 @@ def chemical_potential(n_e: u.m**-3, T: u.K) -> u.dimensionless_unscaled:
...
     The definition for the ideal chemical potential is implicit, so it must
     be obtained numerically by solving for the Fermi integral for values
     of chemical potential approaching the degeneracy parameter. Since values
     returned from the `~plasmapy.formulary.mathematics.Fermi_integral`
-    are complex, a nonlinear Levenberg-Marquardt least squares method is
-    used to iteratively approach a value of :math:`μ` which minimizes
-    :math:`I_{1/2}(β μ_a^{ideal}) - χ_a`
+    are complex, the Broyden–Fletcher–Goldfarb–Shanno algorithm is
+    used to iteratively approach a value of :math:`μ^{ideal}` which minimizes
+    :math:`\lvert I_{1/2}(β μ^{ideal}) - χ \rvert`.
```
rocco8773, line 456: I'm thinking as a follow up PR, it would be good to break out the code for the ideal chemical potential into its own function.  @JaydenR2305 What do you think about this and would you be willing to do it?  Again, this would be a followup PR and I'm not suggesting to do it in this PR.

plasmapy/formulary/tests/test_quantum.py
```diff
@@ -156,15 +156,25 @@ def setup_class(self):
...
         """
-        methodVal = chemical_potential(self.n_e, self.T)
-        testTrue = u.isclose(methodVal, self.True1, rtol=1e-16, atol=0.0)
-        errStr = f"Chemical potential value should be {self.True1} and not {methodVal}."
+        methodVal = chemical_potential(n_e, T).value
+        testTrue = np.isclose(methodVal, expected_beta, rtol=1e-16, atol=0.0)
+        errStr = (
+            f"Chemical potential value should be {expected_beta} and not {methodVal}."
+        )
         assert testTrue, errStr
```
rocco8773, line 178: Can you add another `assert` to ensure the returned unit is `dimensionless_unscaled`?  It'll likely look something like...

```python
assert testTrue, errStr
assert chemical_potential.unit == u.dimensionless_unscaled
```

Also, could you please refactor this to meet PEP8 naming standards?  e.g. `methodVal` -> `method_val`

plasmapy/formulary/quantum.py
JaydenR2305: Sure. That means this function would just be a wrapper for that function multiplied by beta?

changelog/1726.trivial.rst
JaydenR2305: Whoops, that shouldn't say beta anyways. It should be the dimensionless chemical potential

plasmapy/formulary/quantum.py
rocco8773: Yeah, in the end that would be the case.

rocco8773 (approved):
Looks good to me!  Thank you for working on this.

PR #1727, opened 2022-09-03
Bumping `astropy` dependency to `>= 5.0.1` due to `numpy.asscalar` deprecation

JaydenR2305:
CI failures regarding `numpy.asscalar` began popping up ~2 days ago, which correlates with numba release 0.56.2. This pull request declares a strict version of 0.56.0 to be used for the python 3.8 minimal environment

      docstrings.

### DIFF
--- a/changelog/1727.trivial.rst
+++ b/changelog/1727.trivial.rst
@@ -1,1 +1,1 @@
-Increased minimum version of Astropy_ to 5.0.
+Increased minimum version of Astropy_ to 5.0.1.
--- a/changelog/1727.trivial.rst
+++ b/changelog/1727.trivial.rst
@@ -1,0 +1,1 @@
+Set minimum dependency version for numba to 0.56.0 in minimal test environment
--- a/changelog/1727.trivial.rst
+++ b/changelog/1727.trivial.rst
@@ -1,1 +1,1 @@
-Bump minimum astropy version to 5.0
+Bump minimum astropy version to 5.0.1
--- a/changelog/1727.trivial.rst
+++ b/changelog/1727.trivial.rst
@@ -1,1 +1,1 @@
-Bump minimum astropy version to 5.0.1
+Increased minimum version of Astropy_ to 5.0.
--- a/changelog/1727.trivial.rst
+++ b/changelog/1727.trivial.rst
@@ -1,1 +1,1 @@
-Set minimum dependency version for numba to 0.56.0 in minimal test environment
+Bump minimum astropy version to 5.0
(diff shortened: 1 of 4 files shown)


JaydenR2305:
Testing this locally seems to verify that this works! ignoring the failing the `chemical_potential` docstring of course, which I hope to resolve in another pull request

pheuer:
@JaydenR2305 Can you post the CI error that was being raised, and identify where it is being raised? As far as I can see, `np.asscalar` doesn't seem to be supported by numba, so it shouldn't be appearing in numba functions anyway? 

I'm hesitant to lock numba on a past version if we can avoid it... it would be preferable to resolve the issue so that it works with the current version of numba.

rocco8773:
This is actually an `astropy` conflict, since we don't use `asscalar` in our functionality.  Essentially, `numpy` deprecated `asscalar` back in `v1.16` and officially removed it in `v1.23`.  Astropy did not removed their use of `asscalar` until `v5.0`.  So, we have two options here:

1. bump the minimum version of astropy to `v5.0`
2. set an upper limit on numpy to `<1.23`

After discussing with @namurphy, we think bumpy the minimum version of astropy to `astropy >= 5.0` is the option to proceed with.  @JaydenR2305 would you be willing to make the necessary changes?  Any file that refers to `astropy` `v4.3.1` would need to be modified.  This includes `./setup.cfg`, `./requirements/install.txt`, and `./tox.ini`.

JaydenR2305:
@rocco8773 Looks like `chemical_potential` decided to pass anyway this time🤷

changelog/1727.trivial.rst
```diff
@@ -0,0 +1 @@
+Bump minimum astropy version to 5.0.1
```
rocco8773, line 1:
```suggestion
Increased minimum version of Astropy_ to 5.0.
```

Just to remain consistent with similar past entries and our [changelog guide](https://docs.plasmapy.org/en/latest/contributing/changelog_guide.html#changelog-guidelines) (i.e. use of past tense).  The ``Astropy_`` is an explicit link that will generate a link to https://www.astropy.org/.

rocco8773:
> @rocco8773 Looks like `chemical_potential` decided to pass anyway this time🤷

😆 Well, that how the `chemical_potential` issue has presented itself.

You, can leave the minimum version of astropy at `5.0.1`.  The PR (https://github.com/astropy/astropy/pull/12535) that fixed this happened Nov. 30, 2021, so after the `v5.0` release and before the `v5.0.1` release.

changelog/1727.trivial.rst
```diff
@@ -0,0 +1 @@
+Increased minimum version of Astropy_ to 5.0.
```
rocco8773, line 1:
```suggestion
Increased minimum version of Astropy_ to 5.0.1.
```

My bad on this one.

namurphy (approved):
LGTM! Thank you for doing this!

rocco8773 (approved):

PR #1732, opened 2022-09-08
Expand usage of `particle_input` in `plasmapy.formulary`

namurphy:
This PR is to use `@particle_input` in more of the functions in `plasmapy.formulary` that accept particles as inputs.  This follows the `particle_input` refactoring in #1057, and complements #1718 (which updates docstrings).  

I'm planning on additional pull requests that follow up on #1057, so I'm not planning on addressing everything here.  In particular, this PR won't address `collisions.py` or `braginskii.py`.


changelog/1732.trivial.1.rst
```diff
@@ -0,0 +1,12 @@
+Decorated the following functions with |particle_input|\ :
```
StanczakDominik, line 1:
```suggestion
The following functions now support |particle_input|\ :
```

plasmapy/formulary/distribution.py
```diff
@@ -142,8 +146,16 @@ def Maxwellian_1D(v, T, particle="e", v_drift=0, vTh=np.nan, units="units"):
...
 def Maxwellian_velocity_2D(
-    vx, vy, T, particle="e", vx_drift=0, vy_drift=0, vTh=np.nan, units="units"
+    vx,
+    vy,
+    T,
+    particle: ParticleLike = "e",
+    vx_drift=0,
+    vy_drift=0,
+    vTh=np.nan,
+    units="units",
```
StanczakDominik, line 158: I had to do a scroll-up-double-take at what the units argument meant here. Just now realizing that it might be too much customization and we should probably validate the units initially (O(1)), force unitless and append them at the end. Obviously out of scope for this PR, but I'll make an issue out of this.

plasmapy/formulary/quantum.py
```diff
@@ -23,18 +23,20 @@
...
+from plasmapy.particles import particle_input, ParticleLike
 from plasmapy.particles.exceptions import InvalidParticleError
 from plasmapy.utils import RelativityError
 from plasmapy.utils.decorators import validate_quantities
 
 __all__ += __aliases__
 
 
-# TODO: Use @check_relativistic and @particle_input
+# TODO: Use @check_relativistic
```
StanczakDominik, line 34: Just chippin' away 😆 always love to see progress like this!

StanczakDominik (approved):
Looks good in general! One question - isn't there anything we should be worrying about from a deprecation standpoint here? Are all the functions backwards compatible?

namurphy:
> Looks good in general! One question - isn't there anything we should be worrying about from a deprecation standpoint here? Are all the functions backwards compatible?

That is indeed an important question!  Overall this would be an expansion of functionality so that `CustomParticle` and `ParticleList` instances should now be able to work with most of these functions (though we still very much need to add tests to make sure that there are no incompatibilities).  There was one difference in the exception being raised in `test_quantum.py` but otherwise there weren't any substantive changes to tests.  It's still possible that there will be different exceptions raised for some cases, though those would be limited to probably rare cases that we're not testing.

plasmapy/formulary/distribution.py
namurphy: I'm wondering if the parameter should be `skip_validations` instead of `units`.  In any case, that's for another day!

namurphy:
Oh, and thank you for the review!

PR #1735, opened 2022-09-09
Use `nox` as a test runner instead of `tox`

### LINKED ISSUE #1734: Run tests with `nox` instead of `tox`
We currently use [tox](https://tox.wiki/en/latest/) to automate our tests.  It is a really good tool, with the tradeoff that the configuration file, [`tox.ini`](https://github.com/PlasmaPy/PlasmaPy/blob/main/tox.ini), is in a rather esoteric format which makes it hard to update & edit tests.  (At the very least, I personally have trouble understanding it.)  Another limitation is that it's hard to dynamically generate test environments.

An alternative is [nox](https://nox.thea.codes/en/stable/), which is similar to tox, except that the test configuration file is written in Python.  We're a lot more familiar with Python, so that will make the tests easier to maintain.  Plus, we can more easily dynamically generate tests.

One motivation for switching to nox would be that we would be able to get the minimum requirements from our requirements files, rather than having to maintain requirements in one more place (currently in our `tox.ini` file).  This would help us with #1276.

A while back I set up a [`noxfile.py`](https://github.com/HinodeXRT/xrtpy/blob/main/noxfile.py) for xrtpy, and was thinking it would be worthwhile to set up a prototype for PlasmaPy as well.

If we do switch to nox, we would also need to update our testing guide.

namurphy:
[`nox`]: https://nox.thea.codes/en/stable/ 
[`tox`]: https://tox.wiki/en/latest/
[XRTpy]: https://github.com/HinodeXRT/xrtpy/blob/main/noxfile.py

This PR adds a prototype testing configuration with [`nox`] instead of [`tox`].  As described in #1734, the motivations for using [`nox`] are:

 - The test configuration files are written in Python.
 - We will be able to more easily (less difficultly?) dynamically generate tests.

The configuration file is `noxfile.py`, which I'm adapting from the test setup for [XRTpy].

Closes #1734.


namurphy:
Yay it's doing stuff!

An issue it's running into now is that it's not finding Python 3.9 for macOS, which is also happening on XRTpy (though it did work in the not-too-distant past, I think).

.github/workflows/testing.yml
```diff
@@ -5,191 +5,83 @@ on:
...
-      if: startsWith(matrix.name, 'Documentation')
-      run: sudo apt-get install graphviz pandoc
-    - name: Run tests
-      run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
-    - name: Upload coverage to codecov
-      if: ${{ contains(matrix.toxenv,'-cov') }}
-      uses: codecov/codecov-action@v3
-      with:
-        file: ./coverage.xml
+        python-version: '3.10'
```
StanczakDominik, line 55: This line needs reverting - as it is, it's installing 3.10 in all three currently attempted test envs. 3.8 Seems to already be there for 3.8/Windows, which is why that one isn't failing. I suspect it should fix the 3.9/MacOS failure

StanczakDominik:
I hope you don't mind me tinkering with your branch a little 😅 see comment on the latest commit for my reasoning!

namurphy:
> I hope you don't mind me tinkering with your branch a little sweat_smile see comment on the latest commit for my reasoning!

Not a problem!  I might not have a chance to look into this for a few days, so please tinker away!

noxfile.py
```diff
@@ -0,0 +1,77 @@
...
+python_versions = ("3.8", "3.9", "3.10")
+
+sphinx_paths = ["docs", "docs/_build/html"]
+sphinx_fail_on_warnings = ["-W", "--keep-going"]
+sphinx_builder = ["-b", "html"]
+sphinx_opts = sphinx_paths + sphinx_fail_on_warnings + sphinx_builder
+sphinx_no_notebooks = ["-D", "nbsphinx_execute=never"]
+sphinx_nitpicky = ["-n"]
+
+pytest_options = []
```
namurphy, line 14:
```suggestion
pytest_options = ["--showlocals"]
```

noxfile.py
```diff
@@ -0,0 +1,77 @@
...
+python_versions = ("3.8", "3.9", "3.10")
+
+sphinx_paths = ["docs", "docs/_build/html"]
+sphinx_fail_on_warnings = ["-W", "--keep-going"]
+sphinx_builder = ["-b", "html"]
+sphinx_opts = sphinx_paths + sphinx_fail_on_warnings + sphinx_builder
+sphinx_no_notebooks = ["-D", "nbsphinx_execute=never"]
+sphinx_nitpicky = ["-n"]
+
+pytest_options = ["--showlocals"]
```
namurphy, line 14: - [ ] Make sure to keep all of the options that are currently in `tox.ini`

namurphy:
pre-commit.ci autofix

namurphy:
Hm...there's still an issue with one test for coverage that's seemingly not being run:

> **Python 3.8, all tests, code coverage** _Expected — Waiting for status to be reported_

namurphy:
And...I broke it! 😹

StanczakDominik:
I see what happened here! The thing you missed:

```
steps:
  - uses: actions/checkout@v2
  - uses: excitedleigh/setup-nox@v2.0.0
  - run: nox
```

Note that, due to the hyphenated list, these are three steps! You tried to stick `setup-nox` and `run nox` into one step.

StanczakDominik:
This actually conflicts with #1758 a tiny bit, because I stuck the config for flake8... in `tox.ini` 😆 It's no problem to move that to `.flake8rc` or whatsitcalled, though.

StanczakDominik:
Maybe let's get #1754 in first.

namurphy:
The main things I still need to cross over:
 - [ ] Code coverage
 - [ ] `codespell`
 - [ ] Building wheels, etc.

namurphy:
Before marking this as ready for review, I'll need to:
 - [ ] Verify that changes made in `.github/workflows/testing.yml` in the last few months have been ported over, including:
 - [ ] Printing out link to troubleshooting section of contributor guide in test output
 - [ ] Printing out link to testing guide in test output

requirements.txt
```diff
@@ -233,6 +233,7 @@ nest-asyncio==1.5.6
     #   jupyter-client
 nodeenv==1.7.0
     # via pre-commit
+nox
```
StanczakDominik, line 236: Remember that you should pin a version here! The command to do that is up top, but it won't work until you add nox to `pyproject.toml` first.

noxfile.py
```diff
@@ -0,0 +1,62 @@
...
+    session.run(
+        "sphinx-build",
+        *sphinx_opts,
+        *session.posargs,
+    )
+    print(post_doc_build_comments)
+
+
+@nox.session
+def build_docs_no_examples(session):
```
StanczakDominik, line 54:
```suggestion
def build_docs_no_examples(session):  # TODO: "parametrize" this? the same
```

I would absolutely not want to hold this up, but I'm curious, and I just want to make sure I notice this when picking it up later 😉

noxfile.py
```diff
@@ -0,0 +1,62 @@
...
+    session.run("python", "-c", "import plasmapy")
+
+
+sphinx_paths = ["docs", "docs/_build/html"]
+sphinx_fail_on_warnings = ["-W", "--keep-going"]
+sphinx_builder = ["-b", "html", "-n"]
+sphinx_opts = sphinx_paths + sphinx_fail_on_warnings + sphinx_builder
+sphinx_skip_notebooks = ["-D", "nbsphinx_execute=never"]
+
+post_doc_build_comments = """
```
StanczakDominik, line 36: I half expect to see this become `tenured_doc_build_comments` later.

StanczakDominik (commented):
If we're doing this in halves, then it might make sense to merge this to a feature branch first. Then I'll work off that before we put this on main.

PR #1745, opened 2022-09-19
Show local variables in test failures

namurphy:
This PR adds the `--showlocals` flag to our call to `pytest` in `tox.ini` in order to show local variables.  Here's example test output for a purposefully failing test:

```python
    def test_that_fails():
        x = 1
        y = 2
        z = 624
>       assert x == y
E       assert 1 == 2

x          = 1
y          = 2
z          = 624

plasmapy/tests/test_whatev.py:5: AssertionError
```

This can be really helpful for debugging, so I figured it'd be easier to have it on as the default.  The downside is that the test output is longer.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ allowlist_externals =
 setenv =
     MPLBACKEND = agg
     COLUMNS = 180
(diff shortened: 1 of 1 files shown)


StanczakDominik (approved):
Hey, didn't know that one, thanks!

PR #1746, opened 2022-09-25
Add temperature isotropization and thermal equilibration

### LINKED ISSUE #1709: Add temperature isotropization and thermal equilibration to CollisionFrequencies
The NRL formulary on pages 33 and 34 (2019 version) defines collision/relaxation frequencies for temperature isotropization and thermal equilibration. These could be added as properties to the `CollisionFrequencies` class

Temperature isotropization is the rate at which an anisotropic temperature distribution relaxes to isotropic. There are different forms for positive and negative anisotropy A. 

image

Thermal equilibration is the time that it takes two Maxwellian species with different temperatures to equilibrate. 
image

JaydenR2305:
## Description

This pull request will add functionality to calculate temperature isotropization and thermal equilibration for two Maxwellian populations as part of the `MaxwellianCollisionFrequencies` class.

## Motivation and context

Expands functionality of the `MaxweillianCollisionFrequencies` class beyond basic collision frequencies.

## Related issues

Resolves #1709

### DIFF
--- a/plasmapy/formulary/collisions.py
+++ b/plasmapy/formulary/collisions.py
@@ -66,6 +66,7 @@ from astropy.constants.si import e, eps0, hbar, k_B, m_e
 from functools import cached_property
 from numbers import Real
 from numpy import pi
+from typing import Optional
 
 from plasmapy import particles, utils
 from plasmapy.formulary.lengths import Debye_length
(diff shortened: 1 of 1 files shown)


JaydenR2305:
@pheuer I'm still somewhat conflicted about including this in `MaxweillianCollisionFrequencies`.. Are these frequencies considered collision frequencies?

pheuer:
> @pheuer I'm still somewhat conflicted about including this in `MaxweillianCollisionFrequencies`.. Are these frequencies considered collision frequencies?

That's a reasonable question. These aren't exactly collision frequencies, but they are frequencies and are closely related to collisions (since the collisions are what is causing the isotropization or equilibrium). So, I agree maybe not a perfect fit but maybe close enough that it still makes sense to put them here? 

I suppose the other option would be to just implement these as standalone functions in the collisions module?

PR #1759, opened 2022-10-08
First try at reusing tox env setup steps

### LINKED ISSUE #1754: Use reusable workflows to improve modularity of our GitHub Actions
### Feature description

Implement [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for our continuous integration testing.

### Motivation

Right now we have a lot of repeated code in our GitHub Actions for continuous integration testing. For example, we have the steps for running `tox` for our tests twice in `.github/workflows/testing.yml` (for `initial-tests` and `comprehensive-tests`) and then again in `.github/workflows/fortnightly.yml`. This means that if we want to change the steps for doing these tests, we would have to do it in multiple places.  The repetition makes our GitHub Actions complicated and difficult to understand, instead of just...difficult to understand.  

### Implementation strategy

We could possibly create a file (maybe `run_test.yml`?) that contains the steps needed to run `tox` (or possibly `nox`).  That one would have `on: workflow_call`.  

We could possibly create a separate repo for these, but for the moment I'm inclined to keep it in this repo.

### Additional context

This is related to and inspired by #1734 and #1735. It might be easier to work on this issue first before finishing up #1735.

This also looks like a helpful resource: [Automating build/deploy CI/CD with GitHub Actions](https://www.viget.com/articles/automating-build-deploy-ci-cd-with-github-actions/).

StanczakDominik:
## Description

I ended up using https://docs.github.com/en/actions/creating-actions/creating-a-composite-action - we're not reusing *workflows*, as that'd reuse a set of jobs for a bunch of repositories, apps or projects. Instead, we're reusing a set of steps of a particular job, and that's essentially a custom action.

## Related issues

Closes #1754


StanczakDominik:
I don't get why it fails, it looks like it follows https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow nicely.

StanczakDominik:
All in all I don't think this did much. But it's nice to have.

StanczakDominik:
macos, why?

PR #1772, opened 2022-10-31
Delete `codemeta.json` in favor of `CITATION.cff`

namurphy:
Right now, we use both `codemeta.json` and `CITATION.cff` to include metadata for our project. While CodeMeta and Citation File Format each have useful fields that the other one does not have, it'll be a pain to keep both up-to-date...in particular because the bulk of the information (i.e., the author list) will be duplicated in both.  Thinking about the tradeoffs between the two of them, `CITATION.cff` has better integration with GitHub and Zenodo, and thus would end up making our lives easier.  

We could eventually go back to `codemeta.json`, especially since there are tools that will do the conversion from CodeMeta → CFF, though I had trouble finding ones that went CFF → CodeMeta and covered all of the fields that we needed and were out of beta.

(I do really like the CodeMeta format though, but alas!)


StanczakDominik (approved):
Lovely addition-to-deletion ratio <3

PR #1775, opened 2022-10-31
Test on Python 3.11 and remove split between initial and comprehensive tests

### LINKED ISSUE #1767: Add Python 3.11 to the test suite
### Feature description

We will need to add a testing environment for Python 3.11.

### Motivation

Python 3.11 was released yesterday, and there was much rejoicing.

### Implementation strategy

This will have to go in `tox.ini` and/or our `.github/workflows/testing.yml` (or possibly `noxfile.py` if #1734 is merged).

### Additional context

It will probably take a few weeks for NumPy to become compatible with Python 3.11, and probably ∼1–3 months for Numba to become compatible with Python 3.11 too.

namurphy:
The PR includes:

 - Adding test support for Python 3.11 (now that Numba v0.57.0 has been released)
 - Reducing the number of Python checks so that we're doing one each of {Windows, Linux, MacOS} and one each of Python {3.9, 3.10, 3.11}.  This will allow us catch the vast majority of problems, and we still have the weekly tests which are more thorough.  We still have coverage reporting and one case using develop mode, and the 3.9 tests still check the minimal dependencies.
 - Consolidated the "initial tests" and "comprehensive tests".  We ran the initial tests first, followed by the comprehensive tests if the initial tests passed. The downside is that we had duplicated test code for each of the initial and comprehensive tests, which made tests harder to maintain. Since this PR reduces the number of Python test runs, it doesn't help much to split it up into initial vs. comprehensive tests, and it also delays reporting and the like. So, have them all start running at the same time.

Because I renamed tests, there are some "required" tests that are no longer able to be run, so the review should ignore the tests that have a perpetual orange dot. When this is merged, I will need to update the branch protection rules to update the names of tests. 

The RTD test failure is unrelated, and is being fixed in #2103.  Closes #1767.

~~Waiting on https://github.com/numba/numba/issues/8304.~~

Thank you!


.github/workflows/testing.yml
```diff
@@ -134,7 +139,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: 3.11
```
namurphy, line 142: We might need to drop this to 3.10 since `codespell` doesn't yet work on Python 3.11, but I'm guessing it'll end up working on 3.11 sooner than Numba.

namurphy:
As expected...

```
Collecting numba
  Downloading numba-0.56.3.tar.gz (2.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.4/2.4 MB 143.3 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [8 lines of output]
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-tb_gttl2/numba_723f16d3caa44ac482a53bdd4c422953/setup.py", line 51, in <module>
          _guard_py_ver()
        File "/tmp/pip-install-tb_gttl2/numba_723f16d3caa44ac482a53bdd4c422953/setup.py", line 48, in _guard_py_ver
          raise RuntimeError(msg.format(cur_py, min_py, max_py))
      RuntimeError: Cannot install on Python version 3.11.0; only versions >=3.7,<3.11 are supported.
      [end of output]
```

.github/workflows/testing.yml
```diff
@@ -134,7 +139,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: '3.11'
```
namurphy, line 142: It turns out the the codespell test is failing because of the Numba dependency too. However, codespell doesn't actually depend on Numba.  Hm...

namurphy:
It appears that I created this pull request about four months too soon. 😐

.github/workflows/testing.yml
```diff
@@ -20,6 +20,11 @@ jobs:
       matrix:
         include:
 
+        - name: Python 3.11
+          os: ubuntu-latest
+          python: '3.11'
+          toxenv: py311
```
namurphy, line 26: We'll need to go to branch protection rules on the GitHub repository for `main` to update the names of required tests.

namurphy:
Numba 0.57.0 was just released!

.github/workflows/testing.yml
ejohnson-96: Hmmm.... does Numba include dependence on code spell?

requirements.txt
```diff
@@ -81,10 +81,6 @@ docutils==0.18.1
     #   sphinxcontrib-bibtex
 entrypoints==0.4
     # via jupyter-client
-exceptiongroup==1.1.1
```
ejohnson-96, line 84: No replacement code here, is that right?

ejohnson-96 (approved):
Waiting on last few tests to pass, providing they do I will enable auto-merge 👍

requirements.txt
namurphy: Correct!  I think this is no longer needed since the pinned dependencies are for Python 3.11.

namurphy:
Thanks for the review!  I changed the names of the tests, so I had to update the branch protection rules on GitHub.

PR #1729, opened 2022-09-08
Add a GitHub Action to post a code review checklist on pull requests

namurphy:
The goal of this PR is to make it so that a code review checklist gets added in a comment on each PR, since I've seen a lot of resources here and there that have recommended having a code review checklist as a way to facilitate the code review process.  I'm thinking of having this take on some of the role that our PR template currently has.

Some of the resources I've looked at to come up with this list are:

 - [Astropy's PR template](https://github.com/astropy/astropy/blob/ca5531bbf9ffbb63b98e44098c3072569a0839ef/.github/PULL_REQUEST_TEMPLATE.md)
 - [Effective Code Review Checklist](https://engineerchirag.medium.com/effective-code-review-checklist-c735abbcd613) by Chirag Goel

I'm going to attempt to use [this GA](https://github.com/NejcZdovc/comment-pr).  I'd like to generate a link to the documentation build preview based on the PR number, though I don't know how to do that yet.

PR #1776, opened 2022-10-31
Move checklist back into PR template

StanczakDominik:
## Description

I moved the PR checklist back into the PR template, and removed that particular step of the github action.

## Motivation and context

The comment is nice and the checklist is splendid, but I found it pretty noisy to get the comment on every single PR. This goes up to two comments for new contributors, and let's not overwhelm them to much!

I'm putting this PR up as a proposal; we can always just close it.

## Related issues

This essentially reverts #1729


StanczakDominik:
^ and that's what I'm talking about! Already a notification that essentially conveys no useful info. :(

namurphy:
This is indeed worth thinking about.  Thank you for bringing this idea up!

Here are some thoughts on tradeoffs, without attempting to decide one way or the other.  Feel free to edit this if you think of tradeoffs that are missing.

 - Reasons why the code review checklist should be included in the template
   - Fewer notifications 
   - Contributors could more easily delete the parts of the checklist that don't apply to a particular PR
 - Reasons why the code review checklist should be added as a comment to PRs
   - The PR template is more cluttered (i.e. additional coupling between the PR description and the checklist).
   - We already get at least one notification on PRs for the codecov report, and ∼2–3 roughly simultaneous notifications are arguably not much noisier than a single one. (at least somewhat an opinion)  
   - Contributors might end up deleting the checklist in its entirety. (speculative, though this is what I've usually done in the past)

I'm wondering if it would be worth asking new contributors about this too.

PR #1777, opened 2022-10-31
Add `flake8` extensions

namurphy:
This PR adds the following flake8 extensions:

 - ~~flake8-useless-assert — to make sure that `assert` statements won't always pass or fail~~
 - flake8-use-pathlib — to enforce recommend usage of `pathlib`
 - flake8-builtins — to avoid name conflicts with Python builtins (like `min` and `max`)
 - flake8-comments — to avoid redundant comments

This PR also makes a few minor edits in order for the new flake8 extensions to pass.

Before merging, I'll need to:

 - [ ] Test the plasma calculator

### DIFF
--- a/changelog/1777.trivial.rst
+++ b/changelog/1777.trivial.rst
@@ -1,0 +1,3 @@
+Added the flake8_ extensions ``flake8-use-pathlib``,
+``flake8-builtins``, and ``flake8-comments`` to the testing
+requirements.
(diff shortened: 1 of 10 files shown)


namurphy:
I dropped flake8-useless-assert since it has a `flake8 < 5.0` requirement that led to a dependency conflict.

StanczakDominik (approved):

PR #1782, opened 2022-11-01
Add `tryceratops` as a flake8 extension

namurphy:
## Description

This PR adds [`tryceratops`](https://github.com/guilatrova/tryceratops) as a flake8 extension.

## Motivation and context

This flake8 extension will catch common antipatterns for `try`/`except` blocks and improve code quality.  Also, it has a cool name.

## Notes

I'm not yet going to enable the check corresponding to [TC003](https://github.com/guilatrova/tryceratops/blob/main/docs/violations/TC003.md), even though it seems like a good idea that we'd want to implement for things like `InvalidParticleError`.


setup.cfg
```diff
@@ -87,6 +87,7 @@ tests =
     pytest-regressions
     pytest-xdist
     tomli
+    tryceratops
```
namurphy, line 90: Alas, there doesn't seem to be a triceratops emoji insert sad cat face here.

setup.cfg
```diff
@@ -189,7 +190,9 @@ extend-ignore =
     RST213,
     # RST305 so that ReST substitutions in a different file can be used
     RST305,
-    RST306
+    RST306,
+    # TC003 to allow long messages to be specified outside the exception class
+    TC003
```
namurphy, line 195: [TC003](https://github.com/guilatrova/tryceratops/blob/main/docs/violations/TC003.md) seems like a good thing to implement, but would require comparatively large changes. 

Also, it gets applied to builtin exceptions like `TypeError` for which we can't straightforwardly take this approach unless we create a `TypeError` subclass.

namurphy:
The `codecov/patch` check is failing because this PR changes several lines (mostly for raising exceptions) that were previously untested.

setup.cfg
StanczakDominik: Wow yeah, I didn't even think about it but it makes a lot of sense. I don't even mind making those TypeError subclasses, honestly, if it reduces duplication.

StanczakDominik (approved):
Looks good! Seems like a handy tool. I always forget to `raise WhateverError from source_error`, so it's nice to have this as an automation :)

setup.cfg
namurphy: Good idea!  I created these in #1789 and #1790.

PR #1788, opened 2022-11-04
project: intermittency code for plasmapy

dianajh:
## Description

First component of intermittency subpackage (?)


StanczakDominik:
If it's the first component of many, let's redirect it to a feature branch rather than to main, then we can merge it to main once it's useful :) I can do that later, just noting it down for now.

PR #1757, opened 2022-10-08
Update discussion of parameter type specifications in docstrings

namurphy:
This PR expands the discussion on type lines in docstrings, since it didn't have enough information before.  I'm planning to base a lot of it on the [matplotlib doc guide](https://matplotlib.org/stable/devel/documenting_mpl.html#parameter-type-descriptions) and on https://github.com/numpy/numpydoc/pull/367.  In particular, I want to add a bunch of type line examples, including for infrequent cases.

PR #1793, opened 2022-11-04
Update docstrings in `formulary.collisions.frequencies`

namurphy:
This PR makes a few changes to `plasmapy.formulary.collisions.frequencies`:

 - Updates parameter type descriptions to try out #1757 (which is based on matplotlib's dev guide and numpydoc)
 - Makes some math multi-line and more spaced out
 - Additional docstring updates


docs/common_links.rst
```diff
@@ -7,6 +7,7 @@
 
 .. |ClassicalTransport| replace:: :class:`~plasmapy.formulary.braginskii.ClassicalTransport`
 .. |RelativisticBody| replace:: :class:`~plasmapy.formulary.relativity.RelativisticBody`
+.. |SingleParticleCollisionFrequencies| replace:: :class:`~plasmapy.formulary.collisions.frequencies.SingleParticleCollisionFrequencies`
```
namurphy, line 10: This might be our longest namespace yet!

plasmapy/formulary/collisions/frequencies.py
```diff
@@ -25,125 +24,144 @@
...
+    Notes
+    -----
+    The frequency of collisions between a test particle (subscript
+    :math:`α`) and a field particle (subscript :math:`β`) each with mass
+    :math:`m` and charge :math:`e` are given by four differential
+    equations:
 
-            momentum loss: :math:`\frac{d\textbf{v}_{α}}{dt}=-ν_{s}^{α\backslashβ}\textbf{v}_{α}`
+        momentum loss: :math:`\frac{d\mathbf{v}_α}{dt}
+        = - ν_{s}^{α \backslash β} \mathbf{v}_α`
```
namurphy, line 70: For this and the follow equations, I still need to verify that they are rendering just like they did before.

plasmapy/formulary/collisions/frequencies.py
```diff
@@ -25,125 +24,144 @@
...
-        T_b: u.K,
-        n_b: u.m**-3,
-        Coulomb_log: u.dimensionless_unscaled,
-    ):
-        r"""
-        Compute collision frequencies between test particles
-        (labeled 'a') and field particles (labeled 'b').
+    r"""
+    Compute collision frequencies between test particles (labeled 'a')
+    and field particles (labeled 'b').
```
namurphy, line 29: Putting the docstring right after the class definition rather than in `__init__` to match what we've typically done elsewhere.  It probably doesn't change how it shows up in the docs.

docs/common_links.rst
StanczakDominik: Have we thought about moving the stuff from `collisions` to `formulary`, btw?

StanczakDominik (approved):
Looks good, at a skim. Apparently back in November you said you'd still look through these to see if the math's rendering right. Would indeed be great to do that! 😅

namurphy:
> Looks good, at a skim. Apparently back in November you said you'd still look through these to see if the math's rendering right. Would indeed be great to do that! sweat_smile

Thanks for the reminder!  

 - [ ] Make sure the math is rendering awesomely.

docs/common_links.rst
namurphy: I can't remember if we've talked about that.  I've also wondered if it would be appropriate to have `plasmapy.collisions` be its own subpackage, since it is a really big topic and there's a lot more that could go in it beyond what would be appropriate for the formulary.

PR #1795, opened 2022-11-08
Adjust signature of `MagneticDipole.magnetic_field` method

### LINKED ISSUE #1575: Adjust signature of `magnetic_field` method for magnetostatic classes
Right now the [MagneticDipole.magnetic_field](https://github.com/PlasmaPy/PlasmaPy/blob/646f7ed52b89a1254be474fe54bdd672f7d27fb3/plasmapy/formulary/magnetostatics.py#L74) method (from `plasmapy.formulary.magnetostatics` requires a position vector, `p` which makes it harder to do array operations.  We could instead do something like having keyword-only parameters like `x`, `y`, & `z`...which would also let us have additional keyword-only parameters like `r`, `z`, and `phi` or `r`, `theta`, `phi` for cylindrical and circular coordinates, respectively.

bucket420:
## Description

I added keyword-only arguments for different coordinate systems to the `MagneticDipole.magnetic_field` method from `plasmapy.formulary.magnetostatics`.

## Related issues

Closes #1575.


bucket420:
pre-commit.ci autofix

Issue #1725, opened 2022-09-02
[feature request] `proton radiography diagnostic` with an additional `grid` for externally applied magnetic field


weipengyao:
Dear developers,

Now that more and more research groups are coupling their laser-target interaction with externally applied stable (~us), large volume (~cm^3), and strong (~10s T) magnetic field, taking account of this external B-field in the proton radiography (PR) diagnostic seems more and more crucial for the proper interpretation (for example, a recent paper: https://aip.scitation.org/doi/10.1063/5.0054172)

The PR tool in `PlasmaPy` has a `grid` between the `source` and the `detector`, in which we can add the EM fields, generated from the laser-target interaction, etc. It typically has a volume of mm^3, with a resolution of 10s micron.

Normally, we apply the simulated or analytical EM fields into this `grid` and check the results from the detector. 
However, with the applied magnetic field, we need one more grid with cm^3 in volume and, probably, lower resolution, to allow the source particles (i.e., p+) to propagate through, before they reach the detector.

Is it possible to add such a feature in the PR tool, that allows the consideration of the externally applied magnetic fields?

Thanks very much!
Weipeng YAO

pheuer:
Hello! 

Yes, I strongly agree that this would be a useful addition, and I've been playing with how to do it for a while. I think the best solution is to allow multiple grids to be provided in an ordered list. If particles are on the first grid, then the fields should be interpolated from that grid. If not, the code should then check if they are on the next grid in the list, etc. So, for your example, the small micron scale grid would be first, followed by the cm-scale background field grid. 

I can bump this up my queue to work on, but if you or a student would like to try to make the modifications as a new pull request I would be very happy to work with you on that!

weipengyao:
Hi Peter, 

Thanks for your reply and support. I would be happy to give it a try with your guidance.

Just one comment about my example case -- since the micron-scale grid is immersed inside the coil that generates the B-field, so even before the small grid, there is already some B-field (with not-so-small gradient).
For example, a typical spatial distribution of the B-field can be found in this paper, Fig. 5 (https://aip.scitation.org/doi/10.1063/1.4795551)

<img width="867" alt="image" src="https://user-images.githubusercontent.com/37392893/188655908-e46d2d59-e471-4692-9fee-6f4f04df7158.png">


But I guess the essential strategy is still the same. 

So, where should I start with?

Thanks again!

pheuer:
[...]

PR #1274, opened 2021-09-22
Charged Particle Radiography: Detector Stacks

pheuer:
Charged particle radiography detectors are usually comprised of multiple layers of active medium, each of which detects different energies of particles based on their Bragg peak. This PR adds a couple objects that represent these detector stacks and calculates the energies deposited in each layer using stopping power data from NIST. 

This is Step 1 of adding support for detector stacks. Step 2, in a future PR, will create a new `synthetic_radiograph` function that takes a film stack object and produces a synthetic radiograph in each layer. 

~~This PR depends on #1486 and should be merged after that PR~~ Done

PR #1799, opened 2022-11-15
DRAFT: Prad multiple field grids

pheuer:
My attempt to address #1725 

Wait to merge until #1274 is done, since that will refactor the structure of the charged particle radiography code. 

TODO list

- [x] Allow a list of 'grid' objects to be passed into the 'grid' argument. If only one grid is passed, wrap it in a list so that there is always a list. 

- [x] Modify  max theta hit grid for each grid in the list, then take the maximum

- [x] Similarly, for the rest of the `__init__ method`, modify so that all the validation etc. for the grids is being run on each grid in the list. 

- [x] Modify `_coast_to_grid` to calculate `dist` to each grid and then coast to the minimum distance (e.g. the edge of the largest grid). 

- [x] Within `_push`, run `on_grid` on each of the grids. `self.on_grid` will need to be a list of lists now, recording which particles are on each grid. NOTE: -> now on_grid has shape [nparticles, ngrids]

- [x] (The big one): the interpolation section of  `_push` needs to be modified so that you interpolate the correct fields for the correct particles and then sum those fields as I proposed above.  NOTE -> We interpolate every particle on every grid, and just add it all together for each particles. So grids are additive.

- [x] Modify `_adaptive_dt` to calculate the grid resolution of each grid _that currently has particles on it_  based on the ongrid method. We'll need a different dt for each grid. The timestep for each particle needs to be governed by the highest resolution grid it is on. E.g. if it is on both, the smaller dt wins. To avoid unnecessary runtime, we should probably calculate dt for each particle separately (so that those on larger grids can move faster than those on smaller grids). NOTE -> Now calculating a separate dt for every particle, based on what grids it is on.  When a particle is on no grid, give it the largest gridstep as dt.

- [x] Modify `_push` to push the particles in batches based on which timestep they have. If we push everything based on the smaller timestep, the particles will take a very long time to move through the large length scale low res grid. NOTE -> We now handle this by having every particle have its own dt.

- [x] Modify `_stop_condition` as necessary to deal with changes to `entered_grid`. We want to know what fraction of the particle have entered A grid and are now on NO grid. 

- [x] Implement at least one test with two grids of different scales and an expected analytical outcome and demonstrate that the new functionality works. That will go in this file: https://github.com/PlasmaPy/PlasmaPy/blob/main/plasmapy/diagnostics/tests/test_charged_particle_radiography.py


pheuer:
@weipengyao I took a stab at this and it seems to be working - give it a try (using this branch) and let me know what you think!

weipengyao:
> @weipengyao I took a stab at this and it seems to be working - give it a try (using this branch) and let me know what you think!

@pheuer Hi Peter, thanks again for the updates! Yes, it works! Now I am playing with the size, resolution, and the profile of the additional grid. But the function itself is good, and I do see a difference in the final results with and w/o the 2nd grid.

plasmapy/diagnostics/charged_particle_radiography.py
```diff
@@ -733,56 +755,78 @@ def load_particles(
...
-        dist = np.min(np.linalg.norm(self.grid_arr - self.source, axis=3))
+        # Distance from the source to the nearest point on any grid
+        dist = np.min(
+            np.array(
+                [
+                    np.min(np.linalg.norm(arr - self.source, axis=3))
+                    for arr in self.grids_arr
+                ]
+            )
+        )
```
StanczakDominik, line 820: Unless you have a WHOLE LOT of grids, there's not going to be much of a gain in having the list of grids be an array as well, so we can simplify here:
```suggestion
        dist = min(
                    np.min(np.linalg.norm(arr - self.source, axis=3))
                    for arr in self.grids_arr
        )
```

StanczakDominik (commented):
Looks good! Pointed out a few possible rewrites, and the test automation issue seems challenging, but... yeah, looks good!

plasmapy/diagnostics/charged_particle_radiography.py
StanczakDominik: If we were reusing the array of minima along axis=3 sure, but these are passed straight to min. So you're just adding an iteration over all the elements to change their container before iterating over them again to find the scalar minimum, unless I'm misunderstanding this? Try it in ipython with a %timeit!

pheuer: Got it - thanks!

pheuer:
@StanczakDominik I've updated from main and merged with the changes from #1274 , and made the changes you suggested above. Writing an automated test to ensure this works would be heavy (require a high resolution Tracker run) and I don't really think it's worth it: if it runs to completion without error it should be working correctly.

StanczakDominik:
I'm happy to skip it, then!

I'll just figure out the doc errors and we can probably merge it. I'll take it from here :)

pheuer:
> I'll just figure out the doc errors and we can probably merge it. I'll take it from here :)

Doc issue resolved! It was easy, just a stupid typo

StanczakDominik (approved):

pheuer:
Awesome, thanks @StanczakDominik

StanczakDominik:
Nah, thank you for doing this and for finding the doc bug ;)

PR #1467, opened 2022-03-05
Add workflow to label PR's based on size

RAJAGOPALAN-GANGADHARAN:
Using https://github.com/marketplace/actions/add-size-label-to-pr to add size labels to PR's. Triggers when new PR's are created and the existing ones are updated.

- [ ] I have added a changelog entry for this pull request.

Closes #1404

PR #1800, opened 2022-11-16
Update size labeler GitHub workflow

namurphy:
We had added the size labeler in #1467, but for some reason, it hasn't been working.  This PR:

 - Renames the file and workflow name to 
 - Updates the action to match the [updated usage example](https://github.com/CodelyTV/pr-size-labeler)

### DIFF
--- a/.github/workflows/size-labeler.yml
+++ b/.github/workflows/size-labeler.yml
@@ -1,18 +1,13 @@
-name: Pull Request Labeler
-on:
-- pull_request_target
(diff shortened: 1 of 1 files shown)


namurphy:
This might be one of those PRs where we need to merge it to make sure it works.  It hasn't added size labels so far, though.

StanczakDominik (approved):
Let's try it out, then :)

namurphy:
Thanks!

PR #1808, opened 2022-11-29
Change fortnightly tests to weekly even though "fortnightly" sounds waaaay cooler than "weekly"

namurphy:
This PR makes some edits to our cron tests, in particular making our tests run early on Tuesday, timed to be not long before our community meeting.  This will increase the likelihood that we'll notice new failures before they show up in unrelated PRs.

I'll admit being just a little sad about reducing our usage of "fortnightly" in this repository by ∼70%, but I shall take heart in that we still get to use it at least a few times!  🗓️ 📅 📆 ⌛ 🕢


.github/workflows/testing.yml
```diff
@@ -161,7 +161,7 @@ jobs:
 
   build-n-publish:
     name: Packaging
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
StanczakDominik, line 164: I would actually prefer to pin this to a particular target, rather than a moving one - for builds, we really want as much stability as possible. We can absolutely try something the current `latest` though:
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/weekly.yml
```diff
@@ -1,8 +1,8 @@
-name: fortnightly tests
+name: weekly tests
 
 on:
   schedule:
-  - cron: 17 7 1,15 * *
+  - cron: 17 7 * * 2
```
StanczakDominik, line 5: I like how this would run in the 48h before our telecon. I wonder about the timezone here, though. Probably UTC.

Could do this to be on the safe side:
```suggestion
  - cron: 17 7 * * 1
```

StanczakDominik (changes requested):

.github/workflows/weekly.yml
```diff
@@ -1,8 +1,8 @@
-name: fortnightly tests
+name: weekly tests
 
 on:
   schedule:
-  - cron: 17 7 1,15 * *
+  - cron: 37 7 * * 1
```
namurphy, line 5: I just changed the `17` → `37` because I often forget which of the first two slots in a cron schedule correspond to hours and which to minutes.  This will remove the hours/minutes ambiguity when running cron jobs on all planets in the solar system, except for Mercury and Venus.

StanczakDominik (approved):

PR #1813, opened 2022-12-01
Added relativistic functionality to gyroradius

### LINKED ISSUE #1537: Extend `gyroradius` to the relativistic limit
I just noticed that [`gyroradius`](https://docs.plasmapy.org/en/latest/api/plasmapy.formulary.lengths.gyroradius.html?highlight=gyroradius#plasmapy.formulary.lengths.gyroradius) is presently only applicable in non-relativistic cases, but there are a lot of applications in astrophysics that involve relativistic particles in magnetic fields, and it would be helpful to be able to calculate the [gyroradius in the relativistic limit](https://en.wikipedia.org/wiki/Gyroradius#Relativistic_case) as well: 

I'm also wondering if we could have the Lorentz factor γ be a keyword-only argument that could be used instead of the velocity.

MuhammadHMasood:
## Description

I made it so that the gyroradius is calculated, it accounts for relativistic effects by calculating the lorentz factor and multiplying the gyroradius by it. I also added the lorentz factor as an optional keyword if the user wants high precision with their calculation or they can set it to 1 if they don't want it to be relativistic. I also fixed a misleading error message (instead of erroring only when both V and T are provided, it errors when both or neither are provided, since both cases are invalid)

## Motivation and context

Issue #1537 brought up how currently it has inaccurate behavior, so I fixed it so that now the function gives the actual gyroradius. If a user wants to go back to the old behavior then they can use the optional keyword with a lorentz factor of 1.0

## Related issues

Closes #1537


plasmapy/formulary/lengths.py
```diff
@@ -270,7 +289,22 @@ def gyroradius(
 
     omega_ci = frequencies.gyrofrequency(B, particle)
 
-    return np.abs(Vperp) / omega_ci
+    if np.isscalar(lorentzfactor):
```
StanczakDominik, line 292: Maybe simpler:
```suggestion
    if not np.isnan(lorentzfactor):
```

plasmapy/formulary/lengths.py
```diff
@@ -270,7 +289,22 @@ def gyroradius(
...
+        # if it is not finite aka not defined then calc from vperp
+        if not isfinite_lorentzfactor:
+            lorentzfactor = np.copy(lorentzfactor)
```
StanczakDominik, line 295:
```suggestion
```
You don't need this one since the next line overwrite the reference; if you were doing `lorentzfactor[:] = RelativisticBody...`, then sure.

plasmapy/formulary/lengths.py
```diff
@@ -225,12 +235,21 @@ def gyroradius(
...
     isfinite_T = np.isfinite(T)
     isfinite_Vperp = np.isfinite(Vperp)
+    isfinite_lorentzfactor = np.isfinite(lorentzfactor)
 
     # check 1: ensure either Vperp or T invalid, keeping in mind that
     # the underlying values of the astropy quantity may be numpy arrays
-    if np.any(np.logical_and(isfinite_Vperp, isfinite_T)):
-        raise ValueError(
-            "Must give Vperp or T, but not both, as arguments to gyroradius"
+    if np.any(np.logical_not(np.logical_xor(isfinite_Vperp, isfinite_T))):
```
StanczakDominik, line 242: Do you think we could add tests for these cases? If this logic was incorrect before, I suspect the existing ones must not have shown that, which is strange...

plasmapy/formulary/lengths.py
```diff
@@ -225,12 +235,21 @@ def gyroradius(
...
+        # if neither are provided calc Vperp from lorentz factor but but you need lorentz factor to have no undefined elements
+        if np.all(isfinite_lorentzfactor):
+            Vperp = RelativisticBody(particle, lorentzfactor).velocity
+        else:
+            raise ValueError(
+                "Must give one of Vperp or T, but not both, as arguments to gyroradius, and if not then provide the lorentzfactor"
```
StanczakDominik, line 248: Probably clearer this way?
```suggestion
                "Must give either the lorentzfactor or one of: Vperp or T"
```

plasmapy/formulary/lengths.py
```diff
@@ -270,7 +289,22 @@ def gyroradius(
...
+            lorentzfactor = RelativisticBody(particle, Vperp).lorentz_factor
+        # else it is defined already so you are all good and don't need to do anything
+    elif np.all(
+        isfinite_lorentzfactor
+    ):  # if it is not a scalar you want to calculate Lorentz factor based on Vperp for all undefined entries of the array
+        rbody = RelativisticBody(particle, Vperp)
+        lorentzfactor = np.copy(lorentzfactor)
+        lorentzfactor[not isfinite_lorentzfactor] = rbody.lorentz_factor()[
+            not isfinite_lorentzfactor
+        ]
```
StanczakDominik, line 305: Hey, nice job providing an example of what I wrote before! 😜  Pretty sure you're actually going to need to do it like this, because `not` will try to negate the whole array object while `~` can act elementwise:
```suggestion
        lorentzfactor = np.copy(lorentzfactor)
        lorentzfactor[~isfinite_lorentzfactor] = rbody.lorentz_factor()[
            ~isfinite_lorentzfactor
        ]
```

Does RelativisticBody even work with non-scalar array values?

StanczakDominik (dismissed):

changelog/1813.breaking.rst
```diff
@@ -0,0 +1 @@
+Changed the gyroradius function so that it takes relativistic effects into account.
```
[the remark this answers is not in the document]
MuhammadHMasood: yeah I called it breaking just in case somehow someone wanted to keep the exact same behavior before and after

plasmapy/formulary/lengths.py
MuhammadHMasood: > Functions that accept [array_like](https://numpy.org/doc/stable/glossary.html#term-array_like) or [Quantity](https://docs.astropy.org/en/stable/api/astropy.units.Quantity.html#astropy.units.Quantity) inputs should accept and return [nan](https://numpy.org/doc/stable/reference/constants.html#numpy.nan) ([not a number](https://en.wikipedia.org/wiki/NaN)) values.

Upon looking at the style guide I think leaving the output as NaN may have been intended so I will revert it to that.

MuhammadHMasood:
I think I got all the tests working that I physically can since the last 3 changes are because of the expected behavior changing 
At least with the use of tox there are only 3 left, and I'm pretty sure they don't work because of the tests not expecting the relativistic correction, and I don't know how to modify the tests correctly to fix that.

I guess i'll try figuring that out right now.

MuhammadHMasood:
Update: Yeah I played around with the margin of error and it works now, passes all the tests which heavily implies its just the relativity corrections, however I'm not sure what im supposed to do, should I modify the test itself?

changelog/1813.breaking.rst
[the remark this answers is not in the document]
StanczakDominik: Yeah, I absolutely see why! It's a tough question.

MuhammadHMasood:
Also I'm not really sure whats going on with the docs, because building them fails the first time but if the test is run again then it works (at least when I try it with tox locally).

MuhammadHMasood:
Alright everything should be good now, if it fails on the documentation building then rerunning the command will cause it to work, or at least that's how it turned out for me.

MuhammadHMasood:
@StanczakDominik I'm really sorry to bother you but I straight up don't understand whats happening with documentation, do you have any idea how I can stop it from failing. The docs build on my end just fine but for some reason it errors once and when I run the command again it works

changelog/1813.breaking.rst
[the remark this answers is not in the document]
namurphy: I think it'd be fine for the changelog to be either a feature or a breaking change, so my inclination would be to keep it as a breaking change...since that'd be less work!

changelog/1813.bugfix.rst
```diff
@@ -0,0 +1 @@
+Modified an error message to properly trigger if both `T` and `Vperp` (and `lorentzfactor`) are missing as well, instead of just when there are both of them.
```
namurphy, line 1:
```suggestion
Modified an error message in `~plasmapy.formulary.lengths.gyroradius` to
properly trigger if both ``T`` and ``Vperp`` (and ``lorentzfactor``) are 
missing as well, instead of just when there are both of them.
```
When single backticks are used, Sphinx tries to create a link to a Python object, like for `gyroradius` above (with the `~` indicating to hide the namespace prefixes). Changing them to double backticks will make it get formatted as literal text (like for a code snippet) without trying to create any links.

The error message that you were getting really is confusing!  I'm going to make a note to create an issue in the Sphinx repository to see if they can improve it.

changelog/1813.feature.rst
```diff
@@ -0,0 +1 @@
+Added `lorentzfactor` optional keyword depending on needs of user - can also be set to 1.0 if they want the old functionality back.
```
namurphy, line 1:
```suggestion
Added ``lorentzfactor`` optional keyword to `~plasmapy.formulary.lengths.gyroradius` depending on needs of user — can also be set to 1.0 if they want the old functionality back.
```

changelog/1813.trivial.rst
```diff
@@ -0,0 +1,2 @@
+changed Gyroradius::test_value 6 10 and 12 since it errors due to assuming the nonrelativistic case
+added more tests to check functionality for new relativistic features
```
namurphy, line 2: I'm wondering if we can delete this changelog entry, since the changes to tests can be considered part of the `breaking` changelog entry.

plasmapy/formulary/lengths.py
```diff
@@ -175,13 +184,13 @@ def gyroradius(
...
+        r_{Li} = \frac{\gamma V_{\perp}}{ω_{ci}}
 
     where :math:`V_⟂` is the component of particle velocity that is
-    perpendicular to the magnetic field and :math:`ω_{ci}` is the
-    particle gyrofrequency.  If a temperature is provided, then
+    perpendicular to the magnetic field, :math:`ω_{ci}` is the
+    particle gyrofrequency, and :math: `γ` is the lorentz factor.  If a temperature is provided, then
     :math:`V_⟂` will be the most probable thermal velocity of a
-    particle at that temperature.
+    particle at that temperature. :math: `γ` can manually be set to 1 to avoid the relativistic correction
```
namurphy, line 193:
```suggestion
    particle at that temperature. :math: `γ` can manually be set
    to 1 to avoid the relativistic correction.
```
...since there's a PEP 8 style guideline to keep docstrings and comments at ≤ 72 characters per line.

plasmapy/formulary/lengths.py
```diff
@@ -167,6 +173,9 @@ def gyroradius(
     -----
     One but not both of ``Vperp`` and ``T`` must be inputted.
 
+    If ``lorentzfactor`` is inputted then the previous two can be
+    omitted although at high energies this can lead to rounding errors
+
```
namurphy, line 178: In this case, the rounding errors occur when `Vperp` is provided, and it's really close to `c`.

plasmapy/formulary/lengths.py
```diff
@@ -193,16 +202,18 @@ def gyroradius(
...
     >>> gyroradius(400*u.G, particle='Fe+++', Vperp=1e7*u.m/u.s)
-    <Quantity 48.23129... m>
+    <Quantity 48.25815... m>
     >>> gyroradius(B=0.01*u.T, particle='e-', T=1e6*u.K)
     <Quantity 0.003130... m>
     >>> gyroradius(0.01*u.T, 'e-', Vperp=1e6*u.m/u.s)
     <Quantity 0.000568... m>
     >>> gyroradius(0.2*u.T, 'e-', T=1e5*u.K)
-    <Quantity 4.94949...e-05 m>
+    <Quantity 4.94957...e-05 m>
```
namurphy, line 211: Huh, this changed more than I thought it would!  Though I guess electrons really are going fast at $10^5$ K... 🤔

plasmapy/formulary/lengths.py
```diff
@@ -225,13 +236,27 @@ def gyroradius(
...
+        if not np.isscalar(lorentzfactor):
+            raise ValueError(
+                "Inferring velocity(s) from more than one lorentz factor is not currently supported"
+            )
+        Vperp = np.copy(Vperp)
+        rbody = RelativisticBody(particle, lorentz_factor=lorentzfactor)
+        Vperp[~isfinite_Vperp] = rbody.velocity
+    elif np.any(isfinite_lorentzfactor):
+        warnings.warn(
+            "lorentzfactor is given along with Vperp or T, will lead to inaccurate predicitions unless they correspond"
```
namurphy, line 258:
```suggestion
            "lorentzfactor is given along with Vperp or T, will lead to inaccurate predictions unless they correspond"
```

plasmapy/formulary/lengths.py
```diff
@@ -138,14 +140,18 @@ def gyroradius(
         The particle temperature in units convertible to kelvin.
         Note: Deprecated. Use ``T`` instead.
 
+    lorentzfactor : `float` or `~numpy.ndarray`, optional, |keyword-only|
+        The Lorentz factor for the particles, set to 1.0 if you want to use
```
namurphy, line 144:
```suggestion
        The Lorentz factor for the particles. Set to 1.0 if you want to use
```

changelog/1813.breaking.rst
namurphy, line 1:
```suggestion
Changed the `~plasmapy.formulary.lengths.gyroradius` function so that it takes relativistic effects into account.
```
... to add a link to `gyroradius` in the docs.

namurphy (commented):
Thank you for doing this...we really appreciate it!  My suggestions are mostly pretty minor, including one about the documentation errors.  

It's a good idea to include `lorentzfactor` as an argument, since that'll help avoid the rounding errors that could arise when `Vperp` is really close to $c$.  I'm wondering if we'd want to have a `relativistic` keyword that defaults to `True`, which might make the intent of the code more clear when people are using this function.

Thank you again!

changelog/1813.trivial.rst
MuhammadHMasood: Yeah that makes sense, I agree.

plasmapy/formulary/lengths.py
MuhammadHMasood: I re ran the command to make sure and yeah it is that big of a change, hopefully I didn't make a logic error

The lorentz factor was calculated to be: 1.0000168641271152126
and the gyroradius was:  4.949576847389539e-05 m

changelog/1813.bugfix.rst
MuhammadHMasood: Oh yeah based on the codebase guidance I reversed this change, it should just return nan.

MuhammadHMasood:
Holy smokes

MuhammadHMasood:
@StanczakDominik @namurphy Thank you so much for the help with my first commit. It say all checks has passed and I think all requests have been addressed, what now?

namurphy (approved):
Awesome!  I have a few very slight edits which I'll merge, and I'll set this pull request to auto-merge when it's ready. Thank you again!  I'm looking forward to trying this out for an example notebook!

MuhammadHMasood:
Nice, this was really fun and rewarding contributing to an open source project!

PR #1822, opened 2022-12-07
Remove pytest-xdist from test requirements

namurphy:
Right now we have an issue where tests run via `pytest-xdist` with more than one process are not able to be run because the test order is not deterministic (i.e., because sets or dicts are used when parametrizing tests).  See also #750.  This has apparently recently caused the tests to fail when they're run via GitHub Actions.

This PR removes `pytest-xdist` from our requirements as a way to (hopefully) resolve this issue.


StanczakDominik (approved):
OK *for now* and I'll fix the root cause. 👍🏻

PR #1827, opened 2022-12-08
Fix weekly test configuration error

namurphy:
Our weekly tests are failing because `3.10` was being treated as `3.1` in `.github/workflows/weekly.yml`.  This PR fixes that by making the `3.10` into a string.  

If only this was the first time I've introduced this error! 👀 😅 😹 🎈 🎆 🎂

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -27,7 +27,7 @@ jobs:
 
         - name: Python 3.10
           os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


namurphy:
This will need to be merged before we can check that it's working.  After it's merged we should try manually running the weekly tests.

StanczakDominik (approved):
Oh hey, this one again! I was just telling a story about it last week at work 😅

Issue #1728, opened 2022-09-06
Implement parallel processing for `charged_particle_radiography.Tracker`


pheuer:
Particle tracing in the `charged_particle_radiography.Tracker` class is an [embarrassingly parallelizable problem ](https://en.wikipedia.org/wiki/Embarrassingly_parallel), but the calculations are currently being performed in serial. Execution time is limited by processing the interpolation steps. Implementing parallel processing using [the multiprocessing library ](https://docs.python.org/3/library/multiprocessing.html) should allow a ~3-4x speed increase in tracing on laptops and very large increases on computing resources with more cores. This would significantly improve the module and make it feasible to run large batches of simulations for things like training machine learning algorithms.

Only the push step of the algorithm needs to be parallelized. The code should probably split particles into groups (#/ processing core) and then push them in parallel, stopping each process independently when the particles have left the grid(s). 

There are a few challenges to doing this: 
- The grids (from which fields for each particle are interpolated) are large (often >1 GB) and so replicating them in memory for each core is undesirable, especially on smaller computers. They are read-only, so it should definitely be possible to hold the grid in [shared memory](https://docs.python.org/3/library/multiprocessing.shared_memory.html) for use by all the processes. I'm not sure how this would work in detail, or whether it would be worth making this optional (maybe if the RAM is available, making copies of the grid will allow faster interpolation?). 
- The structure of the code needs minor modifications to make it compatible with parallel processing. The `_push` and `run` methods will need to be modified.

namurphy:
Do you think [dask](https://www.dask.org/) would be a suitable tool for this?

namurphy:
@MKastek also mentioned [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html).

pheuer:
@namurphy I have little experience with either, but this is a simple parallelization problem so I'm sure many libraries could work.  I've used multiprocessing recently, so I know that could work.

JaydenR2305:
I'll give this a try

I think dask would be the superior option since it allow both local and distributed computation with a high level API

pheuer:
@JaydenR2305 great, thanks for taking a look at it! As you do, keep in mind also issue #1725: whatever parallel solution we will also eventually want to extend to working with multiple grids. Something to think about when choosing an implementation.

Sobeskes:
Hello, I've been looking into this issue, and I would like to give it a try as my first contribution.

namurphy:
[...]

PR #1830, opened 2022-12-11
Modified the Tracker class in ~plasmapy/diagnostics/charged_particle_…

Sobeskes:
…radiography.py to run in parallel

## Description

Modified ~plasmapy/diagnostics/charged_particle_radiography.py to split array of simulated particles into partitions that are then run in parallel
Added tests to ~plasmapy/diagnostics/tests/test_charged_particle_radiography.py that test performance on extremely large inputs; extremely large tests are left commented to not dramatically slow down the test suite.

## Motivation and context

@pheuer identified this as an embarrassingly parallelizable problem in #1728. In theory, by parallelizing this step of the simulation we should see a significant increase in speed performance.

## Related issues
This issue was brought up in #1728


StanczakDominik:
Going to run pre-commit.ci on this first...

pre-commit.ci autofix

StanczakDominik (changes requested):
> In theory, by parallelizing this step of the simulation we should see a significant increase in speed performance.

In theory, sure! I'm not sure this approach is going to accomplish that - happy to be convinced with data, though! Did you try to benchmark this before and after this change? What kind of speedup, if any, did you get?

This is not really a "reject" based on "needing changes to the code", rather on needing to prove that this works as intended!

Sobeskes:
Absolutely, I totally understand! Right now, when testing on extremely large test cases, there seems to be a modest increase of speed, but definitely not what you'd hope to see. My classes are finishing up right now, but when I get more free time I'd be very interested in coming back to see if dask would be the right solution for this. In the mean time, I made a minor typo change in order to include the CI tests in my final report.

StanczakDominik:
That makes sense!

Would you mind, then, if I converted this PR to draft for now? I'm trying to be better about reviewing/merging stuff in a timely manner and it helps to keep that effort focused 😉

StanczakDominik:
I'm going to take that as a yes, we can always revert the draftiness 😉

Sobeskes:
Absolutely! Sorry for being slow with replying., finals season has been eventful.

StanczakDominik:
No worries whatsoever. Good luck with the finals! :)

Issue #1821, opened 2022-12-07
Calculate Thomson Spectral Density given scattering angle


pheuer:
### Feature description

Calculate spectral density given scattering angle directly.

### Motivation

The Thomson scattering spectral density function depends on the scattering angle (angle between the incoming and scattered light). In the current code, this is calculated from provided parameters `probe_vec` and `scatter_vec', which are unit vectors describing the geometry. This is convenient on larger experimental facilities (where the unit vectors of ports and beams are fixed and easy to look up). 

### Implementation strategy

 I've gotten requests to add an additional keyword `scattering_angle` that, if provided, directly provides the scattering angle. This keyword OR the existing probe_vec/scatter_vec keywords should be accepted, but not both at the same time. 

In addition to logic + validation that only one set of keywords is being used, the only line that should need to be updated is 

```
scattering_angle = np.arccos(np.dot(probe_vec, scatter_vec))
```

Where, if the scattering angle is provided explicitly, this can just be set directly rather than calculated.

tranqver:
I'd like to work on this. I assume this would require additional test inputs?

tranqver:
To check my understanding, this feature would only be added to ``spectral_density_lite``? 

In spectral_density_lite, the line ``k_vec = scatter_vec - probe_vec`` would need ``scatter_vec`` and ``probe_vec``. Would these values need to be initialized to their default values when not given?

pheuer:
I would: 

1) Change spectral_density_lite to only take the scattering_angle keyword. 
2) Add scattering_angle to spectral_density (keeping the other two keywords)
3) Add logic inside spectral_density that either takes the scattering angle provided or calculates it from the other keywords,
then passes that through to spectral_density_lite

suyashnaikare:
I would like to work on it

pheuer:
@suyashnaikare Excellent! Please see the ideas above - you can create your own branch of PlasmaPy, start implementing the feature, then open a pull request and request me as a reviewer. Your pull request does not need to be complete - just do what you can then we can discuss and suggest changes then.

pheuer:
Duplicating this comment here from #1831 


@JaydenR2305 and I looked into this, and I'm now thinking adding this keyword isn't such a good idea. The problem is in the lines here 


```python
scattering_angle = np.arccos(np.dot(probe_vec, scatter_vec))
```

```python
    # Compute the wavenumber shift (required by momentum conservation)
[...]

PR #1831, opened 2022-12-11
Add scattering_angle keyword to thomson.py spectral_density functions

tranqver:
## Description

Adds scattering_angle as an optional input to spectral_density. Users now have the option of providing probe_vec and scatter_vec, or scattering_angle. If provided both scattering_angle and probe_vec or scatter_vec it will throw a ValueError. 

Moves scattering angle out of spectral_density_lite, which will now take in scattering_angle as a required argument.

Adds tests to verify the functionality of using scattering_angle and verify that ValueErrors are thrown from invalid input combinations.

## Motivation and context

Scattering_angle is currently always being computed using probe_vec and scatter_vec, according according to the feature request adding a scattering_angle keyword has been a requested feature.

## Related issues

#1821


StanczakDominik:
Looks like linters are failing, let's deal with that first:

pre-commit.ci autofix

(you'll have to `git pull` locally!)

plasmapy/diagnostics/thomson.py
```diff
@@ -341,6 +334,9 @@ def spectral_density(
         Defaults to [0, 1, 0] which, along with the default ``probe_vec``,
         corresponds to a 90° scattering angle geometry.
 
+    scattering_angle : float
+        Defaults to np.arccos(np.dot(probe_vec, scatter_vec))
```
StanczakDominik, line 338: Same as above!

StanczakDominik (changes requested):

plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -246,6 +246,70 @@ def test_spectral_density_minimal_arguments(single_species_collective_args):
...
 
+def test_invalid_input_combination(single_species_collective_args):
+    """
+    Checks that exception gets raised if scattering_angle and vec_probe or scatter_vec
+    as only scattering_angle OR (vec_probe OR scatter_vec) can be given at once
+    """
+    args_fixture_copy1 = copy.copy(single_species_collective_args)
+    args_fixture_copy2 = copy.copy(single_species_collective_args)
+    # should raise error if scattering_angle, probe_vec and scatter_vec are given
+    args, kwargs = spectral_density_args_kwargs(single_species_collective_args)
```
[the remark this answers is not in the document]
tranqver: Thank you, I will separate the tests.

plasmapy/diagnostics/thomson.py
```diff
@@ -119,14 +120,8 @@ def spectral_density_lite(
...
-    probe_vec : float `~numpy.ndarray`, shape (3, )
-        Unit vector in the direction of the probe laser. Defaults to
-        ``[1, 0, 0]``.
-
-    scatter_vec : float `~numpy.ndarray`, shape (3, )
-        Unit vector pointing from the scattering volume to the detector.
-        Defaults to [0, 1, 0] which, along with the default ``probe_vec``,
-        corresponds to a 90 degree scattering angle geometry.
+    scattering_angle: float
+        np.arccos(np.dot(probe_vec, scatter_vec))
```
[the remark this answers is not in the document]
tranqver: I didn't mean to delete those, thanks for noticing!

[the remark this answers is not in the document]
tranqver: Where can I find a good description of scattering angle? I do not know much about plasma physics and I was hoping someone could help me with the documentation.

tranqver:
pre-commit.ci autofix

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
StanczakDominik: Well, looking below:

```
    For details, see "Plasma Scattering of Electromagnetic Radiation" by
    :cite:t:`sheffield:2011`. This code is a modified version of the
    program described therein.
    For a concise summary of the relevant physics, see Chapter 5 of
    the :cite:t:`schaeffer:2014` thesis.
```

But for us, pragmatically (and sorry, I wasn't as communicative as I had wanted to), I'm pretty sure this suffices:

> Pre-calculated angle between the vector pointing from the scattering volume defined by ``scatter_vec`` and the probe vector defined by ``probe_vec``. If provided, those two must not be provided.

I guess what I most wanted to convey is that... you had the right idea, honestly, but it's good to put what you wrote in words for the documentation! If someone wants the actual implementation, they can skim the code.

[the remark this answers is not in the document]
StanczakDominik: And we can, of course, ask @pheuer if this looks like it's enough :)

StanczakDominik (commented):
Solid rewrite on the tests! I still think we can simplify them a little bit more, but it's probably faster for me to show you. Mind if I push the change to your branch?

StanczakDominik (approved):
I misunderstood the way the tests were being setup a little, so my imagined simple fix isn't as simple after all. I'll still want to refactor this, but I won't hold this PR up for that.

@pheuer, if you want to take a look at it, go ahead, if not just hit merge :)

StanczakDominik:
Wow, codecov has been acting wonky lately. The check failure appears to be unrelated.

plasmapy/diagnostics/thomson.py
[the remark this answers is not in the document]
pheuer: For more details the Schaeffer thesis is very good and accessible. But, at the simplest level, the scattering angle is just the angle labeled 'chi' on the graphic on this wiki page: https://en.wikipedia.org/wiki/Thomson_scattering

plasmapy/diagnostics/thomson.py
```diff
@@ -128,6 +129,11 @@ def spectral_density_lite(
         Defaults to [0, 1, 0] which, along with the default ``probe_vec``,
         corresponds to a 90 degree scattering angle geometry.
 
+    scattering_angle: float
```
pheuer, line 132: As written now, it looks like this function can ONLY take the scattering angle. I think that's actually the right way to do it, since otherwise you need to validate that not all three are provided and we try to avoid validation in docstrings. So, I think you should remove `scatter_vec` and `probe_vec` from the docstring and function signature here and make 'scattering_angle` a required keyword argument

plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -246,6 +247,72 @@ def test_spectral_density_minimal_arguments(single_species_collective_args):
     return (alpha, wavelengths, Skw)
 
 
+def test_invalid_input_all_three(single_species_collective_args):
```
pheuer, line 250: Good job understanding my convoluted test structures! These tests look good: my only comment is that you could probably use `pytest.parameterize` to cover all these cases in just one function. That would reduce the amount of duplicated testing code.

plasmapy/diagnostics/thomson.py
```diff
@@ -341,6 +345,11 @@ def spectral_density(
         Defaults to [0, 1, 0] which, along with the default ``probe_vec``,
         corresponds to a 90° scattering angle geometry.
 
+    scattering_angle : float
```
pheuer, line 348: You could phrase this as `either scattering_angle` or both `probe_vec` and `scatter_vec` keywords must be provided.`

plasmapy/diagnostics/thomson.py
```diff
@@ -431,12 +445,24 @@ def spectral_density(
...
+    # if both scattering_angle and probe_vec or scatter_vec, raise error
+    if scattering_angle is not None and (
+        scatter_vec is not None or probe_vec is not None
+    ):
+        raise ValueError(
+            "Keywords scattering_angle and scatter_vec or probe_vec are both given."
+            "Give only scattering_angle or scatter_vec and probe_vec"
+        )
+
     if probe_vec is None:
```
pheuer, line 457: The more I think about it, having default values for these might be confusing. Now that we have the nice scattering angle keyword, I'd suggest removing these two default value blocks and just add a third block that ensures that at least scattering_angle OR probe_vec and scatter_vec are provided.

plasmapy/diagnostics/thomson.py
```diff
@@ -557,6 +583,7 @@ def spectral_density(
         electron_vel=electron_vel.to(u.m / u.s).value,
         probe_vec=probe_vec,
         scatter_vec=scatter_vec,
+        scattering_angle=scattering_angle,
```
pheuer, line 586: Following my other comments, you would then delete the probe_vec and scatter_vec keywords here and always just pass in the scattering angle (regardless of how it was calculated above)

plasmapy/diagnostics/tests/test_thomson.py
StanczakDominik: I certainly agree, but I'll want to do a pass like that soon anyway, so don't worry too much about it here :)

plasmapy/diagnostics/thomson.py
StanczakDominik: That is a great idea. Let's have this function do one thing and do it well, and the helper utility of calculating the angle from the vectors can be moved to the "top level" `spectral_density` function. Awesome!

pheuer:
@JaydenR2305 and I looked into this, and I'm now thinking adding this keyword isn't such a good idea. The problem is in the lines here 


```python
scattering_angle = np.arccos(np.dot(probe_vec, scatter_vec))
```

```python
    # Compute the wavenumber shift (required by momentum conservation)
    # Eq. 1.7.10 in Sheffield
    k = np.sqrt(ks**2 + kl**2 - 2 * ks * kl * np.cos(scattering_angle))
    # Normal vector along k
    k_vec = scatter_vec - probe_vec
    k_vec = k_vec / np.linalg.norm(k_vec)

    # Compute Doppler-shifted frequencies for both the ions and electrons
    # Matmul is simultaneously conducting dot products over all wavelengths
    # and ion components
    w_e = w - np.matmul(electron_vel, np.outer(k, k_vec).T)
    w_i = w - np.matmul(ion_vel, np.outer(k, k_vec).T)
```

If the only use of `scatter_vec` was to calculate `k`, then the new keyword would make sense. But they are also used to defined `k_vec`, which then allows us to accept vector electron and ion velocity arguments. If vectors for the probe and scattering vector aren't supplied, I don't think we could support flow velocities (or we'd have to assume something about their direction). That seems to me like a significant drawback.

pheuer:
Per discussion in #1821, I'm closing this PR.

PR #1833, opened 2022-12-12
Remove `plasmapy.formulary.parameters`

namurphy:
We had deprecated `plasmapy.formulary.parameters` in 0.8.1, after the Great PlasmaPy Formulary Refactoring of 2022. 🎩  We had said with a deprecation warning that it was going to be removed in 0.9.0, though we didn't get around to it.  This PR will remove the module for our first release of 2023.

Instead of using `formulary.parameters`, users should use the modules based on physical types: `formulary.lengths`, `formulary.frequencies`, etc.


StanczakDominik (approved):
Parameters begone!

PR #1834, opened 2022-12-13
Fixed functional typo in testing_guide.rst

svincena:
Minor/trivial change to the testing guide.
I was reading the sample code output, but the sample output did do what the sample code said it should do.
Reversed order in reverse_order 'if' statement (a+b) -> (b+a)


StanczakDominik:
I can see how that might have been missed 😁

PR #1836, opened 2022-12-16
Ensure ordering for particles in tests

StanczakDominik:
## Description

- I wrapped most of the lists of particles that go into `pytest.mark.parametrize` into `list(sorted(...))`. Turns out that's enough. Not entirely sure why, tbh. See https://pytest-xdist.readthedocs.io/en/latest/known-limitations.html#workarounds for where I found this trick.
- I reverted the removal of pytest-xdist.

## Motivation and context

https://github.com/StanczakDominik/PlasmaPy/actions/runs/3701946161 was just one example of the recently happening test failure resulting from parameters for tests in the particle subpackage being "insufficiently ordered", as it turns out. See commit message for the first one, because I'm still pretty confused about it. We previously removed pytest-xdist from dependencies as a temporary measure in https://github.com/PlasmaPy/PlasmaPy/pull/1822, where I promised to fix the root cause. Well, here it is.

### DIFF
--- a/plasmapy/particles/tests/test_parsing.py
+++ b/plasmapy/particles/tests/test_parsing.py
@@ -331,16 +331,17 @@ def test_parse_InvalidParticleErrors(arg, kwargs):
         )
 
 
-@pytest.mark.parametrize("arg", particle_zoo.everything - {"p+"})
-def test_parse_InvalidElementErrors(arg):
+def test_parse_InvalidElementErrors(particle):
     r"""Tests that _parse_and_check_atomic_input raises an
     InvalidElementError when the input corresponds to a valid
     particle but not a valid element, isotope, or ion."""
+    if particle == Particle("p+"):
+        return  # TODO verify this case should still be a skip
     with pytest.raises(InvalidElementError):
-        parse_and_check_atomic_input(arg)
+        parse_and_check_atomic_input(particle)
         pytest.fail(
             "An InvalidElementError was expected to be raised by "
-            f"{call_string(parse_and_check_atomic_input, arg)}, "
+            f"{call_string(parse_and_check_atomic_input, particle)}, "
             f"but no exception was raised."
         )
 
--- a/plasmapy/particles/tests/test_parsing.py
+++ b/plasmapy/particles/tests/test_parsing.py
@@ -336,7 +336,7 @@ def test_parse_InvalidElementErrors(particle):
     InvalidElementError when the input corresponds to a valid
     particle but not a valid element, isotope, or ion."""
     if particle == Particle("p+"):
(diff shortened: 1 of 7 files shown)


namurphy (approved):

StanczakDominik:
I still have a few tset failures because I was too hasty with the fixturizing

StanczakDominik:
I still have a few tset failures because I was too hasty with the fixturizing

namurphy:
> I still have a few tset failures because I was too hasty with the fixturizing

Oops!  I hastily clicked approve before I checked that the tests were passing since I need to leave for a dentist appointment in a minute.  But: it all looks good, and if there are more than minor changes, I can look it over again in a few kiloseconds.

StanczakDominik:
Might take a bit longer. I found another failure that, well... I want to write these tests properly this time. ^^'

StanczakDominik:
I also removed a couple tests that I thought were doing something checked elsewhere. Let ~dog~ codecov sort them changes out, I say!

StanczakDominik:
Think we could merge this, @namurphy?

plasmapy/particles/tests/test_parsing.py
```diff
@@ -331,16 +331,17 @@ def test_parse_InvalidParticleErrors(arg, kwargs):
...
 
 
-@pytest.mark.parametrize("arg", particle_zoo.everything - {"p+"})
-def test_parse_InvalidElementErrors(arg):
+def test_parse_InvalidElementErrors(particle):
     r"""Tests that _parse_and_check_atomic_input raises an
     InvalidElementError when the input corresponds to a valid
     particle but not a valid element, isotope, or ion."""
+    if particle == Particle("p+"):
+        return  # TODO verify this case should still be a skip
```
namurphy, line 339:
```suggestion
    if particle == Particle("p+"):
        return
```
Yep, protons should still be skipped here.  The reason is that protons are the only particles that count as both a special particle and an element, so this test doesn't apply to them.

tox.ini
```diff
@@ -12,7 +12,7 @@ allowlist_externals =
 setenv =
     MPLBACKEND = agg
     COLUMNS = 180
-    PYTEST_COMMAND = pytest --pyargs plasmapy --durations=25 --showlocals
+    PYTEST_COMMAND = pytest --pyargs plasmapy --durations=25 --showlocals -n=auto --dist=loadfile
```
namurphy, line 15: Awseome!  🎉 🎆 ☄️

namurphy (approved):
I have one minor suggestion to remove a `TODO` comment`, and otherwise I'm happy for this to be merged.

StanczakDominik:
pre-commit.ci autofix

PR #1845, opened 2022-12-21
Apply changes from `refurb` and `ruff` to `plasmapy.utils`

namurphy:
This PR applies changes from [`refurb`](https://github.com/dosisod/refurb) and [`ruff`](https://github.com/charliermarsh/ruff) to `plasmapy.utils`.  Right now both of these will pass so long as we run ruff with `--ignore F401,E501` where `F401` corresponds to unused imports (which are to be expected in `__init__.py` files) and `E501` corresponds to line length violations which are handled by `black` anyway.


namurphy:
I think the reason why there are so many changed lines in certain files is because running `ruff` on them cause the line endings to change between `LF` and `CRLF`.  Yeesh.  If we decide to add `ruff` to `pre-commit` (which would probably be helpful) then we'll probably end up having such changes on a lot of our files, so I suppose we might as well get it over with for `utils`.

I miss the good old days one month ago when I didn't know that Windows and macOS/Linux had different default line endings.\n

plasmapy/utils/data/tests/test_downloader.py
```diff
@@ -26,20 +27,20 @@ def test_get_file(filename, expected, tmp_path):
 
     """
     # Delete file if it already exists, so the test always downloads it
-    dl_path = os.path.join(tmp_path, filename)
-    if os.path.exists(dl_path):
-        os.remove(dl_path)
+    dl_path = tmp_path / filename
+    if dl_path.exists():
+        dl_path.unlink()
```
StanczakDominik, line 32: 🤯

plasmapy/utils/pytest_helpers/pytest_helpers.py
```diff
@@ -368,7 +365,7 @@ def test_func(inputs):
     if expected["result"] is None:
         return None
 
-    if type(result) != type(expected["result"]):
+    if type(result) != type(expected["result"]):  # noqa: E721
```
StanczakDominik, line 368: Guessing it wanted to use `isinstance`, or something along these lines?

StanczakDominik (approved):

PR #1854, opened 2022-12-23
Add `ruff` to `pre-commit` configuration

namurphy:
This PR is the final in the series of PRs to get [`ruff`](https://github.com/charliermarsh/ruff) ~~and [`refurb`](https://github.com/dosisod/refurb)~~ applied to the repo and into our `pre-commit` suite.  

`ruff` is awesome in that it runs really quickly (since it's written in Rust) and automagically applies a lot of fixes that would be found by `flake8`. We might have some duplication between `flake8` and `ruff`, but it'd be helpful to keep both for the time being.

~~A less nice thing about `refurb` is that it takes a long time to run compared to other pre-commit hooks. However, some of the fixes that refurb caught were pretty important.  In particular, it caught a couple of cases where there were two tests named the same thing, which probably meant that the first of those tests were not being run.~~

We should also keep in mind if there are any `ruff` ~~or `refurb`~~ error codes that we'd want to ignore.

### DIFF
--- a/.github/workflows/citation_updater.py
+++ b/.github/workflows/citation_updater.py
@@ -43,7 +43,7 @@ def update_citation_files(args):
     citation_rst_file = pathlib.Path(args.citation_rst_file)
     citation_rst_text = citation_rst_file.read_text()
 
-    for source_regex, target_value in [
+    for source_regex, target_value in (
         (
             r"\|version_to_cite\| replace:: (.*)",
             f"|version_to_cite| replace:: {args.version}",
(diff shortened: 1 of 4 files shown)


namurphy:
Oh, cupcakes.  `refurb` can only be run on Python ≥3.10, so we should not put it into our pre-commit configuration.  I'm wondering if it could go in the `linters` tox env.  But...for this PR...I'm going to drop `refurb`.

StanczakDominik (approved):

PR #1852, opened 2022-12-22
Pin GitHub Action to specific version for posting code review checklists

namurphy:
## Description

This PR pins the GitHub Action used for posting the code review checklist to v0.1.0 instead of the development branch.

## Motivation and context

Whilst hanging out on another repository, I heard some good security advice that it is best to pin GitHub Actions to specific releases.  This is to prepare for the unlikely eventuality that malicious code ends up getting put into a development branch.  Pinning it to a specific release is better than pinning it to a particular commit also (because a force push could have the same commit hash).  

## Related issues

#1729

PR #1855, opened 2022-12-23
Fix version of GitHub Action

namurphy:
This PR fixes the error I introduced in #1852 (oops!) by reverting the version of `actions/checkout` to `v3` instead of the incorrect `v0.1.0`.


namurphy:
I used admin superpowers to merge this so that I can make sure it works.

PR #1860, opened 2022-12-24
Adjust ruff for pre-commit

StanczakDominik:
`ruff`, added in #1854, was missing the `--fix` flag. We can use that one to have pre-commit.ci autofix work as expected.


namurphy:
Good catch!

namurphy (approved):

PR #1864, opened 2023-01-03
Bring back dependabot

### LINKED ISSUE #1861: Bring back Dependabot
The recent failures of CI in multiple places - https://github.com/PlasmaPy/PlasmaPy/commit/f36b7d766e998bc595bb6675a2f2d2208e4fbd8c, https://github.com/PlasmaPy/PlasmaPy/commit/9684bc8f5b834ecfb7360aec59695d1f49196fc5, and whichever one pinned tox to <4, show pretty unambiguously that we have a problem with versioning of our dependencies.

I've also heard recently that Dependabot no longer spams forks. I can't quite give you a source on that. But I'm willing to try.

So, my plan is:

- pin every dependency we have to <= what the newest version existing right now is
    * [ ] how does that work with dev dependencies?
- reenable dependabot and keep the upper bounds current

Then when something like pytest-xdist or tox or whatever breaks our CI, we know about it from the get-go.

I'll happily do it, but wanted to discuss first.

StanczakDominik:
This closes #1861, where the motivation is also laid out. The setup follows https://www.b-list.org/weblog/2022/may/13/boring-python-dependencies/, https://iscinumpy.dev/post/bound-version-constraints/ and https://github.com/dependabot/dependabot-core/issues/3290.

To quote from that last link there:

> as long as dependabot recommends lockfiles, and Python doesn’t have a standard lockfile format, it makes sense to say “we recommend using poetry.lock, requirements.txt AS LOCKFILES”, while of course the recommended format for Python package metadata remains PEP 622.

The lockfile concept is the answer to all my doubts from #1861. Essentially, I was right upper pinning dependencies is problematic; but it is not the only way to ensure tests are ran with a particular set of dependencies. Instead, tools (classically poetry, more recently `pip-compile`) can inspect requirements,  e.g. lower and upper bounds on package versions, and create a lockfile, which is basically "at this time, the most recent set of packages that solves all of these constraints". A single point in dependency space right at the bound of the allowed region.

Essentially, the process goes like:
- specify dependencies with lower bounds and upper bounds in `pyproject.toml`
- use `pip-compile --resolver=backtracking --all-extras` from [`pip-tools`](https://pypi.org/project/pip-tools/) to generate a `requirements.txt` file that pins everything to specific versions, here's an excerpt:
```
#
# This file is autogenerated by pip-compile with Python 3.10
# by the following command:
#
#    pip-compile --all-extras --resolver=backtracking
#
alabaster==0.7.12
    # via sphinx
anyio==3.6.2
    # via jupyter-server
argon2-cffi==21.3.0
    # via jupyter-server
argon2-cffi-bindings==21.2.0
    # via argon2-cffi
asteval==0.9.28
    # via lmfit
astor==0.8.1
    # via flake8-simplify
astropy==5.2
    # via plasmapy (setup.py)
```
- run our 3.10 test suite off of dependencies specified in that file
- enable dependabot to keep checking for updates in that requirements.txt file.

This might need a "merge first, adjust later" approach for the actual dependabot part.


.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+
+updates:
+
+  # Maintain dependencies for GitHub Actions
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
```
StanczakDominik, line 9: Figured we might as well add these, too.

StanczakDominik:
The pre-commit.ci failure seems unrelated. I think we can remove it from `required` temporarily.

.github/dependabot.yml
namurphy: Yay for automation!

requirements.txt
```diff
@@ -0,0 +1,525 @@
...
+    # via plasmapy (setup.py)
+docutils==0.17.1
+    # via
+    #   nbsphinx
+    #   plasmapy (setup.py)
+    #   pybtex-docutils
+    #   restructuredtext-lint
+    #   sphinx
+    #   sphinx-rtd-theme
+    #   sphinxcontrib-bibtex
```
namurphy, line 84: It looks like there are some `docs` dependencies mixed in here.  Perhaps we could remove those for the Python 3.10 tests.

tox.ini
```diff
@@ -43,7 +43,7 @@ commands = coverage erase
 
 [testenv:build_docs]
 changedir = {toxinidir}
-extras = docs
+deps = -r{toxinidir}/requirements.txt
```
namurphy, line 46: If we wanted to have a separate lockfile for the doc build, we could put it at `docs/requirements.txt` instead.

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  # Maintain dependencies for python
+- package-ecosystem: pip
+  directory: /
+  schedule:
+    interval: daily
```
namurphy, line 15:
```suggestion
    interval: weekly
```
Judging from RSS feeds of packages we depend on, a daily interval could end up leading to a few dependabot PRs per week, so I'm wondering if weekly or monthly would be preferable.  Daily might be a good way to start things out to make sure it's working though.

namurphy:
Should we take this approach, then we might want to update the list of CI checks that are required before merging, and not require the other tests that might have spontaneous failures due to upstream package changes.  That'd make it so that weird spontaneous failures are less likely to block PRs from being merged.  

Thank you for doing this!

requirements.txt
StanczakDominik: Yeah, I'm not sure how best to do this:
- Using two lockfiles means they might, theoretically, get out of sync (if a doc dependency has an upper bound on a test dependency, for example).
- But using a single one with `--all-extras` means the install get larger (which could be sidestepped to a degree by finally setting up proper caching for test envs in github actions).
- Using a single lockfile just for the test extra means we aren't catching any doc issues. That's still helpful, then, just less so.

.github/dependabot.yml
StanczakDominik: I did not originally consider either of these points... but they're both good, so let's go with weekly for now 😁 
```suggestion
    interval: daily  # TODO s/daily/weekly once dependabot has settled
```

StanczakDominik:
- [ ] try out two lockfiles after release

StanczakDominik:
I removed temporarily the requirement for the Documentation job, as it's been renamed.

tox.ini
```diff
@@ -69,6 +78,10 @@ setenv =
 commands =
     sphinx-build -D nbsphinx_execute='never' docs docs{/}_build{/}html -b html {posargs}
 
+[testenv:py310-pins]
```
StanczakDominik, line 81: works?

StanczakDominik:
- [ ] requirements.txt file that combines the two

```
requirements/docs.txt
requirements/tests.txt
requirements.txt
```

```
-r requirements/docs.txt
-r requirements/tests.txt
```

StanczakDominik:
Welp. I tried installing the combining requirements.txt file and it already has issues on docutils:
```
docs.txt:docutils==0.17.1
tests.txt:docutils==0.19
```
So let's go with a single file.

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+
+  # Maintain dependencies for python
+- package-ecosystem: pip
+  directory: /
+  schedule:
+    interval: daily  # TODO s/daily/weekly once dependabot has settled
```
namurphy, line 15: It's good to start out with daily for the time being, while making sure that dependabot is working as expected.

I had been leaning towards going with monthly in the long-term, but the more I think about it, weekly would probably be best.  In particular, having this be monthly would increase the chances of us having to deal with more than one thing breaking in a single PR.  And too often in my experience...doubling the problems that we have to deal with more than doubles the effort needed to fix the combined problems.

changelog/1864.trivial.rst
```diff
@@ -0,0 +1 @@
+Brought back Dependabot and turned the root-level ``requirements.txt`` into a lockfile for CI purposes.
```
namurphy, line 1:
```suggestion
Brought back Dependabot and turned the root-level :file:`requirements.txt` into a lockfile for CI purposes.
```

namurphy (approved):
LGTM! I only have one minor reST edit as a suggestion for this PR.  I'm happy to try it out!

PR #1866, opened 2023-01-03
Expand charge number functionality for `CustomParticle`

namurphy:
## Description

I made the following changes to `CustomParticle`:

- Added `Z` as a keyword-only parameter (which can be real, and not just an integer)
- Added the `charge_number` attribute
- Updated the serialization functionality

## Motivation and context

Long story short, this is to make it easier to change `particle_input` and the particle factory function to allow non-integer charges, which will allow `particle_input` to decorate functions that have `Z_mean` as a parameter, and thus let us deprecate `Z_mean` in favor of `Z`.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -2155,20 +2155,22 @@ class CustomParticle(AbstractPhysicalParticle):
             'module': 'plasmapy.particles.particle_class',
             'date_created': '...',
             '__init__': {'args': (), 'kwargs': {'mass': '5.12 kg', 'charge': '6.2 C',
-            'symbol': 'ξ'}}}}
+            'charge_number': '3.869735626...e+19', 'symbol': 'ξ'}}}}
         >>> import pytest
         >>> custom_particle = CustomParticle(mass=1.5e-26 * u.kg)
         >>> custom_particle.json_dict
         {'plasmapy_particle': {'type': 'CustomParticle',
             'module': 'plasmapy.particles.particle_class',
             'date_created': '...',
-            '__init__': {'args': (), 'kwargs': {'mass': '1.5e-26 kg', 'charge': 'nan C',
+            '__init__': {'args': (), 'kwargs': {'mass': '1.5e-26 kg',
+            'charge': 'nan C', 'charge_number': 'nan',
             'symbol': 'CustomParticle(mass=1.5e-26 kg, charge=nan C)'}}}}
         """
         particle_dictionary = super().json_dict
         particle_dictionary["plasmapy_particle"]["__init__"]["kwargs"] = {
             "mass": str(self.mass),
             "charge": str(self.charge),
+            "charge_number": str(self.charge_number),
             "symbol": self.symbol,
         }
         return particle_dictionary
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -23,7 +23,7 @@ from abc import ABC, abstractmethod
 from collections import defaultdict, namedtuple
 from datetime import datetime
 from numbers import Integral, Real
-from typing import Iterable, NoReturn, Optional, Sequence, Set, Union
+from typing import Dict, Iterable, NoReturn, Optional, Sequence, Set, Union
 
 from plasmapy.particles import _elements, _isotopes, _parsing, _special_particles
 from plasmapy.particles.exceptions import (
@@ -2079,9 +2079,9 @@ class CustomParticle(AbstractPhysicalParticle):
         self,
         mass: u.kg = None,
         charge: u.C = None,
-        symbol=None,
+        symbol: Optional[str] = None,
         *,
-        Z: Real = None,
+        Z: Optional[Real] = None,
     ):
 
         if isinstance(charge, Real):
@@ -2107,7 +2107,7 @@ class CustomParticle(AbstractPhysicalParticle):
                 f"{mass} and a charge of {charge}."
             ) from exc
 
-    def __repr__(self) -> str:
+    def __repr__(self):
         """
         Return a string representation of a |CustomParticle|.
 
@@ -2136,7 +2136,7 @@ class CustomParticle(AbstractPhysicalParticle):
         )
 
     @property
-    def json_dict(self) -> dict:
+    def json_dict(self) -> Dict[str, str]:
         """
         A `json` friendly dictionary representation of the |CustomParticle|.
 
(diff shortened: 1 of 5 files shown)


plasmapy/particles/particle_class.py
```diff
@@ -2023,9 +2023,11 @@ class CustomParticle(AbstractPhysicalParticle):
...
         The electric charge of the custom particle.  If provided as a
         `~astropy.units.Quantity`, then it must be in units of electric
-        charge.  If provided as a real number, then it is treated as the
-        ratio of the charge to the elementary charge. Defaults to |nan|
-        C.
+        charge. Defaults to |nan| C.
+
+    Z : ~numbers.Real, optional, |keyword-only|
+        The :term:`charge number`, which is equal to the ratio of the
+        charge to the elementary charge.
```
StanczakDominik, line 2030: So, wait, the only difference between the two is that Z can take floats, but charge can take quantities? Otherwise these look the same, conceptually, for the end user. Are we sure this separation is necessary?

StanczakDominik: I'm probably not remembering some earlier discussion...

namurphy: Right now there's potential for ambiguity when `charge` is given a real number.  In a lot of our functionality, if an argument expected to be a `Quantity` is instead a real number, it will often be assumed that the real number is in SI units (along with a warning).  Separating `Z` from `charge` will remove that ambiguity.  

We use `Z` (or in some places, `Z_mean`) as the charge number pretty much everywhere else, so having `Z` mean charge number here would make it a lot more consistent with the rest of our functionality.

I'm also finding it more convenient when working with `particle_input` and the particle factory function in #1867 to have `CustomParticle` accept `Z` and have it mean the same thing as for a `Particle`.  

Anyway, these are my thoughts for this!

StanczakDominik: Right, thanks for the reminders! 😅 That does make sense.

plasmapy/particles/particle_class.py
```diff
@@ -2053,30 +2055,46 @@ class CustomParticle(AbstractPhysicalParticle):
...
+        charge: u.C = None,
+        symbol: Optional[str] = None,
+        *,
+        Z: Optional[Real] = None,
+    ):
+
+        if isinstance(charge, Real):
+            warnings.warn(
+                "Providing a real number to 'charge' is deprecated. Use "
+                "'Z' as a keyword argument instead.",
```
StanczakDominik, line 2088:
```suggestion
                "Providing a real number to 'charge' is deprecated. To specify the charge as a multiple of the elementary charge, use "
                "'Z' as a keyword argument instead, or pass e.g. `2 * u.e` into `charge`.",
```
will ofc need a pre-commit autofix for line length but you get the drift. Makes things more explicit for how we want people to use this.

plasmapy/particles/particle_class.py
```diff
@@ -2053,30 +2055,46 @@ class CustomParticle(AbstractPhysicalParticle):
...
+    def __init__(
+        self,
+        mass: u.kg = None,
+        charge: u.C = None,
+        symbol: Optional[str] = None,
+        *,
+        Z: Optional[Real] = None,
+    ):
+
+        if isinstance(charge, Real):
```
StanczakDominik, line 2085:
```suggestion
        # TODO py3.10 replace ifology with structural pattern matching
        if isinstance(charge, Real):
```

It's not that far out, let's be prepared!

plasmapy/particles/particle_class.py
```diff
@@ -2116,7 +2134,7 @@ def __repr__(self) -> str:
         )
 
     @property
-    def json_dict(self) -> dict:
+    def json_dict(self) -> Dict[str, str]:
```
StanczakDominik, line 2137: I appreciate this addition of more type hinting!

plasmapy/particles/particle_class.py
```diff
@@ -2087,7 +2105,7 @@ def __init__(self, mass: u.kg = None, charge: (u.C, Real) = None, symbol=None):
                 f"{mass} and a charge of {charge}."
             ) from exc
 
-    def __repr__(self) -> str:
+    def __repr__(self):
```
StanczakDominik, line 2108: I'm confused about this removal of type hinting. Wasn't it fine before?
```suggestion
    def __repr__(self) -> str:
```

plasmapy/particles/particle_class.py
```diff
@@ -2172,6 +2192,12 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
             warnings.warn(
                 f"CustomParticle charge set to {q} times the elementary charge."
             )
+            warnings.warn(
+                "The ability to provide a real number to the 'charge' "
+                "parameter to has been deprecated and will be removed "
+                "in a future release of PlasmaPy. Use 'Z' instead.",
+                PlasmaPyDeprecationWarning,
```
StanczakDominik, line 2199: [Optional] could also do a similar note as before here. But it's arguably less important as the particle creation is the most important point of contact for people for this class.

StanczakDominik (approved):
Looks good! A few minor suggestions.

- [ ] if you accept them remember to autofix this via precommit!

plasmapy/particles/particle_class.py
namurphy: It was.  I was thinking that `__repr__` methods will always return strings so the type hint was redundant, but I'm happy to put it back since explicit is better than implicit.

plasmapy/particles/particle_class.py
namurphy: Good point!  I wonder if this is something that these new automagical refactoring tools would be able to handle by that time.

plasmapy/particles/particle_class.py
namurphy: Sounds good!  I made a slight change of `.si` for the elementary charge because of how `astropy.units` handles electromagnetism.

plasmapy/particles/particle_class.py
```diff
@@ -2172,6 +2187,13 @@ def charge(self, q: Optional[Union[u.Quantity, Real]]):
             warnings.warn(
                 f"CustomParticle charge set to {q} times the elementary charge."
             )
+            warnings.warn(
+                "Providing a real number to 'charge' is deprecated. To "
+                "specify the charge as a multiple of the elementary "
+                "charge, use 'Z' as a keyword argument instead, or pass, "
+                "e.g., '2 * astropy.constants.e.si' into 'charge'.",
+                PlasmaPyDeprecationWarning,
+            )
```
namurphy, line 2196: I realized that this warning was issued twice, so I consolidated it here.

plasmapy/particles/particle_class.py
StanczakDominik: Oh, that thing! I always forget about it. Thanks!

Issue #1865, opened 2023-01-03
Add tests for usage of `CustomParticle` and `ParticleList` in `formulary` functions


namurphy:
### Feature description

We should add tests for functions in `plasmapy.formulary` that decorated with `particle_input` that can accept `CustomParticle`-like and `ParticleList`-like arguments, in addition to the `Particle`-like arguments.

### Motivation

After the improvements to `particle_input` in #1057, functions decorated with `particle_input` can now nominally accept `CustomParticle`, `ParticleList`, and even some `Quantity` arguments for appropriately annotated parameters.  However, this new functionality is untested, so we can't be confident that it actually works.  Adding tests will give us confidence that things are working, and also help us figure out edge cases that need special handling (i.e. collisions functionality).

### Implementation strategy

This issue is dependent upon #341, in that the tests would only be applicable for functions that have been decorated with `particle_input`.  

For functions that have been decorated with `particle_input`, we should add tests for inputs that include:

 - `CustomParticle`
 - `Quantity` of physical type mass or charge, as appropriate 
 - `ParticleList`
 - `ParticleList`-like (i.e., a `list` like `["p+", "He-4 2+"]`)

### Additional context

The following are function in `plasmapy.formulary` that have particle inputs.  They should be checked off when they have the tests as described above.

 - [ ] Alfven_speed
 - [ ] Coulomb_logarithm
 - [ ] Hall_parameter
 - [ ] Knudsen_number
 - [ ] Lundquist_number
 - [ ] MaxwellianCollisionFrequencies
 - [ ] Maxwellian_1D 
 - [ ] Maxwellian_speed_1D
 - [ ] Maxwellian_speed_2D
 - [ ] Maxwellian_speed_3D
 - [ ] Maxwellian_velocity_2D
 - [ ] Maxwellian_velocity_3D
 - [ ] RelativisticBody
 - [ ] SingleParticleCollisionFrequencies
 - [ ] Spitzer_resistivity
 - [ ] cold_plasma_permittivity_LRP
 - [ ] cold_plasma_permittivity_SDP
 - [ ] collision_frequency
 - [ ] coupling_parameter
 - [ ] critical_density
 - [ ] deBroglie_wavelength
 - [ ] electron_thermal_conductivity
 - [ ] electron_viscosity
 - [ ] force_drift
 - [ ] fundamental_electron_collision_freq
 - [ ] fundamental_ion_collision_freq
 - [ ] gyrofrequency
[...]

PR #1867, opened 2023-01-04
Enable `particle_input` to return a `CustomParticle` for non-integer `Z`

namurphy:
## Description

This PR changes the private particle factory function used by `particle_input` so that it accepts a value of the charge number `Z` that is a real number but not an integer.  After this PR, `particle_input` will create a `CustomParticle` with such a value for `Z`.

## Motivation and context

There are a number of functions that accept something like `Z_mean`, which thus far hasn't been handled by `particle_input` and a lot of our particle functionality.  When `Z_mean` is accepted by a function, there often ends up being special code to handle it (like in [Alfven_speed](https://github.com/PlasmaPy/PlasmaPy/blob/a9701300b97b636950ff581542aa03774aa5693d/plasmapy/formulary/speeds.py#L167)) which can complexify functions.  If we have `particle_input` handle situations like that, then we'll be able to simplify a bunch of functions.

## Related issues

This depends on #1866 being merged first.  Related to #341, #1057, and #1865.

### DIFF
--- a/plasmapy/particles/_factory.py
+++ b/plasmapy/particles/_factory.py
@@ -8,14 +8,43 @@ __all__ = []
 
 import astropy.units as u
 import contextlib
+import numpy as np
 
 from astropy.units.physical import electrical_charge, mass
-from numbers import Integral
-from typing import Union
+from numbers import Integral, Real
+from typing import Any, Optional, Union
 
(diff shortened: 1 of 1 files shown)


namurphy:
Superseded by #1884.

PR #1868, opened 2023-01-04
Use `pyproject-fmt` instead of `pretty-format-toml`

namurphy:
## Description

This PR adds a pre-commit hook using [`pyproject-fmt`](https://pyproject-fmt.readthedocs.io/en/latest/) and drops the pre-commit hook using `pretty-format-toml`.  

## Motivation and context

It looks like the only TOML file that we have is our `pyproject.toml`, so it makes sense to use a tool that has more specific and standardized formatting for `pyproject.toml`s.  (Or is it `pyprojects.toml`?  I don't know what the plural of `pyproject.toml` is.)

We're also running into a problem with our pre-commit config due to a new inconsistency upstream of `pretty-format-toml`.  Removing `pretty-format-toml` will address this problem (though that issue will probably get fixed in a few days anyway).

This time I did *not* hear about this tool from a podcast!  I read about it in the show notes of a podcast instead. 😬

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
   "setuptools>=61.2",
   "setuptools-scm",
   "wheel>=0.29",
-]  # ought to mirror 'requirements/build.txt'
+]
 
 [project]
 name = "plasmapy"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,16 +1,46 @@
 [build-system]
-requires = ["setuptools>=61.2",
-            "setuptools-scm",
-            "wheel >= 0.29.0"]  # ought to mirror 'requirements/build.txt'
 build-backend = "setuptools.build_meta"
+requires = [
+  "setuptools>=61.2",
+  "setuptools-scm",
+  "wheel>=0.29",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,16 +1,46 @@
 [build-system]
-requires = ["setuptools>=61.2",
-            "setuptools-scm",
-            "wheel >= 0.29.0"]  # ought to mirror 'requirements/build.txt'
 build-backend = "setuptools.build_meta"
+requires = [
+  "setuptools>=61.2",
+  "setuptools-scm",
+  "wheel>=0.29",
+]  # ought to mirror 'requirements/build.txt'
```
StanczakDominik, line 7:
```suggestion
]
```

Oh look, someone forgot to remove this part! 😅

StanczakDominik (approved):

pyproject.toml
namurphy: And looks like someone forgot to point this out in code review too! 😅

namurphy:
Thank you for the quick review!  Happy to get this in to make pre-commit happy.

PR #1869, opened 2023-01-04
Decorate `gyrofrequency` with `particle_input`

namurphy:
## Description

This PR applies `particle_input` to `plasmapy.formulary.frequencies.gyrofrequency`,and also makes some updates to the docstring.  It also removes that the charge number defaults to 1, since explicit is better than implicit, and it now allows the mass number to be provided too.

## Motivation and context

These changes will enable 

## Related issues

This addresses approximately 2% of #341.

PR #1874, opened 2023-01-06
Add separate constructor for `CustomParticle` and improve |ParticleList| compatibility with |Quantity| objects

### LINKED ISSUE #1872: Make `ParticleList` compatible with `Quantity` objects with physical type of mass or electrical charge
### Feature description

Right now if we try to create a `ParticleList` using a `Quantity` array, we get an `InvalidParticleError`:
```Python
>>> import astropy.units as u
>>> ParticleList([5, 6] * u.kg)
plasmapy.particles.exceptions.InvalidParticleError: The object 5.0 kg supplied to ParticleList is not a particle-like object.
```
What I'd like to happen is for a `CustomParticle` to be created based on each element of the `Quantity` array, i.e.:
```Python
>>> ParticleList([5, 6] * u.kg)
ParticleList(['CustomParticle(mass=5.0 kg, charge=nan C)', 'CustomParticle(mass=6.0 kg, charge=nan C)'])
```

### Motivation

As part of the refactoring of `particle_input` in #1057, we're making it so that a `Quantity` of physical type mass or electrical charge will be used to construct a `CustomParticle` instance.  I discovered in #1871 that `ParticleList` works for `Quantity` non-arrays but not for `Quantity` arrays.  The code downstream of `ParticleList` will be cleaner and have fewer special cases to handle if `ParticleList` can handle `Quantity` iterables of these physical types.

### Implementation strategy

We'll need to make these changes to not only `ParticleList.__init__`, but also the `append`, `extend`, and `insert` methods.  

There's some repeated code in these methods, so it might be worth creating a private function that does the particleification or particlizing.

### Additional context

Implementing this would probably be cleaner by allowing mass and charge as positional arguments in either order in `CustomParticle`, kind of like what I attempted in #1364, but maybe having a signature like `CustomParticle(*quantities, mass: Optional[u.kg] = None, charge=Optional[u.C] = None, symbol: Optional[str]=None)`.

### LINKED ISSUE #1873: Add mechanism to instantiate `CustomParticle` based off of positional arguments
### Feature description

I'd like it to be possible to put the mass and charge of a `CustomParticle` in either order as positional arguments.  For example, I would like to be able to do:
```pycon
>>> CustomParticle(1 * u.C)
>>> CustomParticle(1 * u.kg)
>>> CustomParticle(1 * u.kg, 1 * u.C)
>>> CustomParticle(1 * u.C, 1 * u.kg)
```

### Motivation

While working on #1871 and #1872, I've been having to do a bunch of special handling for when the input to `CustomParticle` is a charge or a mass, including when writing tests. Code downstream of `CustomParticle` will be cleaner if we have this flexibility.

### Implementation strategy

Right now the signature of `CustomParticle.__init__` is:
```
```
I'd like to be able to do `CustomParticle(1e-27 * u.kg)` or `CustomParticle(1e-18 * u.C)`, which would correspond to a signature like:
```
def __init__(self, *quantities, symbol=None):
```

This change would lead to breaking changes, so we might want to retain `mass` and `charge` as keyword arguments, even if we deprecate them.  We could perhaps even make it so that if a string is provided in `quantities`, it would get interpreted as the symbol (though `quantities` might not be the best name in that case).

### Additional context

This is inspired by #1364 and discussed further in #1872.  This should be addressed after #1866 is merged since they're both working on the same section of code.

namurphy:
This PR makes the following changes:
 - Changed the signature of `CustomParticle` to `CustomParticle(*quantities, ...)`, where `quantities` can include `Quantity` objects of physical type mass or charge, or (as its last item) the symbol.  See #1873 for some motivation and follow-up discussion.  This is a follow-up attempt to do more or less the same thing as #1364, which didn't get merged. Closes #1873.
 - Fixed a new edge case in the particle factory function by assuming that a string passed as the first positional argument to it is meant for `Particle` and not as a symbol for `CustomParticle`.
 - Made `ParticleList.__init__` accept an iterable containing `Quantity` objects. Closes #1872.
 - Made `ParticleList` methods accept `Quantity` objects of mass or electrical charge.

The changes to `CustomParticle` made the changes to `ParticleList` pretty straightforward to implement, so I addressed them here.

This PR should make it easier to address #1869 and #1871.


namurphy:
This PR is still in its early stages, and I'm still thinking about the implementation details.

I'm going to try to reduce breaking changes, but it's likely this would make `symbol` keyword-only which would be a breaking changes.  I'm also wondering if the `mass` and `charge` parameters should be deprecated.

plasmapy/particles/tests/test_particle_collections.py
```diff
@@ -44,6 +44,8 @@ def various_particles():
             CustomParticle(),
             CustomParticle(mass=7 * u.kg),
             CustomParticle(charge=11 * u.C),
+            5 * u.kg,
+            5 * u.C,
```
StanczakDominik, line 48: So these two will create, respectively, a charge-less 5kg chonkyboi of a particle and a massless one with a massive 5 C charge, both custom ones, right?

plasmapy/particles/_factory.py
```diff
@@ -99,18 +99,27 @@ def _physical_particle_factory(
...
-        with contextlib.suppress(TypeError, InvalidParticleError):
+        with contextlib.suppress(TypeError, InvalidParticleError, ValueError):
+
+            # Do not interpret a string as the first positional argument
+            # as the symbol for a CustomParticle or something to be
+            # iterated over for a ParticleList.
+            if particle_type is not Particle and args and isinstance(args[0], str):
+                continue
+
             return particle_type(*args, **kwargs)
```
StanczakDominik, line 110: Might be myself being tired but this whole block seems difficult to reason about!  This needs a refactor:
- let's enter `suppress` context once instead of within the for loop
- the case for `Particle` should probably be separate

plasmapy/particles/particle_class.py
```diff
@@ -2016,21 +2017,27 @@ class CustomParticle(AbstractPhysicalParticle):
...
+
+    symbol : str, |keyword-only|, |keyword-only|, optional
+        The symbol to be assigned to the custom particle, if not
+        provided as the final positional argument.
 
-    symbol : str, optional
-        The symbol to be assigned to the custom particle.
+    Z : ~numbers.Real, optional, |keyword-only|
+        The |charge number|, which is equal to the ratio of the charge
+        to the elementary charge.
```
StanczakDominik, line 2040: I kind of dislike how this whole change actually made the user-facing complexity, as measured by the length of docstring necessary, higher! Nothing to change here, probably, just voicing my concern.

plasmapy/particles/particle_class.py
```diff
@@ -2042,48 +2049,79 @@ class CustomParticle(AbstractPhysicalParticle):
...
+                    )
+                charge = physical_type_dict[u.physical.electrical_charge]
+
+            if u.physical.mass in physical_type_dict:
+                if mass is not None:
+                    raise TypeError(
+                        "Cannot provide mass as both a positional and keyword argument."
+                    )
+                mass = physical_type_dict[u.physical.mass]
+
```
StanczakDominik, line 2124: Since this whole block is pretty separate from other functionality, how about moving it to a separate semi-private constructor like this...
```python
@classmethod
def _from_positional(cls, args):
    ... # this whole block
```

and then calling that from here if `quantities` is non-empty?

StanczakDominik (changes requested):
Could use some cleaning up imo!

plasmapy/particles/particle_class.py
namurphy: Agreed — the downside of this change is the increased user-facing complexity.  My original plan was to deprecate the `mass` and `charge` keyword arguments, which I haven't done but am still considering.  The extra complexity comes from attempting to preserve backward compatibility where `mass` and `charge` can be used as keyword arguments too.

plasmapy/particles/particle_class.py
namurphy: Agreed.  This would be cleaner by extracting a bunch of this into its own method.

plasmapy/particles/tests/test_particle_collections.py
namurphy: Yep!

plasmapy/particles/_factory.py
namurphy: Yeah, this block is a bit hard to wrap one's mind around.  

The reason why `contextlib.suppress` is used within the `for` block is that it tries to create a `Particle`, `CustomParticle`, and `ParticleList` sequentially.  There's a chance for one of these exceptions to be raised each time.  If the `suppress` context were outside the `for` loop, then it would exit the context environment upon the first exception...I think.  If it failed to create a `Particle`, then it wouldn't try to create a `CustomParticle` or `ParticleList` afterwards.

I'll keep thinking about how to clean this up and reduce the cognitive complexity.

namurphy:
My main question at this point is: would it be better to deprecate the `mass` and `charge` keyword arguments, and make the mass and charge positional-only?  Right now, this is looking like the optimal approach to reduce complexity not only in `CustomParticle`, but also in `ParticleList` and its attributes, `particle_input`, and the particle factory function.

plasmapy/particles/_factory.py
```diff
@@ -98,19 +100,29 @@ def _physical_particle_factory(
...
+        return Particle(*args, **kwargs)
+    except _particle_creation_exceptions:
+        # If the first argument is a string but is not a valid particle,
+        # then don't try to create a CustomParticle or ParticleList.
+        if args and isinstance(args[0], str):
+            raise
+
+    for particle_type in (CustomParticle, ParticleList):
+        with contextlib.suppress(*_particle_creation_exceptions):
             return particle_type(*args, **kwargs)
```
namurphy, line 113: Hm...it's looking like I'll need to restructure this part again in #1867 to enable `@particle_input` to take values for `Z` that are real but not integers.

StanczakDominik:
Can we instead hold off on this until we've discussed #1873 in more detail? I'm writing up some thoughts on it right now, and we do have a telecon today, right?

namurphy:
The approach suggested by @StanczakDominik at today's telecon is to implement a pattern like:

```Python
class CustomParticle:
    def __init__(self, mass: u.kg, charge: u.C):
        # current broad-strokes signature of CustomParticle creator
        pass

    @classmethod
    def _from_positionals(cls, *args):
        # use this from @particleinput
        ... # figure out physical type of argument -> create a dict -> initialize particle
        d = dict(mass = mass_value_kg, charge = charge_value_C)
        return cls(**d)
```
I think I might try this in a separate PR...

PR #1880, opened 2023-01-10
Add `strict` and `allowed_physical_types` keyword arguments to `_get_physical_type_dict`

namurphy:
## Description

Added additional keyword arguments to `plasmapy.utils._units_helpers._get_physical_type_dict`.

## Motivation and context

This PR adds capabilities that will be used when implementing a custom constructor method for `CustomParticle`.

## Related issues

This PR was originally part of #1874, which I'm splitting up into different pull requests.

PR #1881, opened 2023-01-10
Add separate constructor for `CustomParticle`

### LINKED ISSUE #1873: Add mechanism to instantiate `CustomParticle` based off of positional arguments

namurphy:
## Description

This adds a new private method to `CustomParticle` that acts as an alternative to `__init__`.  This method accepts the `charge` and `mass` of `CustomParticle` as positional arguments in either order (along with `CustomParticle` keyword arguments).  This pattern was suggested by @StanczakDominik.

## Motivation and context

The places where this will be directly applied (in future PRs) are:
 - `ParticleList` methods, including `__init__`, `append`, `extend`, etc. 
 - The particle factory function in `plasmapy.particles._factory`.

For the latter, these changes will help expand the capabilities of `particle_input`.

## Related issues

This PR takes an alternate approach to #1874. Closes #1873. This PR will need a few minor updates after #1880 is merged to handle cases where it should raise exceptions.

### DIFF
--- a/plasmapy/particles/particle_class.py
+++ b/plasmapy/particles/particle_class.py
@@ -38,6 +38,7 @@ from plasmapy.particles.exceptions import (
     ParticleWarning,
 )
 from plasmapy.utils import PlasmaPyDeprecationWarning, roman
+from plasmapy.utils._units_helpers import _get_physical_type_dict
 
 _classification_categories = {
     "lepton",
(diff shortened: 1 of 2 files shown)


plasmapy/particles/particle_class.py
```diff
@@ -2100,6 +2101,51 @@ def __init__(
...
+        """
+
+        if not quantities:
+            return CustomParticle(symbol=symbol, Z=Z)
+
+        physical_type_dict = _get_physical_type_dict(
+            quantities,
+            only_quantities=True,
+            # strict=True,
+            # allowed_physical_types={u.physical.mass, u.physical.electrical_charge},
```
namurphy, line 2136: These lines will be added back in after #1880 is merged.

plasmapy/particles/tests/test_particle_class.py
```diff
@@ -1501,6 +1501,23 @@ def test_CustomParticle_setters(attr, input):
...
+        ([5 * u.C, 6 * u.kg], {}, CustomParticle(mass=6 * u.kg, charge=5 * u.C)),
+        ([7 * u.kg], {"Z": 8.9}, CustomParticle(mass=7 * u.kg, Z=8.9)),
+        ([], {"symbol": "..."}, CustomParticle(symbol="...")),
+    ],
+)
+def test_CustomParticle_from_quantities(args, kwargs, expected):
+    actual = CustomParticle._from_quantities(*args, **kwargs)
+    assert actual == expected
+
+
```
namurphy, line 1520: After #1880 is merged, I'll add tests of the exceptions that this method should raise, such as when the wrong physical type or a non-Quantity is passed to this method.

StanczakDominik (approved):

Issue #1796, opened 2022-11-08
Support Python 3.9+ for releases after April 2023


namurphy:
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) proposes a drop schedule for old versions of NumPy and Python.  NEP 29 calls for support of Python 3.8 through releases made up until April 14, 2023.  Since NEP 29 has emerged as a community standard, it makes sense for us to follow it as well.  In particular, packages that we strong depend on (NumPy, Astropy, etc.) are following that schedule.

For releases after mid-April 2023, we should require Python ≥3.9 and drop Python 3.8 support.  This will probably be for PlasmaPy ≥0.11 if we want to make releases every ∼3–4 months.  

When we do this, we would be able to automagically apply a lot of the improved syntax by changing `--py38-plus` to `--py39-plus` in `.pre-commit-config.yaml`.  We should consider dropping `--keep-runtime-typing` as well for postponed evaluation of annotations.  I did a test run, and most of the changes from `pyupgrade` were things like `typing.List` → `list` in type hint annotations.

We'd also be able to start using [`typing.Annotated`](https://docs.python.org/3/library/typing.html#typing.Annotated), which could potentially help with decorators like `@particle_input` and let us use physical types in type hints as well.

Here's the link to [what's new in Python 3.9](https://docs.python.org/3/whatsnew/3.9.html).

This is the long awaited sequel to #1262.

PR #1883, opened 2023-01-11
Edit the docstring for `particle_input`

namurphy:
This PR makes a few minor updates to the docstring for `@particle_input`, including some edits to the content and fixes for minor formatting issues.

The docstring is also significantly more verbose than would be optimal, which makes it hard to parse and find things in it.  I might try to add subsections to the examples and/or notes sections (though I'm not sure if Sphinx can handle it), or maybe a bullet point list near the top of what `@particle_input` can do.  This might end up needing an example notebook though.

It also removes a mention regarding Python 3.9+, so this should be merged **after** the `2023.1.0` release when we drop support for Python 3.8 (#1796).

### DIFF
--- a/changelog/1883.doc.rst
+++ b/changelog/1883.doc.rst
@@ -1,0 +1,1 @@
+Updated the docstring for |particle_input|.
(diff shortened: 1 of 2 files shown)


StanczakDominik (approved):
While the docstring is indeed lengthy, this is good as an incremental improvement, so, 👍🏻

PR #1884, opened 2023-01-11
Apply new `CustomParticle` constructor to particle factory

namurphy:
## Description

This PR changes `plasmapy.particles._factory._physical_particle_factory` to use `CustomParticle._from_quantities` that was created in #1881.  This change will enable the physical particle factory to handle `Quantity` instances with units of mass or charge. This will not work with `ParticleList` instances; that'll need to wait until a sequel PR.

This PR needs #1880 before it can be merged.

## Motivation and context

This change is to improve handling of `particle_input` when dealing with `Quantity` objects of physical types mass or electrical charge.  This and related changes are described further in #1874. 

## Related issues

This supersedes #1874.


plasmapy/particles/_factory.py
```diff
@@ -110,6 +110,10 @@ def _physical_particle_factory(
                 f"physical type of {physical_type}."
             )
 
+    if args and isinstance(args[0], u.Quantity):
+        with contextlib.suppress(TypeError, InvalidParticleError):
+            return CustomParticle._from_quantities(*args, **kwargs)
+
```
namurphy, line 116: I should be able to move this to the above loop after #1880 is merged.

namurphy:
Something I learned today: we can't do `isinstance(..., ParticleLike)` for Python 3.8 or possibly Python 3.9, since that gives us a hard-to-understand `TypeError` about subscripted generics.  Works for Python 3.10, though.  This is including `u.Quantity` as one of the things encompassed in `ParticleLike`.

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,105 @@
...
+
+def _make_custom_particle_with_real_charge_number(*args, **kwargs):
+    """
+    Create a |CustomParticle| for mean or composite ions.
+
+    This function is intended to produce |CustomParticle| instances
+    provided a string representing an element or isotope without charge
+    information (e.g., ``"He"`` or ``"He-4"`` but not ``"He-4 2+"``)
+    along with a charge number that is a real number but not an integer.
```
namurphy, line 49: I'm thinking it would be worthwhile to make this capability (along with the rest of the particle factory functions) part of our public API, but I want to be able to freely change the API while getting some practice using them.

StanczakDominik: w i s d o m

No, seriously, it's just hit me that we've come a long way from the early days 😀

namurphy: Hm...I probably shouldn't mention that my true motivation was trying to find a way to justify being lazy! 😹

plasmapy/particles/particle_class.py
```diff
@@ -2555,7 +2558,14 @@ def molecule(
 * **Custom particles**
 
     `~plasmapy.particles.particle_class.CustomParticle` instances are
-    particle-like because particle properties are provided in physical units.
+    particle-like because particle properties are provided in physical
+    units. A `~astropy.units.Quantity` with a physical type of mass or
+    charge is |particle-like| because it can be used to generate a
+    |CustomParticle|.
```
StanczakDominik, line 2564: Yes, but that's a bit of a crippled particle, no? If you just pass in a charge, you get a massless charge, which is weird. I would prefer not to extend this definition to "whatever works to create a CustomParticle, too" and keep `ParticleLike` simple.

namurphy: The main reason for including `Quantity` here is to be consistent with how `particle_input` currently works. For consistency with type checkers, type hint annotations of `ParticleLike` would often need to become `Union[ParticleLike, u.Quantity]`, which would be inconsistent with the annotation that `particle_input` currently needs.  There are times too (like in `RelativisticBody`) where a mass is all that's needed. 

In #1724, I made an attempt to [decompose `ParticleLike` into different & more specific typing constructs](https://github.com/PlasmaPy/PlasmaPy/blob/645382e400b2bdca54e2b0c15bd01a4cea2c1047/plasmapy/utils/typing.py#L32), where `ParticleLike` would include `CustomParticle`, `Particle`, or `ParticleList` -like arguments, and there would be other typing constructs specific to `Particle`, `CustomParticle`, or `ParticleList` -like arguments. I'm still trying to figure out how to organize the typing constructs, so feedback would be much appreciated.  Would it be okay to continue this discussion in #1724?

I didn't realize this back in 2017 or 2018, but the biggest downside of having lots of flexibility for `Particle` is that the type hint annotations become kind of a pain to deal with! 😮‍💨 Sigh!

StanczakDominik: Absolutely, 1724 it is, then!

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,105 @@
...
+        errmsg += repr(args)
+        if kwargs:
+            errmsg += " and "
+
+    if kwargs:
+        errmsg += repr(kwargs)
+
+    errmsg += (
+        ". For information on creating particles, see: "
+        "https://docs.plasmapy.org/en/stable/glossary.html"
```
namurphy, line 36: - [ ] Update this to include the anchor for the "particle-like" entry

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+    Z = kwargs.pop("Z")
+
+    if not isinstance(Z, (Real, u.Quantity)):
+        raise InvalidParticleError("The charge number must be a real number.")
+
+    base_particle = Particle(*args, **kwargs, Z=0)
+
+    if not base_particle.is_category(require="element", exclude="ion"):
+        # Add a test if this function becomes part of public API
+        raise InvalidParticleError("Cannot create CustomParticle.")  # coverage: ignore
```
namurphy, line 90: This line shouldn't get called by anything in the public API, but I put it in here as a failsafe in case we do make this function part of the public API.  I didn't test it directly since testing the private API can make implementation details harder to change.

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+        errmsg += repr(args)
+        if kwargs:
+            errmsg += " and "
+
+    if kwargs:
+        errmsg += repr(kwargs)
+
+    errmsg += (
+        ". For information on creating particles, see:"
+        "https://docs.plasmapy.org/en/stable/api/plasmapy.particles.ParticleLike.html"
```
namurphy, line 36: I'm a bit nervous about having a link in here that won't get updated if the namespace for `ParticleLike` changes, but I added a comment next to where `ParticleLike` is defined to look for this.

namurphy:
I've made all the changes I was planning for this, so this is ready for review. Thanks to the feedback from two community meetings ago, I think this did end up cleaner than what I originally attempted in #1866 and #1364.

StanczakDominik:
I'll most likely get to it by tomorrow!

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+    Parameters
+    ----------
+    *args : (1,) tuple of str
+        A `tuple` containing a representation of an element or isotope,
+        without charge information.
+
+    *kwargs : dict of str to real numbers
+        Keyword arguments like those that can be passed to |Partile|,
+        except where the charge number ``Z`` is a real number but not
+        an integer.
```
StanczakDominik, line 60: So, why can't the signature of this be along the lines of  `_make_custom...(s: str, Z: Optional[Union[Number, u.Quantity]], mass_numb: Optional[Number])`? If `args` is gonna be a 1-tuple, it just feels like an extra step, no?

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+    CustomParticle(mass=6.64511...e-27 kg, charge=2.40326...e-19 C)
+    """
+
+    if len(args) != 1 or "Z" not in kwargs or not isinstance(args[0], (Integral, str)):
+        raise InvalidParticleError(
+            "Cannot create CustomParticle with this function with "
+            f"{args = } and {kwargs = }."
+        )
+
+    Z = kwargs.pop("Z")
```
StanczakDominik, line 81: Then this part gets simplified...

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+
+    if not isinstance(Z, (Real, u.Quantity)):
+        raise InvalidParticleError("The charge number must be a real number.")
```
StanczakDominik, line 84: But if it's a Quantity, it's not a real number in the first place...?

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,106 @@
...
+        # Add a test if this function becomes part of public API
+        raise InvalidParticleError("Cannot create CustomParticle.")  # coverage: ignore
+
+    if Z > base_particle.atomic_number:
+        raise InvalidParticleError("The charge number cannot exceed the atomic number.")
+
+    mass = base_particle.mass - m_e * Z
+    symbol = kwargs.get("symbol")
+
+    return CustomParticle(mass=mass, Z=Z, symbol=symbol)
```
StanczakDominik, line 98: This can then stay here...

plasmapy/particles/_factory.py
```diff
@@ -84,36 +175,22 @@ def _physical_particle_factory(
...
         return args[0]
 
     if not args and not kwargs:
         raise TypeError("Particle information has not been provided.")
 
-    for particle_type in (Particle, CustomParticle, ParticleList):
+    for constructor in _particle_constructors:
         with contextlib.suppress(TypeError, InvalidParticleError):
-            return particle_type(*args, **kwargs)
+            return constructor(*args, **kwargs)
```
StanczakDominik, line 191: And you can still call the function above just like this, I think! It should just work out of the box without doing that weird "this needs a 1-tuple first argument" dance. Unless kwargs may have more than those two keyword arguments accepted by the long-named function above. If so, we'd have to sidestep that somehow.

StanczakDominik (changes requested):
So I think the reason why it took me a while to review this is because the design is a bit confusing 😅

plasmapy/particles/_factory.py
namurphy: I was trying to account for things like `5 * u.dimensionless_unscaled`...which will probably not get passed to `Z` often, but might get passed to it occasionally.

plasmapy/particles/_factory.py
namurphy: I think you're right.  The main thing we need to do with this is make sure that when this fails, it raises an exception that then gets handled. When arguments don't match the signature of a function, a `TypeError` gets raised, which is fortunately one of the ones that gets handled by `contextlib.suppress` further below.

plasmapy/particles/_factory.py
```diff
@@ -9,15 +9,105 @@
...
+from typing import Any, Optional, Union
 
-from plasmapy.particles.exceptions import InvalidParticleError
-from plasmapy.particles.particle_class import CustomParticle, Particle
+from plasmapy.particles.exceptions import ChargeError, InvalidParticleError
+from plasmapy.particles.particle_class import CustomParticle, Particle, ParticleLike
 from plasmapy.particles.particle_collections import ParticleList
 
 
+def _generate_error_message(args: tuple, kwargs: dict[str, Any]) -> str:
```
namurphy, line 21: - [ ] Rename this to be more specific

StanczakDominik (approved):
This looks much better, as far as I can tell :+1:

namurphy:
Thank you for the review!  The review process did help quite a bit with this one.

Issue #1320, opened 2021-10-26
Allow type annotations like `Particle | CustomParticle`


namurphy:
A new feature in Python 3.10 is a [new type union operator](https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator) for built-in Python types.  Currently if we want to have a type hint annotation for multiple types, we need to use something like `typing.Union[int, str]`.  This new capability will eventually enable us to write `int | str` instead.  I don't know if we'd be able to implement these changes for built-in types before PlasmaPy's minimum supported version of Python is 3.10, though we might be able to by enabling deferred evaluation of annotations.  

However, we could implement this ourselves for commonly used PlasmaPy classes such as `Particle` and `CustomParticle`.  I think we'd need to specify the `__or__` dunder.  So, this issue is to make `Particle | CustomParticle` return `typing.Union[Particle, CustomParticle]` and `Particle | None` return `typing.Optional[Particle]`.  

One thing we'll have to think about is if it would be worth starting to implement `Particle | CustomParticle` before we can do `int | str`.  The benefit would be a cleaner implementation, but the drawback would be that we'd have multiple notations for type constructs.

namurphy:
From @StanczakDominik: It might be enough to use regularly `from __future__ import annotations` and not need to do this.  

Alternative could be to use string annotations instead.

We'll need to look into this more...

namurphy:
More from today's meeting: if possible, we should avoid implementing a `__or__` dunder if at all possible, since we'd have to do that repeatedly.

namurphy:
It turns out that we can do this on Python 3.10:
```pycon
>>> from plasmapy.particles import CustomParticle, Particle
>>> particle_union = Particle | CustomParticle
>>> type (particle_union)
<class 'types.UnionType'>
```
...so we don't actually have to do anything to enable this more generally, except wait.  For approximately 1.92 years.  😑

StanczakDominik:
I think it might be possible in older versions as well (how old, I'm not sure) via `from __future__ import annotations` in files that would use this.

Issue #1262, opened 2021-09-14
Drop support for Python 3.7


namurphy:
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) suggests a community schedule for dropping support old versions of Python & NumPy.  If we follow that schedule, we would drop support for Python 3.7 in PlasmaPy 0.8.0 which will likely be released in early 2022.  Python 3.10 is due to be released in ≲1 month, so we would be supporting the last three releases of Python.  Besides that, the next release of Astropy will not include support for Python 3.7, so we'd need to update our requirements fairly soon anyway.

The most significant changes in Python 3.8 are:
 - [f-string support for self-documenting expressions](https://docs.python.org/3.8/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging) (I'll probably start using these immediately)
 - [walrus operators](https://docs.python.org/3.8/whatsnew/3.8.html#assignment-expressions) (will likely improve readability here & there)
 - [positional-only parameters](https://docs.python.org/3.8/whatsnew/3.8.html#positional-only-parameters) (maybe appropriate but not necessary for mathematical stuff like the plasma dispersion function?)

namurphy:
Turns out that NumPy will be dropping support for Python 3.7 in their next release too, which is extra motivation for us dropping it too.

namurphy:
When we address this issue, it would probably be worth making use of pyupgrade (#1438) to automate a lot of the changes from 3.7+ to 3.8+ syntax.

PR #1885, opened 2023-01-12
Change minimum version of Python from 3.8 to 3.9

### LINKED ISSUE #1458: Intermittent test failure in `TestSyntheticRadiograph.test_optical_density_histogram`

### LINKED ISSUE #1796: Support Python 3.9+ for releases after April 2023

namurphy:
## Description

This PR changes package, testing, and pre-commit config files and applies changes from `pyupgrade` and `ruff` for the switch from Python 3.8+ to Python 3.9+. 

The changes from `pyupgrade` and `ruff` are primarily related to type hint annotations (e.g., changing `typing.List[str]` → `list[str]`).

This PR should be reviewed and merged after the release of `v2023.1.0`.

## Motivation and context

[NumPy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) lays out a recommended community schedule for dropping support of old versions of packages.  As per the schedule, Python 3.8 support can be dropped for releases after April 14, 2023.

## Related issues

Closes #1796. See also #912, #1661, #1883

Cross-references: #1262, #980, #1320

Closes #1458, hopefully.  #1458 was about the seemingly Python 3.8 only intermittent test failure in `TestSyntheticRadiograph`.


namurphy:
- [ ] Change the GitHub repo setting to not require 'Python 3.8, all tests, code coverage'

pyproject.toml
```diff
@@ -17,26 +17,26 @@ keywords = [
...
   "ipywidgets>=7.6.5",
   "lmfit>=1",
   "matplotlib>=3.3",
   "mpmath>=1.2.1",
   "numba",
-  "numpy>=1.20",
+  "numpy>=1.21",
   "packaging",
-  "pandas>=1",
+  "pandas>=1.2",
```
namurphy, line 32: I think it was the `pandas` dependency that was causing problems on the tests with the minimum versions.  For most of the dependencies that I changed, I looked for a release that was roughly two years old and set that as the minimum version, which is what NEP 29 recommends doing for NumPy and is probably a good balance between reducing maintenance burden and having a wide enough window of allowed package versions.

namurphy:
**Important note:** I temporarily turned off the branch protection rule for requiring the Python 3.8 minimal dependencies check to pass before merging.  After this PR is merged, we'll need to re-enable it on the Python 3.9 equivalent (and perhaps some of the other checks too).  We'll need to do the same thing for switching the tests to nox as well.

docs/common_links.rst
```diff
@@ -104,7 +104,7 @@
 .. |contributor guide| replace:: :ref:`contributor guide`\
 .. |documentation guide| replace:: :ref:`documentation guide`\
 .. |glossary| replace:: :ref:`glossary`\
-.. |minpython| replace:: 3.8
+.. |minpython| replace:: 3.9
```
StanczakDominik, line 107: Wow.

plasmapy/analysis/fit_functions.py
```diff
@@ -32,21 +32,21 @@ class AbstractFitFunction(ABC):
...
 
-    _param_names = NotImplemented  # type: Tuple[str, ...]
+    _param_names = NotImplemented  # type: tuple[str, ...]
 
     def __init__(
         self,
-        params: Tuple[float, ...] = None,
-        param_errors: Tuple[float, ...] = None,
+        params: tuple[float, ...] = None,
+        param_errors: tuple[float, ...] = None,
```
StanczakDominik, line 40: Shouldn't we also go through `future` imports, btw? see if we can't remove them? Or would pyupgrade remove them by itself?

StanczakDominik (changes requested):
LGTM! A couple minor issues. I'll run through this to check the `future` imports.

plasmapy/analysis/fit_functions.py
StanczakDominik: Ah. Guess not; the only `__future__` imports we have are `from __future__ import annotations`, and [that still isn't going to be the default for a while](https://docs.python.org/3/library/__future__.html#id2).

StanczakDominik (approved):
Applied the changelog merge; feel free to revert it if you prefer doing them separately, though.

changelog/1885.removal.rst
```diff
@@ -0,0 +1 @@
+Changed the minimum required version of Python from 3.8 to 3.9.
```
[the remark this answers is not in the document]
namurphy: Yeah, I wasn't sure where to put this one!  Happy to go with `breaking` instead.

namurphy:
Thank you for the reviews!

PR #1887, opened 2023-01-12
Add a documentation page on performance tips

### LINKED ISSUE #1760: Add a page on performance tips to the documentation
It would be helpful to add a new page to the documentation that describes how to improve computational performance when using PlasmaPy.  I'm thinking that this section should:

 - Mention that for most code that's written, performance is not as important as readability and maintainability.
 - Link to [Astropy's performance tips for units](https://docs.astropy.org/en/stable/units/index.html#astropy-units-performance) (since `Quantity` operations are one of the major slowing-things-down things)
 - Mention that pre-creating a `Particle` will save time if it is used in repeated function calls, i.e. that doing:
   ```Python
   >>> for i in range(50000): Alfven_speed(..., particle="p+")
   ```
   will be slower than:
   ```Python
   >>> proton = Particle("p+")
   >>> for i in range(50000): Alfven_speed(..., particle=proton)
   ```
 - Discuss lite-functions, while avoiding duplication between this new section, the glossary entry, and the section in the coding guide. (Judging from experience, it would probably be best to have this be its own pull request, since it was a little tricky due to coupling between documentation sections.)
 - Mention that if there is functionality that is proving to be a performance bottleneck, that it is really helpful to raise an issue on GitHub since this is something we can often do by the next release, and it can also be a good first contribution.

namurphy:
This PR adds a page including performance tips to the online documentation. This is meant to be an initial version that can be updated later, when necessary.

Closes #1760.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -71,6 +71,7 @@ Example highlights
    about/credits
    bibliography
    glossary
+   performance_tips
    Vision Statement <about/vision_statement>
    PlasmaPy Enhancement Proposals <https://github.com/PlasmaPy/PlasmaPy-PLEPs>
    PlasmaPy.org <https://www.plasmapy.org>
--- a/docs/performance_tips.rst
+++ b/docs/performance_tips.rst
@@ -1,0 +1,32 @@
+.. _performance-tips:
+
+****************
+Performance Tips
+****************
+
+.. important::
+
+   Use these
+
+
+Most of the time, optimizing software
+
+Using Astropy Units
+===================
+
+Because many of
+
+Particles
+=========
+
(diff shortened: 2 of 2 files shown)


namurphy:
The documentation failure looks to be unrelated.  It appears to be a problem with downloading packages from an Ubuntu archive. 🖥️

StanczakDominik (approved):

StanczakDominik:
Yeah, there's no way that's related. Let's get it in :)

StanczakDominik:
Right, that job's marked as required. Suppose we waited a couple das for this server issue to resolve itself?

PR #1890, opened 2023-01-13
Merge in v2023.1.0 changes into main

StanczakDominik:
A quick one :)


CITATION.cff
```diff
@@ -1,10 +1,10 @@
 title: PlasmaPy
 type: software
-version: 0.9.0
+version: 2023.1.0
```
namurphy, line 3: Awesome! 🎆 🍰 🎉 🥇 🥦

docs/whatsnew/index.rst
```diff
@@ -12,6 +12,7 @@ including bug fixes and changes to the application programming interface
    :maxdepth: 1
 
    dev
+   2023.1.0
```
namurphy, line 15: Awesome that we can do 2022 level-ups in a single release! 🙃

namurphy (approved):
LGTM!  

This is one where we need to do a merge commit on this one in order to retain the tag history, I think?  I ran into a problem with that in #1616 and #1624 where we did a squash merge and then had to revert it.

StanczakDominik:
> This is one where we need to do a merge commit on this one in order to retain the tag history, I think? I ran into a problem with that in #1616 and #1624 where we did a squash merge and then had to revert it.

Well, glad someone remembers we should do that 😁 

I think it'd be fine with a squash, just the tag wouldn't live directly on the main branch. It'd still live in the git history, just with less visibility. But a non-squash merge for these is completely fine.

PR #1892, opened 2023-01-13
Make `ParticleList` not accept strings

### LINKED ISSUE #1436: Decide on behavior of `ParticleList` when it is provided with a string
`ParticleList` accepts iterables.  When provided with a `str`, `ParticleList` will attempt to create a `Particle` or `CustomParticle` from each of the individual characters in the string.  This could be a helpful shortcut in some cases, but lead to unexpected behavior at other times.
```Python
>>> from plasmapy.particles import ParticleList
>>> ParticleList("pe")  # potentially useful shortcut
ParticleList(['p+', 'e-'])
>>> ParticleList("He")  # might expect He instead of H and e⁻
ParticleList(['H', 'e-'])
```
We have a few options for how `ParticleList` should behave, including but not limited to:

 1. Keep the current behavior (to be consistent with how `list` behaves)
 2. Make `ParticleList` raise a `TypeError` when provided with a string

If we keep the current behavior, we would also have the options of:

 3. Include an admonition in the docstring for `ParticleList` about this behavior 
 4. Include an example in the documentation so users would know that this could happen  
 5. Issue a warning for cases like `ParticleList("He")` where the whole string is a valid particle representation.  This could be done with something more or less like:
    ```Python
    if isinstance(particles, str) and len(particles) > 1:
        try:
            Particle(particles)
        except InvalidParticleError:
            pass
        else:
            warnings.warn("...")
     ```

namurphy:
## Description

This PR makes it so that `ParticleList` raises a `TypeError` when provided with a string. The error message provides an alternative that the user might have intended.

## Motivation and context

Strings are iterables, and so when `ParticleList` is given a string, it tries to create a `Particle` for each of the characters in the string. This is fine for things like `ParticleList("ep")` which is equivalent to `ParticleList(["electron", "proton"])`, but it can lead to unforeseen behavior for things like `ParticleList("He")` which becomes `ParticleList(["H", "e"])` instead of the possibly expected outcome of `ParticleList(["helium"])`.  By disallowing strings, we can avoid situations like that.

Having `ParticleList` not accept strings could also make it a little easier to deal with the particle factory function, in particular in the loop where there's an attempt to sequentially create a `Particle`, `ParticleList`, and `CustomParticle`: if a string is known to not be an input to `ParticleList`, then we could avoid potentially ambiguous situations.

There are alternatives to this, like issuing a warning instead of an error when `ParticleList` is provided with a string, or only allowing a string if there's no chance of ambiguity. These alternatives will probably lead to more complicated code without much benefit.

## Related issues

Closes #1436.


StanczakDominik (approved):
Hell yeah! 👍🏻

namurphy:
Thank you for the quick review!  

We could potentially revisit this in the future if there's enough demand for being able to do things like `ParticleList("ep")`, which would admittedly be convenient, though not as readable.

Issue #1920, opened 2023-01-21
Decompose `mass_density` into multiple, more specific functions


namurphy:
### Feature description

This issue proposes to deprecate `plasmapy.formulary.misc.mass_density` and instead create multiple utility density functions in `plasmapy.particles` that each have a specific purpose: 

 - `species_mass_density` — given one or more particles with number densities, calculate the mass density for each particle (element-wise)
 - `quasineutral_electron_density` — given one or more ions with number densities, calculate the electron number density under the assumption of quasineutrality.
 - `quasineutral_mass_density` — given one or more ions and associated number densities, calculate the total mass density under the assumption of quasineutrality.
 - `total_mass_density` — given multiple particles and associated number densities, calculate the total mass density of all particles.

### Motivation

Right now, we have the `plasmapy.formulary.misc.mass_density`.  While playing around with implementing `particle_input` in `Alfven_speed`, I've been working with `mass_density`.  I've been finding it a little confusing to work with, in particular with how to deal with `z_mean`. I'm also finding it to be more general and potentially ambiguous than I'd prefer.

Having multiple mass & electron density utility functions would lead to simpler functions that each have a specific focus.  These functions could then have names that clearly state what each function does.  For example, including "quasineutral" in the name makes it clear that the net charge density is zero, and that these functions are not applicable to non-neutral plasmas.  We can also have functions that are more general, such as `total_mass_density`, that could be applied to non-neutral plasmas.

### Implementation strategy

At the moment, I'm thinking it would be best to put these in `plasmapy.particles` because they would get used in `IonizationState` and `IonizationStateCollection`.  Other possibilities would be `plasmapy.formulary` (if we don't use them in the ionization state classes), `plasmapy.plasma`, or `plasmapy.utils`.

For some of these, we might want to avoid enforcing the assumption that the number density is positive (i.e. for applications that deal with changes in densities rather than absolute densities).

### Additional context

One thing to consider is whether these functions would be too simple, i.e. if it would be better to always inline them.  My inclination is that code will be more readable by using the explicit names, and since we'll be able to test those functions directly, we'll be less likely to introduce bugs when we use them.  It can also take a bit of time and math to remind ourselves about how to calculate the electron number density from one or more ion densities in a quasineutral plasma when $Z ≠ 1$.  There wouldn't be much improvement of readability for simple cases, but there would be for multi-species plasmas.

[...]

PR #1921, opened 2023-01-21
Decorate `reduced_mass` with `particle_input` & `validate_quantities` whilst expanding its docstring

namurphy:
- Decorated `plasmapy.particles.atomic.reduced_mass` with `@particle_input` and `@validate_quantities`
   - Removed a function inside `reduced_mass` that extracted the particle mass
 - Expanded the docstring of `reduced_mass` to include its equation and some properties


namurphy:
Coincidentally, I'm wondering about moving `reduced_mass` to a separate module in `plasmapy.particles`, since `atomic.py` might not be the best fit.  If we address #1920, we could possibly move `reduced_mass` into a module with the density functions, like maybe `plasmapy.particles.utilities`.  But, that's not for this issue.

plasmapy/particles/atomic.py
```diff
@@ -833,46 +844,27 @@ def reduced_mass(test_particle, target_particle) -> u.Quantity:
...
-                f"Unable to find the reduced mass because the mass of "
-                f"{particle} is not available."
-            ) from None
-        else:
-            return particle.mass.to(u.kg)
-
-    test_mass = get_particle_mass(test_particle)
-    target_mass = get_particle_mass(target_particle)
-
-    return (test_mass * target_mass) / (test_mass + target_mass)
```
namurphy, line 875: Feels nice to be able to delete this much code!

namurphy:
This is pretty close, but this might be a good time to improve the test coverage of `reduced_mass` a bit.  It would be good to merge #1884 before this one, so that I can add some additional test cases that #1884 would enable.

StanczakDominik (approved):
Skimmed, looks great for now 👍🏻

PR #1933, opened 2023-01-25
Rerun particle_tracing notebooks due to high resource usage

StanczakDominik:
I reran the particle tracing notebooks. Seems they somehow became un-pre-executed, and now they're slowing down the doc builds by a major factor. It's been causing intermittent doc build failures on main for a bit, now.

We might also have to disable the epub/pdf builds again because we might not actually be latex compatible - we'll see.


namurphy (approved):
Strange that they get un-pre-executed!  In any case, LGTM, and thanks for doing this!

I haven't looked at the pdfs to tell how compatible they are with LaTeX, though that was a good point to bring up.

StanczakDominik:
I think I forgot to re-execute them after the last PR that touched them, and that's that mystery solved 😅

PR #1953, opened 2023-02-01
Move codespell to pre-commit

namurphy:
## Description

This PR moves `codespell` from `tox` and into the `pre-commit` configuration.

## Motivation and context

We use `codespell` to catch common misspellings.  We currently have a separate `tox` environment and GitHub Action for codespell. However, if we move `codespell` to `pre-commit`, we can remove both the `tox` environment and the GitHub Action.

## Related issues

This is based off of https://github.com/astropy/astropy/pull/13982 and https://github.com/astropy/astropy/pull/13985.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,6 +59,14 @@ repos:
   - id: rst-directive-colons
   - id: rst-inline-touching-normal
 
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.2.2
+  hooks:
+  - id: codespell
+    args: [--write-changes]
+    additional_dependencies:
+    - tomli
+
 - repo: https://github.com/charliermarsh/ruff-pre-commit
   rev: v0.0.237
   hooks:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -122,25 +122,6 @@ jobs:
     - name: Run tests
       run: tox -e build_docs_pins -- -q
 
-  codespell:
-    name: codespell
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-    steps:
-    - name: Checkout code
-      uses: actions/checkout@v3
-      with:
-        fetch-depth: 0
-    - name: Set up Python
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -59,6 +59,14 @@ repos:
   - id: rst-directive-colons
   - id: rst-inline-touching-normal
 
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.2.2
+  hooks:
+  - id: codespell
+    args: [--write-changes]
```
namurphy, line 66: The `--write-changes` option will make pre-commit automatically apply codespell fixes. There's a small chance that this could end up making some incorrect fixes, but I figure we can give it a try and change it if we run into any problems.

StanczakDominik: Hmm. This is concerning - codespell finds enough false positives that I'm worried it might add to the busy work of reverting stuff.

But sure, let's try it and revert it if necessary.

StanczakDominik: Like, I don't know about you, but my preferred workflow is "small, fast, conceptually minimal commits, then once I'm happy with the result, clean stuff up". To that end I often end up doing `commit -n` to skip pre commit checks, then just fix them up eventually.

I guess this is mostly because they end up fixing style issues - like typos -for which I don't want to spare the mental bandwidth while I'm in the weeds, rather than pointing out stuff like "hey, your for loop syntax is wrong here".

Guess I'm ranting. Sorry.

StanczakDominik (approved):

.pre-commit-config.yaml
namurphy: Yeah, I hear you on that.  I tend to use IDEs for showing where broken loops are, rather than pre-commit.  And sometimes I just want to commit code that doesn't work!

I didn't realize that you could do `git commit -n` instead of `git commit --no-verify`!  That is definitely going to speed up my workflow by approximately 0.7 seconds about three times per week!

In any case, I'll merge this, and we can remove it if we run into a problem with it.

PR #1987, opened 2023-03-04
Make `ParticleList` accept `Quantity` arrays

namurphy:
The goal of this PR is to make `ParticleList` accept `Quantity` arrays with physical types of mass or electrical charge.  This also makes `ParticleList.append` accept `Quantity` non-arrays.  Each `Quantity` of the appropriate physical type will be converted into a `CustomParticle`.  

The motivation is primarily to allow `particle_input` to process `Quantity` arrays, which is needed for #1871 (if I recall correctly).

This PR originated as part of #1874.

### DIFF
--- a/plasmapy/particles/particle_collections.py
+++ b/plasmapy/particles/particle_collections.py
@@ -19,7 +19,28 @@ from plasmapy.particles.particle_class import (
 )
 
 
-def _turn_quantity_into_custom_particle(quantity):
+def _turn_quantity_into_custom_particle(
+    quantity: u.Quantity[u.physical.electrical_charge, u.physical.mass]
+) -> CustomParticle:
+    """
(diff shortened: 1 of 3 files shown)


plasmapy/particles/particle_collections.py
```diff
@@ -199,8 +247,9 @@ def _get_particle_attribute(self, attr, unit=None, default=None):
 
     def append(self, particle: ParticleLike):
         """Append a particle to the end of the |ParticleList|."""
-        # TODO: use particle_input when it works with CustomParticle and ParticleLike
```
namurphy, line 202: Because `particle_input` can now create `ParticleList` objects, we want to avoid using `particle_input` here to avoid circularity.

ejohnson-96 (approved):
Good job :)

PR #1990, opened 2023-03-06
Remove deprecated `pytest-xdist` option and update `CODEOWNERS`

namurphy:
This PR removes the `looponfailroots` option, which has been deprecated for `pytest-xdist`.  I also added myself as the code owner for the `pre-commit` configuration file.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,6 +2,7 @@
 
 # any file, anywhere
 *                                                        @PlasmaPy/plasmapy-reviewers
+.pre-commit-config.yaml                                  @namurphy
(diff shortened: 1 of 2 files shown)


pheuer (approved):
Seems straightforward - I couldn't find documentation on what looponfailroots does (did?) but it looks like it's deprecated so maybe that's why?

namurphy:
> I couldn't find documentation on what looponfailroots does (did?) but it looks like it's deprecated so maybe that's why?

I think it was something along the lines of re-running tests that had failed, but yeah, I wasn't able to find current documentation on it either.  Thanks for the review!

PR #1994, opened 2023-03-07
Add a GitHub Action for automatically updating `requirements.txt`

namurphy:
## Description

This PR adds a new GitHub Action that will run once per week (or upon maintainer request) to use `pip-compile` to regenerate `requirements.txt`.

I used ChatGPT pretty heavily to come up with this GitHub Action. It didn't work perfectly, but it did work pretty well!

## Motivation and context

We currently using `requirements.txt` to have a stable environment for CI and for working with IDEs. At the moment, we are using [dependabot](https://github.com/dependabot) to submit pull requests that update `requirements.txt`, one requirement at a time.  The difficulty here is that a some of the requirements are coupled with each other, in which case updating them in isolation will occasionally lead to CI failures.  The goal of this PR is to replace dependabot.


namurphy:
I want to test this action on a separate repository first to make sure that it is working, which I won't have a chance to do now.

The pre-commit error appears to be an issue not with this pull request, but rather with the requirements fixer pre-commit hook.  Will need to check on that later too.

namurphy:
I'm wondering about the cadence for updating `requirements.txt`.  Changing things too frequently would mean extra work with code reviewing, and we'd possibly have to deal with more merge conflicts in `requirements.txt` for PRs that add (for example) a Sphinx extension.  Our IDEs might also give us more warnings about our environments being out-of-date.  Longer times will mean that we don't find out about broken upstream dependencies as quickly.  

What if we updated `requirements.txt` monthly, while adding a weekly cron job to check that things aren't breaking?

I'm also wondering if there's a way to generate a requirements file for the minimum supported dependencies. Maybe I'll ask ChatGPT that too...

namurphy:
One thing to consider here is whether it would make sense to use something like a [poetry](https://python-poetry.org/) lock file to handle what we're using `requirements.txt` for.  My understanding is that poetry isn't the ideal choice for libraries, but we might be able to make it work.  

My inclination at the moment is to try out this GitHub Action, since it should suffice for what we want it to do.  If it ends up not working that well, then we might want to look into poetry and other tools later on.

(I'm waiting on this PR since I want to try it out on a different test repository first.)

.github/workflows/update-pinned-reqs.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Install pip-tools
+      run: python -m pip install pip-tools
+
+    - name: Generate requirements.txt
+      run: pip-compile --all-extras --output-file=requirements.txt --resolver=backtracking -r -U
+
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v5
+      with:
+        commit-message: Run pip-compile
```
namurphy, line 36:
```suggestion
        commit-message: Regenerate requirements.txt
```

ejohnson-96 (approved):
👍

PR #2004, opened 2023-03-09
Rename label: "Good first contribution" → "good first issue"

namurphy:
This PR updates our documentation to refer to the "good first issue" label rather than the "Good first contribution" label.  The reason is that "good first issue" has become the most common wording for this, and there are sites that look specifically for "good first issue" labels.  So, it'd be helpful to update it for us too.


ejohnson-96 (approved):
Looks good, all labels changed :)

Issue #1711, opened 2022-08-30
Remove upper limit on version of `towncrier`


namurphy:
Version 22.9.0 of [towncrier](https://towncrier.readthedocs.io/en/latest/) was just released, and it started causing some documentation build errors like:

```
/home/runner/work/PlasmaPy/PlasmaPy/docs/whatsnew/dev.rst:3: CRITICAL: render_fragments() got multiple values for argument 'top_underline'
/home/runner/work/PlasmaPy/PlasmaPy/docs/whatsnew/index.rst:11: WARNING: toctree contains reference to document 'whatsnew/dev' that doesn't have a title: no link will be generated
```

In #1710, there was an upper limit placed on towncrier.  We should update PlasmaPy so that it becomes compatible with the newest release of towncrier, in particular since upper limits on versions increase the likelihood of a version conflict.

ConorMacBride:
This issue should be resolved when https://github.com/OpenAstronomy/sphinx-changelog/pull/14 is released.

PR #2018, opened 2023-03-14
Update versions of `towncrier` and `sphinx-changelog`

### LINKED ISSUE #1851: Remove upper limit on allowed versions of `towncrier`
In #1850, we ended up pinning `towncrier` to `22.8.0` because there ended up being a discrepancy between `towncrier` and `sphinx-changelog`.  This appears to be because packages that depend on `towncrier` made use of its `Config` object, which is not part of the `towncrier` public API, and then `Config` got converted into a `dataclass`.  Once that issue is addressed upstream, we should revert that upper limit and change to `towncrier >= 22.8.0` or `towncrier >= 22.12.0`.  We'll need to wait until that issue is addressed upstream.

(I also added a title to `dev.rst` in #1850 to fix a separate (unrelated?) warning, so we should make sure the title still needed and appears okay.)

namurphy:
There had been an incompatibility between recent releases of towncrier and sphinx-changelog, so we had previously pinned the versions of these packages. The most recent release of sphinx-changelog should have resolved this incompatibility, so it should now be interoperable with the most recent release of towncrier.  This PR sets the most recent versions of towncrier and sphinx-changelog as lower limits in the requirements.  

Closes #1851. See also #1711 and #1717.


ejohnson-96 (approved):
Looks good :)

ejohnson-96 (changes requested):
Merge conflicts arise when updating

namurphy:
@ejohnson-96 — just fixed the conflicts that resulted from merging #2019.  Thanks!

ejohnson-96 (approved):
👍

PR #1654, opened 2022-07-30
Restrict `sphinx_gallery < 0.11`

rocco8773:
Today (2022-07-29) `sphinx_gallery` released [version `0.11.0`](https://pypi.org/project/sphinx-gallery/0.11.0/).  The changes introduced in this version are negatively affecting the display of thumbnail galleries generated by `nbsphinx`, as such I'm restricting `sphinx_gallery` to `< 0.11.0`.

There are likely multiple changes that are sources to this issue, all of which are probably related to https://github.com/sphinx-gallery/sphinx-gallery/issues/905.  I do think the issue can be traced back to the change of the `.spx-glr-thubcontainer` style to a flex container.  **Instead of spending days trying to fix this my(our)selves.  I'm just restricting the `sphinx_gallery` version and hoping `nbsphinx` comes up with their own solution that we can adopt later.** :)

 on the change to flex containersThe root cause of the display issues comes from changes made to the `.spx-glr-thubcontainer` style...

## OLD

```css
.sphx-glr-thumbcontainer {
  background: #fff;
  border: solid #fff 1px;
  -moz-border-radius: 5px;
  -webkit-border-radius: 5px;
  border-radius: 5px;
  box-shadow: none;
  float: left;
  margin: 5px;
  min-height: 230px;
  padding-top: 5px;
  position: relative;
}
```

image

## NEW

```css
.sphx-glr-thumbcontainer {
  background: transparent;
  -moz-border-radius: 5px;
  -webkit-border-radius: 5px;
  border-radius: 5px;
  box-shadow: 0 0 10px var(--sg-thumb-box-shadow-color);

  /* useful to absolutely position link in div */
  position: relative;

  /* thumbnail width should include padding and borders
  and take all available space */
  box-sizing: border-box;
  width: 100%;
  padding: 10px;
  border: 1px solid transparent;

  /* align content in thumbnail */
  display: flex;
  flex-direction: column;
  align-items: center;
  gap: 7px;
}
```

image

---

`nbsphinx` has already taken not of this... https://github.com/spatialaudio/nbsphinx/issues/655

PR #2019, opened 2023-03-14
Remove upper limit on `sphinx-gallery` and specify documentation-related requirements

### LINKED ISSUE #1972: Remove upper limit on `sphinx-gallery`
In #1654, we added an upper limit to `sphinx-gallery` because of an incompatibility with `nbsphinx` that made our example notebook galleries get formatted weird.  We should hopefully or eventually be able to remove this upper limit, so long as newer versions of `sphinx-gallery` work well with `nbsphinx`.  It's probably best to wait to do this until later though, so I'm marking it as "on hold".

namurphy:
Previously, we placed an upper limit on `sphinx-gallery` because using the newer version led to the thumbnails for our example galleries to show up incorrectly.  This PR removes the upper limit on `sphinx-gallery`.

This PR also sets some new minimum versions of packages for building the documentation, since it may be updates to `sphinx_rtd_theme` or `nbsphinx` that are what fixed the problem.  It's worth noting that the new requirements for several of these packages are for the most recent release, but since this is for optional requirements for the documentation, I don't expect this to cause any problems.

Related to #1654.  Closes #1972.


namurphy:
The [examples page](https://plasmapy--2019.org.readthedocs.build/en/2019/examples.html) for the Read the Docs documentation build for this pull request appear to be fine, so this PR should be good to go!

ejohnson-96 (approved):
Looks good :)

namurphy:
Thanks for the review!

PR #2022, opened 2023-03-17
Decorate `two_fluid` dispersion solver with `@particle_input`

namurphy:
## Description

This PR decorates `plasmapy.dispersion.analytical.two_fluid_.two_fluid` with `@particle_input`. 

A new capability is that `two_fluid` will now accept `mass_numb` separately from ``ion``.

There are some breaking changes:

 - The `z_mean` parameter got changed to `Z` for `two_fluid` (for consistency with how we're handling the charge number
 - `z_mean` will no longer override the charge provided in `ion` (since having contradictory  arguments make code confusing to read without referring to the docstring)
 - The charge number will no longer default to 1 (to avoid implicit particle assumptions)

## Motivation and context

Using `@particle_input` here will allow us to drop two blocks of code which were validating ``ion`` and ``z_mean``.  Plus, using the mechanics allowed by `@particle_input` will give a more consistent user experience across subpackages.

## Related issues

This is motivated by #1869, which at present is marking some `z_mean` override tests as expecting to fail.  

This is an offshoot of #341, except for `plasmapy.dispersion` instead of `plasmapy.formulary`.


plasmapy/dispersion/analytical/tests/test_two_fluid_.py
```diff
@@ -139,34 +140,6 @@ def test_on_bellan2012_vals(self, kwargs, expected):
...
-        ],
-    )
-    def test_z_mean_override(self, kwargs, expected):
-        """Test overriding behavior of kw 'z_mean'."""
-        ws = two_fluid(**kwargs)
-        ws_expected = two_fluid(**expected)
-
-        for mode in ws:
-            assert np.isclose(ws[mode], ws_expected[mode], atol=0, rtol=1.7e-4)
-
```
namurphy, line 169: Dropping this test since the `z_mean` will no longer override a charge number provided in `ion`.

ejohnson-96 (approved):
👍 could users possible get confused between charge state and z mean?

PR #2027, opened 2023-03-18
Make `@particle_input` accept `z_mean` instead of `Z` whilst issuing a deprecation warning

### LINKED ISSUE #2025: Enable `@particle_input` to handle `z_mean` and issue `DeprecationWarning`
There are going to be some PRs like #1869 that drop the `z_mean` parameter in favor of `Z` for some functions, since `Z` is the parameter expected by `@particle_input`.  We can make this change now since `@particle_input` now allows non-integer charge numbers, so we can use `Z` consistently.

Rather than having a sudden breaking change, we could make `@particle_input` accept `z_mean`, in which case it would issue a `DeprecationWarning` saying that users should switch to `Z`, while also setting `Z = z_mean`.

Given that `z_mean` is used in a bunch of places in `plasmapy.formulary`, we'd probably want to keep the `DeprecationWarning` in place for ∼3–5 releases.

namurphy:
This PR adds a clause within `particle_input` so that functions decorated with it can accept `z_mean` as a keyword argument that then gets passed to `Z`.  If `z_mean` is provided, then then a deprecation warning will be issued.

Closes #2025.

To do:

 - [ ] Add tests
 - [ ] Decide if we'd want to do this for other parameters, like `m` for a particle's mass (see #1871), in which case we might want to change this into an instance method.

### DIFF
--- a/plasmapy/particles/decorators.py
+++ b/plasmapy/particles/decorators.py
@@ -22,6 +22,7 @@ from plasmapy.particles.exceptions import (
 )
 from plasmapy.particles.particle_class import CustomParticle, Particle, ParticleLike
 from plasmapy.particles.particle_collections import ParticleList, ParticleListLike
+from plasmapy.utils import PlasmaPyDeprecationWarning
 
 _basic_particle_input_annotations = (
     Particle,  # deprecated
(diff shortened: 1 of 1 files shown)


namurphy:
> * [x]  Decide if we'd want to do this for other parameters, like `m` for a particle's mass 

At least for now, I decided against special handling of any other parameter names.  The reason for handling the deprecation warning for `z_mean` is that `z_mean` gets used in a whole bunch of different places, so it makes sense to do this more generally to enable current behavior for a while longer.

namurphy:
Thank you for the quick review!

PR #2033, opened 2023-03-22
Update requirements

namurphy:
We've been having an CI issue over the last few days, apparently from `ipykernel` not having been found.  This PR updates the pinned requirements as a means of investigating the issue.  

I'm curious if a way to fix this would be to change `ipykernel` from an optional to a required dependency in `pyproject.toml`, but I haven't tried that on my first commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dynamic = [
 dependencies = [
   "astropy>=5.0.1",
   "h5py>=3.1",
-  "ipykernel",
+  "ipykernel>=6.6.1",
   "ipywidgets>=7.6.5",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dynamic = [
 dependencies = [
   "astropy>=5.0.1",
   "h5py>=3.1",
+  "ipykernel>=6.6.1",
```
namurphy, line 39: This is needed for the plasma calculator, so it should be a required dependency.

ejohnson-96 (approved):
Providing the documentation successfully builds, should be good :)

PR #2034, opened 2023-03-22
Allow `@particle_input` to accept custom ions when parameter is named `ion`

namurphy:
A feature of `@particle_input` is that if a parameter to a function is named `ion`, then it will ensure that the particle is indeed an ion. This feature is syntactic sugar for the categorization criteria that can be specified in `@particle_input`.  

However, this feature doesn't work for positively charged particles that represent (for example) *mean* ions. The purpose of this pull request is to enable `@particle_input` to accept `CustomParticle` objects that have a mass and are positively charged.  

This is needed for #2022 to enable compatibility with charge numbers that are not integers.

### DIFF
--- a/changelog/2034.breaking.rst
+++ b/changelog/2034.breaking.rst
@@ -1,0 +1,3 @@
+|particle_input| no longer enforces that |parameters| named
+``ionic_level`` are ions or neutral atoms. For equivalent behavior,
+name the parameter ``ion`` instead.
--- a/plasmapy/particles/decorators.py
+++ b/plasmapy/particles/decorators.py
@@ -619,7 +619,7 @@ def particle_input(
     allow_custom_particles: bool = True,
     allow_particle_lists: bool = True,
 ) -> Callable:
-    """
+    r"""
     Convert |particle-like| |arguments| into particle objects.
 
     When a callable is |decorated| with |particle_input|,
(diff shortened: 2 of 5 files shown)


plasmapy/particles/decorators.py
```diff
@@ -413,21 +413,22 @@ def verify_particle_name_criteria(self, parameter, particle):
...
+            and particle.charge.value > 0
+            and particle.mass.value > 0
+        ):
+            return
+
         name_categorization_exception = [
             ("element", {"require": "element"}, InvalidElementError),
             ("isotope", {"require": "isotope"}, InvalidIsotopeError),
-            # TODO: In the future, "ion" should be changed to allow only
-            # ions and not neutral atoms with an explicit charge.
```
namurphy, line 420: Right now, I'm thinking that it would be fine to have `ion` be a catch-all and include neutral atoms because...

 - It makes the code simpler
 - We'd could avoid breaking changes
 - It's really often helpful to have functions work on neutrals, like `gyroradius` returning `np.inf * u.m`.

changelog/2034.breaking.rst
```diff
@@ -0,0 +1,3 @@
+|particle_input| no longer enforces that |parameters| named
+``ionic_level`` are ions or neutral atoms. For equivalent behavior,
+name the parameter ``ion`` instead.
```
namurphy, line 3: This is technically a breaking change but it's pretty obscure behavior so no deprecation period is needed.

plasmapy/particles/tests/test_decorators.py
```diff
@@ -593,6 +601,22 @@ def test_particle_list_all_in_category(self, case):
...
+    custom_ion = CustomParticle(mass=2e-27 * u.kg, charge=3e-19 * u.C)
+    result = get_ion(custom_ion)
+    assert result == custom_ion
+
+
+def test_creating_mean_particle_for_parameter_named_ion():
+    Z = 1.3
+    ion = get_ion(ion="He", Z=Z)
+    assert u.isclose(ion.charge, Z * const.e.si)
+
```
pheuer, line 618: Do you need an additional test here enforcing that this statement is true? 

```
|particle_input| no longer enforces that |parameters| named
``ionic_level`` are ions or neutral atoms. For equivalent behavior,
name the parameter ``ion`` instead.
```

Something that fails to raise an error for `ionic_level` but does for `ion`?

pheuer (commented):
Sorry for the delayed review! 

I find `ion` accepting neutral particles to be potentially confusing: if the decorator is going to do validation based on the parameter name, then I would expect the name `ion` to only allow particles with an explicit non-zero charge to pass.  

Thinking about your `gyroradius` example: sure it makes sense to allow neutral particles to that function, but then why call the parameter `ion` instead of `particle`?

plasmapy/particles/decorators.py
```diff
@@ -413,21 +414,22 @@ def verify_particle_name_criteria(self, parameter, particle):
...
+            and not np.isnan(particle.charge)
+            and particle.mass.value > 0
+        ):
+            return
+
         name_categorization_exception = [
             ("element", {"require": "element"}, InvalidElementError),
             ("isotope", {"require": "isotope"}, InvalidIsotopeError),
-            # TODO: In the future, "ion" should be changed to allow only
-            # ions and not neutral atoms with an explicit charge.
```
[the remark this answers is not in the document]
namurphy: The way the `Particle` class works is that `Particle("H-1")` and `Particle("H-1 0+")` are different.
 - In `Particle("H-1")`, there is no explicit charge, so `Particle("H-1").charge` is `np.nan` C.  (However, the mass is calculated assuming that it is neutral so there's an inconsistency there.)
 - In `Particle("H-1 0+"), the charge is explicitly defined, so `Particle("H-1 0+").charge` returns 0 C.

This is a design decision that I regret making, so I started working on #1515 to address this.  However, it's turning out to be a pain to deal with this.

changelog/2034.breaking.rst
namurphy: We have a function called `ionic_levels`, but I don't think we use `ionic_level` as a parameter to a function anywhere. So, this is very unlikely to cause anyone any problems.

plasmapy/particles/tests/test_decorators.py
namurphy: Given the obscurity of `ionic_level`, I'd say we can safely skip this test.

namurphy:
@pheuer — Thank you for the review!  I just created #2049 as an issue where we can discuss how to handle parameter names like `ion`, and whether or not a neutral atom could reasonably fit in there. That question is beyond the scope of this PR, so I'd prefer to defer the discussion to that issue.  Did you have any other thoughts on this PR?  Thank you again!

plasmapy/particles/decorators.py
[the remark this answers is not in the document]
pheuer: I agree this is out of scope for this PR - maybe the design issue is allowing particles without a charge specified to begin with?

plasmapy/particles/decorators.py
pheuer: I agree - 'atom' definitely implies neutral to me!

pheuer (approved):
@namurphy My only objections are now covered in #2049, so I'm fine merging this and discussing a larger scale solution over there.

PR #2043, opened 2023-03-29
Repair GitHub Action that labels pull requests

### LINKED ISSUE #2042: Fix labeler GitHub Action
We have a [labeler](https://github.com/PlasmaPy/PlasmaPy/blob/main/.github/labeler.yml) GitHub Action, but it doesn't seem to be working at the moment.  We should look into this and try to fix it.

namurphy:
This PR (hopefully!) repairs our implementation of the [labeler](https://github.com/marketplace/actions/labeler) GitHub Action. 

It appears that the problem was that we deleted our `.github/workflows/labeler.yml` action in #1800, when we were trying to repair the size-labeler GitHub Action.  My bad!

This is one of those pull requests where we won't be able to test it until after we merge it, but if there are any problems with it, then I'll submit a follow-up pull request shortly thereafter.

Closes #2042. Hopefully!

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,32 +1,45 @@
-Analysis: plasmapy/analysis/**/*
+Breaking:
+- changelog/*.breaking*rst
 
-Continuous integration: .github/**/*
+Contributor Guide:
+- docs/contributing/**/*
 
-Contributor Guide: docs/contributing/*
+dependencies:
+- requirements.txt
 
-Diagnostics: plasmapy/diagnostics/**/*
-
-Dispersion: plasmapy/dispersion/**/*
-
-Formulary: plasmapy/formulary/**/*
-
-Mathematics: plasmapy/formulary/mathematics.py
-
-parameters.quantum: plasmapy/formulary/quantum.py
-
-Particles: plasmapy/particles/**/*
-
-# Add Documentation label to any change in docs but notebooks
 Documentation:
 - any: [docs/**/*]
   all: ['!docs/notebooks/**/*', '!docs/contributing/*']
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
namurphy, line 1: The format for this file is:
```yaml
label-name:
- path/to/file
```
...where `*` is the usual wildcard character and `**` represents any number of directories.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v4
+      with:
+        sync-labels: true
```
namurphy, line 14: The `sync-labels` option means that if the pull request is updated, then the labels will be updated too.

.github/labeler.yml
```diff
@@ -1,32 +1,45 @@
...
+GitHub Actions:
+- .github/**/*
 
-Formulary: plasmapy/formulary/**/*
+Notebooks:
+- docs/notebooks/**/*
 
-Mathematics: plasmapy/formulary/mathematics.py
+plasmapy.analysis:
+- plasmapy/analysis/**/*
```
namurphy, line 21: I updated the labels to include the full namespace, which makes it a little more explicit.  It's also a little easier to maintain since all of the subpackage labels will be next to each other in an alphabetical list.

namurphy:
I'm going to merge this without waiting for review so that I can figure out if it's working more quickly.

PR #2046, opened 2023-03-30
Update pre-commit versions and fix linter error

namurphy:
We just got a new linter error, which is essentially that we should switch `something.get('key', None)` with `something.get('key')`.  

I'm also going to take a moment to update our pre-commit configuration, I think.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ci:
   autofix_prs: false
-  autoupdate_schedule: monthly
+  autoupdate_schedule: weekly
 
 repos:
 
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -2,9 +2,6 @@ ci:
   autofix_prs: false
   autoupdate_schedule: monthly
 
-default_language_version:
-  node: 16.15.0
-
```
namurphy, line 7: If this PR runs into any problems, it might be because we removed these lines.  We needed to add them here for some reason, back in the day, but we shouldn't need them anymore.

ejohnson-96 (approved):
Looks Good

PR #2058, opened 2023-04-04
Add notch option to thomson spectral density function

bryancfoo:
## Description

A "notches" keyword option was added to the Thomson spectral density function. It takes in ranges of wavelengths (input via the endpoints) where the output Skw is set to 0.

## Motivation and context

When measuring the EPW spectrum in collective TS, the low omega regime is notched out as the IAW term dominates there. This option lets the user easily generate synthetic EPW spectra which have the notch included to better resemble real data.


ejohnson-96 (dismissed):
Tests need to be added for the new functionality.
Apart from that looks good :)

changelog/2058.feature.rst
```diff
@@ -0,0 +1 @@
+Added a 'notches' keyword option to thomson.spectral_density, which allows users to notch out the output spectrum over a wavelength range. Also made some changes to the thomson fitting models to include the notch as an option.
```
pheuer, line 1: thomson.spectral_density -> `~plasmapy.diagnostics.thomson.spectral_density` should link the function name.

pheuer, line 1: thomson -> Thomson

Maybe rephrase last sentence 

"Changed the `~plasmapy.diagnostics.thomson.spetctral_density_model` function to allow a notch to be applied during fitting."


Right now a notch is added manually in the `run_fit()` function in the Thomson test file. Step one of testing  this PR would be to replace that manual implementation with your new implementation and make sure all the tests pass.

plasmapy/diagnostics/thomson.py
```diff
@@ -63,6 +63,7 @@ def spectral_density_lite(
     probe_vec: np.ndarray,
     scatter_vec: np.ndarray,
     instr_func_arr: Optional[np.ndarray] = None,
+    notches: Optional[np.ndarray] = None,
 ) -> tuple[Union[np.floating, np.ndarray], np.ndarray]:
     r"""
```
pheuer, line 68: You need to add this keyword to the docstring. What is the shape of the required array? Can it include multiple notches?

plasmapy/diagnostics/thomson.py
```diff
@@ -250,6 +251,17 @@ def spectral_density_lite(
...
+
+    # add notch(es) to the spectrum if any are provided
+    if notches is not None:
+        if np.shape(notches) == (2,):
+            notches = np.array([notches])
+
+        for notch in notches:
+            x0 = np.argmin(np.abs(wavelengths - notch[0]))
+            x1 = np.argmin(np.abs(wavelengths - notch[1]))
+            Skw[x0:x1] = 0
```
pheuer, line 263: Not sure you want to set it to zero. The way I've handled notches is: 

1) Set it to np.nan, then it will be ignored in the fit. This causes issues when also using an insturment function, though. 
2) Cut it out of both the wavelength and Skw arrays, e.g. 

```
wavelength = np.concatenate( (wavelength[:x0], wavelength[x1:] ))
Skw = np.concatenate( (Skw[:x0], Skw[x1:] ))
```

plasmapy/diagnostics/thomson.py
```diff
@@ -276,6 +288,7 @@ def spectral_density(
     probe_vec=None,
     scatter_vec=None,
     instr_func: Optional[Callable] = None,
+    notches=None,
```
pheuer, line 291: Why not include the type hint here too?

plasmapy/diagnostics/thomson.py
```diff
@@ -276,6 +288,7 @@ def spectral_density(
...
 ) -> tuple[Union[np.floating, np.ndarray], np.ndarray]:
     r"""Calculate the spectral density function for Thomson scattering of
```
pheuer, line 293: Add to docstring here too

plasmapy/diagnostics/thomson.py
```diff
@@ -542,6 +555,24 @@ def spectral_density(
...
 
+    # Valildate notch input
+
+    if notches is not None:
+        notches_unitless = notches.to(u.m).value
+
+        if len(np.shape(notches_unitless)) == 1:
+            notches_unitless = np.array([notches_unitless])
+
+        for notch in notches_unitless:
```
pheuer, line 566: I see - you do have the exceptions + validation here. So, you can just remove them from the lite function.

plasmapy/diagnostics/thomson.py
```diff
@@ -542,6 +555,24 @@ def spectral_density(
     else:
         instr_func_arr = None
 
+    # Valildate notch input
+
+    if notches is not None:
+        notches_unitless = notches.to(u.m).value
+
+        if len(np.shape(notches_unitless)) == 1:
```
pheuer, line 563: ```if len(np.shape(notches_unitless)) == 1:```

What are you checking here? Is there a more direct way to check it?

plasmapy/diagnostics/thomson.py
```diff
@@ -936,6 +970,9 @@ def spectral_density_model(wavelengths, settings, params):
             "`numpy.delete`."
         )
```
pheuer, line 972: Add to docstring of `spectral_density_model`

plasmapy/diagnostics/thomson.py
```diff
@@ -250,6 +251,17 @@ def spectral_density_lite(
     # Apply an instrument function if one is provided
     if instr_func_arr is not None:
         Skw = np.convolve(Skw, instr_func_arr, mode="same")
+
+    # add notch(es) to the spectrum if any are provided
+    if notches is not None:
+        if np.shape(notches) == (2,):
```
pheuer, line 257:
```
        if np.shape(notches) == (2,):
            notches = np.array([notches])
```
What is this code checking? Is there a more direct way of doing it?

pheuer (changes requested):
Looks like a good start: in addition to adding tests, make sure to add the new keyword to the docstrings. 

Check the way you are removing the values: setting to zero isn't ideal because then you're still fitting, just expecting a zero value there, so the notch has to exactly match the notch in the data. If you just ignore this region, the notch can be made a bit bigger than in the data and not impact the fit.

plasmapy/diagnostics/thomson.py
bryancfoo: So my thought was that in most instances, there will only be one notch, so the user can input a 1D array with 2 endpoints. But I also wanted to give the flexibility to include multiple notches, so the user can also input a 2D array which is treated as a list of pairs, where each pair is a single notch. I have some checks here to parse both of those possible input types. I did discover np.ndim the other day, so that might be a more concise way to check for the dimensionality of the input.

plasmapy/diagnostics/thomson.py
bryancfoo: I can see how setting it to 0 in a larger range than the true notch would increase the value of \chi^2 (or some other goodness of fit metric) during fitting, but assuming that the fitting is only trying to find minima of \chi^2, shouldn't it still manage to properly fit the outer wings of the EPW spectrum, because the notch region just adds some fixed contribution to \chi^2? The reason I don't want to just cut it out of both arrays is because I eventually want to make the fitting functions do integral normalization, and by cutting it out of both, the entire trapezoidal region over the notch still contributes to the integral, which then becomes sensitive to noise at the edges of the notch. Also, np.trapz doesn't seem to have an elegant way to ignore nan values, which would also complicate the integral normalization in the nan approach,

changelog/2058.feature.rst
```diff
@@ -1 +1 @@
-Added a 'notches' keyword option to thomson.spectral_density, which allows users to notch out the output spectrum over a wavelength range. Also made some changes to the thomson fitting models to include the notch as an option.
+Added a 'notches' keyword option to ~plasmapy.diagnostics.thomson.spectral_density, which allows users to notch out the output spectrum over one or multiple wavelength ranges. Changed the ~plasmapy.diagnostics.thomson.spetctral_density_model function to allow notches to be applied during fitting.
```
namurphy, line 1:
```suggestion
Added a ``notches`` keyword option to `~plasmapy.diagnostics.thomson.spectral_density`, which allows users to notch out the output spectrum over one or multiple wavelength ranges. Changed the `~plasmapy.diagnostics.thomson.spectral_density_model` function to allow notches to be applied during fitting.
```
The single back ticks will make it show up in the online docs with a link to the function, and the double back ticks will make it appear as an in-line code snippet.

plasmapy/diagnostics/thomson.py
```diff
@@ -143,6 +143,13 @@ def spectral_density_lite(
         convolved with the spectral density function before it is
         returned.
 
+    notches : (2,) or (N, 2) `~numpy.ndarray`, |keyword-only|
```
namurphy, line 146:
```suggestion
    notches : (2,) or (N, 2) `~numpy.ndarray`, |keyword-only|, optional
```

plasmapy/diagnostics/thomson.py
```diff
@@ -357,6 +360,13 @@ def spectral_density(
         resulting array will be convolved with the spectral density
         function before it is returned.
 
+    notches : (2,) or (N, 2) `~astropy.units.Quantity`, |keyword-only|
```
namurphy, line 363:
```suggestion
    notches : (2,) or (N, 2) `~astropy.units.Quantity`, |keyword-only|, optional
```

plasmapy/diagnostics/thomson.py
```diff
@@ -357,6 +360,13 @@ def spectral_density(
...
+        A pair of wavelengths which are the endpoints of a notch over
+        which the output Skw is set to 0. Can also be input as a 2D array
+        which contains many such pairs if multiple notches are needed.
+        Should be in units convertible to u.m. Defaults to no notch.
```
namurphy, line 367:
```suggestion
        Should be in units convertible to meters. Defaults to no notch.
```

plasmapy/diagnostics/thomson.py
```diff
@@ -143,6 +143,13 @@ def spectral_density_lite(
         convolved with the spectral density function before it is
         returned.
 
+    notches : (2,) or (N, 2) `~numpy.ndarray`, |keyword-only|
+        A pair of wavelengths in meters which are the endpoints of a notch over
+        which the output Skw is set to 0. Can also be input as a 2D array
+        which contains many such pairs if multiple notches are needed.
+        Defaults to no notch.
+
```
namurphy, line 151:
```suggestion
```

plasmapy/diagnostics/thomson.py
pheuer: Hm, I guess that makes sense? Let's try it your way, maybe this will work better.

plasmapy/diagnostics/thomson.py
pheuer: Yes, that's good! Another option would be to accept either a two element np.array or a list of the same. Then you can do something like 

```python
if isinstance(notches, np.ndarray):
    notches = [notches, ]
elif isinstance(notches, list):
   pass
else:
    ValueError

# Check each element in the list notches to make sure it is a valid notch 
```

bryancfoo:
It looks like some of the tests in test_thomson.py already include notched spectra, although it's done manually outside of the thomson.spectral_density call. Would replacing those with my keyword version be enough, or are additional tests also needed?

pheuer:
> It looks like some of the tests in test_thomson.py already include notched spectra, although it's done manually outside of the thomson.spectral_density call. Would replacing those with my keyword version be enough, or are additional tests also needed?

Definitely replace those with your keyword, but organizationally it's nice to have a specific test for each element of functionality, so if the other test changes your keyword doesn't drop out of coverage. A simple test would be running the same function with and without the notch, verifying they are equal outside the notch region but the region within the notch is nan/zero (whichever it is?) inside the notch.

plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -215,6 +215,30 @@ def test_single_species_collective_spectrum(single_species_collective_spectrum):
         "instead of expected 17.7899"
     )
 
+def test_notched_spectrum(single_species_collective_spectrum):
```
pheuer, line 218: Good test! Since your code supports multiple notches, I think you should parametrize this test (using pytest parameterize) for two examples: 
1) what you have here - just one notch
2) A set of 2-3 notches 

That will verify the code is interpreting the specified notch input correctly

plasmapy/diagnostics/thomson.py
```diff
@@ -63,6 +63,7 @@ def spectral_density_lite(
     probe_vec: np.ndarray,
     scatter_vec: np.ndarray,
     instr_func_arr: Optional[np.ndarray] = None,
+    notches: Optional[np.ndarray] = None,
```
pheuer, line 66: Here you declare the type as ndarray, but earlier you are providing tuples. I think that works because they are acted upon the same, but it might be better to be consistent, or explicitly accept and handle both?

plasmapy/diagnostics/thomson.py
```diff
@@ -250,6 +257,13 @@ def spectral_density_lite(
     # Apply an instrument function if one is provided
     if instr_func_arr is not None:
         Skw = np.convolve(Skw, instr_func_arr, mode="same")
+
+    # add notch(es) to the spectrum if any are provided
```
pheuer, line 261: Subtle point - do we want to apply the notch before or after the instrument function? I think before - otherwise the IRF will convolve data from inside the notch region into the Skw on the edges of the notch, but that data doesn't exist in the experiment because the notch filter probably comes before the spectrometer slit? Check with Derek on this one, correct the order if necessary and add a comment clarifying why this is important (if it is) since it's subtle and we don't want it changed back!

plasmapy/diagnostics/thomson.py
pheuer: The notches keyword still needs to be added to the list of `settings` in the docstring for the model function.

pheuer (approved):
Looks good! This is a nice capability to have, thank you for adding it @bryancfoo !

PR #2059, opened 2023-04-04
Update changelog guide and Giles configuration

namurphy:
This PR makes the following changes:
 - Updated the changelog guide, in particular to fix an old link
 - Updated the configuration for Giles, which is what we use to check changelog entries. In particular, I updated the error messages including adding a link to the changelog guide.

I didn't include a changelog entry for the Giles configuration update, since it's not user facing and a pretty minor change anyway.

PR #2060, opened 2023-04-04
Update YAML file defining labels to apply to pull requests

namurphy:
The GitHub Action for labeling PRs has been working inconsistently.  My suspicion is that there need to be quotes surrounding the patterns in lists in `.github/labeler.yml`. I'm thinking (guessing?) that the quotes are needed because all of the examples in the [labeler documentation](https://github.com/actions/labeler#pull-request-labeler) have quotes.  

The quotes were being deleted by the YAML reformatter in our pre-commit configuration, so I set it to exclude that particular file.

I'm going to merge this quickly since it's one of those PRs that need to be merged before they can be tested (sigh).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,16 +2,16 @@ Breaking:
 - changelog/*.breaking*rst
 
 Continuous Integration:
-- any: [.codecov.yaml, .pre-commit-config.yaml, CODEOWNERS, noxfile.py, tox.ini]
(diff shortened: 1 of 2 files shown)


namurphy:
Looking at how the labels were removed when I updated #2059, it's looking like the change isn't working. I think the problem is with the `any` and/or `all` fields in `.github/labeler.yml`.  Hm...

PR #2062, opened 2023-04-04
Run packaging GitHub Action sooner

namurphy:
This PR modifies when the GitHub Action to check packaging PlasmaPy is run during continuous integration.  

At the moment, the packaging check is started after all other checks have passed. This means that the time to finish our continuous integration checks is $t_\mathrm{doc build} + t_\mathrm{packaging}$ (since the doc build takes significantly longer than everything else and is thus our CI critical path).

This PR makes it so that the packaging check is done after the initial tests are run without waiting for the comprehensive tests and the documentation build. As a consequence, our CI timescale will become $∼t_\mathrm{doc build}$.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -146,9 +146,6 @@ jobs:
     runs-on: ubuntu-20.04
     needs:
     - initial-tests
-    - comprehensive-tests
-    - documentation
-    - import-plasmapy
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


namurphy:
> Sounds good - just to make sure I understand, all you need to do is delete these three lines because they are basically requirements that hold up the rest of the test executing?

Yep!  That's exactly right.  

Thank you for the prompt review!  😸

PR #2061, opened 2023-04-04
Update configuration for labeler GitHub Action

namurphy:
This PR (sigh) is a follow-up to #2060, and an attempt to get the labeler GitHub Action to work.  

My hypothesis this time is that the problem is with labels that have a space in them not getting recognized.  I'm changing...

```yaml
Contributor Guide:
 - docs/contributing/**/*
```
to 

```yaml
'Contributor Guide':
 - docs/contributing/**/*
```
Maybe this will work?

PR #1811, opened 2022-11-29
Delete size-labeler.yml

StanczakDominik:

PR #2064, opened 2023-04-04
Add GitHub Action for labeling pull request size

namurphy:
Having a GitHub Action that labels PR size is helpful when doing code review triaging.  For example, it will often take only a minute or two to review a PR that is ≤ 10 lines of code (LOC), whereas it can take a multiple code review sessions to do one that is ≥1000 LOC.

We've tried adding a size PR labeler in the past, but for some reason we weren't able to get it to work.  I'm wondering if the reasons described in #2060 & #2061 might have contributed to the problem. 

This PR attempts to add a [PR size labeler](https://github.com/actions-ecosystem/action-size#example) GitHub Action.  Hopefully it'll work this time?

See also: #1492, #1800, #1801, #1811


namurphy:
Might be best to wait on this, since there's a deprecation warning for this action regarding a feature that'll be removed on 2023 June 1.

namurphy:
Before merging this, we'll need to:

 - [ ] Create the rest of the size labels, as defined in the action

PR #2066, opened 2023-04-06
Remove pydocstyle from dependencies

namurphy:
Looking through `pyproject.toml` and `tox.ini`, I don't see where `pydocstyle` is getting used now.  This PR attempts to remove `pydocstyle`.  There's a chance that it's getting used by flake8, however, which I need to look into a bit more.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,6 @@ tests = [
   "flake8-use-fstring",
   "hypothesis",
   "pre-commit",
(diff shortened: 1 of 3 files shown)


namurphy:
Putting this on hold for a moment, since it might be worth looking for a tool that will automagically check if we have unused dependencies.

namurphy:
Superseded by #2075.

PR #2047, opened 2023-03-30
Re-add GitHub Action to comment on someone's first pull request

namurphy:
This PR adds back and updates a GitHub Action that adds a comment when someone makes their first pull request to PlasmaPy.  We had tentatively removed it because it ran into some problems (discussed in #2044).

In the long term, I'm thinking it would be worthwhile to change the checklist (which we haven't gotten into the habit of using...) into a bullet point list of code/doc/test tips.  I also want to add a link that goes directly to the documentation preview, since it's now necessary to navigate an extra step.  I don't think I'll do that in this PR, though.

Closes #1965.  (along with #2044)

PR #2071, opened 2023-04-11
Update the comment that gets posted on pull requests

namurphy:
## Description

This PR changes the comment that gets posted on pull requests from a checklist into a collection of resources and useful information for especially new contributors.

## Motivation and context

We haven't gotten in the habit of using a code review checklist, so I figured that it would be more helpful to have a comment that provides useful information to contributors, including links to the appropriate resources. We might want to add a code review checklist back later on, though, since it is a recommended practice for code reviews.


namurphy:
We probably want to review and merge #2047 before this one, since I'll need to make sure that information doesn't get repeated in both.

namurphy:
I haven't quite finished the message to be posted on PRs, but I think I'm at the point where feedback would be really helpful.  There are tradeoffs between brevity (important) and including information that would be helpful for contributors (also important), which I haven't quite figured out.  At the very least, it shouldn't get any longer than this.  So, feedback would be quite helpful!

PR #2072, opened 2023-04-12
Extend allowed defaults for type specifications in docstrings

namurphy:
## Description

This PR expands allowed ways to specify defaults in the type specifications in the "Parameters" section of docstrings. This makes it explicit that we can include things like: `default: 32`, `default: [1, 2, 3]`, and `default: p+`.

## Motivation and context

Currently, we do "nitpicky" documentation builds to make sure that we do not end up with any unexpected broken links to Python objects in our documentation. However, when processing the type specification for a parameter in a docstring, Sphinx tries to link most words to Python objects...which leads to problems. We have to specify Python regular expressions for patterns to ignore so that Sphinx knows not to issue a warning (which gets converted into an error).  We then add these regular expressions to `nitpick_ignore_regex` in `docs/conf.py`.

I tried to be more specific so that we will still catch mismatch errors like `default: [1, 3)` and cases where we forget to include the default (i.e. `default:` without anything after).

I tested these patterns on [regex101.com](https://regex101.com).

## Related issues

This should resolve a problem found in #1984 by @gregordecristoforo.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -40,7 +40,7 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
     - name: Install Python dependencies
-      run: python -m pip install --progress-bar off --upgrade "tox<4" codecov
+      run: python -m pip install --progress-bar off --upgrade "tox"
     - name: Run tests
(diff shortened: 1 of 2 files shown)


namurphy:
Looks like we have a new requirements issue, seemingly from a new release of codecov.  I'll need to work on that later...

ejohnson-96 (approved):
Looks good :)

PR #2075, opened 2023-04-17
Update dependency minimum versions, package metadata, and coding guide

namurphy:
[NEP 29]: https://numpy.org/neps/nep-0029-deprecation_policy.html

This PR makes a couple of changes:

 - I set exact minimum requirements for ∼all of our dependencies in `pyproject.toml`. 
   - My motivation for setting exact minimum requirements is to make it easier to (eventually) automate the test where we install the minimum requirements. At present we have to manually update `tox.ini` with minimum dependencies, but if we switch to `nox`, we would be able to automatically extract the minimum dependencies from `pyproject.toml`. It's easy to forget to update `tox.ini`. For example, I just realized that I forgot to update `tox.ini`.
   - For required dependencies, I usually set the minimum version as something that was released ∼2 years ago (as inspired by [NEP 29]). 
   - For optional dependencies needed for running tests and building docs, I usually required more recent versions since there's much less chance of causing a version conflict with another package.
 - I updated keywords and classifiers in `pyproject.toml`.
   - I removed the `Python 3.9` and `Python 3.10` classifiers because they're redundant with `python_requires`, and it's common for packages to forget to update them.
 - I updated the coding guide discussion on exact/maximum requirements. 
 - I allowed the most recent version of `jinja2`, which we had previously not allowed because it had caused a documentation build error in #1487.


.pre-commit-config.yaml
```diff
@@ -32,11 +32,6 @@ repos:
     # run this hook on `labeler.yml`.
     exclude: .github/labeler.yml
 
-- repo: https://github.com/tox-dev/pyproject-fmt
-  rev: 0.9.2
-  hooks:
-  - id: pyproject-fmt
-
```
namurphy, line 39: I'm removing this pre-commit hook since it was doing things like changing `2.5.0` → `2.5`, where my goal is to have exact versions for the minimum requirements for when we automate it.  This hook doesn't have a way to turn it off though.

pyproject.toml
```diff
@@ -1,103 +1,105 @@
...
-  "wheel>=0.29",
+  "setuptools >= 61.2.0",
+  "setuptools-scm >= 6.3.2",
+  "wheel >= 0.37.0",
 ]
 
 [project]
 name = "plasmapy"
-description = "Python package for plasma physics"
+description = "Python package for plasma science"
```
namurphy, line 11: ...since there exists such a thing as plasma chemistry.

requirements.txt
namurphy, line 1: The changes in `requirements.txt` were done automatically.

.pre-commit-config.yaml
ejohnson-96: :thumbsup:

.pre-commit-config.yaml
```diff
@@ -95,7 +90,7 @@ repos:
   rev: 1.13.0
```
ejohnson-96, line 90: In line with your earlier comment about the trailing zeros, should rev: on ln: 90 and additional_depencies: ln: 93, also have their zeros truncated?

ejohnson-96 (approved):
A trailing white space needs to be removed from 'docs/contributing/coding_guide.rst', apart from that looks good 👍

.pre-commit-config.yaml
namurphy: Good eye!  The tool that automatically updates pre-commit hooks pins them to specific versions for stability, so the zeros should be there.  Thanks!

PR #2073, opened 2023-04-12
Update requirements and fix issue with version conflict related to coverage

namurphy:
The goal of this PR is to fix an issue we just started having with codecov.

```
  python -m pip install --progress-bar off --upgrade "tox<4" codecov
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.10/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.10/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.10/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.10/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.10/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.10/x64/lib
Collecting tox<4
  Downloading tox-3.28.0-py2.py3-none-any.whl (86 kB)
ERROR: Could not find a version that satisfies the requirement codecov (from versions: none)
ERROR: No matching distribution found for codecov
Error: Process completed with exit code 1.
```

Issue #2030, opened 2023-03-21
Fix docstring examples after upcoming change in order of printing Astropy units


namurphy:
In https://github.com/astropy/astropy/pull/14439, the method used in `astropy.units` to print out the units in a `Quantity` changed.  Looking at a recent weekly test failure, this affects a few examples in our docstrings which we'll need to update.

A complication is that we test against both newer and older versions of Astropy. Astropy's [changelog entry](https://docs.astropy.org/en/latest/whatsnew/5.3.html#change-in-order-in-unit-string-representations) for this change (https://github.com/astropy/astropy/pull/14526/files) lists some potential workarounds.  

I'm also wondering if we could change the order of units when create compound unit objects (i.e. changing `u.Ohm * u.m` to `u.m * u.Ohm` in the return annotation for functions decorated with `@validate_quantities`.   

Nice to see that our weekly test against the dev version of something was useful!

namurphy:
Closed by #2139.

PR #2079, opened 2023-04-25
Remove pip installation of `codecov` in weekly tests

namurphy:
This follows up on #2073 by removing `pip install codecov` from our weekly tests.  See #2073 for more details.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -76,7 +76,7 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
     - name: Install Python dependencies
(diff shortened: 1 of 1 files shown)


namurphy:
Merging this quickly because it's one of those PRs that needs to be merged to tell whether or not it works.

namurphy:
Looks like this worked!  We're still getting some weekly test failures, but for different reasons:
 - The astropy-dev tests are failing because the order that units are printed out for a `Quantity` got changed.  #2030
 - The numpy-dev tests are failing because of a problem finding the nightly wheel. There's a reasonable chance that this is a transient problem, so I'm going to wait a week before trying to fix it.

PR #2080, opened 2023-04-26
Enable more `ruff` rules and disable some redundant `flake8` checks

namurphy:
Up until recently, we used `flake8` as our primary linter. However, `flake8` is slow and does not make any automatic fixes.  In the last year, `ruff` is an extremely fast linter and autoformatter that includes style rules from `flake8` and a whole bunch of its extensions.

This PR enables more `ruff` style rules, and drops a few (now redundant) extensions from our `flake8` configuration.  

This is a very long pull request, but most of the changes in this PR were made automatically and don't include any changes to the logic.  I'm also reviewing the changes carefully so that the code review can hopefully go relatively quickly.


plasmapy/analysis/nullpoint.py
```diff
@@ -42,8 +42,6 @@ class NullPointError(Exception):
        change in future releases.
     """
 
-    pass
-
```
namurphy, line 46: The `pass` statement isn't necessary because the docstring is effectively a command.

plasmapy/dispersion/analytical/stix_.py
```diff
@@ -206,7 +206,7 @@ def stix(
     elif n_i.size == 1:
         n_i = np.repeat(n_i, len(ions))
 
-    species = ions + [Particle("e-")]
+    species = [*ions, Particle("e-")]
```
namurphy, line 209: This fix looks clever!  I hadn't thought about this before, but it's more clear that `ions` is a list here.

plasmapy/particles/tests/test_ionization_collection.py
```diff
@@ -814,9 +814,7 @@ def test_equality(self, this, that):
             print(f"{that} kwargs:\n {self.dict_of_kwargs[that]}\n")
             self.instances[that].summarize()
             descriptor = "equal" if expect_equality else "unequal"
-            pytest.fail(
-                f"Cases {this} and {that} should be {descriptor} but " f"are not."
```
namurphy, line 818: Sometimes when `black` autoformats things, it leaves two strings on the same line, since they get implicitly concatenated.

pyproject.toml
```diff
@@ -221,25 +219,33 @@ type_incorrect_long = "The filename of the changelog entry must be of the form `
...
+  "E501", # line-too-long
+  "RUF001", # ambiguous-unicode-character-string
+  "RUF002", # ambiguous-unicode-character-docstring
+  "RUF003", # ambiguous-unicode-character-comment
+  "RUF100", # unused-noqa (remove if we drop flake8 completely)
+  "TRY003",
+  "TRY301",
+  "UP025", # unicode-kind-prefix
+]
+extend-exclude = [
```
namurphy, line 243: Using `extend-exclude` instead of `exclude` lets us avoid having to re-specify files/paths that are already ignored by default.

plasmapy/plasma/tests/test_grids.py
```diff
@@ -225,10 +225,10 @@ def test_print_summary(abstract_grid_uniform, abstract_grid_nonuniform):
 ]
 
 
-@pytest.mark.parametrize("attr,type,type_in_iter,value", abstract_attrs)
+@pytest.mark.parametrize("attr,type_,type_in_iter,value", abstract_attrs)
```
namurphy, line 228: ...since `type` is a builtin.

.pre-commit-config.yaml
```diff
@@ -43,13 +43,6 @@ repos:
   - id: absolufy-imports
     exclude: docs/plasmapy_sphinx
 
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.3.1
-  hooks:
-  - id: pyupgrade
-    args: [--keep-runtime-typing, --py39-plus]
-    files: ^plasmapy/
-
```
namurphy, line 52: Changes made by `pyupgrade` are now included in `ruff`.

plasmapy/plasma/grids.py
```diff
@@ -1404,7 +1402,7 @@ def _nearest_neighbor_interpolator(self):
     def nearest_neighbor_interpolator(
         self, pos: Union[np.ndarray, u.Quantity], *args, persistent=False
     ):
-        r""" """
+        r""" """  # noqa: D419
```
namurphy, line 1405: See #2081.

docs/conf.py
```diff
@@ -545,6 +545,6 @@
 
 
 def setup(app: Sphinx) -> None:
-    app.add_config_value("revision", "", True)
+    app.add_config_value("revision", "", rebuild=True)
```
namurphy, line 548: Fixing a "boolean trap", where a boolean is used as an argument without the corresponding parameter name.

plasmapy/analysis/nullpoint.py
```diff
@@ -152,14 +146,14 @@ def _vector_space(
     x_arr=None,
     y_arr=None,
     z_arr=None,
-    x_range=[0, 1],
-    y_range=[0, 1],
-    z_range=[0, 1],
+    x_range=(0, 1),
+    y_range=(0, 1),
+    z_range=(0, 1),
```
namurphy, line 151: ...to make the default arguments immutable rather than mutable.

plasmapy/diagnostics/charged_particle_radiography/tests/test_synthetic_radiography.py
```diff
@@ -638,7 +639,7 @@ def test_saving_output(tmp_path):
 
     # Test that output cannot be saved prior to running
     with pytest.raises(RuntimeError):
-        _ = sim.results_dict
+        pass
```
namurphy, line 642:
```suggestion
        _ = sim.results_dict
```

namurphy:
The vast majority of the automatic changes are being made in subsequent PRs, and thus those should be merged before reviewing this one.

namurphy:
For reference for later, here are some ruff rule sets that I want to look into more.
 - COM
 - YYT
 - T10
 - EM
 - EXE
 - G
 - T20
 - PT
 - Q
 - RSE
 - RET
 - SLF
 - ERA
 - PGH
 - PLR
 - PLW
 - NPY ← the rule about the deprecated random number functionality

Other thoughts:
 - We should enable the NPY rule about deprecated random number functionality
 - Should we add more import conventions (i.e. `import numpy as np`)?
 - We should enable more FBT rules to check for boolean types

Not sure why the Python 3.9 checks started failing in this PR but not in #2092.  Weird!

ejohnson-96 (approved):
0% means nothing is changed right? 👍

namurphy:
> 0% means nothing is changed right? +1

Correct, I think!

PR #2098, opened 2023-05-02
Update pinned requirements and fix errors associated with NumPy 1.24

namurphy:
I noticed after updating requirements in #2080 that we started getting a failure from null point functionality.  This was probably related to the release of NumPy 1.24.  Updating the requirements here was able to reproduce the error, which I will fix when I have a chance to look into it.

PR #2099, opened 2023-05-02
Update codecov configuration

namurphy:
This PR makes some adjustments to our CodeCov configuration. See the [CodeCov documentation](https://docs.codecov.com/docs/codecovyml-reference) for more details on each item.

 - I increased the threshold to 1% since we already have really high test coverage, and there are occasionally situations where there is code that is harder to test (i.e. in the plasma calculator). 
 - I set `wait_for_ci` to `no` to make CodeCov submit its report before all of the status checks have finished. This should mean that it won't need to wait for the doc build to finish, for example.
 - I set `removed_code_behavior` to `adjust_base` will change how CodeCov deals with removed lines when calculating coverage.  This is essentially dealing with situations like when a PR  removes only lines that were covered by tests, so that the coverage percentage drops.


namurphy:
The CI failures are unrelated, so this is ready for review.

pheuer (approved):
Sounds like reasonable changes - if you're sure the CI errors are unrelated, then seems ready to merge.

namurphy:
Thank you for the review!  Merging this since the NumPy errors are being addressed in #2098.

PR #2102, opened 2023-05-03
Update codeowners for different files and subpackages

namurphy:
This PR updates the `CODEOWNERS` file, which is used to specify who gets requested for review by default.  I attempted to make it more fine-grained, which will make it less likely that we'll need to change reviewers around and that we preferentially get assigned reviews of subpackages that we're familiar with.  

I've also commented out a few lines so that those of us who aren't available to review right now are less likely to get assigned reviews, but I'd be happy to uncomment those lines when they're available to review again.  (By the way, I hope you're doing okay!)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,18 +1,42 @@
-# in decreasing order of priority
+# Define individuals or teams who are responsible for code in a
+# repository (i.e. who gets requested for a review).
 
-# any file, anywhere
-*                                                        @PlasmaPy/plasmapy-reviewers
-.pre-commit-config.yaml                                  @namurphy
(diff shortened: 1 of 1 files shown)


ejohnson-96 (changes requested):
Assuming the CI issues are related to the earlier PR, the docs are failing to build.
Something to do with the configuration being incorrect?

namurphy:
I fixed the NumPy errors in #2098 (not merged yet), and will submit another pull request soon to address the Read the Docs failures (sigh! 🫠).  In the meantime, I set the RTD check to be optional instead of required.

namurphy:
Thanks for the review!  We'll probably update this again within a few months as our availability changes.

PR #2103, opened 2023-05-03
Fix and update Read the Docs configuration

namurphy:
Our Read the Docs (RTD) configuration started failing within the last day, presumably after their newest release of [v9.12.0](https://docs.readthedocs.io/en/stable/changelog.html#version-9-12-0).

The purposes of this PR are to:
 - Fix the RTD build somehow.
 - See what happens when we use the melty face emoji in a branch name.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,7 +10,7 @@ build:
   os: ubuntu-22.04
   tools:
     # Use most recent version supported by Numba
-    python: '3.9'
+    python: '3.11'
   apt_packages:
   - graphviz
   jobs:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,10 +10,12 @@ build:
   os: ubuntu-22.04
   tools:
     # Use most recent version supported by Numba
-    python: '3.10'
+    python: '3.9'
   apt_packages:
   - graphviz
   jobs:
+    pre_create_environment:
+    - python -m pip install --upgrade setuptools pip
     post_build:
     - echo https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting
 
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -25,4 +25,3 @@ python:
     path: .
     extra_requirements:
     - docs
-  system_packages: true
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,7 +14,7 @@ build:
   - graphviz
   jobs:
     post_build:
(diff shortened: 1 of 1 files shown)


namurphy:
Cool!  For the second part, I got a warning that: 
```
The head ref may contain hidden characters: "readthedocs-\u{1FAE0}"
```
...which itself warrants the melty face emoji.

.readthedocs.yml
```diff
@@ -10,12 +10,10 @@ build:
   os: ubuntu-22.04
   tools:
     # Use most recent version supported by Numba
-    python: '3.10'
+    python: '3.11'
   apt_packages:
   - graphviz
   jobs:
-    pre_create_environment:
-    - python -m pip install --upgrade setuptools pip
```
namurphy, line 18: I don't know if this is necessary or not.

.readthedocs.yml
```diff
@@ -9,20 +9,16 @@ formats:
...
     post_build:
-    - echo https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting
+    - echo $'\n'For help deciphering documentation build error messages, see:$'\n\n'\ \ https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting
```
namurphy, line 17: I'm putting this on one line because RTD prints out the command before running it.

namurphy:
It's looking good to go!  Removing the `system_packages` option seems to have done the trick for us. It's usually better to go with default options unless we have a good reason otherwise, anyway.

The underlying issue appears to have been described in: https://github.com/readthedocs/readthedocs.org/issues/10285 and https://github.com/readthedocs/readthedocs.org/pull/10287

.readthedocs.yml
```diff
@@ -9,20 +9,16 @@ formats:
 build:
   os: ubuntu-22.04
   tools:
-    # Use most recent version supported by Numba
-    python: '3.10'
+    python: '3.11'
```
namurphy, line 12: A new version of Numba was released, so we can use Python 3.11 now!

.readthedocs.yml
pheuer: I suppose the answer is no, since you removed it and all the tests are still passing?

pheuer (approved):
The docs build, so I'd say this is good to go

I assume the `pre_create_enviroment` line turned out to not be necessary, since everything is building alright

.readthedocs.yml
namurphy: That does appear to be the case!

PR #2110, opened 2023-05-04
Auto-generate a list of modules

namurphy:
This PR is an attempt to remove the need for `api_static` stub files with a script that generates `docs/modules.rst`, which includes a list of all subpackages and modules in PlasmaPy.  This might work?


namurphy:
Only ∼4500 warnings!  The thing I forgot was that there needs to be an `automodapi` directive (I think) for each of the modules/subpackages.  🐰 🕳️ 🐉 

Another possibility would be to:
 - create `x :=` the set of all modules/subpackages by finding directories & files (as in this PR)
 - create `y :=` set of all of the modules/subpackages which have an automodapi directive in the docs (would be able to use `grep` for this were we using the command line)
 - `assert not (x - y), f"need api_static pages for {x-y}` ...

...maybe?

PR #2114, opened 2023-05-05
Remove test helper functionality from public API

namurphy:
This PR changes `plasmapy.tests.helpers` into `plasmapy.tests._helpers` and makes the module and `plasmapy.utils.pytest_helpers` into `plasmapy.utils._pytest_helpers`.  We've talked about removing these from PlasmaPy and possibly putting them into a separate package, so I've been meaning to mark them with a preceding underscore to mark them as private for a while now.


ejohnson-96 (approved):
There is a merge conflict with, pyproject.toml. I can't figure out why though, apart from that, looks good 👍

namurphy:
Thank you for the reviews!

namurphy:
Thanks!  It looks like I have to update the name of these files in the `ruff` settings (since I had set it to ignore a few error codes under the old filenames).

PR #2116, opened 2023-05-07
Transition from numba to rust

### LINKED ISSUE #2082: Re-implement lite-functions using Rust back-end (alternative to #2104)
### Feature description

We should consider re-implementing PlasmaPy's lite-functions using Rust instead of Numba.

### Motivation

We currently use formulary lite-functions for computationally intensive applications.  These functions operate on `np.ndarray` objects rather than `Quantity` objects.  

At present, we use Numba to get compiled speeds.  However, it takes a long time for Numba to get updated for new Python releases.  For example, it's been six months since Python 3.11 was released, but there hasn't been a release of Numba yet that works on Python 3.11 (apart from a recent pre-release). This means that we are preventing downstream packages (i.e. fiasco) from upgrading to Python 3.11.  Users may also run into problems where they install Python 3.11, but have trouble installing PlasmaPy.

There has been a lot of progress and excitement in the last few years about using Rust and Python together, and it looks like enough tooling is in place for us to implement a Rust back-end for our lite-functions.

### Implementation strategy

We could probably use [PyO3](https://pyo3.rs) to do this.

Since the Rust stuff is likely to be a small part of what's in PlasmaPy, we might want to put the Rust stuff in a separate package, and then import the lite-functions separately.  

### Additional context

The main alternatives are:
 - Stick with Numba, which has the downside of needing to sometimes wait 6 months to be able to use the newest version of Python.
 - Use Cython instead of Rust as the back-end.  In one benchmark, [Rust seems to be mostly faster than Cython](https://github.com/timleslie/rust_vs_cython).  Rust also has thread safety built-in.  And, there's more excitement about Rust, which makes it likely that there will be more great things in the future.  It would probably be quicker for Python users to learn Cython than Rust, since Cython is a superset of Python.
 - Use Julia instead of Rust as the back-end.  I've heard of a lot less activity in the Pythoniverse about connecting Julia and Python, so we'd want to make sure there's an active contributor community and that packaging won't be horrible. If we expand beyond lite-functions for Julia/Rust/Cython, Julia would have the advantage of having a whole lot of functionality for scientific computing already in the ecosystem, which doesn't exist yet in Rust.

JaydenR2305:
Resolves #2082 

A proof-of-concept lite function written in rust.

As per the advice of @pheuer I decided to first tackle the `plasma_frequency_lite` function. The biggest obstacle so far has been sanitizing data for use in rust, which is strongly typed. This makes it difficult to pass a mix of numeric types and ndarrays where python would not have any issues.

If anyone had any ideas for improving how input is sanitized I would love to hear them!


JaydenR2305:
I will also likely work on a cython prototype for comparison. I'm guessing that cython will be considerably better for working with ndarrays. We also won't see the overhead we're seeing here from having to convert python representations into rust arrays.

namurphy:
Thank you for working on this!  I wasn't expecting anyone to try this out so soon.

> The biggest obstacle so far has been sanitizing data for use in rust, which is strongly typed. This makes it difficult to pass a mix of numeric types and ndarrays where python would not have any issues.

This is an important lesson.  I wasn't sure if or how rust handles things like this.  I have very little experience with rust so I'm not sure.

> I'm guessing that cython will be considerably better for working with ndarrays. We also won't see the overhead we're seeing here from having to convert python representations into rust arrays.

Good point!  Cython is pretty well established, so I think it's likely that they've figured things like this out.

PR #2117, opened 2023-05-08
Fix doc build warning about frozen modules being used

namurphy:
When we updated to Python 3.11 (yay!) it appears that we started getting a noisy warning in our doc builds:
```
Debugger warning: It seems that frozen modules are being used, which may
make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
to python to disable frozen modules.
Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
```
This didn't cause a doc build warning/failure, but it made it hard to find and parse actual warnings.  The goal of this PR is to make it so that we don't see this warning.

Possible alternatives for our doc builds would include dropping down to Python 3.10 or not using a strict `requirements.txt` file.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -81,16 +81,16 @@ jobs:
       with:
         fetch-depth: 0
 
+    - name: Install graphviz and pandoc
+      run: sudo apt-get install graphviz pandoc
+
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
         python-version: '3.11'
 
-    - name: Install Python dependencies
+    - name: Install tox
       run: python -m pip install --progress-bar off --upgrade tox
-
-    - name: Install language-pack-fr and tzdata
-      run: sudo apt-get install graphviz pandoc
 
     - name: Build docs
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -81,17 +81,17 @@ jobs:
...
       with:
         python-version: '3.11'
 
-    - name: Install Python dependencies
+    - name: Install tox
       run: python -m pip install --progress-bar off --upgrade tox
 
-    - name: Install language-pack-fr and tzdata
-      run: sudo apt-get install graphviz pandoc
-
```
namurphy, line 94: Looks like we had this step named wrong, which I fixed above to reflect the packages that we're actually installing here.

requirements.txt
namurphy, line 1: I auto-generated the updates to `requirements.txt` here.  We dropped a few flake8 extensions (since they're now in ruff) but didn't update this file at the time.

pheuer (approved):
LGTM: the auto-generated requirements is clearly working if the tests are all building, I guess?

PR #2118, opened 2023-05-08
Enable NPY rules for ruff

namurphy:
Following up on #2080, this PR enables both of the [NPY rules](https://beta.ruff.rs/docs/rules/#numpy-specific-rules-npy) for ruff.  Previously, only `NPY001` had been enabled.  `NPY002` warns about usage of deprecated NumPy random number generation tools.  I also added `# noqa: NPY002` to lines that currently use the deprecated tools to silence the warning.

I plan to create an issue to update to newer NumPy random number functionality, but I don't think we need to be in any hurry about it.


ejohnson-96 (approved):
Looks good 👍

PR #2119, opened 2023-05-08
Enable ruff rule for cyclomatic (McCabe) complexity

namurphy:
[Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) is "a quantitative measure of the number of linearly independent paths through a program's source code."  When cyclomatic complexity is high, functions tend to be harder to understand, modify, & test.

This PR enables ruff rule [`C901`](https://beta.ruff.rs/docs/rules/#mccabe-c90) that issues warnings when the cyclomatic complexity exceeds a certain threshold.  I set the threshold to the default value of 10, but we may want to adjust it later since I don't know if that's the best value for scientific software development.  

A good first strategy for [reducing cyclomatic complexity](https://blog.codacy.com/reduce-cyclomatic-complexity/) is to write small functions that do one thing with no side effects.  A high level (big picture) function can call other mid-level functions, and the mid-level functions can call functions implementing low-level implementation details.  It also helps to reduce the number of decision structures that show up, and to avoid having too many flags in functions.  When we're making good use of Python classes, we can also [replace conditionals with polymorphism](https://refactoring.guru/replace-conditional-with-polymorphism).

If we have a good reason for exceeding the maximum cyclomatic complexity, we can add a `# noqa: C901` comment on the line with the `def` or `class` statement.  I also did that for the functions/methods that have a higher cyclomatic complexity than the threshold.

### DIFF
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -558,7 +558,7 @@ def _reduction(vspace, cell):
     return passX and passY and passZ
 
 
-def _bilinear_root(a1, b1, c1, d1, a2, b2, c2, d2):
+def _bilinear_root(a1, b1, c1, d1, a2, b2, c2, d2):  # noqa: C901
     r"""
     Return the roots of a pair of bilinear equations of the following
     format.
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -629,7 +629,7 @@ def _bilinear_root(a1, b1, c1, d1, a2, b2, c2, d2):
             return np.array([(x1, y1), (x2, y2)])
 
 
-def _trilinear_analysis(vspace, cell):
+def _trilinear_analysis(vspace, cell):  # noqa: C901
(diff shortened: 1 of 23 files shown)


namurphy:
My main hesitation about enabling this rule is that it might trip up new contributors.  Then again, it could end up being a learning opportunity.  Getting into the habit of writing short functions that do exactly one thing will very likely reduce future frustration.

pheuer (commented):
EDIT: responded to wrong PR 

This does potentially seem confusing to new contributors: would it for example flag functions with lots of input validation as `too complex` because they have many if/else statements?

ejohnson-96 (approved):
👍

namurphy:
Just talked with @pheuer about this PR and #2121, and we essentially decided to merge this on a trial basis.  

Part of the issue is that an error message like `function is too complex (14 > 10)` is not going to be clear for new contributors.  I'll create another PR to add a little more info about this error message and pre-commit checks to the message that gets posted on PRs.

PR #2121, opened 2023-05-08
Enable ruff rules for pylint refactorings

namurphy:
This PR enables some ruff [rules](https://beta.ruff.rs/docs/rules/#refactor-plr) for pylint refactorings.  I added a bunch of `# noqa` comments for the functions that were flagged already.

There are some rules that we might want to revisit again in the future.

 - `PLR0912` limits the number of branches (i.e., conditionals) in a function to 12
 - `PLR0915` limits the number of statements in a function to 50

If we have a good reason to exceed these limits, we can add a comment like `# noqa: PLR0912" to turn off the rule for a function.

### DIFF
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -629,7 +629,7 @@ def _bilinear_root(a1, b1, c1, d1, a2, b2, c2, d2):  # noqa: PLR0911, PLR0912
             return np.array([(x1, y1), (x2, y2)])
 
 
-def _trilinear_analysis(vspace, cell):  # noqa: PLR0911, PLR0912
+def _trilinear_analysis(vspace, cell):  # noqa: PLR0911, PLR0912, PLR0915
     r"""
     Return a true or false value based on whether a grid cell which has
     passed the reduction step, contains a null point, using trilinear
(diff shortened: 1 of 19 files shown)


namurphy:
> I guess I'm fine with these, although my hunch is that I'm going to end up disabling both of these on many functions. I'm not sure I agree that shorter functions are more readable than longer functions _if the shorter functions aren't conceptual subsections of the larger function_. For example, even just input validation often takes more than 50 lines (and a lot of if/else statements) for functions with many arguments?

There are definitely occasions where we would want to skip linter rules like these!

One way I like to think about this is that when we have a function that's really long, it's worth considering if there is another design pattern that we might implement.  

For example, I originally implemented `@particle_input` as a monolithic function with waaay too many nested conditionals and loops. I gave myself a whole bunch of headaches later on when I tried to change it.  I ended up refactoring [`@particle_input`](https://github.com/PlasmaPy/PlasmaPy/blob/59c24f2ba21f7c674a0e755d859fce4d4aa89111/plasmapy/particles/decorators.py#L145) into a class-based decorator, with the overall functionality broken down into a bunch of methods that each focused on a single task.  Coming back to it now, I'm finding it a lot easier to expand functionality.  

With regards to input validations, some alternatives would be to create a private function to handle them, create a decorator, or apply the [introduce parameter object](https://refactoring.guru/introduce-parameter-object) pattern (i.e., create a `Settings` class which itself would contain the input validations).

ejohnson-96 (approved):

PR #2122, opened 2023-05-08
Enable PLW rule set for ruff

namurphy:
This PR enables the [PLW](https://beta.ruff.rs/docs/rules/#warning-plw) rule set for ruff, and adds `noqa` comments for lines that ruff flagged.


pheuer (approved):
This seems like a reasonable rule set to include: these all seem like design patterns worth avoiding.

namurphy:
Thank you for the reviews! 😺

PR #2126, opened 2023-05-10
Parallelize `Tracker` in `synthetic_radiography`

### LINKED ISSUE #1728: Implement parallel processing for `charged_particle_radiography.Tracker`

JaydenR2305:
Attempt to solve embarrassingly parallelizable problem outlined in #1728 

Resolves #1728


JaydenR2305:
I seem to be getting a strange pickle error... 

```py
TypeError: ('Could not serialize object of type HighLevelGraph', '<ToPickle: HighLevelGraph with 1 layers.\n<dask.highlevelgraph.HighLevelGraph object at 0x7f2f48c782d0>\n 0. 139840971823616\n>')
```

JaydenR2305:
image

Is there a way around this?

pheuer:
@namurphy is in the process of configuring some new [ruff rules](https://beta.ruff.rs/docs/rules/#error-ple), and he may not have flagged all these old functions to disable the rule checking on them. First try merging in main to see if he's done this on a subsequent PR 

If not, I think you need to add `# noqa: PLR0912` to the first line of the function call. I think this is in main already now: 

```python

 def _make_grid(  # noqa: C901, PLR0912
        self,
        start: Union[int, float, u.Quantity],
        stop: Union[int, float, u.Quantity],
        num=100,
        units=None,
        **kwargs,
    ):
```

JaydenR2305:
@pheuer Thank you! Ruff is definitely more strict. I put up a prototype, but I seem to be getting nothing but noise on the synthetic radiograph.

PR #2128, opened 2023-05-10
Enable more `pydocstyle` rules for `ruff`

namurphy:
This PR goes through and enables several `pydocstyle` rules for `ruff` that had previously been disabled. This will better help consistentify our docs, which should make it easier for people to find the information that they need in docs sooner.  Thankfully, many of the checks have autofixes. 🛠️ These `ruff` checks will also make sure that things like missing docstrings will be found sooner, since it's easier to add a docstring at the time we add a function than a year later for a function someone else wrote.

I did end up disabling a bunch of pydocstyle rules for tests.  While documentation for tests can be quite important, there are times when it makes sense to not have a docstring (i.e., if the docstring would verbatim reproduce the name of the test function).


plasmapy/utils/decorators/converter.py
```diff
@@ -11,9 +10,9 @@
 
 
 def angular_freq_to_hz(fn):
-    """A decorator that adds to a function the ability to convert the
-    function's return from angular frequency (rad/s) to frequency
-    (Hz).
+    """
+    A decorator that enables a function to convert its return
+    value angular frequency (rad/s) to frequency (Hz).
```
namurphy, line 15:
```suggestion
    value from angular frequency (rad/s) to frequency (Hz).
```

pyproject.toml
```diff
@@ -251,41 +251,28 @@ extend-select = [
...
+  "PLE0605", # invalid-all-format (flags `__all__ += __lite_funcs`, etc.)
   "PLR0913", # too-many-arguments
   "PLR2004", # magic-value-comparison
   "PT007", # pytest-parametrize-values-wrong-type
   "PT011", # pytest-raises-too-broad
-  "PT012", # pytest-raises-multiple-statements
+  "PT012", # pytest-raises-multiple-statements (enable later?)
   "PT019", # pytest-fixture-param-without-value (enable later?)
-  "RUF001", # ambiguous-unicode-character-string
+  "RUF001", # ambiguous-unicode-character-string (flags
```
namurphy, line 272:
```suggestion
  "RUF001", # ambiguous-unicode-character-string (messes up mathematical symbols)
```

namurphy, line 272:
```suggestion
  "RUF001", # ambiguous-unicode-character-string (messes up mathematical symbols)
```

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -451,7 +451,7 @@ def _apply_wire_mesh(
         mesh_vdir=None,
     ):
         """
-        Apply wire meshes that were added to self.mesh_list
+        Apply wire meshes that were added to self.mesh_list.
```
ejohnson-96, line 454: I see that you added the full stop (period) but in this case, could it potentially cause confusion? I am thinking of a newish user who may think it is included in the code

namurphy, line 454:
```suggestion
        Apply wire meshes that were added to ``self.mesh_list``.
```

plasmapy/particles/data/test/__init__.py
```diff
@@ -1,6 +1,6 @@
 """
 Test`.h5` files came from
-https://github.com/openPMD/openPMD-example-datasets/tree/b4f87b817629b99a048026a2724a5b265810d8be
+https://github.com/openPMD/openPMD-example-datasets/tree/b4f87b817629b99a048026a2724a5b265810d8be.
```
ejohnson-96, line 3: Again, with the full stop (period), wouldn't the link be broken?

ejohnson-96 (approved):
👍

PR #2139, opened 2023-05-13
Fix failing astropy weekly tests

JaydenR2305:
Numpy has made some updates to unit ordering (still trying to find which specific commit, or even if this was intended), causing tests dependent on string representation of quantities to fail.

### DIFF
--- a/plasmapy/formulary/tests/test_magnetostatics.py
+++ b/plasmapy/formulary/tests/test_magnetostatics.py
@@ -39,7 +39,7 @@ class Test_MagneticDipole:
     def test_repr(self):
         """Test __repr__ function"""
         B1 = MagneticDipole(self.moment, self.p0)
-        assert repr(B1) == r"MagneticDipole(moment=[0. 0. 1.]A m2, p0=[0. 0. 0.]m)"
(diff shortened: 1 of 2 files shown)


plasmapy/formulary/tests/test_magnetostatics.py
```diff
@@ -39,7 +39,7 @@ def test_value2(self):
     def test_repr(self):
```
namurphy, line 39:
```suggestion
    @pytest.mark.skipif(astropy.__version__ < "5.3.0", "change in unit representation")
    def test_repr(self):
```
What would you think of setting this test to run only for more recent versions of Astropy?  We'd need an `import astropy` at the top of the file too.

namurphy (commented):
Thank you for doing this!  I had a few suggestions in case we'd want to merge this sooner.

plasmapy/formulary/tests/test_magnetostatics.py
JaydenR2305: Good thinking. Should we plan on bumping the version in the near future?

namurphy: We generally try to support versions of Astropy that have been released within the past ∼2 years, so it'll be a while before we require Astropy ≥ 5.3.0 since it hasn't been released yet.

namurphy (approved):

PR #2140, opened 2023-05-15
Add `mass_numb` and `Z` as parameters to functions decorated with `@particle_input`

namurphy:
If a function is decorated with `@particle_input` and also has `mass_numb` and `Z` as parameters to the function, then we can specify the mass number via `mass_numb` and charge number via `Z`.  This PR adds `mass_numb` and `Z` to a few functions in `plasampy.formulary.lengths` and `plasmapy.formulary.distribition` since I forgot to add it before.

Follows up on #1732.  Related to #341.


ejohnson-96 (approved):
👍

PR #2142, opened 2023-05-15
Mark tests as slow

namurphy:
I did `pytest -m "not slow" --durations=50` to find the 50 tests that were taking the longest to run.  I'm decorating them with `@pytest.mark.slow()` so that we can run `pytest -m "not slow"` to skip tests that take a relatively long time to run.  

I usually define tests as "slow" if they take longer than half a second to run, with a couple of exceptions for functionality that takes significantly longer the first time it is run (i.e., something decorated with `@numba.jit`).


namurphy:
Thank you for the review!

> You said there were 50 tests: are most of those cases of the `temp_ratio` `test_raises` function? I find it surprising that an exception test is taking up so much time...

No, I didn't see these taking up too much time.  I'm not sure why it would take a long time, but if it ends up being a bottleneck, it might be worth looking into.

PR #2145, opened 2023-05-21
Add a weekly test for conda-forge installation

### LINKED ISSUE #1638: Add a weekly test that most recent release can be installed via `conda` & `conda-forge`
We are currently running into some issues with `conda` (#1637) that we didn't notice.  It'd be helpful to have a cron test that makes sure that the most recent release can be installed via `conda` and `conda-forge`.  It should be set to run by workflow dispatch too so that we can run it manually after a release.

I had thought about adding this to our conda-forge feedstock repo, but we might end up forgetting that it's there and miss test failures.

JaydenR2305:
Resolves #1638

Adds a weekly test to the workflow to ensure that PlasmaPy can be installed via the conda-forge channel.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -7,15 +7,14 @@ on:
 
 jobs:
   conda_installs:
-    name: ${{ matrix.name }}
+    name: Install PlasmaPy from ${{ matrix.channel }} channel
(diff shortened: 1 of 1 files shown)


JaydenR2305:
Should I add a changelog for this?

namurphy (approved):

PR #2149, opened 2023-05-23
Update some `ruff` rules and consistentify `pyproject.toml` formatting

namurphy:
This PR makes a few slight changes to the `ruff` configuration.

 - For the RUF rule set, I made it so that we specify which rules are followed rather than which rules are ignored. The RUF rule set will likely to change a lot, so this will make it so that we have fewer surprises.  The tradeoff is that we'll need to update this later, but we'll need to update `ruff` rules periodically until it stabilizes anyway.
 - I made some other small updates to the rules, with autofixes.


pheuer (approved):
Agreed that this is a better way of doing it. I guess just set a reminder to revisit this list, maybe before the next release?

PR #2152, opened 2023-05-23
Add GitHub Action to remove labels when a PR is merged and update PR labeler

namurphy:
This PR adds a GitHub Action that will remove a set of labels from PRs when a PR is merged.  The labels to be removed are only relevant to open PRs (i.e., `status: in progress`).  Previously, I had been occasionally removing these labels manually, so this will automate a slightly tedious task.

I used ChatGPT 4.0 to create this GitHub Action.  It might be necessary to merge this to test whether or not it's working.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -6,43 +6,52 @@
 # If a label includes a space or a pattern begins with an asterisk or
 # special character, enclose it in quotes.
 
-Breaking:
+'changes existing API':
 - changelog/*.breaking*.rst
 
-Bug:
+bug:
 - changelog/*.bugfix*.rst
 
-'Continuous Integration':
+'CI':
 - .codecov.yaml
 - .github/workflows/*test*.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/new-pull-request.yml
```diff
@@ -1,4 +1,4 @@
-name: Comment on PR
+name: Open PR
```
namurphy, line 1: For naming symmetry!

.github/labeler.yml
```diff
@@ -6,43 +6,52 @@
 # If a label includes a space or a pattern begins with an asterisk or
 # special character, enclose it in quotes.
 
-Breaking:
+'changes existing API':
```
namurphy, line 9: I renamed & consolidated a few labels yesterday, so it seems a good time to update this.

ejohnson-96 (approved):
👍

PR #2154, opened 2023-05-24
Add `scripts` folder for Python scripts run during GitHub Actions

namurphy:
This PR adds a `.github/scripts/` folder to contain the Python script that was formerly in `.github/workflows/`.  This helps separate the Python stuff from the workflow file stuff. 

(I'm also using as a test to see if the ChatGPT generated PR label remover GitHub Action actually works.)


pyproject.toml
```diff
@@ -310,7 +310,7 @@ max-complexity = 10
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["D104", "E402", "F401", "F402", "F403"]  # ignore import errors
-".github/workflows/citation_updater.py" = ["D103"]
+".github/scripts/citation_updater.py" = ["D103", "INP001"]
```
namurphy, line 313:
```suggestion
".github/scripts/*.py" = ["D103", "INP001"]
```

PR #2155, opened 2023-05-24
Add a GitHub Action to check that authors are included in `CITATION.cff`

namurphy:
The purpose of this PR is to add a GitHub Action that checks that the GitHub username of all authors of a PR (excluding bots) are appropriately included in `CITATION.cff`. If any authors are missing, then this check will fail and a message will be posted on how to add oneself as an author.

This GitHub Action was mostly generated by ChatGPT-4.  Or more accurately, I wanted a reason to try out using ChatGPT-4, so I generated this GitHub Action. 🥳 

I tested most of this action on a [separate repo](https://github.com/namurphy/TestGitHubActions), except for the part that checks that the comment is only posted once.

Closes #1601.

PR #2156, opened 2023-05-24
Generate author list in `docs/about/credits.rst` from `CITATION.cff`

### LINKED ISSUE #1602: Generate author list from `codemeta.json`

namurphy:
This PR extracts authors from `CITATION.cff`, puts it into a reStructuredText formatted list, and outputs it to `docs/about/_authors.rst`.  Then the generated author list is included directly by `docs/about/credits.rst`. 

Another ChatGPT-4 PR!  This one probably would have taken ∼10–20 times as long had I done solely by myself. I still need to update docs on how to add oneself to an author list.  This should be merged after #2155.

Closes #1602.


docs/conf.py
```diff
@@ -13,6 +13,11 @@
 from pkg_resources import parse_version
 from sphinx.application import Sphinx
 
+from . import cff_to_rst  # noqa
```
namurphy, line 16:
```suggestion
from . import cff_to_rst
```

namurphy:
- [ ] Double check that no authors got dropped (i.e., that all authors from `credits.rst` are in `CITATION.cff`)

ejohnson-96 (approved):
👍

PR #2166, opened 2023-05-25
Update GitHub Action to regenerate requirements

namurphy:
Given that I had trouble with getting tests to run in this Action, I decided to add some comments. Also I needed to fix the link in the Easter egg.

### DIFF
--- a/.github/workflows/update-pinned-reqs.yml
+++ b/.github/workflows/update-pinned-reqs.yml
@@ -49,13 +49,11 @@ jobs:
       with:
         token: ${{ steps.generate_token.outputs.token }}
         commit-message: Regenerate requirements.txt
-        title: Automagically update pinned requirements
+        title: Update pinned requirements
         body: |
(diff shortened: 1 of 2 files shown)


namurphy:
The problems I ran into:
 - I kept getting: `Error: secretOrPrivateKey must be an asymmetric key when using RS256`. The problem was that when I created the private key from the GitHub App, I had copied the SHA instead of the full contents of the file that got downloaded when I created the key. This error came from the `tibdex/github-app-token@v1` action.
 - The App ID does not refer to the name of the App, but rather the number that is provided on the settings page for the App.  (This was confusion I had from the [instructions](https://github.com/peter-evans/create-pull-request/blob/4ddb8c8fe7f58d02d4d1d62359fd91fe4d54361d/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens) on the `create-pull-request` action about authenticating with GitHub app generated tokens.)

PR #2170, opened 2023-05-26
Drop `flake8` and its extensions from linters

### LINKED ISSUE #1837: Migrate from `flake8` to `ruff`
We currently use [`flake8`](https://flake8.pycqa.org/en/latest/) as our primary linter. `flake8` is widely used and well-established, with a wide ecosystem of [awesome extensions](https://github.com/DmytroLitvinov/awesome-flake8-extensions).

[`ruff`](https://github.com/charliermarsh/ruff) is a linter written in [Rust](https://www.rust-lang.org/).  At present, `ruff` appears to have feature parity with the checks done by `flake8`, and has been gaining a lot of traction.  A possibility would be for us to migrate from `flake8` → `ruff`, or perhaps make use of both.

Some advantages of `ruff` over `flake8`:

 - `ruff` has autofix support (major)
 - `ruff` is ∼10–100 times faster while having feature parity (minor)
 - `ruff` is compatible with `pyproject.toml` (minor, see #1835)
 - `ruff` could replace other tools like `isort` and `pyupgrade` (minor)

Some disadvantages of switching from `flake8` to `ruff`:

 - It'd take some time and effort
 - Our `flake8` extensions might not have equivalents in `ruff`
 - We'd presumably have a harder time writing a `ruff` extension than a `flake8` extension (though I hope we'll never need to do either)

I'm labeling this as a proposal and on hold since we should talk about it at a community meeting, and low priority since flake8 has been working well for us.  This could end up being a good issue for someone who would like to contribute to PlasmaPy but and has experience with Python/flake8 but not plasma physics.

namurphy:
We've been rapidly switching from `flake8` and its extensions to `ruff`, in particular because `ruff` can replace almost all of the `flake8` extensions that we have used.  Hence, it doesn't benefit us very much to have `flake8` in addition to `ruff`.  Dropping it will also reduce the dependencies we need for our tests/linters and in our pinned dependencies. Now seems like as good a time as any to drop `flake8` entirely in favor of `ruff`.

One of the `flake8` extensions not yet implemented by `ruff` is `flake8-rst-docstrings`, though `ruff` does have a bunch of `pydocstyle` rules which take care of the most important things that `flake8-rst-docstrings` checks. Plus, most of the errors that `flake8-rst-docstrings` would catch are also checked for by Sphinx.

Closes #1837.


changelog/2170.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Dropped flake8_ and its extensions. Instead, ruff_ is now used as the
+primary linter.
```
namurphy, line 2:
```suggestion
Dropped flake8_ and its extensions as linters. Instead, ruff_ is now used as the primary linter.
```

ejohnson-96 (approved):
👍

namurphy:
Thank you for the reviews!

Issue #2178, opened 2023-05-30
Make certain functions correctly account for mass of electrons in ions when `z_mean` is provided


namurphy:
There are a few functions that accept the charge number as `z_mean`, but do not use it to calculate the mass of an ion.  For example, if a proton is represented with `particle="H-1"` and `z_mean=1`, then the mass used in the formula will be the mass of a neutral hydrogen atom (including the electron) rather than the mass of a proton.

Consequently, the mass used within these functions could be off by perhaps $\lesssim \frac{1}{1837} \approx 0.054$%. This shouldn't noticeably affect anything, except for when the square root of the mass is what gets used in the formula.

From what I can tell, the following functions are affected.

 - [ ] `plasmapy.formulary.frequencies.gyrofrequency` (to be fixed by #2026)
 - [ ] `plasmapy.formulary.speeds.Alfven_speed`
 - [ ] `plasmapy.dispersion.analytical.two_fluid_.two_fluid`
 - [ ] `plasmapy.dispersion.analytical.hollweg_.hollweg`
 - [ ] `plasmapy.dispersion.numerical.kinetic_alfven_.kinetic_alfven`
 - [ ] `plasmapy.formulary.braginskii.ClassicalTransport` (need to double check...)

This problem can be fixed by decorating these functions by `@particle_input`, which accounts for the charge number to create the particle mass.  

This shouldn't directly affect lite functions, though the masses provided to lite functions may have this error in some occasions.

PR #2026, opened 2023-03-18
Decorate `plasma_frequency` with `particle_input`

namurphy:
## Description

This PR applies `particle_input` to `plasmapy.formulary.frequencies.plasma_frequency`, and also makes some updates to the docstring.  It also changes the `z_mean` parameter to `Z` for this function and `plasma_frequency_lite`.  

This PR also corrects a bug where the mass used to calculate the plasma frequency did not correctly account for the number of electrons when the charge number was provided to `z_mean` (see comments for details).

## Motivation and context

This PR is making `plasma_frequency` compatible with `CustomParticle`-like and `ParticleList`-like inputs.

## Related issues

This PR follows up on #1869.  Part of #341.

PR #2179, opened 2023-05-30
Decorate `Alfven_speed` with `@particle_input`

namurphy:
This PR decorates `Alfven_speed` with `@particle_input` in order to:

 - Broaden the acceptable particle-like inputs that can be provided to it
 - Fix a bug related to the electron masses not being correctly accounted for when `z_mean` was provided separately (which also affects `Lundquist_number`)
 - Deprecates `z_mean` as a parameter in favor of `Z` (note that `z_mean` can still be provided, in which case `@particle_input` will issue a deprecation warning)
 - Simplify the implementation within `Alfven_speed`

Related to #2178 and #341.  See also #2026.


ejohnson-96 (approved):
👍

PR #2181, opened 2023-05-31
Decorate `hollweg` with `@particle_input`

namurphy:
This PR decorates `plasmapy.dispersion.numerical.hollweg_.hollweg` with `@particle_input`.  It also makes some style changes to the docstring, etc.  This also fixes (for this function) the problem in #2178 about the mass of electrons not being correctly taken into account for ion mass if `z_mean` is provided.

Related to #341. Similar to #2022.


ejohnson-96 (approved):
👍

PR #2201, opened 2023-06-08
Lower ipykernel requirement

namurphy:
This PR lowers the minimum requirement for `ipykernel` so that it is able to work within Google Colab.


namurphy:
Oops, based this off of the wrong branch.

Issue #2197, opened 2023-06-06
Add dispersion relation for MHD waves


namurphy:
### Feature description

Add functionality for the dispersion relations for Alfvén, slow magnetosonic, and fast magnentosonic waves.

### Motivation

We currently have some two-fluid dispersion relation solvers, but we don't have anything for plain MHD waves.  Adding functionality for MHD wave dispersion relations would be particularly useful as a tool for pedagogy, and would also be useful as a limiting case for researchers.

### Implementation strategy

We probably want to think carefully about the interface to make sure it is intuitive, in particular for newcomers to plasma physics.

Some of the possibilities would be:

 - Have separate classes for `AlfvenWave`, `FastMagnetosonicWave`, and `SlowMagnetosonicWave` with a method for calculating $\omega$ as a function of $k$ and $θ$.
   - There could be a helper function that would return a namedtuple or that includes an instance of each of the three different types of waves with appropriately named attributes.
 - There could be a single `MHDWaves` class

Overall I'm not sure what the best approach would be.

We might want to add a method for plotting a Friedrichs diagram too.

### Additional context

If I remember right, there's also an entropy mode that doesn't propagate.  We might want to add this too.

TrestanSimon:
I would be very interested in working on this.

> Have separate classes for `AlfvenWave`, `FastMagnetosonicWave`, and `SlowMagnetosonicWave` with a method for calculating ω as a function of k and θ.

Since the two fluid DR solver, for example, is just a function in the [plasmapy.dispersion.analytical](https://docs.plasmapy.org/en/stable/api_static/plasmapy.dispersion.analytical.html#module-plasmapy.dispersion.analytical) subpackage, would it make sense that the MHD wave DR functionality also be implemented in this way? (I have not contributed to PlasmaPy in a while; maybe I am missing something here.)

namurphy:
@TrestanSimon — thank you!

> Since the two fluid DR solver, for example, is just a function in the [plasmapy.dispersion.analytical](https://docs.plasmapy.org/en/stable/api_static/plasmapy.dispersion.analytical.html#module-plasmapy.dispersion.analytical) subpackage, would it make sense that the MHD wave DR functionality also be implemented in this way? (I have not contributed to PlasmaPy in a while; maybe I am missing something here.)

I've been wondering about if there would be a way to improve the user-friendliness of the dispersion functionality.  For the MHD case, I'm wondering if we would want to do something like:

[...]

Issue #2198, opened 2023-06-06
Add example notebook for MHD wave dispersion relations


namurphy:
Following up on #2197, it would be really helpful to include an example notebook that covers the three different types of MHD waves. The notebook should give an understandable introduction to Alfvén, fast magnetosonic, and slow magnetosonic waves.  It would be helpful to include the formulae for the dispersion relations, and talk about what is physically happening in each of the waves.  Diagrams indicating the nature of each of the waves would be helpful too.

I keep meaning to do this every year, but it would be really helpful to have this in place by the SULI summer course in June 2024.

TrestanSimon:
Hi @namurphy. Now that #2206 has merged, I am interested in working on this if that is alright. Before I begin, I have two quick questions:

1. Do you think a brief derivation of the dispersion relations should be worked into the notebook? (Maybe something like [this](https://en.wikipedia.org/wiki/Magnetosonic_wave#Dispersion_relation))
2. Do you think the "diagrams indicating the nature of each of the waves" should be generated with PlasmaPy/Python or embedded from elsewhere (something like [this](https://commons.wikimedia.org/wiki/File:Magnetohydrodynamic_shear_Alfv%C3%A9n_wave_linear_perturbations_diagram_perspective.svg))?

namurphy:
> Now that #2206 has merged, I am interested in working on this if that is alright. 

That would be fantastic!   In case you'll be at the APS DPP meeting next week, we'll have a table there if you'd like to stop by.

>     1. Do you think a brief derivation of the dispersion relations should be worked into the notebook? (Maybe something like [this](https://en.wikipedia.org/wiki/Magnetosonic_wave#Dispersion_relation))

I think a brief derivation would be helpful.  An alternative place would be the docstrings of the classes or module.  We wouldn't be able to directly use the Wikipedia text since it is under the copyleft CC-BY-SA license and PlasmaPy is under the permissive BSD 3-clause license, but we could add something similar.

>     2. Do you think the "diagrams indicating the nature of each of the waves" should be generated with PlasmaPy/Python or embedded from elsewhere (something like [this](https://commons.wikimedia.org/wiki/File:Magnetohydrodynamic_shear_Alfv%C3%A9n_wave_linear_perturbations_diagram_perspective.svg))?

I think it'd be fine to use a graphic from elsewhere, and I like the one you pointed to.  I almost mentioned license compatibility here too, but it looks like you're the one who created that figure which is awesome!

Thank you again!

[...]

PR #2206, opened 2023-06-11
Add dispersion relation for MHD waves

TrestanSimon:
## Description

>Add functionality for the dispersion relations for Alfvén, slow magnetosonic, and fast magnentosonic waves.

## Motivation and context

From  https://github.com/PlasmaPy/PlasmaPy/issues/2197:
> We currently have some two-fluid dispersion relation solvers, but we don't have anything for plain MHD waves. Adding functionality for MHD wave dispersion relations would be particularly useful as a tool for pedagogy, and would also be useful as a limiting case for researchers.

## Implementation

There are three separate classes for each wave mode (`AlfvenWave`, `FastMagnetosonicWave`, and `SlowMagnetosonicWave`) which all derive from an abstract base class `AbstractMHDWave`. Each wave mode class overrides the abstract method `angular_frequency` from the base class. The associated constructors have required arguments for $B$, the ion species, and the ion number density* with optional arguments for the temperatures and adiabatic indices of the electrons and ions and the mean ionization state.

*Along with the wavenumber and angle of propagation provided when calling the `angular_frequency` method, these are the only parameters needed for calculating the angular frequency of the Alfvén wave and of the fast and slow magnetosonic waves in the cold plasma limit. (The temperatures will default to zero.)

Note also that some input validation code has already been written for [plasmapy/dispersion/analytical/two_fluid_.py](https://github.com/PlasmaPy/PlasmaPy/blob/e125e69c15dacbbaa6509c8d5c274d7c4d416374/plasmapy/dispersion/analytical/two_fluid_.py), so some of it is reused here.

## Related issues

Closes https://github.com/PlasmaPy/PlasmaPy/issues/2197
See also https://github.com/PlasmaPy/PlasmaPy/issues/2198


namurphy (changes requested):
@TrestanSimon — thank you for submitting this!  I'm excited to have it be added. 🚀 

I have a few suggested refinements (mostly related to implementing `@particle_input`), which I meant to post earlier but was traveling.

Thank you again! 😸

plasmapy/dispersion/analytical/mhd_wave_class.py
```diff
@@ -0,0 +1,262 @@
...
+    ValueError
+        If ``ion`` is not of category ion or element.
+
+    ValueError
+        If ``B``, ``n_i``, ``T_e``, or ``T_I`` are not single valued
+        `astropy.units.Quantity` (i.e. an array).
+    """
+
+    def angular_frequency(self, k: u.rad / u.m, theta: u.rad):
+        """
```
namurphy, line 184:
```suggestion
        r"""
```

plasmapy/dispersion/analytical/tests/test_mhd_wave_class.py
```diff
@@ -0,0 +1,101 @@
...
+        [
+            (  # Cold plasma limit
+                {**_kwargs_plasma_cold},
+                {
+                    "alfven": [_v_A, 0 * u.m / u.s],
+                    "fast": [_v_A, _v_A],
+                    "slow": [0, 0] * u.m / u.s,
+                },
+            ),
+            (  # No magnetic field
```
namurphy, line 52: Good idea to test this limit!

namurphy (commented):
Thank you for working on this!  I have a few minor suggestions now, and will hopefully have a chance to look through this in a bit more detail later this week.

Also — sorry about the delay!  I was traveling last week. ✈️

plasmapy/dispersion/analytical/tests/test_mhd_wave_class.py
TrestanSimon: Thanks. I actually have a question about this particular limit. Without a magnetic field, the MHD fast mode would reduce to a "sound wave" regardless of the ratio between $T_e$ and $T_i$; however, from the two-fluid model, ion acoustic waves could only exist if $T_e \gg T_i$. Do you think some sort of warning should be added?

TrestanSimon:
@namurphy Thanks for all of your suggestions.

I am unfamiliar with how docstring inheritance works between child and parent classes and for overridden methods with Sphinx. I have had trouble finding consistent info about this online and have not seen any examples of it within PlasmaPy. Is there some decorator I should use for this? Or would something like `__doc__ += ...` work?

plasmapy/dispersion/analytical/tests/test_mhd_wave_class.py
namurphy: Good question!  Usually for MHD, it's assumed that the ion & electron temperatures are the same.  Most of the references I've found assume that the sound speed is given by the ideal relation

$$ c_s ≡ \frac{γ k_B T}{m_i} $$

(with $γ = \frac{5}{3}$) which I believe should be used instead of the ion sound speed used for ion acoustic waves. I'll double check this with some friends who work with plasma waves a lot more than I do.  If we assume the same temperature for ions & electrons, we would remove the need for that warning.

namurphy:
> I am unfamiliar with how docstring inheritance works between child and parent classes and for overridden methods with Sphinx. I have had trouble finding consistent info about this online and have not seen any examples of it within PlasmaPy. Is there some decorator I should use for this? Or would something like `__doc__ += ...` work?

Another good question! I usually end up putting entirely new docstrings in child classes.  This sometimes leads to repetition of the content in child class docstrings, but for the most part, I've found it simpler to have new child class docstrings than to try to add code/logic that modifies (or appends to) the docstrings of the parent classes for the child classes.  Adding in code/logic that modifies docstrings could end up being worth it in more complicated cases (like with a broader range of child classes).

namurphy (commented):
Thank you again for working on this!  I'm liking how this is coming together!  Here are a few thoughts & suggestions before I head home for the weekend.  I want to spend a bit of time playing around with the classes too, hopefully by next week.  This will be great to have in our next release in September!

plasmapy/dispersion/analytical/mhd_wave_class.py
```diff
@@ -0,0 +1,766 @@
...
+                   [4.36121946e+01, 3.08384785e+01, 2.67047673e-15]] rad / s>
+        """
+        theta, k = super()._validate_k_theta(k, theta)
+        omega = k * self._v_a * np.cos(theta)
+        return super()._validate_angular_frequency(omega)
+
+
+class FastMagnetosonicWave(AbstractMHDWave):
+    r"""
+    A class to represent fast magnetosonic waves.
```
namurphy, line 384: I wonder if we'd want to add a note somewhere (maybe in the notes section?) that this is also known as the fast magnetoacoustic wave (and similarly for the slow wave below).  That'd help in case people do a search for that term instead.

plasmapy/dispersion/analytical/mhd_wave_class.py
```diff
@@ -0,0 +1,766 @@
...
+        self._op = plasma_frequency(_n, ion)
+
+    @property
+    def alfven_speed(self):
+        """The Alfvén speed of the plasma."""
+        return self._v_a
+
+    @property
+    def sound_speed(self) -> u.m / u.s:
+        """The sound speed of the plasma."""
```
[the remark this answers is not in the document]
TrestanSimon: I think it would be a good idea.

plasmapy/dispersion/analytical/mhd_wave_class.py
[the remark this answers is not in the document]
TrestanSimon: I was thinking about doing this previously, but I was not sure of what the new name should be. I think this will work.

TrestanSimon:
@namurphy Thanks for taking a look at this again. I have implemented all of your suggested changes.

Regarding `mhd_waves` returning a `collections.namedtuple`, I am not sure how to go about writing the new "Returns" section of the docstring. I have attempted to make the change, but I do not think it is correct; could you take a look at it?

Also, since `collections.namedtuple` are indexable, I have changed the applicable tests to loop over the three modes returned in `mhd_waves` by using stacked `@pytest.mark.parametrize` (as you suggest) like so:
```
@pytest.mark.parametrize("mode", range(3))
...
sample_waves[mode].angular_frequency(**kwargs_wave)
```
Utilizing indices seems like the simplest way to go about this, although I am not very familiar with using named tuples and suspect there is a better way. Any thoughts?

ejohnson-96 (approved):
👍

TrestanSimon:
I was able to reproduce the Friedrichs diagrams from Figure 25 of [Huang and Lyu 2019](https://doi.org/10.1063/1.5110991) (with the addition of the Alfvén mode here) using the methods from this PR.

@namurphy You mentioned in the original issue (#2197) maybe having a method to plot Friedrichs diagrams, so I was also wondering if you could expand on this idea.

In the figure below, the Alfvén mode is in green, the fast mode is in red, and the slow mode is in blue.

Friedrichs_diagrams_HuangLyu19

The code I wrote to reproduce these diagrams is in the attached zip file. (GitHub does not allow me to attach py files, and I felt that the code was a little too long to copy and paste here.)
[Friedrichs_diagrams_HuangLyu19.zip](https://github.com/PlasmaPy/PlasmaPy/files/12049469/Friedrichs_diagrams_HuangLyu19.zip)

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        Parameters
+        ----------
+        k : `~astropy.units.Quantity`, single valued or 1-D array
+            Wavenumber in units convertible to rad/m`.  Either single
+            valued or 1-D array of length :math:`N`.
+        theta : `~astropy.units.Quantity`, single valued or 1-D array
+            The angle of propagation of the wave with respect to the
+            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units must be
+            convertible to radians. Either single valued or 1-D array of
+            size :math:`M`.
```
namurphy, line 275:
```suggestion
        k : `~astropy.units.Quantity`
            Wavenumber in units convertible to rad/m`.  Either single
            valued or 1-D array of size :math:`N`.

        theta : `~astropy.units.Quantity`
            The angle of propagation of the wave with respect to the
            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units
            convertible to radians. Either single valued or 1-D array of
            size :math:`M`.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+
+        Returns
+        -------
+        phase_velocity : `~astropy.units.Quantity`
+            An :math:`N × M` array of computed phase velocities in units
+            m/s.
```
namurphy, line 281:
```suggestion
            of m/s.
```

plasmapy/dispersion/analytical/tests/test_mhd_wave_class.py
```diff
@@ -0,0 +1,159 @@
...
+            ({**kwargs_plasma_cold, "density": 2 * u.s}, u.UnitTypeError),
+            ({**kwargs_plasma_cold, "T": "wrong type"}, TypeError),
+            ({**kwargs_plasma_cold, "T": [1.4e6, 1.7e6] * u.K}, ValueError),
+            ({**kwargs_plasma_cold, "T": -10 * u.eV}, ValueError),
+            ({**kwargs_plasma_cold, "T": 2 * u.s}, u.UnitTypeError),
+            ({**kwargs_plasma_cold, "gamma": "wrong type"}, TypeError),
+        ],
+    )
+    def test_raises_init(self, kwargs, error):
+        """Test scenarios that raise an `Exception`."""
```
namurphy, line 47:
```suggestion
        """Test scenarios that raise an exception."""
```
...since the docstrings for tests don't show up in the docs.

plasmapy/dispersion/analytical/tests/test_mhd_wave_class.py
```diff
@@ -0,0 +1,159 @@
...
+            ({"k": 0 * u.rad / u.m, "theta": 45 * u.deg}, ValueError),
+            ({"k": -1.0 * u.rad / u.m, "theta": 45 * u.deg}, ValueError),
+            ({"k": 1e-5 * u.eV, "theta": 45 * u.deg}, astropy.units.core.UnitTypeError),
+            ({"k": 1e-5 * u.rad / u.m, "theta": np.ones((3, 2)) * u.deg}, ValueError),
+            ({"k": 1e-5 * u.rad / u.m, "theta": 5 * u.eV}, u.UnitTypeError),
+        ],
+    )
+    @pytest.mark.parametrize("mode", range(3))
+    def test_raises_angular_frequency(self, kwargs, error, mode):
+        """Test scenarios that raise an `Exception`."""
```
namurphy, line 67:
```suggestion
        """Test scenarios that raise an exception."""
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        then the ions are assumed to be singly ionized.
+    T : `~astropy.units.Quantity`, |keyword-only|, optional
+        The plasma temperature in units of K or eV, which defaults
+        to zero.
+    gamma : `float` or `int`, |keyword-only|, optional
+        The adiabatic index for the plasma, which defaults to 3/5.
+    mass_numb : `int`, |keyword-only|, optional
+        The mass number corresponding to ``ion``.
+    Z : `float` or `int`, |keyword-only|, optional
+        The charge number corresponding to ``ion``.
```
namurphy, line 331:
```suggestion
    B : `~astropy.units.Quantity`
        The magnetic field magnitude in units convertible to T.

    density : `~astropy.units.Quantity`
        Either the ion number density :math:`n_i` in units convertible
        to m\ :sup:`-3` or the total mass density :math:`ρ` in units
        convertible to kg m\ :sup:`-3`\ .

    ion : |particle-like|
        Representation of the ion species (e.g., ``'p'`` for protons,
        ``'D+'`` for deuterium, ``'He-4 +1'`` for singly ionized
        helium-4, etc.). If no charge state information is provided,
        then the ions are assumed to be singly ionized.

    T : `~astropy.units.Quantity`, |keyword-only|, optional
        The plasma temperature in units of K or eV, which defaults
        to zero.

    gamma : `float` or `int`, |keyword-only|, default: 5/3
        The adiabatic index for the plasma.

    mass_numb : `int`, |keyword-only|, optional
        The mass number corresponding to ``ion``.

    Z : `float` or `int`, |keyword-only|, optional
        The charge number corresponding to ``ion``.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        Parameters
+        ----------
+        k : `~astropy.units.Quantity`, single valued or 1-D array
+            Wavenumber in units convertible to rad/m`.  Either single
+            valued or 1-D array of length :math:`N`.
+        theta : `~astropy.units.Quantity`, single valued or 1-D array
+            The angle of propagation of the wave with respect to the
+            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units must be
+            convertible to radians. Either single valued or 1-D array of
+            size :math:`M`.
```
namurphy, line 466:
```suggestion
        k : `~astropy.units.Quantity`, single valued or 1-D array
            Wavenumber in units convertible to rad/m`. Either single
            valued or 1-D array of length :math:`N`.

        theta : `~astropy.units.Quantity`
            The angle of propagation of the wave with respect to the
            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units
            convertible to radians. Either single valued or 1-D array of
            size :math:`M`.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        then the ions are assumed to be singly ionized.
+    T : `~astropy.units.Quantity`, |keyword-only|, optional
+        The plasma temperature in units of K or eV, which defaults
+        to zero.
+    gamma : `float` or `int`, |keyword-only|, optional
+        The adiabatic index for the plasma, which defaults to 3/5.
+    mass_numb : `int`, |keyword-only|, optional
+        The mass number corresponding to ``ion``.
+    Z : `float` or `int`, |keyword-only|, optional
+        The charge number corresponding to ``ion``.
```
namurphy, line 542:
```suggestion
    B : `~astropy.units.Quantity`
        The magnetic field magnitude in units convertible to T.

    density : `~astropy.units.Quantity`
        Either the ion number density :math:`n_i` in units convertible
        to m\ :sup:`-3` or the total mass density :math:`ρ` in units
        convertible to kg m\ :sup:`-3`\ .

    ion : |particle-like|
        Representation of the ion species (e.g., ``'p'`` for protons,
        ``'D+'`` for deuterium, ``'He-4 +1'`` for singly ionized
        helium-4, etc.). If no charge state information is provided,
        then the ions are assumed to be singly ionized.

    T : `~astropy.units.Quantity`, |keyword-only|, optional
        The plasma temperature in units of K or eV, which defaults
        to zero.

    gamma : `float` or `int`, |keyword-only|, default: 5/3
        The adiabatic index for the plasma.

    mass_numb : `int`, |keyword-only|, optional
        The mass number corresponding to ``ion``.

    Z : `float` or `int`, |keyword-only|, optional
        The charge number corresponding to ``ion``.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        Parameters
+        ----------
+        k : `~astropy.units.Quantity`, single valued or 1-D array
+            Wavenumber in units convertible to rad/m`.  Either single
+            valued or 1-D array of length :math:`N`.
+        theta : `~astropy.units.Quantity`, single valued or 1-D array
+            The angle of propagation of the wave with respect to the
+            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units must be
+            convertible to radians. Either single valued or 1-D array of
+            size :math:`M`.
```
namurphy, line 609:
```suggestion
        k : `~astropy.units.Quantity`
            Wavenumber in units convertible to rad/m`.  Either single
            valued or 1-D array of length :math:`N`.

        theta : `~astropy.units.Quantity`
            The angle of propagation of the wave with respect to the
            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units
            convertible to radians. Either single valued or 1-D array of
            size :math:`M`.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+            (:math:`ω ≪ ω_c,ω_p`) assumption of the dispersion relation.
+
+        Notes
+        -----
+        The angular frequency :math:`ω` of a fast magnetosonic wave
+        is given by the equation
+
+        .. math::
+
+            ω^2 = \frac{k^2}{2} \left(c_{ms}^2 + \sqrt{c_{ms}^4 - 4 v_A^2 c_s^2 \cos^2 θ}\right)
```
namurphy, line 642:
```suggestion
            ω^2 = \frac{k^2}{2} \left(
               c_{ms}^2 + \sqrt{c_{ms}^4 - 4 v_A^2 c_s^2 \cos^2 θ}
            \right)
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        Parameters
+        ----------
+        k : `~astropy.units.Quantity`, single valued or 1-D array
+            Wavenumber in units convertible to rad/m`.  Either single
+            valued or 1-D array of length :math:`N`.
+        theta : `~astropy.units.Quantity`, single valued or 1-D array
+            The angle of propagation of the wave with respect to the
+            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units must be
+            convertible to radians. Either single valued or 1-D array of
+            size :math:`M`.
```
namurphy, line 697:
```suggestion
        k : `~astropy.units.Quantity`
            Wavenumber in units convertible to rad/m`. Either single
            valued or 1-D array of length :math:`N`.

        theta : `~astropy.units.Quantity`
            The angle of propagation of the wave with respect to the
            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units
            convertible to radians. Either single valued or 1-D array of
            size :math:`M`.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
```
namurphy, line 939:
```suggestion
        k : `~astropy.units.Quantity`
            Wavenumber in units convertible to rad/m`. Either single
            valued or 1-D array of length :math:`N`.

        theta : `~astropy.units.Quantity`
            The angle of propagation of the wave with respect to the
            magnetic field, :math:`\cos^{-1}(k_z / k)`, in units
            convertible to radians. Either single valued or 1-D array of
            size :math:`M`.
```

plasmapy/dispersion/analytical/mhd_waves_.py
```diff
@@ -0,0 +1,1078 @@
...
+        helium-4, etc.). If no charge state information is provided,
+        then the ions are assumed to be singly ionized.
+    T : `~astropy.units.Quantity`, |keyword-only|, default: 0 K
+        The plasma temperature in units of K or eV.
+    gamma : `float` or `int`, |keyword-only|, optional
+        The adiabatic index for the plasma, which defaults to 3/5.
+    mass_numb : 'int', |keyword-only|, optional
+        The mass number corresponding to ``ion``.
+    Z : `float` or 'int', |keyword-only|, optional
+        The charge number corresponding to ``ion``.
```
namurphy, line 1032:
```suggestion
    B : `~astropy.units.Quantity`
        The magnetic field magnitude in units convertible to T.

    density : `~astropy.units.Quantity`
        Either the ion number density :math:`n_i` in units convertible
        to m\ :sup:`-3` or the total mass density :math:`ρ` in units
        convertible to kg m\ :sup:`-3`\ .

    ion : |particle-like|
        Representation of the ion species (e.g., ``'p'`` for protons,
        ``'D+'`` for deuterium, ``'He-4 +1'`` for singly ionized
        helium-4, etc.). If no charge state information is provided,
        then the ions are assumed to be singly ionized.

    T : `~astropy.units.Quantity`, |keyword-only|, default: 0 K
        The plasma temperature in units of K or eV.

    gamma : `float` or `int`, |keyword-only|, default: 5/3
        The adiabatic index for the plasma.

    mass_numb : 'int', |keyword-only|, optional
        The mass number corresponding to ``ion``.

    Z : `float` or 'int', |keyword-only|, optional
        The charge number corresponding to ``ion``.
```

namurphy (approved):
Apologies for the delay in this review!  I really like how it turned out.  I made a few minor suggestions (mostly for consistency with other docstrings) which I'll accept and then merge if tests pass so we can get it out in the next release.  Thank you again so much!

PR #2213, opened 2023-06-21
Update ruff rules

namurphy:
This PR enables some more ruff rules, and applies some changes and `# noqa` comments to get the rules to pass.


namurphy:
This is ready for review. The drop in code coverage is because of `ruff` adding explicit `return None` statements that had previously been implicit.  This was only done for functions which had possible return values other than `None`.

I'm planning to create an issue about modifying these functions so that the `return None` isn't a possibility.  In the meantime, it is useful information to have these lines marked as not covered by tests.

plasmapy/particles/tests/test_decorators.py
```diff
@@ -344,10 +343,6 @@ def test_preserving_signature_with_stacked_decorators(decorator1, decorator2):
     assert undecorated_signature == decorated_signature_1_2 == decorated_signature_2_1
 
 
-@pytest.mark.xfail(
-    condition=sys.version_info < (3, 9),
```
pheuer, line 348: I assume this passes now?

pheuer (approved):
LGTM

plasmapy/particles/tests/test_decorators.py
namurphy: This was a test failure specific to Python 3.8, and we now require Python ≥3.9, so we're all set with this.

namurphy:
Thank you for the review!

PR #2222, opened 2023-06-23
Shell command correction (In contribution workflow)

mohawk811:


namurphy (approved):
Thank you for doing this!

PR #2229, opened 2023-06-27
Raise ValueError for implicit cases in `distribution`

### LINKED ISSUE #2221: Address functions in `distributions.py` with unexpected possible `return None`
While working on #2213, I noticed that there were a few functions in `distributions.py` which had an implicit `return None` that should not be possible.  We should update these functions so that an exception is raised instead of `None` being returned.

We'll need to check #2213 to see all of the functions.  (If I have time and the wherewithal, I'll make this update later.)

JaydenR2305:
Resolves #2221

### DIFF
--- a/changelog/2229.trivial.rst
+++ b/changelog/2229.trivial.rst
@@ -1,0 +1,1 @@
+Distributions defined in the `~plasmapy.distribution` module will now raise a ValueError for an improper ``units`` parameter.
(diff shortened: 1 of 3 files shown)


pheuer:
Any idea why CodeCov thinks the exception lines aren't being covered? It looks to me like you wrote tests for them...

JaydenR2305:
> Any idea why CodeCov thinks the exception lines aren't being covered? It looks to me like you wrote tests for them...

No idea...

JaydenR2305:
Fixed it!

pheuer (approved):
LGTM, thanks Jayden! (merge when ready)

PR #2230, opened 2023-06-28
Add `ParticleTracker`

JaydenR2305:
Reorganizes the `Tracker` class, which includes routines specific to proton radiography, into a specific `ChargedParticleRadiographyTracker` class and parent `ParticleTracker` class.

### DIFF
--- a/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -5,7 +5,11 @@ methods of 'inverting' experimentally created radiographs to reconstruct the
 original fields (under some set of assumptions).
 """
 
-__all__ = ["Tracker", "synthetic_radiograph"]
+__all__ = [
+    "ParticleTracker",
+    "ChargedParticleRadiographyTracker",
+    "synthetic_radiograph",
+]
 
 import astropy.constants as const
 import astropy.units as u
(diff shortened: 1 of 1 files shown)


changelog/2230.breaking.rst
JaydenR2305, line 1: I *think* breaking is the correct term here? As opposed to 'feature'

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -71,7 +75,447 @@ def _coerce_to_cartesian_si(pos):
...
+                " of particles "
+                f"({x.shape[0]} and {v.shape[0]} respectively)."
+            )
+        else:
+            self.nparticles = x.shape[0]
+
+        self.x = x.to(u.m).value
+        self.v = v.to(u.m / u.s).value
+
+        self.theta = np.arccos(
```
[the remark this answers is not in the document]
JaydenR2305: glad these made it!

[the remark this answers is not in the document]
JaydenR2305: This function was part of the `load_particle` function, which we decided was the generalized `create_particle` method. Should I create a `load_particle` for `ChargedParticleRadiography` that calculates these quantities but calls the `load_particle` inside `ParticleTracker`?

JaydenR2305:
What should be done with the `_remove_deflected_particles` method? Should we account for both tracked and untracked particles in this method? I understand we only need `self.x_tracked` for input into the particle tracker but does `_remove_deflected_particles` need both tracked and untracked particles?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,439 @@
...
+            disable=not self.verbose,
+            desc="Particles on grid",
+            unit="particles",
+            bar_format="{l_bar}{bar}{n:.1e}/{total:.1e} {unit}",  # noqa: FS003
+            file=sys.stdout,
+        )
+
+        # Push the particles until the stop condition is satisfied
+        # (no more particles on the simulation grid)
+        while True:  # not stop_condition():
```
JaydenR2305, line 410: Since we aren't defining the radiography tracer to be a subclass of the general tracer, I ran into some issues with passing arguments to the `stop_condition` method. How would you recommend I proceed?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -755,53 +677,14 @@ def load_particles(
...
-        # with the largest gridstep
-        return np.where(dt == np.inf, np.max(gridstep), dt)
+        Update the x_all and v_all quantities.
+        This function is called after each push to ensure the position and velocities
+        properties contain up-to-date quantities.
+        """
+        self.x_all = np.concatenate((self.x_not_tracked, self.Tracker.x))
+        self.v_all = np.concatenate((self.v_not_tracked, self.Tracker.v))
 
     def _coast_to_grid(self):
```
pheuer, line 689: Problem to consider if you want: is there a better way of evolving particles to the grid so they all go right to the grid, rather than just the 'first' particle?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -866,9 +752,9 @@ def _remove_deflected_particles(self):
         Removes any particles that have been deflected away from the detector
```
pheuer, line 752: Try `removing` by setting self.x = NaN?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1129,54 +902,24 @@ def run(
...
 
-        self._log(
-            "Fraction of tracked particles deflected away from the "
-            "detector plane: "
-            f"{self.fract_deflected*100}%"
-        )
+    def _push(self):
+        # entered_grid is zero at the end if a particle has never
+        # entered any grid
+        self.entered_grid += np.sum(self.Tracker.on_grid, axis=-1)
```
pheuer, line 919: Move back to particle tracker

pheuer (commented):
Looks good! Start thinking about tests for a single particle: gyro orbit/energy conservation and maybe ExB drift.

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -897,107 +771,6 @@ def _remove_deflected_particles(self):
...
-
-    @property
-    def on_any_grid(self):
-        """
-        Binary array for each particle indicating whether it is currently
-        on ANY grid.
-        """
-        return np.sum(self.on_grid, axis=-1) > 0
-
     def _stop_condition(self):
```
pheuer, line 774: Let's have this be a function (could be a static method of this class) that gets passed to ParticleTracker 

e.g. 

```python
class SyntheticRadiograh: 

@staticmethod
def _stop_condition(particle_tracker):

# Determine how many particles have entered and then left the grid 

if < threshold: 
return fract_entered # for progress bar 

else:
return True


class ParticleTracker:

def run( stop_condition=None):

# If None, use a default stopping condition function

stop=False
while not stop:
stop = _stopping_condition(self)
if stop is not True: 
nentered = stop
stop = False



```

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1047,137 +843,65 @@ def run(
...
-            )
+        self._validate_inputs(field_weighting)
 
         # If meshes have been added, apply them now
         for mesh in self.mesh_list:
             self._apply_wire_mesh(**mesh)
 
-        # Store a copy of the initial velocity distribution in memory
-        # This will be used later to calculate the maximum deflection
-        self.v_init = np.copy(self.v)
```
pheuer, line 1105: This line should stay here (only store the init arrays for the charged particle radiography case)

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -5,24 +5,25 @@
...
 from typing import Union
 
-from plasmapy import particles
 from plasmapy.formulary.mathematics import rot_a_to_b
-from plasmapy.particles import Particle
+from plasmapy.particles import Particle, particle_input
 from plasmapy.plasma.grids import AbstractGrid
-from plasmapy.simulation.particle_integrators import boris_push
+from plasmapy.simulation.particle_tracker import (
+    ParticleTracker as GeneralParticleTracker,
```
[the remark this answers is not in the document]
JaydenR2305: Should I change the radiography tracker to `RadiographyTracker` then? I do this to avoid namespace collision

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -720,23 +655,7 @@ def load_particles(
...
-                "Provided x and v arrays have inconsistent numbers "
-                " of particles "
-                f"({x.shape[0]} and {v.shape[0]} respectively)."
-            )
-        else:
-            self.nparticles = x.shape[0]
-
-        self.x = x.to(u.m).value
-        self.v = v.to(u.m / u.s).value
+        self.load_particles(x, v, particle=particle)
```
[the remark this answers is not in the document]
JaydenR2305: oops should be super()

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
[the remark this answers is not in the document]
pheuer: Oh sorry, I see what you're doing: yes this is fine to avoid the local namespace collision.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,440 @@
...
+    @property
+    def on_any_grid(self):
+        """
+        Binary array for each particle indicating whether it is currently
+        on ANY grid.
+        """
+        return np.sum(self.on_grid, axis=-1) > 0
+
+    @staticmethod
+    def _stop_condition():
```
pheuer, line 318: Assuming this is just a placeholder, but I think it needs a reasonable default stop condition option, like running to a pre-determined time.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,440 @@
...
+
+    @property
+    def vmax(self):
+        """
+        Calculate the maximum velocity.
+        Used for determining the grid crossing maximum timestep.
+        """
+        return np.max(np.linalg.norm(self.v, axis=-1))
+
+    def _validate_inputs(self, field_weighting: str):
```
pheuer, line 329: I'm not sure this really belongs in it's own function: it's only called once. If so, I would call it `_validate_inputs_run` and have it handle all the arguments for `run`.  Or `_validate_field_weighting_str` if it only does this

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,440 @@
...
+        if field_weighting in field_weightings:
+            self.field_weighting = field_weighting
+        else:
+            raise ValueError(
+                f"{field_weighting} is not a valid option for ",
+                "field_weighting. Valid choices are",
+                f"{field_weightings}",
+            )
+
+        # Check to make sure particles have already been generated
```
pheuer, line 341: Either way, this should definitely just be in run

plasmapy/simulation/particle_tracker.py
JaydenR2305: It's called first inside the function body for `run` defined in both classes. I was running into an issue with the test cases where the radiography `run` would stop some particles causing the `grid_ind` mask to not represent all of the cases not being tested. I converted the check into a function to ensure we are testing the value of field weighting at the beginning of both `run` functions for consistency. I agree, `_validate_field_weighting_str` would probably be a better name though

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -71,7 +75,447 @@ def _coerce_to_cartesian_si(pos):
...
+        # Used for determining the grid crossing maximum timestep
+        self.vmax = np.max(np.linalg.norm(self.v, axis=-1))
+
+        # Initialize a "progress bar" (really more of a meter)
+        # Setting sys.stdout lets this play nicely with regular print()
+        pbar = tqdm(
+            initial=0,
+            total=self.nparticles_grid + 1,
+            disable=not self.verbose,
+            desc="Particles on grid",
```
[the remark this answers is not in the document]
JaydenR2305: What should we do for the total in the progress bar? Currently we have `total=self.nparticles_tracked + 1,` which obviously wouldn't work for the general tracker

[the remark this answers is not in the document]
pheuer: Now I'm maybe leaning back towards making the ChargedParticleRadiography class a subclass of ParticleTracker (your idea of using NaN to separate the particles makes that simpler). Then you could just overwrite the entire run method and the stopping condition? What do you think?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: That's weird - maybe it doesn't like that it takes an instance of the class as an argument? 

What if each stopping condition was just a method (e.g. `_time_elapsed_stopping_condition`, `_particles_on_grid_stopping_condition` and then `self._stopping_condition` was an attribute that you could set equal to one of the other methods, choosing which one is used?

plasmapy/simulation/particle_tracker.py
JaydenR2305: The `test_run_options` test doesn't like it when I do the validation in the general `run` function so I had to make sure this condition is checked in both `run` functions

pheuer (commented):
Looks good! Here are some comments

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,590 @@
...
+
+class TimestepSavingRoutine(AbstractSavingRoutine):
+    """Saving routine corresponding to saving a hdf5 file every time step."""
+
+    def __init__(self, output_directory):
+        self.output_directory = output_directory
+
+    def post_push_hook(self, particle_tracker):
+        """Save an hdf5 file containing simulation positions and velocities after every push."""
+        if not isinstance(particle_tracker.dt, float):
```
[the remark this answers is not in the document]
JaydenR2305: Should we save when t=0 as well?

plasmapy/simulation/particle_tracker.py
JaydenR2305: I've moved this to the `_enforce_particle_creation` method which seems like a better spot for it

JaydenR2305:
This pull request is being split in two, with a request for the `ParticleTracker` in the `simulation` module and another for the tracker in the `charged_particle_radiography` module.

PR #2231, opened 2023-06-30
Add "decorators" section to coding guide

### LINKED ISSUE #2223: Add minimal example for `@validate_quantities` in coding guide
Right now the coding guide includes an example of using `@validate_quantities`, which includes additional checks for different parameters.  It would be helpful to include a minimal example which does not have any of those additional checks.  It could be something like:

```python
@validate_quantities
def magnetic_pressure(B: u.T) -> u.Pa:
    return B**2 / (2 * const.mu0)
```

This might also be a good time to create a specific subsection for "Validating quantities" so that it can be linked to, instead of including it in a bullet point list.  Should be placed near `@particle_input` too.  Maybe a "Decorators" section with subsections for "Validating quantities" and "Particle inputs"?

JaydenR2305:
Resolves #2223

Adds a decorators section demonstrating use of the `quantity_input` and `particle_input` decorators.


docs/contributing/coding_guide.rst
```diff
@@ -488,6 +488,44 @@ Requirements
   year. However, it may take a few months before packages like |NumPy|
   and |Numba| become compatible with the newest minor version of |Python|.
 
+Decorators
+==========
+
+.. _particle_inputs:
+
+Particle inputs
+---------------
```
namurphy, line 497:
```suggestion
Transforming particle-like arguments into particle objects
----------------------------------------------------------
```

docs/contributing/coding_guide.rst
```diff
@@ -488,6 +488,44 @@ Requirements
...
+
+* Use |particle_input| to coerce objects to relevant |Particle| type
+  (see :ref:`particles`)
+
+.. _validating_quantities:
+
+Validating quantities
+---------------------
```
namurphy, line 505:
```suggestion
Validating Quantity arguments
-----------------------------
```

namurphy (approved):
Thank you again for doing this, and apologies for the delay!  I have a few minor suggestions, and then we can get this merged prior to our forthcoming release.

docs/contributing/coding_guide.rst
```diff
@@ -488,6 +488,44 @@ Requirements
...
+         ...
+
+* Use |validate_quantities| to enforce |Quantity| type hints
+
+  .. code-block:: python
+
+   @validate_quantities
+   def magnetic_pressure(B: u.T) -> u.Pa:
+       return B**2 / (2 * const.mu0)
+
```
namurphy, line 528:
```suggestion
Use |validate_quantities| to verify function arguments and impose
relevant restrictions

.. code-block:: python

   from plasmapy.utils.decorators.validators import validate_quantities

   @validate_quantities(
       n={"can_be_negative": False},
       validations_on_return={"equivalencies": u.dimensionless_angles()},
   )
   def inertial_length(n: u.m**-3, particle) -> u.m:
       ...

Use |validate_quantities| to enforce |Quantity| type hints

.. code-block:: python

   @validate_quantities
   def magnetic_pressure(B: u.T) -> u.Pa:
       return B**2 / (2 * const.mu0)

```

PR #2239, opened 2023-07-04
Appease Ruff in `nullpoint`

JaydenR2305:
Resolves PERF401 error raised in #2238 by using list comprehensions in the `nullpoint` module

### DIFF
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -712,9 +712,14 @@ def _trilinear_analysis(vspace, cell):  # noqa: C901, PLR0911, PLR0912, PLR0915
         dy + gy * yConst1,
         fy + hy * yConst1,
     )
-    for root in root_list:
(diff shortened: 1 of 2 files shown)


JaydenR2305:
@namurphy Should I add a changelog entry for this?

namurphy:
> @namurphy Should I add a changelog entry for this?

I think it'd be fine either way!  I'll add the "no changelog entry needed" label, but will leave it up to you whether you want to add one or not.

JaydenR2305:
According to the coverage report we are missing 37 lines that were previously being tested. Maybe I made a typo somewhere?

namurphy:
I think I've noticed at times that preliminary coverage results sometimes get posted before tests finish.  Not sure why.  It looks like the coverage results are showing that everything is being covered now, so this looks set to merge.  Thank you for doing this!

namurphy (approved):

PR #2240, opened 2023-07-06
Add ruff rule that prohibits implicit `typing.Optional`

namurphy:
A few years ago, a convention with type hints allowed us to skip using `typing.Optional` when the default value was `None`.  That convention is now not recommended.  This PR applies ruff rule `RUF012` which changes things like `f(x: int = None)` to `f(x: Optional[int] = None)`.

Feel free to merge if this looks good!


JaydenR2305 (approved):
LGTM!

PR #2245, opened 2023-07-12
Add `ParticleTracker`

JaydenR2305:
Implements a general `ParticleTracker` class in the `simulation` module. This class is intended to eventually be the parent class of the `ParticleTracker` in the `charged_particle_radiography` module.


plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,715 @@
...
+        # self.grid is the grid object
+        if isinstance(grids, AbstractGrid):
+            self.grids = [
+                grids,
+            ]
+        elif isinstance(grids, collections.abc.Iterable):
+            self.grids = grids
+        elif isinstance(grids, BasePlasma):
+            raise TypeError(
+                "This class has been deprecated. Please revert to PlasmaPy version 2023.5.1 to use this version of ParticleTracker."
```
pheuer, line 222: Maybe "It appears you may be trying to access an older version of the ParticleTracker class. This class has been deprecated..."

plasmapy/simulation/tests/test_particle_tracker.py
```diff
@@ -0,0 +1,53 @@
...
+
+        grid.add_quantities(B_z=Bz)
+
+        simulation = ParticleTracker(
+            grid,
+            req_quantities=["E_x", "E_y", "E_z", "B_x", "B_y", "B_z"],
+            verbose=False,
+        )
+
+        x = np.array([[1, 0, 0], [2, 0, 0], [3, 0, 0]]) * u.m
```
pheuer, line 37: Great test! To avoid hardcoding these numbers, you could define v as v_x, determine the corresponding gyroradius r_L using the plasmapy function (for this particle + bfield + perp velocity) R_L, then set the initial position as (0, R_L, 0). That way the particle should orbit around the origin without hardcoding these values.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,759 @@
...
+
+    @property
+    def get_description(self):
+        """The time elapsed stop condition depends on elapsed time,
+        therefore the relevant quantity is time remaining.
+        """
+        return "Time remaining"
+
+    @property
+    def get_units(self):
```
pheuer, line 95: I would return an astropy object here, `u.s`

JaydenR2305:
@pheuer What would you suggest for other tests on the tracker? I could probably do something with ExB drift and measure the distance from the guiding center right?

pheuer:
> @pheuer What would you suggest for other tests on the tracker? I could probably do something with ExB drift and measure the distance from the guiding center, right?

Yeah, that would be a good one. For the gyro-orbit one you already have, you should check that the kinetic energy  before and after hasn't changed. 

Another nice one would be a particle that goes across a constant E-field region and making sure the energy it gains is what you expect.

pheuer:
@JaydenR2305 For now, the gyroradius test is probably a good start, maybe with the ExB one added if you want (and testing each with a few different charges + masses). 

The stopping and saving conditions will all also need tests. Look at the proton radiography film RCF pack tests for an example of how to save to a temporary directory during tests (for the save to disk tests).

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,833 @@
...
+    def is_finished(self):
+        """The simulation is finished when no more particles are on any grids."""
+        is_not_on_grid = self._particle_tracker.on_grid == 0
+
+        return is_not_on_grid.all() and self._particle_tracker.iteration_number > 0
+
+    @property
+    def progress(self):
+        """The progress of the simulation is measured by how many particles are no longer on a grid."""
+        is_not_on_grid = self._particle_tracker.on_grid == 0
```
pheuer, line 163: Is `on_grid` on the grid NOW, or on the grid EVER? For this application I the relevant quantity is how many of the particles have ever been on the grid, but are now off. Otherwise it will terminate on the first step when no particles have yet entered the grid. To terminate, we want: 

1) Many particles (say >99%) have entered the grid at some point. 
2) All the particles are now off the grid.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,833 @@
...
+
+        if self._particle_tracker.time - self.time_of_last_save >= self.save_interval:
+            self.time_of_last_save = self._particle_tracker.time
+
+            return True
+        else:
+            return False
+
+
+class AbstractDiskSaveRoutine(AbstractSaveRoutine, ABC):
```
pheuer, line 267: Is there a reason to not just have  `save_to_memory` and `save_to_disk` methods on the abstract save class, then chose which one to call based on a switch variable? I'm thinking that the way you have it now, you'll have to implement two classes for each type of save condtion (e.g. a save to memory class for interval saving, then another one for saving to disk with the same `save_now` function?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,833 @@
...
+
+        self.v[particles_to_stop_mask] = np.NaN
+
+    def _remove_particles(self, particles_to_remove_mask):
+        if len(particles_to_remove_mask) != self.x.shape[0]:
+            raise ValueError(
+                f"Expected mask of size {self.x.shape[0]}, got {len(particles_to_remove_mask)}"
+            )
+
+        self.x[particles_to_remove_mask] = np.NaN
```
pheuer, line 476: Can you add notes to the docstring for this class defining what the NaNs mean (for velocity and position). If both are NaN it means the particle was removed: what if it is stopped?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,833 @@
...
+    ):
+        if not isinstance(stop_condition, AbstractStopCondition):
+            raise TypeError("Please specify a valid stop condition.")
+
+        if not isinstance(save_routine, AbstractSaveRoutine):
+            raise TypeError("Please specify a valid save routine")
+
+        # Will the simulation follow a uniform time step?
+        # This must be the case for certain stopping conditions and saving routines
+        self.is_uniform_time = isinstance(dt, u.Quantity) and isinstance(
```
pheuer, line 667: There are three cases to consider here: 

1) Adaptive timestep (nonuniform) computed separately for every particle , dt=[nparticles,]
2) Adaptive timestep (nonuniform)  but all particles use the same one (the min. of those computed for all particles). This needs to be added as a keyword option somehow, and you should be able to use save functions on this? 
3) Uniformly spaced timesteps (behavior you get if you just set dt to a fixed value).

plasmapy/simulation/tests/test_particle_tracker.py
```diff
@@ -0,0 +1,137 @@
...
+        grid_shape = (num,) * 3
+
+        Bz = np.full(grid_shape, B_strength) * u.T
+        grid.add_quantities(B_z=Bz)
+
+        return ParticleTracker(
+            grid, req_quantities=["E_x", "E_y", "E_z", "B_x", "B_y", "B_z"]
+        )
+
+    @pytest.mark.slow()
```
pheuer, line 39: Just curious: is this actually slow? I'd hope it would be pretty quick for just a couple particles. You only need ~50-100 iterations/gyroperiod and ~10 gyroperiods, so you shouldn't need more than 1000 iterations.

plasmapy/simulation/tests/test_particle_tracker.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Set the initial position to the gyroradius
+        # This means the particle will orbit the origin
+        R_L = gyroradius(B_strength, self.point_particle, Vperp=v_x)
+        x = np.array([[0, R_L_element.value, 0] for R_L_element in R_L]) * u.m
+
+        simulation = self.instantiate_simulation(B_strength=B_strength)
+        simulation.load_particles(x, v, self.point_particle)
+
+        stop_condition = TimeElapsedStopCondition(10 * u.s)
```
pheuer, line 55: Makes sure the stop condition and dt are something like my comment above so you aren't under or over-resolving

plasmapy/simulation/tests/test_particle_tracker.py
```diff
@@ -0,0 +1,137 @@
...
+
+        positions = np.asarray(save_routine.x_all) * u.m
+        distances = np.linalg.norm(positions, axis=-1)
+
+        assert np.isclose(distances, R_L, rtol=1e-2).all()
+
+    @pytest.mark.slow()
+    @given(arrays(int, (5,), elements=st.integers(1, 100)))
+    @settings(deadline=1e5, max_examples=3)
+    def test_kinetic_energy(self, v_x):
```
pheuer, line 68: I would run the simulation once (maybe in the instantiate method) then just run this test on the same output as the previous test. No need to run it again.

pheuer (commented):
Really good progress! I really like these class stopping and saving routines - that divides things really nicely!

plasmapy/simulation/particle_tracker.py
JaydenR2305: I get `TypeError: Unsupported operand type(s) for ufunc add: 'str,Quantity'`

plasmapy/simulation/particle_tracker.py
JaydenR2305: I think the issue is then how to deal with the output path. Before, the abstract stopping condition class took an argument for the output path, which doesn't apply when saving to memory. I transferred the output directory component of the code to its own class for the sake of being explicit that the output path parameter is only relevant to disk saving routines.

It would be a good idea to implement the `save_to_memory` and `save_to_disk` methods in the abstract class. My only concern would be that the `save_to_disk` method would have to make reference to some sort of `output_path` parameter, which I hesitate to include in the abstract constructor. Maybe this isn't a problem as long as the documentation is clear?

pheuer: Ok, yes I see. In this case I think it's better to have 'output_path' be an optional keyword to the abstract class, then just note in the documentation that it only needs to be specified if saving to disk. `output_dir` might be better, since you should probably save each timestep to a different h5 file. You can add a quick check that `self.output_diri is not None` to the save to disk method to verify it was set.

plasmapy/simulation/tests/test_particle_tracker.py
JaydenR2305: It seems to do fine with a step of dt=1e-2, the resulting relative differences are usually ~5e-2, so these are the parameters I'll use for the assertion. Good catch!

plasmapy/simulation/particle_tracker.py
JaydenR2305: I use this stopping condition in tests related to `test_particle_tracker_potential_difference` to stop the simulation when the particles have reached the end of the grid, and thus no longer require interpolation to determine their simulated final energies. Would you recommend I change the stopping condition to instead focus on particles that have ever entered a grid and have since exited for sake of generality?

plasmapy/simulation/particle_tracker.py
JaydenR2305: I have been using "uniform" to mean that the simulation uses the same time step for all of the particles. Is this term accurate? Or should I switch to something a little less ambiguous.. 

Either way, I fail to account for the case of a uniform adaptive time step, something that should definitely be implemented.

pheuer: I think it's accurate but potential ambiguous, since it could be understood in either way. Maybe the best way to handle this is just by raising appropriate errors if the timestep (stored as dt in the Tracker object) doesn't match the requirements of the save condition?

plasmapy/simulation/particle_tracker.py
pheuer: Got it - no that's fine. I would maybe add the stopping condition for the proton radiography class now though, just in case implementing it requires any changes to this code? Otherwise we'll have to come back and modify these files again when refactoring that.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,835 @@
...
+        # Create flags for tracking when particles during the simulation
+        # on_grid -> zero if the particle is off grid, 1
+        # shape [nparticles, ngrids]
+        self.on_grid = np.zeros([self.nparticles, self.num_grids])
+
+        # Entered grid -> non-zero if particle EVER entered a grid
+        self.entered_grid = np.zeros([self.nparticles])
+
+        # Update the `particle_tracker` attribute so that the stop condition & save routine can be used
+        stop_condition.particle_tracker = self
```
pheuer, line 771: Here you are setting the `particle_tracker` attribute manually, but at least the save routine has a setter method. I think maybe it's best to add a setter method to the stop condition as well, then use those here? That will make it easier to document that this has to be done.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,835 @@
...
+            what fields are felt by the particles. Options are:
+
+            * 'nearest neighbor': Particles are assigned the fields on
+                the grid vertex closest to them.
+            * 'volume averaged' : The fields experienced by a particle are a
+                volume-average of the eight grid points surrounding them.
+
+            The default is 'volume averaged'.
+
+        stop_condition : callable
```
pheuer, line 736: `function` -> `class`

Add a docstring entry for the save_routine

Maybe specify that this class is passed an instance of the particle tracer, and uses that to determine when to stop/save.

plasmapy/simulation/particle_tracker.py
JaydenR2305: Should I add the radiography stopping condition to this pull request or wait until we create a PR for the radiography component of the simulation?? I was able to write a proof of concept:

```Python

class RadiographyStoppingCondition(AbstractStopCondition):
    def require_uniform_dt(self):
        """Radiography simulations do not require a uniform time step."""
        return True

    @property
    def progress_description(self):
        """The progress of a synthetic radiography simulation is given by the number of particles still being pushed
        through the grid."""
        return "Particles on grid"

    @property
    def units(self):
        """The progress of a radiography simulation is tracked in terms of number of particles."""
        return "particles"

    @property
    def is_finished(self):
        """
        The stop condition is that most of the particles have entered the grid
        and almost all have now left it.
        """
        # Count the number of particles who have entered, which is the
        # number of non-zero entries in entered_grid
        num_entered = np.nonzero(self.tracker.entered_grid)[0].size

        # How many of the particles have entered the grid
        fract_entered = np.sum(num_entered) / self.tracker.nparticles_tracked

        # Of the particles that have entered the grid, how many are currently
        # on the grid?
        # if/else avoids dividing by zero
        if num_entered.sum() > 0:
            # Normalize to the number that have entered a grid
            still_on = np.sum(self.tracker.on_any_grid) / np.sum(num_entered)
        else:
            still_on = 0.0

        if fract_entered <= 0.1 or still_on >= 0.001:
            return False

        # Warn user if < 10% of the particles ended up on the grid
        if num_entered < 0.1 * self.tracker.nparticles_tracked:
            warnings.warn(
                f"Only {100 * num_entered / self.tracker.nparticles_tracked:.2f}% of "
                "particles entered the field grid: consider "
                "decreasing the max_theta to increase this "
                "number.",
                RuntimeWarning,
            )

        return True

    @property
    @abstractmethod
    def progress(self):
        """Return the number of particles still being pushed through the grid."""
        return self.tracker.on_any_grid.sum()

    @property
    @abstractmethod
    def total(self):
        """Return the total number of particles created."""
        return self.tracker.nparticles

```

plasmapy/simulation/particle_tracker.py
JaydenR2305: Is this what you were looking for?

https://github.com/PlasmaPy/PlasmaPy/blob/61630328929ae6e8ac09bec9c0517e9f4ed07033/plasmapy/simulation/particle_tracker.py#L681

An error is raised if we require a uniform (over all particles) time step and one is not explicitly set. The adaptive time step is used otherwise

plasmapy/simulation/tests/test_particle_tracker.py
```diff
@@ -0,0 +1,153 @@
...
+@pytest.mark.parametrize(
+    ("stop_condition", "save_routine", "is_disk_routine"),
+    [
+        (NoParticlesOnGridsStoppingCondition(), MemoryIntervalSaveRoutine, False),
+        (TimeElapsedStopCondition(10 * u.s), MemoryIntervalSaveRoutine, False),
+        (NoParticlesOnGridsStoppingCondition(), DiskIntervalSaveRoutine, True),
+        (TimeElapsedStopCondition(10 * u.s), DiskIntervalSaveRoutine, True),
+    ],
+)
+def test_interval_save_routine(tmp_path, stop_condition, save_routine, is_disk_routine):
```
JaydenR2305, line 34: This isn't an ideal way to do this.. Would you have any recommendations @pheuer ? The problem is that I need a way of accessing `tmp_path` in order to instantiate a disk save routine, meaning the instantiation has to happen inside of the test function (unless there is another way of doing this?)

JaydenR2305: Maybe something like `construct_save_routine_class` would clarify things (which you can pass `tmp_path` into)? Then we would just have to use a dictionary to retrieve the appropriate arguments

plasmapy/simulation/particle_tracker.py
pheuer: That draft looks good! Now you mention it, it probably makes sense to wait and do this in the PR that refactors the radiography code (since that would be the natural place to test this). So, put this code somewhere you won't loose it and pull it out then!

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+
+    def __init__(self, interval: u.Quantity):
+        super().__init__(interval)
+
+    def save(self):
+        """Save the state of the particle tracker to memory."""
+        self.save_to_memory()
+
+
+class ParticleTracker:
```
pheuer, line 317: This whole class needs a docstring.

Can you add notes to the docstring for this class defining what the NaNs mean (for velocity and position). If both are NaN it means the particle was removed: what if it is stopped?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+
+        return is_not_on_grid.sum()
+
+    @property
+    def total(self):
+        """The progress of the simulation is measured against the total number of particles in the simulation."""
+        return self._particle_tracker.nparticles
+
+
+class AbstractSaveRoutine(ABC):
```
pheuer, line 172: Looking at the `adaptive_dt`, it occurs to me that we require a saving condition be provided, but what if the user doesn't care to save the data? We need a `NoSaveRoutine` that just does nothing when the hook is called. That would be a natural keyword default if the user doesn't specify a save routine.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+
+        return np.max(np.linalg.norm(self.v[tracked_mask], axis=-1))
+
+    def _validate_run_inputs(
+        self, stop_condition, save_routine, dt, field_weighting: str
+    ):
+        if not isinstance(stop_condition, AbstractStopCondition):
+            raise TypeError("Please specify a valid stop condition.")
+
+        if not isinstance(save_routine, AbstractSaveRoutine):
```
pheuer, line 667: Here is where we need the 'NullSaveRoutine' or something

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+            raise TypeError("Please specify a valid save routine")
+
+        require_uniform_time = (
+            stop_condition.require_uniform_dt or save_routine.require_uniform_dt
+        )
+
+        # Will the simulation follow a uniform time step?
```
pheuer, line 674: `self.is_uniform_time` should probably be a class property

plasmapy/simulation/particle_tracker.py
pheuer: Yes, I think this is good. Maybe to solve the ambiguity we can introduce two class properties 

`self.is_adaptive_timestep` -> This property, says whether or not an adaptive timestep is being used. I think this is a little more clear? 

`self.is_syncronized_timesteps` -> Timesteps are the same for each particle, or different for each particle. 

At least right now, if `self.is_adaptive_timestep=False`, then `self.is_syncronized_timesteps=False`. But if you are using the adaptive timestep, you could either be applying one timestep to all particles (`self.is_syncronized_timesteps=True`) or you could be calculating a different one for each particle (`self.is_syncronized_timesteps=False`) which is what the radiography tracer does right now by default (since the particles are not interacting). 

Making these class properties will make it easier to document clearly what each means in the docstring.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+
+    @property
+    def nparticles_tracked(self):
+        """Return the number of particles that don't have NaN position or velocity."""
+        return self._tracked_particle_mask().sum()
+
+    def run(
+        self,
+        stop_condition: AbstractStopCondition,
+        save_routine: Optional[AbstractSaveRoutine] = None,
```
pheuer, line 711: Here `save_routine` needs a default of class `AbstractSaveRoutine` for the case where the user doesn't want to save the data (i.e. they only care about the final state of the particles).

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+        """The time elapsed stop condition depends on elapsed time,
+        therefore the relevant quantity is time remaining.
+        """
+        return "Time remaining"
+
+    @property
+    def units(self):
+        """The units for the time elapsed condition have the units of seconds."""
+
+        return "seconds"
```
pheuer, line 113: In PlasmaPy we should use astropy units to represent units for consistency, so this should be `u.s`. 

Then, when you initiate the progress bar, you can convert that to a string 

```python
pbar = tqdm(
            initial=0,
            total=stop_condition.total,
            disable=not self.verbose,
            desc=stop_condition.progress_description,
            unit=str(stop_condition.units),
            bar_format="{l_bar}{bar}{n:.1e}/{total:.1e} {unit}",  # noqa: FS003
            file=sys.stdout,
        )
```

That will end up printing `s` instead of `seconds`, but I think it's better to do that and remain consistent with the expectation that astropy. If you really wanted to have it print seconds, you could do something like 

```python
if len(stop_condition.units.long_names)>0:
    unit_name = stop_condition.units.long_names[0] + 's'
else:
    unit_name = str(stop_condition.units)
```

The `long_names` list contains any fully spelled out unit names associated with a unit object (not plural)

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,842 @@
...
+
+        Parameters
+        ----------
+        stop_condition : `~plasmapy.simulation.particle_tracker.AbstractStopCondition`
+            A class responsible for determining when the simulation has reached
+            a suitable termination condition. The class is passed an instance
+            of `~plasmapy.simulation.particle_tracker.ParticleTracker`. See `~plasmapy.simulation.particle_tracker.AbstractStopCondition`
+            for more details.
+
+        save_routine: `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine`
```
pheuer, line 729: I would word the description here like this

An instance of `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine` which determines which timesteps of the simulation to save. 

That makes it clear that what is being passed in is an instance of that class, not a reference to the class itself.

plasmapy/simulation/tests/test_particle_tracker.py
pheuer: Try pytest fixture
https://docs.pytest.org/en/6.2.x/fixture.html

pheuer:
@namurphy 

@JaydenR2305 is nearing completion on this PR. We're still working on docstrings, some additional tests etc. but if you wanted to glance over the general structure in the next week or so that would be great! 

Also tagging @StanczakDominik since this would replace the existing ParticleTracker (as we previously discussed).

JaydenR2305:
@pheuer I am finishing the final edits for the new time step properties. Is there a reason why self.dt is automatically interpreted to be the limits for a potential time step? It is becoming problematic in the case where we require a synchronized time step but the user does not specify a specific value, meaning an adaptive time step is calculated.

pheuer:
> @pheuer I am finishing the final edits for the new time step properties. Is there a reason why `self.dt` is automatically interpreted to be the limits for a potential time step? It is becoming problematic in the case where we require a synchronized time step but the user does not specify a specific value, meaning an adaptive time step is calculated.
> 
> https://github.com/PlasmaPy/PlasmaPy/blob/d38c099c7bff235c9cb05921226cae385a863a3d/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py#L797

I agree using a single keyword for the range (for an adaptive keyword) or to set a fixed timestep is somewhat confusing - not sure if having two keywords `dt` and `dt_range` would be better?  Maybe it would be good though to have a separate keyword `synchronized_dt`. If True, take the minimum adaptive dt and use for all particles, if False, use a separate `dt` for each particle. Would that solve the problem you're talking about? 

Note that if the user doesn't provide a dt, it's set to a range spanning all possible values 

```python

 # By default, set dt as an infinite range (auto dt with no restrictions)
        self.dt = np.array([0.0, np.inf]) * u.s if dt is None else dt
        self.dt = (self.dt).to(u.s).value

```

JaydenR2305:
> 
> Note that if the user doesn't provide a dt, it's set to a range spanning all possible values
> 
> ```python
>  # By default, set dt as an infinite range (auto dt with no restrictions)
>         self.dt = np.array([0.0, np.inf]) * u.s if dt is None else dt
>         self.dt = (self.dt).to(u.s).value
> ```

This actually creates another headache, since the interval save routine requires that the time step does not exceed the save interval... Should there be some sort of property that handles this in the save routine?

pheuer:
> This actually creates another headache, since the interval save routine requires that the time step does not exceed the save interval... Should there be some sort of property that handles this in the save routine?

This is kind of an edge case: it would really only come up if the user tried to save way too many points for the given scenario, since the adaptive timestep by design picks a physically reasonable timestep. 

The interval save routine doesn't currently guarantee that you save at exactly the specified interval, it just saves at the requested interval +/- one timestep.  Doing otherwise would involve a complicated step where you sometimes choose the timestep to make sure you exactly land on the requested interval, which is too complicated for now I think. 

I would suggest we just drop the error in the save routine, and in this pathological case just let it save every timestep if the save interval is smaller than the timestep?

JaydenR2305:
>
> This is kind of an edge case: it would really only come up if the user tried to save way too many points for the given scenario, since the adaptive timestep by design picks a physically reasonable timestep.
> 
Yeah that makes sense.

> You could include a warning if the interval not much larger than the timestep (e.g. if not timestep <= interval/10 or something similar).

This would correspond to excessive saving right? (saves / timestep <= 10)

plasmapy/simulation/particle_tracker.py
JaydenR2305: By using `unit_string` am I able to still return "seconds"??

plasmapy/simulation/particle_tracker.py
JaydenR2305: I don't think we're using the stopping or removing mechanisms yet. Should I remove them from this pull request and put them somewhere for a later PR?

JaydenR2305:
image
The time has come...

I'll work on rewriting this notebook so we can get the doc tests passing

pheuer:
> This would correspond to excessive saving, right (saves / timestep >= 10)? Or is this about par for these simulations

Yes, exactly. 

> I'll work on rewriting this notebook so we can get the doc tests passing

Great! Feel free to modify parts of it as you see fit, especially to demonstrate the specific functionality of the new tracker.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,886 @@
...
+        self._has_run = False
+
+        self._is_adaptive_time_step = False
+        self._is_synchronized_time_step = False
+
+        # *********************************************************************
+        # Validate required fields
+        # *********************************************************************
+
+        if req_quantities is None:
```
JaydenR2305, line 361: How should we handle required quantities?? Should it be the baseline that 

```python
req_quantities = ["E_x", "E_y", "E_z", "B_x", "B_y", "B_z"]
```

and the `req_quantities` keyword can be used to specify additional quantities?

pheuer: Good question - ultimately this depends on what other options are being used. If you are using the current pusher, you need E&B at a minimum. Once we have scattering + stopping, a density quantity will be required (along with maybe some scalar quantities like the mean ionization energy as we discussed). 

Maybe we should actually move this validation into a separate helper function, then call that in the `run` method so it use all the run settings and pick the required settings appropriately?

plasmapy/simulation/particle_tracker.py
pheuer: Yep sure, I think that can be whatever string representation you want.

plasmapy/simulation/particle_tracker.py
```diff
@@ -285,9 +285,31 @@ def save_now(self):
 
 
 class ParticleTracker:
-    """
-    General particle tracer.
+    r"""A many body particle tracker for electric and magnetic fields with no particle-particle interactions.
+
+    Particles are created and pushed through calculated E and B fields. These fields are specified as part of a grid
```
pheuer, line 290: created -> instantiated 

Calculated -> "a grid of provided E&B fields" We want to differentiate from trackers that calculate the fields from specified charge locations, or from the particles themselves.

plasmapy/simulation/particle_tracker.py
```diff
@@ -285,9 +285,31 @@ def save_now(self):
...
+    which can then be interpolated to determine local field strength. The time step used can be specified, otherwise an
+    adaptive time step is calculated based off grid resolution and the particle's gyroradius.
+
+    The simulation is terminated when a specified stop condition is met. The results of a simulation can be exported
```
pheuer, line 294: Stopping conditions are specified as instances of the ___ class as arguments to the `run` method. 

Save routines are specified as instances of the __ class as arguments to the `run` method.

plasmapy/simulation/particle_tracker.py
```diff
@@ -285,9 +285,31 @@ def save_now(self):
...
+    using a save routine.
 
+    Parameters
+    ----------
+    grids : `~plasmapy.plasma.grids.AbstractGrid` or subclass thereof, or list
+        of same.
+        A Grid object or list of grid objects containing the required
```
pheuer, line 301: An instance of `link to grid class`

For req_quantities I think you can just discuss all of this under the `req_quantities` keyword below

plasmapy/simulation/particle_tracker.py
```diff
@@ -725,15 +755,17 @@ def run(
...
             time steps of the simulation to save. See `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine` for more details.
 
         dt : `~astropy.units.Quantity`, optional
             An explicitly set time step in units convertible to seconds.
             Setting this optional keyword overrules the adaptive time step
-            capability and forces the use of this time step throughout. If a tuple
-            of time steps is provided, the adaptive time step will be clamped
+            capability and forces the use of this time step throughout.
+
+        dt_range : tuple of shape `(2,)` of `~astropy.units.Quantity`, optional
```
pheuer, line 774: Looks good - what happens if the user tries to set both? That should raise an exception in the _validate_run_inputs method.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,883 @@
...
+        grids: Union[AbstractGrid, Iterable[AbstractGrid]],
+        req_quantities=None,
+        verbose=True,
+    ):
+        # self.grid is the grid object
+        if isinstance(grids, AbstractGrid):
+            self.grids = [
+                grids,
+            ]
+        elif isinstance(grids, collections.abc.Iterable):
```
pheuer, line 327: These elif and else statements aren't currently being hit in testing (according to codecov)

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,883 @@
...
+
+    def _stop_particles(self, particles_to_stop_mask):
+        if len(particles_to_stop_mask) != self.x.shape[0]:
+            raise ValueError(
+                f"Expected mask of size {self.x.shape[0]}, got {len(particles_to_stop_mask)}"
+            )
+
+        self.v[particles_to_stop_mask] = np.NaN
+
+    def _remove_particles(self, particles_to_remove_mask):
```
pheuer, line 479: Write just one test each for _remove_particles and _stop_particles that shows they work as intended (to satisfy codecov)

pheuer:
@JaydenR2305 When you get a chance, take a look at the codecov/patch coverage: there's a number of sections here that each just need a simple test to make sure those lines get hit. Everything should get hit in testing unless there's a good reason why it can't really be tested.

plasmapy/simulation/particle_tracker.py
JaydenR2305: Should I keep these methods in this PR?? They aren't being used yet

pheuer: I think that's ok - since they're pretty basic functionality for a particle tracer (and we're just going to add them in the immediate next PR anyway)

plasmapy/simulation/particle_tracker.py
pheuer: Let's discuss this in our next meeting - I think it's a deeper philosophical question about what quantities go at run time (describe the run?) vs at instantiation (describe the grid?).  Either way, you cannot know just from the instantiation step what quantities are required (since that depends on the run conditions) so we would just end up moving lots of stuff to the instantiation which makes less sense to me than having it at the run method? But let's discuss.

pheuer: I've been thinking more about this - maybe the answer is to just move the specification of the termination + saving classes (and eventually whatever flags are provided to turn on stopping + scattering) to init, so we can validate the grid there?

pheuer:
@JaydenR2305 Suddenly occurred to me that maybe we should call the class that determines when the simulation ends a 'TerminationCondition' rather than "stopping" to avoid future confusion when we add the particle stopping functionality?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,896 @@
...
+        if not self._is_adaptive_time_step:
+            return self.dt
+
+        # candidate time steps includes one per grid (based on the grid resolution)
+        # plus additional candidates based on the field at each particle
+        candidates = np.ones([self.nparticles, self.num_grids + 1]) * np.inf
+
+        # Compute the time step indicated by the grid resolution
+        ds = np.array([grid.grid_resolution.to(u.m).value for grid in self.grids])
+        gridstep = 0.5 * (ds / self.vmax)
```
pheuer, line 517: Courant (sp?) number

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,898 @@
...
+    ):
+        # self.grid is the grid object
+        if isinstance(grids, AbstractGrid):
+            self.grids = [
+                grids,
+            ]
+        elif isinstance(grids, collections.abc.Iterable):
+            self.grids = grids
+        elif isinstance(grids, BasePlasma):
+            raise TypeError(
```
JaydenR2305, line 373: Since we move the specification of the termination condition and the saving routine to the constructor this way of detecting old code no longer works and instead throws a `ValueError` because no termination condition is specified.. What do you recommend we do with this? @pheuer

JaydenR2305:
Oh perfect, thank you! I am not too familiar with creating notebooks

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        Sets the time of last save attribute and log the timestamp.
+        """
+        super().save()
+
+        self.time_of_last_save = self.tracker.time
+        self.t_all.append(self.tracker.time)
+
+
+class ParticleTracker:
+    r"""A particle tracker for particles in electric and magnetic fields with inter-particle interactions.
```
pheuer, line 286: withOUT inter-particle interactions

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+    adaptive time step is calculated based off grid resolution and the particle's gyroradius.
+
+    Termination conditions for the simulation are provided as an instance of the `~plasmapy.simulation.particle_tracker.AbstractTerminationCondition` class as arguments to the ``run`` method.
+    The results of a simulation can be exported by specifying an instance of the `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine` class to the ``run`` method.
+
+    Parameters
+    ----------
+    grids : An instance of `~plasmapy.plasma.grids.AbstractGrid`
+        A Grid object or list of grid objects containing the required
+        quantities.
```
pheuer, line 310: "The list of required quantities varies depending on other keywords" ?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+
+    termination_condition : `~plasmapy.simulation.particle_tracker.AbstractTerminationCondition`
+        A class responsible for determining when the simulation has reached
+        a suitable termination condition. The class is passed an instance
```
pheuer, line 314: The sentence about passing in an instance of ParticleTracker might be confusing - I'd leave that for the docstring for the TerminationCondition class.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+    The results of a simulation can be exported by specifying an instance of the `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine` class to the ``run`` method.
+
+    Parameters
+    ----------
+    grids : An instance of `~plasmapy.plasma.grids.AbstractGrid`
+        A Grid object or list of grid objects containing the required
+        quantities.
+
+    termination_condition : `~plasmapy.simulation.particle_tracker.AbstractTerminationCondition`
+        A class responsible for determining when the simulation has reached
```
pheuer, line 313: I prefer the way you phrased this for the save routine... "An instance of ___ which determines ___"

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+    dt : `~astropy.units.Quantity`, optional
+        An explicitly set time step in units convertible to seconds.
+        Setting this optional keyword overrules the adaptive time step
+        capability and forces the use of this time step throughout.
+
+    dt_range : tuple of shape (2,) of `~astropy.units.Quantity`, optional
+        If specified, the calculated adaptive time step will be clamped
+        between the first and second values.
+
+        field_weighting : str
```
pheuer, line 331: Formatting - extra indent

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        # self.grid_arr is the grid positions in si units. This is created here
+        # so that it isn't continuously called later
+        self.grids_arr = [grid.grid.to(u.m).value for grid in self.grids]
+
+        self.verbose = verbose
+
+        # This flag records whether the simulation has been run
+        self._has_run = False
+
+        self._is_adaptive_time_step = False
```
pheuer, line 390: It seems like these should be determined by some logic, not just set to False? Presumably down where you have self.dt and self.dt_range being set

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        self._is_synchronized_time_step = False
+
+        # By default, the gyration of a particle is divided into twelve steps
+        self._steps_per_gyroperiod = 12
```
pheuer, line 394: It would be better to call `setup_adaptive_time_step()` here, so the default values aren't repeated in two places in the code. Otherwise, if you change one later, it could lead to weird behavior.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+
+        self._Courant_parameter = 0.5
+
+        # *********************************************************************
+        # Validate required fields
+        # *********************************************************************
+
+        base_req_quantities = ["E_x", "E_y", "E_z", "B_x", "B_y", "B_z"]
```
pheuer, line 402: Thinking about how we should handle this for cases where we might add more quantities to this list in response to other keywords. I'd suggest 

1) Make this list a private attribute of the class, define right at the beginning of init 
2) If other keywords are turned on (e.g. stopping soon) we can then append to this list
3) Then have this validate structure at the very end of init. To avoid init being a long method, maybe break this into a separate method `_validate_grids()`

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        self,
+        time_steps_per_gyroperiod: Optional[int] = 12,
+        Courant_parameter: Optional[float] = 0.5,
+    ):
+        """Set parameters for the adaptive time step candidates.
+
+        Parameters
+        ----------
+        time_steps_per_gyroperiod : int, optional
+            The number of subdivisions of the particle's gyration. The default is twelve.
```
pheuer, line 488: The minimum ratio of the particle gyroperiod to the timestep. Higher numbers correspond to higher temporal resolution.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+
+        Courant_parameter : float, optional
+            The Courant parameter is the fraction of a grid cell length by which to subdivide
```
pheuer, line 491: ...is the the minimum ratio of the timestep to  the grid crossing time, grid cell length / particle velocity. Lower Courant numbers correspond to higher temporal resolution.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        A Grid object or list of grid objects containing the required
+        quantities.
+
+    termination_condition : `~plasmapy.simulation.particle_tracker.AbstractTerminationCondition`
+        A class responsible for determining when the simulation has reached
+        a suitable termination condition. The class is passed an instance
+        of `~plasmapy.simulation.particle_tracker.ParticleTracker`. See `~plasmapy.simulation.particle_tracker.AbstractTerminationCondition`
+        for more details.
+
+    save_routine : `~plasmapy.simulation.particle_tracker.AbstractSaveRoutine`
```
pheuer, line 318: For each of these, add a sentence saying what the default value is, and what does.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        Notes
+        -----
+        Two candidates are calculated for the adaptive time step: a time step based on the gyroradius
+        of the particle and a time step based on the resolution of the grid. The candidate associated
+        with the gyroradius of the particle takes a ``time_steps_per_gyroperiod`` parameter that specifies
+        how many times the orbit of a gyrating particles will be subdivided. The other candidate,
+        associated with the spatial resolution of the grid object, calculates a time step using the time
+        it would take the fastest particle to cross some fraction of a grid cell length. This fraction is the Courant number.
+        """
+
```
pheuer, line 504: Should calling this method raise an exception if `self.dt` is not None, in other words if a fixed timestep is already set?

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+            Positions for N particles
+
+        v : `~astropy.units.Quantity`, shape (N,3)
+            Velocities for N particles
+
+        particle : |particle-like|, optional
+            Representation of the particle species as either a |Particle| object
+            or a string representation. The default particle is protons.
+
+        distribution : str
```
pheuer, line 539: Delete 'distribution` parameter from docstring, that's from the proton radiography class

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+                " of particles "
+                f"({x.shape[0]} and {v.shape[0]} respectively)."
+            )
+        else:
+            self.nparticles = x.shape[0]
+
+        self.x = x.to(u.m).value
+        self.v = v.to(u.m / u.s).value
+
+    def _stop_particles(self, particles_to_stop_mask):
```
pheuer, line 574: These methods need docstring - they can be a bit shorter (since they are private) but should still describe what the method does and define the parameters.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        """
+        Binary array for each particle indicating whether it is currently
+        on ANY grid.
+        """
+        return np.sum(self.on_grid, axis=-1) > 0
+
+    @property
+    def vmax(self):
+        """
+        Calculate the maximum velocity.
```
pheuer, line 770: This first sentence needs to be a more complete description, because it shows up by itself in the docs in the parameter table:

https://plasmapy--2245.org.readthedocs.build/en/2245/api/plasmapy.simulation.particle_tracker.ParticleTracker.html#plasmapy.simulation.particle_tracker.ParticleTracker

How about "The maximum velocity of any particle in the simulation"

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        Used for determining the grid crossing maximum time step.
+        """
+        tracked_mask = self._tracked_particle_mask()
+
+        return np.max(np.linalg.norm(self.v[tracked_mask], axis=-1))
+
+    def _validate_constructor_inputs(
```
pheuer, line 777: Just for sake of organization for future coders, I think this method should come right after the constructor.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+            raise TypeError("Please specify a valid save routine")
+
+        require_synchronized_time = termination_condition.require_synchronized_dt or (
+            save_routine is not None and save_routine.require_synchronized_dt
+        )
+
+        if dt is not None and dt_range is not None:
+            raise ValueError("Please only specify either dt or dt_range")
+
+        # Will the simulation follow a synchronized time step?
```
pheuer, line 800: This is good - you don't need to set these attributes to `False` in the constructor then, just let this method do it.

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+        # This must be the case for certain termination conditions and saving routines
+        if isinstance(dt, u.Quantity):
+            if isinstance(dt.value, np.ndarray):
```
pheuer, line 803: This logic looks like it needs to be updated to use the new `dt_range` variable? 

An exception should be raised here if you try to set `dt` AND `dt_range`

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,945 @@
...
+            self._is_adaptive_time_step = True
+
+        # Raise a ValueError if a synchronized dt is required by termination condition or save routine but one is not
+        # given. This is only the case if an array with differing entries is specified for dt
+        if require_synchronized_time and not self._is_synchronized_time_step:
+            raise ValueError(
+                "Please specify a synchronized time step to use the simulation with this configuration!"
+            )
+
+        # Load and validate inputs
```
pheuer, line 820: Nice!

pheuer:
Tests look good! Very comprehensive.

plasmapy/simulation/particle_tracker.py
JaydenR2305: I should have caught that. My bad

plasmapy/simulation/particle_tracker.py
JaydenR2305: Yeah the logic for this is currently in `_validate_constructor_inputs` because they're used to determine whether an error should be raised based on the provided save routine and termination condition. Should I move them to the constructor body for the sake of organization?

plasmapy/simulation/particle_tracker.py
JaydenR2305: I think this is what we're doing with the `not self.is_adaptive_time_step` condition?

JaydenR2305:
Okay a quick status update (doubles as a note to self when I get back)-- Ruff is complaining about the constructor being too complex in terms of the number of if statements, which makes sense. I will likely move the logic that involves setting attributes related to the time step into its own function to reduce complexity. This is especially the case with `_is_synchronous_time_step`, since the value is dependent on the save routine and stopping condition (if an adaptive time step is used, then the synchronicity of the time step is dependent on whether it is required by the saving and termination conditions.)

pheuer:
> Ruff is complaining about the constructor being too complex in terms of the number of if statements, which makes sense. I will likely move the logic that involves setting attributes related to the time step into its own function to reduce complexity. 

I agree - that will make the code more readable, which is the goal of the Ruff warning. Moving the grid validation into a separate `_validate_grids` would also move a lot of code (more loops than if statements) out of the constructor.

plasmapy/simulation/particle_tracker.py
pheuer: I wouldn't have them both places - having them be created in `validate...` is a bit confusing since I wouldn't expect validation to create new variables. Maybe a new small method `_determine_timestep_type` or something that just does this, then you can call that in the constructor after you validate the input? 

It seems like a lot of methods, but if the names are good that can make the code self-documenting and make it easier to find relevant parts of the code.

plasmapy/simulation/particle_tracker.py
pheuer: Ah, I see, yes that makes sense.

plasmapy/simulation/particle_tracker.py
JaydenR2305: (use set)

pheuer: And it should be a set! Good call

plasmapy/simulation/particle_tracker.py
```diff
@@ -351,7 +373,7 @@ def __init__(
         self,
         grids: Union[AbstractGrid, Iterable[AbstractGrid]],
         termination_condition: AbstractTerminationCondition = None,
-        save_routine: Optional[AbstractSaveRoutine] = None,
+        save_routine: AbstractSaveRoutine = None,
```
pheuer, line 376: You could also just set the default here, instead of in the if statement below? 

`save_routine: Optional[AbstractSaveRoutine] = DoNotSaveSaveRoutine()`

plasmapy/simulation/particle_tracker.py
```diff
@@ -455,16 +424,47 @@ def __init__(
...
         self.save_routine = save_routine
 
-    @property
-    def num_grids(self):
-        """The number of grids specified at instantiation."""
-        return len(self.grids)
+    def _set_time_step_attributes(self, dt, termination_condition, save_routine):
+        """Determines whether the simulation will follow a synchronized or adaptive time step."""
+
+        if not isinstance(termination_condition, AbstractTerminationCondition):
```
pheuer, line 435: These are input validation - I think they should go in the constructor or in another validator, not in this `set` method

plasmapy/simulation/particle_tracker.py
```diff
@@ -494,14 +494,99 @@ def setup_adaptive_time_step(
...
+                                arr[:, :, -1],
+                            )
+                        ]
+                    )
+                )
+
+                if edge_max > 1e-3 * np.max(arr):
+                    unit = grid.recognized_quantities[rq].unit
+                    warnings.warn(
+                        "Fields should go to zero at edges of grid to avoid "
```
pheuer, line 577: Let's change "Fields" -> "Quantities" throughout this message, since we'll soon be adding non-field quantities like mass density

plasmapy/simulation/particle_tracker.py
```diff
@@ -547,6 +632,11 @@ def load_particles(
         self.v = v.to(u.m / u.s).value
 
     def _stop_particles(self, particles_to_stop_mask):
+        """Stop tracking the particles specified by the stop mask.
+
+        This is accomplished by setting the particle's velocity to NaN.
```
pheuer, line 637: "This is represented by setting..." 

Just to be clear that this is a convention we are using.

plasmapy/simulation/particle_tracker.py
```diff
@@ -555,6 +645,12 @@ def _stop_particles(self, particles_to_stop_mask):
         self.v[particles_to_stop_mask] = np.NaN
 
     def _remove_particles(self, particles_to_remove_mask):
+        """Remove the specified particles from the simulation.
+
+        For the same of keeping consistent array lengths, the position and velocities of the
+        particles are set to NaN.
```
pheuer, line 651: "of the removed particles" 

Does it say anywhere explicitly that 

- x,v != NaN -> Tracked particle
- x != NaN, v=NaN -> Stopped particle (still in the simulation, but position is no longer evolved). 
- x,v = NaN -> Removed particle

docs/notebooks/simulation/particle_stepper.ipynb
```diff
@@ -4,10 +4,7 @@
    "cell_type": "code",
```
pheuer, line 4: <p>Is it still in need of a rewrite for speed? Maybe! But I think we can take this out</p><p> </p><p>Also, stepper -&gt; Tracker everywhere in the notebook, for consistency </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/2245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

pheuer, line 4: <p>Can you make this <code>grid</code> a link to the CartesianGrid object in the docs?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/2245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

pheuer, line 4: <p>"ParticleTracker takes arrays of particle positions and velocities of the shape [nparticles, 3], so these arrays represent one particle starting at the origin." </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/2245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

pheuer, line 4: <p>Hm, this isn't the nicest way of getting out the times and positions from the save method. Maybe we should have a <code>results</code>  method or something on save routine that returns this stuff without the user having to know the names of variables like <code>t_all</code> ?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaPy/PlasmaPy/pull/2245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,962 @@
...
+        self.verbose = verbose
+
+        # This flag records whether the simulation has been run
+        self._has_run = False
+
+        # Instantiate the do not save save routine if no save routine was specified
+        if save_routine is None:
+            save_routine = DoNotSaveSaveRoutine()
+
+        self._set_time_step_attributes(dt, termination_condition, save_routine)
```
pheuer, line 408: Name - also sets termination condition

plasmapy/simulation/particle_tracker.py
```diff
@@ -0,0 +1,962 @@
...
+        if field_weighting in field_weightings:
+            self.field_weighting = field_weighting
+        else:
+            raise ValueError(
+                f"{field_weighting} is not a valid option for ",
+                "field_weighting. Valid choices are",
+                f"{field_weightings}",
+            )
+
+    def _validate_grids(self, additional_required_quantities):
```
pheuer, line 528: "preprocess" grids

docs/notebooks/simulation/particle_stepper.ipynb
pheuer: <p>t, x, v = save_routine.result </p><p> </p><p>do list -&gt; array conversion in that method </p>

pheuer:
@namurphy This PR is getting close to done - let me know if you want to do your own review, otherwise I will probably merge it in a week or two once we are done with the few reorganization and documentation updates we're still working on.

docs/changelog/0.6.0.rst
```diff
@@ -83,7 +83,7 @@ Features
   `plasmapy.dispersion`, which gives an analytical solution to the dispersion relation as
   derived by :cite:t:`bellan:2012`\ . (`#932 <https://github.com/plasmapy/plasmapy/pull/932>`__)
 - Refactor out the `~plasmapy.simulation.particle_integrators.boris_push` tracking
-  integrator algorithm from `~plasmapy.simulation.particletracker.ParticleTracker`. (`#953 <https://github.com/plasmapy/plasmapy/pull/953>`__)
+  integrator algorithm from ``~plasmapy.simulation.particletracker.ParticleTracker``. (`#953 <https://github.com/plasmapy/plasmapy/pull/953>`__)
```
namurphy, line 86:
```suggestion
  integrator algorithm from ``plasmapy.simulation.particletracker.ParticleTracker``. (`#953 <https://github.com/plasmapy/plasmapy/pull/953>`__)
```

namurphy:
@JaydenR2305 & @pheuer — thank you for working on this! I'm excited to see this so close.  

I'll try to look at it this week, but if I don't get around to it, @pheuer should feel free to merge it.  

One minor cosmetic thing is that we should keep docstrings to ≲ 72 characters to be in line with PEP 8.  (There really ought to be an autoformatter for numpydoc-style docstrings, but I haven't found one that fits well enough for us yet.)

Thank you again!

pheuer:
Thanks @namurphy ! You mean the first line of the doc string should be <72 right - the short one sentence description?

namurphy:
> Thanks @namurphy ! You mean the first line of the doc string should be <72 right - the short one sentence description? Or the length of each line?

The length of each line should be ≲ 72 characters so that it shows up better when we use `help()`.  This is a soft limit so it's fine to go past that if there's a reason to, like a long reStructuredText construct or something.  Thanks!

JaydenR2305:
pre-commit.ci autofix

JaydenR2305:
> One minor cosmetic thing is that we should keep docstrings to ≲ 72 characters to be in line with PEP 8. (There really ought to be

Are there any specific classes/methods that are not following this?? I tried to follow this for the docstrings I add here but sometimes get closer to ~80 characters just because I don't want to cut off sentences in weird spots

StanczakDominik:
Hey guys, I'm still alive :) I'll also give this a look through, but don't feel the need to hold up the merge, I can suggest or apply fixes post merge.

pheuer:
> Are there any specific classes/methods that are not following this?? I tried to follow this for the docstrings I add here but sometimes get closer to ~80 characters just because I don't want to cut off sentences in weird spots

I think some of the main docstring for ParticleTracker is breaking it? For those docstrings that are just barely over (e.g. 80 characters) the thing to do is break the line earlier, so the first line is say ~60 characters and the rest goes on the next line. The exception being the first line of the docstring, which needs to be descriptive and strictly less than 72 characters. 


> Hey guys, I'm still alive :) I'll also give this a look through, but don't feel the need to hold up the merge, I can suggest or apply fixes post merge.

Great to hear you're alive @StanczakDominik ! Let us know if you have any suggestions :)

pheuer:
@JaydenR2305 You should merge main into this branch to resolve the branch conflict -due to #2437 and #2439 you may then get some new CI errors due to type hinting we need to address before merging this. Basically, just adding type hints for all the methods, even with trivial returns.  #2437 shows how you can do this using a tool called autotyping if you don't want to do it manually.

JaydenR2305:
@pheuer Do you know how I would go about fixing the conflict? I can't resolve it in github desktop as I'm usually able to. Re-adding the file using the contents on the main branch got rid of the conflict, but deleting it again caused the conflict to reappear..

pheuer:
> @pheuer Do you know how I would go about fixing the conflict? I can't resolve it in github desktop as I'm usually able to. Re-adding the file using the contents on the main branch got rid of the conflict, but deleting it again caused the conflict to reappear..

I think I fixed this (not sure how) - the issue is that this file got renamed from `particletracker` to `particle_tracker`? Was that intentional? My only concern is that it will making the breaking change worse, because `particletrakcer.Tracker` won't resolve and a user wouldn't even get to the depreciation error you added for the new tracker?

namurphy:
@pheuer — I think the mypy errors not being ignored is happening because the filenames were changed: `particletracker.py` → `particle_tracker.py` and `test_particletracker.py` to `test_particle_tracker.py`.  They should go away if the names are updated in `mypy.ini`.

pheuer:
> @pheuer — I think the mypy errors not being ignored is happening because the filenames were changed: `particletracker.py` → `particle_tracker.py` and `test_particletracker.py` to `test_particle_tracker.py`. They should go away if the names are updated in `mypy.ini`.

That would probably be it! 

@JaydenR2305 Can you update that? For some reason my pre-commit hook isn't working right now (getting a 'ruby executable not found' error?)

JaydenR2305:
> @JaydenR2305 Can you update that? For some reason my pre-commit hook isn't working right now (getting a 'ruby executable not found' error?)

Will do! I got that as well, you just have to install ruby

JaydenR2305:
image

JaydenR2305:
pre-commit.ci autofix

JaydenR2305:
Does anyone know why the particle input decorator is raising that error? It doesn't like when I use ParticleLike either.

namurphy:
> Does anyone know why the particle input decorator is raising that error? It doesn't like when I use ParticleLike either.

That is very strange. It looks like it's in `load_particles`, though it appears to be as expected. I also haven't figured out how to reproduce the bug in a simpler test case yet, which would make debugging easier.  I'll try to take a look at it tomorrow.  Thank you for finding and reporting this!

JaydenR2305:
Found it! It didn't like that I was using future annotations. It was convenient because I could avoid having to use ``"ParticleTracker"`` as a type before the class was defined. It does however change the formatting of the annotations dunder which broke the functionality of the decorator.

namurphy:
> Found it! It didn't like that I was using future annotations.

Nicely done!  Would you be willing to [submit a bug report](https://github.com/PlasmaPy/PlasmaPy/issues/new?assignees=&labels=Bug&projects=&template=bug_report.yml) to indicate that `@particle_input` doesn't work with `from __future__ import annotations`?

pheuer:
@namurphy There are a few mypy errors here that I'm just ignoring because I don't see how to deal with them: 

"""
plasmapy/simulation/particle_tracker.py:739: error: Unsupported operand types for / ("float" and "None")  [operator]
            candidates[:, self.num_grids] = gyroperiod / self._steps_per_gyroperiod
                                            ^
"""
`self._steps_per_gyroperiod` can be None, but it is never None when this code block is reached. 


"""
plasmapy/simulation/particle_tracker.py:865: error: Value of type "Union[ndarray[Any, dtype[floating[Any]]], float]" is not indexable  [index]
                self.time[tracked_mask] += dt
                ^~~~~~~~~~~~~~~~~~~~~~~
"""
Same issue here - `self.time` can be either a float or an array of floats depending on keywords - this block is only reached if it is an array, but mypy seems worried about it still. 

"""
plasmapy/simulation/particle_tracker.py:938: error: Attribute "on_grid" already defined on line 781  [no-redef]
            self.on_grid: NDArray[np.bool_] = np.zeros(
"""
I think this was happening because a reassignment of this variable happened in a method that was higher up in the file than the method where the variable is initialized?? 


"""
plasmapy/simulation/particle_tracker.py:795: error: No overload variant of "__rmul__" of "ndarray" matches argument type "None"  [operator]
            gridstep = self._Courant_parameter * (ds / self.vmax)
"""
Same issue here - this line won't be reached when the variable is None

namurphy:
> @namurphy There are a few mypy errors here that I'm just ignoring because I don't see how to deal with them

For the time being, I think adding the error codes to the ignore lists in `mypy.ini` would be perfectly reasonable.  Thank you!

pheuer (approved):
After systematically ignoring most of the mypy errors, I declare victory! @JaydenR2305, I'm going to merge this PR so you can move on to the next steps in the project: refactoring the proton radiography code :)

PR #2246, opened 2023-07-14
Update author check error message

namurphy:
This PR amends the author check error message with some contextual information about how some of the fields are needed for abstract submission. It also adds the `email` field to the example entry for `CITATION.cff`.


JaydenR2305 (approved):
LGTM!

PR #2256, opened 2023-07-26
Cythonify lite functions

namurphy:
This is a prototype for converting the lite-functions from Numba to Cython.


namurphy:
Thinking more about this, it might be better to keep all of the Cython stuff packaged outside of core.  I'm remembering back to https://github.com/astropy/astropy/issues/7660, for example.

PR #2264, opened 2023-08-01
Apply `ruff` rule RUF100 to remove unused `noqa` directives

namurphy:
This PR adds a `ruff` rule that goes through code and removes the `# noqa` comments that are no longer needed.  There are no actual substantive changes.


pyproject.toml
```diff
@@ -258,6 +258,7 @@ extend-select = [
   "RUF010", # explicit-f-string-type-conversion
   "RUF011", # static-key-dict-comprehension
   "RUF013", # implicit-optional
+  "RUF100", # unused-qa
```
namurphy, line 261:
```suggestion
  "RUF100", # unused-noqa
```

PR #2267, opened 2023-08-03
fix typo in test for ExcessStatistics class

gregordecristoforo:
## Description

While working on the next PR for the time_series sub-module I discovered that I didn't update the docstring of a test for the ExcessStatistics class. Originally I implemented the code as a function and forgot to update the docstring when I turned it into a class.

## Related issues

None


namurphy (approved):
Thank you for the fix!

PR #2275, opened 2023-08-07
Add class for calculating conditional average of time series.

### LINKED ISSUE #2125: Conditional averaging for time series
### Feature description

For a given time series and given threshold values, find the conditional average and conditional variance.

### Motivation

This functionality has been developed and applied for time series analysis of intermittent fluctuations in the SOL region of fusion experiments. Some examples are listed below:
https://doi.org/10.1016/j.nme.2016.11.008 (figure 11, 12)
https://doi.org/10.1088/0741-3335/58/4/044006 (figure 10)
https://doi.org/10.1088/1741-4326/aa7e4c (figure 4)

### Implementation strategy

Implementing this feature should be straight forward since the method is already developed and available in: https://github.com/uit-cosmo/fpp-analysis-tools/blob/main/fppanalysis/conditional_averaging.py

The code still needs refactoring and appropriate tests and documentation needs to be added. I am planning to take a look at this in the upcoming days (could I be assigned to this issue?).

### Additional context

_No response_

gregordecristoforo:
## Description

This PR contains a class for calculating the conditional average and variance of events in a time series. These events are defined as a peak above a given threshold. Optionally, an upper threshold can be used. The class also calculated the peaks, arrival times and waiting times of the events.

## Motivation and context

This functionality is widely used for time series analysis of intermittent fluctuations such as fluctuations in the SOL region of fusion experiments. Some examples are listed below:
https://doi.org/10.1016/j.nme.2016.11.008 (figure 11, 12)
https://doi.org/10.1088/0741-3335/58/4/044006 (figure 10)
https://doi.org/10.1088/1741-4326/aa7e4c (figure 4)

## Related issues

Closes #2125


plasmapy/analysis/time_series/conditional_averaging.py
```diff
@@ -0,0 +1,358 @@
...
+
+    # This astropy unit checks are quite ugly in my view.
+    # If a code reviewer has a better idea how to handle this I would be very grateful.
+    def _check_units_consistency(self, signal, lower_threshold, upper_threshold):
+        if isinstance(signal, u.Quantity):
+            if not isinstance(lower_threshold, u.Quantity):
+                raise TypeError(
+                    "signal/reference_signal and lower_threshold must have same astropy unit"
+                )
+            if signal.unit != lower_threshold.unit:
```
namurphy, line 288: One thing we can do with `Quantity` objects is compare their physical types.  That would be something like `signal.unit.physical_type != lower_threshold.unit.physical_type`, which would be `True` if the units were km and cm, for example.  In this case, the code later on requires them to be in the same units.  

Another possibility would be to use a `try`/`except` block:
```Python
try:
    lower_threshold.to(signal.unit)
except u.UnitConversionError:
    raise u.UnitConversionError("<error message>")
```

namurphy (commented):
Thank you for submitting this!  It would be great to include this in our `v2023.9.0` release which will be next month.  I'll be on vacation for the next week but will hopefully be able to read through this again soon after I get back.

Since you asked about how to handle `astropy.units`, I'm curious about the reason for separating the units and values so that the operations are done using NumPy arrays.  I'm wondering about this because `astropy.units.Quantity` objects behave very similarly to NumPy arrays, and usually the same code works for both.  Most of the functions in NumPy and SciPy are compatible with `astropy.units` too.  Sticking with units would also let us use the [`@validate_quantities`](https://docs.plasmapy.org/en/stable/api/plasmapy.utils.decorators.validators.validate_quantities.html) decorator which would handle some of the validations of inputs.  

Thank you again, it's hugely appreciated!

gregordecristoforo:
> Since you asked about how to handle astropy.units, I'm curious about the reason for separating the units and values so that the operations are done using NumPy arrays. I'm wondering about this because astropy.units.Quantity objects behave very similarly to NumPy arrays, and usually the same code works for both. Most of the functions in NumPy and SciPy are compatible with astropy.units too. Sticking with units would also let us use the [@validate_quantities](https://docs.plasmapy.org/en/stable/api/plasmapy.utils.decorators.validators.validate_quantities.html) decorator which would handle some of the validations of inputs.

The reason for separating the units and values lies in the scipy.signal.find_peaks function:
When using a sequence with a unit the returned values don't show any unit anymore:
```python
>>> find_peaks([1,2,1,2,1]*u.m, height = 1.5)
(array([1, 3]), {'peak_heights': array([2., 2.])})
```
When using a threshold with a unit we get the following error:
```python
>>> find_peaks([1,2,1,2,1]*u.m, height = 1.5*u.m)
...
ValueError: array size of lower interval border must match x
```
If there is a more elegant workaround I would be happy to know.

plasmapy/analysis/time_series/conditional_averaging.py
```diff
@@ -0,0 +1,358 @@
...
+        signal,
+        time,
+        lower_threshold,
+        upper_threshold=None,
+        reference_signal=None,
+        length_of_return=None,
+        distance=0,
+    ):
+        # This astropy unit checks are quite ugly in my view.
+        # If a code reviewer has a better idea how to handle this I would be very grateful.
```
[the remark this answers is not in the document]
gregordecristoforo: The current implementation does all of the calculations in the constructor of the class. A setter would therefore not change anything, right? Do you think it makes sense to implement a specific method that does the calculations instead?

plasmapy/analysis/time_series/conditional_averaging.py
gregordecristoforo: I adopted the `signal.unit.physical_type != lower_threshold.unit.physical_type` approach. This is better than my first implementation since it would have thrown an error for `u.cm` and `u.m`.

gregordecristoforo:
The current implementation doesn't allow for the `time` parameter to have an astropy unit. This is also the case for the `ExcessStatistics` class (see issue #2285). Should we take care of this in this PR? Would it make sense to forbid non-time units for `time`?

namurphy:
> The reason for separating the units and values lies in the scipy.signal.find_peaks function: When using a sequence with a unit the returned values don't show any unit anymore:

Thank you, that is very helpful to know!  There is some machinery within `astropy.units` for handling functions like those from `scipy.special`.  If I remember correctly it shouldn't be *that* hard to make functions from `scipy.signal` work with `astropy.units`, at least for the next release of Astropy.  I'll try to look into that soon.

In the meantime, one possibility would be to temporarily create a wrapper function that would  use `scipy.signal.find_peaks` on the values, and then return the results with the units re-attached.

plasmapy/analysis/time_series/conditional_averaging.py
```diff
@@ -0,0 +1,402 @@
...
+    `UnitsError`:
+        If astropy units of ``signal``/``reference_signal`` and ``lower_threshold`` do not match.
+        If astropy units of ``signal``/``reference_signal`` and ``upper_threshold`` do not match.
+        If astropy units of ``time``, ``length_of_return`` and ``distance`` do not match.
+
+    Notes
+    -----
+    A detailed analysis of the conditional averaging method is presented in
+    Rolf Nilsen's master thesis: "Conditional averaging of overlapping pulses"
+    https://munin.uit.no/handle/10037/29416
```
ejohnson-96, line 57: A small section overviewing the basics of conditional averaging would be instrumental here, also could you change the URL to be a linked reference?

plasmapy/analysis/time_series/conditional_averaging.py
```diff
@@ -0,0 +1,402 @@
...
+        distance,
+        signal,
+        time,
+        reference_signal,
+        length_of_return,
+        upper_threshold,
+        lower_threshold,
+    ):
+        if distance < 0:
+            raise ValueError("distance can't be negative")
```
ejohnson-96, line 286: Capitalise the start of the error messages please

ejohnson-96 (commented):
Looks very good, few minor changes 👍

plasmapy/analysis/time_series/conditional_averaging.py
gregordecristoforo: `distance` is a parameter starting with a lower case d, should it still be capitalized in the error message?

gregordecristoforo:
I rephrased the error messages so that they all start with a capitalized letter without changing the parameter names. Let me know if anything else is left to do.

plasmapy/analysis/time_series/conditional_averaging.py
namurphy: Good point!  In that case, perhaps we could put distance in single quotes?

namurphy (approved):
LGTM!  Thank you again for doing this, and my apologies for the delay in merging it.

PR #2276, opened 2023-08-07
Update pre-commit configuration

namurphy:
Still a work in progress.  I'll add more detail here in a bit...


JaydenR2305 (approved):
LGTM! Good call on changing the autoupdate frequency to monthly so we don't have to fix everything that broke in the new ruff version as frequently 👍

namurphy:
> LGTM! Good call on changing the autoupdate frequency to monthly so we don't have to fix everything that broke in the new ruff version as frequently +1

Awesome, thank you for the quick review!

PR #2282, opened 2023-08-08
Make imports from `astropy.units` and `astropy.constants` more consistent

namurphy:
This PR enables the [`pre-commit-search-and-replace`](https://github.com/mattlqx/pre-commit-search-and-replace) hook.

This hook makes these replacements:

 - `from astropy import units as u` → `import astropy.units as u`
 - `from astropy import constants as const` → `import astropy.constants as const`

I also added some common typos for this.

This PR looks big, but the main things to check are `.pre-commit-config.yaml` and `.pre-commit-search-and-replace.yaml`


.pre-commit-search-and-replace.yaml
```diff
@@ -0,0 +1,6 @@
+- search: from astropy import units as u
+  replacement: import astropy.units as u
+- search: from astropy import constants as const
+  replacement: import astropy.constants as const
+- search: Plasmapy|PlasmPy|Plasmpy|Palsmapy
+  replacement: PlasmaPy
```
namurphy, line 6: ...to catch common typos that have snuck in before.

.pre-commit-config.yaml
```diff
@@ -42,6 +42,12 @@ repos:
   - id: absolufy-imports
     exclude: docs/plasmapy_sphinx
 
+- repo: https://github.com/mattlqx/pre-commit-search-and-replace
+  rev: v1.0.5
+  hooks:
+  - id: search-and-replace
+    exclude: changelog/.*|docs/changelog/.*
```
namurphy, line 49: This uses regular expressions, where the `.*` means any character, any number of times; and the `|` means "or".

.pre-commit-search-and-replace.yaml
ejohnson-96: Good Idea 👍

ejohnson-96 (approved):
👍

PR #2284, opened 2023-08-09
Fix and expand pre-commit-search-and-replace configuration

namurphy:
I had set up one of the patterns to replace incorrectly, so this PR fixes that.

I'm also adding a few more to automagically change to use some substitutions.

I had to exclude `.pre-commit-search-and-replace.yaml` from the yaml autoformatter in pre-commit because it kept removing quotes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
     # be put in quotes. However, the pretty-format-yaml hook will
     # remove the quotes which will break that action, so we should not
     # run this hook on `labeler.yml`.
-    exclude: .github/labeler.yml|.pre-commit-search-and-replace.yaml|docs/common_links.rst
+    exclude: .github/labeler.yml|.pre-commit-search-and-replace.yaml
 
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
     # be put in quotes. However, the pretty-format-yaml hook will
     # remove the quotes which will break that action, so we should not
     # run this hook on `labeler.yml`.
```
namurphy, line 31:
```suggestion
    # run this hook on certain files.
```

.pre-commit-config.yaml
```diff
@@ -28,8 +28,8 @@ repos:
     # For the labeler GitHub Action, labels with spaces in them must
     # be put in quotes. However, the pretty-format-yaml hook will
     # remove the quotes which will break that action, so we should not
-    # run this hook on `labeler.yml`.
-    exclude: .github/labeler.yml
+    # run this hook on certain files.
```
namurphy, line 31:
```suggestion
    # run this hook on `labeler.yml` (or certain other files).
```

PR #2294, opened 2023-08-21
Use pre-commit-search-and-replace to change LaTeX commands into Unicode characters.

### LINKED ISSUE #2293: Use `pre-commit-search-and-replace` to replace LaTeX-style characters with Unicode characters?
We recently added a pre-commit hook that searches files for text patterns (or more general regular expressions) and replaces them (i.e. `Plasmapy` → `PlasmaPy`).  One possibility for using this would be to have search-and-replace transformations like `\pm` → `±` and `\alpha` → `α` defined in `.pre-commit-search-and-replace.yaml`.

We'd have to create a regular expression (maybe with ChatGPT's help?) that would help us avoid changes that we wouldn't want to make (i.e. not changing things like `\\\\alpha` or things like `\alphaetc`).   

The advantages are:
 - It would automate converting LaTeX style code to Unicode characters
 - Equations that show up in docstrings would be more readable since we wouldn't have to parse LaTeX code.

Potential disadvantages are:
 - Slightly more complexity for contributors
 - `.pre-commit-search-and-replace.yaml` would become a lot longer
 - Unicode characters might possibly show up as garbled in a terminal (though I don't yet know of any modern terminals that can't handle the more common UTF-8 characters)

namurphy:
This PR tries out the pre-commit-search-and-replace tool to see if it could reasonably be used to convert things like `\alpha` to `α`.  It's a prototype so far.  

See #2293 for details.  Closes #2293.


namurphy:
It's pretty tricky getting the regular expression to match only the LaTeX character commands that I want to match, since there are a bunch of false positives.  

If we decide to proceed, we could...

 - Skip LaTeX characters with lots of false positives like `\alpha` in the hook
 - Do this as a once-over for the repo.

PR #2296, opened 2023-08-22
Enable `ruff` linter rules that check for `print` and `pprint`

namurphy:
This PR enables two rules for `ruff` that check for the usage of `print` [T201](https://beta.ruff.rs/docs/rules/print/) and ``pprint`` [T202](https://beta.ruff.rs/docs/rules/pprint/), which should should be avoided in most cases in production code.  Instead, the `logging` library should be used in most situations.


plasmapy/particles/ionization_state.py
```diff
@@ -912,4 +912,5 @@ def summarize(self, minimum_ionic_fraction: Real = 0.01) -> None:
             output += attributes
 
         for line in output:
-            print(line)
+            # We'll need to switch from print() to using logging library
```
namurphy, line 915:
```suggestion
```
I *think* that since we're intentionally printing this to the console, that `print` should be fine here.  If any issues arise, we can change it.

plasmapy/particles/ionization_state_collection.py
```diff
@@ -987,4 +987,5 @@ def summarize(self, minimum_ionic_fraction: Real = 0.01) -> NoReturn:
         else:
             output_string = output[0]
 
-        print(output_string.strip("\n"))
+        # We'll need to switch from print() to using logging library
```
namurphy, line 990:
```suggestion
```

PR #2307, opened 2023-08-28
Update author information and improve functionality that generates the author list for the docs

namurphy:
- Added ORCID IDs for several contributors who I could disambiguate
 - Updated and fixed GitHub usernames which had been causing errors when doing `make linkcheck`
 - Improved the functionality that converts author information in `CITATION.cff` to a reStructuredText list.  In addition to breaking up large functions into smaller functions, this PR fixed a problem where the `:user:` role was being used even when the `alias` key representing GitHub username is unavailable (which was causing broken links to be generated).
 - Perform a Unicode to ASCII conversion in the sorting key for authors.  This way, if someone's surname starts with an Á, it will get sorted as A. 

I recently also just updated the author list in our `v2023.5.1` release.

### DIFF
--- a/docs/cff_to_rst.py
+++ b/docs/cff_to_rst.py
@@ -6,6 +6,14 @@ import pathlib
 import yaml
 
 from typing import Union
+from unidecode import unidecode
+
+# If a contributor changed their contributor name and we don't know
+# the new name to update in in CITATION.cff, add their old username to
+# the following set to prevent generating a broken link that would get
+# detected when doing a `make linkcheck`.
+
+obsolete_github_usernames = {}
 
 
 def parse_cff(filename: str) -> dict[str, Union[str, list[dict[str, str]]]]:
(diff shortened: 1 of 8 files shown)


docs/cff_to_rst.py
```diff
@@ -21,41 +29,124 @@ def parse_cff(filename: str) -> dict[str, Union[str, list[dict[str, str]]]]:
...
-    ------
-    `yaml.YAMLError`
-        If :file:`CITATION.cff` cannot be parsed.
     """
     with pathlib.Path(filename).open() as stream:
-        try:
-            return yaml.safe_load(stream)
-        except yaml.YAMLError as e:
-            print(e)  # noqa: T201
+        return yaml.safe_load(stream)
```
namurphy, line 34: By raising the exception directly, we won't risk hiding any errors resulting from this file.

PR #2318, opened 2023-08-29
Make it so no warnings are issued during tests by `plasmapy.diagnostics`

namurphy:
See also #844.


plasmapy/diagnostics/tests/test_thomson.py
```diff
@@ -241,8 +241,6 @@ def test_spectral_density_minimal_arguments(single_species_collective_args):
 
     alpha, Skw = thomson.spectral_density(*args, **kwargs)
 
-    return (alpha, wavelengths, Skw)
-
```
namurphy, line 245: pytest issues a warning when a test returns a value, which is likely to be changed to an error in the next version.

PR #2331, opened 2023-09-06
Add a `.gitattributes` file

namurphy:
This PR adds a [`.gitattributes`](https://git-scm.com/docs/gitattributes) file to our repository.  The main benefits are:

 - Consistency in line endings between different text files
 - Better git diff headers for providing context to changes

This PR also makes some minor edits to `.editorconfig` (see also #2129), and changes some files to make sure that the line endings are consistent.

PR #2338, opened 2023-09-09
Add some pre-commit hooks

namurphy:
This PR adds a few linter checks, which I'll comment on directly.


.pre-commit-config.yaml
```diff
@@ -8,16 +8,37 @@ repos:
...
+  hooks:
+  - id: fix-smartquotes
+  - id: fix-spaces
+  - id: fix-ligatures
+  - id: forbid-bidi-controls
+
+- repo: https://github.com/Lucas-C/pre-commit-hooks
+  rev: v1.5.4
+  hooks:
+  - id: remove-crlf
```
namurphy, line 34: This makes sure that the line ending is consistently LF (Unix style) rather than CRLF (Windows style).  See also #2331.

.pre-commit-config.yaml
```diff
@@ -8,16 +8,37 @@ repos:
...
+
+- repo: https://github.com/MarcoGorelli/cython-lint
+  rev: v0.15.0
+  hooks:
+  - id: cython-lint
+  - id: double-quote-cython-strings
```
namurphy, line 40:
```suggestion
```
We don't really need a Cython linter yet since we don't have any Cython code...

.pre-commit-config.yaml
```diff
@@ -8,16 +8,31 @@ repos:
...
   - id: check-toml
   - id: check-yaml
 
+- repo: https://github.com/sirosen/texthooks
+  rev: 0.5.0
+  hooks:
+  - id: fix-smartquotes
+  - id: fix-spaces
+  - id: fix-ligatures
+  - id: forbid-bidi-controls
```
namurphy, line 29: These are some changes to make our code more ASCII-like and avoid weird Unicode issues.  `fix-spaces` is to make sure that all spaces are just regular spaces (and not special Unicode spaces), and the last one is to prevent Unicode markers that indicate direction changes.

PR #2346, opened 2023-09-26
Make `@validate_quantities` accept `Quantity[Unit]` style annotations

### LINKED ISSUE #1413: Adopt ``Quantity`` type annotations that preserve units

namurphy:
## Description

This PR makes it so that `@validate_quantities` accepts annotations of the style:

```python
@validate_quantities
def convert_to_meters(x: u.Quantity[u.m]) -> u.Quantity[u.m]:
    return x
```

I still need to update the documentation and make more thorough tests.

~~**This might not work for Python 3.9, but it should work for Python 3.10+.** We can add the capability now, but we will not be able to apply it in our repository until we drop Python 3.9 support for our 2024.5.0 release.~~

I was pleasantly surprised that this is working for Python 3.9 as well as 3.10 & 3.11.  Not sure why...but I'll take it!

## Motivation and context

Before this pull request, annotations used by `@validate_quantities` had to be unit-like objects, like `f(x: u.m):`.  However, this is incompatible with mypy, which is expecting `Quantity` as an annotation.  This PR will make it so that annotations can actually be type hint annotations.

## Related issues

Closes #1413. Related to #268.

### DIFF
--- a/changelog/2346.trivial.rst
+++ b/changelog/2346.trivial.rst
@@ -1,0 +1,2 @@
+Enabled |validate_quantities| to accept annotations of the form
+``u.Quantity[u.m]``, where we have run :py:`import astropy.units as u`.
--- a/plasmapy/utils/decorators/checks.py
+++ b/plasmapy/utils/decorators/checks.py
@@ -217,8 +217,8 @@ class CheckValues(CheckBase):
         Returns
         -------
         Dict[str, Dict[str, bool]]
-            A complete 'checks' dictionary for checking function input arguments
-            and return.
+            A complete 'checks' dictionary for checking function input
+            arguments and return.
         """
         # initialize validation dictionary
         out_checks = {}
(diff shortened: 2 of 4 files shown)


ejohnson-96 (approved):
👍

PR #2361, opened 2023-10-05
Consolidate plasma dispersion function lite-functions with regular functions

### LINKED ISSUE #2330: Update or consolidate lite function versions of plasma dispersion function and derivative
I just noticed in #2271 that `plasma_dispersion_func_lite` and `plasma_dispersion_func_deriv_lite` are currently not decorated with any Numba decorators, and thus are not being just-in-time compiled.

I don't know if we need separate lite functions for the plasma dispersion function and its derivative because they are intended for complex numbers and arrays, rather than `Quantity` objects (although there could be dimensionless `Quantity` usage).  I'm wondering if we could make the plasma dispersion function itself a lite function (though not in name).  

Our options are:
 - Decorate `plasma_dispersion_func_lite` and `plasma_dispsersion_func_deriv_lite` with `@numba.jit` so that it gets just-in-time compiled.
 - Cythonize these functions instead.
 - Remove the plasma dispersion function lite functions entirely
 - Consolidate the lite functions into cythonized/jitted regular functions (if we can figure out how to handle dimensionless `Quantity` objects.

namurphy:
## Description

This PR removes the `lite` attributes of `plasma_dispersion_func` and `plasma_dispersion_func_deriv`, while improving the performance of the original functions.  

Closes #2330.

## Motivation and context

The lite-functions for the plasma dispersion function and its derivative were not jitted by Numba, and the only difference was the lack of validations in the lite function.  As of Python 3.11, [try/except blocks now have minimal overhead](https://docs.python.org/3.11/whatsnew/3.11.html#misc), so this motivates us to do EAFP rather than LBYL.  

In general, we shouldn't need lite-functions when the original functions generally don't deal with Quantity objects.  In those cases, we should make the original function fast.  

## Additional context

One might almost say that we are turning the lites off. 💡 😹


plasmapy/dispersion/dispersion_functions.py
```diff
@@ -219,41 +120,17 @@ def plasma_dispersion_func_deriv(
...
-
     """
-
-    if not isinstance(
-        zeta, (numbers.Integral, numbers.Real, numbers.Complex, np.ndarray, u.Quantity)
-    ):
+    try:
+        return -2 * (1 + zeta * plasma_dispersion_func(zeta))
+    except u.UnitTypeError as wrong_units:
+        raise u.UnitsError(
```
ejohnson-96, line 127: Coverage on tests is missing here

ejohnson-96 (commented):
Documentation is failing because in the doctoring of dispersion_functions.py (plasmapy.dispersion.dispersion_functions.plasma_dispersion_func_deriv) the numpy. references cannot be found

namurphy:
@ejohnson-96 — Might you have a chance to look at this again since it's changed since when you approved it a few weeks ago?

plasmapy/dispersion/dispersion_functions.py
namurphy: Thanks! I'll add a test for this before merging.

ejohnson-96 (changes requested):
TestPlasmaDispersionFunction is causing a failure on the Python 3.9 test, specifically the test_plasma_dispersion_func_symmetry test. AssertionError is raised, I'm not sure if it's the complex numbers that is throwing it? 

L127 in dispersion_functions.py is not covered in test cases

PR #2367, opened 2023-10-16
Update metadata for `v2023.10.0`

namurphy:
I also updated the DOI that cites all versions of PlasmaPy that are under the `team@plasmapy.org` account.


ejohnson-96 (approved):
👍

Issue #2319, opened 2023-08-30
Fix `DeprecationWarning` in grids functionality regarding deleting a single level of a `MultiIndex`


namurphy:
While working on reducing the number of warnings (see also #844), I noticed that there is a DeprecationWarning that we probably should fix.  

> `DeprecationWarning: Deleting a single level of a MultiIndex is deprecated. Previously, this deleted all levels of a MultiIndex. Please also drop the following variables: {'ax'} to avoid an error in the future.`

PR #2369, opened 2023-10-16
Reduce warnings issued in tests & deprecate passing real number to `charge` in `CustomParticle`

namurphy:
This PR reduces the number of warnings issued by tests.  I'm resolving a few of them and filtering others.

This PR also removes the (previously deprecated) ability to pass a real number to the `charge` parameter in `CustomParticle`, since the preferred behavior is now to pass it to `Z`.  I made sure to issue a descriptive error.

I'm filtering deprecation warnings for `MultiIndex` in `test_grids.py` (see #2319) and traitlets (see #2370).  These deprecations could possibly cause problems later on, but we don't need to see the warnings every time tests run since we have issues now created.

### DIFF
--- a/plasmapy/formulary/tests/test_dielectric.py
+++ b/plasmapy/formulary/tests/test_dielectric.py
@@ -127,7 +127,7 @@ class Test_permittivity_1D_Maxwellian:
                 "T": 30 * 11600 * u.K,
                 "n": 1e18 * u.cm**-3,
                 "particle": "Ne",
-                "z_mean": 8,
+                "Z": 8,
                 "omega": 5.635e14 * 2 * np.pi * u.rad / u.s,
             },
             (-6.729556105413448e-08 + 5.761632594678113e-07j)
--- a/plasmapy/formulary/tests/test_dielectric.py
+++ b/plasmapy/formulary/tests/test_dielectric.py
@@ -127,7 +127,7 @@ class Test_permittivity_1D_Maxwellian:
                 "T": 30 * 11600 * u.K,
                 "n": 1e18 * u.cm**-3,
                 "particle": "Ne",
-                "Z": 8,
+                "z_mean": 8,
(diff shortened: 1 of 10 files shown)


plasmapy/formulary/tests/test_dielectric.py
```diff
@@ -127,7 +127,7 @@ class Test_permittivity_1D_Maxwellian:
                 "T": 30 * 11600 * u.K,
                 "n": 1e18 * u.cm**-3,
                 "particle": "Ne",
-                "z_mean": 8,
+                "Z": 8,
```
namurphy, line 130:
```suggestion
                "z_mean": 8,
```

namurphy:
With this PR, we'll be down to ≤8 warnings that show up during tests!  Not bad, considering we used to have a few thousand.  Eventually I want to set the pytest flag that will turn warnings into errors so that, moving forward, we don't have any new unfiltered warnings come up. We'll need to update the testing guide to discuss this. 

I should note that the behavior being warned about is still happening, though often the warning is not the behavior being tested (which is often the case when a `PhysicsWarning` is issued).

plasmapy/utils/calculator/main_interface.py
```diff
@@ -22,6 +23,12 @@
     _process_queue,
 )
 
+warnings.filterwarnings(
+    action="ignore",
+    message="Passing unrecognized arguments",
+    category=DeprecationWarning,
+)  # see issue 2370 for a deprecation warning from traitlets
+
```
namurphy, line 31: Strangely, we were getting deprecation warnings from this file when running tests, even though this file is not run during tests (I think).

ejohnson-96: Hmmmmmm?

ejohnson-96 (approved):

namurphy:

PR #2372, opened 2023-10-17
Temporarily remove Numba dependency for `v2023.10.0` release before returning it in `v2023.10.1`

namurphy:
## Description

This PR temporarily (and experimentally!) removes Numba as a dependency for our `v2023.10.0`.  This PR will be reverted in a subsequent PR for a `v2023.10.1` release.

## Motivation and context

After a new version of Python is released in October of each year, it can take close to half a year before Numba becomes compatible with it.  Because PlasmaPy requires Numba, we are unable to use PlasmaPy on Python 3.12 during that time period.  By removing our Numba dependency on lite-functions, we will have a recent release that can be used on Python 3.12.  The lite functions won't be jitted, but everything will at least work.  

## Additional context

I am going to put the changelog entry in a different PR so that this PR can be reverted more easily.

### DIFF
--- a/plasmapy/formulary/frequencies.py
+++ b/plasmapy/formulary/frequencies.py
@@ -14,7 +14,6 @@ import numbers
 import numpy as np
 
 from astropy.constants.si import e, eps0
-from numba import njit
 from typing import Optional
 
 from plasmapy import particles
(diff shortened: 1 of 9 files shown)


namurphy:
There's something weird happening here with units and `scipy.special.wofz`, which appears to be related to the plasma dispersion function.  I might put this idea on hold.

namurphy:
Closing this since there are some errors that I'm having trouble tracking down, and don't want to delay the release.  If Numba hasn't released a version compatible with Python 3.12 by January, then we can try out this idea then.

PR #2378, opened 2023-10-19
Update `plasmapy/__init__.py`

namurphy:
This PR makes a few minor updates to `plasmapy/__init__.py`, including the docstring at the top.

### DIFF
--- a/plasmapy/__init__.py
+++ b/plasmapy/__init__.py
@@ -1,8 +1,13 @@
 """
-Welcome to the `plasmapy` package, an open source community-developed
-Python package for the plasma community. Documentation is available in
-the docstrings and `online <https://docs.plasmapy.org>`__ (accessible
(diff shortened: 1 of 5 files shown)


plasmapy/__init__.py
```diff
@@ -18,21 +23,15 @@
...
-        "Please upgrade to a newer version."
+        "This version of PlasmaPy does not support Python "
+        f"{sys.version.split()[0]}. Please upgrade to a newer version "
+        "of Python."
     )
 
-# Packages may add whatever they like to this file, but
-# should keep this content at the top.
-# ----------------------------------------------------------------------------
-from importlib.metadata import PackageNotFoundError
```
namurphy, line 34: I don't think we need a `PackageNotFoundError` here, since it was not being used in this file.  Running the tests will let us find out!

PR #2381, opened 2023-10-20
Remove `# coverage: ignore` in most places for more realistic coverage reports

namurphy:
## Description

This PR removes most comments which say that code coverage checks should be ignored.

## Motivation and context

We currently use `# coverage: ignore` when we have a block of code that do not want to include in code coverage checks.  We've used that somewhat liberally in the past.  The problem with this practice is that we end up sweeping lines of code under the rug, which means that we think our code coverage is higher than it is.  Removing the `# coverage: ignore` comments will make it so that the lines of code that aren't tested show up as not being tested.  Also, some lines of code with that comment might actually be run during tests.

## Related issues

It's common for other projects to use `# pragma: no cover`, but we decided on using `# coverage: ignore` because it's less confusing to newcomers to the project.

This PR is also a check to make sure that changes to GitHub Actions in #2379 and #2380 are still working okay.

### DIFF
--- a/plasmapy/diagnostics/langmuir.py
+++ b/plasmapy/diagnostics/langmuir.py
@@ -166,7 +166,7 @@ class Characteristic:
         if len(np.unique(self.bias)) != len(self.bias):
             raise ValueError("Bias array contains duplicate values.")
 
-    def get_padded_limit(self, padding, log=False):  # coverage: ignore
+    def get_padded_limit(self, padding, log=False):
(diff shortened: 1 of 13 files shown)


namurphy:
This PR is probably going to set the record for largest drop in code coverage!

namurphy:
Interestingly, most of the lines that had `# coverage: ignore` were actually covered by tests.  Go us!

PR #2392, opened 2023-10-23
Separate GitHub Action for checking hyperlinks into its own action

namurphy:
## Description

This PR moves the `make linkcheck` GitHub Action from the weekly tests into a separate file so that it can be run separately.

## Motivation and context

Because [link rot](https://en.wikipedia.org/wiki/Link_rot) is so pervasive, our hyperlink check frequently fails.  We will sometimes get intermittent test failures when a hyperlink temporarily becomes unavailable before being restored.  The nature of this failure is quite different than other failures in our cron tests since it is in an external, changing environment.

### DIFF
--- a/.github/workflows/check-hyperlinks.yml
+++ b/.github/workflows/check-hyperlinks.yml
@@ -1,0 +1,31 @@
+name: Check hyperlinks
+
+on:
+  schedule:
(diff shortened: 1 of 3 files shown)


namurphy:
I'm merging this one now in order to be able to test that this action is working.

PR #2421, opened 2023-12-20
Change unit annotations to `Quantity` annotations

namurphy:
## Description

This PR changes annotations like `u.kg` to `u.Quantity[u.kg]` so that they are true type hint annotations:

```python
import astropy.units as u

# old style unit annotations
@validate_quantities
def magnetic_pressure(B: u.T) -> u.Pa:
    ...

# new style with actual type hint annotations
@validate_quantities
def magnetic_pressure(B: u.Quantity(u.T)) -> u.Quantity(u.Pa):
    ...
```

## Motivation and context

At present, we use annotations like `u.kg` to specify the required units for `@validate_quantities`, which then verifies that the quantities are of the appropriate physical type and converts them to the correct units.  However, this prevents us from using static type checking tools like [mypy](https://mypy-lang.org). While these are annotations, they are not *type hint* annotations because things like `u.kg` are not a type.

Enabling mypy (#2424) will let us catch errors in type hint annotations, which we currently have no way of doing.  

Here's an example docstring [before (with unit annotations)](https://docs.plasmapy.org/en/v2023.10.x/api/plasmapy.formulary.collisions.dimensionless.coupling_parameter.html#plasmapy.formulary.collisions.dimensionless.coupling_parameter) and [after (with `Quantity` annotations)](https://plasmapy--2421.org.readthedocs.build/en/2421/api/plasmapy.formulary.collisions.dimensionless.coupling_parameter.html#coupling-parameter).

## Related issues

Related to: #268, #2424

PR #2424, opened 2023-12-21
Add mypy configuration file that ignores existing errors on a per-file and per-error basis

### LINKED ISSUE #268: Use mypy for static type checking in continuous integration

namurphy:
## Description

The PR adds an initial mypy configuration file, `mypy.ini`. To get it to pass, I set it to ~~ignore a whole bunch of files~~ ignore specific error codes in a bunch of files.

## Motivation and context

We currently have no method for checking that our type hint annotations are accurate. Enabling the `mypy` static type checker will fill this gap in our continuous integration testing.

Prior to #2421 and this PR, running `mypy plasmapy` led to 719 errors, of which ∼200 are from untyped packages like Astropy and SciPy. Rather than fixing all of these errors at once, I wanted to add a configuration for `mypy` that passes, even if it means that we ignore errors in a whole bunch of files. After merging this, we can gradually go through PlasmaPy file-by-file and error code by error code to get `mypy` to pass in strict mode. For errors that are difficult to deal with we could add `# type: ignore[specific-error-code]` comments so that the rest of the file gets type checked.

## Related issues

Closes #268. Depends on #2421.


namurphy:
I just got mypy to pass for the first time in continuous integration ever on PlasmaPy, and all I needed to do was set it to completely ignore 79 files. 🥲  Oh well...it's a starting point!

namurphy:
The number of files that aren't passing mypy went down after #2421, as expected!

.pre-commit-config.yaml
```diff
@@ -85,6 +85,12 @@ repos:
     additional_dependencies:
     - tomli
 
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.7.1
+  hooks:
+  - id: mypy
+    exclude: ^docs|^\.jupyter|^\.github
+
```
namurphy, line 93:
```suggestion
```
Following [this comment](https://github.com/astropy/astropy/issues/15170#issuecomment-1870944481) in https://github.com/astropy/astropy/issues/15170, I will remove the pre-commit hook before merging. This PR would then only add the configuration file (plus a few minor edits here and there to get mypy to pass).  In a future PR, we can set up a tox environment and enable it in CI.

pyproject.toml
```diff
@@ -84,11 +84,14 @@ docs = [
...
 tests = [
   "hypothesis >= 6.35.1",
+  "pandas-stubs >= 1.5.0.221010",
   "pre-commit >= 3.0.0",
   "pytest-regressions >= 2.3.1",
   "pytest-xdist >= 3.0.2",
   "tomli >= 2.0.1",
   "tox >= 4.3.1",
+  "types-requests >= 2.27.1",
+  "types-tqdm >= 4.66.0.3",
```
namurphy, line 94: For packages that don't include type hints themselves, there sometimes exists stub packages that contain typing information.

pheuer: Is it worth adding a line comment on these packages explaining what they are (type information for the associated package)?

.pre-commit-config.yaml
pheuer: So, this PR just adds the capability to run mypi locally during development, but doesn't turn it on in CI?

mypy.ini
```diff
@@ -0,0 +1,553 @@
...
+exclude = (?x)(
+          docs
+          )
+
+enable_error_code = ignore-without-code
+
+# We can get to passing mypy's strict mode by incrementally changing the
+# following global settings to true. This comes from mypy's documentation.
+
+warn_unused_configs = true
```
pheuer, line 13: Right now they are all `true` - you mean you'd change all/some of these to `false` prior to adding this to CI?

pheuer (approved):
Looks good - shouldn't really change much for users/developers until added to CI, except for the few new minor added requirements. 

One note - presumably between now and when CI is turned on, more changes will be made to the code and it's possible the list of exception files will change and need to be updated in the PR that turns on CI?

mypy.ini
namurphy: Good catch!  I was debating back and forth about this.  If we keep all of these true, then we can ensure that new files will be consistent with mypy's strict configuration.  Adding type hint annotations is much easier while writing code, so I'm thinking we should try out strict mode.  If any of the settings cause any headaches, we can set them to be false.

In this case...I need to update the comment!

pheuer: I agree. On the other hand, if you want this in the package I'm in favor of getting it into CI as soon as possible to at least stem the flow of new files without type hints. Maybe we can change some of these to false when adding this to CI to ease into it.

namurphy:
@pheuer — thank you for the review!  This is by no means the final version of the mypy configuration.  We'll need to play around with it to see what it's like.

> Looks good - shouldn't really change much for users/developers until added to CI, except for the few new minor added requirements.

True!  I'm hoping to submit a PR to add mypy to CI soon, but we'll need to add some content to the contributor guide along with it.

> One note - presumably between now and when CI is turned on, more changes will be made to the code and it's possible the list of exception files will change and need to be updated in the PR that turns on CI?

Yes, this is unfortunately true too.  I'm planning to regenerate the configuration file unless I get it added to CI sooner rather than later.  My biggest worry about this approach is that it'll be hard to maintain manually.  But...this configuration is only a starting point.

Thank you again for the review — I appreciate it!

.pre-commit-config.yaml
namurphy: Correct!  Adding it to CI will come soon, probably.

pyproject.toml
namurphy: Good idea!

plasmapy/diagnostics/charged_particle_radiography/detector_stacks.py
```diff
@@ -60,8 +61,8 @@ def __init__(
         self,
         thickness: u.Quantity[u.m],
         energy_axis: u.Quantity[u.J],
-        stopping_power: [u.J / u.m, u.J * u.m**2 / u.kg],
-        mass_density: [u.kg / u.m**3, None] = None,
+        stopping_power: u.Quantity[u.J / u.m, u.J * u.m**2 / u.kg],
+        mass_density: Optional[u.Quantity[u.kg / u.m**3]] = None,
```
namurphy, line 65: I missed these in #2421.

Issue #2150, opened 2023-05-23
Enable `@particle_input` to work for callables that accept positional-only and var-positional arguments


namurphy:
At present, the `plasmapy.particles.particle_input` decorator doesn't work when a function is provided with var-positional or (I think) positional-only arguments.  This is because `@particle_input` assigns arguments to a dictionary with parameters as keys and corresponding arguments as values, and then [passes the dictionary to the decorated callable as keyword arguments](https://github.com/PlasmaPy/PlasmaPy/blob/c52e5135765c4e592c840d9e3ca45b5a43fdc3a1/plasmapy/particles/decorators.py#L900). We should change it so that positional arguments remain positional arguments when provided to the decorated callable.  This would involve changing `callable__(**new_kwargs)` to `callable__(*new_args, **new_kwargs)` in that line, where the only arguments that get changed are the ones with appropriate annotations.

I ran into this problem with #1029, but it's rare that we run into this problem, so it's not high priority. On the other hand, we would improve the fidelity of the decorator by making this change, so it would still be worth doing at some point.

For definitions of positional-only and var-positional, see Python glossary entry on [parameters](https://docs.python.org/3/glossary.html#term-parameter). See also FAQ on [difference between arguments and parameters](https://docs.python.org/3/faq/programming.html#faq-argument-vs-parameter).

PR #2428, opened 2023-12-27
Improve `@particle_input`'s ability to handle variadic positional arguments

namurphy:
## Description

This PR enables `@particle_input` to decorate functions with a signature like:

```python
@particle_input
def f(*args, particle: ParticleLike = "p+"):
    ...
```

This fix does not extend to functions where there are positional arguments prior to variadic positional arguments, and instead adds an xfailing test for this.

This PR also does some minor refactoring of `@particle_input` (including how it processes arguments) and updates some docstrings.

## Motivation and context

I ran into this problem again when coming back to #1029, since I'm considering adding a signature that uses this pattern. 

I can't think of a straightforward way to address the case where there are positional arguments before variadic positional arguments, so I decided to not address that here and instead do an incremental change.

## Related issues

This is a partial fix of #2150.

### DIFF
--- a/changelog/2428.feature.rst
+++ b/changelog/2428.feature.rst
@@ -1,2 +1,2 @@
-Enable |particle_input| to decorate functions which have variadic
+Enabled |particle_input| to decorate functions which have variadic
 positional arguments followed by keyword arguments. See :issue:`2150`.
--- a/changelog/2428.feature.rst
+++ b/changelog/2428.feature.rst
@@ -1,0 +1,2 @@
+Enable |particle_input| to decorate functions which have variadic
+positional arguments followed by keyword arguments. See :issue:`2150`.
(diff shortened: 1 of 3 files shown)


pheuer (approved):
So replacing the dict of keyword arguments with `BoundArguments` basically resolves the problem for this case, but not for the case of positional args followed by variadic positional args then keyword args? Does `insepct.BoundArguments` just not distinguish between the two groups of positional arguments then? 

Anyway, this is a clear improvement on the decorator (functionality and probably cleaner) and looks like it works.

namurphy:
Thank you again!

> So replacing the dict of keyword arguments with `BoundArguments` basically resolves the problem for this case, but not for the case of positional args followed by variadic positional args then keyword args? Does `insepct.BoundArguments` just not distinguish between the two groups of positional arguments then?

I'm not completely sure why, but I think that's what's happening.

PR #2429, opened 2023-12-27
Improve type hint annotations in `plasmapy.particles`

namurphy:
## Description

This PR makes some incremental improvements to typing in `plasmapy.particles` with help from `mypy`.

## Motivation and context

Instead of adding types all at once, it's preferable to add them in small, more reviewable chunks.

## Related issues

#268, #2424

PR #2431, opened 2023-12-29
Add a tox environment for running mypy

### LINKED ISSUE #268: Use mypy for static type checking in continuous integration

namurphy:
This PR adds a tox environment for running the mypy static type checking tool.  This tox environment will be added to our suite of continuous integration tests in #2432.

I am separating this PR out so that I can use this tox environment in other PRs, like #2429.

Related to #268. The motivation for using mypy is discussed in #2424.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,19 +82,13 @@ docs = [
   "tox >= 4.4.0",
   "unidecode >= 1.3.6",
 ]
-# Some of these packages contain typing information for packages that
-# don't themselves contain type hints. We might want to move these from
-# the tests section and into the specific mypy environment.
 tests = [
   "hypothesis >= 6.35.1",
-  "pandas-stubs >= 1.5.0.221010",  # used by mypy
   "pre-commit >= 3.0.0",
   "pytest-regressions >= 2.3.1",
   "pytest-xdist >= 3.0.2",
   "tomli >= 2.0.1",
   "tox >= 4.3.1",
-  "types-requests >= 2.27.1",  # used by mypy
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -82,19 +82,13 @@ docs = [
...
 tests = [
   "hypothesis >= 6.35.1",
-  "pandas-stubs >= 1.5.0.221010",  # used by mypy
   "pre-commit >= 3.0.0",
   "pytest-regressions >= 2.3.1",
   "pytest-xdist >= 3.0.2",
   "tomli >= 2.0.1",
   "tox >= 4.3.1",
-  "types-requests >= 2.27.1",  # used by mypy
-  "types-tqdm >= 4.66.0.3",  # used by mypy
```
namurphy, line 97: Turns out that the type stub packages can be installed using the `--install-types` flag for mypy, so they're not needed here. Using that flag has the advantage that it'll probably get updated with type stub packages as more are created or obsoleted, which means less work for us to do!

tox.ini
```diff
@@ -153,3 +153,9 @@ basepython = python3.9
 extras =
 deps =
 commands = python -c 'import plasmapy'
+
+[testenv:mypy]
+deps =
+    mypy >= 1.8.0
+commands =
+    mypy plasmapy --install-types --non-interactive --show-error-context --show-error-code-links --pretty
```
namurphy, line 161: We'll probably want to play with these flags to get the error reporting settings that work best for us.

PR #2356, opened 2023-10-03
Update docstrings and annotations in formulary length calculations

namurphy:
## Description

I'm updating the docstring of `gyroradius`, including some reorganization and more narrative examples.

## Motivation and context

I was looking through the online documentation and noticed that some of the symbols in the docstring for `gyroradius` were specific to ions.

PR #2432, opened 2023-12-29
Add mypy to continuous integration suite

### LINKED ISSUE #268: Use mypy for static type checking in continuous integration

namurphy:
## Description

This PR adds [`mypy`](https://mypy.readthedocs.io/en/stable/) to our continuous integration suite. 

## Motivation and context

mypy is the *de facto* static type checker in the pythoniverse. It analyzes source code directly and makes sure that type hint annotations are used consistently throughout the package.

Adopting mypy in our CI will help us:

 - Find bugs and common programming errors.
 - Ensure our type hint annotations are accurate and complete. 
 - Better understand what our functions are doing, especially in complicated situations.

Additional considerations or tradeoffs are:

 - Code with type hint annotations is more verbose.  
 - We would need a good way to introduce type hint annotations to new contributors, such as in the contributor guide and in the comment that gets posted to every PR.  We can point to mypy's typing cheatsheet, and discuss some considerations that are specific to PlasmaPy.  
 - I've found error messages from mypy to...take some getting used to.  Additionally, mypy may point to an error on a particular line, when it's a different line that actually needs fixing.
 - mypy is unable to type check objects that are created dynamically rather than being defined statically in the source code (see also https://github.com/astropy/astropy/issues/15808).
 - Our ability to type check things will be limited when upstream packages are not typed, including `astropy.units`.  We can also add type stub files (like in #2442) for important functionality, or `# type: ignore[error-code]` comments for less important situations.  These would need only include the functionality that we actually use.

## Related issues

Closes #268, which was raised almost six years ago!

In #2424, we added a mypy configuration file that ignores existing errors on a per-file and per-error basis. In #2431, we added a tox environment for running mypy.  This PR adds that tox environment to our suite of continuous integration checks.

We also changed `@validate_quantities` to work with `u.Quantity[u.m]` style comments in #2346, following by actual changes to the annotations in #2356 and #2421.


tox.ini
```diff
@@ -158,4 +158,4 @@ commands = python -c 'import plasmapy'
 deps =
     mypy >= 1.8.0
```
namurphy, line 159: I'm wondering if there's an alternative to having to update this manually...  dependabot, maybe?

tox.ini
```diff
@@ -158,4 +158,4 @@ commands = python -c 'import plasmapy'
...
 commands =
-    mypy plasmapy --install-types --non-interactive
+    mypy plasmapy --install-types --non-interactive --show-error-context --show-error-code-links --pretty
```
namurphy, line 161: We'll probably want to play with these flags to figure out which error output works best.

namurphy:
It's still really surprising to see mypy passing in our CI suite!  I mean...we've got it set to ignore ∼700 errors, but still!  😺

tox.ini
ejohnson-96: :thinking:

ejohnson-96 (approved):
:+1: All looks good

StanczakDominik (approved):
Looks quite sensible, let's try it out :) Please allow me to do something I haven't done in a while...

namurphy:
Thank you for the reviews!  I'm curious about what @rocco8773 thinks about this too.  I'm really happy to have this added to our CI suite since it's been something I've been hoping to do for several years.  

Enabling mypy will level up the quality of our code and documentation, but my main worry is that it'll be an extra barrier to entry for new contributors.  So, we should keep thinking of our mypy configuration as experimental.  I currently have it set so that new modules are expected to follow mypy's strict settings, but we may want to loosen those settings later on for some modules.

@StanczakDominik — It's super awesome to hear from you and I hope you've been doing really well!

PR #2437, opened 2024-01-02
Annotate dunder methods using `autotyping` and change `NoReturn` annotations to `None`

namurphy:
This PR uses [autotyping](https://github.com/JelleZijlstra/autotyping) to add type hint annotations to dunder methods like `__init__`, `__str__`, etc.

These are the commands I used in the top-level directory, after having temporarily added a config file as per the [autotyping usage guidelines](https://github.com/JelleZijlstra/autotyping#usage).

```bash
pip install autotyping
python -m libcst.tool codemod autotyping.AutotypeCommand plasmapy --annotate-magics
```

I automatically updated `mypy.ini` to reflect the necessary changes in per-file ignores, and `ruff` did some autoformatting too.

I also changed `NoReturn` to `None` for return annotations.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,8 +3,8 @@ python_version = 3.9
 pretty = true
 exclude = (?x)(
           docs|
-          .run|
-          .tox
+          \.run|
+          \.tox
           )
 
 enable_error_code = ignore-without-code
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,6 +3,7 @@ python_version = 3.9
 pretty = true
 exclude = (?x)(
           docs|
+          .run|
           .tox
           )
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -68,7 +68,7 @@ ignore_errors = true
 # remove the disabled error codes file-by-file.
 
 [mypy-plasmapy]
-disable_error_code = no-untyped-def,no-redef
+disable_error_code = no-untyped-def
 
 [mypy-plasmapy.analysis.fit_functions]
 disable_error_code = assignment,attr-defined,misc,name-match,no-any-return,no-untyped-call,no-untyped-def,type-arg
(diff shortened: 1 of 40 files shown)


plasmapy/plasma/sources/openpmd_hdf5.py
```diff
@@ -72,7 +74,12 @@ def __enter__(self):
     def close(self):
         self.h5.close()
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(
+        self,
+        exc_type: Optional[type[BaseException]],
```
namurphy, line 79: This line originally used `typing.Type`, but `ruff` reported that this was deprecated and we should use `type` instead.

mypy.ini
```diff
@@ -488,13 +483,13 @@ disable_error_code = arg-type,no-untyped-def
 disable_error_code = no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.utils.calculator.widget_helpers]
-disable_error_code = attr-defined,no-untyped-call,no-untyped-def,union-attr,var-annotated
+disable_error_code = assignment,attr-defined,no-untyped-call,no-untyped-def,var-annotated
```
namurphy, line 486: Going from one type of error to another...progress?  😹

mypy.ini
```diff
@@ -104,13 +105,13 @@ disable_error_code = arg-type,no-untyped-call,no-untyped-def
 disable_error_code = no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.analysis.time_series.excess_statistics]
-disable_error_code = no-untyped-call,no-untyped-def
+disable_error_code = no-untyped-call,no-untyped-def,var-annotated
```
namurphy, line 108: I suspect new errors are being reported because untyped functions & methods were not previously being type checked.

namurphy:
The mypy documentation contains [suggestions about annotating init methods](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods).  Essentially, it would be fine to drop the annotation on `__init__` if at least one parameter in the signature is annotated.  I'm inclined to keep the `-> None` annotations for `__init__` methods for the time being, until we make more progress on typing.

pheuer (approved):
Individually these all look like good changes - my only question is, what's the long-term plan here? Is this a one-off improvement? Since this isn't enforced by CI on new PRs, do you plan to re-run this occasionally to add these typings to new code?

namurphy:
Thank you again for the reviews!

> Individually these all look like good changes - my only question is, what's the long-term plan here? Is this a one-off improvement? Since this isn't enforced by CI on new PRs, do you plan to re-run this occasionally to add these typings to new code?

Merging #2432 will make it so that mypy will perform checks like these in CI, depending on the strictness settings that we choose. I'm also hoping to extend the ruff settings to include more of the [ANN rule set](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann).  These are important questions — thank you for asking!

PR #2439, opened 2024-01-02
Use autotyping to add `-> None` return annotations

namurphy:
## Description

I used `autotyping` to add `-> None` annotations to functions and methods with no `return` statement. I'm excluding special methods, which were covered in #2437.

## Context

One thing to note is that this adds a lot of `-> None` annotations to test functions and methods.  The main advantage of having the `-> None` annotations in tests is that it enforces that tests are type checked with mypy, since functions and methods without any annotations are seemingly not type checked.  The `-> None` annotation will also let type checkers verify that tests do not have any `return` statements in them.   

## Related issues

This is similar to #2437, only I used the ``--none-return`` flag instead.

### DIFF
--- a/plasmapy/diagnostics/tests/test_langmuir.py
+++ b/plasmapy/diagnostics/tests/test_langmuir.py
@@ -24,7 +24,7 @@ class Test__fitting_functions:
     T0 = 3
     Delta_T = 15
 
-    def test_fit_func_lin(self):
+    def test_fit_func_lin(self) -> None:
         r"""Test linear fitting function"""
 
         value = langmuir._fit_func_lin(self.x, self.x0, self.y0, self.c0)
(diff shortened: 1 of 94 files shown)


plasmapy/diagnostics/tests/test_langmuir.py
```diff
@@ -198,7 +198,7 @@ class DryCharacteristic(langmuir.Characteristic):
     unique values.
     """
 
-    def __init__(self, bias, current):
+    def __init__(self, bias, current) -> None:
```
namurphy, line 201:
```suggestion
    def __init__(self, bias, current):
```
...since it's covered in another PR.

plasmapy/formulary/relativity.py
```diff
@@ -328,7 +328,7 @@ def __init__(
         Z: Optional[Integral] = None,
         mass_numb: Optional[Integral] = None,
         dtype: Optional[DTypeLike] = np.longdouble,
-    ):
+    ) -> None:
```
namurphy, line 331:
```suggestion
    ):
```

pheuer (approved):
Same comment as #2437 - is the plan to just occasionally run this to catch new issues? Or will this somehow become enforced in CI?

PR #2442, opened 2024-01-03
Add type stub file for `@wrapt.decorator`

namurphy:
## Description

This PR adds a [type stub file](https://mypy.readthedocs.io/en/stable/stubs.html) for [`wrapt`](https://wrapt.readthedocs.io/).

## Motivation and context

A few of our widely used decorators like `@particle_input` make use of `@wrapt.decorator`. Because `wrapt` is untyped, we were running into issues with mypy not type checking some of our code because `@wrapt.decorator` is untyped. Adding a type stub file lets us get past that.  This also serves as a prototype type stub file.

## Related issues

This came up in #2429.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -54,9 +54,6 @@ ignore_missing_imports = true
 [mypy-setuptools.*]
 ignore_missing_imports = true
 
-[mypy-wrapt.*]
-ignore_missing_imports = true
-
 [mypy-plasmapy._dev.scm_version]
 ignore_errors = true
 
@@ -67,9 +64,6 @@ ignore_errors = true
 # typing errors should not be reported for specific modules. As we
 # gradually make PlasmaPy consistent with mypy's strict mode, we can
 # remove the disabled error codes file-by-file.
-
-[mypy-plasmapy]
-disable_error_code = no-untyped-def
 
 [mypy-plasmapy.analysis.fit_functions]
 disable_error_code = assignment,attr-defined,misc,name-match,no-any-return,no-untyped-call,no-untyped-def,type-arg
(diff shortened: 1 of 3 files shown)


mypy.ini
```diff
@@ -1,10 +1,11 @@
...
 pretty = true
+mypy_path = ./type_stubs
 exclude = (?x)(
-          docs|
-          \.run|
-          \.tox
+          /docs/
+          | /\.run/
+          | /\.tox/
           )
```
namurphy, line 9: This corrects the paths to exclude when doing static type checking with mypy.

mypy.ini
```diff
@@ -68,9 +66,6 @@ ignore_errors = true
 # gradually make PlasmaPy consistent with mypy's strict mode, we can
 # remove the disabled error codes file-by-file.
```
namurphy, line 68: All of the changes in this file made below this line were done automatically, as described in #2424.

mypy.ini
pheuer: Independent of the new stub file, correct?

pheuer (approved):
Sounds good - are there any other dependencies with the same problem/solution?

mypy.ini
namurphy: Correct, this is an independent change.

namurphy:
> Sounds good - are there any other dependencies with the same problem/solution?

Yes.  Several. The most pressing and most difficult will be `astropy.units`, which is being discussed in https://github.com/astropy/astropy/issues/15808.  I don't want to permanently add a type stub file for `astropy.units` here, since it would make a lot more sense to add that in Astropy itself.

PR #2443, opened 2024-01-04
Improve type hint annotations for `@particle_input`

### LINKED ISSUE #2434: Get `@particle_input` to pass mypy's strict mode
We use `@particle_input` throughout the package, so it should be a high priority to get it to work with mypy.  In particular, we should get it to work with mypy's strict mode. 

See also #2424 & #2429.

namurphy:
## Description

This PR adds & improves type hint annotations in `plasmapy.particles.decorators` which is where `@particle_input` is defined, along with the tests.

There are some `# type: ignore[...]` statements for errors due to mypy's limitation as a static (rather than dynamic) type checker. We'll probably need to add that to the contributor guide.  

## Motivation and context

See #2434.  

## Related issues

This PR is a spinoff of #2429, and the motivation for #2442.  Closes #2434.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -118,34 +118,34 @@ disable_error_code = no-untyped-call,no-untyped-def
 disable_error_code = no-untyped-def
 
 [mypy-plasmapy.diagnostics.charged_particle_radiography.detector_stacks]
-disable_error_code = call-overload,no-untyped-def
+disable_error_code = call-overload,misc,no-untyped-def,type-arg,valid-type
 
 [mypy-plasmapy.diagnostics.charged_particle_radiography.synthetic_radiography]
-disable_error_code = assignment,attr-defined,has-type,misc,no-untyped-call,no-untyped-def,type-var,var-annotated
+disable_error_code = assignment,attr-defined,has-type,misc,no-untyped-call,no-untyped-def,type-arg,type-var,var-annotated
 
 [mypy-plasmapy.diagnostics.charged_particle_radiography.tests.test_detector_stacks]
 disable_error_code = no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.diagnostics.charged_particle_radiography.tests.test_synthetic_radiography]
-disable_error_code = no-untyped-call,no-untyped-def
+disable_error_code = arg-type,no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.diagnostics.langmuir]
-disable_error_code = assignment,attr-defined,misc,no-untyped-call,no-untyped-def,operator,var-annotated
+disable_error_code = assignment,attr-defined,misc,no-any-return,no-untyped-call,no-untyped-def,operator,type-arg,valid-type,var-annotated
 
 [mypy-plasmapy.diagnostics.tests.test_langmuir]
-disable_error_code = no-untyped-call,no-untyped-def
+disable_error_code = attr-defined,no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.diagnostics.tests.test_thomson]
 disable_error_code = attr-defined,no-untyped-call,no-untyped-def
 
 [mypy-plasmapy.diagnostics.thomson]
-disable_error_code = arg-type,assignment,attr-defined,misc,no-untyped-call,no-untyped-def,type-arg
+disable_error_code = arg-type,assignment,attr-defined,misc,no-untyped-call,no-untyped-def,type-arg,union-attr,valid-type
 
 [mypy-plasmapy.dispersion.analytical.mhd_waves_]
-disable_error_code = attr-defined,has-type,misc,no-untyped-call,no-untyped-def,syntax
+disable_error_code = assignment,attr-defined,has-type,misc,no-any-return,no-untyped-call,no-untyped-def,return-value,syntax,type-arg,valid-type
 
 [mypy-plasmapy.dispersion.analytical.stix_]
-disable_error_code = arg-type,assignment,attr-defined,call-overload,misc,no-untyped-call,no-untyped-def
+disable_error_code = arg-type,assignment,attr-defined,call-overload,misc,no-untyped-call,no-untyped-def,type-arg,valid-type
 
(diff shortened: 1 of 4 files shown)


plasmapy/particles/decorators.py
```diff
@@ -26,6 +27,19 @@
...
+class _CallableDataDict(TypedDict, total=False):
+    allow_custom_particles: bool
+    allow_particle_lists: bool
+    annotations: dict[str, Any]
+    any_of: Optional[Union[str, Iterable[str]]]
+    callable_: Callable[..., Any]
+    exclude: Optional[Union[str, Iterable[str]]]
+    parameters_to_process: list[str]
+    require: Optional[Union[str, Iterable[str]]]
+    signature: inspect.Signature
```
namurphy, line 40: An alternative to static type checking for this would be using [pydantic](https://docs.pydantic.dev/latest/).  But that's for another day.

plasmapy/particles/decorators.py
```diff
@@ -351,7 +367,7 @@ def verify_charge_categorization(self, particle) -> None:
 
         if isinstance(uncharged, Iterable):
             uncharged = any(uncharged)
-            lacks_charge_info = any(lacks_charge_info)
+            lacks_charge_info = any(lacks_charge_info)  # type: ignore[arg-type]
```
namurphy, line 370: For some reason, the built-in `any` and `all` annotations aren't matching the types that we're using here.

plasmapy/particles/decorators.py
```diff
@@ -562,7 +600,10 @@ def process_argument(
 
     parameters_to_skip = ("Z", "mass_numb")
 
-    def perform_pre_validations(self, Z, mass_numb):
+    def perform_pre_validations(
+        self, Z: Optional[float], mass_numb: Optional[int]
+    ) -> None:
+        # TODO: too specific?
```
namurphy, line 606:
```suggestion
```

plasmapy/particles/decorators.py
```diff
@@ -638,15 +682,19 @@ def process_arguments(
         return bound_arguments
 
 
+# TODO: callable_: Callable[[...], ...]
+# TODO: -> Callable[[...], ...]
+
+
```
namurphy, line 688:
```suggestion
```

plasmapy/particles/tests/test_decorators.py
```diff
@@ -145,15 +156,16 @@ class ClassWithDecoratedMethods:
 
     @particle_input
     def method_decorated_with_no_parentheses(self, particle: ParticleLike) -> Particle:
-        return particle
+        # Because run-time logic is needed, mypy will assume that
```
namurphy, line 159:
```suggestion
        # Because run-time logic is needed, mypy is unable to infer
        # that @particle_input is doing a type conversion here. We 
        # would need a dynamic type checker to address this case.
```

plasmapy/particles/tests/test_decorators.py
```diff
@@ -480,12 +524,14 @@ def __init__(
...
+def test_particle_input_verification(
+    kwargs_to_particle_input: dict[str, Any], arg: Any
+) -> None:
     """Test the allow_custom_particles keyword argument to particle_input."""
 
     @particle_input(**kwargs_to_particle_input)
-    def f(particle: ParticleLike):
-        return particle
+    def f(particle: ParticleLike) -> Union[Particle, CustomParticle, ParticleList]:
+        return particle  # type: ignore[return-value]
```
namurphy, line 534: The `# type: ignore` comment here reflects that static type checkers like mypy are unable to tell that there is an implicit object conversion happening here, which would require actually running the code in `@particle_input` instead of statically analyzing it.  This is going to show up in a bunch of places throughout the code.  In these cases, it's probably fine to explicitly ignore the mypy error since it is something that mypy is unable to check.

plasmapy/particles/decorators.py
namurphy: Using `TypedDict` here lets us define the types we expect for different keys.

plasmapy/particles/decorators.py
```diff
@@ -43,7 +57,7 @@
 )
 
 
-def _get_annotations(callable_: Callable):
+def _get_annotations(callable_: Callable[..., Any]) -> dict[str, Any]:
```
namurphy, line 60: For the `Callable[..., Any]` annotation, the `...` represents that the callable can accept whatever arguments, and the `Any` represents that the function can have any return type.

plasmapy/particles/decorators.py
```diff
@@ -240,10 +254,10 @@ def annotations(self) -> dict[str, Any]:
         -------
         `dict` of `str` to `object`
         """
-        return self._data.get("annotations")
+        return self._data.get("annotations")  # type: ignore[return-value]
```
namurphy, line 257: I didn't want to bother with fixing this since there's a new way to get annotations in Python 3.10, and we're dropping Python 3.9 support soon anyway.

plasmapy/particles/decorators.py
ejohnson-96: 🤔

ejohnson-96 (approved):
👍

mypy.ini
```diff
@@ -326,9 +326,6 @@ disable_error_code = type-arg,var-annotated
 [mypy-plasmapy.particles.atomic]
 disable_error_code = arg-type,assignment,misc,no-any-return,no-untyped-call,no-untyped-def,union-attr
 
-[mypy-plasmapy.particles.decorators]
-disable_error_code = arg-type,assignment,call-arg,index,misc,no-any-return,no-untyped-call,no-untyped-def,operator,return-value,type-arg,union-attr,var-annotated
```
namurphy, line 330: These are the mypy errors that existed in `plasmapy.particles.decorators` before this pull request. My process was essentially to:

1. Remove one error code and find the errors that mypy was reporting
2. Either fix those errors, or add a `#type: ignore` comment if the reported error wasn't actually a problem with the code
3. `GOTO 1`

namurphy:
Thank you for the reviews, @ejohnson-96!

PR #2448, opened 2024-01-06
Use autotyping on return values

namurphy:
This is another in a series of PRs that use autotyping to automatically add type hints. This one is using the `--scalar-return` flag.


ejohnson-96 (approved):
A very bool tea full pull request if I do say so 😆 👍

PR #2455, opened 2024-01-10
Specify a version for Ruby in pre-commit configuration

namurphy:
@pheuer and @JaydenR2305 recently ran into a problem in #2245 where pre-commit complained that it couldn't find the Ruby executable which was needed by a particular pre-commit hook.  I happened across https://github.com/pre-commit/pre-commit/issues/1930 that says that we can bypass this problem by specifying the version of Ruby that pre-commit should use.  Non-intuitive but helpful to know!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
   autoupdate_schedule: monthly
 
 default_language_version:
-  ruby: 3.3.0
+  ruby: 2.7.2
 
 repos:
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,9 @@ ci:
   autofix_prs: false
   autoupdate_schedule: monthly
 
+default_language_version:
+  ruby: 3.3.0
+
 repos:
 
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,7 @@ ci:
   autofix_prs: false
   autoupdate_schedule: monthly
 
+# Specify the default version of Ruby in case it is not installed
 default_language_version:
   ruby: 2.7.2
 


pheuer:
@namurphy I'm not sure what I'm doing wrong - I pulled this branch and reinstalled the pre-commit hook (`pre-commit install` in the PlasmaPy directory), but I still get missing executable errors for `ruby` and `echo`. I've tried installing pre-commit and setting the hook both on my local python environment and in my virtualenv with the same result. Am I missing something about how to set this up? 

@JaydenR2305 what did you do to install ruby? `pip install ruby` manually does install something for me, but pre-commit apparently still can't find it.

namurphy:
@pheuer — thank you for trying this out!  I'll take a look into pre-commit's documentation to try to figure out what's happening (or not happening) here. 

Here are the instructions for [installing Ruby](https://www.ruby-lang.org/en/documentation/installation/).

pheuer:
@namurphy  Thanks! Installing Ruby helped (after also restarting my PC a couple times). 

I'm still getting another error, ironically in the hook that prints a troubleshooting message. It looks like it can't find the 'echo' command, although 'echo' seems to be working fine on my terminal? 

image

namurphy:
Thank you, @pheuer!  

I'm less familiar with Windows so I'm not sure how to handle this.  It may have something to do with [this line](https://github.com/PlasmaPy/PlasmaPy/blob/e9e3a439d3f0d3f3872dbd320d68ae9c6ca05160/.pre-commit-config.yaml#L114) where it defines `language: system`.  Hm...

Could you take what you just posted and create a new issue with it?  If you click on the "..." button for the comment, there should be an option to "Reference in new issue".

pheuer:
@namurphy Where are we at for this? I got pre-commit working by installing Ruby 3.3.0, but ideally that wouldn't be required (or else it would be more clearly in our docs). Should we merge this fix and see what happens with it in the future?

namurphy:
I'm not 100% sure that this PR will fix the problem, but at the very least, I don't think it'll do any harm.  We can give it a go, and revisit the problem later if we run into it again.  Thank you for the reminder!

PR #2458, opened 2024-01-11
Improve type hint annotations to `plasmapy.particles`

namurphy:
This PR is part of the process of adding & correcting type hints in `plasmapy.particles`, as a follow-up to my prior attempt in #2429.

I started this PR by using MonkeyType to generate type hints based on the types of variables while running pytest. I also made some manual updates following this.


plasmapy/particles/particle_class.py
```diff
@@ -2,6 +2,13 @@
 
 from __future__ import annotations
 
+from io import StringIO
+from typing import Any, Optional, Union
+
+from astropy.units.quantity import Quantity
+
+from plasmapy.particles.particle_collections import ParticleList
```
namurphy, line 10: Ah. This automatically added line seems to be what's causing problems.  Need to add an `if typing.TYPE_CHECKING` style flag in here.  Probably.

plasmapy/particles/_factory.py
```diff
@@ -92,26 +97,34 @@ def _make_custom_particle_with_real_charge_number(
...
+] = (
     Particle,
     CustomParticle,
     CustomParticle._from_quantities,  # noqa: SLF001
     ParticleList,
     _make_custom_particle_with_real_charge_number,
 )
 
-_particle_types = (Particle, CustomParticle, ParticleList)
+_particle_types: tuple[type, ...] = (Particle, CustomParticle, ParticleList)
```
namurphy, line 122: The `...` in `tuple[type, ...]` indicates that there can be any number of objects in the `tuple`.  A type hint of `tuple[type]` would have indicated that the `tuple` should only contain a single object of type `type`.

plasmapy/particles/_factory.py
```diff
@@ -92,26 +97,34 @@ def _make_custom_particle_with_real_charge_number(
...
 
 
 def _physical_particle_factory(
-    *args, **kwargs
+    *args: ParticleLike | u.Quantity | CustomParticle | Sequence[ParticleLike],
```
namurphy, line 126: I thought about creating a type alias variable for all of this since it gets repeated, but if we make `_physical_particle_factory` into a public function, the type alias variable would need to also be public so that the documentation build doesn't break.

plasmapy/particles/_factory.py
ejohnson-96: I didn't know that

PR #2462, opened 2024-01-15
Move link to pre-commit troubleshooting guide

namurphy:
[description to be added later]


pheuer:
@namurphy Thanks - confirmed that this resolves the pre-commit error on my Windows PC

.pre-commit-config.yaml
```diff
@@ -51,15 +54,11 @@ repos:
     # run this hook on `labeler.yml` (or certain other files).
     exclude: .github/labeler.yml|.pre-commit-search-and-replace.yaml
 
-- repo: https://github.com/pre-commit/mirrors-csslint
-  rev: v1.0.5
-  hooks:
-  - id: csslint
```
namurphy, line 57: It is quite rare for us to modify a CSS file, so I'm dropping this.  But, we should remember that `csslint` exists for when we do need it one of these years.

namurphy:
I expanded this PR a fair bit, mostly to rename hooks to be more accessible to new contributors and add links to the ruff rules and say how to deal with false positives in codespell.  

> Question: was this `echo` hook added recently? If so, I wonder why it only became an issue for me recently. I did re-install PlasmaPy on a new virtual environment in the newest version of Anaconda.

@pheuer — We added that hook relatively recently, perhaps a few months ago.  Thank you again for reporting the issue!

.pre-commit-config.yaml
ejohnson-96: I'm assuming CSS here is not cascading style sheets?

namurphy: It is indeed cascading style sheets.  We have a couple of CSS files for our website, for better or for worse.

ejohnson-96: Argh I see, I have some previous experience working with web development and hosting. If you need something looking at ever I could probably lend a hand

Issue #1877, opened 2023-01-09
Use a `source` layout in `docs`


namurphy:
I was playing around with Sphinx the other year, and it's now the case that when you initialize a Sphinx documentation project, it'll create a `source` folder that contains all of the `.rst` files for the content.  I was thinking we could switch our docs to this `source` layout, since it would separate files that are not content (like `docs/conf.py`) with ones that are (like `docs/index.rst`).  It would make it a little harder to get to the content in the docs, but easier to navigate through the content once you're in `docs/source`.  

If we did this, we'd need to update the doc guide as well with probably a bunch of little changes.  We'd also need to decide where some of the files should go (i.e., should `bibliography.bib` stay in `docs` or go to `docs/source`)?

I wanted to get feedback on this before putting any effort into it, but overall I do think it would be an improvement (and make our docs more consistent with newer Sphinx projects).

PR #2463, opened 2024-01-16
Modernize `docs/Makefile` with help from `sphinx-quickstart`

namurphy:
The `Makefile` we use for our Sphinx documentation build was autogenerated many years ago by `sphinx-quickstart`.  This PR updates it to be closer to what `sphinx-quickstart` provides nowadays.

One thing to note: we currently do not have separate `docs/source` and `docs/build` directories, which now seems to be the preferred way to organize our documentation.  That's discussed in #1877.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,4 @@ untitled*
 .mypy_cache
 node_modules
 package-lock.json
+monkeytype.sqlite3
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -21,7 +21,7 @@ html-nonb: Makefile
 	@$(SPHINXBUILD) -M html "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) -D nbsphinx_execute='never' $(O)
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html. Jupyter notebooks were skipped."
 
-# Catch-all target: route all unknown targets to Sphinx using the new
+# Catch-all target: route all unknown targets to Sphinx using the
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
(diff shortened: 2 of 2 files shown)


docs/Makefile
```diff
@@ -18,15 +17,3 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
-
-clean:
-	rm -rf $(BUILDDIR)/*
-	rm -rf api cache gen_modules __pycache
```
namurphy, line 24: We don't need to add this separately.

docs/Makefile
```diff
@@ -18,15 +17,3 @@ help:
...
-
-clean-api:
-	rm -rf api
```
namurphy, line 27: I don't think this is a defined environment within Sphinx, but we could add this back later if we so choose.

.gitignore
```diff
@@ -149,3 +149,4 @@ untitled*
 .mypy_cache
 node_modules
 package-lock.json
+monkeytype.sqlite3
```
namurphy, line 152: This is a file that MonkeyType creates to store a database of types, for example when running pytest.

docs/Makefile
```diff
@@ -1,32 +1,27 @@
...
-
 clean-api:
 	rm -rf api
 
 html-nonb: Makefile
 	@echo $(SPHINXBUILD) -M html "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) -D nbsphinx_execute='never' $(O)
 	@$(SPHINXBUILD) -M html "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) -D nbsphinx_execute='never' $(O)
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html. Jupyter notebooks were skipped."
+
+# Catch-all target: route all unknown targets to Sphinx using the new
```
namurphy, line 24:
```suggestion
# Catch-all target: route all unknown targets to Sphinx using the
```

PR #2470, opened 2024-01-18
Revise changelog entries prior to v2024.1.0

namurphy:
Some minor updates to changelog entries.  Part of #2467.


namurphy:
I just added two changelog entries to try something out.  Occasionally I've noticed that starting the second line of a changelog entry with a backtick can cause an error in the documentation build.  I added two news fragments to test this.

PR #2477, opened 2024-01-25
Add new hyperlink redirects and add links to docstring of documentation configuration file

namurphy:
There was an update in `sphinx-issues` which set the default for the `:user:` role to point to `github.com/sponsors/username` instead of `github.com/username` (https://github.com/sloria/sphinx-issues/issues/129, https://github.com/sloria/sphinx-issues/pull/131). This led to a number of redirects to show up when we did a linkcheck build, so I added a line to `docs/conf.py` to specify that these redirects should be allowed when doing a linkcheck.  

I added some links to the docstring for `docs/conf.py` that point to helpful locations like our documentation guide and Sphinx's documentation.

I also updated the GitHub username of a past contributor.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -416,7 +416,7 @@ authors:
   family-names: Rajashekar
   affiliation: PES University
   orcid: https://orcid.org/0000-0002-4914-6612
-  alias: DarkAEther
+  alias: vrajashkr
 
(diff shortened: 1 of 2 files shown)


ejohnson-96 (approved):
👍

PR #2478, opened 2024-01-25
Revise testing environment for performing linkchecks of documentation builds

namurphy:
- I renamed the `tox` environment from `build_docs_linkcheck` to `linkcheck` since that's what I kept trying to use.
 - I moved the (incredibly helpful!) `-q` flag from the GitHub Action to the `tox` environment. This way, it is in place when we run linkcheck through `tox` locally. The new setup makes it so that only errors and warnings show up in the output of doing `tox -e linkcheck`.
 - I updated the documentation guide to mention `tox -e linkcheck`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,14 +63,13 @@ commands =
     echo "Troubleshooting guide: https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting"
 deps = -r{toxinidir}/requirements.txt
 
-[testenv:build_docs_linkcheck]
+[testenv:linkcheck]
 changedir = {toxinidir}
 setenv =
     HOME = {envtmpdir}
     PYDEVD_DISABLE_FILE_VALIDATION=1
 commands =
-    sphinx-build docs docs{/}_build{/}html -W -n --keep-going -b linkcheck {posargs}
-    cat docs/_build/html/output.txt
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -63,14 +63,13 @@ commands =
     echo "Troubleshooting guide: https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting"
 deps = -r{toxinidir}/requirements.txt
 
-[testenv:build_docs_linkcheck]
+[testenv:linkcheck]
```
namurphy, line 66: I settled on this name since I kept typing `tox -e linkcheck` even after getting multiple reminders that this wasn't the name of an environment.  🙃

tox.ini
```diff
@@ -63,14 +63,13 @@ commands =
...
 changedir = {toxinidir}
 setenv =
     HOME = {envtmpdir}
     PYDEVD_DISABLE_FILE_VALIDATION=1
 commands =
-    sphinx-build docs docs{/}_build{/}html -W -n --keep-going -b linkcheck {posargs}
-    cat docs/_build/html/output.txt
+    sphinx-build docs docs{/}_build{/}html -W -n -q --keep-going -b linkcheck {posargs}
```
namurphy, line 72: The `-q` flag makes it so that only broken links and redirects show up in the linkcheck output.  Without it, the results for every link that's checked gets outputted to the screen.

ejohnson-96 (approved):
👍

PR #2479, opened 2024-01-26
Make `@particle_input` compatible with `from __future__ import annotations`

namurphy:
## Description

@JaydenR2305 found a bug that `@particle_input` did not work on an instance method when preceded by `from __future__ import annotations`, which delays the evaluation of annotations.  The purpose of this PR is to fix that bug.  

However, at the moment I don't actually know how to fix it, so I'm going to start with writing some purposefully failing tests that reproduce the bug, and then try to get it to work somehow.

Wish me luck!  🙃 

## Related issues

Bug originally observed in #2245.


plasmapy/particles/decorators.py
```diff
@@ -9,7 +9,7 @@
 from collections.abc import Callable, Iterable, MutableMapping
 from inspect import BoundArguments
 from numbers import Integral, Real
-from typing import Any, Optional, TypedDict, Union
+from typing import Any, Optional, TypedDict, Union, get_type_hints
```
namurphy, line 12: ChatGPT suggested using `typing.get_type_hints` because other methods returned the string representations of the annotations rather than the actual objects.  Probably saved me an hour or two of work...

namurphy:
This PR should make `@particle_input` compatible for when postponed evaluation of annotations becomes default (PEP 563, PEP 649).  Probably.  Hopefully.

plasmapy/particles/decorators.py
ejohnson-96: ChatGPT :scream:

ejohnson-96 (approved):
What an annoying little bug :+1:

namurphy:
> What an annoying little bug 👍

If postponed evaluation of annotations had become the default in a future version of Python before this was fixed, this would have been an annoying little bug that broke the entire package! 🙃

namurphy:
@JaydenR2305 — thank you for finding this bug in #2245!  It should be fixed once this is merged.

PR #2482, opened 2024-01-31
Update requirements for building docs

namurphy:
## Description

This PR updates the `tests` and `docs` requirements sets as defined in `pyproject.toml` mostly to the most recent releases (excluding pytest, see #2480).  It also does a `pre-commit autoupdate`.

## Motivation

With PlasmaPy's regular dependencies, it is helpful to support a range of versions so as to reduce the chances of our dependencies conflicting with the dependencies of other packages. In contrast, it is generally helpful for the developers of a package to use the same or similar environments for testing and documentation builds. This is especially the case when there are breaking changes (i.e., going from pytest 7.4 → 8.0).


namurphy:
pre-commit.ci autofix

plasmapy/particles/decorators.py
```diff
@@ -744,8 +744,7 @@ def particle_input(
           class SomeClass:
               @particle_input
               @validate_quantities
-              def instance_method(self, particle: ParticleLike, B: u.Quantity[u.T]):
-                  ...
+              def instance_method(self, particle: ParticleLike, B: u.Quantity[u.T]): ...
```
namurphy, line 747: When I run the `blacken-docs` hook for `pre-commit` locally on my computer, I get the `...` on the following line. When I do `pre-commit.ci autofix`, it puts them on the same line.  Hm...

tox.ini
```diff
@@ -109,7 +109,7 @@ conda_deps =
     matplotlib >= 3.5.1
     mpmath >= 1.2.1
     numpy >= 1.21.0
-    pytest >= 5.4.0
+    pytest == 7.4.4
```
namurphy, line 112: Will remove this strict requirement in #2480.

ejohnson-96 (approved):
:+1:

PR #2484, opened 2024-02-01
Specify version of `black` in `pre-commit` hook for `blacken-docs` 

namurphy:
While we mostly switched over to the `ruff` formatter, we still use `black` via `blacken-docs` to autoformat code in documentation.  There was a breaking change in the most recent release, so this PR specifies the version of `black` used by `blacken-docs`.


namurphy:
pre-commit.ci autofix

PR #2487, opened 2024-02-01
Make `synthetic_radiograph.Tracker` deterministic by adding a random seed keyword to the `create_particles` method

pheuer:
@namurphy If this looks good to you, let's try to get it merged in - quickly. @JaydenR2305 is about to refactor this `synthetic_radiography` module using his new `Tracker` class, so I'd like to have this fix in when he starts that. 

## Description
1) Small refactoring to use new `numpy.Generator` functions
2) Added `random_seed` keyword to `create_particles` that, when set, makes the method deterministic 

## Motivation and context
Should resolve #1458 (knock on wood). 

## Related issues
#1458

PR #2488, opened 2024-02-01
Bump minimum versions of dependencies

namurphy:
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#) provides a recommended community schedule for supporting old versions of Python and NumPy.  Continuing to support older versions of packages helps reduce the chances of dependency conflicts, but with an increased maintenance burden and inability to use new features. NEP 29 provides a community standard to balance these competing needs.   

NEP 29 states that minor releases of NumPy within the last two years should be supported by packages across the broader scientific pythoniverse. Correspondingly, I bumped the minimum version of NumPy to 1.23.0.  Dropping support for older versions of NumPy will be beneficial because a lot of the stuff for type hint annotations has been improving in the last several years. 

While I was at it, I dropped support for releases older than two years old for a bunch of our other dependencies.


namurphy:
@pheuer — If these changes look good to you, please feel free to merge and see if they help with #2487.  Even if the changes don't help, it's probably worth it to stay in alignment with NEP 29 so as to reduce the chances of encountering bugs related only to older versions of dependencies.

ejohnson-96 (approved):
👍

PR #2492, opened 2024-02-03
Fix linkcheck GitHub Action along with minor updates to workflows

namurphy:
#2490 ended up disabling the workflow dispatch option for doing a linkcheck.  This PR reverts some of #2490 so that the linkcheck workflow does not attempt to run in PRs, and instead adds a workflow step for CI that does the linkcheck.

I also took a moment to make the formatting of workflow files a bit more consistent, as well as some other minor updates.


namurphy:
Merging this so I can test that it works.

PR #2493, opened 2024-02-04
Update synthetic radiography `Tracker` class to use `ParticleTracker`

JaydenR2305:
In #2245 we implemented a new particle tracker class that was meant to generalize the already written radiography tracker. This pull request partially rewrites the synthetic radiography tracker to use this new particle tracker class.


plasmapy/simulation/particle_tracker.py
```diff
@@ -858,6 +858,9 @@ def _push(self) -> None:
             .T
         )
 
+        # Don't count untracked particles as being on the grid
+        self.on_grid[~tracked_mask] = False
```
JaydenR2305, line 862: While testing the new stopping condition for proton radiography I discovered that this behavior wasn't already implemented

JaydenR2305:
TODO:

- [ ] Write new save routine specific to proton radiography
- [ ] "Generalize" the proton radiography termination condition and move it to the particle tracker module
- [ ] Implement new synthetic radiography test case that compares against a deterministic result of the tracker
- [ ] Improve documentation about improvements to `_coast_to_grid` and `_coast_to_plane` methods
- [ ] Convert `_tracked_particle_mask` to a property???

JaydenR2305:
image
Looks like it's working now 🎉

plasmapy/simulation/particle_tracker.py
```diff
@@ -252,6 +252,10 @@ def post_push_hook(self, force_save=False) -> None:
         if self.save_now or force_save:
             self.save()
 
+    def post_run_hook(self) -> None:
+        """Function called after the simulation has finished running."""
+        return
```
JaydenR2305, line 257: This is the best way I could come up with to create an "optional" abstract method (i.e. a method that is guaranteed to be implemented but whose implementation doesn't necessarily have to come from the child class)

changelog/2493.trivial.rst
```diff
@@ -0,0 +1 @@
+Implement |ParticleTracker| in `~plasmapy.diagnostics.charged_particle_radiography.synthetic_radiography.Tracker`
```
pheuer, line 1: "Refactor __ to be a subclass of ___"

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,89 @@ def total(self) -> float:
         return float(self.tracker.nparticles)
 
 
+class ParticlesPushedThroughGridsTerminationCondition(AbstractTerminationCondition):
```
pheuer, line 173: "AllParticlesOffGridTerminationCondition"? Since all these are for particles being pushed through a grid?

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,89 @@ def total(self) -> float:
...
+    """Termination condition corresponding to stopping the simulation when a provided
```
pheuer, line 174:
```suggestion
    """Termination condition corresponding to stopping the simulation when a specified
```

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,89 @@ def total(self) -> float:
...
+    proportion of particles have entered and exited the grids.
```
pheuer, line 175:
```suggestion
    fraction of particles have entered and exited the grids.
```

to match the language in the keywords 'fraction'

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,89 @@ def total(self) -> float:
...
+
+    Parameters
+    ----------
+    fraction_entered_threshold: float, optional
```
pheuer, line 179: So, the purpose of this line in the original tracker 

`if self.fract_entered <= 0.1`

Was just to make sure the tracker didn't terminate on the very first step (if no particle had yet traveled onto the grid enough to count as being on the grid). 

Maybe it's simpler to just remove this keyword and below make the condition 

```python
if self._particle_tracker.num_entered > 0 and self.fraction_exited_threshold < still_on:
    return True
```

So the termination condition basically turns on after the first particle gets to the grid? Then only one user-specified keyword "fraction_exited_threshold" should be enough.

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,89 @@ def total(self) -> float:
...
+        if self._particle_tracker.num_entered > 0:
+            # Normalize to the number that have entered a grid
+            still_on = (
+                np.sum(self._particle_tracker.on_any_grid)
+                / self._particle_tracker.num_entered
+            )
+        else:
+            still_on = 0.0
+
+        if (
```
pheuer, line 231: See above, also `fraction_exited_threshold` either needs a default or to be required - otherwise this terminates immediately?

pheuer (commented):
Initial review of the save routine, changelog

pheuer (commented):
Initial minor review of Tracker

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -70,7 +73,37 @@ def _coerce_to_cartesian_si(pos):
...
+
+    @property
+    def require_synchronized_dt(self) -> bool:
+        return False
+
+    @property
+    def save_now(self) -> bool:
+        return False
+
+    def save(self) -> None:
```
pheuer, line 90: If this is a subclass of AbstractSaveRoutine, then shouldn't it really overwrite `_save_to_disk` and `_save_to_memory`, and leave `save` alone? 

Maybe we should have included a more extendable way of saving quantities in `AbstractSaveRoutine`, say a dict `self.results = {}` For the abstract class, that would have keys `x` and `v`. For the charged particle radiograph class, you could extend it to also save these other variables. Then, `_save_to_disk` would just always save whatever was in the dictionary to the HDF5 file?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -127,22 +160,35 @@ def __init__(
         grids: AbstractGrid | Iterable[AbstractGrid],
         source: u.Quantity[u.m],
         detector: u.Quantity[u.m],
+        dt=None,
```
pheuer, line 163: Update docstring for these new parameters

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -127,22 +160,35 @@ def __init__(
...
+        save_routine = (
+            _SyntheticRadiographySaveRoutine(output_file)
+            if output_file is not None
+            else None
+        )
 
-        # self.grid_arr is the grid positions in si units. This is created here
-        # so that it isn't continuously called later
-        self.grids_arr = [grid.grid.to(u.m).value for grid in self.grids]
+        termination_condition = AllParticlesOffGridTerminationCondition(
```
pheuer, line 179: Good to make this a default, but maybe still keep this a keyword, or make `fraction_exited_threshold` a keyword? In some cases I think I've had to fiddle with that parameter to get things to run?

pheuer:
In `_apply_wire_mesh()`, particles are being removed like this: 

```python
 self.x = self.x[keep_these_particles, :]
 self.v = self.v[keep_these_particles, :]
```
Maybe this should change to use the `remove` functionality in ParticleTracker?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -820,16 +747,20 @@ def _coast_to_grid(self) -> None:
...
         t = dist / vmax
 
         # Coast the particles to the advanced position
-        self.x = self.x + self.v * t[:, np.newaxis]
+        self.x[tracked_mask] = (
+            self.x[tracked_mask] + self.v[tracked_mask] * t[:, np.newaxis]
+        )
 
-    def _coast_to_plane(self, center, hdir, vdir, x=None):
+    def _coast_to_plane(self, center, hdir, vdir, x=None, mask=None):
```
pheuer, line 763: Can you add parameter descriptions to this docstring? Especially for mask, to just to make that clear

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1110,62 +882,43 @@ def run(
...
-        self.fract_tracked = self.nparticles_grid / self.nparticles
-
-        # Create flags for tracking when particles during the simulation
-        # on_grid -> zero if the particle is off grid, 1
-        # shape [nparticles, ngrids]
-        self.on_grid = np.zeros([self.nparticles_grid, self.num_grids])
-        # Entered grid -> non-zero if particle EVER entered a grid
-        self.entered_grid = np.zeros([self.nparticles_grid])
+        # Determine which particles should be removed
+        # These particles will not be pushed through the fields
```
pheuer, line 886:
```suggestion
        # These particles will not be pushed through the fields but instead will be automatically advanced 
        # to the detector plane.
```

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1110,62 +882,43 @@ def run(
...
-        self.nparticles_grid = len(self.grid_ind)
-        self.fract_tracked = self.nparticles_grid / self.nparticles
-
-        # Create flags for tracking when particles during the simulation
-        # on_grid -> zero if the particle is off grid, 1
-        # shape [nparticles, ngrids]
-        self.on_grid = np.zeros([self.nparticles_grid, self.num_grids])
-        # Entered grid -> non-zero if particle EVER entered a grid
-        self.entered_grid = np.zeros([self.nparticles_grid])
+        # Determine which particles should be removed
```
pheuer, line 885: I think this comment is in the wrong place?

```suggestion
```

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1110,62 +882,43 @@ def run(
...
+            )
 
         # Remove particles that will never reach the detector
         self._remove_deflected_particles()
 
         # Advance the particles to the image plane
         self._coast_to_plane(self.detector, self.det_hdir, self.det_vdir, x=self.x)
 
-        # Log a summary of the run
+        self.save_routine.post_push_hook(force_save=True)
```
pheuer, line 919: I still think here it would be OK to just call `save` explicitly? What is the post_push_hook method doing for us now?

plasmapy/diagnostics/charged_particle_radiography/tests/test_synthetic_radiography.py
```diff
@@ -646,30 +674,30 @@ def test_optical_density_histogram(self) -> None:
...
-    path = tmp_path / "temp.npz"
-    sim.save_results(path)
-
-    # Load result
-    results_2 = dict(np.load(path, "r", allow_pickle=True))
-
-    assert set(results_1.keys()) == set(results_2.keys())
-    for key in results_1:
-        assert np.allclose(results_1[key], results_2[key])
+# @pytest.mark.slow()
```
pheuer, line 677: This test pertains to the old save behavior, right? You can just delete it (better to not leave commented out code in main)

JaydenR2305:
Ok so now all save routines defined in simulation/particle_tracker have a `results` property that defines the relevant results for the simulation. This dictionary is updated by the `_save_to_memory` method which is called by the `save` method regardless of whether `output_directory` has been specified. This means we would implement a `SaveOnceOnCompletition` class by overriding the `save_now` method. A save method for the synthetic radiography simulation would be a child class of `SaveOnceOnCompletition` but also extend the `_save_to_memory` method to include magnification, initial positions, etc.

Does this all sound good to you @pheuer ?

JaydenR2305:
image

Any idea what's going on here??

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
JaydenR2305: So maybe something like having to specify either `termination_condition` or `fraction_exited_threshold`?

plasmapy/simulation/particle_tracker.py
```diff
@@ -219,37 +290,81 @@ def save_now(self) -> bool:
...
-        self.v_all.append(np.copy(self._particle_tracker.v))
+        """Update the results dictionary with the current state of the simulation."""
+        # """Append simulation positions and velocities to save routine object."""
+        # self.x_all.append(np.copy(self._particle_tracker.x))
+        # self.v_all.append(np.copy(self._particle_tracker.v))
+
+        self._results["x"].append(np.copy(self._particle_tracker.x))
+        self._results["v"].append(np.copy(self._particle_tracker.v))
+
+    def _apply_units_to_results(self):
```
pheuer, line 343: The one thing I don't like about this approach is that, when you add more things to `results` in a subclass you now have to updated two methods: `_save_to_memory` and `_apply_units_to_results`, which seems like an easy place for bugs to creep in. I do see the issue with units, though. 

What about defining what is saved as a list of tuples attribute of the save class: 

```python
_quantities = [
('x', u.m, 'dataset'),
('v', u.m/u.s, 'dataset'),
...
('magnification', u.dimensionless_unscaled, 'attribute'),
]
```

Then, when adding new things to save, you just overwrite his list and the none of the other classes should need to get changed?

plasmapy/simulation/particle_tracker.py
```diff
@@ -169,6 +170,75 @@ def total(self) -> float:
         return float(self.tracker.nparticles)
 
 
+class AllParticlesOffGridTerminationCondition(AbstractTerminationCondition):
```
pheuer, line 173: There's kind of a lot of these already, and I expect we may add more in the future! To head off a potential future refactor, what if we moved these files into different modules now: 

simulation/particle_tracker/particle_tracker.py 
simulation/particle_tracker/termination_conditions.py 
simulation/particle_tracker/save_routines.py

plasmapy/simulation/particle_tracker.py
```diff
@@ -219,37 +290,81 @@ def save_now(self) -> bool:
...
     def _save_to_memory(self) -> None:
-        """Append simulation positions and velocities to save routine object."""
-        self.x_all.append(np.copy(self._particle_tracker.x))
-        self.v_all.append(np.copy(self._particle_tracker.v))
+        """Update the results dictionary with the current state of the simulation."""
+        # """Append simulation positions and velocities to save routine object."""
+        # self.x_all.append(np.copy(self._particle_tracker.x))
+        # self.v_all.append(np.copy(self._particle_tracker.v))
+
+        self._results["x"].append(np.copy(self._particle_tracker.x))
```
pheuer, line 340: Results is understandably a dict of lists, since we are appending to them as the simulation runs. But, when the user accesses this property, I think it would be more convenient for plotting etc. to return a dict of Quantities? I think your unit application already does this? 

Interesting: I didn't know that list * quantity = quantity without having to do the conversion list -> np.ndarray first.

pheuer:
> Ok so now all save routines defined in simulation/particle_tracker have a `results` property that defines the relevant results for the simulation. This dictionary is updated by the `_save_to_memory` method which is called by the `save` method regardless of whether `output_directory` has been specified. This means we would implement a `SaveOnceOnCompletion` class by overriding the `save_now` method to only save when `final_save == True` . A save method for the synthetic radiography simulation would be a child class of `SaveOnceOnCompletion` but also extend the `_save_to_memory` method to include magnification, initial positions, etc.
> 
> edit: I have changed the results dictionary to `_results`. The `results` property will now return `_results` but with proper units applied :)
> 
> Does this all sound good to you @pheuer ?

Mostly! See comments above about a possible better way to handle adding quantities to save and associating them with units, dataset/attribute. The rest sounds good!

plasmapy/simulation/particle_tracker.py
JaydenR2305: Would `_save_to_memory` then assume that the quantities of interest can be found just by doing something like

```py
getattr(self._particle_tracker, value[0])
```
where value would be some element of the `_quantities` list?

pheuer: Yes, exactly

JaydenR2305:
image
image

Is this a bug with ruff?

plasmapy/simulation/particle_tracker.py
JaydenR2305: Yep this is what I ahve:
```py

    @property
    def results(self) -> dict[str, u.Quantity]:
        """Return the results of the simulation.
        The quantities returned are the times, positions, and velocities, respectively.
        """
        return self._apply_units_to_results()
   ```

Is that the behavior you were looking for?

Definitely a plus that astropy handles the list to array conversion!

File: plasmapy/simulation/particle_tracker.py (hunk 18 above)
Comment (JaydenR2305) [reply plasmapy/simulation/particle_tracker.py:173]: How should we handle the namespace? Should all of this still be accessible with something like
```py
from plasmapy.simulation.particle_tracker import AbstractSaveRoutine
```
or were you thinking
```py
from plasmapy.simulation.particle_tracker.save_routines import AbstractSaveRoutine
```

pheuer: I think the latter? Separate documentation pages would be good so it's not too cluttered on the main particle_tracker page.

pheuer:
> Is this a bug with ruff?

Not sure what's best here - in general it would be good to not define the variable if it isn't necessary (e.g. an unnecessary enumerate) but here you do need one half of the tuple and defining `_units` instead of just `_` does seem cleaner?

pheuer:
A couple good ideas from a conversation with @JaydenR2305 ! 

GitHub apparently generates an API token 'GITHUB_TOKEN' at the beginning of every workflow job. If we can access this in `test_downloader.py`, we could pass it into an optional keyword in `Downloader` and make authenticated API calls! Those have a limit of 5,000/hr, which we would absolutely not exceed at the rate of <~5 calls per commit. All I need to do is 

1) Figure out how to get this variable in `test_downloader.py`. Might be as simple as running `os.environ['GITHUB_TOKEN`]`? 
2) Add a keyword for an API token to Downloader, which might be nice to have anyway. 

Another backup idea we had was to create a `RateLimitError` exception that is raised in `test_downloader.py` if the API becomes rate limited and causes the tests to automatically pass in that case. Since I expect this to be rare (only occur if several people are pushing commits all at once) this would still lead to the test running most of the time. 

I vastly prefer the first idea though, so I'll look into that!

JaydenR2305:
@pheuer Any idea why it doesn't like `|ParticleTracker|` ?

pheuer:
> @pheuer Any idea why it doesn't like `|ParticleTracker|` ?

Yes, so those shortcuts are defined somewhere and I always forget where! Might be in the contributor guide? OR @namurphy will know! 

Anyway, since you moved that file the link is now pointing to something that doesn't exist.

JaydenR2305:
I think they're defined in global substitutions? I changed the definition there to 
```py
"ParticleTracker": "`~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker`",
```

pheuer:
> I think they're defined in global substitutions? I changed the definition there to
> 
> ```python
> "ParticleTracker": "`~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker`",
> ```

Oh, you're right! Except I think you also need to specify that it's a class, so it looks like it should be 

"ParticleTracker": ":class:`~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker`",

JaydenR2305:
Why is the CI doc build failing but read the docs is fine?

pheuer:
> Why is the CI doc build failing but read the docs is fine?

Ok, now I'm also very confused: the links to `|ParticleTracker|` are working in readthedocs too! But not in the documentation build... 

By the way, I think we can remove this in `particle_tracker.rst`: 

"""
This module is highly unstable and is expected to change a lot in the
future.
"""

pheuer:
Any other @PlasmaPy/peer-reviewers have any idea what's going wrong with the documentation build here?

plasmapy/simulation/particle_tracker/save_routines.py
```diff
@@ -0,0 +1,215 @@
...
+        self._quantities = {
+            "x": (u.m, "dataset"),
+            "v": (u.m / u.s, "dataset"),
+        }
+
+        self._particle_tracker = None
+
+    @property
+    def tracker(self):
+        """Return the `ParticleTracker` object for this stop condition."""
```
ejohnson-96, line 53: I think the problem is here is `ParticleTracker` set up in the reference file? I believe the "WARNING: py:obj" error in building the docs is trying to refer to the `ParticleTracker` and it doesn't exist?

pheuer: Ah, I see! So even without the tilde
```
`ParticleTracker`
```
Is a reference to an object by that name, which is no longer in the same file! The error should go away if we did a string literal 

```
``ParticleTracker``
```

But it would be better to add `|ParticleTracker|` back to the `_global_substitutions` file, and then just use that throughout all of the docstrings. 

Does that make sense @JaydenR2305 ?

plasmapy/simulation/particle_tracker/save_routines.py
```diff
@@ -0,0 +1,215 @@
...
+    After every push, the `post_push_hook` method is called with the
+    respective `~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker` object passed as a parameter.
+    Then, the hook calls `save_now` to determine whether or not the simulation state should be saved.
+    """
+
+    def __init__(self, output_directory: Path | None = None) -> None:
+        self.output_directory = output_directory
+
+        self._results = {}
+        self._quantities = {
```
pheuer, line 44: Save time in AbstractSave: should be a shape[ntimes] array if the timestep is synchronized, otherwise a shape[ntimes,nparticles]. I think this should be the default since without time you can't really plot x,v. In subclasses, like the charged particle radiography subclass, you can overwrite it and not save time as it makes sense.

plasmapy/simulation/particle_tracker/save_routines.py
```diff
@@ -0,0 +1,215 @@
...
+            self._save_to_disk()
+
+    def _save_to_disk(self) -> None:
+        """Save a hdf5 file containing simulation positions and velocities."""
+
+        path = self.output_directory / f"{self.tracker.iteration_number}.hdf5"
+
+        with h5py.File(path, "w") as output_file:
+            for key, (_units, data_type) in self._quantities.items():  # noqa: B007
+                match data_type:
```
pheuer, line 98: Wow, somehow I completely missed that Python added this switch statement architecture in 3.10...

docs/simulation/particle_tracker/save_routines.rst
```diff
@@ -0,0 +1,10 @@
+.. _save_routines:
+
+******************************
+Particle Tracker Save Routines
+******************************
+
```
pheuer, line 6:
```suggestion
Save routines tell a |ParticleTracker| when to save information about the particles. Information can be saved either to memory (accessible as a dictionary after the simulation is complete) 
```

docs/simulation/particle_tracker/termination_conditions.rst
```diff
@@ -0,0 +1,10 @@
+.. _termination_conditions:
+
+***************************************
+Particle Tracker Termination Conditions
+***************************************
+
```
pheuer, line 6:
```suggestion
Termination conditions tell a |ParticleTracker| simulation when to stop.
```

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -70,7 +73,37 @@ def _coerce_to_cartesian_si(pos):
     return pos_out
 
 
-class Tracker:
+class _SyntheticRadiographySaveRoutine(AbstractSaveRoutine):
```
pheuer, line 76: I think this class needs to be updated to match the changes in `AbstractSaveRoutine`.

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -70,7 +73,37 @@ def _coerce_to_cartesian_si(pos):
...
+    ARRAY_KEYS = ["x", "y", "v", "x0", "y0", "v0"]
```
pheuer, line 77: Should save everything in the Tracker's `results_dict`

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: The tracker sort of assumes this termination condition, so I would make `fraction_exited_threshold` a keyword. The current value of 0.001 is a good default.

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
JaydenR2305: Do you mean that everything inside of `results_dict` should be saved to the HDF file-- say as an attribute by default?

JaydenR2305:
@namurphy Hey Nick we're having trouble merging main into this PR to match with the new package structure. Do you have any recommendations with how we should proceed?

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
JaydenR2305: I *think* it should be good now. It works locally, at least.

JaydenR2305:
https://github.com/PlasmaPy/PlasmaPy/blob/9a27fd47b2de46133d754af46182675c7ab794cf/plasmapy/simulation/particle_tracker/particle_tracker.py#L590C1-L601C1

@pheuer What are your thoughts on converting the `on_grid` to a function definition like this?

pheuer:
> https://github.com/PlasmaPy/PlasmaPy/blob/9a27fd47b2de46133d754af46182675c7ab794cf/plasmapy/simulation/particle_tracker/particle_tracker.py#L590-L601
> 
> @pheuer What are your thoughts on converting the `on_grid` to a function definition like this?

Do you mean as a property? Sure that's a good idea!

plasmapy/simulation/particle_tracker/save_routines.py
JaydenR2305: Ok I just added "time" to all of the `_quantities` lists for the save routines. By default it will follow this behavior since that's how the underlying `ParticleTracker` keeps track of time anyways (one number for synchronized timestep, otherwise an array with entries for each particle)

plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Yep - for the charged particle radiography case we want everything in that dict

PR #2501, opened 2024-02-06
Drop python 3.9 support

namurphy:
## Description

This PR drops support for Python 3.9 and bumps the minimum required version of Python to 3.10.

This PR also applies some autofixes from `ruff`, including:
 - Changes like `isinstance(..., (int, float))` → `int | float`
 - Explicitly using the `strict` keyword in calls to `zip`.

After this PR, we can enable the `UP006` & `UP007` rules from ruff which will automagically update our annotations to use the clean new syntax.  Didn't want to do that here since it's big enough as it is, and there may be some nuances with `Optional[int]` vs. `int | None` and how it's handled with our decorators.

## Motivation and context

[NumPy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) lays out a recommended community standard for supporting old versions of Python and NumPy.  NEP 29 balances the need for a package to work with older versions of Python with the ability to use new language features.  As per this schedule, Python 3.9 no longer needs to be supported in releases after April 4, 2024.  We just released `v2024.2.0`, and it's likely that our next release will be in May or June.

[New features in Python 3.10](https://docs.python.org/3/whatsnew/3.10.html) that we can now use include:

 - Structural pattern matching (i.e., `match` statements)
 - Length checking in `zip` statements with the `strict` keyword argument
 - Parenthesized context managers
 - Writing union types as `X | Y` ← which I've been wanting to start using for the last three years!

## Related issues

#987, #1262, #1796


namurphy:
pre-commit.ci autofix

docs/notebooks/analysis/fit_functions.ipynb
```diff
@@ -291,7 +291,7 @@
     "plt.legend(fontsize=14, loc=\"upper left\")\n",
     "\n",
     "txt = f\"$f(x) = {explin.latex_str}$\\n$r^2 = {explin.rsq:.3f}$\\n\"\n",
-    "for name, param, err in zip(explin.param_names, explin.params, explin.param_errors):\n",
+    "for name, param, err in zip(explin.param_names, explin.params, explin.param_errors, strict=False):\n",
```
namurphy, line 294: The `strict=False` flag for `zip` allows the zipped iterables to have different lengths. In general, it will probably be better to set `strict=True` for most of the time that we use `zip`.  However, I'm not doing that here, since it will take a bit of work to think about each usage of `zip` to make sure it is desirable to set `strict=True`.  I plan to raise an issue about `strict=True` after this is merged.

namurphy:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -214,7 +214,7 @@ type_incorrect = "Incorrect changelog type (see `changelog/README.rst`)"
 type_incorrect_long = "The filename of the changelog entry must be of the form `changelog/NUMBER.TYPE.rst` where `TYPE` is one of `feature`, `trivial`, `doc`, `bugfix`, `breaking`, or `removal`.\n\nFor details, see: https://docs.plasmapy.org/en/latest/contributing/changelog_guide.html#adding-a-changelog-entry"
```
ejohnson-96, line 215: The background colour of the text here. Is it significant or just denoting something?

plasmapy/utils/roman.py
```diff
@@ -94,7 +94,7 @@ def to_roman(n: Union[Integral, np.integer]) -> str:
     'MMDXXV'
 
     """
-    if not isinstance(n, (Integral, np.integer)):
+    if not isinstance(n, Integral | np.integer):
         raise TypeError(f"{n} cannot be converted to a Roman numeral.")
     if not (0 < n < 5000):
         raise OutOfRangeError("Number is out of range (need 0 < n < 5000)")
```
ejohnson-96, line 100: I didn't know we had this functionality, but I am so happy we do

docs/notebooks/analysis/fit_functions.ipynb
ejohnson-96: Would it be easier to set strict to False as the default in Zip for the time being then? Else wont all these strict=False statements have to be altered later down the line?

ejohnson-96 (approved):
Couldn't see any instances that were incorrectly altered. I do like the new | variable denotation, helps reduce the number of brackets you are keeping track off.

pyproject.toml
namurphy: I guess it's how GitHub now renders TOML files now?

docs/notebooks/analysis/fit_functions.ipynb
namurphy: This came from ruff rule [`B905`](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) which requires an explicit `strict=True` or `strict=False` in calls to `zip`. I decided against changing current behavior in this PR, and keeping `strict=False` will make the code more explicit in what it's actually doing.  You're right though that most (all?) of the `strict=False` statements will need to be altered later on.

PR #2504, opened 2024-02-06
Apply ruff rules `UP006` and `UP007`

### LINKED ISSUE #2502: Enable the ruff rules for upgrading type hint annotations to use new syntax
Enabling `UP006` & `UP007` rules in `ruff` will upgrade our type hint annotations to use the new syntax.  For example, `typing.Union[int, float]` will be changed to `int | float`.  

I'm assigning myself to this since there are probably going to be some nuances in `@particle_input` and maybe `@validate_quantities`, such as whether `int | None` will get treated the same as `typing.Optional[int]`, which may cause some headaches.

namurphy:
I updated the `ruff` configuration to enable rules [`UP006`](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) and [`UP007`](https://docs.astral.sh/ruff/rules/non-pep604-annotation/).  

All of the changes were associated with `UP007`, which changed the `typing.Union[X, Y]` syntax to `X | Y`. This PR is a follow-up to #2501 which dropped Python 3.9 support. This syntax was introduced in Python 3.10.

In the definition of `ParticleLike` and `ParticleListLike`, I kept the `typing.Union` syntax in the definition, since we weren't able to redefine `ParticleLike.__doc__` when using the `X | Y` syntax.

I had been worried that `@particle_input` would handle annotations like `Optional[ParticleLike]` and `ParticleLike | None` differently, but it worked out okay.

Closes #2502.

### DIFF
--- a/plasmapy/particles/decorators.py
+++ b/plasmapy/particles/decorators.py
@@ -9,7 +9,7 @@ import warnings
 from collections.abc import Callable, Iterable, MutableMapping
 from inspect import BoundArguments
 from numbers import Integral, Real
-from typing import Any, TypedDict, get_type_hints
+from typing import Any, TypeAlias, TypedDict, get_type_hints
 
 import numpy as np
 import wrapt
@@ -40,7 +40,7 @@ class _CallableDataDict(TypedDict, total=False):
     signature: inspect.Signature
 
 
-_basic_particle_input_annotations = (
+_basic_particle_input_annotations: tuple[type | TypeAlias, ...] = (
     Particle,  # deprecated
     ParticleLike,
     ParticleListLike,
@@ -49,7 +49,9 @@ _basic_particle_input_annotations = (
 )
 _optional_particle_input_annotations = tuple(
     annotation | None
-    for annotation in _basic_particle_input_annotations
+    # remove [:-1] index in following line when dropping (Particle, Particle)
+    # as a valid annotation
+    for annotation in _basic_particle_input_annotations[:-1]
     if annotation != (Particle, Particle)  # temporary hack
 )
 _particle_input_annotations = (
--- a/plasmapy/particles/decorators.py
+++ b/plasmapy/particles/decorators.py
@@ -32,11 +32,11 @@ class _CallableDataDict(TypedDict, total=False):
     allow_custom_particles: bool
     allow_particle_lists: bool
     annotations: dict[str, Any]
-    any_of: Optional[Union[str, Iterable[str]]]
+    any_of: str | Iterable[str] | None
     callable_: Callable[..., Any]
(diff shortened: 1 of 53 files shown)


namurphy:
I've been wanting to make these changes for approximately 39 months! 🙃

plasmapy/particles/decorators.py
```diff
@@ -32,32 +32,34 @@ class _CallableDataDict(TypedDict, total=False):
...
+    ParticleLike | ParticleListLike,
     (Particle, Particle),  # deprecated
 )
 _optional_particle_input_annotations = tuple(
-    Optional[annotation]
-    for annotation in _basic_particle_input_annotations
+    annotation | None
+    # remove [:-1] index in following line when dropping (Particle, Particle)
+    # as a valid annotation
+    for annotation in _basic_particle_input_annotations[:-1]
```
namurphy, line 54: This includes some manual additions to this PR. I had intended to put it in a different PR, but got branches mixed up.  I got a mypy error around here.

plasmapy/particles/particle_class.py
```diff
@@ -395,10 +395,10 @@ class Particle(AbstractPhysicalParticle):
         integer representing the atomic number of an element; or a
         |Particle|.
 
-    mass_numb : integer, |keyword-only|, optional
+    mass_numb : int, |keyword-only|, optional
         The mass number of an isotope.
 
-    Z : integer, |keyword-only|, optional
+    Z : int, |keyword-only|, optional
```
namurphy, line 401: The mypy cheat sheet recommends using `int` for situations like this.

plasmapy/particles/particle_class.py
```diff
@@ -1465,7 +1465,7 @@ def electron_number(self) -> Integral:
             raise InvalidIonError(_category_errmsg(self, "ion"))
 
     @property
-    def isotopic_abundance(self) -> Real:
+    def isotopic_abundance(self) -> float:
```
namurphy, line 1468: Similarly, the mypy cheat sheet recommends using `float` rather than `numbers.Real`.

plasmapy/particles/particle_class.py
```diff
@@ -2494,7 +2492,7 @@ def molecule(
 # If ParticleLike is renamed or moves out of particle_class.py, check
 # for a link to its doc page in error messages in _factory.py.
 
-ParticleLike = Union[str, Integral, Particle, CustomParticle, u.Quantity]
+ParticleLike: TypeAlias = Union[str, Integral, Particle, CustomParticle, u.Quantity]  # noqa: UP007
```
namurphy, line 2495: Needed to keep this as using `typing.Union` because using the `|` made it so that we couldn't redefine `ParticleLike.__doc__` below.

PR #2505, opened 2024-02-06
Update `@particle_input` docstring to use type union syntax instead of `Optional`

namurphy:
Prior to #2504, optional arguments to functions decorated with `@particle_input` had annotations like `typing.Optional[ParticleLike]`. Since PlasmaPy is now Python 3.10+, we can instead use annotations like `ParticleLike | None` or `ParticleListLike | None`. This PR updates the docstring of `@particle_input` to reflect that.

Follows up on #2501, #2502, #2504.


ejohnson-96 (approved):
👍

PR #2506, opened 2024-02-06
Make `@validate_quantities` compatible with `from __future__ import annotations`

namurphy:
Currently, `@validate_quantities` does not work in files when there is a `from __future__ import annotations` statement which postpones the evaluation of annotations. The purpose of this PR is to change that. 

With postponed evaluation of annotations, the annotations become strings. For `@particle_input`, we were able to fix this by using `typing.get_type_hints`.

I'm starting with a failing test to get a better idea of where this goes wrong.  

Closes #2492. Similar to #2479.


namurphy:
It turns out this is more difficult than `@particle_input` and I'm not sure yet how to fix this.  Hm...

namurphy:
> I think `wrapped_sign = inspect.signature(f)` -> `wrapped_sign = inspect.signature(f, eval_string=True)` does the trick :) @namurphy

@JaydenR2305 — Great job figuring this out!  I had been pretty lost on how to address this, so thank you!  I made sure to include you as a co-author on the commit.

PR #2509, opened 2024-02-08
Parametrize tests of `Debye_length`

namurphy:
PlasmaPy's tests of `Debye_length` are currently consolidated into a single test function with multiple assert statements.  This PR changes the test setup to use `@pytest.mark.parametrize`. 

I've been meaning to get to this for a couple of years, but better late than never!


ejohnson-96 (approved):
👍

PR #2512, opened 2024-02-11
Fix typo in gyroradius formula

Bzero:
## Description

This pull request fixes a small typo in the non-relativistic gyroradius formula.


namurphy (approved):
Good catch!  Thank you for finding and fixing this!

PR #2516, opened 2024-02-14
Add known limitations to docstring of `@particle_input`

namurphy:
`@particle_input` is able to handle most of the functions we use it with, but there are some edge cases and situations where it can behave unexpectedly. This PR adds a "Notes" section to its docstring that describes issues that can arise when using `@particle_input`.

The tradeoff is that the docstring for `@particle_input` is longer than it already was, but it's still important to have this documented...especially since it may take a while before we address these limitations.


ejohnson-96 (approved):
👍

PR #2518, opened 2024-02-15
Add a "See Also" to the `formulary.radiation` docs on `astropy.modeling.BlackBody`

namurphy:
Rather than adding our own function for blackbody radiation, this PR will point users to Astropy's.


ejohnson-96 (approved):
👍

PR #2523, opened 2024-02-22
Use `uv` to regenerate pinned requirements to be used in CI for multiple versions of Python

namurphy:
## Description

 - Changed from using `pip-tools` to [`uv`](https://github.com/astral-sh/uv) for creating pinned dependencies
 - Created a `tox` environment called `requirements` that is used to rebuild the requirements files. This environment creates now creates multiple requirements files: one for each supported Python version for tests, one for minimal dependencies with the oldest Python, and for building docs with the most recent Python.
 - Updated tox environments to run tests with the pinned dependencies for all Python environments
 - Made it so that doctests are only being run in Python 3.12 in CI
 - Reinstating a requirement for `sphinx == 7.2.6` because an error that I thought had gone away is still there.

`requirements.txt` as well as the files in `requirements/` are auto-generated, and don't need a close review.

## Motivation and context

`uv` is a new (mostly) drop-in replacement for `pip` and `pip-tools`. Because it is written in Rust with improved caching and dependency resolution algorithms, it has significantly better performance.

`uv` also has some new features. The `--resolution=lowest` option for the `uv pip compile` command will let us create a requirements file for our minimal dependencies. We have previously been updating that environment in `tox.ini` manually, which has sometimes led to inconsistencies for this between `tox.ini` and `pyproject.toml`.  `uv` can also resolve dependencies for arbitrary versions of Python, which makes it easier to generate requirements files for all supported versions of Python.


namurphy:
I just realized something...maybe.  I set up caching of environments for our CI using the `pip` option. However, the environments used to run tests and build docs are created using `tox`.  Hence, it may be that the caching doesn't include the actual environments run during tests.  Yeesh...

If I don't deal with that in this PR, I will:

 - [ ] Create an issue about ↑

changelog/2523.internal.rst
namurphy, line 1: - [ ] Update this

.github/workflows/update-pinned-reqs.yml
```diff
@@ -23,14 +23,8 @@ jobs:
...
-    - name: Upgrade pip
-      run: python -m pip install --upgrade pip
-
-    - name: Install pip-tools
-      run: python -m pip install pip-tools
-
-    - name: Generate requirements.txt
-      run: pip-compile --output-file=requirements.txt --rebuild --upgrade --all-extras
+    - name: Rebuild requirements
+      run: tox -e requirements
```
namurphy, line 27: Creating a tox environment for rebuilding requirements makes it a lot more straightforward to rebuild requirements locally with `tox -e requirements` in the same way that is done in CI.

.github/workflows/weekly-tests.yml
```diff
@@ -20,17 +20,22 @@ jobs:
         - name: Python 3.12 (macOS)
           os: macos-latest
           python: '3.12'
-          toxenv: py312-all
+          toxenv: py312-pins-all-doctest
```
namurphy, line 23: This PR makes it so that doctests in PRs are only run in Python 3.12 with pinned dependencies.  This has the features that:
 - When there is a breaking change in fundamental constants or how arrays or quantities are displayed, we won't have to try to make our docstrings compatible with both the older and newer versions of our dependencies.
 - The code coverage metric will reflect how much the test suite itself covers lines of code, without contributions from doctests. This will make it less likely that we'll forget to include tests for new functionality.

.github/workflows/weekly-tests.yml
```diff
@@ -20,17 +20,22 @@ jobs:
...
         - name: Python 3.10 (Windows)
           os: windows-latest
           python: '3.10'
-          toxenv: py310-all
+          toxenv: py310-pins-all
+
+        - name: Python 3.10 (Windows), minimal dependencies
+          os: windows-latest
+          python: '3.10'
+          toxenv: py310-pins-minimal-all
```
namurphy, line 38: Turns out our CI tests were not being run on our minimal dependencies from changes I made recently.  Oops!

pyproject.toml
```diff
@@ -120,15 +120,13 @@ namespaces = false
 write_to = "plasmapy/_version.py"
 
 [tool.pytest.ini_options]
-minversion = "7.0.1"
```
namurphy, line 123: ...since it's defined elsewhere in this file.

pyproject.toml
```diff
@@ -120,15 +120,13 @@ namespaces = false
...
 testpaths = ['plasmapy']
 norecursedirs = ['build', 'docs', 'plasmapy[\/]_dev']
 xfail_strict = true
 doctest_optionflags = """
 NORMALIZE_WHITESPACE
 ELLIPSIS
 NUMBER"""
-addopts = "--doctest-modules --doctest-continue-on-failure"
```
namurphy, line 131: The flags for running doctests are now defined in `tox.ini`.

requirements.txt
namurphy, line 1: `requirements.txt` and the files in `requirements/` are auto-generated, and don't need a close review.

tox.ini
```diff
@@ -14,7 +14,8 @@ setenv =
     COLUMNS = 180
     PIP_INDEX_URL = {env:PIP_INDEX_URL:https://pypi.anaconda.org/scipy-wheels-nightly/simple}
     PIP_EXTRA_INDEX_URL = {env:PIP_EXTRA_INDEX_URL:https://pypi.org/simple}
-    PYTEST_COMMAND = pytest --pyargs plasmapy --durations=5 -n=auto --dist=loadfile
+    PYTEST_COMMAND = pytest --pyargs plasmapy --durations=5 -n=auto --dist=loadfile --tb=short
```
namurphy, line 17: The `--tb=short` makes it so that the traceback reports are shorter.  Without it, the default is to print the entire function for the first and last error, which can make it harder to navigate pytest output.

ejohnson-96 (approved):
👍

namurphy:
Thank you as always for the quick review!

PR #2524, opened 2024-02-22
Refactor script for checking `CITATION.cff`

namurphy:
This PR does some refactoring of the script that checks whether each author of a pull request is included in `CITATION.cff`.  

This PR also fixes an error in the link to edit `CITATION.cff`.  The link had previously pointed to the `PlasmaPy/PlasmaPy` repo instead of the fork.  It does still assume that the fork is named "PlasmaPy" which will be true most of the time but possibly not all.

### DIFF
--- a/.github/scripts/authors_in_cff.py
+++ b/.github/scripts/authors_in_cff.py
@@ -82,7 +82,6 @@ persistent identifier used by researchers, authors, and open source
 contributors. Sign up at: https://orcid.org/register
 
 Thank you for contributing!
-
 """
 
     logging.info(instructions_to_add_author)
--- a/.github/scripts/authors_in_cff.py
+++ b/.github/scripts/authors_in_cff.py
@@ -34,7 +34,11 @@ def find_missing_github_usernames(authors: list[str]) -> list[str]:
     """Verify that all authors of a PR are included in :file:`CITATION.cff`."""
     with pathlib.Path("CITATION.cff").open() as file:
         lines = file.read()
-        return [author for author in authors if f"alias: {author}" not in lines]
+        return [
+            author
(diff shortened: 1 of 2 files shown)


.github/scripts/authors_in_cff.py
```diff
@@ -23,56 +26,66 @@ def get_pr_authors() -> set[str]:
...
+https://github.com/{username}/PlasmaPy/edit/{branch}/CITATION.cff
+
+Each entry should be of the form:
 
 - given-names: <given names>
   family-names: <family names>
   affiliation: <affiliation>
   orcid: https://orcid.org/<ORCiD number>
-  alias: {missing_github_username}
-  email: <email address>
```
namurphy, line 57: We don't have a need for people's emails, so we can remove this. Plus, very few people have added the email field to their `CITATION.cff` entry, except for the most frequent contributors.

.github/scripts/authors_in_cff.py
```diff
@@ -23,56 +26,66 @@ def get_pr_authors() -> set[str]:
...
+To ensure that you get credit for your contribution, please add the
+following authors to CITATION.cff: {", ".join(missing_github_usernames)!r}
 
-        error_message = f"""
-To ensure that you get credit for your contribution to PlasmaPy, please
-add {missing_github_username!r} as an author to CITATION.cff.
+This file can be edited at:
 
-The entry should be of the form:
+https://github.com/{username}/PlasmaPy/edit/{branch}/CITATION.cff
```
namurphy, line 68: This link was incorrect before since it pointed to `PlasmaPy/PlasmaPy` instead of the fork.

PR #2525, opened 2024-02-23
Loosen requirements for `pytest` and `sphinx` and drop `pytest` as runtime dependency

### LINKED ISSUE #2497: Remove limit on version of Sphinx
We just started getting an `ImportError` related to `plasmapy_sphinx` in our weekly test with the development version of Sphinx:  
```
Extension error:
Could not import extension plasmapy_sphinx (exception: cannot import name 'PyField' from 'sphinx.domains.python' (/home/runner/work/PlasmaPy/PlasmaPy/.tox/build_docs-sphinxdev/lib/python3.12/site-packages/sphinx/domains/python/__init__.py))
build_docs-sphinxdev: exit 2 (7.26 seconds) /home/runner/work/PlasmaPy/PlasmaPy> sphinx-build docs docs/_build/html -W -n --keep-going -b html -q pid=2260
```

I think the change happened in either [this commit](https://github.com/sphinx-doc/sphinx/commit/e9dcfebcf9f3cb5db6a47448e10e9b61c5abc416) or [this other commit](https://github.com/sphinx-doc/sphinx/commit/d491d3e04c0b95ea9603a82a5fb8cb83c1e192f1).

In #2498, the requirement got hard-coded to `sphinx == 7.2.6` so that this doesn't end up surprising us in the coming months.  I decided against fixing it now because the change hasn't been released yet.

namurphy:
* Removed pytest as a runtime dependency because it is only needed for test helper functionality which is now private and only used when running our tests. `pytest` is now a `tests` dependency.
* Updated requirement to `pytest >= 8.0.1`.  There were some breaking changes in v8.0.0 which required some fixes. In making this upgrade, we had to update where there were overlapping `with pytest.warns` and `with pytest.raises` blocks (i.e., situations where both a warning was issued and an exception was raised).  It seems that when an exception is now raised, the warning isn't captured by pytest.
* Update requirement to `sphinx >= 7.2.6`.  Closes #2497. We had previously started getting an error in `plasmapy_sphinx` when doing building the docs with the development version of Sphinx, but that error appears to have gone away.


pyproject.toml
```diff
@@ -48,7 +48,7 @@ dependencies = [
   "numpy >= 1.23.0",
   "packaging >= 22.0",
   "pandas >= 1.4.0",
-  "pytest == 7.4.4",
+  "pytest >= 7.0.0",
```
namurphy, line 51: `pytest` is one of our runtime dependencies because of (now private) test helper functionality, so it will be helpful to loosen the requirement for installation.

namurphy:
There's a small change that removing pytest as a runtime dependency will break something later on, but so far it appears to be working fine.

ejohnson-96 (approved):
👍

PR #2526, opened 2024-02-23
Drop the `py310-conda` tox environment

namurphy:
For a while we've had a test environment for conda with Python 3.10.  However, we haven't used it in CI for quite a while.  It's also one more place where we have to specify requirements.  Since we're not using it, it also has the potential to become outdated.  This PR removes this unused environment.


ejohnson-96 (approved):
👍

PR #2533, opened 2024-02-23
Resolve deprecation warnings in tests

namurphy:
This PR reduces the number of `DeprecationWarnings` that show up in our tests.

### DIFF
--- a/plasmapy/analysis/nullpoint.py
+++ b/plasmapy/analysis/nullpoint.py
@@ -483,7 +483,7 @@ def _trilinear_jacobian(vspace, cell):
                 return v
             if len(v) != 1:
                 raise ValueError("Unable to get scalar from multi-element array.")
-            return v[0]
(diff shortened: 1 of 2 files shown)


plasmapy/analysis/nullpoint.py
```diff
@@ -473,17 +473,29 @@ def jacobian_func(xInput, yInput, zInput):
...
+
+        def get_scalar(v: float | np.ndarray):
+            """
+            If a `float`, return the argument. If a `numpy.ndarray` with
+            a single element, return that element.
+            """
+            if not hasattr(v, "__len__"):
+                return v
+            if len(v) != 1:
+                raise ValueError("Unable to get scalar from multi-element array.")
```
namurphy, line 485: I put this block in primarily as an extra safeguard, though it may not actually be needed here.

Issue #2435, opened 2023-12-30
Get `@validate_quantities` to pass mypy strict mode


namurphy:
We use `@validate_quantities` throughout the package, so it should be a high priority to get it to work with mypy.  In particular, we should get it to work with mypy's strict mode. 

See also #2434.

PR #2543, opened 2024-02-27
Update type hints and docstrings in `plasmapy.formulary`

namurphy:
This PR adds and updates a bunch of type hint annotations in `plasmapy.formulary`. It also makes a few minor refactoring changes.

A limiting factor in doing this is that `@validate_quantities` is currently largely unannotated (#2435), but is being used to decorate a whole bunch of different functions. However, I'm going to leave that for another day (or more likely, another month).

### DIFF
--- a/plasmapy/formulary/collisions/frequencies.py
+++ b/plasmapy/formulary/collisions/frequencies.py
@@ -18,7 +18,7 @@ from astropy.constants.si import e, k_B, m_e
 from plasmapy import particles
 from plasmapy.formulary.collisions import coulomb, lengths, misc
 from plasmapy.formulary.speeds import thermal_speed
-from plasmapy.particles import ParticleLike
+from plasmapy.particles.particle_class import ParticleLike
 from plasmapy.utils.decorators import deprecated, validate_quantities
 from plasmapy.utils.exceptions import PhysicsError, PlasmaPyFutureWarning
 
--- a/plasmapy/formulary/collisions/lengths.py
+++ b/plasmapy/formulary/collisions/lengths.py
@@ -8,20 +8,21 @@ import astropy.units as u
 import numpy as np
 from astropy.constants.si import eps0, hbar
 
-from plasmapy import particles
 from plasmapy.formulary.collisions import frequencies, misc
 from plasmapy.formulary.lengths import Debye_length
 from plasmapy.formulary.quantum import Wigner_Seitz_radius
+from plasmapy.particles.decorators import particle_input
+from plasmapy.particles.particle_class import Particle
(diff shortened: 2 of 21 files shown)


plasmapy/formulary/collisions/lengths.py
```diff
@@ -8,20 +8,21 @@
...
 
 @validate_quantities(
     T={"can_be_negative": False, "equivalencies": u.temperature_energy()}
 )
-@particles.particle_input
+@particle_input
 def impact_parameter_perp(
     T: u.Quantity[u.K],
-    species: (particles.Particle, particles.Particle),
+    species: (Particle, Particle),
```
namurphy, line 25: To be more consistent with the import conventions we use elsewhere.

plasmapy/formulary/dielectric.py
```diff
@@ -3,17 +3,22 @@
     "cold_plasma_permittivity_SDP",
     "cold_plasma_permittivity_LRP",
     "permittivity_1D_Maxwellian",
+    "RotatingTensorElements",
+    "StixTensorElements",
```
namurphy, line 7: I needed to expose these to the public API so that they could be used as return annotations without causing a "missing reference" error when doing the doc build.

plasmapy/formulary/collisions/frequencies.py
```diff
@@ -18,7 +18,7 @@
 from plasmapy import particles
 from plasmapy.formulary.collisions import coulomb, lengths, misc
 from plasmapy.formulary.speeds import thermal_speed
-from plasmapy.particles import ParticleLike
+from plasmapy.particles.particle_class import ParticleLike
```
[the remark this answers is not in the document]
ejohnson-96: I think I remember this issue

plasmapy/formulary/collisions/lengths.py
ejohnson-96: I love it when a review is just basically import statements, it makes the review a lot easier 😆

plasmapy/formulary/dielectric.py
ejohnson-96: Could you explain this when we next talk please?

ejohnson-96 (approved):
👍

plasmapy/formulary/collisions/lengths.py
namurphy: It makes the pull request a lot easier too!

plasmapy/formulary/dielectric.py
namurphy: Would be happy to, with the caveat that I only understand 83% of what's happening here!

PR #2546, opened 2024-02-29
Create universal wheel in tox environments

namurphy:
This PR follows the advice in this blog post on [turbocharging tox](https://hynek.me/articles/turbo-charge-tox/).

> ```toml
> [testenv]
> package = wheel
> wheel_build_env = .pkg
> ```
>
> The first line tells tox to build and install wheels instead of source distributions, and the second line tells it to share the same build environment – and thus wheel – across all tox environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,10 @@
 [tox]
 envlist = clean,py312,build_docs
-
 isolated_build = True
 
 [testenv]
+package = wheel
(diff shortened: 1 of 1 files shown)


namurphy:
When I tested this locally, it greatly sped up the stages before running tests...at least the second time I ran it.  It's a small enough change, so I'll go ahead and merge it.

PR #2551, opened 2024-02-29
Update dependabot settings

namurphy:
An unintended consequence of the creation of a pinned requirements file for minimal dependencies in #2523 is that dependabot has begun creating pull requests to update the outdated versions of dependencies in that file.  

The purpose of this PR is to get dependabot to not do those updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ version: 2
 updates:
 
 - package-ecosystem: github-actions
-  directory: /
+  directory: /.github/workflows
   labels:
   - GitHub Actions
   - no changelog entry needed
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,11 +3,11 @@ version: 2
 updates:
 
 - package-ecosystem: github-actions
-  directory: /
+  directory: /.github/workflows
```
namurphy, line 6: My thinking is that this will make dependabot skip the `requirements` directory. Don't know if this will work though...

namurphy:
Not sure if this will suffice, but I'm going to attempt it.

Issue #2258, opened 2023-07-28
Speed up documentation builds done via GitHub Actions in CI by caching previous build


namurphy:
Right now our documentation builds currently take ∼10–15 minutes.  A consequence of this is that we often have to push a minor change to a reStructuredText file to try to resolve a warning in the documentation, and then wait ∼10—15 minutes to see if it goes through.

It would be really helpful if we could speed up the documentation build in CI.  One possibility would be to have a GitHub Action that would cache the output of the previous documentation build, and then rebuild the documentation from that.  Doing this would allow Sphinx to only regenerate the documentation pages that correspond to files that have changed, which should greatly speed things up.

There are a few things we would need to be careful about. For example, there will be occasions where we would want to do a fresh rebuild of everything, like possibly when we edit configuration stuff in `docs/conf.py` and want to make sure that everything gets updated.  For that, we could potentially create a label that would make Sphinx rebuild the docs from scratch, or manually remove the cached build. It's also possible that errors could be hidden by rebuilding the docs from scratch.  The RTD build should be able to take care of that, in particular if we require that the RTD build doesn't have any warnings.

PR #2552, opened 2024-02-29
Speed up tests by caching `.tox` in GitHub Actions and update tox environments

namurphy:
Continuous integration tests are most beneficial when they provide rapid 🚀 feedback. One of the time sinks in tests has been that `tox` has had to recreate Python environments every time tests are run.  

This PR enables caching for our test suite and documentation build via GitHub Actions. **Caching speeds up our Python 3.12 tests in CI (which skip slow tests but include doctests) from ∼3.5 min to ≲ 1 min!** We save the ∼2.5 minutes by not having to always recreate Python environments.

I renamed tox environments to make them easier to keep track of (at least for me). In particular, I reduced the extent to which `tox.ini` makes use of [generative environments](https://tox.wiki/en/4.13.0/user_guide.html#generative-environments) when specifying the commands for pytest to run. This is less clever, but probably a bit more maintainable.

I also did a bit of cleanup of `tox.ini` and GitHub Actions.

## Related issues

I attemped to take care of #2258 here by caching `docs/_build`, `docs/api`, and `docs/about/_authors.rst`, but I got a weird doc build error that I didn't attempt to figure out.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,7 +65,7 @@ jobs:
       run: python -m pip install --progress-bar off --upgrade tox
 
     - name: Run tests
-      run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
+      run: tox -e ${{ matrix.toxenv }}
 
     - name: Upload coverage to codecov if requested
       if: ${{ contains(matrix.toxenv,'-cov') }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,6 +83,7 @@ jobs:
 
     - name: Save cache
       uses: actions/cache/save@v3
+      if: steps.restore-cache.outputs.cache-hit != 'true'
       with:
         path: |
           .tox
@@ -115,26 +116,19 @@ jobs:
     - name: Restore cache
       uses: actions/cache/restore@v3
       id: restore-doc-cache
-      if: ${{ github.event_name == 'pull_request' }}
       with:
         path: |
-          docs/api
-          docs/_build
-          docs/about/_authors.rst
           .tox
-        key: ${{ runner.os }}-${{ matrix.toxenv }}-${{ hashFiles('requirements.txt', 'docs/conf.py') }}-${{ github.event.number }}
+        key: ${{ runner.os }}-build_docs_pins-${{ hashFiles('requirements.txt') }}
 
     - name: Build docs
       run: tox -e build_docs_pins -- -q
 
-    - name: Save cache for pull requests
+    - name: Save cache
       uses: actions/cache/save@v3
-      if: ${{ github.event_name == 'pull_request' }}
+      if: steps.restore-doc-cache.outputs.cache-hit != 'true'
       with:
         path: |
-          docs/api
-          docs/_build
-          docs/about/_authors.rst
           .tox
         key: ${{ steps.restore-doc-cache.outputs.cache-primary-key }}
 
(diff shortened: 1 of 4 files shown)


namurphy:
pre-commit.ci autofix

tox.ini
```diff
@@ -12,26 +12,27 @@ allowlist_externals =
...
 commands =
-    !all: {env:PYTEST_COMMAND} {posargs} -m 'not slow'
-    all: {env:PYTEST_COMMAND} {posargs}
-    cov-all: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}{/}pyproject.toml --cov-append --cov-report xml:coverage.xml
-    doctest: {env:PYTEST_COMMAND} {posargs} --doctest-modules --doctest-continue-on-failure
+    pytest_all: {env:PYTEST_COMMAND} {posargs}
+    pytest_skipslow: {env:PYTEST_COMMAND} {posargs} -m "not slow"
+    pytest_doctests_all: {env:PYTEST_COMMAND} {posargs} --doctest-modules --doctest-continue-on-failure
+    pytest_doctests_skipslow: {env:PYTEST_COMMAND} {posargs} --doctest-modules --doctest-continue-on-failure -m "not slow"
+    pytest_cov_all: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}{/}pyproject.toml --cov-append --cov-report xml:coverage.xml
```
namurphy, line 34: A problem with the previous setup was that having both `!all` and `all` made it so that pytest ended up being run twice in some environments.

I wasn't able to find a good explanation in the tox documentation about how the `command` field depended on the names of environments for the generative whatnot, so I went with having unique strings and including pytest in it.

This is making me want to switch to `nox`...

.github/workflows/tests.yml
```diff
@@ -26,18 +26,17 @@ jobs:
         - name: Python 3.12 (Ubuntu), with doctests, skip slow
           os: ubuntu-latest
           python: '3.12'
-          toxenv: py312-pins-doctest
-          toxargs: --develop
```
namurphy, line 30: I had thought that the `--develop` flag here was to enable Python development mode, but I think this is just passed to tox.

.github/workflows/tests.yml
```diff
@@ -62,13 +61,20 @@ jobs:
...
-    - name: Install and upgrade tox
+    - name: Install and upgrade test tools
       run: python -m pip install --progress-bar off --upgrade tox
 
+    - name: Cache
+      uses: actions/cache@v4
+      with:
+        path: |
+          .tox
+        key: ${{ matrix.toxenv }}-${{ runner.os }}-${{ hashFiles('requirements.txt', 'pyproject.toml') }}
```
namurphy, line 72: By creating a hash from these two files and including it in the cache name, we can make it so that the cache gets regenerated whenever `requirements.txt` or `pyproject.toml` change.

.github/workflows/tests.yml
```diff
@@ -62,13 +61,20 @@ jobs:
...
+
     - name: Run tests
-      run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
```
namurphy, line 69: We weren't using `toxargs` or `toxposargs` except for the `--develop` flag, but that's been removed.  We can instead define flags to pass to pytest, etc. in the `tox` environments themselves.

namurphy:
pre-commit.ci autofix

ejohnson-96 (approved):
👍

namurphy:
Thank you for the review!  I'm happy with how this PR turned out, but it is safe to say that dealing with both `tox.ini` and GitHub Actions at once is not the most fun experience in the universe.  But, getting quicker feedback in CI will be worth it.

PR #2555, opened 2024-03-04
Implement NIST STAR Data

### LINKED ISSUE #2517: Implement NIST STAR (Stopping Power & Ranging) calculations in PlasmaPy
### Feature description

Implement NIST STAR (Stopping Power & Ranging) calculations in PlasmaPy, something similar to the calculations in the NIST PSTAR website: https://physics.nist.gov/PhysRefData/Star/Text/PSTAR.html 

Something like 

`stopping_power =StoppingPower(test_particle, material, energies)`

Ideally this would work for as many materials as possible (not just elements): maybe there is a way to combine stopping_powers from elements to approximate a compound? 

### Motivation

The NIST PSTAR database data is currently used in the charged_particle_radiography.detector_stacks package (currently awkwardly being imported by users from the NIST website), and soon we would like to use it in charged_particle_radiography.Tracker. 

Having this data directly in PlasmaPy would be much more direct for users! 

### Implementation strategy

Start with an existing but unmaintained repository for this: 
https://github.com/Zelenyy/nist-calculators/tree/master

EDIT: This package doesn't actually do the calculations, it has all the NIST data downloaded into this HDF5 file and then just interpolates from that. That would also be a fine way to do it... 

The HDF5 is 2.2 MB, but we could put it in the plasmapy_data repo and then use the downloader to download it when called? 

### Additional context

Details on physics behind the NIST databases: 
https://physics.nist.gov/PhysRefData/Star/Text/programs.html 

Project for @JaydenR2305 ?

JaydenR2305:
Implements the ability to calculate stopping power for alpha particles and protons in a specified material using the NIST ASTAR and PSTAR databases, respectively. The data for this pull request can be found [here](https://github.com/PlasmaPy/PlasmaPy-data/blob/main/NIST_STAR.hdf5)

Resolves #2517


JaydenR2305:
@namurphy I wasn't quite sure where to put this function.. Where do you think it should go?

JaydenR2305:
image
Just to ensure that the pulled hdf5 file has accurate data, here is data I pulled directly from the PSTAR website plotted with data from the hdf5 file. This particular plot is for the total stopping power of hydrogen

ejohnson-96 (approved):
👍

JaydenR2305:
Going to hold off on this until #2570 is merged

pheuer:
> Going to hold off on this until #2570 is merged

Agreed - let me know what you think about the class structure in that PR!

pheuer:
> Going to hold off on this until #2570 is merged

#2570 is now merged!

pheuer:
@JaydenR2305 I took a closer look at this file (had to use stopping powers for a calculation today) and realized just how many elements the NIST database doesn't actually include! We can (and should) go ahead and get this merged in, but a future project would be filling in some of the missing data (especially important filter quantities like protons in Tantalum) using output from SRIM calculations.

pheuer:
> @pheuer Do you know what's causing these ReadTimeout errors??

Hm, does the test pass locally? The file/url it is trying to reach is definitely valid? 

One thing you might want to do to limit test time/number of API queries is to use an unvalidated downloader in the tests (except for maybe one, that confirms the file is accessible). See example of how I did that here: https://github.com/PlasmaPy/PlasmaPy/blob/534b8e960050488db4c9e2977bc4142388afc153/tests/utils/data/test_downloader.py#L11

But, that shouldn't explain the read timeout, since even in the unvalidated mode `downloader` still has to download from the `raw` github URL...

JaydenR2305:
@pheuer What are your thoughts on implicitly setting the API token if it is detected that the code is being run in a CI environment? That is, if an API key is not explicitly passed during instantiation then ``os.environ["GH_TOKEN"]`` would be used.

pheuer:
> @pheuer What are your thoughts on implicitly setting the API token if it is detected that the code is being run in a CI environment? That is, if an API key is not explicitly passed during instantiation then `os.environ["GH_TOKEN"]` would be used.
> 
> This way we don't need to complicate things by introducing optional keywords for `Downloader` objects and remembering that we have to pass one in every time we write tests involving the downloader

Hm, that's a good idea for now! It's not exactly equivalent to unvalidated (still makes queries to find the file I think) but that's ok. 

Make sure you include a comment in the test explaining what you're doing.

JaydenR2305:
How should I add coverage for these lines? 

https://github.com/PlasmaPy/PlasmaPy/blob/fe4ed92f5b9eb30b19e26ec6a4f7efcbc0a662f8/src/plasmapy/particles/atomic.py#L1238-L1241

The returned result can have some ~100 entries.

JaydenR2305:
also why does this have partial coverage? It looks like the code under the elseif statement is being executed fine? 

https://app.codecov.io/gh/PlasmaPy/PlasmaPy/pull/2555/blob/src/plasmapy/particles/atomic.py#L595

pheuer:
> How should I add coverage for these lines?
> 
> https://github.com/PlasmaPy/PlasmaPy/blob/fe4ed92f5b9eb30b19e26ec6a4f7efcbc0a662f8/src/plasmapy/particles/atomic.py#L1238-L1241
> 
> The returned result can have some ~100 entries.

It looks like this isn't being hit just because all the test cases supply `energies`? You could include one separate test that doesn't supply an `energies` vector, then just confirm that you get the tuple out?

JaydenR2305:
any ideas on this? 
https://github.com/PlasmaPy/PlasmaPy/actions/runs/9291692287/job/25570832682#step:6:118

JaydenR2305:
pre-commit.ci autofix

pheuer (approved):
Looks good!

Issue #1680, opened 2022-08-19
Move from `setup.py` and `setup.cfg` to `pyproject.toml`


namurphy:
The Python community has been converging on using [`pyproject.toml`](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/). The old way of doing it with `setup.py` and `setup.cfg` is [now considered legacy](https://pip.pypa.io/en/stable/reference/build-system/setup-py/).  The Python Packaging User Guide's tutorial on [packaging Python projects](https://packaging.python.org/en/latest/tutorials/packaging-projects/) now includes `pyproject.toml` and doesn't include `setup.cfg` or `setup.py`.

We should move away from using our [`setup.cfg`](https://github.com/PlasmaPy/PlasmaPy/blob/main/setup.cfg) for configuration and packaging stuff when we have the option to, and instead use our [`pyproject.toml`](https://github.com/PlasmaPy/PlasmaPy/blob/main/pyproject.toml) file.  We should also drop our [`setup.py`](https://github.com/PlasmaPy/PlasmaPy/blob/main/setup.py`) and have `pyproject.toml` handle those responsibilities.

We are also currently using `setup.cfg` for configuring `flake8`, `pytest`, `coverage`, and `codespell`.  We should move the configurations out of `setup.cfg` and into `pyproject.toml` or elsewhere, probably splitting this up into multiple smaller pull requests to make each individual chunk more manageable.

This isn't high priority, but it would be good to do this in the next year or two.

StanczakDominik:
I've taken the liberty of adjusting the labeling here a little - this can essentially be done by trial and error, which is how I'm gonna tackle it if nobody wants to try in the meantime 😄

namurphy:
I looked into using `pyproject.toml` for flake8, and found that flake8 doesn't support `pyproject.toml`, and is unlikely to support it in the next few years (see https://github.com/PyCQA/flake8/issues/234). However, we could use [Flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) in order to switch to `pyproject.toml`.

Cadair:
Cross posting from wtbarnes/fiasco#189:

Note that not having a setup.py file present in the repo requires a (at the time of writing) very new version of pip to support development installs.

Having a two line setup.py file and moving everything else in to the other config files is what I would currently recommend.

namurphy:
[...]

PR #2558, opened 2024-03-04
Remove `setup.py`

namurphy:
Since PlasmaPy has `pyproject.toml`, our `setup.py` file is no longer necessary. In particular, the `setup.py` configuration [is now considered legacy](https://pip.pypa.io/en/stable/reference/build-system/setup-py/). This PR removes `setup.py`.

We had previously kept a simple `setup.py` around because `pip install -e` had only recently been implemented for projects without `setup.py`.

Cross-references: #1680, #1758

### DIFF
--- a/changelog/2558.internal.rst
+++ b/changelog/2558.internal.rst
@@ -1,0 +1,1 @@
+Removed :file:`setup.py`.
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,0 @@
-from setuptools import setup
-
-setup()


ejohnson-96 (approved):
Read the Docs does not compile

namurphy:
Thank you for the review!  Looks like the Read the Docs build failed because of too many concurrent operations, and it was able to pass the most recent time around.

namurphy:
One possible use case where we might want to keep `setup.py` would be if we wanted to include tests in the sdist but not in wheels.  There's some discussion about this in https://github.com/astropy/astropy/issues/16152.  (It's fun hanging out in Astropy's repo since I get to learn lots of things!)

For the time being, I think our distribution and tests are small enough that it's fine to keep the tests.  If we start putting test data in a lot of the package, or if someone requests it, then it'd be worth revisiting this.

PR #2559, opened 2024-03-05
Make `common_isotopes`, `known_isotopes`, and `stable_isotopes` each return a `ParticleList`

namurphy:
Whilst working on #2458, I noticed that a few of the functions in `plasmapy.particles.atomic` each returned a `list` containing `Particle` objects.  This PR changes it so that they return a `ParticleList`.

### DIFF
--- a/plasmapy/particles/atomic.py
+++ b/plasmapy/particles/atomic.py
@@ -783,7 +783,7 @@ def stable_isotopes(
         return [
             isotope
             for isotope in KnownIsotopes
-            if _isotopes.data_about_isotopes[isotope.symbol]["stable"] == stable_only
+            if _isotopes.data_about_isotopes[isotope.isotope]["stable"] == stable_only
         ]
 
     if argument is not None:
--- a/plasmapy/particles/atomic.py
+++ b/plasmapy/particles/atomic.py
@@ -493,7 +493,7 @@ def half_life(particle: Particle, mass_numb: int | None = None) -> u.Quantity[u.
     return particle.half_life
 
 
-def known_isotopes(argument: str | int | None = None) -> list[str]:
+def known_isotopes(argument: str | int | None = None) -> ParticleList:
     """
     Return a list of all known isotopes of an element, or a list of all
     known isotopes of every element if no input is provided.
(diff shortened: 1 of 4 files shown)


plasmapy/particles/atomic.py
```diff
@@ -664,11 +664,11 @@ def common_isotopes_for_element(
         CommonIsotopes = [
             isotope
             for isotope in isotopes
-            if "abundance" in _isotopes.data_about_isotopes[isotope]
+            if "abundance" in _isotopes.data_about_isotopes[isotope.isotope]
```
namurphy, line 667: `Particle.isotope` returns a string with the symbol of the isotope, which is expected by that `dict`.

plasmapy/particles/atomic.py
```diff
@@ -783,7 +783,7 @@ def stable_isotopes_for_element(
         return [
             isotope
             for isotope in KnownIsotopes
-            if _isotopes.data_about_isotopes[isotope]["stable"] == stable_only
+            if _isotopes.data_about_isotopes[isotope.symbol]["stable"] == stable_only
```
namurphy, line 786:
```suggestion
            if _isotopes.data_about_isotopes[isotope.isotope]["stable"] == stable_only
```

plasmapy/particles/tests/test_atomic.py
```diff
@@ -483,18 +483,14 @@ def test_isotopic_abundance() -> None:
         isotopic_abundance("Og-2")
 
 
-isotopic_abundance_elements = (
-    atomic_number(atomic_numb) for atomic_numb in range(1, 119)
-)
```
namurphy, line 488: I seem to have discovered Python generators in Dec 2017 and perhaps maybe had just a little bit too much fun with them...

plasmapy/particles/tests/test_atomic.py
namurphy, line 1: This file has some incidental refactorings that I made while trying to debug a thing.

plasmapy/particles/atomic.py
ejohnson-96: I thought we didn't return dict objects?

plasmapy/particles/atomic.py
ejohnson-96: Updated to isotope

ejohnson-96 (approved):
👍

plasmapy/particles/atomic.py
namurphy: The dict here was a private one containing isotope data, so it wasn't part of our public API.  I'd say that there are times when a dict is a good choice for a return value, and other times when it's worth creating something like a class or namedtuple.  As with everything, there are tradeoffs!

namurphy:
pre-commit.ci autofix

PR #2565, opened 2024-03-08
Update documentation guide on how to locally build docs

namurphy:


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py312,build_docs_pins
+envlist = py312-pins-pytest_all,build_docs
```
namurphy, line 2: These are the environments that get run when we run `tox` without specifying any environments. They needed updating since I recently renamed the environments. We don't actually use this in our CI, so I didn't notice it then.

ejohnson-96: 👍

ejohnson-96 (approved):
👍

StanczakDominik (approved):

PR #2567, opened 2024-03-09
Add a function to update downloaded files 

pheuer:
## Description

This PR adds a function `util.downloader.update_downloads()` that update any downloaded resource files. 

## Motivation and context

The way the `downloader.get_file` function works, once a file exists in the directory with the requested name that file path will always be returned. There is no way of detecting when a file of the same name is updated on the github repository. I could do this using the GitHub file SHA, but that would require a GitHub API token, which I really don't want to involve here. 

My short-term solution is to add a function `update_downloads()` that manually goes through and re-downloads any files the user has already downloaded. The idea is that users could call this function manually if desired. We can add a note to functions that use the `get_file` function about this option. 

The function will only delete files that match a file name on the repository, and will never delete directories. So, I think this is pretty safe (e.g. users cannot accidentally delete their home directory or something).

### DIFF
--- a/plasmapy/utils/data/downloader.py
+++ b/plasmapy/utils/data/downloader.py
@@ -6,10 +6,11 @@ downloading files from |PlasmaPy's data repository|.
 
 from pathlib import Path
 from urllib.parse import urljoin
+from shutil import rmtree
 
 import requests
 
-__all__ = ["get_file"]
+__all__ = ["get_file", 'update_downloads']
 
 # Note: GitHub links have a problem where the Content-Encoding is set to
 # 'gzip' but the file is not actually compressed. This header is just ignored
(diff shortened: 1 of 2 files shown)


JaydenR2305:
you sure you would need an API token??

https://api.github.com/repos/PlasmaPy/PlasmaPy/commits

pheuer:
> you sure you would need an API token??
> 
> https://api.github.com/repos/PlasmaPy/PlasmaPy/commits
> 
> Definitely a solid idea to add a way to update data files!

Hm you're right, it looks like you can access this without an API token! Maybe it is worth implementing it that way, then...

https://api.github.com/repos/PlasmaPy/PlasmaPy-data/contents/NIST_PSTAR_aluminum.txt

Issue #2535, opened 2024-02-26
Create `NuclearReaction` class to support crosssection data


pheuer:
### Feature description

We need a way to incorporate more information about nuclear reactions in to the Particles module. 

### Motivation

We need crosssections for common fusion reactions to make notebooks on topics like the Lawson criteria. 

### Implementation strategy

1) Refactor `particles.nuclear.nuclear_reaction_energy` into a `NuclearReaction` class. Most of the input parsing is already done, but should be updated to use Particle Lists! The total baryon number, charge, and mass energy methods should also be updated to be based on Particle Lists. @namurphy Do you think you could tackle this? 

2) Add a method to the new `NuclearReaction` class that searches for a matching reaction in an h5 file of nuclear data and returns the relevant data arrays if the reaction is included in the file. 

### Additional context

I've created an example h5 file with one cross-section coming from the ENDF database. I can collect these for a bunch of common reactions once we agree on a format etc.

namurphy:
Thank you for raising this!  I can work on the first part.

Cross-references: #110, #1026, #1450

pheuer:
@namurphy Perfect, thank you! I would just start with a NuclearReaction class now, then we can take a step back and see what features are in common with other classes, like the collision classes, and see if a parent class makes sense? I don't immediately see lots of features shared between collisions and nuclear interactions.

pheuer:
@namurphy I've collected the cross-section data for the fusion reactions mentioned in #110 from ENDF, and created a draft h5 file in the PlasmaPy-data repository: 

https://github.com/PlasmaPy/PlasmaPy-data/blob/main/nuclear_reactions.h5 

We can easily add more later, but I think this is a good start. The most important convention for this file is how to label the reactions. To best match ENDF, I've gone with [Projectile]([Target],[Products])[Remnants] although what is considered a product vs. a remnant for different reactions sometimes seems a bit arbitrary? We'll have to figure out the best way to take search this file given a nuclear reaction.

PR #2570, opened 2024-03-11
Create a class to manage local and online resource files

pheuer:
## Description

Replaces the `get_file()` function with a class that fetches resource files, but also keeps a register of SHA hashes for each file so that files can be re-downloaded automatically if they are updated on the PlasmaPy-data repository. 

## Motivation and context

The way the `downloader.get_file` function works, once a file exists in the directory with the requested name that file path will always be returned. There is no way of detecting when a file of the same name is updated on the github repository. 

This new file uses the GitHub API to get the SHA hash for the file online, and keeps a JSON file registry of the SHA hashes for already downloaded files. If the SHA of the local file doesn't match the online one, the file will be re-downloaded. If the online connection cannot be made and the local file exists, it will be returned with a warning that it may be out of date. 

## Related issues

Revision of #2567 after @JaydenR2305  helpfully pointed out that the GitHub API doesn't actually require an API token for this! 

Inspired by #2555 (NIST data) and #2535 (Nuclear XS data), both of which will explicitly depend on resource files that are too large to include in the base installation of PlasmaPy.


pheuer:
@JaydenR2305 Let me know what you think of this structure, and if you have any suggestions! I still need to write tests etc.

JaydenR2305:
The class structure looks great so far, it's all very intuitive. My only suggestion is to make sure you go through the rigmarole to ensure that the `.plasmapy` directory is created if necessary

pheuer:
> The class structure looks great so far, it's all very intuitive. My only suggestion is to make sure you go through the rigmarole to ensure that the `.plasmapy` directory is created if necessary

Good catch, thanks! I'll add a mkdir

pheuer:
@JaydenR2305 Unfortunately it looks like there is an API limit that I'm already hitting through CI without an API token >< I will refactor to see if I can minimize requests 

"""
ValueError: Resource could not be found locally or retrieved from the PlasmPy-data repository: NIST_PSTAR_tissue_equivalent.txt. Exception raised: URL https://api.github.com/repos/PlasmaPy/PlasmaPy-data/contents/NIST_PSTAR_tissue_equivalent.txt returned JSON file, but missing expected keys 'sha' and 'download_url`. JSON contents: {'message': "API rate limit exceeded for 13.105.49.222. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
"""

pheuer:
So, the rate limit for these API requests is actually quite sever: 60/hr. That leaves us with two issues

1) It will limit the number of times CI can run successfully severly, to only a few times per hour across the entire package.  The only solution would be to somehow set up a plasmapy github API token as a secret and pass that in to the class just when running the CI. That sounds challenging? 

2) We'd have to handle this really carefully for users. Something like, when the API usage is exceeded, defaulting back to the unverified method (not checking the repo before trying to download a file). Or saving a copy of the repo response locally and using that if it's recent enough? It's a bit of a mess. 

@JaydenR2305 @namurphy What do you think?

changelog/2570.feature.rst
```diff
@@ -0,0 +1,4 @@
+Replaced plasmapy.utils.data.downloader.get_file function with a
+new class `~plasmapy.utils.data.downloader.Downloader` with a
+method ~plasmapy.utils.data.downloader.Downloader.get_file` which
```
namurphy, line 3:
```suggestion
Replaced `plasmapy.utils.data.downloader.get_file` with a
new class `~plasmapy.utils.data.downloader.Downloader` with a
method `~plasmapy.utils.data.downloader.Downloader.get_file` which
```

pheuer:
@namurphy 

Great suggestions! 

In theory I like the idea of hosting files somewhere else, but it would need either an API without rate limits or a simple URL file structure that we controlled (e.g. an FTP server?). GitHub conveniently stores the SHA hash for every file - on another platform we would need to access something like that or maybe the date the file was uploaded? Something to determine whether files have been updated. It might require more work on our part when we update or add datafiles, which is pretty easy on GitHub right now. 

Breaking the plan (for now still using GitHub) into two parts: 

**Getting the code to work for users**
I will modify the code yet again to robustly fall back on downloading files without verifying the SHA hash (like we used to do) if the server connection is bad or if the API usage limit is met. The result should at least be as good as what we had before, and should ideally update the files fairly well. 

**Dealing with testing on CI**
I suggest we'd need to implement two of your suggestions to make this work 
1. We should create a cache directory to store files used in any tests so the downloader doesn't need to access the API. You said you know how to do this for testing: can the same or a different cache be used for documentation (e.g. notebooks)? 
2. We need to ignore `test_downloader.py` in tests unless the PR modifies that file, since we can't write tests for the module while using cached files! But we want those tests to run very infrequently.

pheuer:
@namurphy Alright I think this is working now on the user side! The class now makes at most one API call, downloading all the information from the repo in one go and storing it to a file. 

I am still getting a doc error: 

```
/home/runner/work/PlasmaPy/PlasmaPy/docs/changelog/0.9.0.rst:91: WARNING: py:obj reference target not found: plasmapy.utils.data.downloader.get_file
<unknown>:9: WARNING: py:obj reference target not found: plasmapy.utils.data.downloader.get_file
```

Presumably the problem is that this function no longer exists, but it doesn't have a tilde so I don't know why it's being seen as a reference? 

0.9.0.rst:91
```
`plasmapy.utils.data.downloader.get_file`
```

pheuer:
Ok, I think this is almost ready for review, pending setting up a label so that `test_downloader.py` doesn't run unless changes are being made to `downloader.py`? @namurphy is that something you can set up?

For everywhere else in the package, we should be able to use `Downloader(validate=False)` and not incur an API call when downloading the files (since we know we don't have them yet and that they exist, so there's no need to check with the API before downloading).

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        # instantiated.
+        if self._updated_blob_file_from_repo:  # coverage: ignore
+            return None
+
+        reply = self._http_request(self._API_BASE_URL)
+
+        # Extract the SHA hash and the download URL from the response
+
+        # Extract contents to JSON
+        # Not tested, since any URL on the gituhb API that doesn't raise a 404
```
namurphy, line 182:
```suggestion
        # Not tested, since any URL on the GitHub API that doesn't raise a 404 error
```
Incidentally, 404 was the number of a differential equations class I took!  Either that, or I never managed to find the course website. 😬

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+            self._validate = False
+            return None
+
+        for item in info:
+            try:
+                filename = item["name"]
+                repo_sha = item["sha"]
+                download_url = item["download_url"]
+
+            # Not tested, since any URL on the gituhb API that doesn't return a 404
```
namurphy, line 201:
```suggestion
            # Not tested, since any URL on the GitHub API that doesn't return a 404
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        # At the end, write back to the blobfile
+        self._write_blobfile()
+
+        # The blob file has been updated: set this flag so we won't do it
+        # again on this instance of Downloader
+        self._updated_blob_file_from_repo = True
+
+    def _update_blob_entry(
+        self,
+        filename: str,
```
namurphy, line 231:
```suggestion
        filename: str | Path,
```
I don't know if this is consistent with elsewhere, though.

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
-        with path.open(mode="wb") as f:
+        # Extract the 'message' value if it is there
+        # If the file does not exist on the repository, the GitHub API
+        # will return `Not Found` in response to this but not raise a 404 error
+        if reply.status_code == 404:
+            raise ValueError(f"URL returned 404: {url}")
+
+        return reply
+
+    def _filepath(self, filename: str) -> Path:
```
namurphy, line 278:
```suggestion
    def _filepath(self, filename: str | Path) -> Path:
```
Again, not sure if this is needed or consistent.

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        Download a file from a given URL to a specified path.
+
+        Parameters
+        ----------
+        filename : str
+            Name of the file to download: determines the filepath
+
+        dl_url : str
+            URL from which to download
+
```
namurphy, line 293:
```suggestion
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        """Formats a filepath from a filename."""
+        return Path(self._download_directory, filename)
+
+    def _download_file(self, filename: str, dl_url: str) -> Path:
+        """
+        Download a file from a given URL to a specified path.
+
+        Parameters
+        ----------
+        filename : str
```
namurphy, line 288:
```suggestion
        filename : str or pathlib.Path
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+            return filepath
+
+        # Try blindly downloading from the base URL
+        # Note that downloading directly from the RAW url does not
+        # require an API call.
+        try:
+            dl_url = urljoin(self._RAW_BASE_URL, filename)
+            return self._download_file(filename, dl_url)
+        except ValueError:
+            pass
```
namurphy, line 341:
```suggestion
        with contextlib.suppress(ValueError):
            dl_url = urljoin(self._RAW_BASE_URL, filename)
            return self._download_file(filename, dl_url)
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+            f"{filename}."
+        )
+
+    def _get_file_with_validation(self, filename: str) -> Path:
+        """
+        Return file logic with validation.
+        """
+        filepath = self._filepath(filename)
+
+        # Try to update the repo blob dict
```
namurphy, line 355:
```suggestion
```
We can probably avoid this comment since it's very similar to the name of the method being called.

A corollary here is: good choice for the method name!

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        # Otherwise fall back to retrieving the file without validation
+        else:
+            self._validate = False
+            warnings.warn(
+                f"Could not retrieve file {filename} with validation: "
+                "trying again without validation."
+            )
+            return self._get_file_without_validation(filename)
+
+    def get_file(self, filename: str) -> Path:
```
namurphy, line 394:
```suggestion
    def get_file(self, filename: str | Path) -> Path:
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        Returns
+        -------
+        Path : `~pathlib.Path`
+            The local path to the resource file.
+
+        """
+        if self._do_validation:
+            return self._get_file_with_validation(filename)
+        else:
+            return self._get_file_without_validation(filename)
```
namurphy, line 412:
```suggestion
        return self._get_file_without_validation(filename)
```

plasmapy/utils/data/tests/test_downloader.py
```diff
@@ -1,7 +1,79 @@
...
+
+
+@pytest.fixture()
+def downloader_unvalidated(tmp_path):
+    return Downloader(directory=tmp_path, validate=False)
+
+
+test_urls = [
+    # Test with a page we know is up if the tests are running
+    ("https://github.com/PlasmaPy/PlasmaPy", None),
```
namurphy, line 22: Highly appropriate choice for the website! 😺

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        """
+
+        try:
+            reply = requests.get(url)  # noqa: S113
+
+        # No test coverage for this exception since we can't test it without
+        # severing the network connectivity in pytest
+        except requests.ConnectionError as err:  # coverage: ignore
+            raise requests.ConnectionError(
+                "Unable to connect to data " f"repository {self._API_BASE_URL}"
```
namurphy, line 267:
```suggestion
                f"Unable to connect to data repository {self._API_BASE_URL}"
```
I suspect these were originally on two lines, and then were put on one line by ruff-format.  I think we have it set so that this is usually autofixed, but maybe that was skipped since one was an f-string and the other wasn't?

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+                "download_url": download_url,
+            }
+
+    def _http_request(self, url: str) -> requests.Response:
+        """
+        Issue an HTTP request to the specified URL, handling exceptions.
+        """
+
+        try:
+            reply = requests.get(url)  # noqa: S113
```
namurphy, line 261:
```suggestion
            reply = requests.get(url, timeout=10)
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        # Path to the local SHA blob file
+        self._blob_file = Path(self._download_directory, self._blob_file_name)
+
+        # Create the SHA blob file if it doesn't already exist
+        if not self._blob_file.is_file():
+            self._blob_dict = {}
+            self._write_blobfile()
+        # Otherwise, read the SHA blob file
+        else:
+            self._read_blobfile()
```
namurphy, line 85: I'm wondering if we could remove the need for the comment by renaming `_read_blobfile` to `_read_sha_blobfile`.

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        reply = self._http_request(self._API_BASE_URL)
+
+        # Extract the SHA hash and the download URL from the response
+
+        # Extract contents to JSON
+        # Not tested, since any URL on the gituhb API that doesn't raise a 404
+        # should return a JSON
+        try:  # coverage: ignore
+            info = reply.json()
+        except requests.exceptions.JSONDecodeError as err:  # coverage: ignore
```
namurphy, line 186: My personal preference is usually to leave out `# coverage: ignore` comments so that the lines of code that are untested show up more clearly in coverage reports, even though it lowers the code coverage metric.  But, I have no problem leaving it as is, especially since you're the one who is most likely to be editing this file in the future.

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        # require an API call.
+        try:
+            dl_url = urljoin(self._RAW_BASE_URL, filename)
+            return self._download_file(filename, dl_url)
+        except ValueError:
+            pass
+
+        raise ValueError(
+            "Resource could not be found locally or "
+            "retrieved from the PlasmPy-data repository: "
```
namurphy, line 345:
```suggestion
            "retrieved from the PlasmaPy-data repository: "
```

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+
+        try:
+            local_sha = self._blob_dict[filename]["local_sha"]
+        except KeyError:
+            local_sha = None
+
+        try:
+            repo_sha = self._blob_dict[filename]["repo_sha"]
+        except KeyError:
+            repo_sha = None
```
namurphy, line 366:
```suggestion
        local_sha = self._blob_dict[filename].get("local_sha", default=None)
        repo_sha = self._blob_dict[filename].get("repo_sha", default=None)
```
This uses `dict.get()` with a default value of `None` if the key isn't present.  I think I got the syntax here right!

namurphy (commented):
Thank you for doing this!  This PR provides some pretty clear improvements to the downloader functionality. I haven't had much practice using `requests`, so I was able to learn a lot from this PR.  I like the overall structure, so most of my comments and suggestions are fairly localized.

I brought up a longer-term idea in https://github.com/PlasmaPy/PlasmaPy-data/issues/3, but even if we ultimately go the route of creating a separate Python package for data, we'll probably need this functionality for tests and example notebooks and larger data sets.

Thank you again!

docs/notebooks/diagnostics/charged_particle_radiography_film_stacks.ipynb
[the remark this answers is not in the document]
pheuer: I don't think that's true? The purpose of creating the downloader.get_file method in the first place was that this notebook (and tests for this functionality) needed a way to get the relevant data files. It's possible I accidentally pushed it as executed here, but it's not supposed to be!

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
-def get_file(basename, base_url=_BASE_URL, directory=None):
-    r"""
-    Download a file from a URL (if the file does not already exist) and
-    return the full local path to the file.
+class Downloader:
+    """
+    Accesses the PlasmaPy resource files.
+
+    Retrieves local paths to resource files, and downloads those files from
+    |PlasmaPy's data repository| if they cannot be found locally.
```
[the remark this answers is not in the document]
pheuer: They mostly confuse me, but you set this one up previously so it was easy for me to copy over ;)

plasmapy/utils/data/downloader.py
pheuer: Yeah, honestly the comment is not really necessary as is

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        with self._blob_file.open("r") as f:
+            self._blob_dict = json.load(f)
+
+    @property
+    def _api_connected(self) -> bool:
+        """
+        Returns `True` if a connection exists to the API, otherwise `False`.
+        """
+        try:
+            # Requesting this URL does not count as an API query
```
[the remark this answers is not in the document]
pheuer: Once you get used to it, HTTP APIs are great! Basically just send a request to the right URL and it responds with a JSON object with everything you need!

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+        -------
+        repo_blob_dict : dict
+            Dictionary with filenames as keys. Each item is another entry
+            with keys `sha` and `download_url`.
+        """
+        # If the current blob file has been updated in the past 5 minutes,
+        # don't bother doing it again
+        # Ignore in tests, as this won't happen in CI
+        try:
+            if time.time() - self._blob_dict["_timestamp"] < 300:
```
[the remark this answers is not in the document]
pheuer: Good idea, I'll make that change

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,409 @@
...
+                return None  # coverage : ignore
+        # If the _timestamp key hasn't been set yet, the blob file has
+        # never been updated before
+        except KeyError:  # coverage : ignore
+            pass
```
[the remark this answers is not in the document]
pheuer: Ooh fancy!

plasmapy/utils/data/downloader.py
pheuer: I think a few of these were necessary to clear the coverage test: this module has more untestable code than usual because situations like network connectivity and future changes to the GitHub API are hard to test.

plasmapy/utils/data/downloader.py
pheuer: Actually here it does have to be a string (just the file basename, not the path)

plasmapy/utils/data/downloader.py
pheuer: What does this actually do? I think this comment is left over from the old function...

plasmapy/utils/data/downloader.py
pheuer: This one is also just the str

plasmapy/utils/data/downloader.py
pheuer: Oh nice!

docs/notebooks/diagnostics/charged_particle_radiography_film_stacks.ipynb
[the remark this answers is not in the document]
pheuer: That is one way we could prevent an API call when building the docs! But I want to try and avoid that on principle since pre-executed notebooks are way more likely to get buggy/out of date.

pheuer:
Thanks for the review @namurphy! I'll fix the remaining few failing tests soon, but I don't want to merge this until we have a solution for preventing `test_downloader.py` from running on most PRs, otherwise it will cause a ton of sporadic CI failures in other PRs!

plasmapy/utils/data/downloader.py
namurphy: This prevents the request from hanging indefinitely.  I didn't know this until I looked up ruff rule [S113](https://docs.astral.sh/ruff/rules/request-without-timeout/) yesterday.

pheuer:
A couple good ideas from a conversation with @JaydenR2305 ! 

GitHub apparently generates an API token 'GITHUB_TOKEN' at the beginning of every workflow job. If we can access this in `test_downloader.py`, we could pass it into an optional keyword in `Downloader` and make authenticated API calls! Those have a limit of 5,000/hr, which we would absolutely not exceed at the rate of <~5 calls per commit. All I need to do is 

1) Figure out how to get this variable in `test_downloader.py`. Might be as simple as running `os.environ['GITHUB_TOKEN`]`? 
2) Add a keyword for an API token to Downloader, which might be nice to have anyway. 

Another backup idea we had was to create a `RateLimitError` exception that is raised in `test_downloader.py` if the API becomes rate limited and causes the tests to automatically pass in that case. Since I expect this to be rare (only occur if several people are pushing commits all at once) this would still lead to the test running most of the time. 

I vastly prefer the first idea though, so I'll look into that!

pheuer:
@namurphy So regarding getting the token, it looks like there are a couple steps
 
1) Generating the GITHUB API token (you'd have to do this on the PlasmaPy account)
2) Placing the API token in the Actions Secrets for the repository. 
3) In each test that will run pytest, we need to add this to the env: 

```
        env:
             GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
Can you do 1 and 2? I could try 3, but I'm not sure if I've identified the right file? I'm looking at `.github/workflows/tests.yml` but I'm not sure where to make the change. Would it go here? 

```
steps:
    - name: Set tokens
       env:
            GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}

    - name: Checkout code
      uses: actions/checkout@v4
      with:
        fetch-depth: 0

    - name: Set up Python
      uses: actions/setup-python@v5
      with:
        python-version: ${{ matrix.python }}
        cache: pip
```

So it's done before each run of tests?

pheuer:
Support for API authorization is now working! Just need the API token to be accessible through `os.environ` and this should be set!

plasmapy/utils/data/downloader.py
pheuer: Syntax is actually dict().(key, default), and defaults to a default of None. 

However, I had to roll this back because the try/except here is actually catching two errors: 

1) If `filename` is not in `_blob_dict`
2) If `local_sha` is not in the `filename` dict. 

And using `get` only handles the first one. Using get twice doesn't work because then it tries to call get a second time on None.

namurphy:
@pheuer - I might not have a chance to get to this for a few hours, so I just gave you permissions for the organization in case you still need them (which I've been meaning to do anyway).  I'll take a look at this PR later today.  In the meantime, here are the [docs on environment variables for GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/variables).  Thank you again for doing this!

pheuer:
> @pheuer - I might not have a chance to get to this for a few hours, so I just gave you permissions for the organization in case you still need them (which I've been meaning to do anyway).

Thanks - I looked into it, and it looks like only user accounts can make API tokens (as opposed to organization). Does PlasmaPy have a user account on GitHub, or is just an organization? If one doesn't exist, we'll need to make an account to generate the API token from.

pheuer:
One discovery: variables from the environment need to be explicitly passed through to `tox` as well
https://github.com/tox-dev/tox/issues/2543#issuecomment-1343027510

pheuer:
@namurphy Still no luck here - is it possible that using the cached tox environments is part of the issue? I've got tox now passing these variables through, so now they appear as empty strings in the environment variables within python. But I can't seem to get the actual variables set in `tests.yml` successfully...

pheuer:
Arg, this StackOverflow seems to identify the problem correctly. CI running on a fork of a repository won't have access to secrets. I think this means running this CI on other PRs (once this is in main) will also not have access to the secrets...

https://stackoverflow.com/questions/58737785/github-actions-empty-env-secrets

pheuer:
@namurphy Victory! Using the `GITHUB_TOKEN` secret that GitHub creates, I was able to pass that through into `tox` and then access it in the script. With some modifications to the `Downloader` HTTP function, that now works! I wrote a test which confirms this authentication is working on CI (`test_apo_token`), but that test will always fail locally so I put a break in to make it always pass when NOT running on Actions. 

I think this PR is ready for final review!

pheuer:
@namurphy, just a reminder that this is ready for final review!

namurphy (commented):
Thank you for the reminder!  I think once we address the suggestions here, I'm happy for us to merge it.  Thank you again for doing this too!

.github/workflows/tests.yml
```diff
@@ -71,6 +75,9 @@ jobs:
           .tox
         key: ${{ matrix.toxenv }}-${{ runner.os }}-${{ hashFiles('requirements.txt', 'pyproject.toml') }}
 
+    - name: Check available variables
+      run: env
+
```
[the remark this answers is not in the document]
pheuer: Oops, forgot to remove that!

plasmapy/utils/data/downloader.py
```diff
@@ -4,74 +4,421 @@
...
-__all__ = ["get_file"]
+__all__ = ["Downloader"]
 
-# Note: GitHub links have a problem where the Content-Encoding is set to
-# 'gzip' but the file is not actually compressed. This header is just ignored
-# by the get_file function.
-_BASE_URL = "https://raw.githubusercontent.com/PlasmaPy/PlasmaPy-data/main/"
 
 # TODO: use a config file variable to allow users to set a location
 # for the data download folder?
```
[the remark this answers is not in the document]
pheuer: I'm ok with having that be a default, then going to the home directory if that variable isn't defined? But let's implement it in another PR as you say

plasmapy/diagnostics/charged_particle_radiography/tests/test_detector_stacks.py
```diff
@@ -10,14 +10,20 @@
...
 
 @pytest.fixture()
-def hdv2_stack(tmp_path):
+def downloader(tmp_path):
+    # Turn off validation so API calls aren't used
+    return Downloader(directory=tmp_path, validate=False)
+
+
+@pytest.fixture()
+def hdv2_stack(downloader):
```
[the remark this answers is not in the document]
pheuer: Good idea, I'll add a comment.

plasmapy/diagnostics/charged_particle_radiography/tests/test_detector_stacks.py
```diff
@@ -10,14 +10,20 @@
...
 
 
 @pytest.fixture()
-def hdv2_stack(tmp_path):
+def downloader(tmp_path):
+    # Turn off validation so API calls aren't used
+    return Downloader(directory=tmp_path, validate=False)
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
pheuer: Oops, I thought this was what fixtures already did >< Yes that is good! 

I actually tried making this a `package` scope, so it can be used in other test folders, but that causes issues if you try to run those test files without this file. So, instead I've changed the function to use a session-scoped temporary directory, so tests that download files should leave them there, so that if another function in another test tries to get the same file, it doesn't have to re-download it.

plasmapy/utils/data/tests/test_downloader.py
```diff
@@ -1,7 +1,106 @@
...
+    return Downloader(directory=tmp_path, api_token=api_token)
+
+
+def test_api_token(downloader_validated):
+    """
+    Test whether the API connection is valid
+    """
+    # Skip the test if running locally
+    if not in_ci():
+        return None
```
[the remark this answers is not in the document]
pheuer: Yes that's much better, that annotates it nicely as skipped when running locally.

pheuer:
@namurphy Great, thanks for your help, I think this is done! The remaining pre-commit issue seems to be on main (I cannot reproduce it locally, but this PR doesn't touch the file causing the issues?) Once you approve, please feel free to merge.

namurphy (approved):
LGTM! 🚀

namurphy:
The problem with the pre-commit check is related to a bug in ruff v0.3.4 (https://github.com/astral-sh/ruff/issues/10528) that will be included in the next release.  We can ignore it, so I'll use my superpowers to ignore a required check and merge this.  Thank you again!

PR #2571, opened 2024-03-11
Reorganize and improve naming consistency of tests

namurphy:
- PlasmaPy's test suite has included runs using the minimal versions of PlasmaPy's dependencies. However, "minimal" is ambiguous because it's not clear whether it's the minimum versions of just direct dependencies, or _all_ dependencies.  [uv](https://astral.sh/blog/uv) has `--resolution=lowest` and `--resolution=lowest-direct` flags to help with dependency resolution (using `uv pip compile`). `lowest` provides the lowest allowed versions of all dependencies, while `lowest-direct` provides the lowest allowed versions of _direct_ dependencies and the newest allowed versions of all other dependencies.  Using `lowest` wasn't able to resolve the environment for PlasmaPy, so we're using `lowest-direct` (which is consistent with what we've done in the past with `pip-compile` from `pip-tools`).  This PR renames "minimal" tests to "lowest direct" to remove the ambiguity and better reflect what the dependencies are actually doing.
 - I made some updates to testing environment names that I'd been meaning to do previously.
 - I consolidated the `build_docs` and `build_docs_pins` environments for `tox` into `build_docs`.  We want the doc build to be done consistently, so it should always used the pinned requirements.


.pre-commit-config.yaml
namurphy, line 1: I snuck in a `pre-commit autoupdate`.

.github/workflows/tests.yml
```diff
@@ -179,6 +179,10 @@ jobs:
       run: |
         pip install --progress-bar off .
 
+    - name: Uninstall PlasmaPy with pip
+      run: |
+        pip uninstall plasmapy
+
```
namurphy, line 185: To make sure that the environment is fresh before following up with the next step, which is to try an editable installation.

PR #2578, opened 2024-03-12
Explicitly specify UTF-8 for citation generation

JaydenR2305:
Declares `encoding="utf-8"` for reading the `CITATIONS.cff` file. For one reason or another, machines don't always default to this encoding, using instead something like cp1252. These machine-dependent encoding standards are usually incapable of reading certain characters. Explicitly specifying UTF-8 encoding will ensure documentation can be built regardless of a user's operating system.


namurphy (approved):
Good catch!  LGTM!

PR #2584, opened 2024-03-15
Enable a `tox` extension to use `uv` instead of `pip` for installing all the things

namurphy:
This PR adds the [`tox-uv`](https://github.com/tox-dev/tox-uv) extension to [`tox`](https://tox.wiki) to make it so that [`uv`](https://github.com/astral-sh/uv) installs Python packages into `tox` environments and the creation of virtual environments.It also adds `tox-uv` into the `docs` and `tests` dependency sets so that when users run `tox` locally, virtual environment creation will be done with `uv`. 

The consequences of this PR are:

 - CI will be faster when reading from a cache by ∼10%
 - CI will be significantly faster
 - Running tests locally with tox will be faster.

`uv` is very new (the one-month anniversary of its release is today! 🎂), but work for the vast majority of common usages of `pip`.  We recently began using `uv` to generate our pinned requirements files in #2523. We also recently made some significant improvements to our CI by caching `.tox` in #2552. 

The risks of this PR are very low; if one of these packages stops being maintained, then we can revert this PR. 

If we decide to eventually switch from `tox` to [`nox`](https://nox.thea.codes/en/stable/usage.html) (#1734), it looks straightforward to use the `uv` backend for `nox`.

PR #2590, opened 2024-03-21
Update contributor guide discussions about tox and nox

### LINKED ISSUE #2583: Update testing guide to reflect that repeat runs of tox are only a few seconds slower than pytest
The testing guide currently states that running tests with tox is slower because it needs to recreate an environment. More recent versions of tox are doing a better job at caching environments, so I'm finding now that running tests locally with tox is only about 3 seconds longer than running tests with pytest, except for the first tox run when it needs to recreate the environment.  Running tox also makes sure that the tests are done in a consistent environment, so I think we should recommend tox as the recommended way to run tests nowadays.  

(Incidentally, we might be able to use a tox-uv plugin which would speed up the creation of environments, but that's a separate issue.)

namurphy:
This PR makes some temporary changes to the contributor guide during the transition from tox to nox, ahead of the v2024.5.0 release today or tomorrow. I plan to make further updates in the next month or two.

 - I updated the testing guide to reflect that use of `tox-uv` greatly reduces the overhead of running `tox`.  
 - I added an admonition to the testing guide that states that tests will soon need to be run with `tox`.
 - I updated the doc guide to reflect that the documentation is now built using `nox` instead of `tox`.  

Closes #2583.  Cross-reference: #2584.  Related to: #1734, #1735.


namurphy:
Hm...not sure what's happening with the doc build.  Will look into it tomorrow. 

```
Exception occurred:
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_tabs/tabs.py", line 143, in run
    tabs_id = self.env.temp_data["tabs_stack"][-1]
              ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'tabs_stack'
The full traceback has been saved in /tmp/sphinx-err-oiyp82in.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

docs/contributing/doc_guide.rst
```diff
@@ -1153,16 +1153,16 @@ Building documentation
    documentation locally on your own computer. New contributors can
    safely skip this section.
 
-There are two methods for building the documentation: make_ and |tox|.
+There are two methods for building the documentation: make_ and |nox|.
```
ejohnson-96, line 1157: tox to nox to nox to tox argh.... 😆

docs/contributing/doc_guide.rst
```diff
@@ -1184,40 +1184,37 @@ It may also be necessary to install the following software:
...
-   .. group-tab:: tox
+   .. group-tab:: nox
 
-      We can use |tox| to build the documentation locally by running:
+      We can use |nox| to build the documentation locally by running:
 
       .. code-block:: bash
 
-         tox -e build_docs
+         nox -s docs
```
ejohnson-96, line 1204: tox -e replaced with nox -s, assuming they are the correct letters but just checking

docs/contributing/doc_guide.rst
```diff
@@ -1184,40 +1184,37 @@ It may also be necessary to install the following software:
...
 
       To pass any `options to sphinx-build`_, put them after ``--``. For
-      example, use :bash:`tox -e build_docs -- -v` to increase output
+      example, use :bash:`nox -s docs -- -v` to increase output
```
ejohnson-96, line 1207: See previous comment

docs/contributing/doc_guide.rst
namurphy: Yep, this is important to double check!   nox primarily uses `-s` because it is short for "session", while tox uses `-e` primarily because it is short for "environment".

(The `-e` also works for nox, but I figured it'd be better to be consistent with the nox docs.)

namurphy:
Thank you again!

PR #2593, opened 2024-03-21
Add some new ruff rules

namurphy:
This PR adds a few rules in ruff's [`RUF`](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf) rule set that have been taken out of ruff's nursery and thus are ready to be used in production.

### DIFF
--- a/plasmapy/analysis/tests/test_fit_functions.py
+++ b/plasmapy/analysis/tests/test_fit_functions.py
@@ -317,7 +317,7 @@ class BaseFFTests(ABC):
 
         with with_condition:
             results = ff_obj.func_err(x, **kwargs)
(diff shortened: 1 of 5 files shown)


plasmapy/analysis/tests/test_fit_functions.py
```diff
@@ -317,7 +317,7 @@ def test_func_err(self, x, kwargs, with_condition) -> None:
 
         with with_condition:
             results = ff_obj.func_err(x, **kwargs)
-            if "rety" in kwargs and kwargs["rety"]:
+            if kwargs.get("rety"):
```
namurphy, line 320: This one is neat!

PR #2601, opened 2024-03-30
Update project settings

namurphy:
In #2598, I ended up making some miscellaneous changes that would be largely hidden during review.  This PR separates those changes into a separate PR so that it's easier to see what is being changed.

The reduction in coverage is at least partially due to a change in the coverage settings, in particular for [branch coverage](https://coverage.readthedocs.io/en/latest/branch.html#branch).

PR #2609, opened 2024-04-03
Loosen voila requirement and update pinned requirements

namurphy:
This is to allow older versions of `jupyter-server` and `jupyter-core` in the environment for the Python in Heliophysics Community summer school...hopefully!

I had previously set a stricter version of `voila` because I didn't think that it was widely used in packages likely to be installed alongsize PlasmaPy, but then realized with @brianthomas that `voila == 0.5.0` has strict requirements on `jupyter-server`, `jupyter-core`, and perhaps other packages.  The PyHC environment is currently running voila 0.4.2 so this PR should do it.

I also added some comments about setting requirements, including regarding NEP 29 and SPEC 0.


namurphy:
The code coverage check failure is because we changed coverage settings in #2601, and so are unrelated.

ejohnson-96:
> The code coverage check failure is because we changed coverage settings in #2601, and so is unrelated.

Okay thanks, I was wondering what was going on there

ejohnson-96 (approved):
👍

PR #2612, opened 2024-04-03
Include coverage in a weekly test

namurphy:
The code coverage reports have been acting a little weird lately.  In #2598, for example, the BASE commit is from two months ago (with the nearest commits being #2477, #2478, & #2481).  This PR is an attempt to get coverage reports uploaded from `main`.  

Even if this PR does work, it's more of a patch 🩹 rather than a complete fix.  It might be helpful to set it up so that the coverage test gets run every time a pull request is merged.

It may also be that we have code coverage set up incorrectly, or that the github token expired or something.

PR #2617, opened 2024-04-03
Update Codecov settings and usage

namurphy:
## Description

This PR does some clean-up of the Codecov configuration file in `codecov.yml`, as well as how the Codecov action gets used in CI and in weekly tests.

This PR sets `fail_ci_if_error` to `true`, which will help us notice a lot sooner when our Codecov configuration breaks. 

## Motivation and context

I recently noticed that we had not been uploading coverage reports from `main` to Codecov for approximately two months. Coverage reports from forks were still getting uploaded correctly (since they can be done tokenlessly), which meant that we hadn't noticed it for a while.  However, the coverage reports from pull requests were being compared against a commit to `main` from two months ago (around the time of #2478).  I recently fixed this problem by setting `CODECOV_TOKEN` as a GitHub secret, and wanted to follow up with some subsequent clean-up (and a better explanation).  

## Related issues

This is a follow-up to changes begun in: #2612, #2613, #2615, #2616.  I noticed the problem in #2598, with some follow-up in #2601.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,11 +74,12 @@ jobs:
 
     - name: Upload coverage reports to Codecov
       if: ${{ contains(matrix.toxenv,'_cov') }}
-      env:
-        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       uses: codecov/codecov-action@v4
       with:
         file: ./coverage.xml
+        fail_ci_if_error: true
+      env:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   documentation:
 
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,8 @@
+# CODECOV_TOKEN is set in the repository's secrets.
 codecov:
   notify:
-    wait_for_ci: no
-  require_ci_to_pass: no
+    wait_for_ci: false
+  require_ci_to_pass: false
 coverage:
   range: 75..98
   status:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,12 +1,17 @@
 codecov:
-  token: 85c13d8a-e9e1-49fd-97d4-7ba103f3e830
   notify:
     wait_for_ci: no
   require_ci_to_pass: no
 coverage:
+  range: 75..98
   status:
     project:
       default:
-        target: auto
-        threshold: 1%
+        threshold: 0.2%
         removed_code_behavior: adjust_base
+        if_not_found: failure
+    patch:
+      default:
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -74,11 +74,12 @@ jobs:
 
     - name: Upload coverage reports to Codecov
       if: ${{ contains(matrix.toxenv,'_cov') }}
-      env:
-        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       uses: codecov/codecov-action@v4
       with:
         file: ./coverage.xml
+        fail_ci_if_error: true
```
namurphy, line 80: This really should be set to `true` by default...sigh...

codecov.yml
```diff
@@ -1,12 +1,17 @@
...
 coverage:
+  range: 75..98
   status:
     project:
       default:
-        target: auto
-        threshold: 1%
+        threshold: 0.2%
         removed_code_behavior: adjust_base
+        if_not_found: failure
```
namurphy, line 12: I _think_ that this will help us find problems with our Codecov setup more quickly.  If we run into any problems in the near future, this setting could be the source of it.

codecov.yml
```diff
@@ -1,12 +1,17 @@
 codecov:
-  token: 85c13d8a-e9e1-49fd-97d4-7ba103f3e830
   notify:
     wait_for_ci: no
   require_ci_to_pass: no
 coverage:
+  range: 75..98
```
namurphy, line 6: If a file has code coverage < 75%, it'll appear red.  Between 75% to 98%, it'll appear yellow.  Above 98%, it'll appear green.

codecov.yml
```diff
@@ -1,12 +1,17 @@
 codecov:
-  token: 85c13d8a-e9e1-49fd-97d4-7ba103f3e830
```
namurphy, line 2: The token is now set as a GitHub secret called `CODECOV_TOKEN`, so it's redundant to have it here too.  It seemed like this setting has been ignored.  There's a chance this change may break something though.

ejohnson-96: Might be worth a breaking.rst incase people have used to token to set up there own repos on their local machines

ejohnson-96 (approved):
👍

codecov.yml
namurphy: I believe the only usage of this token has been via the `main` branch of the original GitHub repo.

namurphy:
Thank you for the review!  We may still want to fine tune Codecov settings after this, but I think this configuration is close to what we want.

Issue #2620, opened 2024-04-05
`thomson_fitting` error with `lmfit==1.3.0` 


pheuer:
### Bug description

I'm looking into this: I can reproduce this error locally by upgrading my lmfit version from 1.2.0 to 1.3.0 (the latter is what PlasmaPy is now running on, at least in CI).  The error occurs in `thomson_fitting.ipynb`: 

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[9], line 2
      1 fit_kws = {}
----> 2 epw_result = epw_model.fit(
      3     epw_skw,
      4     params=params,
      5     wavelengths=epw_wavelengths.to(u.m).value,
      6     method="differential_evolution",
      7     fit_kws=fit_kws,
      8 )

File ~\anaconda3\envs\dev311\Lib\site-packages\lmfit\model.py:1174, in Model.fit(self, data, params, weights, method, iter_cb, scale_covar, verbose, fit_kws, nan_policy, calc_covar, max_nfev, coerce_farray, **kwargs)
   1168     fit_kws = {}
   1170 output = ModelResult(self, params, method=method, iter_cb=iter_cb,
   1171                      scale_covar=scale_covar, fcn_kws=kwargs,
   1172                      nan_policy=self.nan_policy, calc_covar=calc_covar,
   1173                      max_nfev=max_nfev, **fit_kws)
-> 1174 output.fit(data=data, weights=weights)
   1175 output.components = self.components
   1176 return output

File ~\anaconda3\envs\dev311\Lib\site-packages\lmfit\model.py:1559, in ModelResult.fit(self, data, params, weights, method, nan_policy, **kwargs)
   1557 self.userargs = (self.data, self.weights)
   1558 self.userkws.update(kwargs)
-> 1559 self.init_fit = self.model.eval(params=self.params, **self.userkws)
   1560 _ret = self.minimize(method=self.method)
   1561 self.model.post_fit(_ret)

File ~\anaconda3\envs\dev311\Lib\site-packages\lmfit\model.py:997, in Model.eval(self, params, **kwargs)
    964 def eval(self, params=None, **kwargs):
    965     """Evaluate the model with supplied parameters and keyword arguments.
    966 
    967     Parameters
   (...)
    995 
    996     """
--> 997     return coerce_arraylike(self.func(**self.make_funcargs(params, kwargs)))

[...]

PR #2618, opened 2024-04-05
Add examples to `thermal_bremsstrahlung` docstring

jwreep:
## Description

I've added a few examples to the docstring of `thermal_bremsstrahlung`.

## Related issues

Fixes #2529

PR #2621, opened 2024-04-05
Temporarily pin `lmfit < 1.3.0`

namurphy:
This is due #2620, and was discovered in #2618.  We should remove this upper limit before the next release (#2508).

Thanks to @jwreep & @pheuer for finding and reporting this!


namurphy:
I am not sure why Codecov failed on this one, first time around.  Related to the ongoing saga including #2613, #2615, #2616, etc.

pheuer:
@namurphy See the hotfix in #2620 : I could implement that on a PR trivially if you'd rather use that workaround and remove the cap on `lmfit` until the bug gets patched.

namurphy:
> @namurphy See the hotfix in #2620 : I could implement that on a PR trivially if you'd rather use that workaround and remove the cap on `lmfit` until the bug gets patched.

That would be great!  Having upper limits on dependencies can cause significant headaches for anyone trying to create a Python environment with lots of packages, so we should avoid it at all possible.  Once you open that PR, please feel free to close this one. Thank you!

PR #2622, opened 2024-04-05
Use pinned requirements for Python 3.12 in Read the Docs build 📍

namurphy:
The error described in #2620 suprised us 😮 because we were not using pinned requirements 📍 for the RTD build.  This PR should rectify that so we aren't surprised by Read the Docs failures in normal PRs.   Instead we'll find them in the PRs that regenerate requirements files.  As a consequence of this PR, the GitHub Action to build documentation will use the same Python environment as the 

I'm also updating the RTD build to Python 3.12. 🐍  

📌 I'm considering moving the docs requirements to `docs/requirements.txt`, which would involve changing the `tox` environment used for building the docs.  I'm undecided on that, but that would be best for a different PR.


namurphy:
This PR may require some iteration. 🙃

ejohnson-96 (approved):
👍

namurphy:
Thank you for the review!

Issue #2591, opened 2024-03-21
Improve Thomson module


pheuer:
### Feature description

The Thomson scattering module is getting a good amount of users, and some consistent feature requests are emerging! I will start collecting them here.

PR #2636, opened 2024-04-15
Several improvements to OTS fitting

pheuer:
This PR makes updates to the OTS fitting module to address the following changes listed in #2591: 

Requested new features
- Allow ion_z and ion_mass to be parameters that can be fit.
- Add constant background parameter 

Changes made in this PR
- Added a new `background` parameter that is added as a constant offset to the normalized Skw to represent noise floor. 
- Changed the way the notch keyword in the lite function is applied to fix a bug. 
- Changed _model function to look for `ion_mu` and `ion_z` in params instead of settings. 
- Changed Model creation function to create fixed parameters for any `ion_mu` or `ion_z` parameters not set explicitly.  Remove old `ions` validation that doesn't apply anymore. 
- Refactored tests to get rid of support changes, get rid of some warnings. 
- Added a new test for explicitly setting `ion_mu` and `ion_z` as varying parameters.

The changes to the fitting algorithm are non-breaking, since the Model function can still take a list of ions in settings. There should be no breaking changes.

### DIFF
--- a/changelog/2636.feature.rst
+++ b/changelog/2636.feature.rst
@@ -1,0 +1,2 @@
+Added support for new ``background``, ``ion_mu`` and ``ion_z`` fitting
+parameters to the `~plasmapy.diagnostics.thomson.spectral_density_model` function.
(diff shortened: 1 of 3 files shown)


pheuer:
Per @namurphy in #2637, we should just ignore the failing codecov upload, so otherwise this PR is passing CI and is ready for review.

ejohnson-96 (approved):
👍

Issue #2595, opened 2024-03-21
Different return types for `is_category` method of `ParticleList` than `Particle` & `CustomParticle` 


namurphy:
The return type of `ParticleList.is_category()` is `list[bool]`, while the return type of each of `Particle.is_category()` and `CustomParticle`.is_category()` is a `bool`.  I think this is a violation of the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), at least approximately.

A collection in Python is considered truthy if it contains anything and falsey if it doesn't contain anything.  For example, `ParticleList(["Fe"]).is_category("isotope")` returns `[False]`, but `[False]` evaluates as `True` because the `list` contains an object.  This led to the problem in #2048.  

One possibility to resolve this would be to have `is_category` return a `bool` that is `True` if all of the contents are `True` by default, and then add a keyword argument that would return this particle-wise.  

```Python
>>> particles = ParticleList(["p+", "He"])
>>> particles.is_category("ion")
False
>>> particle.is_category("ion", particlewise=True)
[True, False]
```
This seems a likely way to reduce the chance of errors, but I want to think out the consequences a bit better first.

jwreep:
Thinking a bit about this logic, but isn't this the difference between `np.allclose` and `np.isclose`? 

```
>>> x = [True, False]
>>> np.isclose(x, True)
array([ True, False])
>>> np.allclose(x, True)
False
>>> y = [True, True]
>>> np.isclose(y, True)
array([ True,  True])
>>> np.allclose(y, True)
True
```

Presumably these could be used.

jwreep:
This issue could be closed, presumably.

namurphy:
@jwreep — thank you!  It turns out that one of the less well known ways to help out an open source project is to comment on old issues that can be closed. I do appreciate it!

Issue #2589, opened 2024-03-20
Decide on preferred set of `mypy` rules to adopt


namurphy:
### Description of improvement

In the last few months, we have had a flurry of pull requests to add static type checking with [mypy](https://mypy.readthedocs.io/en/stable/index.html) to PlasmaPy's continuous integration suite. The current version of [`mypy.ini`](https://github.com/PlasmaPy/PlasmaPy/blob/main/mypy.ini) defines which mypy rules to ignore on a per-file and per-error basis (#2424), while having the global settings be approximately mypy's strict mode by default. New files would be checked against mypy's strict mode, while existing files have errors be ignored.  **We should decide on the "final" set of mypy rules to adopt.**

### Motivation

Enabling static type checking with mypy comes with a bunch of tradeoffs.  Having accurate type hint annotations can help with code readability and finding errors, but at the cost of making it hard for making a code contribution. We don't need to enable all of mypy's strict rule set, as it's possible an intermediate level would be most appropriate.

### Implementation strategy

These rules need not be the same for every subpackage, as suggested in #2438. For example, `plasmapy.particles` is used throughout the rest of the package, so having more strict typing requirements would be appropriate.

### Additional context

Cross-references: #268, #2424, #2432, #2534, #2435, #2438, #2444, #2527

pheuer:
I forget the error this raises, but: 

```python

def fcn(x:None| int = None):
    if x is None:
         x = 0

    y = x + 1
```

This raises a mypy error because it thinks x can still be None on the last line, and it drives me crazy since this is a pretty common design pattern!

pheuer:
Ah, maybe it is related to decorators? I'll keep an eye out for next time. Here's another pytest specific one 

```python
def test_fcn(ex : Exception | None):
    with pytest.raises(ex):
            ...
```
For some reason mypy is looking for an overloaded function for `pytest.raises` that has `Exception` as its class, except isn't that just the default (and only?) signature for that function?

namurphy:
[...]

PR #2648, opened 2024-04-25
Add `particlewise` keyword to `particle_collections.is_category`

jwreep:
## Description

Addresses #2595 by adding the requested `particlewise` keyword to the function `particle_collections.is_category`.

This changes the default behavior to return a `bool` rather than a `list` of `bool`.  If the original functionality is preferred to be the default, the default value of `particlewise` could be switched to `True`.  

I've added tests that should (hopefully) address all the new cases, by using `particlewise` as both `True` and `False` and adding one new test that should evaluate to `True`.

I looked at the docs to see if any changes were needed, but the `is_category` function doesn't appear to have any further description there.  Please let me know if I missed something!

## Motivation and context

See the description by @namurphy in #2595.  It was marked as "needs discussion", which I don't know if this has been resolved.

## Related issues

#2595


jwreep:
The ruff error failing is [FBT003](https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/), from this line 
`return category_list if particlewise else np.allclose(category_list, True)`
which I think is complaining about the call to `np.allclose`.  An alternative formulation, which I think wouldn't trigger the warning, is ` np.allclose(category_list, np.array([True]))` but that's ugly coding and the extra call to `np.array` will add a small bit of overhead.

There's some relevant discussion in [this thread](https://github.com/astral-sh/ruff/issues/3247).

Any suggestions?

jwreep:
This caused a lot of other tests to fail because `is_category` appears to be used pretty commonly in the decorators.  I'm guessing to avoid breaking lots of other parts of the code it would be better to switch the default to `particlewise = True`.  Alternatively, we would have to carefully make sure to insert the keyword everywhere the function is called.

Any thoughts?

namurphy:
> The ruff error failing is [FBT003](https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/), from this line `return category_list if particlewise else np.allclose(category_list, True)` which I think is complaining about the call to `np.allclose`.

Perhaps we could add a `# noqa: FBT003` comment at the end of the corresponding line so that `ruff` will ignore the rules on those lines?

namurphy:
> This caused a lot of other tests to fail because `is_category` appears to be used pretty commonly in the decorators. I'm guessing to avoid breaking lots of other parts of the code it would be better to switch the default to `particlewise = True`. Alternatively, we would have to carefully make sure to insert the keyword everywhere the function is called.

In the long term, it would probably be best for `Particle.is_category` and `ParticleList.is_category` to return the same type by default.  My inclination would be to insert the keyword everywhere the function is called.  But, that could possibly be a follow-up PR.  

Thank you for doing this!

jwreep:
Thanks @namurphy!  

I've added the suggested `noqa` and tried to update the decorator code to fix the issues.  It looks like it's solved most of the problems.  There's still one outstanding issue with `AbstractPhysicalParticle` which I'll come back to in a while (looks like it defines its own `is_category`).

I'm not sure that I understand the outstanding issue with the Documentation, though.  Any insight into the issue there?

jwreep:
Sorry for the commits going back and forth.  I'm trying to understand the issue with the static type checking.  

I've changed `ParticleList.is_category` to return either `bool` or `list[bool]` which is throwing an error in the static type check because `any(ParticleList)` might not return an iterable.  I have it nested under an if-statement that should determine whether it has returned a `bool` or `list[bool]`, though, so I'm a bit confused why the issue arises.

jwreep:
@namurphy, please review when you can.  I finally managed to get the issues sorted.

changelog/2648.feature.rst
namurphy, line 1: Since this is technically a breaking change, perhaps we could rename the file to `2648.breaking.rst`?

changelog/2648.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added a new keyword ``particlewise`` to the function `~plasmapy.particles.particle_collections.ParticleList.is_category`,
```
namurphy, line 1:
```suggestion
Added a new keyword ``particlewise`` to the method `~plasmapy.particles.particle_collections.ParticleList.is_category` of |ParticleList|,
```
🔧📚 Slight terminology change to reflect that this is a method attached to a class rather than a standalone function.

It'll also be helpful to explicitly state `ParticleList` here for context, since the `~` causes the namespace to be hidden in the rendered documentation so `ParticleList` itself won't appear in the changelog otherwise.

docs/particles/particle_class.rst
```diff
@@ -259,6 +259,18 @@ and/or |ParticleList| objects together.
...
+
+>>> helium_ions.is_category("ion")
+False
+
+We may also check each particle in the list individually by setting
+the keyword ``particlewise`` to `True`.
+
+>>> helium_ions.is_category("ion", particlewise=True)
+[False, True, True]
+
```
namurphy, line 273: 👍🏻 Thank you for updating this here! I often forget to update the narrative docs until I realize it about six months later... 😅

src/plasmapy/formulary/collisions/helio/collisional_analysis.py
```diff
@@ -211,7 +211,7 @@ def temp_ratio(  # noqa: C901
             f"Instead received {len(ions)} input values."
         )
 
-    if not all(ions.is_category("ion")):
+    if not ions.is_category("ion"):
```
namurphy, line 214: 👍🏻 Always nice to be able to shorten/simplify code!

src/plasmapy/particles/particle_collections.py
```diff
@@ -314,41 +314,79 @@ def insert(self, index, particle: ParticleLike) -> None:
...
+
+    @overload
+    def is_category(
+        self,
+        *category_tuple,
+        require: str | Iterable[str] | None = None,
+        any_of: str | Iterable[str] | None = None,
+        exclude: str | Iterable[str] | None = None,
+        particlewise: Literal[False],
+    ) -> list[bool]: ...
```
namurphy, line 335: 👍🏻 Nice use of `@overload`!

namurphy (approved):

I only have a few quick suggestions, and I think we can get this merged in time for our `v2024.5.0` release on Tuesday or Wednesday.

Thank you again!

src/plasmapy/particles/particle_collections.py
jwreep: So, I did this when I was trying to solve my issues with the static typing.  It didn't solve that issue, but from what I gather reading comments on StackExchange and Github, it seems this is a clear way to specify the output when it changes based on a `bool` input.

changelog/2648.feature.rst
jwreep: Will do.

docs/particles/particle_class.rst
jwreep: I only caught this while trying to track down why the docs were failing to build

src/plasmapy/particles/particle_collections.py
namurphy: That is helpful to know!  

If you run into any mypy errors that feel unduly strict, please feel free to bring it up in #2589.  We're still in the process of figuring out how strict we want the mypy settings to be, since there are tradeoffs between the benefits of type hints compared to the effort needed to implement them.  I suspect that we'll probably need to make mypy somewhat less strict.

jwreep: I had to use `# type: ignore[arg-type]` to resolve the issue here:
```
        if (
            not self.allow_custom_particles
            and isinstance(particle, ParticleList)
            and any(particle.is_category("custom", particlewise=True))  # type: ignore[arg-type]
        ):
```
The third condition can only be called if it is indeed a `ParticleList`, but the mypy check insisted that there would be an issue with possibly return something that is not iterable, which means that the call to `any()` would fail.  It might be too strict for this case?  I'll mention it in the other thread.

jwreep:
Thanks for the help!

PR #2653, opened 2024-04-30
Update Codecov settings

namurphy:
We've been having some problems with our Codecov configuration the last few months.  This PR makes it so that CI can pass if the Codecov report doesn't upload correctly.  I've correspondingly set it so that the codecov/patch check is now required, so it's more apparent if it is missing.  This is a bandaid 🩹 which we'll have to look into further.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,15 +4,13 @@ codecov:
     wait_for_ci: false
   require_ci_to_pass: false
 coverage:
-  range: 75..98
+  range: 80..98
   status:
     project:
       default:
-        threshold: 0.2%
+        threshold: 0.1%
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -4,15 +4,13 @@ codecov:
     wait_for_ci: false
   require_ci_to_pass: false
 coverage:
-  range: 75..98
+  range: 90..98
```
namurphy, line 7: The `range` here defines for Codecov reports when the value should appear as red (below 90%), yellow (between 90% to 98%), or green (98% to 100%).

namurphy:
I think I need to merge this before we can tell if it is working or not.

PR #2650, opened 2024-04-29
Update requirements setup

namurphy:
This PR partially addresses #2647 by:

 - Renaming `requirements/` to `ci_requirements/` to indicate that the files contained therein are specifically for CI. (This may also have side benefit of making dependabot ignore the files in this directory.)
 - Adding `ci_requirements/README.md` to describe the purpose of the directory.
 - Adding some breadcrumbs to `pyproject.toml` about regenerating requirements files.

I considered removing `requirements.txt` in this PR, but I wanted to isolate those changes in case anything unexpected happens.

This PR doesn't need to be included in our v2024.5.0 release, which I'm expected to do on Friday.

PR #2654, opened 2024-04-30
Add nox configuration file with environments for building documentation and running mypy

namurphy:
## Description

This PR adds `noxfile.py`, which is a configuration file for [`nox`](https://nox.thea.codes). So far it includes testing environments for building docs and running mypy. After installing `nox` and `uv`, the environments can be run with commands like `nox -e docs` or `nox -e mypy` in the top-level directory of the repository.

## Motivation and context

We currently use `tox` as our test runner. The configuration for `tox` is given in `tox.ini`. I've personally found `tox.ini` to be incredibly difficult to parse and understand. 

`nox` is a test runner that was originally inspired by `tox`. Unlike `tox`, the configuration file for `nox` is a Python file: `noxfile.py`.  In my limited experience, setting up and modifying test environments defined in `noxfile.py` has been considerably less difficult than doing the same in `tox.ini`.  

This PR is only to add `noxfile.py` without adding it to the CI suite and without removing anything from `tox.ini`. Subsequent steps include:

 - Adding environments for running pytest, including with coverage
 - Adding an environment for regenerating requirements files (should wait until after #2650 is resolved)
 - Switching GitHub Actions to use the nox environments instead of tox. (This ideally would be done incrementally.)
 - Updating the testing guide to reflect usage of nox instead of tox
 - Removing `tox.ini`

## Related issues

#1734, #1735


noxfile.py
```diff
@@ -0,0 +1,56 @@
...
+    "sphinx-build",
+    "docs/",
+    "docs/build/html",
+    "--nitpicky",
+    "--fail-on-warning",
+    "--keep-going",
+)
+
+html = ("-b", "html")
+check_hyperlinks = ("-b", "linkcheck")
```
namurphy, line 22: 📝 By defining the options into discrete groups like this, we can compose different environments to build the docs 📚 with different option sets ⚙️, which should make the commands more reusable. I suspect it'll be more maintainable too, since the commands that get reused are isolated from the details of the testing environment ↔️.

conftest.py
```diff
@@ -9,6 +9,8 @@
 settings.register_profile("debug", max_examples=10, verbosity=Verbosity.verbose)
 settings.load_profile(os.getenv("HYPOTHESIS_PROFILE", "default"))
 
+collect_ignore = ["noxfile.py"]
```
namurphy, line 12: ❔ I wasn't able to get `pytest` to ignore this file from within `pyproject.toml`, since I could only find options for ignoring directories rather than files.  

It should also help to have a global pytest configuration file so that it works for both regular tests (in `tests/`) and doctests (in `src/plasmapy/`).

ejohnson-96: Hmm 🤔 how strange

ejohnson-96 (approved):
👍

namurphy:
Thank you for the reviews! 😸

Issue #285, opened 2018-03-02
Add ionization_energy attribute to Particle class


namurphy:
A commonly needed in plasma physics is the energy it takes to ionize a single atom or ion.  The `Particle` class should have an attribute called `ionization_energy` that returns this value.  Example functionality is:
```Python
>>> from plasmapy.atomic import Particle
>>> ion = Particle('Fe 6+')
>>> ion.ionization_energy
<Quantity 2.00233625e-17 J>
```
When this is called for an element or isotope with no charge specified, the energy to ionize a neutral atom should be returned.  The returned value should be in joules because PlasmaPy uses SI units.

The ionization energy might depend on the mass number for the lightest elements, so it would be worth checking if there are different ionization energies for protium, deuterium, and tritium; and also for `He-3` and `He-4`.  

When ionization energy is not available for a particular ion, a `MissingAtomicDataError` exception should be raised.

The terms "ionization energy" and "ionization potential" are used interchangeably, so `Particle.ionization_potential` would be an alternative.  I prefer `Particle.ionization_energy` because it is explicit that it is an energy, because it is shorter, and because it appears to be somewhat more commonly used (565K Google results for "ionization energy" vs. 491K Google results for "ionization potential").

namurphy:
The [NIST Atomic Spectra Database Ionization Energies Form](https://physics.nist.gov/PhysRefData/ASD/ionEnergy.html) would be a good place to get the data for this.  It's tricky to get all of the data, so we have to list the atomic symbols of the elements that we want the ionization energy for.   We can get a list of the elements by running:
```Python
from plasmapy.atomic import atomic_symbol
elements = ""
for i in range(1,105):  # Information for the highest atomic number elements is not available
    elements += f"{atomic_symbol(i)}, "
print(elements[:-2])
```
This yields:
```
[...]

PR #2657, opened 2024-05-03
Issue 285: Add ionization energy data from NIST

chasepd:
## Description

Adds ionization energy data from [NIST ](https://physics.nist.gov/PhysRefData/ASD/ionEnergy.html) to the particle data json files, as well as a property in the `Particle` class to retrieve this information. Returns a descriptive `MissingParticleDataError` if ionization energy data is not available for the particle.

## Motivation and context

Addresses the requested feature at: https://github.com/PlasmaPy/PlasmaPy/issues/285. 

> A commonly needed in plasma physics is the energy it takes to ionize a single atom or ion. The Particle class should have an attribute called ionization_energy that returns this value. 

## Related issues
https://github.com/PlasmaPy/PlasmaPy/issues/285


chasepd:
pre-commit.ci autofix

chasepd:
pre-commit.ci autofix

tests/particles/test_particle_class.py
```diff
@@ -1573,3 +1573,27 @@ def test_undefined_mass_energy() -> None:
     """Test that a particle with an undefined mass energy returns |nan| J."""
     nu_tau_particle = Particle("nu_tau")
     assert u.isclose(nu_tau_particle.mass_energy, np.nan * u.J, equal_nan=True)
+
+
+@pytest.mark.parametrize(
+    ("particle_symbol", "expected_ionization_energy"),
+    [
+        ("H", u.Quantity((13.598434599702 * u.eV).to(u.J).value, u.J)),
```
pheuer, line 1581: I think this is equivalent, and a little clearer?
```suggestion
        ("H", (13.598434599702 * u.eV).to(u.J) ),
```

src/plasmapy/particles/particle_class.py
```diff
@@ -783,6 +785,12 @@ def _assign_atom_attributes(self) -> None:
 
         categories.add(this_element["category"])
 
+        try:
+            ev_value = this_element["ionization energy"] * u.eV
+            attributes["ionization energy"] = u.Quantity(ev_value.to(u.J).value, u.J)
```
pheuer, line 790: Same comment as below, I think this is simpler
```suggestion
            attributes["ionization energy"] = ev_value.to(u.J)
```

changelog/2657.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added ionization energy data from NIST to the `Particle` class.
```
pheuer, line 1:
```suggestion
Added ionization energy data from NIST to the `~plasmapy.particles.Particle` class.
```

changelog/2657.feature.rst
```diff
@@ -0,0 +1,3 @@
...
+This can now be accessed using the `ionization_energy` attribute
```
pheuer, line 2:
```suggestion
This can now be accessed using the `~plasmapy.particles.Particle.ionization_energy` attribute
```

changelog/2657.feature.rst
```diff
@@ -0,0 +1,3 @@
...
+from the `Particle` class.
```
pheuer, line 3:
```suggestion
from the `~plasmapy.particles.Particle` class.
```

pheuer (changes requested):
Very nice start! Here are a few minor corrections. 

There is one larger bug: right now, this code still returns the ground state ionization energy even if the particle is an ion, e.g. 

```python
Particle('C +2').ionization_energy == Particle('C +0').ionization_energy
>True
```

while, according to the NIST tables, the ionization states above should be 47 eV and 11 eV respectively. I can see two ways of fixing this: 

1) Make this property only available for particles corresponding to neutral atoms

2) Much more ambitious: implement this property for all of the ion charge states! 

If you chose the second path, I would recommend the following: 

1) Download the data for every element in the NIST database as a CSV 
2) Write a python script that takes that data and puts it into a single json file that you can add to the particles/data folder. You could add different isotopes there too. 
3) In the attribute, determine the isotope, charge, and nucleus of the particle and return the appropriate value from your json file. 

No pressure to go the second path if you don't want to, just adding this for neutral atoms would be a nice start! But thought I'd offer it as a suggestion since you seem to have handled this pretty easily :)

chasepd:
@pheuer Thanks for the suggestion! I'll give the second route a shot and see where we get from there. It may take a few days but given the feature request has been open since 2018 I imagine that should be fine 😂

pheuer:
> @pheuer Thanks for the suggestion! I'll give the second route a shot and see where we get from there. It may take a few days but given the feature request has been open since 2018 I imagine that should be fine 😂

Absolutely! Thank you for doing this, getting more fundamental data in the package is really useful!

changelog/2657.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added ionization energy data from NIST to the `~plasmapy.particles.Particle` class.
+This can now be accessed using the `~plasmapy.particles.Particle.ionization_energy` attribute
+from the `~plasmapy.particles.Particle` class.
```
namurphy, line 3:
```suggestion
Added ionization energy data from NIST to the |Particle| class.
This can now be accessed using the `~plasmapy.particles.particle_class.Particle.ionization_energy` attribute
from the |Particle| class.
```
📚 We've defined some reStructuredText substitutions (`docs/_global_substitutions.py`) for things like `Particle` which we can use here.  For PlasmaPy's docs, we also have to put the full namespace of the object, including the module name (for better or for worse).

namurphy (commented):
Thank you for doing this! I'm excited to have this in here.  One quick initial comment in addition to Peter's review.

chasepd:
@namurphy and @pheuer - I *think* I've implemented ionization energy for ions and fixed the bug where ions just returned their base element ionization energy:

``` 
Particle('C +2').ionization_energy == Particle('C +0').ionization_energy
>False
```

src/plasmapy/particles/particle_class.py
```diff
@@ -783,6 +785,19 @@ def _assign_atom_attributes(self) -> None:
 
         categories.add(this_element["category"])
 
+        try:
```
pheuer, line 788: One more addition to this logic: I don't think the isotope number changes the ionization energy, but right now I think this code fails to find the ionization energy if you specify the isotope, e.g. 

```python
> Particle('C +5').ionization_energy.to(u.eV)
Out[13]: 489.99 eV

 

>Particle('C-12 +5').ionization_energy.to(u.eV)

  Cell In[12], line 1
    Particle('C-12 +5').ionization_energy.to(u.eV)

  File ~\GitHub\PlasmaPy\src\plasmapy\particles\particle_class.py:1900 in ionization_energy
    raise MissingParticleDataError(

MissingParticleDataError: The ionization energy of C-12 5+ is not available.
``` 

I think the correct behavior should be that, if the particle is an isotope, the lookup symbol should be the element/ion without the isotope specified? e.g. the following should work

```python
>Particle('U-235 +5').ionization_energy == Particle('U-238 +5').ionization_energy
True
```

src/plasmapy/particles/particle_class.py
```diff
@@ -1863,6 +1878,31 @@ def recombine(self, n: int = 1, inplace: bool = False) -> Particle | None:
...
+        ~plasmapy.particles.exceptions.MissingParticleDataError
+            If the ionization energy is not available for the particle.
+        
+
+        Examples
+        --------
+        >>> hydrogen = Particle("H")
+        >>> hydrogen.ionization_energy
+        <Quantity 2.17870942e-18 J>
+
```
pheuer, line 1897: We should cite the source of the data so people know where it is coming from, and the approximate date you downloaded it, something like 

```
Notes
------
Ionization energy data downloaded from the `NIST Atomic Spectra Database <https://physics.nist.gov/PhysRefData/ASD/ionEnergy.html>`_  on 5/4/2024.

File: tests/particles/test_particle_class.py [hunk 9]
```diff
@@ -1573,3 +1573,27 @@ def test_undefined_mass_energy() -> None:
     """Test that a particle with an undefined mass energy returns |nan| J."""
     nu_tau_particle = Particle("nu_tau")
     assert u.isclose(nu_tau_particle.mass_energy, np.nan * u.J, equal_nan=True)
+
```
pheuer, line 1576: Can you add a couple additional tests here: 

1) Ensure that the ionization energy for `U-235 +1` == `U-238 +1` == `U +11
2) Ensure that `C +0` and `C` return the same ionization energy

docs/particles/particle_class.rst
```diff
@@ -74,6 +74,8 @@ These properties are often returned as a |Quantity| in SI units.
 <Quantity 3.888e+08 s>
 >>> iron56.binding_energy.to('GeV')
```
pheuer, line 75: @namurphy this should really be called `nuclear_binding_energy` to distinguish from the sum of the ionization energies for this particle!!  I will open an issue...

pheuer (commented):
Nice work! I think once we get the tests passing, this is ready to merge. 

While reviewing this, I realized that it wouldn't be hard to use the json data you've already compiled to also add an `electron_binding_energy` attribute. I've opened an issue suggesting it if you feel like tackling it in a follow up pull request :)  #2673

pheuer:
For this error 

```
src/plasmapy/particles/_ionization_energy.py:22: error: Argument 1 to "loads"
has incompatible type "bytes | None"; expected "str | bytes | bytearray" 
[arg-type]
        pkgutil.get_data("plasmapy", "particles/data/ionization_energy.jso...
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Looking at this: https://stackoverflow.com/questions/52599007/python3-pkgutil-get-data-usage

Maybe you can try `pkgutil.get_data("plasmapy.particles.data", 'ionization_energy.json')` ? 


@namurphy the pre-commit error is from too much complexity in `_assign_atom_attributes`. Do you want to break this up in this PR? I supposed it could be split into `_assign_element_attributes`, `_assign_ion_attributes`, etc.?

src/plasmapy/particles/particle_class.py
chasepd: Sounds good. Would it be helpful for me to include the script I wrote to download and format the data somewhere? Like maybe in `src/plasmapy/particles/data/utils/` or something?

pheuer: Yes that would be very helpful! I think @namurphy was suggesting a while back that we create a `/tools` folder in the main directory of the project that would hold code like this that isn't part of the package but is important for assembling data/package maintenance? That would indicate it doesn't need to be tested etc. like the actual package code.

tests/particles/test_particle_class.py
chasepd: @pheuer it is certainly possible I'm misunderstanding something (I have a more-than-basic understanding of elements, isotopes, ions, etc, but ultimately am just a SWE looking for ways to volunteer and help scientific research 😅), but I don't think that NIST has data on isotopes such as `U-235` or `U-238`. If they do, I can't figure out how to input it.

Submitting `U-235` [results](https://physics.nist.gov/cgi-bin/ASD/ie.pl?spectra=U-235&submit=Retrieve+Data&units=1&format=0&order=0&at_num_out=on&sp_name_out=on&ion_charge_out=on&el_name_out=on&seq_out=on&shells_out=on&level_out=on&ion_conf_out=on&e_out=0&unc_out=on&biblio=on) in `Element U was given with negative electric charge. It is not allowed in this query`, suggesting they're interpreting this as U with a bunch of extra electrons rather than the intended isotope. 

On the other hand, just submitting `U` gives [a list of ions](https://physics.nist.gov/cgi-bin/ASD/ie.pl?spectra=U&submit=Retrieve+Data&units=1&format=0&order=0&at_num_out=on&sp_name_out=on&ion_charge_out=on&el_name_out=on&seq_out=on&shells_out=on&level_out=on&ion_conf_out=on&e_out=0&unc_out=on&biblio=on), but doesn't appear to list any isotopes (or perhaps I'm just not recognizing how they're differentiating them as an isotope, which is definitely possible)

src/plasmapy/particles/particle_class.py
chasepd: Read one of your other comments before this one, and I think this one answers my question 😂 

Thanks for checking me here - this was due to me having an incomplete understanding of particle physics and playing it on the safe side

tests/particles/test_particle_class.py
chasepd: Nevermind, I just realized your other comment inadvertently answered my question here, haha

src/plasmapy/particles/particle_class.py
pheuer: No problem, I had to look this up too!

chasepd: This should be fixed now in the latest version

src/plasmapy/particles/particle_class.py
chasepd: I'm good with either location; I'll commit it in a new `/tools` directory for now

chasepd:
> @pheuer hmm this appears to still be causing the same issue 🤔 I'll try to dig in and figure out why

Figured this out - this error appears to be manually in the mypy config for _elements and _isotopes:

https://github.com/PlasmaPy/PlasmaPy/blob/main/mypy.ini#L201-L205

chasepd:
pre-commit.ci autofix

chasepd:
pre-commit.ci autofix

pheuer (changes requested):
This looks great! Thanks for adding the downloader script! 

One remaining bug I found: you have separate data for deuterium (H-2) and hydrogen (H-1), but it looks like with the current logic both of these get interpreted as 'Hydrogen' so you never actually get the value for Deuterium. It's only slightly different, but sine you have the data in there it would be good to have it return the right one! 

You could maybe add one more test for for this exception (since in all the other cases, you do have a test that correctly shows that the different isotopes have equal ionization energies). 

To be clear on the physics, technically all of the isotopes have different ionization energies, but the difference is just vanishingly small above He so it's not measurable. 

```python
Particle('H').ionization_energy == Particle('D').ionization_energy == Particle('H-2').ionization_energy
>True
```

Should return False

chasepd:
pre-commit.ci autofix

pheuer (approved):
Looks great @chasepd , thanks so much for putting this together!

PR #2672, opened 2024-05-07
Update requirements for binder

### LINKED ISSUE #1635: Fix problems with opening binder links

namurphy:
This is an attempt to fix the problem with opening binder links as described in #1635.

### DIFF
--- a/binder/requirements.txt
+++ b/binder/requirements.txt
@@ -1,3 +1,2 @@
-.
-..
+plasmapy
 jupytext
--- a/changelog/2672.bugfix.rst
+++ b/changelog/2672.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the requirements file used by binder to open notebooks.


namurphy:
I wasn't able to get to binder from the PR branch, so I'm going to try merging this and opening a binder link from `main`.

If this doesn't work, then I'll probably remove the unworking binder links for the time being since I don't have a good idea of how else to fix this.

namurphy:
This appears to have worked!  

A caveat is that it takes a significant amount of time for binder to initially spin up an environment that contains PlasmaPy.  However, after the first notebook successfully launches, subsequent notebooks can be opened quickly.

PR #2681, opened 2024-05-09
Add nox sessions for running tests

namurphy:
This PR adds a `tests` 🧪 session to `noxfile.py` which would let us run pytest.  This is intended to replace the current tox environments for running tests.

The `tests` session is parametrized across versions of Python and for:
 - Running all tests (`all`)
 - Running tests except slow tests (`skipslow`)
 - Running all tests with code coverage (`cov`)
 - Running all tests against the minimum versions of direct dependencies (`lowest-direct`).

I set it so that doctests are run only against the newest supported versions of Python and dependencies.  Doctests are not run when code coverage is being performed, since it helps to make sure that the test suite itself is testing everything we need it to.

We are currently running the tox equivalents of the following sessions in our CI:

 - `tests-3.10(lowest-direct)`
 - `tests-3.11(cov)`
 - `tests-3.12(skipslow)`

By default, running `nox` would run the test session for the current version of Python, skipping slow tests.  I suspect that this would be the most frequently run nox session.


tox.ini
```diff
@@ -30,7 +30,9 @@ commands =
     pytest_skipslow: {env:PYTEST_COMMAND} {posargs} -m "not slow"
     pytest_doctests_all: {env:PYTEST_COMMAND} {posargs} --doctest-modules --doctest-continue-on-failure
     pytest_doctests_skipslow: {env:PYTEST_COMMAND} {posargs} --doctest-modules --doctest-continue-on-failure -m "not slow"
-    pytest_cov_all: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}{/}pyproject.toml --cov-append --cov-report xml:coverage.xml
+    pytest_cov_all: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml
+    --cov-config={toxinidir}{/}pyproject.toml
+    --cov-append --cov-report xml:coverage.xml
```
namurphy, line 35:
```suggestion
    pytest_cov_all: {env:PYTEST_COMMAND} {posargs} --cov=plasmapy --cov-report=xml --cov-config={toxinidir}{/}pyproject.toml --cov-append --cov-report xml:coverage.xml
```

noxfile.py
```diff
@@ -59,37 +56,30 @@ def get_requirements_filepath(
...
     session.install("uv >= 0.1.39")
 
-    category_version_resolution = [
-        ("all", maxpython, "highest"),
-        ("docs", maxpython, "highest"),
-        ("tests", minpython, "lowest-direct"),
+    category_version_resolution: list[tuple[str, str, str]] = [
+        ("tests", version, resolution)
+        for version in supported_python_versions
+        for resolution in ("highest", "lowest-direct")
```
namurphy, line 66: Prior to this change, a requirements file was only created for the `lowest-direct` option for Python 3.10.  With this change, this is now done for all supported versions of Python.  This change will enable us to test PlasmaPy against the lowest versions of direct dependencies for all versions of Python (in our weekly tests).

noxfile.py
```diff
@@ -106,11 +93,70 @@ def requirements(session):
...
+test_specifiers: list[nox._parametrize.Param] = [
+    nox.param("all", id="all"),
+    nox.param("with code coverage", id="cov"),
+    nox.param("skip slow tests", id="skipslow"),
+    nox.param("lowest-direct", id="lowest-direct"),
+]
+
+
+@nox.session(python=supported_python_versions)
+@nox.parametrize("test_specifier", test_specifiers)
```
namurphy, line 128: One neat thing about nox is that we can parametrize test sessions with an API that is very close to `@pytest.mark.parametrize`.

The `id` fields here make it so that `test_specifier=` doesn't show up in the list of parametrized nox sessions.

noxfile.py
```diff
@@ -1,28 +1,34 @@
...
 
-import nox
+import sys
 
-nox.options.default_venv_backend = "uv|virtualenv"
-nox.options.sessions = []
+import nox
 
-supported_python_versions = ("3.10", "3.11", "3.12")
+supported_python_versions: tuple[str, ...] = ("3.10", "3.11", "3.12")
```
namurphy, line 22: I was debating about whether the type hint annotations were necessary in this file, but decided that they were since the annotations will help IDEs like PyCharm to highlight configuration problems while we're still making edits to a file.

noxfile.py
```diff
@@ -1,28 +1,34 @@
...
 
 maxpython = max(supported_python_versions)
 minpython = min(supported_python_versions)
 
+current_python = f"{sys.version_info.major}.{sys.version_info.minor}"
+
+nox.options.sessions: list[str] = [f"tests-{current_python}(skipslow)"]
+nox.options.default_venv_backend = "uv|virtualenv"
+
```
namurphy, line 31: 🛗 Moved without changes.

noxfile.py
```diff
@@ -59,37 +53,30 @@ def get_requirements_filepath(
...
-    category_version_resolution += [
-        ("tests", version, "highest") for version in supported_python_versions
+    category_version_resolution: list[tuple[str, str, str]] = [
+        ("tests", version, resolution)
+        for version in supported_python_versions
+        for resolution in ("highest", "lowest-direct")
     ]
 
-    category_flags = {
+    category_flags: dict[str, tuple[str, ...]] = {
```
namurphy, line 66: 📝 An annotation of `tuple[str]` would indicate that the tuple must contain exactly one string, whereas `tuple[str, ...]` indicates any number of strings.  The `...` is not necessary for lists, since lists are mutable.

ejohnson-96 (approved):
👍

ejohnson-96 (changes requested):
Documentation is failing

namurphy:
I ran the failing check again since it looked like an intermittent failure (I think it was due to problems getting dependencies or something), so I'll go ahead and merge it.  

@ejohnson-96 — Thank you for the review, and also for the wonderful dissertation defense! 🎆

PR #2682, opened 2024-05-10
Add `import`, `cff`, and `build` sessions for nox

namurphy:
This PR adds three quick nox sessions:
 - `import`, which installs PlasmaPy and then attempts to run `import plasmapy`
 - `cff`, which validates `CITATION.cff`
 - `build`, which builds a source distribution (sdist) and wheel, and then uses twine to validate them.

📝 The `import` session is based off of an existing tox environment, while the `cff` and `build` are coming from commands in GitHub workflows.

✨ Setting these up in `noxfile.py` probably took me perhaps ∼30–50% of the time it would have taken me to set these up in `tox.ini`, largely because I would have had to go back to the tox docs to look up how to do this.  

While tox is an incredible tool, I never want to edit a `tox.ini` file ever again!  😹

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -174,3 +174,15 @@ def cff(session):
     """Validate CITATION.cff."""
     session.install("cffconvert")
     session.run("cffconvert", "--validate")
+
+
+build_command = ("python", "-m", "build")
+
+
+@nox.session
+def build(session):
+    """Build and verify a source distribution and wheel."""
+    session.install("twine", "build")
+    session.run(*build_command, "--sdist")
+    session.run(*build_command, "--wheel")
+    session.run("twine", "check", "dist/*")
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -160,3 +160,27 @@ def mypy(session):
     session.install("mypy >= 1.10.0", "pip")
     session.install("-r", "requirements.txt")
     session.run(*mypy_command, *mypy_options, *session.posargs)
+
+
+@nox.session(name="import")
+def try_import(session):
```
namurphy, line 166: 📝 `import` is a reserved word in Python, but nox lets us define a name for a session which is different from the function name.  Pretty helpful!

ejohnson-96 (approved):
👍

Issue #2683, opened 2024-05-10
Add a Nox session to build the changelog


namurphy:
We should add a nox session for using towncrier to build the changelog.  At the same time, we should update the changelog guide accordingly to include the command to build the changelog.

We currently have the commands to build the changelog defined [in a workflow](https://github.com/PlasmaPy/PlasmaPy/blob/f35d5b3b16d9ec463153d17c9e5b027415408128/.github/workflows/mint-release.yml#L53).  The nox session shouldn't include the git commands, unless we include sessions with and without git, or had a session posarg to provide this option.

✨ Creating nox sessions for tasks like this will help make it so that we can do all the GitHub workflow tasks locally as well, which can help with maintaining and debugging the code for these tasks.

I'll assign this to myself to go along with the other nox PRs I'm doing.

PR #2664, opened 2024-05-03
Add a nox session for regenerating pinned requirements files used in CI

namurphy:
This PR adds a nox environment that can be used to regenerate requirements files.  It can be run with `nox -s requirements`.  The intention of this nox environment is to eventually replace the current tox environment, but the removal of the tox environment will come after sessions for running tests get added to `noxfile.py`...hopefully soon.

This PR also standardizes the naming scheme for the requirements files, ideally making the names of the files better reflect what they are used for.

All of the files in `ci_requirements/` are auto-generated, so they probably be mostly skipped during the review.

Only a couple pull requests into this, and I'm already hoping I never have to edit a `tox.ini` file ever again! 😹  

## Related issues

#1734, #1735, #2647, #2654, #2656

PR #2656, opened 2024-05-02
Use nox for some testing environments in GitHub workflows

namurphy:
This PR switches some GitHub workflows from being run with tox to being run with nox, including:

 - Documentation
 - mypy
 - linkcheck

## Motivation

Presently, our tox configuration file is difficult to understand and maintain. Using nox lets us have a configuration file that is written in Python.

## Related issues

This PR is a follow-up on #2654. See also #1734 & #1735.

PR #2685, opened 2024-05-13
Switch GitHub workflows used in CI from tox environments to nox sessions

### LINKED ISSUE #1734: Run tests with `nox` instead of `tox`

namurphy:
Most of our GitHub workflows and heretofore been based on tox environments. However, the configuration files for tox is in an INI format, which I've perpetually had trouble understanding and maintaining.  In contrast, nox is a test runner that is directly inspired by tox, with the exception that nox configuration files _are written in Python_. 

I've spent perhaps a tenth as much time working with nox than I have with tox, and I already have a much better understanding of working with nox.  For a Python project, it makes a lot of sense to configure things in Python when possible.  Switching to nox will move us half of the way from removing INI files from our configuration.  (The remaining files is `mypy.ini`, though the configuration for that will eventually go in `pyproject.toml`.)

In the recent past, we've had a few PRs that switched tox environments to nox sessions (#2654, #2664, #2656, #2681, #2683).  This PR **switches several of our GitHub workflows over from using tox environments to using nox sessions.**

Closes #1734 at long last!

> [!IMPORTANT]
> Some of the checks are showing as "Expected" without being performed because I changed the names of some checks in this pull request without changing the names of required checks in the repository settings. So, it's expected that the "Expected" checks should be unexpected. 😅


noxfile.py
```diff
@@ -210,24 +216,24 @@ def mypy(session):
...
     session.run(*build_command, "--sdist")
     session.run(*build_command, "--wheel")
     session.run("twine", "check", "dist/*")
+
+
+@nox.session
+def cff(session: nox.Session):
+    """Validate CITATION.cff."""
+    session.install("cffconvert")
+    session.run("cffconvert", "--validate")
```
namurphy, line 239: 🛗 Moved without change.

.github/workflows/ci.yml
```diff
@@ -13,24 +13,56 @@ on:
...
   DOC_TROUBLESHOOTING: "Warnings like 'reference target not found' occur when Sphinx tries to interpret text as a Python object, but is unable to do so. This warning can often be fixed by surrounding text in double back ticks instead of single back ticks (e.g., by changing `y` to ``y``) so that it gets formatted as an in-line literal. For more information about addressing documentation build failures, please check out the documentation troubleshooting guide at: https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting"
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
 
 
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
```
namurphy, line 24: There were some problems with GitHub Actions today, and our doc build was left hanging for close to an hour.  Since none of the workflows in this file typically last more than 5 minutes, it seemed reasonable to have a timeout that's much shorter than GitHub's default of 360.

.github/workflows/ci.yml
```diff
@@ -13,24 +13,56 @@ on:
...
+          nox_session: tests-3.10(lowest-direct)
+
+        - name: Packaging, Python 3.12, Ubuntu
+          os: ubuntu-latest
+          python: '3.12'
+          nox_session: build
+
+        - name: Import PlasmaPy, Python 3.10, Windows
+          os: windows-latest
+          python: '3.10'
```
namurphy, line 54: Turns out that it is important to have the quotes around `3.10`.  Without them, GitHub interprets the string as $3.1$.  👀

docs/contributing/testing_guide.rst
```diff
@@ -830,8 +825,8 @@ yet able to handle correctly.
 
 To exclude a line from a coverage report, end it with
 ``# coverage: ignore``. Alternatively, we may add a line to
-``exclude_lines`` in the ``[coverage:report]`` section of
-:file:`setup.cfg` that consists of a
+``exclude_lines`` in the ``[tool.coverage.report]`` section of
+:file:`pyproject.toml` that consists of a
```
namurphy, line 829: 🏕️ I noticed that our discussion of coverage settings is out-of-date. Since I was in the area, I decided to leave the campsite nicer than how I found it (hence the 🏕️).

noxfile.py
```diff
@@ -151,6 +154,9 @@ def tests(session, test_specifier: nox._parametrize.Param):
     if session.python == maxpython and test_specifier in {"all", "skipslow"}:
         options += with_doctests
 
+    if gh_token := os.getenv("GH_TOKEN"):
+        session.env["GH_TOKEN"] = gh_token
```
namurphy, line 158: Needed for `test_downloader.py`.

pyproject.toml
```diff
@@ -86,7 +86,6 @@ docs = [
   "pygments >= 2.17.2",
   # install setuptools to get pkg_resources for doc build
   "setuptools; python_version >= '3.12'",
-  # see issue 2497
```
namurphy, line 89: 🏕️ We closed #2497 a while back, but forgot to delete this comment.

namurphy:
I removed labels to test the labeler GitHub Action after #2690. If it worked, the labels should be added back after the next push.

.github/workflows/ci.yml
ejohnson-96: -_- how nice of GitHub 😆

ejohnson-96 (approved):
There are 3 checks waiting, apart from that looks good 👍

namurphy:
> There are 3 checks waiting, apart from that looks good 👍

The three checks that are waiting are because I changed the names of the tests in the GitHub Action, but I haven't yet gone into the repo settings to change the name of which tests are required since that would break all other PRs.  I'll do that right after merging.

Thank you as always for the reviews!

PR #2690, opened 2024-05-14
Update labeler GitHub workflows

namurphy:
This repository makes use of the [labeler](https://github.com/actions/labeler) GitHub Action, which labels pull request based on what files are being changed.  This action saves us a lot of time because we don't need to label issues manually, while resulting in much more consistently applied labels.  The switch to an `src` layout in #2598 broke some of the pull request labeler specifications.  This PR fixes them, and makes some new additions.

I updated the changelog guide for better consistency with labels, and also to update the towncrier build command which was broken.

I also updated the labels to be removed after a pull request is merged.


.github/labeler.yml
```diff
@@ -36,77 +36,126 @@ dependencies:
...
+  - changed-files:
+    - any-glob-to-any-file: [
+      '**/particles/data/**/*',
+      '**/tools/export_ionization_energy.py'
+    ]
+
 plasmapy.analysis:
   - changed-files:
-    - any-glob-to-any-file: 'plasmapy/analysis/**/*'
+    - any-glob-to-any-file: '**/analysis/**/*'
```
namurphy, line 97: 📝 This change makes it so that changes to any file in `src/plasmapy/analysis` or `tests/analysis` makes it so that this label gets applied.

.github/labeler.yml
```diff
@@ -36,77 +36,126 @@ dependencies:
 
 docs:
   - changed-files:
-    - any-glob-to-any-file: ['.readthedocs.yml', 'changelog/*doc*.rst', 'docs/**/*', 'README.md']
+    - any-glob-to-any-file: ['.readthedocs.yml', 'changelog/*doc*.rst', 'docs/**/*']
```
namurphy, line 39: 📝 The `**` here represents any number of directories, including zero.

namurphy:
Merging this now to be able to test if it works.  (One of these years, I'm going to learn how to test GitHub Actions locally...)

Issue #2673, opened 2024-05-08
Rename `binding_energy` property of `Particle` to `nuclear_binding_energy` and add `electron_binding_energy` based on `ionization_energy`


pheuer:
### Feature description

The current property name is ambiguous, and will lead to confusion if we add an `electron_binding_energy` attribute based on the new `ionization_energy` attribute! 

### Motivation

Reducing confusion and adding a new useful attribute!

### Implementation strategy

1) Rename the existing property
2) Implement `electron_binding_energy` by summing up the ionization energy for all charge states above the given ionization, e.g. 
Particle('C+3`).electron_binding_energy = sum(Particle(p).ionization_energy for p in ['C +4', 'C+5', 'C+6'] )

### Additional context

See #2657 for the ionization energy attribute! Only some small additional logic would be required to calculate the electron binding energy for a given element or ion.

PR #2693, opened 2024-05-16
Issue 2673: Rename `binding_energy` to remove ambiguity, add `electron_binding_energy`

chasepd:
## Description

Adds requested feature from https://github.com/PlasmaPy/PlasmaPy/issues/2673.

- Renames `binding_energy` to `nuclear_binding_energy`, but leaves `binding_energy` as a pointer to `nuclear_binding_energy` for backwards compatibility with a warning that it is deprecated and will be removed in a future release.
- Adds `electron_binding_energy`, defined as the sum of `ionization_energy` for all charge states above the given ionization. 

## Motivation and context

See https://github.com/PlasmaPy/PlasmaPy/issues/2673. With the addition of ionization energy data, calculating `electron_binding_data` becomes fairly easy. However, this causes `binding_energy` to be ambiguous/confusing, and is renamed to `nuclear_binding_energy` for clarity and greater specificity.

## Related issues

Resolves https://github.com/PlasmaPy/PlasmaPy/issues/2673.


chasepd:
@pheuer - here's a new PR from a fresh branch!

pheuer:
> @pheuer - here's a new PR from a fresh branch!

Thank you!!! Here's a couple comments: 

1) Can you make it work for isotopes, similar to the `ionization_energy` property? E.g. 

```python
Particle("C +3").electron_binding_energy == Particle("C-12+3").electron_binding_energy
>True
```
Again, should work for everything above hydrogen, but return separate values for the hydrogen isotopes since the ionization energies are measurably different there. 

2) I think you might be adding one too few ionization energies onto the list? 

```python
Particle("C +3").electron_binding_energy.to(u.eV) 
> 882.08 eV 
```

But according to the NIST database, that's the ionization energy of C+4, and C+3 should be `946.57 eV`. Binding energy being the energy required to free every electron the ion has. This is a little confusing, so maybe you could clarify conceptually a bit in the docstring? 

Otherwise this looks great to me, thanks again for doing this!

chasepd:
pre-commit.ci autofix

chasepd:
@pheuer took me a bit to get back to this, but this should have those items fixed!

pheuer (approved):
@chasepd Awesome, this looks great, thanks for adding this feature!!

PR #2694, opened 2024-05-16
Finish transition from tox to Nox for current continuous integration checks and weekly tests

namurphy:
This PR concludes the switchover from tox to Nox.  I wanted to finish this ahead of the PlasmaPy Summer School so that participants won't need to learn how to work with tox (or more specifically, `tox.ini`).  

The changes in this PR include:

 - 📚 I added a Nox session to build docs against the development versions of Sphinx, sphinx_rtd_theme, and nbsphinx.
 - 🧪 I added a Nox session to run tests against the development versions of NumPy, Astropy, xarray, lmfit (since we had a breaking change come up recently), and pandas.
 - 🧪 I switched over weekly tests to entirely using Nox sessions
 - ♻️ I consolidated the separate workflows for building docs and running tests in CI to a single workflow.
 - ✂️ I deleted `tox.ini` in its entirely, and there was much rejoicing.  Using tox has overall been great, except for the INI-formatted configuration files!  (We still have `mypy.ini`...but I'm hoping that's only temporary.)
 - ✂️ I removed `requirements.txt`, as inspired by #2647.  It's not being used in CI anymore or by any Nox sessions, so we don't need it.  Plus, IDEs are much more adept at getting requirements information from `pyproject.toml` than they used to be.
 - ✂️ I removed the requirements files used only by tox environments. These have been superseded by the requirements files used by Nox sessions.
 - ✨ I made it so that weekly tests can be run for PRs that are labeled with "run weekly tests in CI".  

Some notes:
 - There were a few checks that I wasn't able to get working.  
   - 🧪 Testing against the lowest-direct versions of dependencies for Python 3.11 & 3.12.  
   - 🧪 Testing against the development version of h5py.
   - 🌠 Building the docs on Windows or macOS, since the `sudo apt install` step for pandoc and graphviz is specific to Ubuntu.
   - 🦺 I added `mypy` to the `tests` dependency set, primarily because then we won't have requirements information defined in `noxfile.py` that we have to update.

Remaining tasks
 - [x] Clean up `noxfile.py`
 - [x] Add more changelog entries
 - [x] Possibly spin off some of the smaller, more isolated changes into distinct PRs.

PR #2695, opened 2024-05-17
Add Nox sessions for running pre-commit and checking `MANIFEST.in`

namurphy:
- I added a `lint` session for Nox as a shortcut to run `pre-commit run --all-files`.  This is mostly syntactic sugar, but does remove the need to install `pre-commit` manually.
 - I added the `manifest` session for Nox to run `check-manifest`, which checks that all appropriate files are included in `MANIFEST.in`.  
 - I added a configuration for `check-manifest` in `pyproject.toml`
 - I updated `MANIFEST.in` accordingly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,11 @@ jobs:
           python: '3.12'
           nox_session: cff
 
+        - name: Check MANIFEST.in
+          os: ubuntu-latest
+          python: '3.12'
+          nox_session: manifest
+
     steps:
 
     - name: Checkout code
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,27 +1,29 @@
+graft changelog
+graft ci_requirements
 graft docs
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -64,6 +64,11 @@ jobs:
           python: '3.12'
           nox_session: cff
 
+        - name: Check MANIFEST.in
+          os: ubuntu-latest
+          python: '3.12'
+          nox_session: manifest
+
```
namurphy, line 71: This makes it so that `MANIFEST.in` gets checked for every PR, but we can switch it to a weekly test after #2694 is merged.

pheuer (approved):
I only about 50% understand what is going on with this tox->nox transition, but insofar as this PR defines a few new sessions for checking the manifest and running pre-commit, I _think_ I understand and it looks good :)

namurphy:
Awesome, thank you for the review!

PR #2699, opened 2024-05-23
Fix z-dependence in ion contribution of optical Thomson scattering `spectral_density` function.

### LINKED ISSUE #2698: Apparent bug in ion contribution to Thomson scattering spectrum
We received a bug report over email.  

> I believe there may be a bug in the computation of the ion contribution for the Thomson scattering in diagnostics/thomson.py.  Specifically, the contribution from the species' charge ([line 247](https://github.com/PlasmaPy/PlasmaPy/blob/main/src/plasmapy/diagnostics/thomson.py#L247)) appears to be slightly incorrect. Rather than `ion_z​` it should be `ion_z**2 / zbar​` according to [Sheffield](https://www.sciencedirect.com/book/9780123748775/plasma-scattering-of-electromagnetic-radiation) (Section 5.1, Eq. 5.1.2 and following on page 104 - see screenshot below)
unnamed

@pheuer — could you look into this?

pheuer:
A user very helpfully pointed out a small error in the implementation of the Thomson scattering spectral density function. This PR updates our code to match Sheffield, as detailed in #2698 

The changes are: 

1) Re-defined `zbar` in [line175](https://github.com/PlasmaPy/PlasmaPy/blob/534b8e960050488db4c9e2977bc4142388afc153/src/plasmapy/diagnostics/thomson.py#L175) to be the array $\bar z_j$, not the sum $\sum \bar z_j$. As a result, added a sum to the following line. 

2) Modified the ion contribution as suggested on [line 247](https://github.com/PlasmaPy/PlasmaPy/blob/534b8e960050488db4c9e2977bc4142388afc153/src/plasmapy/diagnostics/thomson.py#L247) from $z_j$ to $z_j^2 / \bar z_j$

3) We have a test that compares some metrics of a spectrum against some previously recorded values - this PR slightly changes the spectrum for the multi-ion species test, so I have updated that fixed value accordingly. 

Closes #2698


changelog/2699.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Changed the z dependence of the ion contribution to the optical Thomson scattering
+spectral density function in `~plasmapy.diagnostics.thomson.spectral_density_lite` from ``z``
+to ``z^2 / zbar`` to match Eq. 5.1.2 and following equations in Plasma Scattering of Electromagnetic Radiation
+by Sheffield et al. (2011). The result is a small change in the ion acoustic wave spectrum for plasmas
```
namurphy, line 4:
```suggestion
to ``z^2 / zbar`` to match Eq. 5.1.2 and following equations in :cite:t:`sheffield:2011`. 
The result is a small change in the ion acoustic wave spectrum for plasmas
```
📚 We could use the bibliography!

namurphy (approved):
Thank you for getting to this so quickly!  I have a few minor suggestions, and overall it looks great to me!

pheuer:
@namurphy Thanks for the suggestions! 

One last thing before we merge - is the person who pointed out the error already listed in the citation file? If not, would they like to be? That certainly counts as a contribution in my book!

namurphy (approved):

pheuer:
> @namurphy Thanks for the suggestions!
> 
> One last thing before we merge - is the person who pointed out the error already listed in the citation file? If not, would they like to be? That certainly counts as a contribution in my book!

@namurphy Merging this because I think it should be in the package, but we should add the contributor to the file if they are interested in that.

PR #2700, opened 2024-05-23
Add `typos` as a pre-commit hook

namurphy:
This PR tries out [typos](https://github.com/crate-ci/typos) as a pre-commit hook as a possible addition to or replacement for [codespell](https://github.com/codespell-project/codespell).  It managed to find a couple more typos.  Since it's written in Rust, it's probably quicker (though I didn't time it).

I'll tentatively add this as a hook without removing codespell, but we can consider removing one of them later if it gets annoying to have to deal with both.

Thanks to @nabobalis for suggesting this!


_typos.toml
```diff
@@ -0,0 +1,19 @@
+[default]
+extend-ignore-identifiers-re = [
```
nabobalis, line 2: Do you want these two sections?

_typos.toml
```diff
@@ -0,0 +1,19 @@
...
+    "AttributeID.*Suppress.*",
+]
+
+[default.extend-identifiers]
+# *sigh* this just isn't worth the cost of fixing
+AttributeIDSupressMenu = "AttributeIDSupressMenu"
+
+[default.extend-words]
+aas = "aas"
+arange = "arange"
```
nabobalis, line 13: This was such an annoying thing to have to add to our config as well.

nabobalis (approved):

Issue #2655, opened 2024-05-01
Spin off plasma calculator into separate affiliated package


namurphy:
### Description of improvement

Right now the plasma calculator is installed as part of PlasmaPy.  One possibility in the future would be to split off the plasma calculator into a standalone package.

### Motivation

In the lead-up to the Python in Heliophysics Community Summer School, work is ongoing to create a heliocloud Python environment that contains all PyHC core packages. A problem that arose was that PlasmaPy's dependency on `voila` (and consequently `tornado`) for the plasma calculator ended up conflicting with requirements for other packages that needed web servers.  In a sense, this would let us separate the plasma calculator front-end from the `plasmapy` back-end.  Separating the plasma calculator would decrease the dependency footprint of PlasmaPy.  In particular, we may be able to drop PlasmaPy's dependency on the Jupyter ecosystem, as well as dependencies dealing with web server stuff like `voila` and `tornado`.

### Alternatives

In the short term, we can loosen the requirements for `voila` and other packages.

### Implementation

This would probably be a fair bit of work, and thus I'd place it at low priority.

PR #2705, opened 2024-05-29
Loosen dependencies on `ipykernel`, `ipywidgets`, and `voila` ⬇️ 

namurphy:
This PR loosens requirements on `ipykernel`, `ipywidgets`, and `voila` to be consistent with what is used by Google Colab, which is currently `v5.5.6`.  See #2655 the motivation why.

We should strive to keep the minimum version of `ipykernel` to be no greater than the version used by Google Colab (which tends to run older versions of a bunch of packages, and also Python).


namurphy:
Got this error: 
```
  File "D:\a\PlasmaPy\PlasmaPy\.nox\tests-3-10-lowest-direct\lib\site-packages\_pytest\config\__init__.py", line 802, in import_plugin
    __import__(importspec)
ModuleNotFoundError: No module named 'pytest_tornasync'
```
I'm not sure why this came up, but will look into it later.

PR #2704, opened 2024-05-29
Update synthetic radiography `Tracker` class to use `ParticleTracker`

JaydenR2305:
In https://github.com/PlasmaPy/PlasmaPy/pull/2245 we implemented a new particle tracker class that was meant to generalize the already written radiography tracker. This pull request partially rewrites the synthetic radiography tracker to use this new particle tracker class.

This pull request is a duplicate of #2493, which implemented the same changes prior to the switch to the src/ layout. This pull request manually copies the changes made in the previous PR to a fork with the correct structure.

PR #2708, opened 2024-05-31
Revive Relativistic Boris Push

JaydenR2305:
Reimplements the relativistic Boris push integrator defined in #979 using the new `ParticleTracker` framework

Must be merged after #2704


JaydenR2305:
@pheuer What are your thoughts on defining some sort of integrator class? I'm just thinking we want to make sure the integration functions all have a consistent signature so that they can be easily called from the `_push()` function.. Is there a way of doing this that doesn't require creating a new class hierarchy?

JaydenR2305:
image
image
image
image

JaydenR2305:
It looks like the discrepancy in the factor of `c` came from CGS units

image

pheuer:
> It looks like the discrepancy in the factor of `c` came from CGS units

Not surprising, that happens all the time! 

These plots look great! Maybe we can have one test case as v/c=0.001 (show the pusher gives the same result for non-relativistic particles) then one at v=0.999 (show that the pusher now matches the analytical solution)?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -119,9 +136,14 @@ def __init__(
         dt=None,
         dt_range=None,
         field_weighting="volume averaged",
+        integrator="explicit_boris_relativistic",
```
pheuer, line 139: I think it would be better if the tracker just switched over automatically if the simulation was relativistic? We have the velocities, and there's no circumstance where someone would want to use the wrong integrator?

JaydenR2305:
There does seem to be a discrepancy with the fitting for the classical trajectory?
image

whereas the relativistic fits quite nicely
image

JaydenR2305:
figured it out! There was some dependence on the initial velocity. I've set the initial velocity to zero for now until we can figure out a better alternative (setting it directly to some fraction of the final drift velocity wasn't working...) This just means that we have to discard the first 20 points or so to allow the particle to reach the expected drift velocity.

pheuer:
> figured it out! There was some dependence on the initial velocity. I've set the initial velocity to zero for now until we can figure out a better alternative (setting it directly to some fraction of the final drift velocity wasn't working...) This just means that we have to discard the first 20 points or so to allow the particle to reach the expected drift velocity.

Awesome! Kind of concerning it doesn't work with an initial velocity thought... any idea why that is? Want to make sure that's not an issue with the analytic solution or the pusher or something.

JaydenR2305:
> Awesome! Kind of concerning it doesn't work with an initial velocity thought... any idea why that is? Want to make sure that's not an issue with the analytic solution or the pusher or something.

image
I think the derivation makes the assumption that the particle is starting from rest

pheuer:
>I think the derivation makes the assumption that the particle is starting from rest

Ah, got it, ok that's expected then! In that case I'd say this test looks good enough?

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -186,6 +186,10 @@ class Tracker(ParticleTracker):
         simulation. This does not include particles that have never entered
         the grids.
 
+    relativistic_beta_threshold: `float`, optional
+        The threshold fraction of the speed of light, which once exceeded, will
+        trigger the simulation to switch to a relativistic Boris push.
```
pheuer, line 191: Put the default value here, maybe 1%?

src/plasmapy/simulation/particle_integrators.py
```diff
@@ -150,3 +153,71 @@ def boris_push(x, v, B, E, q, m, dt, inplace: bool = True):
...
+    m : float
+        particle mass, in SI (kg) units.
+    dt : float
+        timestep, in SI (second) units.
+
+    Notes
+    -----
+    For the basic overview of this algorithm, see `boris_push`. This
+    version, based on [1]_, applies relativistic corrections such as
+    TODO.
```
pheuer, line 183: TODO

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -29,6 +31,20 @@
 )
 
 
+class _IntegratorInfo(TypedDict):
```
pheuer, line 34: Hm, I see why you wanted to make these classes. You could do that, and just have each have an `integrate` method? Just as a way to store this information more naturally? 

Otherwise I'd say put this in the file with the push algorithm, but the class idea is better

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -77,6 +93,10 @@ class to the ``run`` method.
         If specified, the calculated adaptive time step will be clamped
         between the first and second values.
 
+    relativistic_beta_threshold: `float`, optional
+        The threshold fraction of the speed of light, which once exceeded, will
+        trigger the simulation to switch to a relativistic Boris push.
```
pheuer, line 98: Same comment here - put default value in the docstring

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -118,10 +138,18 @@ def __init__(
...
         field_weighting="volume averaged",
         req_quantities=None,
         verbose=True,
     ) -> None:
+        # By default, set the integrator to the explicit Boris push
+        # The `_push()` method may change this to a relativistic integrator if the energies of the particles
+        # exceed the threshold specified by `relativistic_beta_threshold`
+        self._integrator = _INTEGRATORS["explicit_boris"]
+        self._beta_threshold = relativistic_beta_threshold
+        self._is_relativistic = False
```
pheuer, line 151: If you do make the integrator a class, then `_is_relativistic can just be a property 

def _is_relativistic(self):
return self._integrator.relativistic

tests/simulation/test_particle_tracker.py
```diff
@@ -438,3 +440,141 @@ def test_particle_tracker_stop_particles(request):
...
+        # Calculate the positions
+        # Eq. 71
+        a = γd * vd / ν
+        x = (a * γd * (ν * 𝜏 - np.sin(ν.si.value * 𝜏.si.value))).to(u.m)
+        z = (a * (np.cos(ν.si.value * 𝜏.si.value) - 1)).to(u.m)
+
+    return x, z
+
+
+def construct_field(
```
pheuer, line 503: Personal preference - if this is only used in the below test function, maybe define it as local inside that function?

tests/simulation/test_particle_tracker.py
```diff
@@ -438,3 +440,141 @@ def test_particle_tracker_stop_particles(request):
...
+
+@pytest.mark.parametrize(
+    "regime",
+    [
+        0.01,
+        0.5,
+        0.99,
+    ],
+)
+def test_Boris_integrator_fitting(regime):
```
pheuer, line 524: This test looks great, thanks for working out those last bugs!

tests/simulation/test_particle_tracker.py
```diff
@@ -438,3 +440,141 @@ def test_particle_tracker_stop_particles(request):
...
+    ----------
+    """
+    save_routine = IntervalSaveRoutine(0.1 * u.ns)
+
+    simulation = ParticleTracker(
+        grids=fields,
+        save_routine=save_routine,
+        termination_condition=termination_condition,
+    )
+    simulation.load_particles(
```
pheuer, line 559: Can you make a test case with an electron, too? Since that's the most likely relativistic use case

pheuer (commented):
Looking good! Few minor comments before we merge

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: Yeah the class idea is definitely the way to go here. Otherwise the typing gets a little dubious because Python's typing for `Callable` is very messy https://docs.python.org/3/library/typing.html#annotating-callable-objects

pheuer: Ok sounds great! Maybe you can prioritize these changes to this PR today and we can try to get this in today/early next week? Looks pretty close to done to me.

tests/simulation/test_particle_tracker.py
JaydenR2305: Good idea! I actually caught an interesting edge case with this. In `_adaptive_timestep` we were calculating the gyroperiod using the expression:
```py
gyroperiod = 2 * np.pi * self.m / (self.q * np.max(Bmag))
```
which doesn't play well with electrons. I wrapped this in an `np.abs()` which should fix the issue

pheuer: Oops!! Yep that's a problem. To be explicit where the sign issue is, you could wrap just q in the abs, `np.abs(q)`

src/plasmapy/simulation/particle_integrators.py
JaydenR2305: I added some more details here. Let me know what you think! (I have next to no experience with relativity)

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: So I have it implemented but I'm not really happy with how I handle `is_relativistic`: https://github.com/PlasmaPy/PlasmaPy/blob/c6aee176c601687992d25e50684ad2936ff33a82/src/plasmapy/simulation/particle_integrators.py#L24-L28

I need a clean way of indicating that this property has to be defined by specifying it in the abstract class somewhere while also not requiring the class to be instantiated-- something like an "abstract property"

pheuer: Does abstract property from ABC not do what you want? https://docs.python.org/3/library/abc.html#abc.abstractproperty

src/plasmapy/simulation/particle_integrators.py
pheuer: Looks good to me!

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: It does, but part of the issue is that we shouldn't be requiring the user to instantiate the integrator (which is required with abstract properties.) I might just stick with the `TypedDict` approach and expose the `AbstractIntegrator` as such a dict

pheuer: I had assumed the classes would all be defined here, then in tracker the function would take a keyword argument to specify which tracker to use, then instantiate it internally, `self._integrator = BorisIntegrator()` then use that for the push function? Maybe I don't understand the issue here?

The documentation is a good thing to think of: at least in this module, we want documentation for each integrator method which the classes seem a natural match for.

JaydenR2305: I'll switch it over to instantiate the integrator, definitely makes it easier to ensure the relevant documentation is included

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: According to the profiler library they both have a percall time of 0.000 so I think it's safe to say they're close enough? What do you think

tests/simulation/test_particle_tracker.py
JaydenR2305: What are your thoughts on moving all of the relevant functions into a class? https://github.com/PlasmaPy/PlasmaPy/blob/2f25b6c847caa2d7a69d815af522f116d8c8772b/tests/simulation/test_particle_tracker.py#L445-L662

pheuer: Looks like a good idea to me!

pheuer (approved):
This looks great, thanks for polishing off this neglected old PR!!

PR #2710, opened 2024-06-03
Make `Downloader` Test Suite Resilient to `ReadTimeout`

### LINKED ISSUE #2711: Flaky Failure in `test_floating_potential`
The `test_floating_potential` module occasionally emits a `JSONDecodeError`. See https://github.com/PlasmaPy/PlasmaPy/actions/runs/9352991385/job/25742427527#step:6:129

It is unclear which of the tests this error is originating from.. usually the console mentions this information?

JaydenR2305:
Resolves `requests.exceptions.ReadTimeout` errors originating from downloader-related tests

Upon merging #2555 it was discovered that one of the associated tests can be flaky (see https://github.com/PlasmaPy/PlasmaPy/actions/runs/9349808820/job/25731939682?pr=2709#step:6:310). ~~I propose we increase the timeout of the get request to 40 seconds to give the server ample time to reply.~~

Additionally, @pheuer suggested implementing a check to ensure if a connection can be established to the server within some timeframe, and simply `@pytest.mark.skip` any relevant tests if ([here](https://pytest.org/en/6.2.x/reference.html#pytest-mark-skipif) for my reference) a connection cannot be established.

Resolves #2711


pheuer:
@JaydenR2305 Can you try leaving the timeout at 10s and instead doing something like 

```python

@pytest.fixture
def check_github_connection():
"""returns True if good connection to github, False otherwise"

@pytest.skipif(check_github_connection) 
def test_any_test_using_api_calls

```

My concern with increasing the timeout so long is that it will make CI take that much longer to run, and if a connection really takes > 10 s I'm not optimistic it will complete in 40s?

JaydenR2305:
@pheuer Any idea what's going on with the ``test_floating_potential`` test? I feel like I've seen it error before like this out of the blue.

pheuer:
> @pheuer Any idea what's going on with the `test_floating_potential` test? I feel like I've seen it error like this before out of the blue.

I don't think so - is that in the Langmuir module? You might open an issue about that. Maybe there's a random component in the test (or using hypothesis) that is failing occasionally on some edge value?

src/plasmapy/utils/data/downloader.py
```diff
@@ -13,14 +13,24 @@
...
+__all__ = ["_API_CONNECTION_ESTABLISHED", "Downloader"]
 
+_API_CONNECTION_ESTABLISHED = False
 _IS_CI = "GH_TOKEN" in os.environ
 
 
 # TODO: use a config file variable to allow users to set a location
 # for the data download folder?
 
+try:
```
pheuer, line 25: I suppose this has to be in this main file rather than in the test file since multiple different test files (anything using downloader) will need to access it?

JaydenR2305: Yep, we just import it from there into the test modules that use the downloader directly/indirectly

pheuer:
> https://github.com/PlasmaPy/PlasmaPy/blob/4d08880e6b44ab0f596e7c99e867a43513a86104/src/plasmapy/utils/data/downloader.py#L29-L32
> 
> Would it be fine if I just `# coverage: ignore` this?

Yes, I don't see a good way to test this (other than simulating a lost connection somehow, but that seems unnecessarily complicated)

pheuer (approved):
@JaydenR2305 This all looks good to me - ready to merge?

PR #2712, opened 2024-06-03
Add Stopping to `ParticleTracker`

JaydenR2305:
Implements the ability to add particle stopping to a particle tracer simulation. Physical properties of the relevant materials is fed into the simulation using either `add_stopping` or `add_Bethe_stopping`-- for the former, an array of material strings is passed in, for the latter, the mean excitation energy and electron density of the relevant materials are required. The stopping power is retrieved from the NIST stopping power database file housed in the PlasmaPy data repository.

merge #2704 first


JaydenR2305:
@pheuer What are your thoughts on implementing the Bethe formula in addition to providing support for the NIST data? Do you think this unnecessarily clutters things?

pheuer:
> @pheuer What are your thoughts on implementing the Bethe formula in addition to providing support for simply using the NIST stopping data? Do you think this unnecessarily clutters things?

I think that's a great idea! Maybe implement it as its own function in PlasmaPy, with tests against the NIST data? Then you could give users the option of which to use. Analytical might be faster than interpolating from the tables, not sure if the Bethe formula starts to get inaccurate at energies people are actually interested in but if so then they might opt for the tables.

JaydenR2305:
pre-commit.ci autofix

src/plasmapy/particles/atomic.py
```diff
@@ -1159,7 +1159,8 @@ def _get_relevant_stopping_information(
             raise NotImplementedError(
                 "Stopping calculations for electrons have not been implemented yet!"
             )
-        elif incident_particle == Particle("H+"):
+        # TODO: is there a difference between H+ and p+ ??
```
pheuer, line 1162: I don't think so, as long as you take the convention that H+ is the most common isotope with no neutrons, they should be the same? I think @namurphy mentioned some subtlety here to me at some point...

JaydenR2305: For some reason it wasn't catching p+ in the if statement

namurphy: There is indeed a difference between `'H+'` and `'p+'`.  When we use `'H+'`, it uses the standard atomic weight of hydrogen (mostly H-1, but some D), which is how it's handled for all other elements.

This is probably one of the most common mix-ups 😕, so I'm wondering if we should have a warning when `Particle` is provided with `'H+'`...or at the very least, better note this in the documentation.

JaydenR2305: What if we just check the charge and mass of the particle that is passed in?

namurphy: Another possibility would be `incident_particle in ParticleList(["H+", "p+"])`.

pheuer: I think Nick's solution is more Plasmapythonic :)

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -413,6 +436,67 @@ def load_particles(
...
+        """
+
+        if len(materials) != len(self.grids):
+            raise ValueError(
+                "Please provide an array of length ngrids for the materials."
+            )
+
+        # Require that each grid has a defined mass density,
+        # if a grid does not define rho, raise an exception
+        # TODO: should this be replaced with only raising an exception when none
```
pheuer, line 488: Remember this function replaces any missing values with zero arrays, so it depends how you handle it later. Probably the fastest way is to not require it, but then only do the stopping calculation if it is present? 

I would raise a warning here if NO grid has rho...

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -602,6 +686,17 @@ def particles_on_grid(self):
 
         return all_particles
 
+    @property
+    def _particle_kinetic_energy(self):
+        r"""
+        Return the non-relativistic kinetic energy of the particles.
+        """
+
+        # TODO: how should the relativistic case be handled?
```
pheuer, line 695: Good question... no idea how relativistic stopping works! 

How about adding an 'is relativistic' property to the base Tracker (maybe like 1% C?) and then raise an exception if you try to use stopping in that regime? Until it's explicitly handled.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -621,70 +716,68 @@ def _push(self) -> None:
         # entered_grid is zero at the end if a particle has never
         # entered any grid
         self.entered_grid += np.sum(self.particles_on_grid, axis=-1).astype(np.bool_)
+        # TODO: should this interpolation step be broken into a separate method
```
pheuer, line 719: No this is a great refactor! But add some comments explaining what's happening in each step a little more clearly to counteract the lower transparency of this compact phrase.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -706,6 +799,55 @@ def _push(self) -> None:
...
+        #     )
+        #     _S = np.nan_to_num(_S, nan=0.0 * u.J / u.m)
+        #
+        #     S += _S
+        speeds = np.linalg.norm(self.v[tracked_mask], axis=-1)[:, np.newaxis]
+        dx = np.multiply(speeds, dt)
+        stopping_power = np.zeros((self.nparticles_tracked, 1))
+
+        for cs in self._stopping_power_interpolators:
+            # TODO: is this the proper way to handle the addition of multiple stopping powers?
```
pheuer, line 822: This should be about right! Stopping is really particle-particle interactions: the lattice structure etc. of the material only matters insofar as it affects the density and thus the probability of a collision. So, if you have a mixture of gold and aluminum, say, I think the decrease in energy should just add, e.g. 

dE = (sp1*rho1 + sp2*rho2)*dx

JaydenR2305:
https://github.com/PlasmaPy/PlasmaPy/blob/749b9dc9d10f8dcae8dcadef6ff4c8cfd1070355/src/plasmapy/simulation/particle_tracker/particle_tracker.py#L850

oh man

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -621,70 +716,68 @@ def _push(self) -> None:
         # entered_grid is zero at the end if a particle has never
         # entered any grid
         self.entered_grid += np.sum(self.particles_on_grid, axis=-1).astype(np.bool_)
+        # TODO: should this interpolation step be broken into a separate method
+        #  for the sake of simplicity?
+        # TODO: how should we handle unrecognized quantities?
```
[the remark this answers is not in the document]
JaydenR2305: Yeah I was just looking at the fact that `_push` could be broken down into interpolation, push, stopping, and (eventually) scattering

JaydenR2305:
Another issue is that the `synthetic_radiograph` function was displaying stopped particles. Is it save to hide these from view if `ignore_grid=True`?

JaydenR2305:
> Another issue is that the `synthetic_radiograph` function was displaying stopped particles. Is it save to hide these stopping particles from view if `ignore_grid=True`?

(I made this adjustment and it caused the optical intensity test to fail https://github.com/PlasmaPy/PlasmaPy/actions/runs/9451367781/job/26032127312?pr=2712#step:6:244)

JaydenR2305:
pre-commit.ci autofix

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -706,6 +799,55 @@ def _push(self) -> None:
...
+        dE = -np.multiply(energy_loss_per_length, dx)
+
+        # Update the velocities of the particles using the new energy values
+        # TODO: again, figure out how to differentiate relativistic and classical cases
+        E = self._particle_kinetic_energy[tracked_mask, np.newaxis] + dE
+
+        speeds = np.sqrt(2 * E / self.m)
+        unit_vectors = np.multiply(1 / speeds, self.v[tracked_mask])
+
+        self.v[tracked_mask] = np.multiply(speeds, unit_vectors)
```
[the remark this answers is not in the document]
pheuer: I'm not sure I understand why the second mask is necessary? Particles that aren't tracked shouldn't be stopping, so they never need to be selected by this statement.

pheuer:
> Another issue is that the `synthetic_radiograph` function was displaying stopped particles. Is it save to hide these stopping particles from view if `ignore_grid=True`?

Hm, I'd say always ignore them, but then there's no way to ignore just the grid. Maybe that keyword should just go away >< Otherwise we have to keep track of WHAT stopped the particles and that sounds like a slippery slope.

pheuer:
> > Another issue is that the `synthetic_radiograph` function was displaying stopped particles. Is it save to hide these stopping particles from view if `ignore_grid=True`?
> 
> (I made this adjustment and it caused the optical intensity test to fail https://github.com/PlasmaPy/PlasmaPy/actions/runs/9451367781/job/26032127312?pr=2712#step:6:244)


Odd - there shouldn't be any stopping in that test, right?

Are NaNs in the positions propagating into the histogram and causing problems in that part of syntehtic_radiograph? If so, maybe using functions like nanmean instead of mean would help, or masking nan->0 in the histogram?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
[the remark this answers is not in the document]
JaydenR2305: The calculated change in energy (and the total energy by extension) is calculated only for the tracked particles (nparticles_tracked). We can use this to select out particles with negative energy but the mask we pass to stop_particles() needs to have shape nparticles. We construct this by initializing an empty book array of length nparticles and setting the indices within particles tracked equal to the E < 0 mask. 

Does that make more sense?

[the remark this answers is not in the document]
pheuer: Yes I think so - just a bit odd that stop_particles takes a length nparticles since the only the tracked ones can be stopped - unless it's also used for wire grids maybe?

JaydenR2305:
> Odd - there shouldn't be any stopping in that test, right?

No it was more so that I was getting problems with my test cases where I visualize the stopping using a synthetic radiograph and make sure the intensity is zero everywhere. Correcting this by hiding the particles with NaN velocity caused the test to fail but provided the expected intensities when stopping is included.

The test seems to pass when we hide these particles for `ignore_grid=False`

pheuer:
> I think particles are being stopped in the test because of the `max_theta` criterion?

Oh, I guess they are getting advanced to the detector then `stopped` prior to running the tracing?

src/plasmapy/particles/atomic.py
```diff
@@ -1225,17 +1193,76 @@ def stopping_power(
...
+    Notes
+    -----
+    The data for stopping power is taken from the National Institute of
+    Standards and Technology's Stopping-Power and Range Tables :cite:p:`niststar:2005`.
+    Valid materials can be found on the NIST STAR website. The default energies
+    are taken from the data points in the STAR database.
+
+    """
+
+    if energies is None:
```
pheuer, line 1260: I don't really understand the need to split this into several functions: can't you just: 

1) Get the relevant stopping power data
2) If energies is None, just return that
3) If energies is not None, then interpolate that over the requested energies and return that

It seems like it can be done in one function? Is this a function complexity issue?

JaydenR2305: The `ParticleTracker` uses the  `_stopping_power_interpolator` function to get the interpolator for each material. These interpolators would otherwise be constructed on every time step-- I think this is similar to something we have in the grid object, something about a persistent interpolator?

JaydenR2305: note to self: recombine into one function with `return_interpolator` keyword

JaydenR2305:
@pheuer Do you mind if I remove Bethe stopping from this PR and make it the focus of another? I'm just anticipating having to change around a few signatures in order to incorporate an external function for calculating stopping powers. Maybe these stopping routines should eventually be reorganized to look like our integrator classes

pheuer:
> @pheuer Do you mind if I remove Bethe stopping from this PR and make it the focus of another? I'm just anticipating having to change around a few signatures in order to incorporate an external function for calculating stopping powers. Maybe these stopping routines should eventually be reorganized to look like our integrator classes

Sounds like a good idea.

JaydenR2305:
New `_push` loop:
https://github.com/PlasmaPy/PlasmaPy/blob/b11b4925398ff310e507c917c17fb11ca6367ce4/src/plasmapy/simulation/particle_tracker/particle_tracker.py#L851-L871

JaydenR2305:
pre-commit.ci autofix

JaydenR2305:
I think this is a bug with mypy? 
https://github.com/PlasmaPy/PlasmaPy/blob/b52e767e52bd8d1f6b0d3450c57eb179fc549edc/src/plasmapy/particles/atomic.py#L1237-L1258
https://github.com/PlasmaPy/PlasmaPy/actions/runs/9714784258/job/26814672404#step:7:32
The `energies` argument is guaranteed to not be `None` since we implicitly rule out this possibility through the combination of two if statements. Explicitly writing out `if energies=None:` makes the error go away:
https://github.com/PlasmaPy/PlasmaPy/blob/94146258aafbbf6ff0488ed87054299dd1191eac/src/plasmapy/particles/atomic.py#L1237-L1261

Am I missing something here?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -73,6 +75,11 @@ class to the ``run`` method.
         time steps of the simulation to save. The default is `~plasmapy.simulation.particle_tracker.save_routines.DoNotSaveSaveRoutine`.
         See `~plasmapy.simulation.particle_tracker.save_routines.AbstractSaveRoutine` for more details.
 
+    particle_integrator : `~plasmapy.simulation.particle_integrators.AbstractIntegrator`, optional
+        An instance of `~plasmapy.simulation.particle_integrators.AbstractIntegrator` that is responsible for implementing the push behavior
```
pheuer, line 79:
```suggestion
        A subclass of `~plasmapy.simulation.particle_integrators.AbstractIntegrator` that is responsible for implementing the push behavior
```

Right?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -429,12 +463,62 @@ def load_particles(
...
+                "The density is not defined on any of the provided grids! Particle stopping will not be "
+                "calculated.",
+                RuntimeWarning,
+            )
+
+            # Don't set `_do_stopping`. The push loop does not have to do stopping
+            # calculations
+            return
+
+        raise ValueError("Requiring rho!")
```
pheuer, line 502: This shouldn't be here, right? Probably from debugging?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -429,12 +463,62 @@ def load_particles(
...
+        # Require that each grid has a defined mass density,
+        # if a grid does not define rho, raise an exception
```
pheuer, line 504: It doesn't raise an exception does it? It just defines it as zeros?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
-        Ez = np.zeros(self.nparticles_tracked) * u.V / u.m
-        Bx = np.zeros(self.nparticles_tracked) * u.T
-        By = np.zeros(self.nparticles_tracked) * u.T
-        Bz = np.zeros(self.nparticles_tracked) * u.T
+        # TODO: how should we handle unrecognized quantities?
+
+        # Construct a dictionary to store the interpolation results
+        # This is accomplished using a dictionary comprehension. Each quantity is
+        # initialized as a zeros array with its respective units
+        summed_field_values = {
```
pheuer, line 732: Name confused me for a bit - this is the total field across all grids? Maybe total_grid_values? 

Also field->grid or something, since it's not just the fields anymore (rho)

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
+            for field_value, field_name in zip(
+                field_values, summed_field_values.keys(), strict=False
+            ):
+                summed_field_values[field_name] += np.nan_to_num(
+                    field_value,
+                    0.0 * AbstractGrid.recognized_quantities[field_name].unit,
+                )
 
         # Create arrays of E and B as required by push algorithm
         E = np.array(
```
pheuer, line 760: Maybe this should be done in the _update_position method, since it is just as input to the integrator? Then you can just return the summed fields variable from this fcn

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
 
-        # Update the tracked particles using the integrator specified at instantiation
-        # TODO: implement "tentative" x and v to prevent speeds faster than light
-        #  from occurring over one step. Possibly based on field strengths?
-        self.x[tracked_mask], self.v[tracked_mask] = self._integrator.push(
-            pos_tracked, vel_tracked, B, E, self.q, self.m, dt
+        return dt
+
+    def _update_position(self, B, E):
+        # TODO: should this be generalized to allow for more fields to be
```
pheuer, line 807: For now this is fine - I think E&B is all the pusher will take.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
+        x_results, v_results = self._integrator.push(
+            pos_tracked, vel_tracked, B, E, self.q, self.m, self.dt
+        )
+
+        self.x[self._tracked_particle_mask], self.v[self._tracked_particle_mask] = (
+            x_results,
+            v_results,
+        )
+
+    def _update_velocity_stopping(self, summed_field_values):
```
pheuer, line 820: Docstring

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
             self.time += dt
 
-        # Update the tracked particles using the integrator specified at instantiation
-        # TODO: implement "tentative" x and v to prevent speeds faster than light
-        #  from occurring over one step. Possibly based on field strengths?
-        self.x[tracked_mask], self.v[tracked_mask] = self._integrator.push(
-            pos_tracked, vel_tracked, B, E, self.q, self.m, dt
+        return dt
+
+    def _update_position(self, B, E):
```
pheuer, line 806: Add docstrings for all these submethods  - just a single sentence defining what it does.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
-        self.dt = dt
+        new_speeds = np.sqrt(2 * E / self.m)
+        self.v[self._tracked_particle_mask] = np.multiply(
+            new_speeds, velocity_unit_vectors
+        )
+
+        self._stop_particles(particles_to_be_stopped_mask)
+
+    def _push(self) -> None:
+        r"""
```
pheuer, line 874: Update docstring? Now uses whatever integrator you supply?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
+
+        # Interpolate fields at particle positions
+        summed_field_values, E, B = self._interpolate_grid()
+
+        # Calculate an appropriate timestep (uniform, synchronized)
+        self.dt = self._update_time(summed_field_values)
+
+        # Update the position and velocities of the particles using timestep
+        # calculations as well as the magnitude of E and B fields
+        self._update_position(B, E)
```
pheuer, line 887: Maybe some slight OCD here, but it bothers me that you have "E,B" above, then "B,E" here! Try to keep the order the same so the signatures are easier to remember. 

IMO E,B is the correct order ;)

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -620,117 +703,205 @@ def particles_on_grid(self):
...
+            if beta_max >= 0.01:
+                self._raised_relativity_warning = True
+
+                warnings.warn(
+                    f"Particles have reached {beta_max}% of the speed of light. Consider using a relativistic integrator for more accurate results.",
+                    RelativityWarning,
+                )
+
+        # Update velocities to reflect stopping
+        if not self._do_stopping:
```
pheuer, line 901: We are going to add more to this function later - you don't want it to stop here if you are doing scattering but not stopping. Why not do 

```
if stopping: 
   self._update_velocity_stopping
```
?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -429,12 +453,99 @@ def load_particles(
...
+        r"""
+        Check to ensure the provided quantity string is defined on at least one grid.
+
+        Returns ``True`` if the quantity is defined on at least one grid and
+        ``False`` if none of the grids have defined the specified quantity.
+        """
+
+        return any(quantity in grid.quantities for grid in grids)
+
+    def add_stopping(self, materials: list[str]):
```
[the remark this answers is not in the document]
JaydenR2305: What if we handled this like the termination condition and saving routines? That way the user can define their own stopping if they wanted

JaydenR2305:
image
Here's a recreation of a plot from the Wikipedia entry for the Bethe formula (https://en.wikipedia.org/wiki/Bethe_formula#The_formula)

According to NIST, they use the Bethe formula (+ higher-order corrections for high energies) for energies above some energy, which they specify to be around 0.5 MeV for protons. Maybe a good test for the new Bethe function would be to compare these results to the data from NIST?

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -674,6 +667,7 @@ def create_particles(
         self.particle_energy = particle_energy.to(u.eV).value
```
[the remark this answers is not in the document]
JaydenR2305: There's a little bit of subtlety here because `_angles_monte_carlo` and `_angles_uniform` both read (and write in `_angles_uniform`) `self.nparticles`, which should in theory only be set by `load_particles`. Should I refactor these methods to take in an `nparticles` argument?

tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
```diff
@@ -985,3 +977,54 @@ def test_radiography_memory_save_routine() -> None:
     sim = cpr.Tracker(grid, source, detector, field_weighting="nearest neighbor")
     sim.create_particles(1e3, 15 * u.MeV, max_theta=8 * u.deg, random_seed=42)
     sim.run()
+
+
+@pytest.mark.slow()
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
JaydenR2305: As it is currently defined the `create_particles` method doesn't allow for multiple calls within the same simulator object. Should this functionality, in addition to being able to call the underlying `load_particles` multiple times, be implemented?

JaydenR2305:
Out of curiosity, should two changelog files be created when a PR includes new features as well as breaking changes (2712.breaking.rst and 2712.feature.rst) ? I'm looking at the draft for the Towncrier changelog and it would make more sense to have some of the changes appear under the "Backwards Incompatible Changes" category.

JaydenR2305:
I ran some tests with using the new Bethe stopping routine. It seems like it works but for lower energies the stopping power is very much underestimated by the theory. This means values for the stopping range are very inaccurate. Is this fine? just a baked-in assumption when using a theory intended for high energy particles?

pheuer:
> Out of curiosity, should two changelog files be created when a PR includes breaking changes as well as new features (2712.breaking.rst and 2712.feature.rst)? I'm looking at the draft for the Towncrier changelog and it would make more sense to have some of the changes appear under the "Backwards Incompatible Changes" category.

Yep! You can make one changelog file for each type. Is "breaking" not the same as "backwards incompatible"? 

> I ran some tests with using the new Bethe stopping routine. It seems like it works but for lower energies the stopping power is very much underestimated by the theory. This means values for the stopping range are very inaccurate. Is this fine? just a baked-in assumption when using a theory intended for high energy particles?

Hmm, yes I think that's fine: can you add a warning that states this? I think it should display every time you use the Bethe stopping, because even if you start with a high energy particle it could be ranged down to low energy. Alternatively we could check at the end and, if any particle drops below some threshold energy (including  being stopped), then display the warning then? Or do both?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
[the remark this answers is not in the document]
pheuer: Not really a stopping routine though, just a model for the stopping power? In principle I agree this would be nice, but in practice I think this adds complexity and isn't maybe necessary yet - if people start using this a lot and that becomes a problem, we could split it off like that then?

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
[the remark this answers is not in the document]
pheuer: Can you make that variable local and only set `self.nparticles` in `load_particles`? 

Or yeah, adding an nparticles argument might be the better way of doing it.

JaydenR2305:
https://github.com/PlasmaPy/PlasmaPy/actions/runs/9763687160/job/26950157611#step:7:28

The joys of sphinx. Does anyone know how to fix this? mypy complains if I don't include the type hinting. I guess I could just revert it to `np.ndarray` and then add it to the exceptions file

pheuer:
> https://github.com/PlasmaPy/PlasmaPy/actions/runs/9763687160/job/26950157611#step:7:28
> 
> The joys of sphinx. Does anyone know how to fix this? mypy complains if I don't include the type hinting. I guess I could just revert it to `np.ndarray` and then add it to the mypy exceptions file

Yeah - is there a reason to force it to be float64 input? Is a float32 ndarray invalid input for some reason?

JaydenR2305:
> Yeah - is there a reason to force it to be float64 input? Is a float32 ndarray invalid input for some reason?
It's supposed to be a float32, yeah. I don't think any of the numpy size aliases work well with sphinx. something about intersphinx interpreting aliases as `py:class` as opposed to `py:attr`

tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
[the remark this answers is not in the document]
JaydenR2305: Just pushed this! It was somewhat difficult but I was able to brush up on my advanced indexing in NumPy. The performance benefits are definitely worth it

JaydenR2305:
> Hmm, yes I think that's fine: can you add a warning that states this? I think it should display every time you use the Bethe stopping, because even if you start with a high energy particle it could be ranged down to low energy. Alternatively we could check at the end and, if any particle drops below some threshold energy (including being stopped), then display the warning then? Or do both?

Ok I've added a warning for <1 MeV since that's the same threshold NIST uses

pheuer:
> > Yeah - is there a reason to force it to be float64 input? Is a float32 ndarray invalid input for some reason?
> 
> It's supposed to be a float32, yeah. I don't think any of the numpy size aliases work well with sphinx. something about intersphinx interpreting aliases as `py:class` as opposed to `py:attr`

I still don't quite understand this: why would the function not work if you passed in the input as, say, np.float128?

JaydenR2305:
> I still don't quite understand this: why would the function not work if you passed in the input as, say, np.float128?

That's a good point. I just went straight to `float32` because I knew that's the size python allocates under the hood for floats. A more accurate type would probably just be `np.generic`, which is the base class for all of NumPy's scalar types. This (hopefully) simplifies the intersphinx side of things as well because `np.generic` is defined as a `py:class`, which is what autodoc is looking for.

JaydenR2305:
Ok so it looks like we're going with `np.integer[Any] | np.floating[Any]` over `np.number` so we don't have to deal with complex numbers

JaydenR2305:
image
Here's a fun inheritance chart showing the relationships between the different kind of scalars defined in numpy

pheuer:
> > I still don't quite understand this: why would the function not work if you passed in the input as, say, np.float128?
> 
> That's a good point. I just went straight to `float32` because I knew that's the size python allocates under the hood for floats. A more accurate type would probably just be `np.generic`, which is the base class for all of NumPy's scalar types. This (hopefully) simplifies the intersphinx side of things as well because `np.generic` is defined as a `py:class`, which is what autodoc is looking for.

I think thats better - users will probably provide input as numpy types, and I think wherever possible/reasonable we want the function to try to execute using the precision provided. Certainly downgrading the precision of the input without a clear reason wouldn't be good, because the user may have a good reason for wanting that precision.

JaydenR2305:
image

Is this implemented in the formulary somewhere already? If not should we just include atomic number, weight, etc. as keyword arguments to the Bethe stopping function? should this logic be transferred into the lite function? I'm unsure whether or not including a few if statements would have an adverse impact on performance-- but we can always do profiling

JaydenR2305:
Why isn't codecov picking up on the coverage here? https://app.codecov.io/gh/PlasmaPy/PlasmaPy/pull/2712/blob/src/plasmapy/formulary/collisions/misc.py#L325

The tests I've written thus far have used the `Bethe_stopping` function which just calls `Bethe_stopping_lite` so both should be covered?

pheuer:
> Why isn't codecov picking up on the coverage here? https://app.codecov.io/gh/PlasmaPy/PlasmaPy/pull/2712/blob/src/plasmapy/formulary/collisions/misc.py#L325
> 
> The tests I've written thus far have used the `Bethe_stopping` function which just calls `Bethe_stopping_lite` so both should be covered?
> 
> Edit: it appears that merging main did the trick?

No idea why this worked

JaydenR2305:
@pheuer this PR is ready for review

src/plasmapy/simulation/particle_integrators.py
```diff
@@ -235,13 +235,16 @@ def push(x, v, B, E, q, m, dt):
...
         uvel = v * γ
 
         uvel_minus = uvel + q * E * dt / (2 * m)
 
-        γ1 = np.sqrt(1 + (np.linalg.norm(uvel_minus) / _c.si.value) ** 2)
+        γ1 = np.sqrt(
+            1 + (np.linalg.norm(uvel_minus, axis=1, keepdims=True) / _c.si.value) ** 2
+        )
 
         # Birdsall has a factor of c incorrect in the definition of t?
```
pheuer, line 249: I forget, did we determine Birdsall actually had this factor of c wrong? Should this comment still be here?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -164,13 +182,19 @@ def __init__(
...
+
         # Raise a ValueError if a synchronized dt is required by termination condition or save routine but one is
         # not given. This is only the case if an array with differing entries is specified for dt
         if self._require_synchronized_time and not self._is_synchronized_time_step:
             raise ValueError(
                 "Please specify a synchronized time step to use the simulation with this configuration!"
             )
 
+        # Ensure that the grids have defined the necessary quantities or at least zeroed
+        self._required_quantities = self._REQUIRED_QUANTITIES.copy()
```
pheuer, line 197: I don't understand what's happening here? is _REQUIRED_QUANTITIES a Tracker property?

EDIT: Oh, I see, it's just defined below! Can you move it up, maybe above init, so one finds it before getting to this point when reading the code?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -429,12 +468,143 @@ def load_particles(
...
+        self,
+        method: Literal["NIST", "Bethe"],
+        materials: list[str] | None = None,
+        I: u.Quantity[u.eV] | None = None,  # noqa: E741
+    ):
+        r"""
+        Enable particle stopping using experimental stopping powers.
+
+        Interpolation of stopping powers is conducted using data from the NIST
+        PSTAR. This information is combined with the mass density quantity
```
pheuer, line 548:
```suggestion
        PSTAR database. This information is combined with the mass density quantity
```

pheuer (commented):
Sorry for taking a little while to get back to this one! Almost there, here's a few minor comments.

pheuer:
> @pheuer this PR is ready for review

Thanks for the reminder! Definitely ping me like this if I leave something on read for a while!

src/plasmapy/simulation/particle_integrators.py
JaydenR2305: Good catch! iirc Birdsall was using CGS units

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -309,24 +329,39 @@ def _validate_constructor_inputs(
...
+        This method is not only called during instantiation of the |ParticleTracker|
+        but is also called when enabling stopping.
         """
 
-        # Some quantities are necessary for the particle tracker to function regardless of other configurations
-        required_quantities = {"E_x", "E_y", "E_z", "B_x", "B_y", "B_z"}
+        # Convert the required quantities set to a list to appease type hinting
+        quantities_list = list(self._REQUIRED_QUANTITIES)
 
         for grid in self.grids:
```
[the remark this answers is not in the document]
JaydenR2305: So we should require that each quantity should be provided on at least one grid, otherwise an error will be emitted? What about simple cases where we only want to specify an E field, like in our Gaussian potential notebook?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -429,12 +468,143 @@ def load_particles(
...
+                    raise ValueError(
+                        "Please specify the relevant materials to use NIST stopping!"
+                    )
+
+                if len(materials) != len(self.grids):
+                    raise ValueError(
+                        "Please provide an array of length ngrids for the materials."
+                    )
+
+                if not self._is_quantity_defined_on_one_grid("rho"):
```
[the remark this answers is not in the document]
JaydenR2305: Sorry which if statement? I see that they are essentially the same checks but with different quantity names. Did you want me to abstract this, with some `req_grid_quantity` (rho, n_e), `req_grid_constant` (materials, mean excitation energy)

pheuer:
Looks good to merge once the CI is passing again!

pheuer (approved):
Awesome, let there be stopping!

PR #2713, opened 2024-06-03
Add `enum` Module

JaydenR2305:
Implements a new `enum` module to store enumerations for data types. Adds a `StoppingMaterial` enum to be used with the NIST stopping power database

should probably wait until after #2712 is merged


JaydenR2305:
Will wait on #2710 to be merged so I can use the new decorator syntax to write the test (making sure all members of the enum are actually valid material strings)

pheuer:
I definitely see why you'd do it this way, but I'm a little averse to creating a new module + enum just to support this one niche feature of the particle tracker. What if we just accepted a string for now, and said in the docstring that it had to be one of the entries in the h5 file? Or listed the acceptable strings at the end of the docstring? 

Or, you could put this enum in the proton tracker file? 

Just trying to manage complexity of the package.

JaydenR2305:
> Or, you could put this enum in the proton tracker file?

Ah this makes a lot more sense. I'll do that instead

pheuer:
> > Or, you could put this enum in the proton tracker file?
> 
> Ah this makes a lot more sense. I'll do that instead

Actually, logically this should really be the same file as the `stopping_power` function, so that they appear in the docs together? The problem is, that's already in quite a large file `particles/atomic`...

What if you created `_stopping_power.py` in `particles` and moved both the `stopping_power` function and this associated enum there? @namurphy what do you think about the best way to organize this?

JaydenR2305:
> What if you created `_stopping_power.py` in `particles` and moved both the `stopping_power` function and this associated enum there? @namurphy what do you think about the best way to organize this?

Yeah that's probably a good idea especially given that I'm splitting the function into multiple routines in #2712. Should this be a public module? Or are you saying the `enum` should only be accessible by the particles subpackage (`plasmapy.particles.StoppingMaterial`)

pheuer:
@JaydenR2305 What's happening with this one? I think this enum should be in the same namespace as the function that gets the NIST data

JaydenR2305:
> @JaydenR2305 What's happening with this one? I think this enum should be in the same namespace as the function that gets the NIST data

Speaking of which, where do you think these functions should be moved?? I don't think the atomic module is the best/most intuitive place

JaydenR2305:
After discussion with @pheuer we are opting for a string input over an enumeration of all possible materials. A link to the PSTAR/ASTAR databases will be included in the relevant documentation to direct users to a list of acceptable material strings

PR #2715, opened 2024-06-03
Prevent array shape mismatch due to NumPy squeeze in `mhd_waves_.py`

TrestanSimon:
## Description

This PR adds an if statement to prevent array shape mismatch due to `np.squeeze` in `mhd_waves_.py`.

`np.squeeze` will convert column vectors to row vectors, for example, which causes problems where the array shape is assumed to be preserved.


namurphy (approved):
Thank you for doing this, and apologies for the delay in merging it!

PR #2716, opened 2024-06-03
Update GitHub workflow to regenerate requirements

namurphy:
This PR updates the GitHub workflow and PR body that regenerates requirements files.  

 - I moved the body of the PR into its own file.  
 - I updated the content of the PR body since it was a bit out-of-date.
 - I made this workflow fortnightly for the cron job, while also letting maintainers trigger this manually by workflow dispatch. 

⚖️ Running this frequently means that we'll find and fix breaking changes sooner, and thus be able to submit bug reports to dependencies more quickly; however, we'd have less stability and would need to expend more effort doing code reviews.  Running this less frequently means it'd take longer to find/fix/report new bugs, but less time spent doing code reviews.  Fortnightly seems like a good balance.

💭 I was wondering if there was a way to set this to run on branches and forks, but the token step makes things complicated.  (I ended up testing this on my fork, but I had to comment out all the parts related to generating a token since they are specific to PlasmaPy/PlasmaPy.)


namurphy:
And...merging this because I need to run it to be able to verify that it is working.

PR #2731, opened 2024-06-27
Add GitHub workflow to check changelog entries

namurphy:
This PR adds enables the [scientific-python/action-towncrier-changelog](https://github.com/scientific-python/action-towncrier-changelog) GitHub workflow which is used to validate changelog files.

This would possibly replace Giles.


namurphy:
This needs to be merged before we can fully test it, so merging it now.

PR #2733, opened 2024-06-28
Add type hints to `plasmapy.utils.roman`

namurphy:
This PR adds type hint annotations to `plasmapy.utils.roman` and its tests.

I'm also using it as a way to test the GitHub workflow used to check for changelog entries (#2731), and to distract me from any U.S. presidential debates that are about to happen.


noxfile.py
```diff
@@ -322,6 +322,7 @@ def mypy(session: nox.Session) -> None:
         version=session.python,
         resolution="highest",
     )
+    session.install("pip")
```
namurphy, line 325: For some reason I was getting an error that `pip` was not installed, at least when running `nox -s mypy` locally. Maybe because I was using a `uv` backend?

PR #2736, opened 2024-06-28
Update docstrings in `plasmapy.formulary.collisions`

namurphy:
This PR makes some minor updates to docstrings in `plasmapy.formulary.collisions`.

Changes include:
 - Using unicode characters instead of LaTeX symbols (which makes docstrings more readable when using `help()`)
 - Removing extra brackets in equations (which makes it easier to keep track of brackets)
 - Justifying docstring text


pheuer (approved):
LGTM! Is unicode paying you? :P

PR #2743, opened 2024-07-03
Drop `|tox|` global substitution and adjust changelog formatting

namurphy:
We recently completed our switch from tox → Nox as our test runner. This is a quick follow-up PR to drop usage of the `|tox|` global substitution (mostly in the changelog) since we no longer need to refer to it, and to drop `tox` and `tox-uv` from the requirements. While I was at it, I also did some minor justifying of text in changelog files (since it sometimes gets formatted awkwardly).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,12 +113,11 @@ Internal Changes and Refactorings
 - Removed :file:`setup.py`. (:pr:`2558`)
 - Added ``sphinx-lint`` as a |pre-commit| hook to find
   reStructuredText errors. (:pr:`2561`)
-- Enabled the `tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin to tox,
-  so that package installation, caching, and the creation of virtual
-  environments will
-  be handled by |uv| instead of |pip|. This change makes it faster to run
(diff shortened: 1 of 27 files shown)


CHANGELOG.rst
```diff
@@ -105,15 +105,15 @@ Internal Changes and Refactorings
...
   caching speeds up several continuous integration tests by ∼2–3 minutes.
   See :issue:`2585` to learn more about recent efforts to drastically
   speed up PlasmaPy's continuous integraiton checks. (:pr:`2552`)
 - Removed :file:`setup.py`. (:pr:`2558`)
 - Added ``sphinx-lint`` as a |pre-commit| hook to find
   reStructuredText errors. (:pr:`2561`)
-- Enabled the `tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin to |tox|,
+- Enabled the `tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin to tox,
   so that package installation, caching, and the creation of virtual
   environments will
```
namurphy, line 118:
```suggestion
- Enabled the ``tox-uv``plugin to tox, so that package installation,
  caching, and the creation of virtual environments will
```

Issue #2647, opened 2024-04-25
Requirements files need to be rethought


rocco8773:
There are numerous issues with the current structure of requirements files...

1. There is no documentation on when or how to regenerate files
2. There is no documentation on the intended purpose of each requirements file
3. There are no breadcrumbs telling the developer to regenerate the requirements files after the `pyproject.toml` is modified.
4. Requirements files are purpose built for CIs (not an issue) but at the expense of the developer use (an issue)
5. When `uv` generates a requirements files it by default adds OS specific packages to that file.  So, if you generate the file on a Windows machine without the `--no-deps` flag, then all associated CIs and RTD runners will fail.  Again, not documented anywhere.
6. IDEs often use the `requirements.txt` for inspection, so a frozen set of dependencies is a nuisance for the developer.  Since the main `requirements.txt` is frozen set of dependencies for a given py version, it is not a good reference file for IDEs when the developer is using a different py version.  It would be better to have a flexible dependency file for developers.
7. I don't think a frozen set of requirements is ideal for the RTD builds, since that would require us to do a dedicated PR on the `plasmapy` side every time we update `plasmpay_sphinx`.

Possible Solutions:

1. Restrict the `requirements/` directory to CI focused requirements only.
2. Add breadcrumbs to `pyproject.toml` to tell the developer to update the CI specific requirement.
3. Have the GitHub actions self generate the requirements files so we offload the burden from the contributor.
4. Add to the documentation instructions for buildings requirements for CIs and personal developer use.
5. Reserve the `requirements.txt` for the developer and add it to `.gitignore` so the developer can create a `requirements.txt` for their own development environment.  (This would be done in combination with the above item.)
6. Instead of the above, have a manually created `requirements.txt` that mirrors `pyproject.toml`, like we used to have.  This is not ideal from a maintainer perspective, but offloads a lot of burden from the contributor side.
7. Update `.readthedocs.yml` so it install dependencies like `plasmapy[docs]` instead of a requirements file.  I'm doing this right now in PR #1643 .  (Currently this does not appear to affect RTD build time.)

namurphy:
> There is no documentation on when or how to regenerate files
> 
> There are no breadcrumbs telling the developer to regenerate the requirements files after the `pyproject.toml` is modified.

Very important points!  Addressing this in #2650. 🔧 

[...]

PR #2745, opened 2024-07-05
Automatically merge dependabot PRs

JaydenR2305:
As part of the discussion on the role of `requirements.txt` in #2647 @namurphy mentioned it would be convenient to implement a workflow that automatically merges dependabot pull requests should all of the necessary CI tests pass.

This is accomplished using the workflow outlined [here](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request). The default behavior of auto-merge is to hold off on merging if CIs fail, which is exactly the behavior we're looking for.

I'm not quite sure what the best way to test this is. Maybe I could make my own repository and link my results here

### DIFF
--- a/.github/workflows/merge-dependabot.yml
+++ b/.github/workflows/merge-dependabot.yml
@@ -1,0 +1,28 @@
+name: Dependabot auto-approve
+on: pull_request
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+    - name: Fetch metadata
+      id: metadata
+      uses: dependabot/fetch-metadata@v2
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Approve PR
+      run: gh pr review --approve "$PR_URL"
+      env:
+        PR_URL: ${{github.event.pull_request.html_url}}
+        GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+    - name: Enable auto-merge
+      run: gh pr merge --auto --merge "$PR_URL"
+      env:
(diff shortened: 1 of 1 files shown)


JaydenR2305:
Ok here are some tests I ran:
https://github.com/JaydenR2305/workflow-testing/pull/1
https://github.com/JaydenR2305/workflow-testing/pull/2

.github/workflows/merge-dependabot.yml
```diff
@@ -0,0 +1,29 @@
...
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
```
namurphy, line 11:
```suggestion
    if: github.actor == 'plasmapy-requirements-bot'
```

🔧 I forgot to mention in #2647 that the automated requirements updates are done through our own custom `plasmapy-requirements-bot` rather than dependabot.  I'm not sure if we'd need to include the `[bot]` at the end here.

namurphy (commented):
Thank you for doing this!  GitHub Actions are a pain to test, but next week after you're ready, we can merge this and manually trigger the "Update pinned requirements" workflow to make sure that it works.

.github/workflows/merge-dependabot.yml
JaydenR2305: Ah okay. Both dependabot and the plasmapy-requrirements-bot have the "app/" prefix when I add them as a filter to the PR search under "author", so I'm going to guess we need to add the [bot] suffix for the plasmapy bot also.

namurphy:
It's taking me longer to get to this than I hoped, but it's still on my radar.  Thank you again for doing this!

PR #2746, opened 2024-07-05
Require `particle` for `ParticleTracker.load_particles`

### LINKED ISSUE #2725: Make  `particle` a required keyword to `ParticleTracker.load_particles`?
### Feature description

Right now this method assumes that the particles being added are protons, but I wonder if that's really a good default? Maybe that keyword should be required, just to be explicit? 

```python
@particle_input
    def load_particles(
        self,
        x,
        v,
        particle: Particle = Particle("p+"), 

```

### Motivation

Avoid anyone accidentally creating particles with the wrong mass/charge 

### Implementation strategy

_No response_

### Additional context

_No response_

JaydenR2305:
Implements `particle` as a required argument for the load particle method of the particle tracker.

Resolves #2725

### DIFF
--- a/changelog/2746.breaking.rst
+++ b/changelog/2746.breaking.rst
@@ -1,0 +1,1 @@
+Convert ``particle`` to a required argument of the ``~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker.load_particles`` method.
--- a/src/plasmapy/simulation/particle_tracker/particle_tracker.py
+++ b/src/plasmapy/simulation/particle_tracker/particle_tracker.py
@@ -394,7 +394,7 @@ class ParticleTracker:
         self,
         x,
         v,
-        particle: Particle = Particle("p+"),  # noqa: B008
+        particle: Particle,
     ) -> None:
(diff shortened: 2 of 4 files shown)


JaydenR2305:
pre-commit.ci autofix

pheuer (commented):
Looks good! One minor request: can you make sure it takes string inputs like "p+" directly as well? I think applying the `particle_input` decorator to the method would automatically do this for any parameter named `particle` (or maybe it uses the type hint?) 

Otherwise looks ready to merge!

JaydenR2305:
> Looks good! One minor request: can you make sure it takes string inputs like "p+" directly as well? I think applying the `particle_input` decorator to the method would automatically do this for any parameter named `particle` (or maybe it uses the type hint?)

Yep, providing a string input works. The `particle_input` decorator inspects the signature for `ParticleLike` annotations, which in turn takes care of the necessary conversions. Did you want this to be implemented as a test? or were you just confirming

pheuer (approved):
LGTM, thanks!

PR #2747, opened 2024-07-05
Add experimental Nox session for running MonkeyType

namurphy:
This PR adds an experimental Nox session that uses [MonkeyType](https://monkeytype.readthedocs.io) to add type hint annotations to different modules.  Essentially, this session uses MonkeyType to build a database of types based on what shows up when running pytest, and then applies those types to a particular module.  For example, if I wanted to add type hints to `plasmapy.particles.atomic`, we would run 

```shell
nox -s monkeytype -- plasmapy.particles.atomic
```

The type hint annotations that get added will still require a careful code review, but this would at least provide a start.  I'm not sure if we'd want to consider this a temporary addition or permanent, since it could potentially be useful for adding type hint annotations to newly created modules.


noxfile.py
```diff
@@ -386,6 +387,43 @@ def autotyping(session: nox.Session, options: tuple[str, ...]) -> None:
...
+
+    session.install(".[tests]")
+    session.install("MonkeyType", "pytest-monkeytype", "pre-commit")
+
+    database = pathlib.Path("./monkeytype.sqlite3")
+
+    if not database.exists():
+        session.log(f"File {database.absolute()} not found. Running MonkeyType.")
+        session.run(
+            "pytest", "-m", "not slow", f"--monkeytype-output={database.absolute()}"
```
namurphy, line 415:
```suggestion
            "pytest", f"--monkeytype-output={database.absolute()}"
```
...for after this gets past the initial testing phase.

docs/_global_substitutions.py
```diff
@@ -192,7 +192,7 @@ def make_global_substitutions_table(
             f"``{global_substitutions[substitution].rstrip()}``",
             f"|{substitution}|",
         )
-        for substitution in sorted(global_substitutions)
+        for substitution in sorted(global_substitutions, key=lambda x: x.lower())
```
namurphy, line 195: Follow-up of #2766.

PR #2748, opened 2024-07-08
Add Scattering to `ParticleTracker`

JaydenR2305:
Adds Coulomb scattering to the `ParticleTracker`

Merge #2712 and #2790 first


JaydenR2305:
Current plan:
- Allow users to specify their own defined differential cross section for scattering
- Implement the Mott-Born formula and the exact Mott phase-shift formula as outlined in [this article](https://journals.aps.org/rmp/pdf/10.1103/RevModPhys.36.881)
- Specified cross sections will be numerically integrated to obtain mean-square rate of scattering:
image
where theta_min is-- in the above reference-- given to be twice the ratio of the impact parameter to the Debye length

pheuer:
Sounds like a good plan! Post some benchmarking of both models vs the ENDF data here for comparison when you get the models working!

JaydenR2305:
image
Here is what I have so far. This is for 5 MeV protons travelling through 2 cm of room temperature aluminum along with a Gaussian potential E field.

Are there any experimental data on differential scattering cross-sections for plamas that we can compare to?

pheuer:
> Here is what I have so far. This is for 5 MeV protons travelling through 2 cm of room temperature aluminum along with a Gaussian potential E field.

Good start! Presumably with stopping turned off, since that much Al would definitely stop the protons! I would do it without the E-field, since we can't simulate that in SRIM. 

> Are there any experimental data on differential scattering cross-sections for plamas that we can compare to?

Probably, but I'm not well versed on that literature - maybe you could poke around for some papers? If not I think we can benchmark in solids vs. SRIM to start - that's probably enough for the paper, although plasma scattering would be a great addition.

JaydenR2305:
image
has the same form as
image
with sigma = sqrt(<theta_0^2>)

JaydenR2305:
Apparently I've been misunderstanding SRIM's definition of lateral straggling, which is given as:
image
given that the ion beam is colinear with the x-axis

(see page 26 here for the definition of other statistical quantities relevant to scattering http://www.srim.org/SRIM/SRIM%2008.pdf) 
In general, the straggling is defined to be the second moment of the (cylindrically symmetric) distribution of ions about the center of the detector 🤷‍♂️

pheuer:
> With these adjustments I'm getting a value of 7.9 microns for the lateral straggling, which is at least _closer_ to the 8.65 micron expected value. One potential reason for this deviation is because I'm not currently removing back-scattered particles, which may represent a portion of the ion population.

Awesome, this looks great! Try excluding back-scattered particles to see if that helps, and test this over a number of different materials/densities. For the paper, I'm envisioning maybe a plot of lateral straggling vs. areal density of the target, and then plotting values for both SRIM and the tracker for 3-4 materials scanning across the areal density range? That should be a good validation. 

Also would be interesting to plot the analytical model there too, just to show where it diverges.

pheuer:
I'll start giving this code a rough review soon since it seems to be working - can you merge in main so the changes from the previous PR don't show up in the diff?

JaydenR2305:
> I'll start giving this code a rough review soon since it seems to be working - can you merge in main so the changes from the previous PR don't show up in the diff?

Should be good now!

changelog/2712.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added the ability to enable particle stopping in the |ParticleTracker|.
+Added a ``return_interpolator`` keyword to `~plasmapy.particles.atomic.stopping_power` to allow the user to specify the return of an interpolator function (`~scipy.interpolate.CubicSpline` under the hood).
```
pheuer, line 2: This line of the changelog was added in the last PR, right?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -156,6 +159,7 @@ def __init__(
         dt_range=None,
         field_weighting="volume averaged",
         req_quantities=None,
+        seed: int | None = None,
```
pheuer, line 162: good idea! Add this to the docstring

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -447,6 +456,8 @@ def load_particles(
         # Raise an error if the run method has already been called.
         self._enforce_order()
 
+        # TODO: it is an assumption that there exists only one specie of particles
```
pheuer, line 459: Yes
```suggestion
```

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -447,6 +456,8 @@ def load_particles(
...
+        #  is this a reasonable assumption?
```
pheuer, line 460:
```suggestion
```

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -584,6 +596,147 @@ def inner_Bethe_stopping(v, n_e):
         self._stopping_method = method
         self._stopping_power_interpolators = stopping_power_interpolators
 
+    # TODO: create a method to validate that these quantities have shape ngrids
+    # TODO: is this signature too much?
+    # TODO: find a way to create a type for the signature of the Callable parameters
```
pheuer, line 601: Yes, this is the way to clean this up, not sure how it works though.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -584,6 +596,147 @@ def inner_Bethe_stopping(v, n_e):
...
+                    u.Quantity[1 / u.m**3],  # Electron Number Density
+                ],
+                u.Quantity[
+                    u.rad**2 / u.s
+                ],  # Mean-square scatter rate in radians^2 per second
+            ]
+            | None
+        ) = None,
+    ):
+        r"""
```
pheuer, line 638: Expand docstring - lots of parameter details, physics stuff to explain here

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -584,6 +596,147 @@ def inner_Bethe_stopping(v, n_e):
...
+            self._scattering_method = "differential cross section"
+            self._scattering_routine = differential_cross_section
+        elif mean_square_scatter_rate is not None:
+            self._scattering_method = "mean square scatter rate"
+            self._scattering_routine = mean_square_scatter_rate
+
+        self._target = target
+        self._do_scattering = True
+
+        # For now the integrals will be evaluated every time step
```
pheuer, line 680: Is there an alternative to this, since the local density and temperature at each particle is different on each timestep?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -675,17 +828,58 @@ def fract_entered(self):
...
+        # Iterate through the `self._total_grid_values` dictionary, keeping only
+        # the necessary values
+        self._total_grid_values = {
+            key: np.compress(
+                keep_grid_value_mask[self._tracked_particle_mask], grid_value
+            )
+            for key, grid_value in self._total_grid_values.items()
+        }
+
+        # TODO: this seems awkward, maybe ``dt`` should be of shape ``nparticles``? as
```
pheuer, line 861: Good idea - can you have the dt method cast it that way?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -584,6 +596,147 @@ def inner_Bethe_stopping(v, n_e):
...
+    # TODO: create a method to validate that these quantities have shape ngrids
+    # TODO: is this signature too much?
+    # TODO: find a way to create a type for the signature of the Callable parameters
+    #  (preferably using only one expression)
+    @particle_input
+    @validate_quantities
+    def add_scattering(
+        self,
+        target: ParticleLike,
+        differential_cross_section: (
```
pheuer, line 608: I don't entirely understand where these callables come from, but here's an idea. 
I suggest: 

1) Make all of these parameters optional keywords to this function along with a 'method' keyword that chooses between the two. Make clear in the docstring which keywords are for which method, and which are for both. 

2) For the given method, ensure that the associated keywords are set, match the grid size, etc. 

3) Create the callable inside this function?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -982,21 +1186,116 @@ def _update_velocity_stopping(self, summed_field_values):
...
+                            theta_prime,
+                            speed,
+                            self._particle,
+                            self._target,
+                            T_e,
+                            n_e,
+                        )
+                    ).si.value
+
+                # Calculate the minimum scattering angle based on screening
```
pheuer, line 1209: Write out all of what happens in both methods and how that is applied to the particles in latex math in the docstring for adding scattering. Bonus, you can then put that in the paper!

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -982,21 +1186,116 @@ def _update_velocity_stopping(self, summed_field_values):
...
+            self._rng.uniform(low=-1, high=1, size=(self.nparticles_tracked, 3)),
+        )
+        # Normalize
+        k_norm = k / np.linalg.norm(k, axis=-1, keepdims=True)
+        # Scale to the perpendicular component of the velocity
+        delta_v = speeds[:, np.newaxis] * np.sin(theta) * k_norm
+
+        # Begin by constructing the new velocities by adding the change in
+        # velocity to the current velocities. The direction of this vector will be
+        # correct but will need to be "rescaled" to unsure that the speed remains
```
pheuer, line 1274:
```suggestion
        # correct but will need to be "rescaled" to ensure that the speed remains
```

pheuer (commented):
Looks pretty good so far! Figuring out how to create the callables in the user-accessible function that defines everything in a well documented easy to apply way is the most complex part I think. 

Make sure to describe the math fully in the docstring - as you've seen scattering is complex, so we need to be detailed about exactly what the code is doing.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: Somewhere in the code I had a block commented out where I was essentially trying to construct a multidimensional interpolator that takes in v, T_s, and n_s and returns the mean-square rate of scattering. To do this I had to come up ranges for the parameters such that every possible combination of v, T_s, and n_s would be within the "domain" of the interpolator. The ranges (sampled in logspace) are v: 0 -> c , T_s: 0 -> max(T_s), and n_s: 0 -> max(n_s) where we cheat by taking the max value of temperature and number density encountered in the relevant grid. Does that sound like a reasonable approach?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: Sounds good. Did you have any recommended resources for scattering? I'm reading through Taylor's Classical Mechanics at the moment

JaydenR2305:
> Awesome, this looks great! Try excluding back-scattered particles to see if that helps

Just to make sure I'm understanding this correctly, back-scattered particles are where the angle between the velocity and src->detector vector is greater than 90 degrees?

pheuer:
> > Awesome, this looks great! Try excluding back-scattered particles to see if that helps
> 
> Just to make sure I'm understanding this correctly, back-scattered particles are where the angle between the velocity and src-to-detector vector is greater than 90 degrees?

Yes exactly

JaydenR2305:
Just realizing that I forgot to move my stopping range test into the proper file before merging our other PR.. I'll add that alongside our lateral straggling tests here for scattering

src/plasmapy/simulation/particle_tracker/particle_tracker.py
pheuer: I'll send you some stuff! I mostly meant just the equations showing how the mean scattering angle is obtained from the dsigma/domega, and provide references for the ENDF data vs. the analytic model (provide an equation for the analytic model). Basically just nicely write out what's in the code in more readable latex and explain where it comes from.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
pheuer: I suppose the question is one of computation time, and consequently on how fine of a resolution for that grid is necessary for the initial computation, before the interpolation? If constructing the interpolator with sufficient resolution is fast enough, then this could work well.

JaydenR2305:
image
thank you IUPAC

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: Just finished this, let me know what you think of how everything is documented :)

pheuer: Two concerns still: 

1) The callables don't know what scattering type they are (?), so there's the potential for conflict. I still think it would be better, for the limited number of modes we have, to just have the callable selected internally based on a keyword that selects which mode to use? 

2) You should add all the other keywords as parameters, e.g. `T_s`, `n_s` to explain in detail what each one is. The docstring for the parameter can say `Only used for routines ....` for each parameter so it's clear which are required when. But we need to clearly define what they all are, and that's easier when they are all  listed as separate parameters in the normal format.

JaydenR2305:
@pheuer What are your thoughts on a `cast_iterable` keyword inside of the `validate_quantities` decorator? I'm running into the problem of implicitly assuming that the user isn't providing a scalar argument for things like `target_rho` inside of the `add_stopping` and `add_scattering` methods. The `cast_iterable` would convert any passed scalars into a (1,) NumPy array.

Also, should `validate_quantities` cast a single `ParticleLike` quantity into a `ParticleList` for arguments annotated with `ParticleList`? This is basically a similar case as above except with particle objects.

pheuer:
> @pheuer What are your thoughts on a `cast_iterable` keyword inside of the associated dictionary defined in the `validate_quantities` decorator? I'm running into the problem of implicitly assuming that the user isn't providing a scalar argument for things like `target_rho` inside of the `add_stopping` and `add_scattering` methods. The `cast_iterable` would convert any passed scalars into a (1,) NumPy array.

@namurphy 
I would say that would be scope creep for that decorator, since it's intended to validate u. Quantity objects. I wouldn't be opposed to creating a `cast_iteratable` decorator separately, although I might call it `cast_list` or `cast_tuple` to be explict about what the type is (or I guess there could be a keyword to `cast_iterable` to set that). 

> Also, should `validate_quantities` cast a single `ParticleLike` quantity into a `ParticleList` for arguments annotated with `ParticleList`? This is basically a similar case as above except with particle objects.

I feel like this could maybe be a keyword option for the @particle_input decorator? Again pinging @namurphy for their input.

namurphy:
> @pheuer What are your thoughts on a `cast_iterable` keyword inside of the associated dictionary defined in the `validate_quantities` decorator? I'm running into the problem of implicitly assuming that the user isn't providing a scalar argument for things like `target_rho` inside of the `add_stopping` and `add_scattering` methods. The `cast_iterable` would convert any passed scalars into a (1,) NumPy array.



> Also, should `validate_quantities` cast a single `ParticleLike` quantity into a `ParticleList` for arguments annotated with `ParticleList`?

This would be under the purview of `@particle_input`




> > @pheuer What are your thoughts on a `cast_iterable` keyword inside of the associated dictionary defined in the `validate_quantities` decorator? I'm running into the problem of implicitly assuming that the user isn't providing a scalar argument for things like `target_rho` inside of the `add_stopping` and `add_scattering` methods. The `cast_iterable` would convert any passed scalars into a (1,) NumPy array.

> I wouldn't be opposed to creating a `cast_iteratable` decorator separately, although I might call it `cast_list` or `cast_tuple` to be explict about what the type is (or I guess there could be a keyword to `cast_iterable` to set that).

`validate_quantities` is currently quite complex, and I've found it difficult to add or change functionality in it (or even to add type hints). My inclination would be to create a separate but focused decorator that would include the needed functionality. 🤔 
 
> Also, should `validate_quantities` cast a single `ParticleLike` quantity into a `ParticleList` for arguments annotated with `ParticleList`?.

Yep, this would fall under the purview of `@particle_input`. One possibility would be that if the parameter is annotated with `ParticleListLike` instead of `ParticleLike`, then the corresponding argument would be transformed into a `ParticleList` instead of a `Particle | CustomParticle`.  I did a major refactoring of `@particle_input` a few years ago, so I'm hoping that it wouldn't be too difficult to modify this functionality! 😅

pheuer:
@JaydenR2305 If you choose to add these decorators, add them in a separate PR so we can review + merge them more quickly (rather than including them in this MR).

JaydenR2305:
Another caveat is that an isotope needs to be specified for the target material in order to use the `mass_number` attribute-- so something like ``Particle("Al")`` wouldn't work but ``Particle("Al-27")`` would.

Maybe this could be a keyword argument to the ``particle_input`` decorator?

JaydenR2305:
> Intuitively I feel like we should change it so `require` applies to all elements in a ParticleList type input?
Yep this is already the case apparently. The `particle_input` decorator passes `particlewise=True` to the `is_category` method of `ParticleList` which takes care of the rest

JaydenR2305:
> `validate_quantities` is currently quite complex, and I've found it difficult to add or change functionality in it (or even to add type hints). My inclination would be to create a separate but focused decorator that would include the needed functionality. 🤔

The main problem with making another decorator is that the only parameters of interest (at least so far) are `Quantity` objects, meaning they wouldn't fall neatly into a category like `cast_list` or `cast_tuple`, since they are technically array values under the hood. Making `cast_iterator` a keyword argument of a decorator meant to deal with sanitizing `Quantity` objects made sense to me since we're technically just imposing another "sanitization" criteria be applied to the quantity before passing it to the decorated function.

JaydenR2305:
@pheuer Also, what would you think about extending the `Particle` class to include array-like behavior by default? Meaning you would be able to instantiate a particle using a sequence of strings, then expressions like `2 * particle.atomic_number` would be supported without any changes to the code

JaydenR2305:
also for the time being I've solved the array quantity problem by:
```py
if not isinstance(target_rho.value, np.ndarray):
    target_rho = np.asarray([target_rho.value]) * target_rho.unit
```

pheuer:
> @pheuer Also, what would you think about extending the `Particle` class to include array-like behavior by default? 

This is an interesting idea - I think if you did this it would replace the ParticleList? Maybe this is a better way to do it? Worth opening an issue and tagging Nick to discuss, but for this PR I think the workaround you described is a good temporary solution.

changelog/2712.feature.rst
pheuer, line 1: This changelog entry appears to be for the wrong PR?  Maybe needs to be updated for this PR?

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -267,7 +270,7 @@ def Bethe_stopping_lite(
 
     beta = v / _c.si.value
 
-    return np.asarray(
+    return -np.asarray(
```
pheuer: Is this sign change to match the docstring which says by convention dEdx is positive? 

Also, in the docstring for this as well as the main function, "stopping energy" -> "stopping power"

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +328,80 @@ def Bethe_stopping(
     return Bethe_stopping_lite(I.si.value, n.si.value, v.si.value, z) * u.J / u.m
 
 
+# TODO: figure out if there's a better name for this. And where it came from.
```
pheuer: Agree this is a bit troubling and should be resolved before we merge this PR. Can you search for an English reference for this model? The Chinese citation looks like just a textbook, so maybe an English graduate plasma textbook. Or, look through the references 5,7-11 for Eq. 1 in this paper to see if any of them take it further? 

I often hear this referred to as Moliere's multiple scattering theory, and that's the title of the Bethe paper that is cited in ref. 9: https://journals.aps.org/pr/abstract/10.1103/PhysRev.89.1256 

Another possible references: https://apps.dtic.mil/sti/citations/trecms/AD0856254 

https://pubs.aip.org/aip/rsi/article/90/12/123503/1059941/MPRAD-A-Monte-Carlo-and-ray-tracing-code-for-the

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +328,80 @@ def Bethe_stopping(
...
+
+    T_s : `~astropy.units.Quantity`
+        The temperature of the target species.
+
+    n_s : `~astropy.units.Quantity`
+        The number density of scatterers in the material.
+
+    Notes
+    -----
+    The mean-square rate of scattering of incident protons is given in Mott & Massey (1950) to be
```
pheuer: I couldn't find this when I went looking for it - can you include an equation number

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -415,9 +449,8 @@ def load_particles(
                 " of particles "
                 f"({x.shape[0]} and {v.shape[0]} respectively)."
             )
-        else:
-            self.num_particles: int = x.shape[0]
 
+        self.num_particles: int = x.shape[0]
```
pheuer: Occurs to me that m,q,x,v num_particles etc. are all public variables but won't have any documentation. Should we make them private with a property to access them so the definitions show up in the docs?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -519,7 +559,7 @@ def add_stopping(
                 self._raised_energy_warning = False
 
                 # These functions are used to represent that the mean excitation energy
-                # does not change over space for a given grid.
+                # and ion number density does not change over space for a given grid.
```
pheuer: I don't really understand this - in a solid (which this model applies to) the ion density and electron density are directly related since it is neutral? So by specifying n_e, n_i is given.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -459,8 +492,15 @@ def _validate_stopping_inputs(
...
 
+            case _:
+                raise ValueError(
+                    f"Please provide one of 'NIST' or 'Bethe' for the method keyword. (Got: {method})"
+                )
+
         return True
 
+    @particle_input(require="isotope")
+    @validate_quantities
```
pheuer, line 592: I think given the package name, we should maybe specify that this model applies for stopping (and the scattering model to?) in solids/liquids/gases, but NOT in plasmas. There are models for that like Li-Petrasso, but this model won't work well in that case. This is more meant for solid components around the plasma in an experiment.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -546,6 +586,161 @@ def inner_Bethe_stopping(v, n_e):
...
+            Relevant keyword arguments are passed to the provided function, where
+            the defined keywords depend on the value of ``scatter_routine_type``. If
+            ``scatter_routine_type`` is set to ``differential cross-section``
+            then the function receives the following keyword arguments: ``theta``,
+            ``speed``, ``beam_material``, ``target_material``, ``T_s``, and ``n_s``.
+            When ``scatter_routine_type`` is set to ``mean square rate``,
+            the passed function receives the following keyword arguments: ``speed``,
+            ``beam_material``, ``target_material``, ``T_s``, and ``n_s``.
+
+        scatter_routine_type : `str`
```
pheuer: Is this keyword necessary if you can tell the routine type apart based on the units?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -546,6 +586,161 @@ def inner_Bethe_stopping(v, n_e):
...
+        # TODO: create a method to validate that these quantities have shape ngrids
+        # TODO: is this signature too much?
+        # TODO: find a way to create a type for the signature of the Callable parameters
+        #  (preferably using only one expression)
+
+    @particle_input(require="isotope")
+    @validate_quantities
+    def add_scattering(
+        self,
+        target: ParticleListLike,
```
pheuer, line 720: define target, target_rho in the docstring. 

Isn't rho already defined on the grid?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -546,6 +586,161 @@ def inner_Bethe_stopping(v, n_e):
...
+                "for the scatter routine keyword."
+            )
+
+        self._scattering_method = scatter_routine_type
+        self._scattering_routine = scatter_routine
+
+        self._set_target_attributes(target, target_rho)
+        self._do_scattering = True
+
+        # For now the integrals will be evaluated every time step
```
pheuer: What are your thoughts now on this block? If it's not going to be used, we should take it out

src/plasmapy/simulation/particle_tracker/particle_tracker.py
pheuer: Following up on this - the docs on the actual 'Mott-Bethe' function are a good start here, I would link to that function in the docstring for the method that turns scattering on, so it's clear what's being used.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -1037,6 +1232,127 @@ def _update_velocity_stopping(self) -> None:
         # Stop particles, which resets the cache
         self._stop_particles(particles_to_be_stopped_mask)
 
+    def _update_velocity_scattering(self):
+        speeds = np.linalg.norm(self.v[self._tracked_particle_mask], axis=-1)
+
+        match self._scattering_method:
+            # Numerically integrate the scattering integral to construct the mean square scatter rate
+            case "differential cross section":
+                # `theta_prime` dummy variable introduced to prevent global shadow of `theta`
```
pheuer: Maybe not the best name choice since I would read that as a time derivative or something. How about _theta or theta2

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -1037,6 +1232,127 @@ def _update_velocity_stopping(self) -> None:
...
+                    * self._target.atomic_number
+                    * const.e.si**2
+                    / (4 * np.pi * const.eps0 * mu * speeds**2)
+                )
+                lambda_D = Debye_length(
+                    self._total_grid_values["T_i"], self._total_grid_values["n_i"]
+                )
+                theta_min = 2 * b_0 / lambda_D
+
+                # Numerically integrate the classical scattering integral
```
pheuer: This does look awfully slow - at some point you mentioned making an interpolator when scattering was added (since we known the full range of T_i, n_i  on the grid before the sim starts) and then just calling that? Sounds like a good idea to me.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -546,6 +586,161 @@ def inner_Bethe_stopping(v, n_e):
...
+        Enable particle scattering by numeric integration of the provided
+        differential cross-section.
+
+        Parameters
+        ----------
+        target : `~plasmapy.particles.particle_collections.ParticleListLike`
+            A `~plasmapy.particles.particle_collections.ParticleListLike` representation of the target material(s).
+            The atomic mass of each particle must be specified (i.e. ``Al-27``).
+
+        scatter_routine : `callable`
```
pheuer: Is there a reason to support both of these? It does make the code a lot more complicated and double the number of cases to test. Let's discuss.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -1069,6 +1385,12 @@ def _push(self) -> None:
         if self._do_stopping:
             self._update_velocity_stopping()
 
+        if self.num_particles_tracked == 0:
```
pheuer: When is this needed? Clarify in a comment

tests/simulation/test_particle_tracker.py
```diff
@@ -648,6 +653,171 @@ def test_particle_tracker_Bethe_warning(
         simulation.run()
 
 
+# def test_particle_tracker_scattering_warning(
```
pheuer, line 700: Can we remove this block of commented out tests? Or is this waiting to be put back in?

tests/simulation/test_particle_tracker.py
```diff
@@ -648,6 +653,171 @@ def test_particle_tracker_Bethe_warning(
...
+
+    v0 = np.reshape(v0, newshape=(nparticles, 3))
+    sim.load_particles(x0 * u.m, v0 * u.m / u.s, particle=beam_particle)
+
+    # Add stopping and scattering to the simulation
+    sim.add_stopping("NIST", target_material, target_density, NIST_material_string)
+    sim.add_scattering(
+        target_material,
+        target_density,
+        scatter_routine=Mott_Bethe_mean_squared_scattering,
```
pheuer: This only tests the MSR option - what about the differential cross section option?

pheuer (commented):
Looking good! Let's discuss tomorrow

src/plasmapy/formulary/collisions/misc.py
pheuer: There should be another negative sign somewhere where this is applied to the particle energy

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -105,9 +110,24 @@ class to the ``run`` method.
 
         The default is 'volume averaged'.
 
+    req_quantities : `list` of `str`, default : `None`
```
pheuer, line 116: Take this keyword out

pheuer: And append rho to this list when yo uadd scattering

src/plasmapy/simulation/particle_tracker/particle_tracker.py
pheuer: "cold matter"

src/plasmapy/simulation/particle_tracker/particle_tracker.py
pheuer: Need to re-think how target is specified: needs to work easily for something like a C4H8O2 or something

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +328,80 @@ def Bethe_stopping(
...
+    where :math:`n_s` is the number density of scatterers, :math:`\ln{(\bar{\Lambda})}`,
+    :math:`q_{\alpha, \beta}`, :math:`m_{\alpha, \beta}` and :math:`v_{\alpha, \beta}` are the particle charge,
+    mass, and velocity in the rest frame, respectively. The subscript :math:`\alpha`
+    refers to the beam species whereas :math:`\beta` refers to the target species.
+    :math:`\ln{\left(\bar{\Lambda}\right)}` is the averaged Coulomb logarithm, given by :math:`\ln{(\lambda_D / \langle b_0 \rangle)}`
+    where :math:`\lambda_D` is the Debye length and :math:`\langle b_0 \rangle` is the
+    average value for the distance of closest approach associated with the
+    classical Coulomb potential in the center of mass frame, given as :math:`b_0 = q_{\alpha}
+    q_{\beta} / 4 \pi \eps_0 \mu \bar{u}^2`, where :math:`\bar{u}` is the average relative velocity.
+    """
```
pheuer, line 840: Modify to work for fractions of elements in a compound, calculate for each species then weight with fraction?

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -546,6 +586,161 @@ def inner_Bethe_stopping(v, n_e):
...
+        ]:
+            raise ValueError(
+                "Please pass one of 'differential cross-section' or 'mean square rate' "
+                "for the scatter routine keyword."
+            )
+
+        self._scattering_method = scatter_routine_type
+        self._scattering_routine = scatter_routine
+
+        self._set_target_attributes(target, target_rho)
```
pheuer, line 743: Would be nice if attributes like this were stored on the Grid object

tests/simulation/test_particle_tracker.py
```diff
@@ -648,6 +653,171 @@ def test_particle_tracker_Bethe_warning(
...
+            "ALUMINUM",
+            [1, 5, 10] * u.MeV,
+            [16.33, 215.93, 709.23] * u.um,
+            [0.9972, 10.29, 31.70] * u.um,
+        )
+    ],
+)
+@particle_input
+@validate_quantities
+def test_particle_tracker_ranges_straggling(
```
pheuer, line 731: "stopping_and_scattering"

src/plasmapy/formulary/collisions/misc.py
JaydenR2305: Is this what you were looking for?
https://github.com/PlasmaPy/PlasmaPy/blob/d1ccae120252b88bb75f8c7d41bc418bc660edbe/src/plasmapy/simulation/particle_tracker/particle_tracker.py#L1013

Did you want me to add a comment about the minus sign to avoid confusion?

pheuer: Great, probably a good idea to just mention that dEdx is by convention positive, so then this gets a minus sign

changelog/2712.feature.rst
JaydenR2305: Am not sure why this is appearing here. It says the associated commit isn't from this PR:

image

I must have messed this up when I merged my previous PR for stopping into this one? I'll be sure to add a changelog for this PR

pheuer: I have a guess as to what happened: 

1) Way back when you created this PR you maybe based it on the not-yet-merged stopping PR, which had this changelog in it. 

2) That PR was merged, and then a new plasmapy version was released, which means the changelog got incorporated into the release and the file went away. 

3) Now the file is still in this PR, so git thinks you want to add it back into main. 

I think you should just be able to delete this file on this branch and have no issues?

JaydenR2305: Oooh yeah that's right. I forgot that they're all deleted after a release. Good catch!

src/plasmapy/simulation/particle_tracker/particle_tracker.py
JaydenR2305: This comment is incorrect. Only the mean excitation energy is constant over a given grid and this excitation energy is passed as an argument to the `add_stopping` method. The number density is a quantity in the grid, so it can vary from location to location.

I'll change this comment to reflect that each grid only has a single value for the mean excitation energy. For the inner function, stopping interpolators expect a `Callable` that takes the velocity of the incident species and a number density of the scatterers, so I must have changed the signature at some point and forgot to update the comment 🤔 

Would you recommend I change this variable from `n_e` to `n_s` to reflect that we're talking about the species responsible for electronic scattering?

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -436,6 +437,40 @@ def scattering_integrand(theta, B):
...
+    L_rad : `~astropy.units.Quantity`
+        The radiation length of the target material. See notes for more details.
+    dt : `~astropy.units.Quantity`
+        The timestep being used in the Monte Carlo simulation.
+
+    Returns
+    -------
+    theta_rms : `~astropy.units.Quantity`
+        The RMS scattering angle in radians.
+    """
```
pheuer, line 1167: Add a reference to the Highland paper in the docstring, and maybe a few notes on the regimes of relevance as described in the paper (e.g. E_s may not be constant below Z=20).

Also include some info on the "radiation length": that's a bit of an odd quantity people may not be familiar with (I'm not outside of this context)

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -436,6 +437,40 @@ def scattering_integrand(theta, B):
...
+    # Fitting constant, value provided by Highland
+    E_s = 17.5 * u.MeV
+    # The length the particle will traverse in this timestep
+    L = v * dt
```
pheuer: For this function, let's take L as the parameter rather than dt. That makes more sense because another common usage of this equation is just to estimate the total scattering through a foil, where L would be the thickness but the user would have to go calculate dt. 

Then calculate L before calling this in the MC code

src/plasmapy/formulary/collisions/misc.py
JaydenR2305: I just pushed a draft of the notes section for the `Highland_scattering` function in the collision formulary. I'm still not 100% on the details of the radiation length, but it seems like it's related to energy loss due to bremsstrahlung? I just wrote in the doc string that it was the relevant distance scale when we're talking about these "charge interactions"

pheuer: Yeah, the wiki article is pretty helpful https://en.wikipedia.org/wiki/Radiation_length

Makes more sense how it is related to this model then, although interesting that this works so well for ions when the radiation length is defined by electron stopping.

JaydenR2305: Would this potentially mean that the Highland formula isn't meant for proton scattering?🤔

pheuer: I think it is used for protons/ions, note that the 'projectile mass' is refered to rather than just the electron mass

src/plasmapy/formulary/collisions/misc.py
JaydenR2305: image

I'll look through our other sources to see if any of them mention compounds

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +334,157 @@ def Bethe_stopping(
...
+    n: int,
+):
+    integral = [
+        quad(_f_n_mol_integrand, 0, np.inf, args=(x, n))[0] for x in ϑ
+    ] * u.dimensionless_unscaled
+
+    return integral / factorial(n)
+
+
+def Bethe_Moliere_scattering(
```
pheuer: I always just hear this called Moliere scattering- I think Bethe mostly just described Moliere's results in english? Either way, I think it's best to stick to that name and then just cite Bethe's paper

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +334,157 @@ def Bethe_stopping(
     return Bethe_stopping_lite(I.si.value, n.si.value, v.si.value, z) * u.J / u.m
 
 
+def _f_n_mol_integrand(
```
pheuer: This is probably old but you don't actually need to compute the integral, just type up the data in Table II and then interpolate over that to get f0, f1, f2, then plug those into Eq. 25 to get the angle probability distribution f(theta)dtheta

You could maybe try explicitly doing the integral if you want, just to test the table, but I think the table will be much faster and avoids numerical integration.

src/plasmapy/formulary/collisions/misc.py
pheuer: Also if this first PR just covers single elemental scattering, that would be a good start, we could add compounds next.

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -27,10 +33,143 @@
...
+    [1e-3 * 2e-8, 1e-3 * 3.831, 1e-3 * 0.8326],
+    [1e-3 * 2e-10, 1e-3 * 2.527, 1e-3 * 0.5368],
+    [1e-3 * 5e-13, 1e-3 * 1.739, 1e-3 * 0.3495],
+    [1e-3 * 1e-18, 1e-3 * 0.9080, 1e-3 * 0.1584],
+    [1e-3 * 3e-25, 1e-3 * 0.5211, 1e-3 * 0.0783],
+    [1e-3 * 1e-32, 1e-3 * 0.3208, 1e-3 * 0.0417],
+    [1e-3 * 1e-40, 1e-3 * 0.2084, 1e-3 * 0.0237],
+]
+
+_f_mol_spline = make_interp_spline(_ϑ_tabulated, _f_mol_tabulated)
```
pheuer, line 182: Is this a 2D spline? Make sure it's treating the three columns independently, e.g. fitting them each with separate splines. Can you check the interpolator vs. the table data to make sure you have a good interpolation over the full range?

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +464,213 @@ def Bethe_stopping(
...
+            )
+
+        B_coefficient = np.asarray([1 / B**i for i in range(3)])
+
+        if use_interpolator:
+            f_n = _f_mol_spline(ϑ.cgs).T
+        else:
+            f_n = [_f_mol_n(ϑ, i) for i in range(3)]
+
+        return np.sum(B_coefficient * f_n, axis=0)
```
pheuer: If this is the full RHS of eq 25 should it include a factor of ϑ multiplying all of these terms?

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +464,213 @@ def Bethe_stopping(
...
+
+    if not return_rms:
+        return scattering_integrand
+    else:
+        # Eq. 30
+        theta_w = χ_c * np.sqrt(B)
```
pheuer: Why not just use pi as the upper bound, and not rely on this approximation that neglects the higher order terms?

src/plasmapy/formulary/collisions/misc.py
```diff
@@ -325,6 +464,213 @@ def Bethe_stopping(
...
+        # Use Gaussian approximation to determine when the distribution has
+        # reached a sufficiently low value
+        upper_bound = 2 * theta_w
+
+        total_scattering_probability, _total_probability_error = quad(
+            scattering_integrand, 0, upper_bound
+        )
+
+        mean_squared, _mean_squared_error = quad(
+            lambda theta: theta**2
```
pheuer: I don't entirely understand the theta**2 here for calculating the RMS. Is that because the theta is left off of eq. 25 earlier?   

Shouldn't the scattering integrand be squared? https://en.wikipedia.org/wiki/Root_mean_square

src/plasmapy/formulary/collisions/misc.py
JaydenR2305: TODO: Figure out heuristic for including higher-order corrections when necessary

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -844,6 +1056,25 @@ def _adaptive_dt(self) -> NDArray[np.float64] | float:
...
+            # can only lose 5% per timestep (or 0.25 MeV)
+            max_delta_E = (
+                self._stopping_parameter
+                * S[tracked_particles]
+                / (np.abs(dS_dx[tracked_particles]))
+            )
+            # max_delta_E = np.maximum(max_delta_E, (0.05 * u.MeV).to(u.J).value)
+
+            timestep = np.nan_to_num(
+                max_delta_E / (S[tracked_particles] * current_speeds), nan=np.inf
```
JaydenR2305, line 1064: Image

EDIT: dS/dx should read dS/dT, where T is the beam's kinetic energy

JaydenR2305: where S is the stoppingf power

JaydenR2305: computing a derivative from a spline in log-log space:
image

Issue #2632, opened 2024-04-12
Show correct version in title of `latest` docs


namurphy:
The title in our `latest` docs is "PlasmaPy Documentation — PlasmaPy 0.1.dev50 documentation".  We should update this to use the current month.  This is probably closely related to #2630.

PR #2751, opened 2024-07-10
Base `plasmapy.__version__` on date and git hash outside of official releases

### LINKED ISSUE #2630: Make `plasmapy.__version__` draw from most recent tag in development installations
Currently when we do an editable install of a development version of PlasmaPy (i.e., with `pip install -e .`), it gives us something like:

```pycon
>>> import plasmapy
>>> plasmapy.__version__
'2023.1.1.dev943+gb1495a48'
```
I'm not sure why, but it's giving us last year.  We should update it so that it gives something like `2024.4.12.dev943+gb1495a48` when the day is (or the last commit was on) 2024 Apr 12.  The last part of it should be the git hash, essentially the same way that it is done now.  I'm not sure if we can use [`setuptools_scm`](https://github.com/pypa/setuptools_scm) for this, but it's worth looking into.

namurphy:
This is an experimental PR to try to fix problems with the version. Rather than following this PR, I think the more robust fix would be to modify the GitHub workflow so that it tags the version and then creates the release branch off of that tagged version.

Closes #2630. Related to #2632. Partially superseded by #2756.


docs/conf.py
```diff
@@ -205,6 +205,7 @@
     "includehidden": False,
 }
 htmlhelp_basename = "PlasmaPydoc"
+github_url = "https://github.com/PlasmaPy/PlasmaPy"
```
namurphy, line 208: Need to verify that the "Edit on GitHub" button on the docs is working. 

There's chance that `github_url` be in the `html_theme_options` dict above instead of as a separate variable.

pyproject.toml
```diff
@@ -147,7 +145,7 @@ exclude = ["docs/automodapi", "src/plasmapy/_dev"]
 namespaces = false
 
 [tool.setuptools_scm]
-write_to = "src/plasmapy/_version.py"
+version_file = "src/plasmapy/_version.py"
```
namurphy, line 148: `write_to` was deprecated.

tests/test_version.py
```diff
@@ -0,0 +1,13 @@
...
+
+from plasmapy import __version__
+
+
+@pytest.mark.skipif("dev" not in __version__, reason="test is irrelevant for ")
+def test_dev_version():
+    today = date.today()  # noqa: DTZ011
+    year = today.year
+    month = today.month
+    assert __version__.startswith(f"{year}.{month}.")
```
namurphy, line 13: I don't know if we actually want this in the test suite, but I'll keep it here until it's closer to ready to merge.

namurphy:
Closing this in favor of #2756, and perhaps also #2752.

namurphy:
Update: My current understanding is that `_dev/scm_version.py` only gets used when the package is *not* installed. This used to be possible or common if we did `import plasmapy` from the top-level directory of this repository, which would import the contents of the `plasmapy` directory without installing it. Now since we have it in `src/plasmapy`, this means that we have to do a significant workaround to be able to import it without installing it.

PR #2752, opened 2024-07-10
Update the workflow for minting a release to create the release branch after it is tagged on `main`

### LINKED ISSUE #2630: Make `plasmapy.__version__` draw from most recent tag in development installations

namurphy:
## Description

This PR makes it so that the release branch is created _after_ the tasks to build the changelog, etc. are completed.

## Motivation and context

When creating the string for a development version of PlasmaPy, `setuptools-scm` guesses the next version based off of the previous tag on `main`. However, the `mint-release.yml` workflow currently creates a tag on the release branch. As a result, we have been getting dev versions like `2023.1.1.dev1015+g9f40495d` even though our most recent release was `2024.5.0`.  

## Related issues

Closes #2630. Probably. Related to #2632.  Alternative to #2751.


.github/workflows/mint-release.yml
```diff
@@ -61,10 +57,13 @@ jobs:
...
         git tag "v${{ github.event.inputs.tag }}" -m "Version v${{ github.event.inputs.tag }}"
-        git tag
 
-    - name: Push tag, branch state
+    - name: Create or checkout release branch
+      run: |
+        BRANCH=$(echo $VERSION | awk 'BEGIN{FS=OFS="."} {$3="x"} 1' | cut -f1,2,3 -d'.' | awk -F'rc' '{print $1 ORS $2}')
+        git checkout $BRANCH 2>/dev/null || git checkout -b $BRANCH
+
+    - name: Push tag and branch to remote repository
```
namurphy, line 66: 🛗 Moved from above, where the only change is extracting `$VERSION` so it's defined globally.

namurphy:
I'm a bit hesitant about this PR since the workflow would require us to bypass branch protection rules.  I think the better approach will be to remember to merge the PR from the release branch back into `main` **without squashing it**.

We can probably close this, but I want to get the other unrelated changes into a separate PR.

PR #2754, opened 2024-07-10
Automatically regenerate changelog index

### LINKED ISSUE #2753: Dynamically generate `docs/changelog/index.rst`
### Description of improvement

We should regenerate the file containing the index of changelogs (`docs/changelog/index.rst`) so that the "Unreleased Changes" page from `dev.rst` is only included if there are any towncrier changelog entries in the `changelog/` directory.  

### Motivation

Our changelog index file currently begins with `dev`, which means that "Unreleased Changes" appears even for official releases. We should make it so that "Unreleased Changes" only appears in the index when 

### Implementation strategy

We should probably create and call a function in `docs/conf.py` that generates `docs/changelog/index.rst` when doing a doc build.  This could implemented similarly to how `authors.rst` is auto-generated.

If there are any valid changelog entries in the top-level `changelog/` directory, we should include `dev` at the beginning of the list of pages in `docs/changelog/index.rst`.  Otherwise, we should exclude it.

I don't think we'd need to regenerate `dev.rst` since it includes `:orphan:`.

There might be an easier way to do this (maybe with `ifconfig`?) but that would probably also have similar complications and might actually be more difficult to implement.

### Additional context

_No response_

namurphy:
The file `docs/changelog/index.rst` is an index for the changelog files for releases. However, it presently includes a line for unreleased changes (via `docs/changelog/dev.rst`), even if there are no unreleased changes. 

This PR makes it so that this index file is autogenerated during the documentation build. The line for unreleased changes is only included when where are unreleased changes. I tested this locally by building the docs after deleting all of the changelog entries.

We also had been automatically updating `docs/changelog/index.rst` as part of the process for updating citations. By auto-generating this file, we no longer need to do that.

Closes #2753.


.github/scripts/citation_updater.py
```diff
@@ -65,18 +65,6 @@ def update_citation_files(args) -> None:
...
-        whatsnew_index_text,
-    )
-    with whatsnew_index_rst_file.open("w") as f:
-        f.write(whatsnew_index_text)
-
-
 if __name__ == "__main__":
     args = parse_arguments()
     update_citation_files(args)
-    update_whatsnew(args)
```
namurphy, line 82: `update_whatsnew` is no longer necessary because the changelog index file is being regenerated during every documentation build.

namurphy:
I just looked at the changelox index [for the stable docs](https://docs.plasmapy.org/en/stable/changelog/index.html) and [for the latest docs](https://docs.plasmapy.org/en/stable/changelog/index.html) just after the release so that there are no unreleased changes, and "Unreleased changes" is not appearing in the toctree which is the intended behavior!

PR #2755, opened 2024-07-11
Update `docs/conf.py`

namurphy:
- Simplified code related to `version` and `release`
 - Removed a deprecated call from `datetime`
 - Reorganized imports
 - Minor configuration changes


docs/conf.py
```diff
@@ -25,41 +25,35 @@
...
+version = __version__
 
 if release.startswith("0"):
     warnings.warn(f"Incorrect version in documentation build ({release = })")
 
-release = "" if release == "unknown" else release
-parsed_version = pkg_resources.parse_version(release)  # deprecated
-release = parsed_version.public
-version = ".".join(release.split(".")[:2])  # short X.Y version
-revision = parsed_version.local[1:] if parsed_version.local is not None else ""
```
namurphy, line 61: Previously, `release` was supposed to be the full version and `version` was supposed to be the `X.Y` without patch version.  But, there isn't really a need for us to make a distinction between them.

.pre-commit-config.yaml
```diff
@@ -79,13 +79,13 @@ repos:
     - tomli
 
 - repo: https://github.com/crate-ci/typos
-  rev: v1.22.9
+  rev: typos-v0.10.26
```
namurphy, line 82:
```suggestion
  rev: v1.23.2
```

namurphy:
For the "Add no changelog label" workflow, this PR has been getting the following error.  I'm not sure what's causing it, but it may be intermittent, or perhaps `GITHUB_TOKEN` has expired or something.
```
RequestError [HttpError]: Resource not accessible by integration
```

Issue #2757, opened 2024-07-11
Intermittent Codecov failures because downloader tests are sometimes skipped


namurphy:
We've been getting some variations in the results of the **codecov/project** check in CI of ∼0.8%. The cause seems to be because tests related to the downloader are sometimes run and sometimes skipped.

I'm not sure how best to address this.  Some possibilities would be:
 - Exclude these lines/files from coverage reports
 - Skip downloader tests entirely when checking coverage (which we could do by including them in the Python 3.12 tests but excluding them in the Python 3.11/coverage tests)
 - Set the tolerance for **codecov/project** to be ∼0.9%.
 - Cache the downloaded data sets within PlasmaPy or via GitHub Actions
 - Use mocking for the downloader so that a connection to the World Wide Web™ isn't needed

PR #2763, opened 2024-07-15
Re-write 'Test independence and parametrization' section of testing guide

shaunagm:
## Description

This PR re-writes the [Test independence and parametrization](https://docs.plasmapy.org/en/latest/contributing/testing_guide.html#test-independence-and-parametrization) section of the testing guide to use extremely simple math.

## Motivation and context

I re-wrote the example math in this section to minimize cognitive load for the reader. This way they only need to think about the concept being taught, rather than also trying to remember the example math.


shaunagm:
Hi @namurphy - not sure how to handle the codecov and pre-commit failures, the details are not informative enough for me to fix whatever needs fixing. (I do not have pre-commit hooks set up, which may be part of that particular failure.) Please let me know how to proceed, but otherwise I think everything should be good.

namurphy (commented):
🎉 Thank you, @shaunagm!  It will helpful to have a description of test parametrization based on simpler math. I have two pretty minor suggestions and then we'll be just about ready to merge it.

✅ We can address the pre-commit failures by commenting on this PR with "pre-commit.ci autofix". 

👀 The Codecov failure is unrelated so we don't need to worry about that (#2757). 

Thank you again! 🙏🏻

shaunagm:
Made the changes you suggested. Does "pre-commit.ci autofix" need to be in a separate comment or was your original comment enough? (I believe that would make it a [speech act](https://en.wikipedia.org/wiki/Speech_act) or whatever the written equivalent is)

namurphy:
> Made the changes you suggested. Does "pre-commit.ci autofix" need to be in a separate comment or was your original comment enough? (I believe that would make it a [speech act](https://en.wikipedia.org/wiki/Speech_act) or whatever the written equivalent is)

Correct, "pre-commit.ci autofix" would need to be in a separate standalone comment.  And thanks!

shaunagm:
pre-commit.ci autofix

namurphy (approved):

PR #2765, opened 2024-07-15
Update comment that gets posted to pull requests

namurphy:
- I removed the notice about the switch to an `src` layout, since it's been several months.
 - I shortened a few parts, such as the links to the contributor guide.  
 - I added more emojis. 😅


ejohnson-96 (approved):
👍

PR #2766, opened 2024-07-15
Generate table of global substitutions to put in the documentation guide

### LINKED ISSUE #1550: Automatically construct a table of substitutions from `_global_substitutions.py` in doc guide
We have a bunch of substitutions and links defined in [`docs/global_substitutions.py`](https://github.com/PlasmaPy/PlasmaPy/blob/main/docs/_global_substitutions.py), and right now the only way to know what links & substitutions are there is by going directly to that file.  We could automatically generate a table of the substitutions and put it in the doc guide.  

I expect that the implementation of this issue would be similar to that of #1222, so these two issues could be addressed at around the same time (though in different pull requests).

namurphy:
This PR adds functionality to `docs/_global_substitutions.py` to generate a table of the global substitutions that are used in the documentation. Here is a link to the RTD [preview of the substitutions section](https://plasmapy--2766.org.readthedocs.build/en/2766/contributing/doc_guide.html#substitutions) which contains the table.

The table is quite long, so I put it under a collapsible tab using `sphinx_collapse`.  The table is also quite wide because one of the substitutions includes some sentences to caution about unstable APIs, but that would take a while to fix so I'm okay with leaving it as-is.

🎨 I also expanded usage of the `:rst:` role so that in-line reStructuredText code snippets get formatted with pygments, which I think makes it considerably more readable.  

😩 All the nested backticks reminded me of writing the documentation guide in the first place!  

Closes #1550.


changelog/2766.doc.2.rst
```diff
@@ -0,0 +1 @@
+Renamed :file:`docs/_cff_to_rst.py` to :file:`docs/_author_list_from_cff.py`.
```
namurphy, line 1: 🏷️ Mentioning in the name that this is for creating an author list should lead to more readable code.

ejohnson-96 (approved):
👍

namurphy:
@ejohnson-96 — thank you for the prompt review!

PR #2772, opened 2024-07-19
Fix tests for `numpy == 2.0.0` and update requirements

### LINKED ISSUE #2724: Fix newly failing tests and remove version upper limit on NumPy
NumPy v2.0.0 was released yesterday. Because we had ∼5 new test failures come up, #2722 added a requirements that `numpy < 2.0.0`.  We should address those test failures, and unpin NumPy.

Thankfully, the test failures seem pretty insignificant and are mostly related to how values are printed out, so there _probably_ won't be any major issues that arise from the release for us.

namurphy:
This PR updates some minor issues we ran into when NumPy 2.0.0 came out.  There weren't any major problems, but there were a few tests and such where changes were needed.

I also updated some pre-commit hooks

See also #2722. Closes #2724.


namurphy:
pre-commit.ci autofix

.pre-commit-config.yaml
```diff
@@ -54,7 +54,7 @@ repos:
     exclude: .github/labeler.yml|.pre-commit-search-and-replace.yaml
 
 - repo: https://github.com/mattlqx/pre-commit-search-and-replace
-  rev: v1.0.5
+  rev: v1.1.3
```
namurphy, line 57: I ran into some problems with v1.1.2 a while back (probably in the pre-commit autoupdate PRs), but v1.1.3 seems to be working. Yay!

namurphy, line 57:
```suggestion
  rev: v1.0.5
```
Nevermind, it seems to be timing out during the installation phase...

pyproject.toml
```diff
@@ -59,20 +59,20 @@ dependencies = [
...
-  "scipy >= 1.8.0",
+  "scipy >= 1.9.0",
   "setuptools >= 61.2.0",
-  "tqdm >= 4.60.0",
+  "tqdm >= 4.65.0",
   # Use loose requirements for voila because it impacts ipykernel version
   "voila >= 0.2.7",
   "wrapt >= 1.14.0",
-  "xarray >= 2022.3.0",
+  "xarray >= 2022.6.0",
```
namurphy, line 75: These updates are all based on the SPEC 0 recommended policy (i.e., that we drop support for dependencies that were released ≥ 24 months ago).

tests/formulary/test_dimensionless.py
```diff
@@ -162,9 +166,6 @@ def test_Hall_parameter() -> None:
     with pytest.warns(u.UnitsWarning):
         Hall_parameter(n, T, B, ion, particle, V=100)
 
-    with pytest.raises(TypeError):
-        Hall_parameter(n, T, B, ion, particle, coulomb_log="test")
-
```
namurphy, line 167: This gave a `TypeError` for older versions of NumPy, and a `ValueError` for newer versions. In this case, I don't think the specifics of the exception are particularly important.

PR #2774, opened 2024-07-19
Generate OS-aware requirements

### LINKED ISSUE #2773: Verify that rebuilding requirements works across operating systems
We currently use `nox -s requirements` to regenerate requirements files. However, I've primarily used this on Linux, and the corresponding [workflow](https://github.com/PlasmaPy/PlasmaPy/blob/main/.github/workflows/update-pinned-reqs.yml) runs on Linux as well. Hence, I don't know how well this works on Windows or macOS. We should compare running `nox -s requirements` on Windows, macOS, and Linux at the same time.  We could perhaps edit the workflow to run on Windows (macOS), and then run the workflow, and see how well CI does.

There was a [comment](https://github.com/PlasmaPy/PlasmaPy/issues/2647#issue-2264244121) made by @rocco8773 in #2647 that brought this up:

> When `uv` generates a requirements files it by default adds OS specific packages to that file. So, if you generate the file on a Windows machine without the `--no-deps` flag, then all associated CIs and RTD runners will fail. Again, not documented anywhere.

If we do run into problems, I'm not sure yet how best to address them. One possibility would be to use the `--python-platform` flag of `uv pip compile` to create requirements files that are specific to the different operating systems that get used in CI.

JaydenR2305:
Updates the requirements `nox` session to generate requirements for each supported OS: Linux, MacOS, and Windows. Under the hood this takes advantage of `uv`'s `--python-platform` flag

Resolves #2773


namurphy (approved):
Thank you for doing this, and just in time for the release!  I made a few minor adjustments since we use the hash of one of the requirements files as a means of cache invalidation for GitHub workflows.  Thank you again!

JaydenR2305:
> I made a few minor adjustments since we use the hash of one of the requirements files as a means of cache invalidation for GitHub workflows.

Nice catch!

PR #2790, opened 2024-07-26
Extend `@particle_input` to `ParticleListLike` annotations

JaydenR2305:
Currently, the `@particle_input` decorator *supports* `ParticleList` annotations but does not provide robust support for casting for `ParticleListLike` objects. This PR will implement support for casting of iterables and strings to `ParticleList` objects for parameters annotated with `ParticleListLike`.

### DIFF
--- a/src/plasmapy/particles/decorators.py
+++ b/src/plasmapy/particles/decorators.py
@@ -586,6 +586,14 @@ class _ParticleInput:
             if not hasattr(argument, "__len__") or len(argument) != 2:
                 raise ValueError(f"The length of {argument} must be 2.")
             return Particle(argument[0]), Particle(argument[1])
+        elif annotation in [ParticleList, ParticleListLike]:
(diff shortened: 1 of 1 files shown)


namurphy:
> I'm trying to figure out if there's some type in Python that includes the usual `Iterable` types like tuples, lists, and sets, but doesn't include strings

Hm...I don't know of any type or typing construct like that, unfortunately.  

This reminds me that we used to have a problem where `ParticleList("He")` was the same as `ParticleList(["H", "e-"])`... 👀

JaydenR2305:
> This reminds me that we used to have a problem where `ParticleList("He")` was the same as `ParticleList(["H", "e-"])`... 👀

Yeah that *should* be fixed now with the above logic. I'll double check just to be 100% :)

namurphy (approved):
Thank you for adding this, and my apologies for taking so long to merge it!  I expect that this will be useful.  I added a changelog entry, and will merge it shortly.

Issue #2764, opened 2024-07-15
Unpin requirement of `sphinx == 7.3.7`


namurphy:
In #2762, we ended up getting a bunch of new warnings in our documentation build with the release of Sphinx 7.4 (see below).  I'm updating that PR to pin `sphinx == 7.3.7`. In a week or two, we should address these errors and remove this restriction on Sphinx.

These are some of the warnings that we started getting. Unfortunately, the warnings do not state which file the warnings are occurring in, which will make these hard to fix.  

```
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.fit_functions' with 'fit_functions'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.nullpoint' with 'nullpoint'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.swept_langmuir' with 'swept_langmuir'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.time_series' with 'time_series'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.swept_langmuir.floating_potential' with 'floating_potential'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.swept_langmuir.helpers' with 'helpers'.
WARNING: Summarised items should not include the current module. Replace 'plasmapy.analysis.swept_langmuir.ion_saturation_current' with 'ion_saturation_current'.
```

JaydenR2305:
That's interesting because I'm just getting a build error on windows:

```
nox > Running session docs
nox > Creating virtual environment (uv) using python.exe in .nox\docs
nox > uv pip install -r ci_requirements/docs-3.12.txt .
nox > Command uv pip install -r ci_requirements/docs-3.12.txt . failed with exit code 1:
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of numpy{platform_system ==
      'Windows'} are available:
          numpy{platform_system == 'Windows'}==1.3.0
          numpy{platform_system == 'Windows'}==1.4.1
          numpy{platform_system == 'Windows'}==1.5.0
          numpy{platform_system == 'Windows'}==1.5.1
          numpy{platform_system == 'Windows'}==1.6.0
          numpy{platform_system == 'Windows'}==1.6.1
          numpy{platform_system == 'Windows'}==1.6.2
          numpy{platform_system == 'Windows'}==1.7.0
[...]

PR #2793, opened 2024-07-26
Fix problem with installation step in Nox sessions to build documentation

namurphy:
Previously, we included a line in the Nox sessions to build documentation like:
```python
session.install("-r", docs_requirements, ".")
```
The problem is that this does not explicitly include the `docs` dependency set, which has the consequence of hiding errors resulting from inconsistencies between the requirements file and `docs` requirements set.

🔧 This PR makes it so that the installation step explicitly includes the `docs` dependency set via `.[docs]`. 

I'm also repinning `sphinx==7.3.7` (sigh!) since the problem with the `docs` Nox session masked the warnings that we were trying to fix (see also #2762 and #2764).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -283,7 +283,7 @@ def docs(session: nox.Session) -> None:
     """
     if running_on_ci:
         session.debug(doc_troubleshooting_message)
-    session.install("-r", docs_requirements, ".")
+    session.install("-r", docs_requirements, ".[docs]")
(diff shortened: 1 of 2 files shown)


namurphy:
@JaydenR2305 — The good news is that you helped us find and fix this bug in the installation step for the documentation build in CI, so thank you! 🥇 The less than good news is that the Sphinx issue hasn't been resolved, so we have to re-pin it to an old version, alas.

PR #2809, opened 2024-07-30
Add an seedling emoji

cacsphysics:
I added a seedling emoji in the index page.


cacsphysics:
I think this emoji should stay or be replaced with a broccoli emoji.

namurphy (approved):

PR #2833, opened 2024-08-01
Add example to docstring of ParticleTracker 

### LINKED ISSUE #2532: Add examples to docstring of `ParticleTracker`
Currently, the docstring for `ParticleTracker` ([permalink](https://github.com/PlasmaPy/PlasmaPy/blob/2c1ee2e74e86d9519d1a306a6f78413683ca9a02/src/plasmapy/simulation/particle_tracker/particle_tracker.py#L32)) doesn't contain any examples. It would be helpful to add a simple working example that shows how to use it.

The example shouldn't be too computationally expensive because it will be run every time the documentation is built, so I was thinking of perhaps running it for 10–20 steps.

One possibility would be to adapt some of the [particle tracker tests](https://github.com/PlasmaPy/PlasmaPy/blob/2c1ee2e74e86d9519d1a306a6f78413683ca9a02/tests/simulation/test_particle_tracker.py).

s-gerow:
Closes #2532 
Adds an example use of the ParticleTracker, simulating a single particle with an initial velocity moving through a uniform magnetic field.


src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -133,6 +133,30 @@ class to the ``run`` method.
...
+    >>> Bz = np.full(grid_shape, 1) * u.T
+    >>> grid.add_quantities(B_z=Bz)
+    >>> x = [[0, 0, 0]] * u.m
+    >>> v = [[1, 0, 0]] * u.m / u.s
+    >>> termination_condition = TimeElapsedTerminationCondition(6.28*u.s)
+    >>> save_routine = IntervalSaveRoutine(6.28*u.s/10)
+    >>> simulation = ParticleTracker(grid, termination_condition, dt=1e-2*u.s, save_routine=save_routine, field_weighting="nearest neighbor")
+    >>> simulation.load_particles(x, v, example_particle)
+    >>> simulation.run()
+    >>> print(simulation.save_routine.results)["time"]
```
s-gerow, line 158:
```suggestion
    >>> print(simulation.save_routine.results["time"][-1], simulation.save_routine.results["x"][-1])
```
This will correctly print the output displayed in line 159.

s-gerow:
pre-commit.ci autofix

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -133,6 +133,45 @@ class to the ``run`` method.
     If the particle's position is not ``NaN``, but the velocity is ``NaN``, the particle has been stopped
     (i.e. it is still in the simulation but is no longer evolved.)
     If both the particle's position and velocity are set to ``NaN``, then the particle has been removed from the simulation.
+
+    Example
+    -----
+    >>> from plasmapy.particles import CustomParticle
+    >>> from plasmapy.plasma.grids import CartesianGrid
```
pheuer, line 140: I think a more common use case would be using `Particle` to create an actual particle, like `Particle('p+')` A 1 kg, 1 C particle isn't a very common scenario!

pheuer (changes requested):
Good idea to add this! Just one suggestion for making the example a bit more relevant to what most people probably plan to do with the tracker, then I think we can merge this in.

JaydenR2305 (approved):
Looks good to me! This is a good intro to the basic workflow of instantiating and running a 'ParticleTracker' simulation. Great work :)

src/plasmapy/simulation/particle_tracker/particle_tracker.py
s-gerow: That's a good point! I made have made and committed that change.

pheuer (approved):
Awesome, thanks again for doing this!

PR #2838, opened 2024-08-23
Magnetic probe analysis

dschaffner:


pheuer:
@dschaffner Maybe a better name would be 'magnetic_flux_probe' or 'dBdt_flux_probe' or something? Magnetic pickup seems too general to me, since someone from another field/application might confuse that with a Hall effect sensor?

Issue #2634, opened 2024-04-12
Enable weekly tests with the JIT compiler to be added in Python 3.13


namurphy:
[JIT compiler]: https://docs.python.org/3.13/whatsnew/3.13.html#whatsnew313-jit-compiler

Python 3.13 is expected to have a [JIT compiler], though it is expected to be disabled by default and thus far only has modest performance gains.  After it is released, we should add a weekly test to make sure that PlasmaPy can be run with it. We will need to wait until after a version of Numba is released for Python 3.13 as well (unless we switch to Cython).

There's a chance that the JIT compiler will encounter some issues with upstream packages (e.g., NumPy & Astropy) so we may need to raise issues in the upstream repositories.  If this doesn't work right away, we can update it to the "revisit in 2026" label instead "revisit in 2025".

Issue #2104, opened 2023-05-04
Re-implement lite-functions using Cython (alternative to #2082)


namurphy:
#2082 proposes that we re-implement lite-functions using Rust.  An alternative would be to re-implement lite-functions using [Cython](https://cython.org/).  

An advantage of Cython is that it'll probably be easier to learn and understand Cython than to learn Rust, assuming you know Python and have no experience with Rust. Cython is a superset of the Python language, with some added type declarations.  It would still be more complex than Numba, but also not require waiting half a year for the first release of Numba after an October Python release.

PR #2841, opened 2024-08-28
Remove Numba as a dependency

### LINKED ISSUE #1342: Handle Numba/LLVM delays for new Python versions
It usually takes a ∼1–2 months for LLVM and hence Numba to become compatible with new versions of Python.  At present, this is blocking us from being compatible with Python 3.10.  This is one of the things that would block us from being able to test PlasmaPy on the development version of Python or alpha/beta/rc releases (see also discussion in #1215).  

One way to get around this would be to have substitute decorators that could be imported when Numba cannot be imported.  Essentially we'd be creating a blank interface within PlasmaPy that could be aliased to `numba`.  

One possibility would be to include in our main `__init__.py` a try/except block that would `import numba as _numba`, and if that raises an `ImportError`, then we'd `import plasmapy.utils.fake_numba as _numba`.  In the rest of the code we'd then do `from plasmapy import _numba` and decorate things like `@_numba.jit`.  I'm not sure which namespace would be best to use.

My preference would be to issue a warning that Numba can't be imported.  We could do this either when PlasmaPy is first imported, or when a function is called that is supposed to have been Numbafied.  I think I prefer the latter.  It'd probably be helpful to include in the warning a mention that for newly released versions of Python, it might take a month or two for Numba to become compatible with it, since that would be really useful context.

namurphy:
## Description

The purpose of this pull request is to test out dropping Numba as a dependency by dropping the `numba.njit` decorator from two formulary lite-functions. The lite-functions still exist as minimal functions that operate on NumPy arrays without any validations.  

## Motivation

It often takes Numba ∼3–5 months before it becomes compatible with a new release of Python each October.  Depending on Numba has the following adverse consequences:

 - PlasmaPy is not able to run on the newest version of Python for a period of ∼3–5 months after the release of a new version of Python.
 - Packages that depend on PlasmaPy (like fiasco) likewise cannot be run on the newest version of Python for ∼3–5 months after a release.
 - The test suite for PlasmaPy cannot be run on release candidates for a new version of Python.

I did some timings (reported in [this comment](https://github.com/PlasmaPy/PlasmaPy/pull/2841#issuecomment-2315920723)) and found that the performance gains from using Numba are ∼1–3% for the charged particle radiography tests, and that our full test suite runs ∼1% quicker without a Numba dependency.

My view is that we get enough performance gains from dropping units and validations for the lite-functions, and that the drawbacks from depending on Numba are much more significant than the slight performance gains.

## Alternatives

I've been hoping we could Cythonize the lite functions so that we could drop the Numba dependency. However, we haven't had the capacity to complete this.

## Related issues and pull requests

Closes #1342.

Related to #1098, #1145, #2104, #2256, #2372, #2634, #2786.


namurphy:
Here are some timings (in progress):

| Test | `main`      | `denumbafy`      |
| ------------- | ------------- |
|  |  |  |
|  |  |  |
|  |  |  |
|  |  |  |

namurphy:
I did some timings to check the performance of functions as 

| inputs | jitted | not-jitted |
|--| ------ | ---------- |
| floats | 295 ns | 772 ns |
| 1000-element arrays | 3.22 μs | 5.72 μs |

For floats, it

StanczakDominik (approved):
Alas, so it is. Perhaps Numba's still better for end users instead of libraries 🫡

namurphy:
> Alas, so it is. Perhaps Numba's still better for end users instead of libraries 🫡

It really is!  Numba is a fantastic package, but the ∼3–5 month delays after a new release of Python outweigh the slight improvements in overall performance, and are causing problems for fiasco which depends on PlasmaPy.  Plus, depending on Numba also means that we can't test against Python beta releases and release candidates before the official release of Python.  Sigh!

Also it is great as always to hear from you!  I hope you have been doing really well!

pheuer (approved):
I also approve - give the scope of PlasmaPy, I think these minor enhancements from njit are not worth the operational issues. Thanks for running the comprehensive timing comparisons!

PR #2857, opened 2024-09-28
Adjust sphinx config for RTD default changes

StanczakDominik:
As per <https://about.readthedocs.com/blog/2024/07/addons-by-default/>,
RTD is currently simplifying their pipeline - they're removing "hidden"
manipulations of the `conf.py` context.

I've been getting emails about needing to add this snippet to preserve
current behavior, so there you go :)


namurphy (approved):

PR #2861, opened 2024-10-04
Add instructions on how to install plasmapy with uv in a virtual envi…

tien-vo:
…ronment and in a uv-managed project. Closes #2858

PR #2866, opened 2024-10-10
Add sphinx-substitutions to docs dependency group, enable it in conf.…

tien-vo:
…py, and populate maxpython into rst_prolog

The ability to dynamically use global substitutions in documentation was mentioned in [#2861](https://github.com/PlasmaPy/PlasmaPy/pull/2861#issuecomment-2394780879). This PR adds the [sphinx-substitution-extensions](https://github.com/adamtheturtle/sphinx-substitution-extensions). To demonstrate its usage, I rewrote the code block in instructions for installing plasmapy with uv (#2861 ) to use |maxpython|.

The substitution variables need to be populated into `prolog_rst`, though. So this may slow down documentation build.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,6 +101,7 @@ extensions = [
     "sphinx_gallery.load_style",
     "sphinx_issues",
     "sphinx_reredirects",
+    "sphinx_substitution_extensions",
     "sphinx_tabs.tabs",
     "sphinx_collapse",
     "sphinxcontrib.bibtex",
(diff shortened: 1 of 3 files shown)


tien-vo:
pre-commit.ci autofix

tien-vo:
Hmm. Odd. The docs builds successfully in my environment, and nox does not give any error similar to the docs build here. @namurphy  maybe there is somewhere else I need to add the dependency requirement to?

PR #2868, opened 2024-10-22
Update saving for CPR synthetic radiography

pheuer:
A couple updates to how Tracker and the CPR module deal with saving, to make it a bit more convenient: 

1) `Tracker` save routines now take an optional `output_name` keyword that provides a basename for the save files. This is useful if you want your output files to have names based on context, or if you're saving multiple files. 

2) `synthetic_radiograph` now also works if you give it a path to an HDF5 file of Tracker output, which is now the default output format for Tracker.

### DIFF
--- a/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -76,9 +76,11 @@ def _coerce_to_cartesian_si(pos):
 
 class _SyntheticRadiographySaveRoutine(SaveOnceOnCompletion):
     def __init__(
-        self, output_directory: Path | None = None, output_name: str = "output"
+        self, output_directory: Path | None = None, output_basename: str = "output"
     ) -> None:
-        super().__init__(output_directory=output_directory, output_name=output_name)
+        super().__init__(
+            output_directory=output_directory, output_basename=output_basename
+        )
 
         self._quantities = {
             "source": (u.m, "attribute"),
@@ -100,7 +102,7 @@ class _SyntheticRadiographySaveRoutine(SaveOnceOnCompletion):
         if self.output_directory is None:
             return
 
-        output_file_path = self.output_directory / Path(f"{self.output_name}.h5")
+        output_file_path = self.output_directory / Path(f"{self.output_basename}.h5")
 
         with h5py.File(output_file_path, "w") as output_file:
             for key, (_units, data_type) in self._quantities.items():
@@ -182,7 +184,7 @@ class Tracker(ParticleTracker):
         Directory for objects that are saved to disk. If a directory is not
         specified then a memory save routine is used.
 
-    output_name : `str`, optional
+    output_basename : `str`, optional
         Optional base name for output files.
 
     fraction_exited_threshold : float, optional
@@ -207,7 +209,7 @@ class Tracker(ParticleTracker):
         ] = "volume averaged",
         detector_hdir=None,
         output_directory: Path | None = None,
-        output_name: str = "output",
+        output_basename: str = "output",
         fraction_exited_threshold: float = 0.999,
         verbose: bool = True,
     ) -> None:
@@ -215,7 +217,7 @@ class Tracker(ParticleTracker):
         # The particle tracker class ensures that the provided grid argument has the proper type and
         # that the necessary grid quantities are created if they are not already specified
         save_routine = (
-            _SyntheticRadiographySaveRoutine(output_directory, output_name)
+            _SyntheticRadiographySaveRoutine(output_directory, output_basename)
             if output_directory is not None
             else None
         )
--- a/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -75,8 +75,10 @@ def _coerce_to_cartesian_si(pos):
 
 
 class _SyntheticRadiographySaveRoutine(SaveOnceOnCompletion):
-    def __init__(self, output_directory: Path) -> None:
-        super().__init__(output_directory=output_directory)
+    def __init__(
+        self, output_directory: Path | None = None, output_name: str = "output"
+    ) -> None:
+        super().__init__(output_directory=output_directory, output_name=output_name)
 
         self._quantities = {
             "source": (u.m, "attribute"),
@@ -98,7 +100,7 @@ class _SyntheticRadiographySaveRoutine(SaveOnceOnCompletion):
         if self.output_directory is None:
             return
 
-        output_file_path = self.output_directory / "output.hdf5"
+        output_file_path = self.output_directory / f"{self.output_name}.h5"
 
         with h5py.File(output_file_path, "w") as output_file:
             for key, (_units, data_type) in self._quantities.items():
@@ -180,6 +182,9 @@ class Tracker(ParticleTracker):
         Directory for objects that are saved to disk. If a directory is not
         specified then a memory save routine is used.
 
+    output_name : `str`, optional
+        Optional base name for output files.
+
     fraction_exited_threshold : float, optional
         The fraction of particles that must leave the grids to terminate the
         simulation. This does not include particles that have never entered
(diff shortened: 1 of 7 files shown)


pheuer:
@namurphy Are we strictly enforcing these pre-commit rules now? I really don't think this function is overly complex (the if-else trees are all straightforward and single level) and I really don't want to break this into a class or bunch of separate input processing functions when it's currently a nice single function...

```
src\plasmapy\diagnostics\charged_particle_radiography\synthetic_radiography.py:
  1106:5 C901 `synthetic_radiograph` is too complex (13 > 10)
  1106:5 PLR0912 Too many branches (14 > 12)
```

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -98,7 +100,7 @@ def save(self) -> None:
         if self.output_directory is None:
             return
 
-        output_file_path = self.output_directory / "output.hdf5"
+        output_file_path = self.output_directory / Path(f"{self.output_name}.h5")
```
namurphy, line 103: 💾 Yay for `pathlib`! 👍🏻

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -201,16 +206,17 @@ def __init__(
...
         verbose: bool = True,
     ) -> None:
         # The synthetic radiography class handles logging, so we can disable logging for the particle tracker
         # The particle tracker class ensures that the provided grid argument has the proper type and
         # that the necessary grid quantities are created if they are not already specified
         save_routine = (
-            _SyntheticRadiographySaveRoutine(output_file)
-            if output_file is not None
+            _SyntheticRadiographySaveRoutine(output_directory, output_name)
+            if output_directory is not None
```
namurphy, line 219: 🤔 Would it make sense to combine `output_directory` and `output_name` to a single `output_path` parameter?

src/plasmapy/simulation/particle_tracker/save_routines.py
```diff
@@ -30,6 +30,9 @@ class AbstractSaveRoutine(ABC):
         Output for objects that are saved to disk. If a directory is not specified
         then a memory save routine is used.
 
+    output_name : `str`, optional
+        Optional string basename for saved files.
+
```
namurphy, line 35:
```suggestion
```
🎨

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: I tried this, the problem is that this can be a directory (if multiple output files are created, e.g. at multiple time steps) or a single file (for a single output file at the end of a simulation). So if this is just one `path` keyword, I would have to depend on looking at file extensions or something to figure out which it was...

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1151,11 +1158,24 @@ def synthetic_radiograph(obj, size=None, bins=None, ignore_grid: bool = False):
     if isinstance(obj, Tracker):
         # results_dict raises an error if the simulation has not been run.
         d = obj.results_dict
+
     elif isinstance(obj, dict):
         d = obj
+
+    elif isinstance(obj, str | Path):
+        d = {}
+        obj = Path(obj)
```
namurphy, line 1167: 🤔 I'm wondering about renaming `d` to something like `results_dict_` to be a bit more explicit.

pheuer: That's definitely a good idea

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -1151,11 +1158,24 @@ def synthetic_radiograph(obj, size=None, bins=None, ignore_grid: bool = False):
...
+        # Create a dictionary of all of the datasets and attributes in the save file
+        # Equivalent to |results_dict|
+        with h5py.File(obj, "r") as f:
+            for k in f:
+                d[k] = f[k][...]
+            for k in f.attrs:
+                d[k] = f.attrs[k][...]
```
namurphy, line 1174: 🤔 I'm wondering if we could use dictionary comprehensions here so that we don't need two `for` loops.  I'm not sure my suggestion is right, though.

```suggestion
        with h5py.File(obj, "r") as f:
            d |= {f[k][...] for k in f} 
            d |= {f.attrs[k][...] for k in f.attrs}
```

It may also help to have a more descriptive name for `k` or to use different variable names for the two loops/comprehensions.

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
namurphy: Makes sense!  I updated my comment around the same time you posted this.

namurphy:
> @namurphy Are we strictly enforcing these pre-commit complexity rules now? I really don't think this function is overly complex (the if-else trees are all straightforward and single level) and I really don't want to break this into a class or bunch of separate input processing functions when it's currently a nice single function...
> 
> ```
> src\plasmapy\diagnostics\charged_particle_radiography\synthetic_radiography.py:
>   1106:5 C901 `synthetic_radiograph` is too complex (13 > 10)
>   1106:5 PLR0912 Too many branches (14 > 12)
> ```

Generally I think of the `ruff` rules for complexity like [`C901`](https://docs.astral.sh/ruff/rules/complex-structure/) as guidelines rather than strict rules, and it's fine to use comments like `# noqa: C901` if the author decides that there's a good reason to ignore the guideline. Usually it's worthwhile to take a look at the code in question to see if there's a way to reduce the complexity and then decide whether it makes sense to simplify the code or to ignore the rule.  It's up to you!

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Good idea!

pheuer: Ok I made the name change, but it turns out you can't make this suggestion because the data contains numpy arrays which aren't hashable? 

`FAILED test_synthetic_radiography.py::test_radiography_disk_save_routine - TypeError: unhashable type: 'numpy.ndarray'`

pheuer: Ok actually this doesn't work  because some of the values in the dict are numpy arrays which are unhashable and that raises an exception: 

`FAILED test_synthetic_radiography.py::test_radiography_disk_save_routine - TypeError: unhashable type: 'numpy.ndarray'`

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Done!

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: I did change k -> key

pheuer:
> It's looking good! Happy to merge this before I do the next release later this week (#2786). I'm taking the remainder of the day off since I'm on the tail end of post-DPP COVID, but will plan on merging it tomorrow. Thank you for doing this!

Sounds good, sorry to hear you have been sick but glad to hear you are getting better! No rush - I'll get the tests passing then you can approve when ready. 

Does the path directory vs. filepath thing make sense? That was annoying me too, but I can't think of a better way around it without making some assumptions about filepath formats...

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Yes good call, I like output_basename: change made

namurphy (approved):

namurphy:
> Does the path directory vs. filepath thing make sense? That was annoying me too, but I can't think of a better way around it without making some assumptions...

Yep, it made more sense after you explained it and after I looked into the code a bit more.  Thank you!

Issue #2870, opened 2024-10-25
Errors when collecting synthetic radiography tests with Python 3.13


namurphy:
When attempting to run tests against Python 3.13 in #2869, we came up against some new errors when collecting some tests in `test_synthetic_radiography.py` which appears to be related to `grids.py`.  I'm not sure what's happening here yet.

```python
==================================== ERRORS ====================================
_ ERROR collecting tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py _
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:561: in <module>
    class TestSyntheticRadiograph:
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:567: in TestSyntheticRadiograph
    tracker_obj_not_simulated = create_tracker_obj()
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:545: in create_tracker_obj
    grid = _test_grid("electrostatic_gaussian_sphere", num=50)
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:120: in _test_grid
    grid.add_quantities(E_x=-Ex, E_y=-Ey, E_z=-Ez, phi=potential)
.nox/tests-3-13-skipslow/lib/python3.13/site-packages/plasmapy/plasma/grids.py:638: in add_quantities
    if key in self.recognized_quantities:
E   TypeError: argument of type 'method' is not iterable
_ ERROR collecting tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py _
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:561: in <module>
    class TestSyntheticRadiograph:
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:567: in TestSyntheticRadiograph
    tracker_obj_not_simulated = create_tracker_obj()
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:545: in create_tracker_obj
    grid = _test_grid("electrostatic_gaussian_sphere", num=50)
tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py:120: in _test_grid
    grid.add_quantities(E_x=-Ex, E_y=-Ey, E_z=-Ez, phi=potential)
.nox/tests-3-13-skipslow/lib/python3.13/site-packages/plasmapy/plasma/grids.py:638: in add_quantities
    if key in self.recognized_quantities:
[...]

PR #2869, opened 2024-10-25
Add Python 3.13 to tests

namurphy:
Now that I'm 95.7% recovered from COVID, here's a PR to finally add Python 3.13 to PlasmaPy's test suite. [Python 3.13](https://docs.python.org/3/whatsnew/3.13.html) was released a few weeks ago while several of us were at the APS DPP meeting in Atlanta, and it has a bunch of awesome new features worth checking out!

We dropped our dependency on Numba recently (#2841), so this PR should go smoothly. 🤞🏻 

This is ahead of our v2024.10.0 release (#2786), which is scheduled for when I'm 96.1% recovered from COVID.  

We'll be dropping Python 3.10 support after this release, but I'm leaving that for a subsequent PR.

PR #2872, opened 2024-10-25
Add discussion in testing guide of how to reduce chances of getting test collection errors

namurphy:
When using pytest, I've noticed that it's harder to diagnose test collection errors than test failures (most recently in #2869 & #2870).  Test collection errors don't result in good tracebacks, at least with the pytest settings from `pyproject.toml`. Test collection errors also prevent pytest from running the full test suite by default, which means we often get incomplete information.

In this PR, I'm planning to add some tips to the testing guide on how to avoid getting test collection errors.  In particular, if a test file contains code that has a chance of failing, that code should be put into either a fixture or an actual test.  Pretty much I want to avoid situations like...

```python
# test_file.py

x = 0 / 0   # causes a ZeroDivisionError in the test collection phase

def test_something():
    assert True
```

I may also look into ways to improve the reporting of test collection errors.

Just wanted to start this PR now as a placeholder since it might take me a month or three to get to it.

### DIFF
--- a/docs/contributing/testing_guide.rst
+++ b/docs/contributing/testing_guide.rst
@@ -150,7 +150,7 @@ Writing Tests
 
 Every code contribution that adds new functionality requires both tests
 and documentation in order to be merged. Here we describe the process of
-write a test.
+writing a test.
 
 Assertions
 ----------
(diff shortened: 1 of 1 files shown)


docs/contributing/testing_guide.rst
```diff
@@ -152,6 +143,15 @@ actual printed output matches the output included in the docstring.
...
 provided in pytest_'s documentation on `test discovery conventions`_.
 
+.. _writing-tests:
+
+Writing Tests
+=============
+
+Every code contribution that adds new functionality requires both tests
+and documentation in order to be merged. Here we describe the process of
+write a test.
```
namurphy, line 153:
```suggestion
writing a test.
```

PR #2877, opened 2024-10-28
Update changelog entries ahead of v2024.10.0 release

namurphy:
This PR makes a few slight updates to recent changelog entries ahead of the v2024.10.0 release which I'm planning to do today-ish.


pheuer:
CL edits look good, and it looks like the pre-commit version increases are working. Does this need a changelog itself? If not, I think it's good to go.

PR #2880, opened 2024-10-29
Drop support for Python 3.10 and bump minimum versions of dependencies

namurphy:
[SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends that packages in the scientific pythoniverse:

 - Support versions of Python from the last three years, and
 - Support versions of dependencies from the last two years.

Following these guidelines, this PR drops support for Python 3.10 following our v2024.10.0 release. PlasmaPy's minimum supported version of Python will be 3.11, starting with our next release (probably v2025.1.0 or v2025.2.0) through about October 2025.  This PR also bumps the minimum required versions of several dependencies.

> [!NOTE]
> I updated which workflows are required for a PR to be merged, so other PRs will be waiting for workflows that won't happen until after this PR is merged.

PR #2887, opened 2024-10-30
Use `typing.Self` in type hint annotations for `particle_class.py`

namurphy:
Once we drop support for Python 3.10, we can use new typing features like [`typing.Self`](https://typing.readthedocs.io/en/latest/spec/generics.html#self)!  This PR applies this to some type hint annotations in `Particle` and `CustomParticle`.

We still need to do `from __future__ import annotations` for delayed evaluation of type hints because we use `ParticleLike` in some annotations before it is actually defined.

This PR will need to wait until after #2880 is merged.

### DIFF
--- a/src/plasmapy/particles/particle_class.py
+++ b/src/plasmapy/particles/particle_class.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Literal
+from typing import Any, Literal, Self
 
 __all__ = [
     "AbstractParticle",
@@ -105,7 +105,7 @@ See Also
 """
 
 
-def _category_errmsg(particle: str | Particle, category: str) -> str:
+def _category_errmsg(particle: str | Self, category: str) -> str:
     """
     Return an error message when an attribute raises an
     `~plasmapy.particles.exceptions.InvalidElementError`,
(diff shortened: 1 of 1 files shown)


src/plasmapy/particles/particle_class.py
```diff
@@ -2634,7 +2634,7 @@ def __hash__(self) -> int:
         return hash(self.__repr__())
 
 
-def molecule(symbol: str, Z: int | None = None) -> Particle | CustomParticle:
+def molecule(symbol: str, Z: int | None = None) -> Self | CustomParticle:
```
namurphy:
```suggestion
def molecule(symbol: str, Z: int | None = None) -> Particle | CustomParticle:
```

src/plasmapy/particles/particle_class.py
```diff
@@ -105,7 +105,7 @@
 """
 
 
-def _category_errmsg(particle: str | Particle, category: str) -> str:
+def _category_errmsg(particle: str | Self, category: str) -> str:
```
namurphy:
```suggestion
def _category_errmsg(particle: str | Particle, category: str) -> str:
```

PR #2892, opened 2024-10-31
Add & update notebook gallery sections

### LINKED ISSUE #2835: "Example Example" all over formulary docs
### Nature of improvement

Example title is doubled in most of the Formulary docs (eg. below)
It does not seem to happen for other sub-packages, but one should check all to be sure.

image

### Location of improvement

This seems to happen because in the  *.rst file we have Examples as the title but also as the nbgallery caption (eg. below).
Removing the caption should solve the issue.

```
Examples
========

For a general overview of how unit-based input works, take a look at the
following examples:

.. nbgallery::
    :caption: Examples
```

### Additional context

_No response_

namurphy:
This PR makes a few updates in the documentation and some comments.  

 - Used section headers instead of `:caption:` for `nbgallery` directives. (When `:caption:` is used, it didn't appear to create anchors that we could link to.)  
 - During the above step, this PR removes duplicate "Examples" headers by not having both a section header and a `:caption:` option in the `nbgallery` directive.  Closes #2835.
 - I made some improvements for consistency, like for spacing for options in reStructuredText directives and "sub-package" → "subpackage" for consistency, (since searches for "sub-package" sometimes yield results for "package").
 - I added missing example notebook gallery sections to several pages, including several `api_static` stub files.


docs/index.rst
```diff
@@ -70,7 +70,7 @@ Example highlights
 
 .. toctree::
    :maxdepth: 1
-   :caption: All the Rest
+   :caption: Project details
```
namurphy, line 73: This is the header used in Astropy's docs, which is a bit more explicit.

docs/api_static/plasmapy.analysis.nullpoint.rst
```diff
@@ -6,3 +6,10 @@
 .. currentmodule:: plasmapy.analysis.nullpoint
 
 .. automodapi:: plasmapy.analysis.nullpoint
+
+Example notebooks
+-----------------
+
+.. nbgallery::
+
+   /notebooks/analysis/nullpoint
```
namurphy, line 15: I keep forgetting that it's helpful to include a thumbnail gallery of example notebooks in the documentation pages for different modules.

PR #2906, opened 2024-11-15
Add multiple electron species to OTS tests

pheuer:
#2905 

This PR adds a second electron species to the multiple species test fixture for the OTS module.


namurphy:
I brought up the RTD build error in #2908, and am attempting to track down if a new release of a dependency is the cause.

pheuer:
@namurphy I forgot about this minor PR, but it would be good to merge in. I think it should be ready now.

namurphy (approved):
LGTM!  Thank you for doing this!

PR #2911, opened 2024-11-18
Optimizing ParticleTracker and cpr.Tracker

pheuer:
This PR contains a long list of mostly minor improvements to the ParticleTracker and charged particle radiography Tracker focused on reducing unnecessary repetitive calculations.

Major changes
- Made all of the axis properties of AbstractGrid cached_properties
- Added a 'soften edges' method to Grid that can help prepare the grid for particle tracking. 

Minor changes (no changelog necessary)
- Wrapped statements in test_particle_tracker with warnings filter to catch a warning about fields not being zero at the edges of the grids. It's ok -these are just tests. 
- ParticleTracker interpolator now handles quantities that aren't provided on the grids a bit more elegantly - we don't have to add arrays of zeros and then interpolate over them anymore. 
- Cached some masks (e.g. _tracked_particle_mask) and reset the cache only as needed - these were cheap to compute, but they were being recomputed a lot, so this helps reduce that.


namurphy:
> @namurphy I added some additional print output which is causing doctest to fail - how does one reset that to accept the new test output as 'normal'?

That's a good question!  We can do that by updating the example section of each docstring directly to have the new output rather than the old.  So it'd be something like changing

```python
    >>> simulation.run()
    >>> simulation.whatevs()
```
to
```python
   >>> simulation.run()
   Awesomely running awesome simulation.
   Finished in 4.6 gigaseconds.
   >>> simulation.whatevs()
```

pheuer:
> That's a good question! We can do that by updating the example section of each docstring directly to have the new output rather than the old. So it'd be something like changing a docstring from

Ah, awesome, I understand how that works now! Thanks!

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -233,11 +233,9 @@ def __init__(
             dt=dt,
             dt_range=dt_range,
             field_weighting=field_weighting,
-            verbose=False,
+            verbose=verbose,
```
pheuer, line 236: This argument should be passed through to ParticleTracker

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -845,8 +843,8 @@ def _coast_to_plane(self, center, hdir, vdir, x=None, mask=None):
 
         # Check that all points are now in the plane
         # (Eq. of a plane is nhat*x + d = 0)
-        plane_eq = np.dot(x - center, normal)
-        assert np.allclose(plane_eq[mask], 0, atol=1e-6)
+        plane_eq = np.dot(x[mask] - center, normal)
```
pheuer, line 846: Just fewer computations if you apply the mask in the first statement

src/plasmapy/plasma/grids.py
```diff
@@ -391,7 +391,7 @@ def unit(self):
     # *************************************************************************
     # 1D axes and step sizes (valid only for uniform grids)
     # *************************************************************************
-    @property
+    @cached_property
```
pheuer, line 394: All of these properties cannot be changed through the user accessible interface after initialization, so there's no reason to not cache them. The axes get called a fair bit in particle tracker, so it's nice to have them cached.

src/plasmapy/plasma/grids.py
```diff
@@ -1107,6 +1107,51 @@ def vector_intersects(self, p1, p2):
 
         return Tmin < Tmax
 
+    def soften_edges(self, width: u.Quantity | None = None) -> None:
```
pheuer, line 1110: This function exists to help get rid of some of those "Quantities at the edges of the grid should go to zero..." warnings the particle tracker always throws. If the grid has a sharp edge, that gradient can mess up the tracking results, so this is a nice pre-filter to run before particle tracing.

src/plasmapy/plasma/grids.py
```diff
@@ -1107,6 +1107,51 @@ def vector_intersects(self, p1, p2):
...
+
+        mask = np.ones(self.shape)
+        for i, pts in enumerate([self.pts0, self.pts1, self.pts2]):
+            w = 0.1 * (np.max(pts) - np.min(pts)) if width is None else width[i]
+
+            sigma = w / 4
+            pad = 2
+            x1 = np.min(pts) + pad * sigma
+            x2 = np.max(pts) - pad * sigma
+            mask *= (
```
pheuer, line 1136: Mask is just two error functions going from 0-1 back to back to create a flat-top Gaussian along each axis.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -105,15 +105,6 @@ class to the ``run`` method.
 
         The default is 'volume averaged'.
 
-    req_quantities : `list` of `str`, default : `None`
```
pheuer, line 108: This really doesn't need to be a keyword - user shouldn't be specifying this

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -175,17 +166,6 @@ class to the ``run`` method.
     6.2999999999... s [[-1.73302...e-08  1.31539...e-05  0.00000...e+00]] m
     """
 
-    # Some quantities are necessary for the particle tracker to function
-    # regardless of other configurations
-    _REQUIRED_QUANTITIES: typing.ClassVar[set[str]] = {
```
pheuer, line 180: Replaced this with a list that gets assembled by different setup methods, so it's more flexible for different configurations.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -195,9 +175,18 @@ def __init__(
         dt=None,
         dt_range=None,
         field_weighting: str = "volume averaged",
-        req_quantities=None,
         verbose: bool = True,
     ) -> None:
+        # Verbose flag controls whether or not output is printed to stdout
```
pheuer, line 180: All of this existed before, lots just got re-ordered

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -309,6 +291,12 @@ def _set_time_step_attributes(
             # Initialize default values for time steps per gyroperiod and Courant parameter
             self.setup_adaptive_time_step()
 
+        sync = (
```
pheuer, line 294: Since there are more ways the tracker can be configured now, I tried to add some more printouts to confirm the configuration at runtime.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -491,33 +406,26 @@ def load_particles(
...
-        r"""
-        Check to ensure the provided quantity string is defined on at least one grid.
-
-        Returns ``True`` if the quantity is defined on at least one grid and
-        ``False`` if none of the grids have defined the specified quantity.
-        """
-
-        return any(quantity in grid.quantities for grid in self.grids)
-
     def _validate_stopping_inputs(
```
pheuer, line 424: The input changes to this function don't really change functionality - they just update it to what it was already supposed to do. Before, you had to include a material for every grid. Now, you can only include it for some, and stopping will only be added on those grids.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -624,6 +544,108 @@ def inner_Bethe_stopping(v, n_e):
...
+
+        This will be called at the beginning of run, since the required
+        quantities depends on how the Tracker has been set up.
+        """
+        # Assemble an array of the resolutions of all of the grids
+        self._grid_resolutions = np.array(
+            [grid.grid_resolution.to(u.m).value for grid in self.grids]
+        )
+
+        # Determine which quantities should be interpolated from each grid
```
pheuer, line 604: This code goes through and figures out which quantities to interpolate on which grids, then creates a dictionary with arrays to hold those values. Reserving that memory now so we don't have to keep creating variables later.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -624,6 +544,108 @@ def inner_Bethe_stopping(v, n_e):
...
+        #  Each quantity is initialized as a zeros array with its respective units
+        # Arrays are ``num_particles`` sized so they don't need to be recreated
+        # when the number of tracked particles changes
+        self._total_grid_values = {
+            field_name: np.zeros(self.num_particles)
+            * AbstractGrid.recognized_quantities()[field_name].unit
+            for field_name in self._interpolated_quantities_any_grid
+        }
+
+        # These variables indicate whether any E or B fields exist
```
pheuer, line 628: The rest of this method sets up the E and B fields

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -681,6 +711,15 @@ def run(self) -> None:
 
             self._push()
 
+            desc = (
```
pheuer: More printout to help the user track the progress of the run. Progress bar is hard for this application since there isn't a definite number of iterations, so all this info is very useful. Also for debugging if you set it up wrong.

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -730,6 +770,9 @@ def _stop_particles(self, particles_to_stop_mask) -> None:
 
         self.v[particles_to_stop_mask] = np.nan
 
+        # Reset the cache to update the particle masks
```
pheuer, line 773: Each of these methods potentially changes the number of tracked particles, so we reset the cache after each one.

tests/simulation/test_particle_tracker.py
```diff
@@ -139,7 +135,7 @@ def test_particle_tracker_constructor_errors(
             CartesianGrid(-1 * u.m, 1 * u.m),
             "no_particles_on_grids_instantiated",
             None,
-            {"req_quantities": ["rho"]},
```
pheuer, line 142: This keyword no longer exist

tests/simulation/test_particle_tracker.py
```diff
@@ -123,6 +116,9 @@ def test_particle_tracker_constructor_errors(
         save_routine = request.getfixturevalue(save_routine)
 
     with pytest.raises(expected_exception):
+        warnings.filterwarnings(
```
pheuer, line 119: Most of the changes in this file are just adding this warning filter, which catches the warning thrown when quantities are not zero on the edge of a grid. That is the case because these are small, contrived grids for testing, and we expect it.

src/plasmapy/utils/data/downloader.py
```diff
@@ -214,7 +214,7 @@ def _update_repo_blob_dict(self) -> None:
             warnings.warn(
                 "URL did not return the expected JSON file: "
                 f"{self._API_BASE_URL}. "
-                f"Response content: {reply.content}. Exception: {err}"
+                f"Response content: {reply.content.decode('utf-8')}. Exception: {err}"
```
pheuer, line 217: This is completely unrelated to this PR, but it popped up in the errors and looked like a simple fix so I just made it...

pheuer:
@namurphy Pending the resolution of those two docs bugs, this PR is ready to go. It's pretty in the weeds on the particle tracker, but the upside is that it doesn't change much at all of the UI, just shuffles around the internals with some optimization, and the tests still pass. 

I know you're busy right now, so this can wait until that's done.

changelog/2911.trivial.3.rst
namurphy, line 1: 🪧 We recently added an `internal` category for changelog entries for things like refactoring. I'm wondering if this entry would better fit as an `internal` change. 🤔

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -87,7 +87,7 @@ def __init__(
             "detector": (u.m, "attribute"),
             "mag": (u.m, "attribute"),
             "max_deflection": (None, "attribute"),
-            "nparticles": (None, "attribute"),
+            "num_particles": (None, "attribute"),
```
namurphy, line 90: Would this count as a breaking change?

pheuer: Maybe? It's technically user-facing, although I at least never access it. I'll add a changelog for it.

changelog/2911.trivial.3.rst
pheuer: Oooh yes that's much better, since it definitely wasn't trivial :)

pheuer:
Two new doc errors. Not sure why this reference is a problem...
```
nox > uv pip install -r ci_requirements/docs-3.13-linux.txt '.[docs]'
nox > sphinx-build docs/ docs/build/html --nitpicky --fail-on-warning --keep-going -q --builder html
<unknown>:37: WARNING: py:obj reference target not found: astropy.unit.Quantity
<unknown>:37: WARNING: py:obj reference target not found: astropy.unit.Quantity
Loading template...
Finding news fragments...
nox > Command sphinx-build docs/ docs/build/html --nitpicky --fail-on-warning --keep-going -q --builder html failed with exit code 1
nox > Session docs failed.
```


And this is also weird, since this is definitely still a valid argument to `np.reshape` on their dev and stable branches? 

https://numpy.org/doc/stable/reference/generated/numpy.reshape.html

```
=========================== short test summary info ===========================
FAILED tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py::test_NIST_particle_stopping[ALUMINUM-density0-energy_projected_range_list0] - TypeError: reshape() got an unexpected keyword argument 'shape'
FAILED tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py::test_NIST_particle_stopping[SILICON-density1-energy_projected_range_list1] - TypeError: reshape() got an unexpected keyword argument 'shape'
= 2 failed, 4382 passed, 4 skipped, 6 xfailed, 1 xpassed, 80 warnings in 518.50s (0:08:38) =
nox > Command pytest --pyargs --durations=5 --tb=short -n=auto --dist=loadfile failed with exit code 1
nox > Session tests-3.11(lowest-direct) failed.
```

changelog/2911.breaking.2.rst
```diff
@@ -0,0 +1,4 @@
+The method `~plasmapy.simulation.particle_tracker.particle_tracker.ParticleTracker.add_stopping`
+argument ``I`` must now be a list of `~astropy.unit.Quantity` objects rather than a single
+`~astropy.unit.Quantity` array. This is because ``None`` entries are now also accepted to that
+keyword as placeholders for grids on which stopping will not be applied.
```
namurphy, line 4:
```suggestion
argument ``I`` must now be a list of `~astropy.units.Quantity` objects rather than a single
`~astropy.units.Quantity` array. This is because ``None`` entries are now also accepted to that
keyword as placeholders for grids on which stopping will not be applied.
```

pheuer: *facepalm*

pheuer:
The error in the tests is probably coming from the fact that the test uses the `np.reshape(shape = ?)` keyword argument, but that keyword argument was only added in numpy 2.1. In the older versions of numpy, shape could only be a positional argument, and maybe on the 3.11 CI tests environment, one of the earlier numpy versions is installed? The better question is, why wasn't this a problem before...

pheuer:
@PlasmaPy/plasmapy-reviewers Can we get this merged in the next few weeks? I've left comments trying to explain what was done - please tag me with any further questions.

tests/plasma/test_grids.py
```diff
@@ -505,6 +505,12 @@ def uniform_cartesian_grid():
     return grid
 
 
+def test_soften_edges(uniform_cartesian_grid):
+    grid = uniform_cartesian_grid
+    grid.soften_edges()
+    assert grid["x"][0, 0, 0] == 0
```
namurphy, line 511:
```suggestion
    u.assert_quantity_allclose(grid["rho"][0, :, :], 0)
    u.assert_quantity_allclose(grid["rho"][:, 0, :], 0)
    u.assert_quantity_allclose(grid["rho"][:, :, 0], 0)
    u.assert_quantity_allclose(grid["rho"][-1, :, :], 0)
    u.assert_quantity_allclose(grid["rho"][:, -1, :], 0)
    u.assert_quantity_allclose(grid["rho"][:, :, -1], 0)
```

🧪 I'm wondering if we could use [`astropy.tests.helper.assert_quantity_allclose`](https://docs.astropy.org/en/latest/api/astropy.tests.helper.assert_quantity_allclose.html) to help check that all of the grid faces are zeroed out. 

I changed it to `"rho"` instead of `"x"` so that it's checking a physical quantity rather than a coordinate.  I haven't tried this, so the latter arguments might need to be changed to `0 * u.kg * u.m**-3` if `grid["rho"]` is a `Quantity`.

We could potentially parameterize this, but I don't know how to do that with the `:` indices. 🤔

tests/simulation/test_particle_tracker.py
namurphy: Reducing the number of warnings that our test suite produces is very welcome! 😅

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -845,8 +843,8 @@ def _coast_to_plane(self, center, hdir, vdir, x=None, mask=None):
 
         # Check that all points are now in the plane
         # (Eq. of a plane is nhat*x + d = 0)
-        plane_eq = np.dot(x - center, normal)
-        assert np.allclose(plane_eq[mask], 0, atol=1e-6)
+        plane_eq = np.dot(x[mask] - center, normal)
+        assert np.allclose(plane_eq, 0, atol=1e-6)
```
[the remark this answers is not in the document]
pheuer: Reasonable - this should be mathematically impossible (only happens if the push to plane broke somehow) so maybe a ValueError.

src/plasmapy/plasma/grids.py
```diff
@@ -1110,6 +1110,51 @@ def vector_intersects(self, p1, p2):
...
+                * (erf((pts - x1) / sigma) + 1)
+                * -0.5
+                * (erf((pts - x2) / sigma) - 1)
+            )
+
+        edge_mask = np.ones(self.shape)
+        edge_mask[0, :, :] = 0
+        edge_mask[-1, :, :] = 0
+        edge_mask[:, 0, :] = 0
+        edge_mask[:, 1, :] = 0
```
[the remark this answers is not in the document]
pheuer: EEK yes!

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -745,46 +788,51 @@ def _remove_particles(self, particles_to_remove_mask) -> None:
...
 
-        # Subdivide the gyroperiod into a provided number of steps
-        # Use the result as the candidate associated with gyration in B field
-        candidates[:, self.num_grids] = gyroperiod / self._steps_per_gyroperiod
+        # Compute the cyclotron gyroperiod for each particle
+        if self._B_on_grids:
+            Bmag = np.linalg.norm(self._B, axis=-1)
+            mask = Bmag != 0
+            gyroperiod = np.full(Bmag.shape, fill_value=np.inf)
+            gyroperiod[mask] = 2 * np.pi * self.m / (np.abs(self.q) * Bmag[mask])
```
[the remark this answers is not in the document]
pheuer: Gyrofrequency is in there, isn't it? Maybe I should be using it...

https://docs.plasmapy.org/en/stable/api/plasmapy.formulary.frequencies.gyrofrequency.html

src/plasmapy/simulation/particle_tracker/particle_tracker.py
```diff
@@ -798,60 +846,22 @@ def _adaptive_dt(self, Ex, Ey, Ez, Bx, By, Bz) -> NDArray[np.float64] | float:
...
-        # Construct a dictionary to store the interpolation results
-        # This is accomplished using a dictionary comprehension. Each quantity is
-        # initialized as a zeros array with its respective units
-        total_grid_values = {
-            field_name: np.zeros(self.nparticles_tracked)
-            * AbstractGrid.recognized_quantities()[field_name].unit
-            for field_name in self._required_quantities
-        }
+        # Zero out the array of results
+        self._total_grid_values = {k: v * 0 for k, v in self._total_grid_values.items()}
```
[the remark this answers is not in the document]
pheuer: Certainly!

src/plasmapy/plasma/grids.py
```diff
@@ -1110,6 +1110,51 @@ def vector_intersects(self, p1, p2):
...
+        ----------
+        width : `~astropy.units.Quantity` (optional)
+            Width of the transition region.
+            Defaults to 10% of the size of the grid.
+
+        """
+        if isinstance(width, u.Quantity):
+            width = [
+                width,
+            ] * 3
```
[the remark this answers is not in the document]
pheuer: Done

pheuer:
@namurphy I'm working on addressing these, but I think one of the recent changes has broken pre-commit on windows. It works for me on old branches (not up to date with main), but on main now I get: 

```
PS C:\Users\pvheu\GitHub\PlasmaPy> pre-commit run --all-files
[INFO] Installing environment for https://github.com/mattlqx/pre-commit-search-and-replace.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('rbenv', 'init', '-')
return code: 1
stdout:
    Executable `bash` not found
stderr: (none)
Check the log at C:\Users\pvheu\.cache\pre-commit\pre-commit.log
```
With the unhelpful log: 

```
### version information

```
pre-commit version: 4.0.1
git --version: git version 2.31.0.windows.1
sys.version:
    3.12.7 (main, Oct 16 2024, 00:21:24) [MSC v.1929 64 bit (AMD64)]
sys.executable: C:\Users\pvheu\.virtualenvs\dev312\Scripts\python.exe
os.name: nt
sys.platform: win32
```

### error information

```
An unexpected error has occurred: CalledProcessError: command: ('rbenv', 'init', '-')
return code: 1
stdout:
    Executable `bash` not found
stderr: (none)
```

```
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\error_handler.py", line 73, in error_handler
    yield
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\main.py", line 417, in main
    return run(args.config, store, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\commands\run.py", line 442, in run
    install_hook_envs(to_install, store)
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\repository.py", line 229, in install_hook_envs
    _hook_install(hook)
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\repository.py", line 85, in _hook_install
    lang.install_environment(
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\languages\ruby.py", line 125, in install_environment
    lang_base.setup_cmd(prefix, ('rbenv', 'init', '-'))
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\lang_base.py", line 86, in setup_cmd
    cmd_output_b(*cmd, cwd=prefix.prefix_dir, **kwargs)
  File "C:\Users\pvheu\.virtualenvs\dev312\Lib\site-packages\pre_commit\util.py", line 111, in cmd_output_b
    raise CalledProcessError(returncode, cmd, stdout_b, stderr_b)
pre_commit.util.CalledProcessError: command: ('rbenv', 'init', '-')
return code: 1
stdout:
    Executable `bash` not found
stderr: (none)
```
```

src/plasmapy/simulation/particle_tracker/particle_tracker.py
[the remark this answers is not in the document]
pheuer: I tried this but there's no lite function yet, so I think I'd prefer to stick with this until someone gets around to adding that for gyrofrequency

pheuer:
pre-commit.ci autofix

pheuer:
Hi @namurphy 

> 1. If I'm understanding this correctly, the `erf` formulation for the mask in `soften_edges` will change all of the values on the grid. At the center of the grid, the original values would be multiplied by a value that is really close to, but slightly less than, one.  Is there a mask we could apply that doesn't affect quantities toward the center of the grid?  Or alternatively, should we add an admonition to the docstring that mentions this?

Yes, I don't see a clean way to fix this and it's a very minor problem since the ERF is very close to one away from the edge, so I've just mentioned it in the docstring.


> 2. If this mask is applied to the magnetic field, I believe this would cause the resulting magnetic field to violate the divergence constraint

Yes, good point, I've added a note to that effect.

pheuer:
Ok, I think all the comments are addressed - my pre-commit is still broken locally, but I was able to run it through the comment bot, so it's passing now.

tests/plasma/test_grids.py
```diff
@@ -505,6 +506,24 @@ def uniform_cartesian_grid():
...
+)
+def test_soften_edges(uniform_cartesian_grid, width):
+    grid = uniform_cartesian_grid
+    grid.soften_edges(width=width)
+    assert_quantity_allclose(grid["rho"][0, :, :], 0 * u.kg / u.m**3)
+    assert_quantity_allclose(grid["rho"][:, 0, :], 0 * u.kg / u.m**3)
+    assert_quantity_allclose(grid["rho"][:, :, 0], 0 * u.kg / u.m**3)
+    assert_quantity_allclose(grid["rho"][-1, :, :], 0 * u.kg / u.m**3)
+    assert_quantity_allclose(grid["rho"][:, -1, :], 0 * u.kg / u.m**3)
+    assert_quantity_allclose(grid["rho"][:, :, -1], 0 * u.kg / u.m**3)
```
namurphy, line 524: Looking back at this, would it be worthwhile to add one more test to make sure that the value at the center of the grid is very close to the original value there?

namurphy (approved):
I have one final optional suggestion for one of the tests which you can consider, and I'd say it's ready to go!  Thank you for doing this!

pheuer:
pre-commit.ci autofix

tests/plasma/test_grids.py
pheuer: Done!

PR #2917, opened 2024-11-23
Remove GitHub-style admonition that does not get recognized on PyPI page

namurphy:
While GitHub has new syntax for admonitions, other sites do not (yet) recognize it.  Hence, I'm removing it here for the `README.md` file. The weird formatting can (or could) be seen on [PlasmaPy's PyPI page](https://pypi.org/project/plasmapy).  

The top-level `README.md` is the primary Markdown file that shows up on different sites, so it should be no problem to use it in other places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,10 +84,9 @@ To install PlasmaPy on macOS or Linux, open a terminal and run:
 python -m pip install plasmapy
 ```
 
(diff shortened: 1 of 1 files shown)


namurphy:
Merging since it's a very minor change.

PR #2937, opened 2025-01-08
Pin requirements with `uv.lock` for use in CI and on Read the Docs

### LINKED ISSUE #2852: Use `uv.lock` file to pin dependencies in tests instead of files in `ci_requirements`
### Description of improvement

We should consider changing the way we pin requirements for tests and documentation to use `uv.lock` files instead of the pinned requirements files in `ci_requirements/`.

### Motivation

Right now, we have a home grown system for pinning the requirements used in tests. We use `nox -s requirements` to update the requirements files, which are then stored as requirements files in `ci_requirements/`.  There are files for different versions of Python, different operating systems, for running tests or docs, or for running tests with the lowest versions of direct dependencies.  

The Nox sessions that run tests and build documentation install those packages via these requirements files.  Pinning the requirements helps us avoid situations where an unrelated PR suddenly stops working because of a change in an upstream dependency (which causes grief to both new and current contributors alike).  We have automated PRs to update the requirements files, which really help us find new problems from changes in upstream dependencies.  The diffs tell us what packages are being updated.

The problems with the current setup are: 
 - It's extra code (and documentation) for us to maintain
 - It's home grown software, so it can cause confusion to people who are new to the project.
 - Few people besides me are familiar with the inner workings of the pinned requirements files.
 - Updates to requirements mean updates to ∼25–30 files which result in big PRs

While I'm really happy with how we pin requirements used in tests, switching to `uv.lock` files could help ameliorate a bunch of these problems and be cleaner and more maintainable in the long term. We'd also offload the need to document a bunch of this stuff.  We'd still want the weekly or fortnightly PRs to do this. Fortunately, `uv.lock` files are supposed to be human readable.

### Implementation strategy

Because uv is growing so rapidly, my inclination is to wait until next year if we decide to do this. 

This also could be part of a larger effort to integrate uv packaging tools, which I also think should wait until 2025 because of how quickly uv is growing.

I don't know how we would handle the optional dependency sets (i.e., `docs` and `tests`) in a `uv.lock` file.  

We'd need to update the installation step for the Nox environments for building docs and running tests.

### LINKED ISSUE #2940: Validate that the requirements in `pyproject.toml` are consistent with `uv.lock`
If #2937 is merged, we'll be pinning requirements for tests, documentation builds, etc. with `uv.lock`.

Before #2937, we had been pinning requirements files in `ci_requirements/`. We also had a [`validate_requirements`](https://github.com/PlasmaPy/PlasmaPy/blob/5bf2abf8be694bf0c468e093a51aa433b63f6845/noxfile.py#L406) session in `noxfile.py`.  That let us do `nox -s validate_requirements` and make sure that the requirements files were consistent with `pyproject.toml`.  This session is being removed in #2937.  

We should re-add a `validate_requirements` session to `noxfile.py` to make sure that the requirements in `pyproject.toml` are consistent with the solution in `uv.lock`.

Running `uv lock` will include the current project in `uv.lock`.  Because PlasmaPy uses dynamic versioning (via `setuptools_scm`) that makes use of the git commit hash, this means that the version of PlasmaPy in `uv.lock` will always be out-of-date after a new commit. There's discussion in https://github.com/astral-sh/uv/issues/7533 related to this issue, so this issue requires an upstream fix. If it ends up being possible to exclude the current project when using `uv.lock` (or alternatively, if we drop the git hash from dynamic versions), then it'll be much more straightforward to verify that `pyproject.toml` and `uv.lock` are consistent.

Mostly, I want to avoid a situation where `pyproject.toml` gets updated and `uv.lock` ends up being inconsistent with it.

namurphy:
## Background

For the last several months, we've been using `uv pip compile` to create separate pinned requirements files in `ci_requirements/` for different versions of Python and for different operating systems. Having to deal with 18 separate files got unwieldy, and the Nox session to  generate these files began to get tricky to maintain.

`uv.lock` holds information about different dependency groups for different operating systems and for different versions of Python. Pinning requirements in `uv.lock` would consolidate all of this information in a single file. 💾 

## Description of changes

This PR switches multiple Nox sessions defined in `noxfile.py` for testing and documentation builds to use requirements that are frozen in `uv.lock`. We can then use `uv sync` to install exact requirements.[^1]

## Context

Updating pinned requirements is still done via `nox -s requirements`, but can also be done with `uv lock --upgrade`. Keeping this as a Nox session ensures that we have a consistent/canonical way to update `uv.lock` (i.e., by making sure we are using the same version of `uv` with the same options) both locally and in CI. With a Nox session, we could add in additional steps too (e.g., logging or exporting a documentation build requirements file).

There are some Nox sessions that will not use `uv.lock`, including the tests that run against the lowest versions of direct dependencies. Consequently, breaking changes in new versions of _indirect_ dependencies could cause spontaneous CI failures.  If this happens more than ∼twice per year, we might be able to create a separate `uv.lock` or requirements file for the lowest direct dependencies.  

This PR does not substantively change the GitHub workflow to update pinned requirements that is run weekly to quarantine new versions of dependencies and make sure that CI for other PRs does not get spontaneous failures due to breaking changes in upstream dependencies.  

Closes #2852. Closes #2940.

## Tasks
 
 - [x] Make the Nox session for requirements generate `uv.lock` instead
 - [x] Install dependencies from `uv.lock` using `uv sync` for the Nox sessions for:
    - [x] Building documentation
    - [x] Running tests
    - [x] Running mypy
 - [x] Update the PR message for the GitHub workflow that regenerate requirements
 - [x] Remove `_get_requirements_filepath` and other unneeded functionality
 - [x] Remove newly obsolete information from contributor guide and README files, if needed
 - [x] Update the Nox session to `validate_requirements`
 - [x] Remove `ci_requirements/` files
 - [x] Update the Read the Docs configuration to install dependencies using `uv`. We could probably adapt [this section of the RTD docs](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv) to use `uv sync` instead of `uv pip install` so that the requirements from `uv.lock` are used.

[^1]:  Whilst `uv.lock` is described as a "universal" lockfile, it is frankly planetary at best. 🪐


.github/workflows/ci.yml
```diff
@@ -67,11 +67,6 @@ jobs:
           python: '3.13'
           nox_session: cff
 
-        - name: Check consistency of pinned & project requirements
-          os: ubuntu-latest
-          python: '3.13'
-          nox_session: validate_requirements
-
```
namurphy, line 74: See #2940 for more about this.

noxfile.py
```diff
@@ -45,123 +43,34 @@
...
-                resolution,
-                *session.posargs,
-                "--python-platform",
-                os_platform,
-            )
-
+    # If it becomes possible to exclude the current project when using
+    # `uv lock`, we should do so here. That would allow us to add a Nox
+    # session to validate the requirements back in.
+    session.run("uv", "lock", "--upgrade", "--no-progress")
```
namurphy, line 73: I'm quite happy that we can reduce this session to two lines of code!

namurphy:
Getting this error again related to sphinx-codeautolink.  I updated the requirements in #2941 to see if this was because of its v0.16.1 release, but the tests were passing there.  Last time, the error happened after using the `docs` requirements for the incorrect version of Python, but there might be something else happening here.  Hm...

```
nox > sphinx-build docs/ docs/build/html --nitpicky --fail-on-warning --keep-going -q --builder html
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/__init__.py", line 46, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/__init__.py", line 182, in create_references
    self.filter_and_resolve(transforms, skipped, doc)
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/__init__.py", line 206, in filter_and_resolve
    key = resolve_location(name, self.inventory)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/resolve.py", line 30, in resolve_location
    call_value(cursor)
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/resolve.py", line 114, in call_value
    cursor.value = get_return_annotation(cursor.value)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/PlasmaPy/PlasmaPy/.nox/docs/lib/python3.12/site-packages/sphinx_codeautolink/extension/resolve.py", line 122, in get_return_annotation
    annotation = get_type_hints(func).get("return")
                 ^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/typing.py", line 2296, in get_type_hints
    raise TypeError('{!r} is not a module, class, method, '
TypeError: <function concatenate at 0x7f1531b16cb0> is not a module, class, method, or function.

Extension error (sphinx_codeautolink.extension):
Handler <bound method SphinxCodeAutoLink.create_references of <sphinx_codeautolink.extension.SphinxCodeAutoLink object at 0x7f15064c58e0>> for event 'env-updated' threw an exception (exception: <function concatenate at 0x7f1531b16cb0> is not a module, class, method, or function.)
```

namurphy:
A source of the problem:
```
>>> import numpy as np
>>> import typing
>>> typing.get_type_hints(np.concatenate)
  File "/run/media/namurphy/d423d80e-c227-4a33-b50f-545b44160ce3/namurphy/.local/share/uv/python/cpython-3.12.6-linux-x86_64-gnu/lib/python3.12/typing.py", line 2295, in get_type_hints
    raise TypeError('{!r} is not a module, class, method, '
TypeError: <function concatenate at 0x7fec740aa270> is not a module, class, method, or function.
```

@rocco8773 pointed out that here that this [try/except block](https://github.com/felix-hilden/sphinx-codeautolink/blame/8b570dc5aeece1699ad184735468d741a67c5706/src/sphinx_codeautolink/extension/resolve.py#L121-L125) should catch when `TypeError` results from `typing.get_type_hints`.   

🐰 🕳️ 🐉

.github/workflows/ci.yml
```diff
@@ -67,11 +67,6 @@ jobs:
           python: '3.13'
           nox_session: cff
```
namurphy, line 69:
```suggestion

        - name: Check consistency of pinned & project requirements
          os: ubuntu-latest
          python: '3.13'
          nox_session: validate_requirements
```

namurphy:
I just purposefully made `pyproject.toml` inconsistent with `uv.lock` in two different ways, and the `validate_requirements` session failed as it should both times.  Yay!

.github/workflows/ci.yml
```diff
@@ -71,7 +71,6 @@ jobs:
           os: ubuntu-latest
           python: '3.13'
           nox_session: validate_requirements
-
     steps:
```
namurphy, line 74:
```suggestion

    steps:
```

docs/conf.py
```diff
@@ -244,6 +244,8 @@
 # sphinx-codeautolink
 
 codeautolink_concat_default = True
+codeautolink_warn_on_failed_resolve = False  # turn on for debugging
+codeautolink_warn_on_missing_inventory = False  # turn on for debugging
```
namurphy, line 248: 🏕️ These are useful flags that can help us figure out problems with `sphinx-codeautolink`. The default is `False` for these, but I'm putting these here for convenience. Later on, we could set these to `True` and fix the broken autolinks.

.github/content/update-requirements-pr-body.md
```diff
@@ -1,18 +1,17 @@
...
 [`noxfile.py`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/noxfile.py
 [`pyproject.toml`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/pyproject.toml
 [Nox]: https://nox.thea.codes/en/stable/
 [uv]: https://github.com/astral-sh/uv
 
-This pull request (PR) regenerates the requirements files in [`ci_requirements/`] that are used for running tests, building documentation, and performing other continuous integration (CI) checks. Pinning requirements files reduces the probability that a CI check will spontaneously start failing in a PR due to a breaking change in a dependency. This PR lets us find unexpected breaking changes in newly released dependencies before they get used in CI. 🛡
+This pull request (PR) regenerates the requirements files in [`uv.lock`] that are used for running tests, building documentation, and performing continuous integration (CI) checks. Pinning requirements files reduces the probability that a CI check will spontaneously start failing in a PR due to a breaking change in a dependency. This PR lets us quarantine breaking changes in newly released dependencies before they get used in CI. 🛡
```
namurphy, line 9:
```suggestion
This pull request (PR) regenerates the requirements files in [`uv.lock`] that are used for running tests, building documentation, and performing continuous integration (CI) checks. Pinning requirements files greatly reduces the probability that a CI check will spontaneously start failing in a PR due to a breaking change in a dependency. This PR lets us quarantine breaking changes in newly released dependencies before they get used in CI. 🛡
```

.readthedocs.yml
```diff
@@ -13,14 +13,24 @@ build:
...
+    - asdf install uv latest
+    - asdf global uv latest
+    - uv venv --python=3.12
+
+    install:
+    - uv sync --extra docs --frozen --exact
+
+    build:
+      html:
+      - uv run --extra docs --frozen --exact nox -s docs
```
namurphy, line 33: Using `--exact` ensures that the documentation builds are as consistent as possible across platforms, with the additional benefit that we'll likely be able to find problems with `uv.lock` sooner rather than later. The disadvantage is that the RTD builds _might_ become more finicky.

noxfile.py
```diff
@@ -45,122 +43,62 @@
...
-        ("tests", minpython, "lowest-direct"),
-        ("docs", maxpython, "highest"),
-        ("all", maxpython, "highest"),
-    ]
-
-    category_flags: dict[str, tuple[str, ...]] = {
-        "all": ("--all-extras",),
-        "docs": ("--extra", "docs"),
-        "tests": ("--extra", "tests"),
-    }
```
namurphy, line 131: Dependency information across operating systems, versions of Python, and dependency sets is handled internally by `uv`, so these sections of the code are no longer needed.

noxfile.py
```diff
@@ -261,23 +204,40 @@ def run_tests_with_dev_version_of(session: nox.Session, repository: str) -> None
...
 
-sphinx_commands: tuple[str, ...] = (
+doc_build_dir = "$READTHEDOCS_OUTPUT" if running_on_rtd else "docs/build/html"
+
+if running_on_rtd:
+    rtd_output_path = pathlib.Path(os.environ.get("READTHEDOCS_OUTPUT")) / "html"
+    rtd_output_path.mkdir(parents=True, exist_ok=True)
+    doc_build_dir = str(rtd_output_path)
+else:
+    doc_build_dir = "docs/build/html"
```
namurphy, line 221: I had a bit of trouble making sure that the documentation build was put into the expected place for RTD, but the [RTD docs](https://docs.readthedocs.com/platform/stable/reference/environment-variables.html#envvar-READTHEDOCS_OUTPUT) helped.

noxfile.py
```diff
@@ -261,23 +204,40 @@ def run_tests_with_dev_version_of(session: nox.Session, repository: str) -> None
...
-    "--fail-on-warning",
     "--keep-going",
-    "-q",
-)
+]
+
+if not running_on_rtd:
+    sphinx_base_command.extend(
+        [
+            "--fail-on-warning",
```
namurphy, line 234: Not using `--fail-on-warning` for RTD builds makes it so that the doc build will be easier to preview from the link at the bottom of a pull request page.  The warnings will still cause failures on the CI documentation build.

noxfile.py
```diff
@@ -261,23 +204,40 @@ def run_tests_with_dev_version_of(session: nox.Session, repository: str) -> None
...
+            "--quiet",
```
namurphy, line 235: Usually, the documentation output is much easier to parse with the `--quiet` option. However, doing the RTD build without the `--quiet` option will let us see a more complete log of the documentation build without having to do a repeat run.  Might be helpful to add this tip to the documentation guide.

namurphy:
@sapols, might you be interested in reviewing this PR to check out some fun infrastructure improvements?  

The purpose of this PR is to pin the dependencies used in PlasmaPy's CI and doc builds with `uv lock`. The `uv.lock` files is similar to the `requirements.txt` like files we have been using, but `uv.lock` stores requirements information for multiple versions of Python, multiple operating systems, and multiple dependency sets in a single file.  Every Monday, an automated PR updates `uv.lock` to make sure that the tests still pass with new versions of the dependencies. This setup may be helpful for other packages in PyHC since changes to `uv.lock` help identify which upstream packages that may be responsible for newly failing tests.

docs/contributing/coding_guide.rst
```diff
@@ -658,16 +658,17 @@ Dependencies and requirements
...
     their initial release.
 
-* The |ci_requirements/|_ directory contains pinned requirements files
-  for use in continuous integration tests (see
-  |ci_requirements/README.md|_).
+* The |uv.lock|_ file contains pinned requirements files
+  for use in continuous integration tests.
 
   - These files are updated periodically via pull requests created by a
     GitHub workflow to `update pinned requirements`_.
```
rocco8773, line 665: Note:  This link currently takes you to the GitHub Action page (https://github.com/PlasmaPy/PlasmaPy/actions/workflows/update-pinned-reqs.yml) instead of the actual workflow script (https://github.com/PlasmaPy/PlasmaPy/blob/main/.github/workflows/update-pinned-reqs.yml).

[Opinion] To me it makes more sense to link to the script and not the Action.  In any case this will not affect this PR.

pyproject.toml
```diff
@@ -86,13 +86,13 @@ docs = [
   "docutils >= 0.20.1",
   "jinja2 >= 3.1.4",
   "nbsphinx >= 0.9.4",
-  "nox >= 2024.4.15",
+  "nox >= 2024.10.9",
   "numpydoc >= 1.7.0",
   "pillow >= 10.4.0",
-  "pygments >= 2.18.0",
+  "pygments == 2.18.0",
```
rocco8773, line 92: You made a comment in the the `.github/content/update-requirements-pr-bocy.md` that a PR should be opened to remove an upper bound on a dependency if one is add.  **Should this also be true if we freeze `==` a dependency?**

rocco8773: Or maybe we only do this for install requirements (i.e. not tests, docs, etc.).

noxfile.py
```diff
@@ -45,122 +43,62 @@
 maxpython = max(supported_python_versions)
 minpython = min(supported_python_versions)
```
rocco8773, line 44: Note: The `supported_python_versions` can easily get out of date with `pyproject.toml`.  Are there breadcrumbs to ensure this gets updated when `pyproject.toml` gets updated?  Or, is there a way to scrape `pyproject.toml` for this info?

noxfile.py
```diff
@@ -45,122 +43,62 @@
...
-    ----------
-    category : str
-        The name of the optional dependency set, as defined in
-        :file:`pyproject.toml`.
+running_on_ci = os.getenv("CI")
+running_on_rtd = os.environ.get("READTHEDOCS") == "True"
 
-    version : str
-        The supported version of Python.
+uv_requirement = "uv >= 0.6.1"
```
rocco8773, line 63: `uv` is not listed as a package dependency, should it be?  If `uv` is included as a dependency, then this line can get out of sync easily.

noxfile.py
rocco8773, line 221: Whenever we build the docs the build directory is `docs/_build` (with an underscore), not `docs/build`.  That's how the make files are constructed.

This build directory is a bit arbitrary though.  However, this would make the build directory different if you built with `make` versus `nox`.

rocco8773: Also, this block can be streamlined more...

```suggestion
doc_build_dir = "docs/build/html"
if running_on_rtd:
    rtd_output_path = pathlib.Path(os.environ.get("READTHEDOCS_OUTPUT")) / "html"
    rtd_output_path.mkdir(parents=True, exist_ok=True)
    doc_build_dir = str(rtd_output_path)
```

rocco8773 (approved):
I left a few minor comments that highlight some potential (unrelated 🤷‍♂️) issues, but nothing I think really holds up this PR.

With my limited knowledge of how `uv` and `nox` works, this looks good to me.  There might be some subtleties I'm missing, but if it works it works!

noxfile.py
namurphy: I've been thinking the same thing!  I did see an example `noxfile.py` somewhere that scraped `pyproject.toml` to get the supported versions of Python, but haven't had it be a high priority since I've been pretty thorough when dropping & adding versions of Python.  I raised #2954 so we don't forget.

noxfile.py
namurphy: I have been wondering the same thing!  I've seen it recommended that `uv` be installed completely independently of Python, which makes sense since `uv` can be used to install and manage multiple versions of Python at once. 

`uv` needs to get installed first because it's what is used to install the dependencies in the tests. I put the `uv` requirement in here explicitly as a safeguard so that running `nox -s docs` will work even if `uv` hasn't been installed separately. 

I'm going to need to take a bit more time to think about it! 😅

pyproject.toml
namurphy: > **Should this also be true if we freeze `==` a dependency?**

Yes! I'll need to update that.

 - [ ] address this ⬆️

namurphy:
@rocco8773 — Thank you for the very helpful review!

namurphy:
Some exciting news for the pythoniverse is that [PEP 751](https://peps.python.org/pep-0751/) was approved yesterday, which defines a standard for `pylock.toml` lock files.  They will be kind of like a next gen `requirements.txt` file.  

I'm inclined to stick with `uv.lock` as PlasmaPy's primary lockfile since:
 - `uv.lock` is more fully featured than `pylock.toml`
 - `uv` is being very actively developed and has a lot of really helpful tooling built around `uv.lock`
 - PlasmaPy uses the lockfile only internally for running tests and building docs, so the lockfile isn't used by end users.

It may be helpful to export `pylock.toml` from `uv.lock`, since that might be useful to have in PlasmaPy Zenodo uploads.

PR #2959, opened 2025-03-04
Use `setup-uv` in GitHub workflows

namurphy:
For approximately eight hundred billion years, we have been using `actions/setup-python` to install Python in our GitHub workflows. Unlike `pip`, `uv` is able to install and manage Python installations. `uv` also has significantly faster dependency resolution.

This PR switches our GitHub workflows so that they use `astral-sh/setup-uv` instead of `actions/setup-python`, so that `uv run` can be used to run Nox or a script.  This PR also makes use of built-in caching in `setup-uv` (similar to caching in `setup-python`).  

The installation of Python environments are done within the sessions defined in `noxfile.py` for consistency between running Nox sessions both via GitHub Actions and on our local computers.

See also: #2585

> [!NOTE]
> We had previously set `fetch-depth: 0` in `actions/checkout` with the _intention_ of saving time by only downloading the most recent commit for `main`.  However, setting `fetch-depth: 0` will _download the entire history of the repository with all branches_.  Oops!  The default is to download only the most recent commit (i.e., `fetch-depth: 1`), so it does not need to be set at all.


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions: {}
```
namurphy, line 18: Including this as a security improvement suggested by `zizmor`.  This will likely conflict with #2975.

.github/workflows/ci.yml
```diff
@@ -77,40 +79,28 @@ jobs:
...
       uses: actions/checkout@v4
       with:
-        fetch-depth: 0
-
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: ${{ matrix.python }}
-        cache: pip
+        persist-credentials: false
```
namurphy, line 82: Another security improvement suggested by `zizmor`.

.github/workflows/ci.yml
```diff
@@ -77,40 +79,28 @@ jobs:
...
       uses: ts-graphviz/setup-graphviz@v2
 
-    - name: Install pandoc for doc builds (requires Ubuntu)
-      if: startsWith(matrix.name, 'Documentation')
+    - name: Install pandoc for doc builds
+      # This step requires Ubuntu
+      if: startsWith(matrix.nox_session, 'docs')
       run: |
         sudo apt update
         sudo apt-get install -y pandoc
```
namurphy, line 93: It would have been nice to make this workflow independent of OS, but I haven't found a reliable cross-platform way of installing `pandoc`.

.github/scripts/authors_in_cff.py
```diff
@@ -1,5 +1,10 @@
 """Check that the authors of a PR are included in ``CITATION.cff``."""
 
+# /// script
+# requires-python = ">=3.13"
+# dependencies = ["requests>=2.32"]
+# ///
+
```
namurphy, line 7: This follows the new standard for [inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata).

.github/workflows/check-author-included.yml
```diff
@@ -12,19 +14,19 @@ jobs:
...
-        python-version: 3.13
+        persist-credentials: false
 
-    - name: Install dependencies
-      run: pip install requests
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+      with:
+        enable-cache: true
+        cache-suffix: author-cff
```
namurphy, line 24: Caching for this workflow is probably worthwhile since this is run in CI.

.github/workflows/linkcheck.yml
```diff
@@ -27,18 +29,22 @@ jobs:
...
-      run: python -m pip install --upgrade nox uv
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
 
     - name: Install graphviz & pandoc
       run: sudo apt-get install graphviz pandoc
 
     - name: Check hyperlinks
-      run: nox -s linkcheck
+      run: uv run --python 3.12 --with nox --frozen nox -s linkcheck
```
namurphy, line 50: The `--frozen` is so that `uv.lock` doesn't get updated during `uv run`.  Without it, `uv.lock` would get updated and the runs would no longer be reproducible.

.github/workflows/linkcheck.yml
```diff
@@ -27,18 +29,22 @@ jobs:
...
+    - name: Install pandoc
+      # This step requires Ubuntu
+      run: |
+        sudo apt update
+        sudo apt-get install -y pandoc
 
-    - name: Install Python dependencies
-      run: python -m pip install --upgrade nox uv
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
```
namurphy, line 44: I didn't enable caching for this step because the workflow is run pretty infrequently (weekly and before releases to check that links in the documentation still work).

namurphy:
@rocco8773 — would you have a chance to review this by ∼Thursday?

.github/workflows/check-author-included.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+permissions: {}
```
rocco8773, line 7:
```suggestion
permissions: {}  # disables all GitHub permissions for the workflow
```

I only suggest this since it is not obvious what `permissions: {}` does, and it's buried in the [GitHub documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions).

.github/workflows/ci.yml
```diff
@@ -77,40 +79,28 @@ jobs:
...
-        key: ${{ matrix.nox_session }}-${{ matrix.python }}-${{ runner.os }}-${{ hashFiles('pyproject.toml', 'uv.lock') }}
-
-    - name: Run Nox session
-      run: nox -s '${{ matrix.nox_session }}'
+        python-version: ${{ matrix.python }}
+        enable-cache: true
+        cache-suffix: ${{ matrix.nox_session }}-${{ matrix.python }}-${{ runner.os }}
+
+    - name: Run the check
+      run: uv run --python ${{ matrix.python }} --with nox --frozen nox -s '${{ matrix.nox_session }}'
```
rocco8773, line 103: Do you need `--python ${{ matrix.python }}` since `setup-uv` already declares the python version to use?

.github/workflows/linkcheck.yml
```diff
@@ -12,6 +12,8 @@ on:
     branches:
     - main
 
+permissions: {}
```
rocco8773, line 15:
```suggestion
permissions: {}  # disables all GitHub permissions for the workflow
```
Same as [comment](https://github.com/PlasmaPy/PlasmaPy/pull/2959/files#r2055075406)

.github/workflows/linkcheck.yml
rocco8773, line 44:
```suggestion
      uses: astral-sh/setup-uv@v5
      with:
        python-version: 3.12
```

Then I do not think you need to use `--python 3.12` in the `uv` command on line 50.

.github/workflows/update-pinned-reqs.yml
```diff
@@ -16,18 +16,13 @@ jobs:
...
 
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.13
-
-    - name: Install dependencies
-      run: python -m pip install --upgrade nox uv
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
```
rocco8773, line 22:
```suggestion
      uses: astral-sh/setup-uv@v5
      with:
        python-version: 3.13
```

Then I do not think you need to use `--python 3.13` in the `uv` command on line 25.

.github/workflows/weekly.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
     - main
 
+permissions: {}
```
rocco8773, line 12:
```suggestion
permissions: {}  # disables all GitHub permissions for the workflow
```
Same as [comment](https://github.com/PlasmaPy/PlasmaPy/pull/2959/files#r2055075406)

.github/workflows/check-author-included.yml
namurphy: That's a good point!  When I first ran `zizmor` and found this potential vulnerability, I needed to look it up to understand what was going on.   I'll include this comment in `ci.yml` since this is the most frequently read & revised workflow file.

.github/workflows/ci.yml
namurphy: Looking at the docs for [`setup-uv`](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#python-version), it appears that we don't need to use `--python` if we already declare it using `with:` in `setup-uv`. I'll revise this.  Thanks!

namurphy:
The documentation failure appears to be unrelated. It was a warning caused by not being able to find the matplotlib intersphinx inventory file.

PR #2964, opened 2025-03-10
Print out the differences between environments when updating requirements

namurphy:
In #2937, we switched to using `uv.lock` to pin dependencies for use in CI.  We had previously been using `requirements.txt` style files, which had the advantage that the diffs were more easily human readable. The purpose of this PR is to make it so that the Nox session that updates requirements for us also prints out the differences between the old and new environments.

### DIFF
--- a/changelog/2964.internal.rst
+++ b/changelog/2964.internal.rst
@@ -1,0 +1,2 @@
+Updated the ``requirements`` session in :file:`noxfile.py` to print
+out the differences between the new and old environments.


.github/workflows/update-pinned-reqs.yml
```diff
@@ -46,11 +46,12 @@ jobs:
...
-        commit-message: Update requirements with 'nox -s requirements'
         title: Update pinned requirements 📍
+        add-paths: uv.lock
+        commit-message: Update requirements with 'nox -s requirements'
+        # The PR body message is generated dynamically via `nox -s requirements` in CI
         body-path: .github/content/update-requirements-pr-body.md
         labels: no changelog entry needed, requirements
-        delete-branch: true
         base: main
-        assignees: namurphy
```
namurphy, line 56: I removed this line since `CODEOWNERS` is the appropriate place for this.

.github/workflows/update-pinned-reqs.yml
```diff
@@ -46,11 +46,12 @@ jobs:
     - name: Create pull request
       uses: peter-evans/create-pull-request@v7
       with:
-        token: ${{ steps.generate_token.outputs.token }}
-        commit-message: Update requirements with 'nox -s requirements'
         title: Update pinned requirements 📍
+        add-paths: uv.lock
+        commit-message: Update requirements with 'nox -s requirements'
+        # The PR body message is generated dynamically via `nox -s requirements` in CI
```
namurphy, line 52: This comment is the only new line.  Most of the other lines are just moved around.

noxfile.py
```diff
@@ -574,5 +619,9 @@ def lint(session: nox.Session) -> None:
     )
 
 
+# /// script
+# dependencies = ["nox"]
+# ///
+
```
namurphy, line 625: Turns out that this does not need to be included at the top of the file.

namurphy:
After a quick follow-up in #2964 to print out lines that were not included in the table, I got it to work in #2966. This should reduce the activation energy to get new CI failures fixed.

PR #2965, opened 2025-03-10
Improve handling of unexpected lines of `uv lock` output in Nox session to update requirements

namurphy:
This PR is a quick follow-up to #2964, since it failed when I tried to run the requirements workflow.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -64,7 +64,7 @@ running_on_rtd = os.environ.get("READTHEDOCS") == "True"
 uv_requirement = "uv >= 0.6.5"
 
 
-def _create_requirements_pr_message(uv_output: str) -> None:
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -92,18 +92,24 @@ def _create_requirements_pr_message(uv_output: str) -> None:
     ]
 
     for package_update in uv_output.splitlines():
-        if package_update.startswith("Resolved"):
+        if not package_update.startswith("Updated"):
+            session.debug(f"Line not added to table: {package_update}")
```
namurphy, line 96: There might be some standard output from `uv lock --upgrade` that this function is currently not handling, so it'll be helpful to print it out as a debug statement.

PR #2968, opened 2025-03-13
Add optional keywords to SyntheticRadiograph to directly control detector orientation & source particle orientation

pheuer:
The SyntheticRadiography module tracks particles (usually randomly generated by the `create_particles` method) through a grid onto a detector plane (defined by two perpendicular vectors and a center point). The initial particle velocities are distributed around the source-detector axis.

 In the current version, these vectors are set up automatically in a reasonable way (e.g. assuming the detector is looking at the origin, particles go along the source-detector axis). However, experimental setups are sometimes non-ideal, and in those cases the user might want direct control over the orientation of the detector and the initial velocity vector. 

This PR just adds a couple keywords to explicitly set the detector_vdir (a keyword already existed for detector_hdir) and the vector around which the velocities of particles in `create_particles` are centered. Both keywords are optional and default to the previous behavior, so this a non-breaking feature.

Also, corrected a bug in the docs where the default detector_hdir was incorrectly described as (source-detector X Z-axis) when in reality it was (origin-detector X Z-axis).

### DIFF
--- a/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -175,10 +175,10 @@ class Tracker(ParticleTracker):
         A unit vector (in Cartesian coordinates) defining the horizontal
         direction on the detector plane. By default, the horizontal axis in the
         detector plane is defined to be perpendicular to both the
-        origin-to-detector vector  (such that the detector is 'looking at' the origin) 
-        and the z-axis (unless the origin-to-detector axis is parallel to the z axis, 
+        origin-to-detector vector  (such that the detector is 'looking at' the origin)
+        and the z-axis (unless the origin-to-detector axis is parallel to the z axis,
         in which case the horizontal axis is the x-axis).
-        
+
     detector_vdir : `numpy.ndarray`, shape (3), optional
         A unit vector (in Cartesian coordinates) defining the vertical
         direction on the detector plane. By default, the vertical axis in the
@@ -669,11 +669,11 @@ class Tracker(ParticleTracker):
             Simulations run in the ``'uniform'`` mode will imprint a grid pattern
             on the image, but will well-sample the field grid with a
             smaller number of particles. The default is ``'monte-carlo'``.
-            
+
         source_vdir : `numpy.ndarray`, shape (3), optional
-            A unit vector (in Cartesian coordinates) defining the direction
-            of the particle velocities. By default, the particle velocities
-            will be distributed around the source-detector axis.
+            A unit vector (in Cartesian coordinates) defining the orientation
+            of the mean of the particle velocities.  By default, the particle
+            velocities will be distributed around the source-detector axis.
 
         random_seed : int, optional
             A random seed to be used when generating random particle
@@ -683,7 +683,7 @@ class Tracker(ParticleTracker):
 
         # Raise an error if the run method has already been called.
         self._enforce_order()
-        
+
         if source_vdir is None:
             source_vdir = self.src_det / np.linalg.norm(self.src_det)
 
--- a/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -670,7 +670,7 @@ class Tracker(ParticleTracker):
             on the image, but will well-sample the field grid with a
             smaller number of particles. The default is ``'monte-carlo'``.
 
-        source_vdir : `numpy.ndarray`, shape (3), optional
+        source_vdir : (3,) |array_like|, default: None
             A unit vector (in Cartesian coordinates) defining the orientation
             of the mean of the particle velocities.  By default, the particle
             velocities will be distributed around the source-detector axis.
--- a/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
+++ b/src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
@@ -175,13 +175,17 @@ class Tracker(ParticleTracker):
         A unit vector (in Cartesian coordinates) defining the horizontal
         direction on the detector plane. By default, the horizontal axis in the
         detector plane is defined to be perpendicular to both the
-        source-to-detector vector and the z-axis (unless the source-to-detector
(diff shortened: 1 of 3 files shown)


pheuer:
@namurphy I'm sure you probably dread seeing a new CPR PR pop up, but this one should be quick and easy to review! :)

tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
```diff
@@ -433,6 +436,17 @@ def test_create_particles() -> None:
...
 
+    # Test specifying direction
+    src_vdir = np.array([0.1, 1, 0])
+    src_vdir /= np.linalg.norm(src_vdir)
+    sim.create_particles(
+        1e3, 15 * u.MeV, particle="p+", random_seed=42, source_vdir=src_vdir
+    )
+    # Assert particle velocities are actually in that direction
+    vdir = np.mean(sim.v, axis=0)
+    vdir /= np.linalg.norm(vdir)
```
namurphy, line 447:
```suggestion
    vdir = np.mean(sim.v, axis=0) / np.linalg.norm(vdir)
```
Would it make sense to combine these lines?

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -211,6 +215,7 @@ def __init__(
             "volume averaged", "nearest neighbor"
         ] = "volume averaged",
         detector_hdir=None,
+        detector_vdir=None,
```
namurphy, line 218: Would it make sense to have it be something like `(0, 0, 1)` for `detector_hdir` and `detector_vdir` rather than `None`? Doing that would make it easier for users to see what kind of input is expected and what the vertical & horizontal directions are. It could also simplify some code below.

It'd be best as a `tuple` since mutable default arguments like lists are fraught with peril.

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -293,8 +298,11 @@ def __init__(
             self.det_hdir = self._default_detector_hdir()
 
         # Calculate the detector vdir
-        ny = np.cross(self.det_hdir, self.det_n)
-        self.det_vdir = -ny / np.linalg.norm(ny)
+        if detector_vdir is not None:
+            self.det_vdir = detector_vdir / np.linalg.norm(detector_vdir)
+        else:
+            ny = np.cross(self.det_hdir, self.det_n)
+            self.det_vdir = -ny / np.linalg.norm(ny)
```
namurphy, line 305: If `detector_vdir` has a default argument, then we might be able to simplify this into a single line.

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
```diff
@@ -661,6 +670,11 @@ def create_particles(
             on the image, but will well-sample the field grid with a
             smaller number of particles. The default is ``'monte-carlo'``.
 
+        source_vdir : `numpy.ndarray`, shape (3), optional
```
namurphy, line 673:
```suggestion
        source_vdir : (3,) |array_like|, default: (0, 0, 1)  # please correct default!
```
I've been trying to be consistent in the way that we describe [type specifications](https://docs.plasmapy.org/en/stable/contributing/doc_guide.html#docstring-guidelines) in docstrings.  Alas, the numpydoc standard hasn't defined this yet (see https://github.com/numpy/numpydoc/pull/423).

I also didn't double check that this would be the right default for this.

namurphy (commented):
Thank you for doing this!  I have a few suggestions but it looks pretty close to ready to merge.

tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
pheuer: Sure!

pheuer: I don't think it will work, because vdir in norm is then not defined yet?

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Sure, I can change to |array_like|. The default is not constant (defaults to the source-detector axis),  so not sure what to write for that? I suppose literally the default is `None`.

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: I don't think so for the same reason below: `ny` needs to be calculated on line 304 before you can do the normalization on line 305? If there was a `np.linalg.normalize` function, that would work, but I don't think there is.

namurphy: Ah, makes sense!  I remember how math works! 🙃

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
namurphy: In that case, `None` will work best!

namurphy (approved):
Thank you for the clarifications!  Happy to approve this.  

To approximately quote Grand Moff Tarkin: "You may [merge] when ready."

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: The problem is that a constant default value would often give nonsensical results. So, when these are set to None, the code below constructs a reasonable guess based on the provided source and detector points (which are required args)

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
pheuer: Sure, done!

src/plasmapy/diagnostics/charged_particle_radiography/synthetic_radiography.py
namurphy: Thanks for the clarification! The way it is now is the better way to do it.

pheuer:
Thanks!

PR #2975, opened 2025-03-25
Automate GitHub workflow security audits with `zizmor`

namurphy:
[`zizmor`](https://woodruffw.github.io/zizmor/) is a static analysis tool to find common security issues in typical GitHub workflows. This PR:

 - Adds a Nox session to run `zizmor`
 - Fixes some security issues found by `zizmor` (like changing `pull_request_target` → `pull_request`, setting `persist-credentials: false`, and setting strict default permissions)
 - Sets up `.github/zizmor.yml` to ignore `zizmor` warnings and errors that are more difficult to fix.

The changes here are probably going to result in a bunch of conflicts in #2959, but it makes sense to merge this first.

Many thanks to the pyOpenSci community for helping me learn about `zizmor` in the first place!

> [!NOTE]
> Several `zizmor` checks require a GitHub token, so running the `zizmor` Nox session locally will skip certain checks.


.github/zizmor.yml
```diff
@@ -0,0 +1,41 @@
...
+      - unlabel-pr-after-merge.yml:8:3
+      - update-pinned-reqs.yml:9:3
+  template-injection:
+    ignore:
+      - create-release-issue.yml:21:7
+      - mint-release.yml:41:7
+      - mint-release.yml:47:7
+      - mint-release.yml:53:7
+      - mint-release.yml:61:7
+      - mint-release.yml:66:7
```
namurphy, line 38: These errors will probably be more difficult to fix than the others.

namurphy:
@sapols — might you be interested in reviewing this PR?  I'm wondering if it would make sense to add a PyHC standard to perform automated security analyses in CI to catch common vulnerabilities. Tools like `zizmor` are really helpful at finding potential security issues that most of us in PyHC wouldn't have realized.  The [`flake8-bandit`](https://docs.astral.sh/ruff/rules/#flake8-bandit-s) rule set in `ruff` can also help catch a lot of potential issues.

.github/workflows/changelog.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: add-no-changelog-label
     steps:
-    - uses: scientific-python/action-towncrier-changelog@v1
+    - uses: scientific-python/action-towncrier-changelog@v1 # zizmor: ignore[unpinned-uses]
```
namurphy, line 66: The `unpinned-uses` rule checks that actions used via `uses` are pinned to a specific commit rather than the version.  We can ignore this check for actions that come from trusted publishers.

.github/workflows/update-pinned-reqs.yml
```diff
@@ -34,14 +34,14 @@ jobs:
     # action.
 
     - name: Generate token
-      uses: tibdex/github-app-token@v2
+      uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```
namurphy, line 37: It seems reasonable to pin actions by the commit hash (for extra security) if they are released by individuals or small organizations.

namurphy:
@rocco8773 — would you have a chance to review this in the next few days?  With the changes merged and conflicts resolved from #2959, I don't think there should be anything controversion in this PR.  Kind thanks!

rocco8773 (approved):
Things looks good to me.

The only thing I go umm about is the use of `# zimor: ignore[artipacked]`.  I'm not concerned about it being used, but more concerned a contributor might not know it exists and when to use it.  I can just see myself updating a workflow in the future that causes a `zizmor` check to fail and I just forget that this inline comment exists.

I do not think adding a note about this to the documentation would help.  The only thing I can think of doing is having a PR comment be made when `zizmor` check fails.

I do not think this needs to be resolved in the PR, but it is something to consider for a follow up PR.

namurphy:
> The only thing I go umm about is the use of `# zimor: ignore[artipacked]`. I'm not concerned about it being used, but more concerned a contributor might not know it exists and when to use it. 

> The only thing I can think of doing is having a PR comment be made when `zizmor` check fails.

That's an important point. ✅ It wouldn't be hard to have the Nox session for zizmor print out a troubleshooting message.  🤔

> I do not think this needs to be resolved in the PR, but it is something to consider for a follow up PR.

If I don't fix this now, I probably won't remember to later... 😅

 - [ ] Do this ↑

Thanks for the helpful review!

PR #2980, opened 2025-04-04
Create an optional dependency set for the plasma calculator

### LINKED ISSUE #2979: Remove `ipykernel`, `ipywidgets` and `voila` as required dependencies
Currently, `ipykernel`, `ipywidgets`, and `voila` are hard dependencies of all of `plasmapy`. In particular, both `ipykernel` and `ipywidgets` have hard dependencies on `jupyter`, `notebook`, and `jupyterlab`. This means that any package which uses any part of plasmapy is required to install the whole Jupyter lab/notebook stack. 

Given that (as far as I can tell) `ipywidgets` and `voila` are only used explicitly in `utils.calculator`, I'd like to propose adding an optional dependency category for `calculator` and move `ipykernel`, `ipywidgets` and `voila` there.

To be transparent, this suggestion is being motivated by my work on [fiasco](https://github.com/wtbarnes/fiasco) where I'm only using a very small part of plasmapy (mostly `plasmapy.particles`).

namurphy:
This PR creates the `calculator` optional dependency set, as suggested by @wtbarnes. Henceforth it will be necessary to do `pip install plasmapy[calculator]` to install the prototype plasma calculator.

 - [x] Create the `calculator` dependency set
 - [x] Update the plasma calculator documentation page with installation instructions
 - [x] Update package installation instructions
 - [ ] Verify that the plasma calculator works with the specified dependencies. Since this is now an optional dependency set and plasma calculator development has been slow over the last few years, this will likely involve putting some upper limits on dependencies.
 - [ ] Update Nox sessions for building docs and running tests with the new dependency sets

Closes #2979.  Related to #2655.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -223,12 +223,13 @@ def tests(session: nox.Session, test_specifier: nox._parametrize.Param) -> None:
 
     match test_specifier:
         case "lowest-direct":
-            session.install(".[tests]", "--resolution=lowest-direct")
+            session.install(".[tests,calculator]", "--resolution=lowest-direct")
         case _:
             # From https://nox.thea.codes/en/stable/cookbook.html#using-a-lockfile
             session.run_install(
                 *uv_sync,
                 "--extra=tests",
+                "--extra=calculator",
                 env={"UV_PROJECT_ENVIRONMENT": session.virtualenv.location},
             )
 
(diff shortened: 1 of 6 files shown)


noxfile.py
```diff
@@ -61,7 +61,7 @@
 running_on_ci = os.getenv("CI")
 running_on_rtd = os.environ.get("READTHEDOCS") == "True"
 
-uv_requirement = "uv >= 0.6.5"
+uv_requirement = "uv >= 0.6.12"
```
namurphy, line 64:
```suggestion
uv_requirement = "uv >= 0.6.5"
```

PR #2981, opened 2025-04-05
Enable a pre-commit hook to format Markdown files

namurphy:
This PR adds and applies a `pre-commit` hook to format GitHub-flavored Markdown files with [mdformat](https://mdformat.readthedocs.io/en/stable/), excluding license files. 

I also fixed a link to our Matrix chat, and put some text in admonitions in the release checklist.

This hook does not wrap `.md` files to 72 characters per line because some files use [semantic line breaks](https://sembr.org/), and we may want to add semantic line breaks later on. Word wrapping can also mess up formatting for GitHub issues created from Markdown files. I applied the `--wrap=72` option to `README.md` files since we've done this by convention, but skipped the files such as `.github/content/release-checklist.md` (which contains a release checklist with one line per item in each list).

> [!NOTE]
> It may make sense to use semantic line breaks in reStructuredText files, but we most likely want to keep docstrings justified to 72 characters so that they appear nicely when we use `help()` on a function in the Python REPL.


.github/content/update-requirements-pr-template.md
```diff
@@ -1,19 +1,18 @@
...
 > When it is necessary to pin or put an upper limit on a requirement in [`pyproject.toml`], please [create an issue] to remove this upper limit before the next release.
 
 ## Updated dependencies
+
+[create an issue]: https://github.com/PlasmaPy/PlasmaPy/issues/new?title=Remove+upper+limit+on+version+of
+[nox]: https://nox.thea.codes/en/stable/
+[`.github/workflows/update-pinned-reqs.yml`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/.github/workflows/update-pinned-reqs.yml
+[`noxfile.py`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/noxfile.py
+[`pyproject.toml`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/pyproject.toml
+[`uv.lock`]: https://github.com/PlasmaPy/PlasmaPy/blob/main/uv.lock
```
namurphy, line 18: A table with package updates is dynamically generated and added to the GitHub post, but putting the links at the bottom shouldn't cause any issues.  We'll notice pretty quickly if the formatting gets messed up.

.pre-commit-config.yaml
```diff
@@ -118,3 +118,19 @@ repos:
...
+    # Skip wrapping line length because doing so can mess up
+    # formatting of GitHub issues created from files in
+    # .github/content, and some files use semantic line breaks.
+    exclude: licenses/.*|LICENSE.md|PATENT.md
+    additional_dependencies:
+    - mdformat-black
+    - mdformat-footnote
+    - mdformat-frontmatter
+    - mdformat-gfm
+    - mdformat-gfm-alerts
```
namurphy, line 136: Here, `gfm` refers to GitHub Flavored Markdown.

uv.lock
namurphy, line 1: I did a `nox -s requirements` out of habit and figured why not. 😅

PR #2997, opened 2025-04-28
Organize `.gitignore` to use templates from GitHub

namurphy:
This PR reorganizes `.gitignore` to put custom & PlasmaPy content is near the top, while having templates below.  I tried to minimally edit the templates so that it's easier to do a `git diff` later on to see what in the templates have changed.

Instead of ignoring `.idea` — the PyCharm settings folder — I put in the template from JetBrains. My hope is that we'll be able to set up certain test configurations that will be common across projects, so that it isn't necessary for everyone to set this up.  

> [!NOTE]
> I thought about including the .gitignore template specific to VS Code, but decided to keep ignoring the entirety of the configuration directory (`.vscode/`).  I plan to maintain the PyCharm configuration, but we would need someone to maintain the VS Code configuration.

I also made a few slight changes to `MANIFEST.in`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,14 +85,14 @@ ipython_config.py
 # pyenv
 #   For a library or package, you might want to ignore these files since the code is
 #   intended to run in multiple environments; otherwise, check them in:
-# .python-version
+.python-version
 
 # pipenv
 #   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
 #   However, in case of collaboration, if having platform-specific dependencies or dependencies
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,37 +1,87 @@
...
+_version.py
+
+# basedpyright (baseline files should be cached in CI if we adopt this tool)
+.basedpyright
+
+# Cython files (uncomment if we adopt Cython)
+# *.[ao]
+# *.[ch]
+# *.cpp
+# *.dll
```
namurphy, line 30: Cross-reference: #2256

PR #2998, opened 2025-05-01
Update usage of `uv` by Nox

namurphy:
After a recent release of Nox, we had a test failure show up in #2975 for Python 3.11 with the `lowest-direct` resolution strategy.  This PR updates the corresponding Nox session to fix the failure. 

 - I updated some `uv run` commands to use `uvx` since the invocation is a little cleaner.
 - I set the backend for Nox to be `uv` without `virtualenv` as a backup. A lot of the Nox sessions included a command to install `uv`, so these commands are no longer necessary.
 - I changed the dev requirement from `nox` to `nox[uv]`.


noxfile.py
```diff
@@ -61,8 +61,6 @@
 running_on_ci = os.getenv("CI")
 running_on_rtd = os.environ.get("READTHEDOCS") == "True"
 
-uv_requirement = "uv >= 0.6.5"
-
```
namurphy, line 70: In retrospect, including a lower bound on `uv` here wasn't necessary or helpful, especially since `uv` gets new releases quite frequently.

noxfile.py
```diff
@@ -229,6 +227,8 @@ def tests(session: nox.Session, test_specifier: nox._parametrize.Param) -> None:
             session.run_install(
                 *uv_sync,
                 "--extra=tests",
+                "--no-default-extras",
+                f"--python={session.virtualenv.location}",
```
namurphy, line 231: These additional lines are from https://github.com/wntrblm/nox/pull/933, which added them to the Nox + uv recipe.

.github/workflows/ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         cache-suffix: ${{ matrix.nox_session }}-${{ matrix.python }}-${{ runner.os }}
 
     - name: Run the check
-      run: uv run --with nox --frozen nox -s '${{ matrix.nox_session }}'
+      run: uvx nox -s '${{ matrix.nox_session }}'
```
namurphy, line 103: I had added `--frozen` here originally to make sure that `uv.lock` was not changed. However, `uvx` runs `nox` from an isolated virtual environment independent of the local `uv.lock`.  Probably. 😅

noxfile.py
```diff
@@ -54,15 +54,13 @@
 current_python = f"{sys.version_info.major}.{sys.version_info.minor}"
 nox.options.sessions = [f"tests-{current_python}(skipslow)"]
 
-nox.options.default_venv_backend = "uv|virtualenv"
+nox.options.default_venv_backend = "uv"
```
namurphy, line 57: Because there were a whole bunch of Nox sessions that installed and depended upon `uv`, it makes sense to have `uv` be the only backend.

noxfile.py
```diff
@@ -223,12 +216,20 @@ def tests(session: nox.Session, test_specifier: nox._parametrize.Param) -> None:
 
     match test_specifier:
         case "lowest-direct":
-            session.install(".[tests]", "--resolution=lowest-direct")
+            session.install(
+                ".[tests]",
+                "--resolution=lowest-direct",
+                "--no-all-extras",
+                f"--python={session.virtualenv.location}",
+                env={"UV_PROJECT_ENVIRONMENT": session.virtualenv.location},
```
namurphy, line 224: These are the lines that fixed the newly failing test.

pyproject.toml
```diff
@@ -112,7 +112,7 @@ optional-dependencies.docs = [
 optional-dependencies.tests = [
   "hypothesis>=6.108.2",
   "mypy>=1.11",
-  "nox>=2024.10.9",
+  "nox[uv]>=2025.5.1",
```
namurphy, line 115: @rocco8773 suggested that it would be worthwhile to have a developer dependency set including `nox[uv]`, and then have both of the `tests` and `docs` dependency sets depend on them (see also https://github.com/pypa/pip/issues/11296).  But, not in this PR.

PR #2999, opened 2025-05-01
Update installation instructions for uv

namurphy:
This PR updates the installation instructions using `uv` to include different commands to activate virtual environments from different operating systems, since there are different commands in PowerShell and Linux/macOS.

I'll need to come back later to make sure that the group-tabs feature is working as intended.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -43,17 +43,20 @@ terminal and run:
 
    python -m pip install plasmapy
 
-On some systems, it might be necessary to specify the |Python| version
-number by using ``python3``, ``python3.8``, ``python3.9``,
-``python3.10``, or ``python3.11`` instead of ``python``.
+
+.. note::
+
+   On some systems, it might be necessary to specify the |Python|
+   version by using ``python3``, ``python3.11``, ``python3.12``,  or
+   ``python3.13`` instead of ``python``.
 
 To install PlasmaPy on Windows, run:
 
 .. code-block:: bash
 
-   py -3.11 -m pip install plasmapy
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -22,38 +22,32 @@ PlasmaPy requires a version of |Python| between |minpython| and
...
 instructions to `download Python`_ and install it. 🐍
 
-.. tip::
-
-   New versions of |Python| are released annually in October, and it can
-   take a few months for the scientific Python ecosystem to catch up. If
-   you have trouble installing `plasmapy` on the most recent |Python|
-   version between October and March, then try installing it on the
-   second most recent version.
-
```
namurphy, line 32: This was primarily an issue for back when PlasmaPy depended on Numba, which often took several months to become compatible with each new version of Python. Since PlasmaPy no longer depends on Numba, the delay will be significantly shorter.

docs/install.rst
```diff
@@ -101,24 +115,24 @@ Installing PlasmaPy with Conda
...
 into an activated Conda_ environment by opening a terminal and running:
 
 .. code-block:: bash
 
    conda install -c conda-forge plasmapy
 
-Here :bash:`-c conda-forge` indicates that `plasmapy` should be installed
-from the conda-forge_ channel.
+Here :bash:`-c conda-forge` indicates that PlasmaPy should be installed
+from tkhe conda-forge_ channel.
```
namurphy:
```suggestion
from the conda-forge_ channel.
```

PR #3006, opened 2025-05-06
Ignore `.idea` folder in `.gitignore` again

namurphy:
In #2997, I added a `.gitignore` template for the `.idea/` folder, which is where JetBrains products such as PyCharm store project configuration files.  This PR updates `.gitignore` to ignore the `.idea/` folder again. 

My motivation for #2997 included that it would be helpful to have common PyCharm settings for PlasmaPy development, such as shared test configurations so that PyCharm users don't need to recreate them independently.  The drawback is that these settings would need to be maintained, and we likely won't have the capacity to maintain these configurations after the conclusion of the current NSF award supporting PlasmaPy development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -222,7 +222,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-.idea/  # To add a PyCharm integration, put it in .idea/.gitignore
+.idea/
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -222,95 +222,14 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/  # To add a PyCharm integration, put it in .idea/.gitignore
```
namurphy, line 225:
```suggestion
.idea/
```

PR #1643, opened 2022-07-21
Remove `plasmapy_sphinx`

rocco8773:
This PR updates `plasmapy` in accordance with moving `plasmapy_sphinx` into its own repository, [`plasmapy_sphinx`](https://github.com/PlasmaPy/plasmapy_sphinx).

Items that changed...

- Completely removed `docs/plasmapy_sphinx` and put it into its own repo [`plasmapy_sphinx`](https://github.com/PlasmaPy/plasmapy_sphinx).
- Removed any stub files `docs/api_static` related to `plasmapy_sphinx`.
- Removed theme templates `docs/_templates` and CSS files `docs/_static/css` and integrated them into `plasmapy_sphinx`.
- Added a `docs/_static/css/overrides.css` stylesheet so style changes can be made quickly, allowing time for them to integrated into `plasmapy_sphinx`.
- Added `plasmapy_sphinx` as a `docs` dependency.
- Updated `docs/conf.py` to used the `plasmapy_theme`, which is available with the `plasmapy_sphinx.theme` extension.
- Added sphinx extensions `plasmapy_sphinx.ext.autodoc` and `plasmapy_sphinx.ext.directives`.
- Added `plasmapy_sphinx` to the `intersphinx_mapping`.
- Removed upper limit on `sphinx`, and repaired build warnings that resulted.
  - Added `("py:class", "dimensionsless")` to `nitpicky_ignore_regex`.  Astropy does not produce `intersphinx` links for `dimensionless_unscaled`, which caused build warning since we use this as a type annotation.
  - Updated `nbsphinx_prolog` `jinja2` template so variable `docname` is always converted to a string.  This was a problem in `sphinx>=8.0`, since `BuildEnvironment.doc2path()` started returning a `_StrPath` object instead of a string.
- Updated `weekly.yml` and `noxfile.py` to included a weekly build test on the dev/latest version of `plasmapy_sphinx`.
- Added `requirements/*` and `.venvs` to `.gitignore`.
---

### Note

The requirements currently have `plasmapy_sphinx` being pulled directly from the GitHub repo, since it has not been released to PyPI yet.  We can consider making the switch to PyPI once `plasmapy_sphinx` is released.  However, my current thought is to stick with pulling from GitHub since we can then get immediate updates.

PR #3007, opened 2025-05-07
Repair the `weekly.yml` workflow

rocco8773:
The `weekly.yml` workflow had been failing since `nox` was trying to be run directly while it had not been explicitly installed.  This appears to have occurred when moving to `setup-uv` in PR #2959.

This was easily resolved by running the `nox` command with `uv run --with nox`.

After solving the above issue, I noticed the documentation tests (specifically for `sphinx`) were not working as intended.  Since `sphinx` had an upper dependency limit placed on it, the installer could not properly resolve dependencies with `plasmapy` and the latest `sphinx`.  To resolve this issue...

1. I created the `_get_dependencies_from_pyproject_toml()` helper function (in `noxfile.py`), so I could extract the package dependencies directly from the `pyproject.toml`.
2. I then removed the target package dependency (e.g. `sphinx`), so there would be no resolution conflict.
3. I installed all the dependencies manually using `nox.Session.install`.
4. I installed `plasmapy` using `nox.Session.install` with the `--no-deps` flag so dependencies would not be overwritten.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -402,7 +402,7 @@ def build_docs_with_dev_version_of(
         *list(deps.values()),
         silent=False,
     )
-    session.install("--no-deps", "--force-install", ".")
+    session.install("--no-deps", ".")
     session.run(*sphinx_base_command, *build_html, *session.posargs)
 
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -123,7 +123,7 @@ def _create_requirements_pr_message(uv_output: str, session: nox.Session) -> Non
 
 def _get_dependencies_from_pyproject_toml(extras: Optional[str] = None):
     _PYTPROJECT_TOML = (_HERE / "pyproject.toml").resolve()
-    with open(_PYTPROJECT_TOML, "rb") as file:
+    with _PYTPROJECT_TOML.open(mode="rb") as file:
         data = tomllib.load(file)
         config = data["project"]
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -360,7 +360,7 @@ def build_docs_with_dev_version_of(
     so that they can be fixed or updated earlier rather than later.
     """
     session.install(".[docs]")
-    session.install("--ignore-installed", f"git+https://{site}.com/{repository}")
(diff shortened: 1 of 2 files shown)


rocco8773:
pre-commit.ci autofix

rocco8773:
@namurphy I did not thing this PR needed a change log so I added the "no changelog entry needed" label.  If you feel otherwise, then feel free to remove the label.

noxfile.py
```diff
@@ -29,8 +29,10 @@
 import re
 import shutil
 import sys
+import tomllib
 
 import nox
+from packaging.requirements import Requirement
```
namurphy, line 35: It's cool to learn that this exists!

noxfile.py
```diff
@@ -358,7 +388,19 @@ def build_docs_with_dev_version_of(
...
+    pkg_name = repository.split("/")[-1]
+    deps = _get_dependencies_from_pyproject_toml(extras="docs")
+    deps.pop(pkg_name, None)
+
+    session.install(
+        f"git+https://{site}.com/{repository}",
+        *list(deps.values()),
+        silent=False,
+    )
+    session.install("--no-deps", ".")
```
namurphy, line 403: 👍🏻 This will be a more robust way than how we had set this up before.

namurphy (approved):
LGTM!  Thank you for fixing my mistakes. 🙏🏻   

We do have a "run weekly tests in CI" label which I should have used in the earlier PR.  Maybe we should add that to the auto-labeler...

noxfile.py
rocco8773: Yeah, I just found this.  I was not looking forward to parsing the dependency strings, but I was happy to find this.

rocco8773:
Note, the one failing weekly test is due to `plasmapy_sphinx` we have in `docs/plasmapy_sphinx` not having the updates for `sphinx>7.4`.  The `main` branch of `plasmapy_sphinx` currently has these fixes, so making a new release of `plasmapy_sphinx` and merging #1643 will resolve this check.

PR #3008, opened 2025-05-07
Automatically add label to run weekly tests in CI when certain files are modified

namurphy:
As gracefully pointed out by @rocco8773 in [this comment](https://github.com/PlasmaPy/PlasmaPy/pull/3007#issue-3047090812), I broke our weekly tests in #2959 because I did not run the weekly tests in CI. 

This PR updates the automatic labeler of GitHub issues so that the label to run weekly tests is automatically added when `.github/workflows/weekly.yml` is modified.

I'm considering whether we should run weekly tests when `noxfile.py` is modified, but we'd probably need to remove that label about 90% of the time. 🤔


rocco8773:
> I'm considering whether we should run weekly tests when `noxfile.py` is modified, but we'd probably need to remove that label about 90% of the time. 🤔

My 2 cents, I do think it would be safer to also automatically add the label when the `noxfile.py`.  I'd rather have to remove the label than forget to added it.

If possible, I think it would be good for an auto comment (not commit) to be made that says the label has be attached.

namurphy:
> My 2 cents, I do think it would be safer to also automatically add the label when the `noxfile.py`. 

✅ Done! I also decided to add this label when `docs/conf.py` is changed, since there might be some sphinx-dev issues that arise when we change stuff.

🌠 It would probably be helpful to eventually separate the doc stuff from the other stuff for weekly tests.

> If possible, I think it would be good for an auto comment (not commit) to be made that says the label has be attached.

It'd be possible, but since I'd have to relearn a lot of GitHub Actions scripting stuff, this would likely be more effort than it is worth.

rocco8773 (approved):
🚀

PR #3010, opened 2025-05-08
Fix `htmlzip` build for public RTD builds

rocco8773:
Since PR #2937, the RTD builds for non-hidden versions have been failing.  This has been a result of the the `htmlzip` build failing, and not the main `html` build.  We do not see this failure on PR builds, because RTD does not produce the html zip file for those builds.

Here's an example of the error...

image

The cause of the error is `python` not being available outside the `nox` environments used to do the html build.  To remedy this I created a new `nox` runner `docs_bundle_htmlzip` and defined a htmlzip build command in the `.readthedocs.yml`.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -32,5 +32,8 @@ build:
       html:
       - uv run --extra docs --frozen --exact nox -s docs
 
+      htmlzip:
+      - uv run nox -s docs_bundle_htmlzip
+
     post_build:
     - echo $'\n'⚠️ For help deciphering documentation build error messages, see:$'\n\n'\ \ https://docs.plasmapy.org/en/latest/contributing/doc_guide.html#troubleshooting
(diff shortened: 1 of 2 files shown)


rocco8773:
pre-commit.ci autofix

rocco8773:
@namurphy I did not feel this needed a changlog entry so I added the https://github.com/PlasmaPy/PlasmaPy/labels/no%20changelog%20entry%20needed label.  If you feel otherwise, then just remove the label.

As I stated in the main comment, the zip file will not get generated to the PR RTD build.  I made sure this PR was working by activating the RTD build for the branch on my fork, see https://plasmapy-rocco8773.readthedocs.io/en/repair_rtd_builds/.  If you go to the flyout, then the html zip file can be downloaded.

namurphy (approved):
Thank you for catching this!  Please feel free to merge when ready.

Issue #3026, opened 2025-05-29
Reduce duration of synthetic charged particle radiography tests


namurphy:
## Description

We should look into speeding up `test_multiple_grids`, `test_multiple_grids2`, and `test_add_wire_mesh` in `test_synthetic_radiography.py`. This will likely be possible by improving the performance of the particle tracker (#2911, #2978), but we could potentially reduce the scale of the tests in other ways (like perhaps the number of particles or the size of the grid).

## Alternatives

We currently have a `@pytest.mark.slow` decorator to mark tests that take ≳ 1 second, and these tests are decorated accordingly. However, the tests that are marked as slow are still run in every pull request. 

We could possibly create another decorator (perhaps `@pytest.mark.intensive`?) that could be used for tests that take ≳ 15 seconds.  Instead of running the numerically intensive tests on every pull request, we could perhaps run them more selectively (i.e., only when certain files are changed in a PR, and also as part of the weekly tests).  🤔

## Timings

Currently, PlasmaPy's tests with the longest durations are:

| Time | File | Test |
| ---- | ---- | ---- |
| 210.51 s | `test_synthetic_radiography.py` | `test_multiple_grids` |
| 198.45 s | `test_synthetic_radiography.py` | `test_multiple_grids2` |
| 20.98 s  | `test_synthetic_radiography.py` | `test_add_wire_mesh` |
| 5.55 s   | `test_nullpoint.py` | `test_null_point_find6` |
| 4.77 s   | `test_nullpoint.py` | `test_null_point_find4` |

The timings are from running the Python 3.11 tests with lowest direct dependencies on Windows in #3001. The three slowest tests took 87% of the time of the full test run. For some reason I don't understand, the Python 3.12 tests were close to twice as fast. 🤔

pheuer:
@JaydenR2305 Could you maybe take a look at this? Scattering PR will presumably add some more tests, so worth thinking about whether we can reduce particle# or scope, or combine some of these tests using fixtures to reduce total run time?

PR #3027, opened 2025-05-29
Skip slow tests when testing against lowest direct dependencies

namurphy:
The synthetic charged particle radiography tests take several minutes to run (see #3026), and for some reason take close to twice as long when running the tests on Windows with Python 3.11. This PR makes it so that slow tests are skipped when testing against the lowest versions of direct dependencies for the purpose of speeding up CI.


namurphy:
@rocco8773 suggested that I :
 - [ ] Make sure that the 3.11 lowest-direct tests are still part of the weekly CI

PR #3037, opened 2025-06-14
[SLA] Add helper function `merge_voltage_clusters`

rocco8773:
Add helper function `merge_voltage_clusters` to the `plasmapy.analysis.swept_langmuir` module.  This function takes the two time series voltage and current arrays of a swept langmuir signal and merges clusters/islands of closely spaced voltage values.  Since several langmuir analysis techniques requires differentiation, the often irregularly spaced voltage signals can cause erratic behavior during differentiation.  Merging clusters can help "smooth out" the signal for differentiation.

### DIFF
--- a/src/plasmapy/analysis/swept_langmuir/helpers.py
+++ b/src/plasmapy/analysis/swept_langmuir/helpers.py
@@ -162,8 +162,16 @@ def sort_sweep_arrays(
     current: np.ndarray,
     voltage_order: Literal["ascending", "descending"] = "ascending",
 ) -> tuple[np.ndarray, np.ndarray]:
-    if voltage_order not in ["ascending", "descending"]:
-        raise ValueError()
+    if not isinstance(voltage_order, str):
+        raise TypeError(
+            "Expected 'voltage_order' to be a string equal to 'ascending' "
+            f"or 'descending', but got type {type(voltage_order)}."
+        )
+    elif voltage_order not in ["ascending", "descending"]:
+        raise ValueError(
+            "Expected 'voltage_order' to be a string equal to 'ascending' "
+            f"or 'descending', but got '{voltage_order}'."
+        )
 
     voltage, current = check_sweep(
         voltage, current, strip_units=True, allow_unsorted=True
--- a/src/plasmapy/analysis/swept_langmuir/helpers.py
+++ b/src/plasmapy/analysis/swept_langmuir/helpers.py
@@ -1,6 +1,6 @@
 """Helper functions for analyzing swept Langmuir traces."""
 
-__all__ = ["check_sweep", "sort_sweep_arrays"]
+__all__ = ["check_sweep", "merge_voltage_clusters", "sort_sweep_arrays"]
 
 from typing import Literal
 
@@ -234,3 +234,98 @@ def sort_sweep_arrays(
         current = current[::-1]
 
     return voltage, current
+
+
+def merge_voltage_clusters(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    voltage_step_size: float = None,
+    force_regular_grid: bool = False,
+):
+    # step_size = 0 (only merge identical voltages)
+    # None : calculate steps size as average of array step sizes
+
+    voltage, current = check_sweep(voltage, current)
+
+    if not isinstance(force_regular_grid, bool):
+        raise TypeError(
+            "Expected 'force_regular_grid' to be a bool, but got type "
+            f"{type(force_regular_grid)}."
+        )
+
+    if voltage_step_size is None:
+        voltage_step_size = np.abs(np.average(np.diff(voltage)))
+    elif voltage_step_size == 0:
+        voltage_step_size = 0.0
+    elif not isinstance(voltage_step_size, float):
+        raise TypeError(
+            "Expected 'voltage_step_size' to be a float or None, got type "
+            f"{type(voltage_step_size)}."
+        )
+    elif voltage_step_size < 0:
+        voltage_step_size = -voltage_step_size
+
+    voltage_ascending = bool(np.all(np.diff(voltage) >= 0))
+    if not voltage_ascending:
+        voltage, current = sort_sweep_arrays(voltage, current, voltage_order="ascending")
+
+    # now merge clusters
+    new_voltage = np.unique(voltage)
+    if voltage_step_size == 0 and new_voltage.size == voltage.size:
+        new_current = current.copy()
+    elif voltage_step_size == 0:
+        new_current = np.empty_like(new_voltage, dtype=current.dtype)
+
+        for ii in range(new_voltage.size):
+            mask = voltage == new_voltage[ii]
+            if np.count_nonzero(mask) == 1:
+                new_current[ii] = current[mask]
+            else:
+                new_current[ii] = np.average(current[mask])
+    else:
+        new_voltage = np.empty_like(voltage, dtype=voltage.dtype)
+        new_voltage[...] = np.nan
+        new_current = np.empty_like(current, dtype=current.dtype)
+        new_current[...] = np.nan
+
+        start_voltage = voltage[0]
+        stop_voltage = start_voltage + voltage_step_size
+
+        ii = 0
+        while start_voltage <= voltage[-1]:
+            mask1 = voltage >= start_voltage
+            mask2 = voltage < stop_voltage
+            mask = np.logical_and(mask1, mask2)
+
+            if np.count_nonzero(mask) > 0:
+                new_voltage[ii] = (
+                    start_voltage + 0.5 * voltage_step_size if force_regular_grid
+                    else np.average(voltage[mask])
+                )
+                new_current[ii] = np.average(current[mask])
+
+                ii += 1
+
+            start_voltage = stop_voltage
+            stop_voltage = start_voltage + voltage_step_size
+
+        # filter out NaN values
+        new_voltage = new_voltage[:ii]
+        new_current = new_current[:ii]
+
+        # force regular spacing
+        if force_regular_grid:
+            # Need to fill array with NaN values to force the regular spacing
+            ...
+
+    if not voltage_ascending:
+        voltage, current = sort_sweep_arrays(voltage, current, voltage_order="descending")
+        new_voltage, new_current = sort_sweep_arrays(
+            new_voltage, new_current, voltage_order="descending"
+        )
+    else:
+        new_voltage, new_current = sort_sweep_arrays(
+            new_voltage, new_current, voltage_order="ascending"
+        )
+
+    return new_voltage, new_current
--- a/src/plasmapy/analysis/swept_langmuir/helpers.py
+++ b/src/plasmapy/analysis/swept_langmuir/helpers.py
@@ -236,61 +236,164 @@ def sort_sweep_arrays(
     return voltage, current
 
 
+def _nan_stuff_regular_grid(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    voltage_step_size: float,
+) -> tuple[np.ndarray, np.ndarray]:
+    size = int(np.round((voltage[-1] - voltage[0]) / voltage_step_size)) + 1
+    reg_voltage = np.linspace(voltage[0], voltage[-1], num=size, dtype=voltage.dtype)
+
+    _, reg_indices, new_indices = np.intersect1d(
+        np.round(reg_voltage, decimals=5),
+        np.round(voltage, decimals=5),
+        return_indices=True,
+    )
+    mask = np.ones_like(reg_voltage, dtype=bool)
+    mask[reg_indices] = False
+    reg_voltage[mask] = np.nan
+
+    reg_current = np.full(size, np.nan, dtype=current.dtype)
+    reg_current[np.logical_not(mask)] = current[...]
+
+    return reg_voltage, reg_current
+
+
 def merge_voltage_clusters(
     voltage: np.ndarray,
     current: np.ndarray,
     voltage_step_size: float = None,
-    force_regular_grid: bool = False,
-):
-    # step_size = 0 (only merge identical voltages)
-    # None : calculate steps size as average of array step sizes
-
-    voltage, current = check_sweep(voltage, current)
-
-    if not isinstance(force_regular_grid, bool):
+    force_regular_spacing: bool = False,
+) -> tuple[np.ndarray, np.ndarray]:
+    # condition force_regular_grid
+    if not isinstance(force_regular_spacing, bool):
         raise TypeError(
-            "Expected 'force_regular_grid' to be a bool, but got type "
-            f"{type(force_regular_grid)}."
+            "Expected 'force_regular_spacing' to be a bool, but got type "
+            f"{type(force_regular_spacing)}."
         )
 
-    if voltage_step_size is None:
-        voltage_step_size = np.abs(np.average(np.diff(voltage)))
-    elif voltage_step_size == 0:
-        voltage_step_size = 0.0
-    elif not isinstance(voltage_step_size, float):
+    # condition voltage_step_size
+    if (
+        voltage_step_size is not None
+        and not isinstance(voltage_step_size, (float, np.floating, int, np.integer))
+    ):
         raise TypeError(
             "Expected 'voltage_step_size' to be a float or None, got type "
             f"{type(voltage_step_size)}."
         )
+    elif isinstance(voltage_step_size, (int, np.integer)):
+        voltage_step_size = float(voltage_step_size)
+
+    voltage, current = check_sweep(voltage, current)
+
+    # check if grid is regularly spaced
+    voltage_diff = np.diff(voltage)
+    mask_zero_diff = np.isclose(voltage_diff, 0.0)
+    if np.count_nonzero(mask_zero_diff) > 0:
+        is_regular_grid = False
+    elif np.allclose(voltage_diff, voltage_diff[0]):
+        # grid is already regularly spaced
+        is_regular_grid = True
+    elif np.allclose(
+        np.rint(voltage_diff / np.min(voltage_diff))
+        - voltage_diff / np.min(voltage_diff),
+        0,
+    ):
+        # grid has a common dV, but at times jumps N * dV times
+        if force_regular_spacing and (voltage_step_size is None or voltage_step_size == 0):
+            # need to stuff with NaN values
+            voltage, current = _nan_stuff_regular_grid(
+                voltage=voltage,
+                current=current,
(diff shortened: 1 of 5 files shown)


rocco8773:
pre-commit.ci autofix

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -1,37 +1,50 @@
...
-        Values should start from a negative ion-saturation current and increase
-        to a positive electron-saturation current.  *No units are assumed or
-        checked, but values should be in amperes.*
+        1D `numpy.ndarray` representing the current of the swept
+        Langmuir trace.  Values should start from a negative
+        ion-saturation current and increase to a positive
+        electron-saturation current.  *No units are assumed or checked,
+        but values should be in amperes.*
 
     strip_units: `bool`
```
namurphy:
```suggestion
    strip_units: `bool`, default: `True`
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -1,37 +1,50 @@
...
-        (Default: `True`) If `True`, then the units on ``voltage`` and/or
-        ``current`` will be stripped if either are passed in as an Astropy
-        `~astropy.units.Quantity`.
+        (Default: `True`) If `True`, then the units on ``voltage``
```
namurphy:
```suggestion
        If `True`, then the units on ``voltage``
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -1,37 +1,50 @@
...
+        and/or ``current`` will be stripped if either are passed in as
+        an Astropy `~astropy.units.Quantity`.
+
+    allow_unsorted: `bool`
+        (Default: `False`) If `True`, then the supplied ``voltage``
```
namurphy:
```suggestion
    allow_unsorted: `bool`, default: `True`
        If `True`, then the supplied ``voltage``
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -145,3 +158,364 @@ def check_sweep(  # noqa: C901, PLR0912
...
+    current: `numpy.ndarray`
+        1D `numpy.ndarray` representing the current of the swept
+        Langmuir trace.  Values should start from a negative
+        ion-saturation current and increase to a positive
+        electron-saturation current.  *No units are assumed or checked,
+        but values should be in amperes.*
+
+    voltage_order: `str`
+        Either 'ascending' or 'descending' to indicate how the
+        ``voltage`` array should be sorted.  (DEFAULT: ``'ascending'``)
```
namurphy:
```suggestion
    voltage_order: `str`, default: ``'ascending'``
        Either ``'ascending'`` or ``'descending'`` to indicate how the
        ``voltage`` array should be sorted.
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -145,3 +158,364 @@ def check_sweep(  # noqa: C901, PLR0912
...
+    new_voltage = new_voltage[nan_mask]
+    new_current = new_current[nan_mask]
+
+    return new_voltage, new_current
+
+
+def merge_voltage_clusters(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    voltage_step_size: float | None = None,
```
namurphy, line 408: I can tell you're not an AI because you didn't use `typing.Optional` here.

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -145,3 +158,364 @@ def check_sweep(  # noqa: C901, PLR0912
...
+        ion-saturation current and increase to a positive
+        electron-saturation current.  *No units are assumed or checked,
+        but values should be in amperes.*
+
+    voltage_step_size: `float` or `None`
+        A non-zero, positive step size for the ``voltage`` array
+        cluster identification.  A value of ``0`` will merge only
+        duplicate voltage values.  A value of `None` will default to
+        95% of the average step size of the ``voltage`` array (only
+        counting duplicate voltages once).  (DEFAULT:`None`)
```
namurphy:
```suggestion
    voltage_step_size: `float` | `None`, default: `None`
        A non-zero, positive step size for the ``voltage`` array
        cluster identification.  A value of ``0`` will merge only
        duplicate voltage values.  A value of `None` will default to
        95% of the average step size of the ``voltage`` array (only
        counting duplicate voltages once).
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -145,3 +158,364 @@ def check_sweep(  # noqa: C901, PLR0912
...
+
+    .. math::
+
+        N = \texttt{floor}\left(
+            \frac{V_{cluster,max} - V_{cluster,min}}{\texttt{voltage_step_size}}
+        \right)
+    """
+    # condition voltage_step_size
+    if voltage_step_size is not None and not isinstance(
+        voltage_step_size, float | np.floating | int | np.integer
```
namurphy:
```suggestion
        voltage_step_size, numbers.Real
```
...I think?

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -145,3 +158,364 @@ def check_sweep(  # noqa: C901, PLR0912
...
+    ):
+        raise TypeError(
+            "Expected 'voltage_step_size' to be a float or None, got type "
+            f"{type(voltage_step_size)}."
+        )
+    elif isinstance(voltage_step_size, int | np.integer):
```
namurphy:
```suggestion
    elif isinstance(voltage_step_size, numbers.Integral):
```
...I think?

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +237,285 @@ def sort_sweep_arrays(
...
+def merge_voltage_clusters(
+    voltage: np.ndarray,
+    current: np.ndarray,
+    voltage_step_size: float | None = None,
+) -> tuple[np.ndarray, np.ndarray]:
+    r"""
+    Search the ``voltage`` array for closely spaced voltage clusters
+    based on the ``voltage_step_size`` parameter and merge those
+    clusters, and associated ``current`` values, into a single point.
+
```
namurphy, line 415: State the purpose of this function?

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +237,285 @@ def sort_sweep_arrays(
...
+    if np.count_nonzero(mask_zero_diff) > 0:
+        # is_regular_grid = False
+        pass
+    elif np.allclose(voltage_diff, voltage_diff[0]):
+        # grid is already regularly spaced
+        is_regular_grid = True
+    elif np.allclose(
+        np.rint(voltage_diff / np.min(voltage_diff))
+        - voltage_diff / np.min(voltage_diff),
+        0,
```
namurphy, line 262: Should we include `atol` and/or `rtol` explicitly?

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +237,285 @@ def sort_sweep_arrays(
         current = current[::-1]
 
     return voltage, current
+
+
+def _is_voltage_regularly_spaced(
```
namurphy, line 243: Does this function need to handle `numpy.nan` values?

..perhaps not, since I think they may get filtered out beforehand.

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +237,285 @@ def sort_sweep_arrays(
...
+    mask_zero_diff = np.isclose(voltage_diff, 0.0)
+    mask_zero_diff = np.append(mask_zero_diff, [mask_zero_diff[-1]])
+
+    # initialize new voltage and current arrays
+    new_voltage = np.full(voltage.shape, np.nan, dtype=voltage.dtype)
+    new_current = np.full(current.shape, np.nan, dtype=current.dtype)
+
+    mask = np.logical_not(mask_zero_diff)
+    new_voltage[mask] = voltage[mask]
+    new_current[mask] = current[mask]
```
namurphy, line 292: This reminded me of [`numpy.ma`](https://numpy.org/doc/stable/reference/maskedarray.html) which handles masked arrays, though I suspect it wouldn't be worth the effort to use them in this particular case.

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +237,285 @@ def sort_sweep_arrays(
...
+) -> bool:
+    """
+    Determine if the voltage difference array ``voltage_diff`` is
+    regularly spaced; that is the differences are all equal or some
+    integer multiple of the smallest difference.
+    """
+    is_regular_grid = False
+    if np.count_nonzero(mask_zero_diff) > 0:
+        # is_regular_grid = False
+        pass
```
namurphy:
```suggestion
        return False
```

tests/analysis/swept_langmuir/test_helpers__merge_voltage_clusters.py
```diff
@@ -0,0 +1,165 @@
+"""
+Tests for helper function `merge_voltage_clusters` contained in
+`plasmapy.analysis.swept_langmuir.helpers`.
+"""
+
+from contextlib import nullcontext as does_not_raise
```
namurphy, line 6: Happy to see good use of `contextlib`!

namurphy (approved):
I put in some minor thoughts and mostly optional suggestions, and overall it looks to be in quite good shape.  Thank you for doing this!

src/plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: Default values for `atol` and `rtol` are `1e-8` and `1e-5` respectively.  Since voltage measurements are unlikely to be on the nV scale, the default values is likely to be adequate.  Will leave the code unchanged. (conclusion made via in-person discussion)

src/plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: If the original `voltage` array has `np.nan` values, then `voltage_diff=np.diff(voltage)` will contain `np.nan` values to either side of the original `np.nan` location.  All the routines in the if-clause can handle `np.nan` values and would result in `is_regular_grid` being `False`, as it should be.

So, I do not think we need to handle it explicitly.

src/plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: I think `numpy.ma` would be useful if we're trying to maintain views onto the original array, versus duplicating.

The top-level function `merge_voltage_clusters` returns new `voltage` and `current` arrays.  I did this because I did not want to manipulate the original arrays if the user was going to keep them around for some purpose.  Additionally, I doubt too many users would be familiar with the mechanics of `numpy.ma`.

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+    mask_zero_diff: np.ndarray,
+) -> bool:
+    """
+    Determine if the voltage difference array ``voltage_diff`` is
+    regularly spaced; that is the differences are all equal or some
+    integer multiple of the smallest difference.
+    """
+    is_regular_grid = False
+    if np.count_nonzero(mask_zero_diff) > 0:
+        # is_regular_grid = False
```
namurphy, line 254:
```suggestion
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+        return False
+    elif np.allclose(voltage_diff, voltage_diff[0]):
+        # grid is already regularly spaced
+        is_regular_grid = True
+    elif np.allclose(
```
namurphy, line 259:
```suggestion
    if np.allclose(voltage_diff, voltage_diff[0]):
        # grid is already regularly spaced
        return True
    if np.allclose(
```
[optional] I have a slight style preference to use `return` statements to avoid having to create an intermediate variable, and so that we can use separate `if` statements rather than a larger `elif` construct.

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+    values.
+    """
+    # generate boolean mask for zero diff locations
+    voltage_diff = np.diff(voltage)
+    mask_zero_diff = np.isclose(voltage_diff, 0.0)
+    mask_zero_diff = np.append(mask_zero_diff, [mask_zero_diff[-1]])
+
+    # initialize new voltage and current arrays
+    new_voltage = np.full(voltage.shape, np.nan, dtype=voltage.dtype)
+    new_current = np.full(current.shape, np.nan, dtype=current.dtype)
```
namurphy, line 288: I keep forgetting `numpy.full` exists. 😅

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+    mask = np.logical_not(mask_zero_diff)
+    new_voltage[mask] = voltage[mask]
+    new_current[mask] = current[mask]
+
+    # merge zero diff clusters
+    while np.any(mask_zero_diff):
+        volt = voltage[mask_zero_diff][0]
+        index = np.where(new_voltage == volt)[0]
+        if len(index) == 0:
+            index = np.where(mask_zero_diff)[0]
```
namurphy: [optional] Reading this, I got a little confused about the different between `voltage` and `volt`.  I'm wondering if there's an alternate name for `volt`. 🤔

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+    if indices.size > 1:
+        indices[2:] = indices[2:] + 1
+    if cluster_mask[-1] and indices[-1] != voltage.size - 1:
+        # last element in voltage is in a cluster
+        indices = np.append(indices, [voltage.size - 1])
+    indices = np.reshape(indices, newshape=(int(indices.size / 2), 2))
+    # ^ using newshape kwarg for backwards compatibility, newshape kwarg
+    #   has been deprecated since numpy v2.1
+
+    # populate
```
namurphy:
```suggestion
    # populate voltages and currents
```
Should we expand the comment into something like this?

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+            \frac{V_{cluster,max} - V_{cluster,min}}{\texttt{voltage_step_size}}
+        \right)
+    """
+    # condition voltage_step_size
+    if voltage_step_size is not None and not isinstance(
+        voltage_step_size, numbers.Real
+    ):
+        raise TypeError(
+            "Expected 'voltage_step_size' to be a float or None, got type "
+            f"{type(voltage_step_size)}."
```
namurphy, line 474:
```suggestion
            f"{type(voltage_step_size).__name__}."
```
...something PyCharm suggested to me earlier 🙃

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+        # check if the ValueError was a result of voltage containing nan
+        # values
+        if not np.any(np.isnan(voltage)):
+            # voltage array has no nan values
+            raise ValueError(*err.args) from err
+
+        if not filter_nan:
+            raise ValueError(
+                "The voltage array contains NaN values.  If you want NaN "
+                "values to be automatically filtered set argument "
```
namurphy:
```suggestion
                "values to be automatically filtered, set argument "
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+        pass
+    elif filter_nan:
+        # mask out current NaN values
+        nan_mask = np.logical_not(np.isnan(current))
+        voltage = voltage[nan_mask]
+        current = current[nan_mask]
+    else:
+        raise ValueError(
+            "The current array contains NaN values.  If you want NaN "
+            "values to be automatically filtered set argument "
```
namurphy:
```suggestion
            "values to be automatically filtered, set argument "
```

src/plasmapy/analysis/swept_langmuir/helpers.py
```diff
@@ -235,3 +238,330 @@ def sort_sweep_arrays(
...
+        warnings.warn(
+            "The supplied ``voltage`` array is already regularly spaced. If "
+            "you want to re-bin the arrays to a different voltage_step_size, "
+            "then use something like numpy.interp.  No merging performed.",
+            PlasmaPyWarning,
+        )
+
+        return voltage.copy(), current.copy()
+
+    # condition voltage_step_size ... Round 2
```
namurphy, line 535: Should the comment say why round 2 is needed? 🤔

tests/analysis/swept_langmuir/test_helpers__merge_voltage_clusters.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+
+        if expected is not None:
+            assert np.allclose(rtn_voltage, expected[0])
+            assert np.allclose(rtn_current, expected[1])
+        else:
+            assert np.allclose(rtn_voltage, voltage)
+            assert np.allclose(rtn_current, current)
+
+        assert 0 < mock_sweep.call_count <= 2
```
namurphy, line 196: Would it make sense to add a comment about why this check is here?

namurphy (approved):

src/plasmapy/analysis/swept_langmuir/helpers.py
rocco8773: Sticking with `f"{type(voltage_step_size)}"` since it creates a printout that includes the full package structure, e.g. `numpy.ndarray` instead of just `ndarray`.

PR #3039, opened 2025-06-17
Reduce CPR test time

### LINKED ISSUE #3026: Reduce duration of synthetic charged particle radiography tests

pheuer:
Closes #3026 

This PR should go a long way to reducing the CPR module test execution time. The following changes are made: 

1) test_multiple_grids and test_multiple_grids2 were literally just duplicate tests - I have deleted the second one. 
2) test_multiple_grids is just a "does it run" test: no actual validation, so it definitely didn't need 1E5 particles. I've dropped it down to just 100 particles, so that should run much faster. 
3) test_add_wire_mesh was a badly structured test that was actually running a BUNCH of simulations to test a variety of things. I've broken the input validation tests into two other parametrized tests. 
4) The remaining `test_add_wire_mesh_accuracy` is actually a pretty important test for the CPR module: it checks that a radiograph of a grid returns the expected image, which is probably the best integrated test of the whole module. So, I left it alone but found I could reduce the particle number some and it still passed. So, I've done that.

### DIFF
--- a/changelog/3039.internal.rst
+++ b/changelog/3039.internal.rst
@@ -1,0 +1,1 @@
+Refactored tests for the `~plasmapy.diagnostics.charged_particle_radiography` module to reduce test runtime.
--- a/tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
+++ b/tests/diagnostics/charged_particle_radiography/test_synthetic_radiography.py
@@ -809,11 +809,11 @@ def test_gaussian_sphere_analytical_comparison() -> None:
     "kwargs",
     [
         # Test a circular mesh
-        {"extent": 1 * u.mm},
+        ({"extent": 1 * u.mm}),
         # Test providing hdir
-        {"mesh_hdir": np.array([0.5, 0, 0.5])},
+        ({"mesh_hdir": np.array([0.5, 0, 0.5])}),
         # Test providing hdir and vdir
-        {"mesh_hdir": np.array([0.5, 0, 0.5]), "mesh_vdir": np.array([0, 0.1, 1])},
+        ({"mesh_hdir": np.array([0.5, 0, 0.5]), "mesh_vdir": np.array([0, 0.1, 1])}),
     ],
 )
 @pytest.mark.slow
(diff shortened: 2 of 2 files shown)


pheuer:
pre-commit.ci autofix

pheuer:
@namurphy ready for review

namurphy (approved):
@pheuer — thank you!  This will be very helpful to keep our tests fast.  Feel free to merge when you're ready to!

Cross-reference: #2585

PR #3046, opened 2025-07-03
Bump minimum versions of dependencies as per SPEC 0

namurphy:
[SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends that packages in the scientific pythoniverse drop core dependencies that are more than 24 months old.  This PR bumps the minimum versions of several of PlasmaPy's dependencies to be consistent with SPEC 0.

🌠🤖 It'd be great to find a way to automate this.  

> [!NOTE]
> PlasmaPy still has upper limits on Astropy and xarray which we should fix before the next release.


.github/workflows/update-pinned-reqs.yml
```diff
@@ -2,7 +2,7 @@ name: Update pinned requirements
 
 on:
   schedule:
-  - cron: 37 11 * * 5
+  - cron: 37 11 * * 4
```
namurphy: Fixing an off-by-one error from #3041.  It must be Thursday. I never could get the hang of Thursdays.

pyproject.toml
```diff
@@ -57,69 +57,70 @@ dynamic = [ "version" ]
...
-  "sphinx>=8.2.3",
-  "sphinx-changelog>=1.5",
-  "sphinx-codeautolink==0.15.2",
+  "docutils>=0.20.2",
+  "jinja2>=3.1.6",
+  "nbsphinx>=0.9.7",
+  "nox>=2025.5.1",
+  "numpydoc>=1.9",
+  "pillow>=11.3",
+  "plasmapy-sphinx>=2025.5.0",
```
namurphy:
```suggestion
  "plasmapy-sphinx>=0.2.0b1",
```

PR #3056, opened 2025-07-29
Fix hyperlinks in documentation

namurphy:
To start the week off with an easy PR, this PR is fixing broken hyperlinks and permanent redirects ahead of the release which we're planning to do this week.

> [!NOTE]
> There may still be some hyperlinks that get tagged as broken because of timeouts, but I'll test them manually.  If they recur later on, we can add them to the linkcheck ignore configuration variable.


src/plasmapy/formulary/laser.py
```diff
@@ -472,7 +472,7 @@ def Gaussian_Rayleigh_length(
 
     Notes
     -----
-    For details, see :wikipedia:`Rayleigh Length`\ .
+    For details, see :wikipedia:`Rayleigh length`.
```
namurphy, line 475: - [ ] Double check this

PR #3080, opened 2025-08-05
Use `GITHUB_TOKEN` in workflow to mint a release

namurphy:
I ran into an authentication error with the `git push` command at the end of `mint-release.yml`. The goal of this PR is to fix that.

### DIFF
--- a/.github/workflows/mint-release.yml
+++ b/.github/workflows/mint-release.yml
@@ -6,11 +6,12 @@ on:
       tag:
         description: Release tag 🏷️
         required: true
-        default: 2024.5.0rc1
+        default: 2025.8.0rc1
       DOI:
         description: Reserved DOI from Zenodo
         required: true
-        default: 10.5281/zenodo.1238132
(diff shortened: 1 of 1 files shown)


.github/workflows/mint-release.yml
```diff
@@ -6,11 +6,12 @@ on:
       tag:
         description: Release tag 🏷️
         required: true
-        default: 2024.5.0rc1
+        default: 2025.8.0rc1
       DOI:
         description: Reserved DOI from Zenodo
         required: true
-        default: 10.5281/zenodo.1238132
+        default: 10.5281/zenodo.16747747
```
namurphy, line 13: The value I just reserved for `v2025.8.0`

.github/workflows/mint-release.yml
```diff
@@ -6,11 +6,12 @@ on:
...
+
```
namurphy, line 14:
```suggestion
```

PR #2943, opened 2025-01-16
Build documentation with Python 3.13

namurphy:
Read the Docs now appears to allow documentation to be built with Python 3.13, so this PR adjusts PlasmaPy's doc build configuration accordingly.

Closes #2874. Potential overlap with #2937.

PR #3091, opened 2025-08-11
Build documentation with Python 3.13 in CI

namurphy:
I apparently forgot to do this in #2943. 😅

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -54,7 +54,7 @@ _HERE = pathlib.Path(__file__).parent
 # the Docs. Because Read the Docs takes some time to support new
 # releases of Python, we should not link docpython to maxpython.
 
-docpython = "3.12"
+docpython = "3.13"
 
 current_python = f"{sys.version_info.major}.{sys.version_info.minor}"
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -131,9 +131,10 @@
 maximum_signature_line_length = 90
 sphinxemoji_style = "twemoji"
 
-suppress_warnings = [
-    "autosummary.import_cycle",
-]
+# suppress_warnings = [
+#    "autosummary.import_cycle",
+#    "docutils",  # to suppress complaints about the add docstring section for to_hz
+# ]
```
namurphy:
```suggestion
```

docs/conf.py
```diff
@@ -131,10 +131,6 @@
 maximum_signature_line_length = 90
 sphinxemoji_style = "twemoji"
 
-suppress_warnings = [
-    "autosummary.import_cycle",
-]
```
namurphy, line 136: The `autosummary.import_cycle` warnings no longer appeared, and the new warnings from this PR actually did refer to a real problem with indentation in docstrings created by the decorator for `to_hz` that we didn't get warnings about before.

PR #3094, opened 2025-08-12
Update mypy configuration and add type hint annotations

namurphy:
...since my flight home to visit family was delayed.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -5,6 +5,11 @@ exclude = (?x)(
           docs|
           \.run|
           \.tox|
+          \.nox|
+          env|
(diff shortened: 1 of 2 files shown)


mypy.ini
```diff
@@ -5,6 +5,11 @@ exclude = (?x)(
           docs|
           \.run|
           \.tox|
+          \.nox|
+          env|
+          .env|
+          venv|
+          \.venv|
```
namurphy, line 12: ...since mypy found errors in some libraries contained in a conda environment from a few years ago on my old work laptop.

PR #3096, opened 2025-08-18
Add a `Material` class

JaydenR2305:
In this PR, I outline a consistent structure for stopping and ranging (STAR) data by defining `MaterialSTARData`. I then use this definition to create an interface for users to create compounds using the `Material` class. This is critical to my PR implementing Moliere scattering (#2748) to ensure that we provide support for users who wish to scatter through a compound

### DIFF
--- a/src/plasmapy/plasma/material.py
+++ b/src/plasmapy/plasma/material.py
@@ -3,6 +3,7 @@ Definition of the material class and helper functions. The module also
 includes type definitions for stopping and ranging data.
 """
 
+from collections.abc import Callable
 from functools import cached_property
 from typing import TypedDict
 
@@ -33,7 +34,9 @@ class MaterialSTARData(TypedDict):
     projected_range: u.Quantity[u.cm]
 
 
-def _construct_log_interpolator(x0: u.Quantity, y0: u.Quantity):
+def _construct_log_interpolator(
+    x0: u.Quantity, y0: u.Quantity
+) -> Callable[[u.Quantity], u.Quantity]:
     """
     Apply a logarithmic transformation to input and output data points,
     allowing for approximation of an exponential relation using
--- a/src/plasmapy/plasma/material.py
+++ b/src/plasmapy/plasma/material.py
@@ -1,0 +1,190 @@
+"""
+Definition of the material class and helper functions. The module also
+includes type definitions for stopping and ranging data.
+"""
+
+from functools import cached_property
+from typing import TypedDict
+
+import astropy.units as u
+import numpy as np
+from scipy.interpolate import CubicSpline
+
+from plasmapy.particles.particle_class import Particle
+
+__all__ = [
+    "MaterialSTARData",
+    "Material",
+]
+
+from plasmapy.utils.decorators import validate_quantities
+
+
+class MaterialSTARData(TypedDict):
+    """Type definition for formatting data related to beam-target interactions."""
+
+    incident_particle: Particle
+
+    energies: u.Quantity[u.MeV]
+    stopping_power: u.Quantity[
+        u.MeV * u.cm**2 / u.g
+    ]  # Units of energy per areal density
+    CSDA_range: u.Quantity[u.cm]
+    projected_range: u.Quantity[u.cm]
+
+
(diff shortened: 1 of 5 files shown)


src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+    ----------
+    A : `~astropy.units.Quantity`
+        An array of the atomic weights of the elements in the compound in units
+        convertible to atomic mass units.
+
+    Z : `~astropy.units.Quantity`
+        An array of the atomic numbers of the elements making up the compound.
+        Does not have any units.
+
+    fractional_weights : `~astropy.units.Quantity`
```
pheuer, line 75: I think it's much better to have these be the number fractions. When you read a chemical formula, that's what you have available, e.g. C2H4, = C:2, H:4. Calculating the weight fractions in the code is then easy using the atomic masses.

src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+class Material:
+    """A representation of a compound material.
+
+    Parameters
+    ----------
+    A : `~astropy.units.Quantity`
+        An array of the atomic weights of the elements in the compound in units
+        convertible to atomic mass units.
+
+    Z : `~astropy.units.Quantity`
```
pheuer, line 71: If A is the atomic number, then Z is usually the ionization state in a plasma. But this is a solid, so I think you only need A. In the Moliere scattering then, "z" is "A".

src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+    fractional_weights : `~astropy.units.Quantity`
+        An array containing the fractional composition of the compound. For
+        example, if the value at index 0 has a value of 0.33, that element would
+        make up 33% of the compound by weight.
+
+    density : `~astropy.units.Quantity`
+        The density of the material in units convertible to kilograms per cubic
+        meter.
+
+    name : str, optional
```
pheuer, line 84: I think I suggested this but now IDK if we really need the name attribute

src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+__all__ = [
+    "MaterialSTARData",
+    "Material",
+]
+
+from plasmapy.utils.decorators import validate_quantities
+
+
+class MaterialSTARData(TypedDict):
+    """Type definition for formatting data related to beam-target interactions."""
```
pheuer, line 25: In theory these are redundant - once you know the stopping power, you could calculate the others. Not saying the class has to do that, but it could in theory.

pheuer: I think it should be possible for some of these to be None - what if all you have, and all you need, is stopping power data?

src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+        make up 33% of the compound by weight.
+
+    density : `~astropy.units.Quantity`
+        The density of the material in units convertible to kilograms per cubic
+        meter.
+
+    name : str, optional
+        The name of the material.
+
+    STAR_data : `~plasmapy.plasma.material.MaterialSTARData`, optional
```
pheuer, line 87: I don't really like the idea of providing this as an iterable of this class in the case of different particles. How about storing this internally as a nested dict, something like d['proton']['stopping_power'], then creating methods like add_stopping_power(particle, energy_axis, data) to populate it, creating the interpolators as you go?

src/plasmapy/plasma/material.py
```diff
@@ -0,0 +1,193 @@
...
+class MaterialSTARData(TypedDict):
+    """Type definition for formatting data related to beam-target interactions."""
+
+    incident_particle: Particle
+
+    energies: u.Quantity[u.MeV]
+    stopping_power: u.Quantity[
+        u.MeV * u.cm**2 / u.g
+    ]  # Units of energy per areal density
+    CSDA_range: u.Quantity[u.cm]
```
pheuer, line 33: How is CSDA range different than projected range? I thought all the projected range data was using the CDSA approximation.

PR #3098, opened 2025-08-22
DEP: drop unused runtime dependency on setuptools

neutrinoceros:
I noticed this while reviewing dependencies. I found neither `setuptools`, `distutils` or `pkg_resources` in the source code, and I got tests passing without it, so I think it's safe to remove it.


namurphy:
Good catch!  I think this is a remnant from needing `pkg_resources` somewhere, but it's no longer needed.  Happy to merge this!

PR #3024, opened 2025-05-27
Remove `sphinx-hoverxref` as an extension to the documentation build

namurphy:
Back in #1353, we added ``sphinx-hoverxref`` as a Sphinx extension to the documentation build so that readers of the documentation could hover over links and get previews. Because that functionality is now included in Read the Docs add-ons, this extension has been deprecated.

I had been planning to wait until after our next release to remove ``sphinx-hoverxref``, but there are some problems with the sphinx-dev doc build in the weekly tests (which are being run in #2998), so it seems better to remove it now.

Closes #3009.

PR #3099, opened 2025-08-22
TST: adjust uv settings to never build third party libraries even with `--resolution=lowest-direct`

neutrinoceros:
I noticed while running `nox -s tests` locally on a non-Linux machine that every session using the `lowest-direct` resolution strategy from uv failed because it tried building numpy from source. This adjusts settings to always prefer pre-comipled binary distributions even in this context (the oldest available *compatibile* binary will be selected on each platform instead of the absolute minimal requirement, which in general is platform dependent)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,6 +139,9 @@ namespaces = false
 [tool.setuptools_scm]
 write_to = "src/plasmapy/_version.py"
 
+[tool.uv]
+no-build-package = [ "numpy", "matplotlib", "scipy", "astropy", "h5py", "pandas", "wrapt" ]
+
 [tool.ruff]
 target-version = "py311"
 namespace-packages = [ ".github/workflows", "docs" ]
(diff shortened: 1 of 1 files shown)


neutrinoceros:
pandas and astropy both produce nightly wheels, so you shouldn't need to build them ever. I'll adjust the noxfile accordingly.

neutrinoceros:
Welp. Pretty weird how lmfit doesn't seem to install properly in doc envs. I'm honestly not sure whether this could be a uv bug. In any case I don't know when I'll be able to investigate this further.

namurphy:
> I found the problem; there was a logic defect in your `noxfile.py`

Many thanks for fixing this!  

I just updated minimum requirements in #3024 keeping this PR in mind, and am happy to merge this.  I'm hoping it'll help with some problems I've been running into in #3110 as well.

And with that, I'm happy to merge this. Thank you once again!

namurphy (approved):

PR #3110, opened 2025-09-22
Drop Python 3.11 and add Python 3.14

### LINKED ISSUE #3042: Separate out weekly tests for development versions of packages
Currently we have one big file for weekly tests that includes running tests and building docs, including with the development versions of upstream packages like Sphinx, pandas, NumPy, etc. It's been fairly common for us to pin an upper limit on Sphinx, which means that there is often a known failure in these tests.

It would be helpful to separate out the comprehensive tests which are run against locked dependencies from the tests against the development versions of package.  We could have a demarcation like:

 - **Comprehensive tests** — a more complete set of test runs and documentation builds, run using only versions of dependencies that have been released and are within the requirements from `pyproject.toml`
 - **Tests against development versions of packages** — Includes running tests and building documentation build with the versions of packages from their repositories

This way, we could have separate issue labels for "run comprehensive tests" along with "run development version tests".

### LINKED ISSUE #3047: Drop Python 3.11 support for first release after October 2025
[SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends that packages in the scientific pythoniverse support minor releases of Python that have been made in the past three years. Python 3.11 was released on 2022 Oct 24, so PlasmaPy should support Python 3.12+ for releases made after that date.

> [!NOTE]
> PlasmaPy is due for a release soon (#2920), but since I don't know if we'll do another release before October, we should wait a few months before implementing this.

See the [What's New in Python 3.12](https://docs.python.org/3/whatsnew/3.12.html) for more details. I don't anticipate that any of the improvements will significantly impact PlasmaPy source code, but we could make use of some typing improvements (quoted verbatim):

 - [PEP 692](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep692), using [TypedDict](https://docs.python.org/3.12/library/typing.html#typing.TypedDict) to annotate [**kwargs](https://docs.python.org/3.12/glossary.html#term-argument)
 - [PEP 695](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep695), type parameter syntax and the [type](https://docs.python.org/3.12/reference/simple_stmts.html#type) statement
 - [PEP 698](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep698), [typing.override()](https://docs.python.org/3.12/library/typing.html#typing.override) decorator

### LINKED ISSUE #3100: Enable tests for Python 3.14
Python 3.14 is scheduled to be released in October 2025. Once that happens, we'll need to start running tests against Python 3.14.  We'll need to make changes in `noxfile.py`, `.github/workflows/ci.yml`, and probably a few other places.

namurphy:
This PR makes the following changes:

 - **Dropped Python 3.11 support.** The minimum version of Python for PlasmaPy v2025.10.0 and newer is now Python 3.12, as per the guidelines in [SPEC 0](https://scientific-python.org/specs/spec-0000/).  
 - **Added Python 3.14** to the suite of continuous integration tests.
 - **Fixed a problem where PlasmaPy was unable to be imported in Python 3.14** (see #3123). 
   - This problem occurred due to a breaking change in how Python handles the attributes of `typing.Union` objects like `ParticleLike` and `ParticleListLike`.  The docstrings of these objects are no longer directly set using `ParticleLike.__doc__` or `ParticleLikeLike.__doc__`.
 - Split up the weekly continuous integration checks into comprehensive checks (which perform a more thorough suite of checks), upstream checks (which run tests and build docs using the newest unreleased versions of several upstream dependencies), and a check in a separate file that PlasmaPy can be installed with conda.
 - Added documentation headers to several GitHub workflow files.

Updates of the Python version for documentation builds are being done in #3091, but are lower priority.

Closes #3047. Closes #3100.  Closes #3042. Partially addresses #3123.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -30,6 +30,11 @@ instructions to `download Python`_ and install it. :snake:
    version between October and March, then try installing it on the
    second most recent version.
 
+.. note::
+
+   PlasmaPy versions through ``v2025.8.0`` require Python 3.13 or
+   earlier and are incompatible with Python 3.14.
+
 .. _install-pip:
 
 Installing PlasmaPy with pip


noxfile.py
```diff
@@ -147,26 +151,33 @@ def _get_dependencies_from_pyproject_toml(extras: str | None = None):
...
-    # When silent is `True`, `session.run()` returns a multi-line string
-    # with the standard output and standard error.
+    # Running `uv lock` fails when `uv.lock` has a merge conflict or
+    # is otherwise invalid. To avoid situations like this, preemptively
+    # delete `uv.lock`. Since `uv lock` uses information in `uv.lock`
+    # even if `--upgrade` is specified, deleting `uv.lock` may make
+    # running `uv lock` more deterministic.
+
+    lockfile = pathlib.Path(root_dir / "uv.lock")
+    lockfile.unlink(missing_ok=True)
```
namurphy, line 171: I considered deleting `uv.lock` only when running `uv lock` locally or when a git merge is underway, but decided it's generally better to do the same thing locally as in CI most of the time.

namurphy:
Oh good, only about 30000 lines of errors.  Most of them are something like this:
```
ERROR src/plasmapy/utils/decorators/helpers.py::plasmapy.utils.decorators.helpers.preserve_signature - AttributeError: 'typing.Union' object attribute '__doc__' is read-only
```

noxfile.py
namurphy: Deleting `uv.lock` removes pins and other settings that are not included in `pyproject.toml`, but we want `pyproject.toml` to be the definitive source for requirements information, so in our case this should be fine.

pyproject.toml
```diff
@@ -34,7 +34,6 @@ classifiers = [
   "Natural Language :: English",
   "Operating System :: OS Independent",
   "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.11",
```
namurphy, line 37: 🫡

noxfile.py
namurphy: Turns out that this change made it so that the tables in the pull requests to upgrade pinned requirements were empty (e.g., #3149). Fixing this in #3150.

Issue #3107, opened 2025-09-09
Switch from `pre-commit` to `prek`?


namurphy:
PlasmaPy currently uses [`pre-commit`](https://pre-commit.com/) to run numerous hooks that perform code quality checks and apply automated changes. I often describe `pre-commit` as "extremely helpful, and only moderately annoying". 😅  

[`prek`](https://prek.j178.dev/) is a new drop-in replacement for `pre-commit` written in Rust, though `prek` is not yet ready for production use (see also [GitHub repo](https://github.com/j178/prek)).  It's significantly faster than `pre-commit`, and has some [new features](https://prek.j178.dev/#prek-provides-a-better-user-experience) that I suspect will make `prek` more user-friendly than `pre-commit`.

I plan on playing around with `prek` locally, and we can consider switching to it for the repository if/when there's an equivalent to [`pre-commit.ci`](https://pre-commit.ci).

PR #3113, opened 2025-09-26
Remove `pre-commit-search-and-replace` hook for `pre-commit`

namurphy:
We currently use `pre-commit-search-and-replace` to search for patterns and make some replacements.  This hook helped us find a bunch of typos and make documentation improvements, which was really helpful.  

This hook has been a bit tricky to use since it sometimes requires pre-commit installing Ruby locally for it to work.  I also had trouble using this hook when using `prek` instead of `pre-commit` in #3107. This PR tentatively removes this hook, though we may add it back in the future.


.pre-commit-config.yaml
```diff
@@ -2,10 +2,6 @@ ci:
   autofix_prs: false
   autoupdate_schedule: quarterly
 
-# Specify the default version of Ruby in case it is not installed
-default_language_version:
-  ruby: 2.7.2
-
```
namurphy, line 8: Specifying a default language version for Ruby was to try to make it so that pre-commit would install Ruby. However, I've still run into problems when trying to use this hook that pre-commit was unable to install Ruby.

.pre-commit-config.yaml
```diff
@@ -53,13 +49,6 @@ repos:
     # run this hook on `labeler.yml` (or certain other files).
     exclude: .github/labeler.yml|.pre-commit-search-and-replace.yaml
 
-- repo: https://github.com/mattlqx/pre-commit-search-and-replace
-  # keep at v1.0.5
-  rev: v1.0.5
```
namurphy, line 58: I ran into problems with more recent releases of this hook, including with the Ruby installation.

Issue #3071, opened 2025-08-02
Intermittent test failure in `test_at_most_one_api_call`


namurphy:
I've noticed in a few recent PRs that there's a test failure in `test_downloader.py` that's been occurring infrequently:

```
__________________________ test_at_most_one_api_call ___________________________
[gw1] darwin -- Python 3.12.11 /Users/runner/work/PlasmaPy/PlasmaPy/.nox/tests-3-12-cov/bin/python3
tests/utils/data/test_downloader.py:238: in test_at_most_one_api_call
    assert used1 <= used0 + 1
E   assert 820 <= (818 + 1)
```

Some possibilities would be to:

 - Keep the test as-is and tolerate the occasional failures
 - Separate out the downloader tests into a separate Nox session that only gets run in certain situations
 - Add an additional tolerance to the test (i.e., change the `+ 1` to `+ 2`) but still spit out a warning
 - Remove the test entirely

PR #3125, opened 2025-10-24
Decorate downloader tests with `@pytest.mark.slow`

namurphy:
We've had a lot of intermittent test failures with the downloader functionality (see #3071).  As a tentative way to reduce the frequency that we run into these intermittent test failures, this PR marks the test as slow. As a consequence, the downloader tests will only be run on a regular pull request for the one version of Python for which coverage is being tested. This doesn't fix the problem, but rather reduces the rate at which we'll run into these false positives.

### DIFF
--- a/tests/utils/data/test_downloader.py
+++ b/tests/utils/data/test_downloader.py
@@ -67,7 +67,6 @@ test_urls = [
 ]
 
 
-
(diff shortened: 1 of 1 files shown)


namurphy:
pre-commit.ci autofix

PR #3126, opened 2025-10-24
Update minimum requirements for docs and tests dependency sets

namurphy:
This PR updates the `tests` and `docs` dependency sets to closer to the most recent releases of each package. Because these dependency sets are not user-facing, it's reasonable to keep the much more recent requirements since there's not much benefit in supporting old versions of pytest, for example.

### DIFF
--- a/.github/workflows/ci-upstream.yml
+++ b/.github/workflows/ci-upstream.yml
@@ -38,7 +38,7 @@ jobs:
         - name: Tests, Python 3.14, astropy-dev, Ubuntu
           os: ubuntu-latest
           python: '3.14'
-          nox_session: docs_upstream(astropy)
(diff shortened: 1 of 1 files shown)


namurphy:
pre-commit.ci autofix

.github/workflows/ci-upstream.yml
```diff
@@ -38,47 +38,47 @@ jobs:
         - name: Tests, Python 3.14, astropy-dev, Ubuntu
           os: ubuntu-latest
           python: '3.14'
-          nox_session: run_tests_with_dev_version_of(astropy)
+          nox_session: docs_upstream(astropy)
```
namurphy:
```suggestion
          nox_session: test_upstream(astropy)
```

PR #3157, opened 2025-11-25
Switch from optional dependencies to dependency groups

namurphy:
The configuration for PlasmaPy in `pyproject.toml` has been defining optional dependencies for `tests` and `docs`. These dependencies have been able to be installed with commands like `pip install plasmapy[docs,tests]` (for installation from PyPI) or with `pip install -e .[docs,tests]` (for an editable local installations). However, the documentation and test dependencies are not intended for use by end users, and it is unnecessary (and probably undesirable) for these optional dependencies to be included in built distributions.

[Dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) were defined in [PEP 735](https://peps.python.org/pep-0735/). This PR switches PlasmaPy from using optional dependencies to dependency groups, which are excluded from built distributions of PlasmaPy.

### Checklist

 - [x] Update `pyproject.toml`
 - [x] Update `uv.lock` (will probably need to do this multiple times)
 - [x] Changelog entry
 - [x] Update multiple sessions in `noxfile.py` to use dependency groups.
 - [x] Update Read the Docs configuration, if necessary
 - [x] Update installation instructions in contributor guide to show how to install these dependency groups with:
   - [x] pip
   - [x] uv
 - [x] Finish #3000
 - [ ] ~~Revive #3001 to simplify the process of setting up a development virtual environment?~~

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -179,24 +179,19 @@ def lock(session: nox.Session) -> None:
 @nox.session
 def validate_lockfile(session: nox.Session) -> None:
     """
-    Ensure consistency of uv.lock with requirements in pyproject.toml.
+    Ensure that uv.lock is consistent with pyproject.toml.
 
-    This session invokes the uv-lock hook for pre-commit to update
-    check and update uv.lock.
-
-    While this check is normally performed locally when running
-    pre-commit (or prek), the uv-lock hook is not run on pre-commit.ci
-    since pre-commit.ci blocks network access. A separate session is
-    necessary so that the validity of uv.lock can be confirmed through a
-    GitHub workflow.
+    This check is normally performed locally when running pre-commit or
+    prek. Because pre-commit.ci blocks network access, this check is
+    instead done in CI via a GitHub workflow that calls this session.
     """
     if running_on_ci:
         errmsg = (
             "The Python environments in file 'uv.lock' are inconsistent "
             "with the requirements defined in 'pyproject.toml'. "
-            "After installing uv, this problem can be fixed by running "
-            "`uvx nox -s validate_lockfile` in the top-level directory "
-            "of your clone of PlasmaPy, and then pushing the updated "
+            "After installing Nox, this problem can be fixed by running "
+            "`nox -s validate_lockfile` in the top-level directory of "
+            "your clone of PlasmaPy, and then pushing the updated "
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
namurphy, line 1: The changes to the pre-commit configuration are essentially adding a docstring. Nothing too important, but helpful for new contributors.

.readthedocs.yml
```diff
@@ -23,17 +23,13 @@ build:
     - asdf plugin add uv
     - asdf install uv latest
     - asdf global uv latest
-    - uv venv --python=3.14
-
-    install:
-    - uv sync --extra docs --frozen --exact
```
namurphy, line 29: I _think_ we can drop the `uv venv` and `uv sync` commands since installing the things is done via `uvx`.

docs/install.rst
```diff
@@ -22,25 +22,17 @@ PlasmaPy requires a version of |Python| between |minpython| and
...
 instructions to `download Python`_ and install it. :snake:
 
-.. tip::
-
-   New versions of |Python| are released annually in October, and it can
-   take a few months for the scientific Python ecosystem to catch up. If
-   you have trouble installing `plasmapy` on the most recent |Python|
-   version between October and March, then try installing it on the
-   second most recent version.
-
```
namurphy, line 32: Since we dropped Numba as a dependency, issues like this should be significantly more rare.

.pre-commit-config.yaml
```diff
@@ -120,19 +137,15 @@ repos:  # https://pre-commit.com/#adding-pre-commit-plugins-to-your-project
...
+  rev: 0.9.24
   hooks:
     # running uv lock requires network access or a warm cache, so it
     # cannot be run via pre-commit.ci, even with the `--offline`,
     # `--no-index`, and/or `--check` flags
   - id: uv-lock
-    name: uv-lock (skipped in CI; run `nox -s lock` for problems)
-    # The cooldown defined here must be identical to the cooldown
-    # defined when running `nox -s lock`.
-    args: [--exclude-newer, 3 days]
```
namurphy, line 132: The cooldown is now defined in the `[tool.uv]` table in `pyproject.toml`.

PR #3194, opened 2026-01-18
Update pytest configuration

namurphy:
- Moved pytest flags from `noxfile.py` to `pyproject.toml`. Keeping configuration variables in `pyproject.toml` makes it so that running `uv sync; pytest` will be very similar to running the `tests` nox session. This is partially motivated because PyCharm doesn't natively support Nox sessions.
 - Dropped `norecursedirs` from the pytest configuration in favor of being more specific with `testpaths`.
 - Enabled the `strict` configuration variable for pytest to help find pytest configuration errors. (Note: this exact strict settings enabled by this variable may change in the future.)
 - Added unique identifiers to parametrized cases in `test_aliases` to allow a stricter pytest configuration.
 - Updated the release checklist to look at issues labeled "needed for release" and fix remaining deprecation warnings.
 - Labeled module-level constants in `noxfile.py` to be in all caps, as per PEP 8.
 - Ran a `pre-commit autoupdate`.

### DIFF
--- a/.github/content/release-checklist.md
+++ b/.github/content/release-checklist.md
@@ -28,7 +28,11 @@ This issue is created from the [release checklist] for releasing a new version o
 
   - Occasionally, certain hooks may need to be manually downgraded after running `pre-commit autoupdate` because of problems with the latest versions of these hooks. Look for comments in [`.pre-commit-config.yaml`] for guidance. 🪝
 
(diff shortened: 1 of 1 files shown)


tests/formulary/test_frequencies.py
```diff
@@ -35,6 +35,7 @@
         (wlh_, lower_hybrid_frequency),
         (wuh_, upper_hybrid_frequency),
     ],
+    ids=["oc_", "wc_", "wlh_", "wuh_"],
```
namurphy, line 38: The unique ids here are needed with `tools.pytest.strict = true` because `oc_` and `wc_` both point to `gyrofrequency`.

PR #3197, opened 2026-01-22
Add a reusable composite action that GitHub workflows can use to run Nox sessions

### LINKED ISSUE #1754: Use reusable workflows to improve modularity of our GitHub Actions

namurphy:
## Description

Add a composite GitHub action that contains all steps necessary to use uv to run a Nox session, and update multiple GitHub workflow files to invoke it. 

The composite action contains steps to install non-Python software needed for building documentation and to upload coverage reports to Codecov, when necessary and appropriate.

Closes #1754.

## Motivation

Several of the GitHub workflows in this repository have the sole purpose of running one or more Nox sessions. These workflows are defined in `.github/workflows/` and include `ci.yml`, `ci-comprehensive.yml`, `linkcheck.yml`, and `upstream-docs.yml`, and `upstream-tests.yml`. Each of these files invokes a different set of Nox sessions that are run under different triggers.

Several of these workflow files are used to run tests, and several are used to build documentation. The documentation workflows each include a step that uses micromamba to install pandoc and graphviz, which is repeated across multiple workflows. That means that if we decide to change how these packages get installed, then we'll need to update it in multiple files. By consolidating the steps into a single file with a composite action, we only need to define that step once.

## Implementation details

 - It turns out that we need to define the GitHub token for both tests and documentation builds.  However, there are some Nox sessions that do not need it, so I set it as not required.
 - The Codecov token is only necessary in `ci.yml` for one case. It would not be unreasonable to define that step only in `ci.yml` instead of in the newly defined workflow.

## Alternatives

A reasonable alternative would be to put most or all of the extra steps into `noxfile.py`. However, this would not include any steps that invoke other GitHub actions. Even if we did go a phase where we were able to put everything into Nox sessions, we could still run into this problem later on if we decide to use a GitHub action.


noxfile.py
```diff
@@ -465,7 +465,7 @@ def docs_upstream(session: nox.Session, site: str, repository: str) -> None:
 """
 
 
-@nox_uv.session(python=DOCPYTHON, uv_groups="docs")
+@nox_uv.session(python=DOCPYTHON, uv_groups=["docs"])
```
namurphy, line 468:
```suggestion
@nox_uv.session(python=DOCPYTHON, uv_groups=["docs"])



```

namurphy:
The downside of consolidating so much into a single action is that the output of each individual step is no longer being shown separately in the GitHub interface.  This would likely make it harder to track down which step failed, so I'm not convinced yet that this PR will ultimately improve contributor experience. 🫤

PR #3201, opened 2026-01-27
Update documentation build to use `sphinx>=9.1`

namurphy:
Following the release of `v2026.1.0` of `plasmapy_sphinx`, this PR updates the documentation build requirement to `sphinx>=9.1`.

 - [x] Add changelog entry
 - [x] Wait three days
 - [x] Re-run `nox -s lock`


rocco8773:
> > These are usually due to a "shortcoming" of the upstream intersphinx mapping. For example, the documentation for `Quantity` is actually `astropy.units.Quantity` even though the definition path is `astropy.units.quantity.Quantity`.
> 
> I tried fixing a bunch of these and I think we're still getting the same warnings. It would have been reaaalllly helpful if the warnings showed the file that this was happening in. I'll try a `sphinx-build` without the `--quiet` option in case that would help. I think it has something to do with usage in type hint annotations somewhere/somehow. 🤔
> 

Yeah, I think they started showing up when we starting using `typing.Annotations`, but I can't say for sure.

namurphy:
I'm setting the documentation build to ignore the `reference target not found` warnings were arising and related to `typing.Annotation[~astropy.units.quantity.Quantity]`, and raised #3312 so that we don't forget it.  Things are passing at this point, thankfully! 🎉🎂💖🎆🪩🌌🥦

PR #3227, opened 2026-02-24
Fix rot a to b

yaochengchen:
Replace exact equality check (== -1) with np.isclose in rot_a_to_b to correctly handle anti-parallel vectors under floating-point rounding.


pheuer (approved):
This is a good idea, thanks!

PR #3244, opened 2026-03-25
Update and re-run tool to export ionization energies from NIST

namurphy:
- Updated ionization energies from NIST (adding tritium).
 - Updated `tools/export_ionization_energy.py` to include inline script metadata and write the resulting JSON file to its final location.
 - Created a session in `noxfile.py` to run this script.

## Tasks

 - [ ] Check if there was special handling for deuterium that needs to be repeated for tritium
 - [ ] Check if there are other isotopes that should be included (like maybe $^3_2\mathrm{He}$ and $^4_2\mathrm{He}$?)
 - [ ] Raise issue about putting data in HDF5 format to hopefully speed up imports?


.gitignore
```diff
@@ -24,11 +28,11 @@ _version.py
...
-# Cython files (uncomment if we adopt Cython)
-# *.[ao]
-# *.[ch]
-# *.cpp
-# *.dll
+# Cython files
+*.[ao]
+*.[ch]
+*.cpp
+*.dll
```
namurphy, line 39:
```suggestion
# Cython files
# *.[ao]
# *.[ch]
# *.cpp
# *.dll

```

namurphy:
The documentation build failure is unrelated to this PR.

PR #3282, opened 2026-05-18
Update internal documentation and GitHub workflows

namurphy:
## Updates to internal documentation

- Updated `.github/workflows/README.md` to list all workflows presenting included in that directory, so that it's easier for future maintainers to get the big picture of which workflows do what.
- Added or revised comments in GitHub workflow files, including a comment at the top of each file that describes the main purpose of the workflow. (These quasi-docstring comments are intended to make it easier for someone to navigate the files in the directory and get an idea of what the workflow defined in each file does.)

## Updates to GitHub workflows

- Removed an exemption for the PyHC workflow because the minimum allowed version of xarray had been newer than PHEP 3 recommended.
- Updated the labels to be removed after merging a PR.
- Removed some obsolete files from the labeler configuration.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -213,7 +213,7 @@ testing:
       '**/test*.py',
       '**/tests/**',
       '*codecov.y*ml',
-      '.github/workflows/ci.yml',
+      '.github/workflows/ci*.yml',
     ]
 
 tools:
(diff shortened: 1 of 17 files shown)


namurphy:
The documentation failure in the comprehensive tests is unrelated.

PR #3294, opened 2026-05-26
Run comprehensive tests less frequently

namurphy:
I've been finding that the comprehensive tests are run much more frequently than needed, so this PR limits how frequently they're run to save computing resources.

🪧 The comprehensive tests will now be run:
 
- For the automated PRs that update `uv.lock`, and
- When the comprehensive workflow file is updated.

🏕️ While in the area, I also changed `nox -s` to `nox --session` to make things a bit more descriptive.  This is one example of "measuring the length of a variable not by the number of characters, but by how long it takes to understand what it means."

🏕️ While also in the area, I changed some YAML formatting so that it works better with prek's check-yaml, and so that diffs appear more cleanly.


.github/labeler.yml
```diff
@@ -8,214 +8,215 @@
 
 'breaking change':
   - changed-files:
-    - any-glob-to-any-file: 'changelog/*breaking*.rst'
+    - any-glob-to-any-file:
+      - 'changelog/*breaking*.rst'
```
namurphy, line 12: Though there is only a single glob pattern here, putting it on its own line will reduce the size of future diffs.

.pre-commit-config.yaml
```diff
@@ -59,8 +59,6 @@ repos:
     # Labels with spaces must be put in quotes in `labeler.yml`.
     # pretty-format-yaml removes those quotes and breaks the action.
     exclude: .github/labeler.yml
-  - id: pretty-format-ini
-    args: [--autofix]
```
namurphy, line 63: We can drop this now that the repository no longer contains `mypy.ini`, `pytest.ini`, or `tox.ini`.  🎉🎂💖🎆🪩🌌🥦

PR #3296, opened 2026-05-26
Automatically add ty ignore flags when upgrading `uv.lock`

namurphy:
## Description

This PR extends the workflow to upgrade `uv.lock` to add new `# ty: ignore` comments.

## Motivation

PlasmaPy uses [ty](https://docs.astral.sh/ty) to perform static type checking. The rule set for ty is still growing, so it's quite possible that new typing failures will show up upon a new release of ty. Fixing typing failures is outside of the scope of these PRs since they take a long time to understand and fix. Moreover, static type checking is mostly intended for human-created PRs.  Hence, it makes sense for these PRs to keep the status quo with regards to typing.


namurphy:
pre-commit.ci autofix

PR #3325, opened 2026-07-06
Allow particle_input and validate_quantities in either order on instance methods

### LINKED ISSUE #2035: Allow `particle_input` and `validate_quantities` to be used in either order for instance methods
### Bug description

At present, `@particle_input` and `@validate_quantities` can be used to decorate most callables in either order (i.e. with `@particle_input` as the inner decorator and `@validate_quantities` as the outer decorator, or vice versa).  

However, this is not true for instance methods.  For instance methods, it's necessary to have `@particle_input` be the outer decorator and `@validate_quantities` to be the inner decorator.

### Expected outcome

We should be able to use `@particle_input` and `@validate_quantities` in either order when decorating an instance method.

### Minimal complete verifiable example

```Python
from plasmapy.particles import particle_input, ParticleLike
import astropy.units as u

class Class:
    @particle_input  # @particle_input as outer decorator
    @validate_quantities
    def method(self, particle: ParticleLike, B: u.T) -> u.T:
        return B

instance = Class()
instance.method("p+", B = 5 * u.T)  # This works fine since @particle_input is the outer decorator

class AnotherClass:
    @validate_quantities  # @validate_quantities as inner decorator
    @particle_input
    def method(self, particle: ParticleLike, B: u.T) -> u.T:
        return B

another_instance = AnotherClass()
another_instance.method("p+", B = 5 * u.T)  # This breaks


Cell In[3], line 8
      5         return B
      7 another_instance = AnotherClass()
----> 8 another_instance.method("p+", B = 5 * u.T)  # This breaks

File ~/Projects/PlasmaPy/plasmapy/utils/decorators/validators.py:206, in ValidateQuantities.__call__.<locals>.wrapper(*args, **kwargs)
    203     bound_args.arguments[arg_name] = arg
    205 # call function
--> 206 _return = f(**bound_args.arguments)
    208 # validate output
    209 if "validations_on_return" in validations:

File ~/Projects/PlasmaPy/plasmapy/particles/decorators.py:882, in particle_input.<locals>.wrapper(callable__, instance, args, kwargs)
    877 @wrapt.decorator
    878 def wrapper(
    879     callable__: Callable, instance: Any, args: tuple, kwargs: dict[str, Any]
    880 ):
    881     new_kwargs = particle_validator.process_arguments(args, kwargs, instance)
--> 882     return callable__(**new_kwargs)

TypeError: AnotherClass.method() missing 1 required positional argument: 'self'
```

### Package versions

_No response_

### Additional context

This might have to do with `@particle_input` making use of `wrapt` functionality for decorators, which isn't used for `@validate_quantities`.  A possibility would be to implement `wrapt` in `@validate_quantities` too, though I do not know if this would actually work.

Mohit-Ak:
## What was wrong

`@particle_input` and `@validate_quantities` can be stacked in either order on
plain functions, but not on instance methods. For methods, `@particle_input` had
to be the *outer* decorator — putting `@validate_quantities` on the outside broke
with:

```
TypeError: AnotherClass.method() missing 1 required positional argument: 'self'
```

Root cause: `@particle_input` is implemented with `wrapt`. When it is accessed as
an instance method (i.e. as the outer decorator), wrapt splits `self` out and
passes it as a separate `instance`, so `_bind_arguments()` correctly pops `self`
before forwarding. But when `@validate_quantities` is the outer decorator, it is
*not* wrapt-based: it binds its own signature (including `self`) and then calls the
inner `particle_input` wrapper as a plain function, with `self` sitting in the
keyword arguments and wrapt's `instance` being `None`. `_bind_arguments()` then
unconditionally popped `self`/`cls`, stripping the argument the wrapped method
still needed.

## The fix

Only pop `self`/`cls` when `instance is not None` — i.e. only when wrapt supplied
the instance separately. When `instance is None`, any `self`/`cls` present came
from the actual call and is preserved so it can be forwarded. `self` carries no
particle annotation, so `process_argument()` already passes it through untouched.

Also removed the now-obsolete "particle_input must be the outer decorator" caveats
from the `particle_input` docstring.

## How it was tested

The two decorator-order cases were already present in
`test_particle_input_with_validate_quantities` marked `xfail(... See #2035)`. With
the fix they pass, so the `xfail` markers were removed (the suite runs
`xfail_strict`, so a lingering xfail on a now-passing case would fail CI). Repro
from the issue now works in both orders.

Test results (Python 3.12, wrapt 2.2.2):

```
tests/particles/test_decorators.py .......................... 84 passed, 1 xfailed
tests/particles/                    ......................... 1883 passed, 2 xfailed
tests/utils/decorators/             ......................... 120 passed
ruff 0.15.15 check .................. All checks passed!
ruff format --check ................. 2 files already formatted
```

The remaining xfails are unrelated (#2150, variadic positional arguments).

Fixes #2035

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -776,3 +776,7 @@ authors:
   family-names: Bonner
   affiliation: Haverford College
   alias: JackBonner888
+
+- given-names: Mohit
+  family-names: Arvind Khakharia
+  alias: Mohit-Ak
--- a/changelog/3325.bugfix.rst
+++ b/changelog/3325.bugfix.rst
@@ -1,0 +1,4 @@
+Enabled |particle_input| and |validate_quantities| to be used in
+either order when decorating an instance method. Previously,
+|particle_input| had to be the outer decorator for instance methods
+(see :issue:`2035`).
--- a/src/plasmapy/particles/decorators.py
+++ b/src/plasmapy/particles/decorators.py
@@ -144,8 +144,17 @@ def _bind_arguments(
 
     bound_arguments.apply_defaults()
 
-    bound_arguments.arguments.pop("self", None)
-    bound_arguments.arguments.pop("cls", None)
+    # When ``instance`` is provided, then wrapt has already separated
+    # ``self``/``cls`` out of ``args``/``kwargs`` and bound it above, so
+    # it must be removed before calling the wrapped callable. But when
+    # ``instance`` is `None` — as happens when |particle_input| is the
+    # *inner* decorator on an instance method and the outer decorator
+    # (e.g. |validate_quantities|) calls it directly with ``self`` among
+    # the arguments — the ``self``/``cls`` argument came from the actual
+    # call and must be preserved so it can be forwarded. See #2035.
+    if instance is not None:
+        bound_arguments.arguments.pop("self", None)
+        bound_arguments.arguments.pop("cls", None)
 
     return bound_arguments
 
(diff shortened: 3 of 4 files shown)


namurphy (approved):
Thank you for figuring out the causes behind #2035 and for making this fix! I appreciate it!

PR #3327, opened 2026-07-08
test: replace bare assert np.isclose/allclose with np.testing.assert_allclose

steps-re:
## Summary

Converts 357 bare `assert np.isclose(...)` and `assert np.allclose(...)` calls
across 41 test files to `np.testing.assert_allclose(...)`.

**Why this matters:** bare `assert np.isclose(a, b)` produces an uninformative
`AssertionError` on failure. `np.testing.assert_allclose` prints the actual vs
desired values, mismatching indices, and the tolerance used — dramatically
reducing debugging time.

## Key details

- **Default tolerances set explicitly** as `rtol=1e-5, atol=1e-8` to match
  `np.isclose` defaults exactly. `np.testing.assert_allclose` defaults to
  `atol=0`, which causes false failures for comparisons against near-zero values.

- **Shape fixes** in `test_nullpoint.py`, `test_mhd_wave_class.py`: unnecessary
  `.reshape(1, 3)` calls converted to `.flatten()`. `np.allclose` silently
  broadcasts mismatched shapes; `assert_allclose` enforces shape equality.

- **Quantity atol** in `test_grids.py`: `float(Quantity)` raises for
  dimensional quantities, so `atol` is stripped via `.value` and both arrays
  via `np.asarray()`.

- **Parametrized tolerance dicts** in `test_speeds.py`: `isclose_kw` dict
  expansion handled via `**{"rtol": 1e-5, "atol": 1e-8, **isclose_kw}` so
  parametrized values override the fallback defaults.

- **Unit-aware list comparison** in `test_mhd_wave_class.py`: Python lists of
  `Quantity` objects wrapped in `u.Quantity()` before comparison so
  `assert_allclose` can compare them correctly.

- **Intentional broadcasting preserved**: particle tracker gyroradius/energy
  conservation tests that compare `(N_timesteps, N_particles)` vs `(N_particles,)`
  are kept as `assert np.allclose(...)` since the broadcasting over all time
  steps is the intended physics check.

## Test results

- 4,343 tests pass (excluding pre-existing errors in `tests/diagnostics/` that
  exist on `main` before this PR).
- No new test failures introduced.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -780,3 +780,7 @@ authors:
 - given-names: Mohit
   family-names: Arvind Khakharia
   alias: Mohit-Ak
+
+- given-names: Mike
+  family-names: German
+  alias: steps-re
--- a/changelog/3327.test.rst
+++ b/changelog/3327.test.rst
@@ -1,0 +1,4 @@
+Replaced bare ``assert np.isclose(...)`` and ``assert np.allclose(...)``
+calls across 41 test files with ``np.testing.assert_allclose(...)``, which
+produces informative failure messages showing actual vs. desired values and
+the tolerance used instead of a bare `AssertionError`.
--- a/tests/analysis/swept_langmuir/test_floating_potential.py
+++ b/tests/analysis/swept_langmuir/test_floating_potential.py
@@ -228,7 +228,7 @@ class TestFindFloatingPotential:
                 if np.isnan(val):
                     assert np.isnan(rtn_val)
                 else:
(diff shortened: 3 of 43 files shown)


namurphy (approved):
Thank you for submitting this!  I found this to be a good use case for AI since it's a useful refactoring that will help clarify future test failures, and it was also straightforward to review.  Happy to merge this now!

PR #3328, opened 2026-07-10
Raise a clear error when the CPR source or detector is at the grid origin

### LINKED ISSUE #2077: Bug in charged particle radiography when source or detector position is at origin
### Bug description

The code calculates the normalize vector from the origin to the source and to the detector 

```python
# Calculate normal vectors (facing towards the grid origin) for both
    # the source and detector planes
    self.src_n = -self.source / np.linalg.norm(self.source)
    self.det_n = -self.detector / np.linalg.norm(self.detector)
```

But if either is at the origin, this returns inf!

### Expected outcome

Either source or detector should be allowed to  be located at the origin

### Minimal complete verifiable example

_No response_

### Package versions

_No response_

### Additional context

_No response_

Mohit-Ak:
## What was wrong

In the charged-particle-radiography `Tracker`, the source and detector planes are
defined by normal vectors that point from each plane towards the grid origin:

```python
self.src_n = -self.source / np.linalg.norm(self.source)
self.det_n = -self.detector / np.linalg.norm(self.detector)
```

If the source or the detector is placed *at* the grid origin, the corresponding
norm is zero, so the normal comes out as `[nan nan nan]`. For a source at the
origin the magnification `1 + norm(detector)/norm(source)` also becomes `inf`.
The `Tracker` was constructed silently in this broken state and only failed
later, in a much more confusing place, which is exactly the "confusing errors"
the issue asks us to avoid.

Reproduction before the fix (both cases build without complaint):

```
CASE A (source at origin):   src_n = [nan nan nan], mag = inf
CASE B (detector at origin): det_n = [nan nan nan]
```

## What the fix does

Adds an explicit check right after the source/detector positions are coerced to
Cartesian, before the normals are computed, that raises a clear `ValueError`
when either point is at the grid origin. This matches the existing
input-validation style already used in `__init__` (e.g. the "vector between the
source and the detector does not intersect the grid" error).

Fully *supporting* a source or detector at the origin would be a design change
rather than a bug fix — the point-projection magnification is genuinely
undefined there — so this PR keeps the scope to turning silent NaN/inf geometry
into an actionable error message, as the maintainer suggested in the issue
thread.

## How it was tested

Added a focused regression test, `test_source_or_detector_at_origin`, that
asserts both the source-at-origin and detector-at-origin cases raise
`ValueError`.

- With the fix removed, the new test fails at the `src_n = ...` line with
  `RuntimeWarning: invalid value encountered in divide` (the original bug).
- With the fix in place, the new test passes.
- The full `test_synthetic_radiography.py` module passes: **34 passed**.
- `ruff check` (v0.15.20, the pinned pre-commit version) reports
  "All checks passed!" and `ruff format --check` is clean on both changed files.

Fixes #2077


pheuer:
@Mohit-Ak  Thanks for identifying this issue and even more for producing a patch! The PR looks good to me, but one question before we merge. 

Is there an issue if the source and detector are put on the same side of the origin? The ValueError text suggests there is, in which case maybe that should also be verified in the input?

Mohit-Ak:
Great question — yes, there is, and I've added a check for it.

I dug into the geometry to be sure. The point-projection magnification is `mag = 1 + |detector| / |source|`, which is only physically meaningful when the grid origin (the target) sits *between* the source and detector, i.e. they're on opposite sides. I confirmed empirically against `main` that same-side placement isn't currently caught:

- `source = (0, 10, 0) mm`, `detector = (0, 200, 0) mm` (both +y) builds a `Tracker` with `mag = 21.0` — a plausible-looking but meaningless number, since the particles can't project through the origin onto the detector in that configuration.
- The existing `vector_intersects` guard does **not** catch this: it returns `True` for the `(0,10,0) → (0,200,0)` segment even though that segment lives entirely at y ≥ 10 mm and never touches a grid spanning ±1 mm (it appears to test the infinite line, not the bounded segment).

The cleanest discriminator is the sign of `np.dot(source, detector)`: it's negative for every valid opposite-side geometry (including the spherical/cylindrical ones in the existing tests) and `>= 0` exactly when they're on the same side or perpendicular through the origin. So I added, right after the origin checks:

```python
if np.dot(self.source, self.detector) >= 0:
    raise ValueError(
        "The source and detector must be placed on opposite sides of "
        f"the grid origin (source = {source}, detector = {detector}). "
        "The origin must lie between them so that the projection "
        "geometry and magnification are well defined."
    )
```

Verified locally (plasmapy 2026.2.0): a new `test_source_and_detector_on_same_side` covers both-+y, both-−y, and the perpendicular (`dot == 0`) case; it goes RED on the current code (Tracker builds, no error) and GREEN with the check. The full non-slow CPR module plus the coordinate-system/`test_init` Tracker-building tests all still pass (no valid geometry is newly rejected), and `ruff check`/`ruff format` are clean. Changelog fragment updated to mention the same-side case.

Let me know if you'd prefer a stricter tolerance (e.g. rejecting near-perpendicular within some angle rather than only `dot >= 0`) — happy to adjust.

pheuer (approved):
I think these are good additions - thank you for putting this together!

PR #3329, opened 2026-07-10
Add fusion cross sections

### LINKED ISSUE #110: Add functionality to find fusion reaction rates, cross sections, and energy generation rates

seanclark-cell:
Adding two functions that give the nuclear cross section and reactivity for some (so far nine) common fusion reactions relevant to fusion energy and astrophysics. All reactions are evaluated through a single Bosch-Hale (1992) Padé parametrization: the four main D-T/D-D/D-³He reactions use the coefficients published by Bosch and Hale, while the remaining five use coefficients obtained by fitting the same Padé form to ENDF/B-evaluated data, so every reaction runs through the identical code path. The goal is to eventually support a few educational notebooks on these topics.

Closes #110


src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,330 @@
...
+    parametrization.
+    """
+    rxn = _rxty_coeff(reaction)
+    T_keV = ion_temp.to(u.keV).value
+    Theta = _rxty_polynomial(T_keV, rxn)
+    xi = _find_xi(Theta, rxn)
+    sv = _find_reactivity(T_keV, rxn, Theta, xi)
+    return sv * (u.cm**3 / u.s)
+
+amu = 1.66053906660e-27           # atomic mass unit [kg]
```
pheuer: For these quantities, consider using astropy.constants

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,450 @@
...
+import numpy as np
+from scipy.interpolate import CubicSpline
+
+from plasmapy.utils import data
+
+"""
+Opening Bosch and Hale Tables IV and Json Files
+"""
+
+DATA_DIR = data.__file__
```
JaydenR2305: @pheuer this is my janky way of doing this :P 
I know you had mentioned something about `importlib`?

pheuer: Yes, it looks like there is a new way of doing this now using this `files` object: 
https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files

Something like this (AI example)

```python
from importlib.resources import files, as_file

# 1. Point to the resource inside your package
resource_ptr = files("my_package.data").joinpath("config.json")

# 2. Safely resolve it to a physical pathlib.Path object
with as_file(resource_ptr) as physical_path:
    print(f"Physical path: {physical_path}")
    
    # You can now pass this path to 3rd-party libraries or open it
    with open(physical_path, "r", encoding="utf-8") as f:
        data = f.read()
```

JaydenR2305: im going to hate on the clanker and say that
```py
with resource_ptr.open("r") as data:
```
is superior


unless of course there is some obscure reason it decided to with...as twice

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,450 @@
...
+    table_v = json.load(f)
+
+with Path.open(DATA_DIR / "bosch_hale_table_vii.json") as f:
+    rxty_coeffs = json.load(f)["reactions"]
+
+with Path.open(DATA_DIR / "bosch_hale_table_viii.json") as f:
+    table_viii = json.load(f)
+
+
+def cross_section(energy: u.Quantity, reaction: str, source: str) -> float:
```
JaydenR2305: BTW you can add further annotations to the `u.Quantity` object. This also applies generally to any `Iterable`: lists, NumPy arrays, etc. Mainly anything that could be considered, in some abstract sense, to be a collection of "things"

```suggestion
def cross_section(energy: u.Quantity[u.MeV], reaction: str, source: str) -> float:
```

If you're familiar with basically any other programming language-- chances are it's a language that is "strongly typed." Meaning, if I annotated some function 

```py
def my_function(my_argument: list[str]) -> str:
     return my_argument[0]
```
and called it with the correct type, but consisting of the wrong entries:
```py

my_list_of_floats = [1, 67, 42]
my_function(my_list_of_floats)

>>> 1
```

In a strongly-typed programming language, this would throw an exception. This is not the case in Python. The point of type annotations in Python is to make life easier for the end-user, and to make it easier to write consistent documentation.

In the case of `Quantity` objects however, it is worth it to get a bit neurotic to make sure our target audience: students and scientists who may not have a professional background in programming are passing units with the correct type hints-- something that again, is NOT a default behavior in Python. That is, if someone were to pass a length (`Quantity[u.m]`) to a function that calls for a time, neither a warning nor error will be emitted. This is where the [`@validate_quantities`](https://docs.plasmapy.org/en/stable/api/plasmapy.utils.decorators.validators.validate_quantities.html) decorator is an absolute necessity. As a general rule of thumb, any function that isn't explicitly marked as a [lite function](https://docs.plasmapy.org/en/stable/glossary.html#term-lite-functions) it is considered best practice to annotate the function with `@validate_quantities`.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,453 @@
...
+
+with Path.open(DATA_DIR / "bosch_hale_table_vii.json") as f:
+    rxty_coeffs = json.load(f)["reactions"]
+
+with Path.open(DATA_DIR / "bosch_hale_table_viii.json") as f:
+    table_viii = json.load(f)
+
+
+@validate_quantities
+def cross_section(energy: u.Quantity[u.keV], reaction: str, source: str) -> float:
```
namurphy:
```suggestion
def fusion_cross_section(energy: u.Quantity[u.keV], reaction: str, source: str) -> u.Quantity[u.m**2]:
```
🏷️ Since there are a bunch of different types of cross sections (like the cross sections for Coulomb collisions), it would be helpful to have a more explicit name to reduce the chance of confusion.  

📏 Since PlasmaPy has had a practice of using SI units most of the time, could we have the function return a `Quantity` that consistently has units of m$^2$?

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,453 @@
...
+        - ``"T(t,2n)A"`` — :math:`T + T → 2n + α`
+        - ``"11B(p,a)2A"`` — :math:`^{11}B + p → 3α`
+
+    source : {``"BH"``, ``"ENDF"``}
+        The data source used to evaluate the cross section:
+
+        - ``"BH"`` — the Padé parametrization of :cite:t:`bosch:1992`,
+          valid for the four reactions in their Table IV over the energy
+          range given for each reaction (at most :math:`0` - :math:`4900`
+          keV).
```
namurphy: I'm wondering if it would be helpful to users to list the valid energy range for each reaction in this docstring, since I imagine most users wouldn't look up the table in the paper. 🤔

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,453 @@
...
+    The Bosch-Hale parametrization writes the cross section in terms of
+    the astrophysical :math:`S`\ -function and the Gamow penetrability
+    factor as
+
+    .. math::
+
+        σ(E) = \frac{S(E)}{E \exp\left(B_G / \sqrt{E}\right)},
+
+    where :math:`B_G` is the Gamow constant and :math:`S(E)` is a Padé
+    approximant of the form
```
namurphy:
```suggestion
    where :math:`B_G` is the :wikipedia:`Gamow factor` and :math:`S(E)`
    is a Padé approximant of the form
```
One of the documentation extension lets us use this handy `:wikipedia:` role so that we can link directly to article like the one on the [Gamow factor](https://en.wikipedia.org/wiki/Gamow_factor). 🔗

It'd be worth double checking this in the Read the Docs preview in case I'm mis-remembering the syntax. 😅

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+def _rxty_coeff(r):
+    return rxty_coeffs[r]
+
+
+"""
+Cross Section Function and its Helper Functions
+"""
+
```
namurphy, line 611:
```suggestion

# Cross Section Function and its Helper Functions

```

changelog/3329.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added a fusion module in the src/formulary that calculates
```
pheuer: Look at some other changelogs to see how you can format these as links to the relevant modules. Expand this a bit more to say Padre polynomial fits (e.g. Bosch and Hale)

docs/bibliography.bib
```diff
@@ -103,6 +103,17 @@ @inproceedings{boris:1970
   editor = {J. P. Boris and R. A. Shanny},
 }
 
+@article{bosch:1992,
```
pheuer, line 106: Good - can you add a citation for ENDF? We should mention what we fit. The files I downloaded are from all different versions of ENDF so maybe just cite their website? TBD.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+from plasmapy.utils.decorators import validate_quantities
+
+"""
+Opening Bosch and Hale Tables IV and Json Files
+"""
+
+DATA_DIR = files("plasmapy.utils.data")
+
+with as_file(DATA_DIR) as physical_path:
```
pheuer: This code will execute whenever you load this module. Can you include these calls in the functions so they only load when needed? 

If you want them to only load once, make a function that returns each table and use the @functools.cache decorator so it's saved in memory once loaded once.

pheuer: Also make these variables/functions if you do that private

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+        - ``"3He(d,p)A"`` — :math:`^3He + D → p + α`
+        - ``"D(d,p)T"`` — :math:`D + D → p + T`
+        - ``"D(d,n)3He"`` — :math:`D + D → n + ^3He`
+        - ``"3He(3He,2p)A"`` — :math:`^3He + ^3He → 2p + α`
+        - ``"3He(t,n+p)A"`` — :math:`^3He + T → n + p + α`
+        - ``"3He(t,d)A"`` — :math:`^3He + T → D + α`
+        - ``"T(t,2n)A"`` — :math:`T + T → 2n + α`
+        - ``"11B(p,a)2A"`` — :math:`^{11}B + p → 3α`
+
+        The first four use the coefficients published in Table IV of
```
pheuer: Explain this is because of an error in the paper

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+
+    gives a problem that is linear in the :math:`A_i` and :math:`B_j`,
+    whose least-squares solution provides a good starting point near the
+    true minimum. Finally, the numerator and denominator coefficients
+    differ by many orders of magnitude, so they are scaled before fitting
+    to keep the Jacobian columns comparable and avoid an
+    ill-conditioned solve. The resulting :math:`A_i` and :math:`B_j` are
+    stored in the same format as the published coefficients and evaluated
+    through exactly the same code path, so ``cross_section`` behaves
+    identically for fitted and published reactions.
```
pheuer, line 318: I need to look at this when rendered to read fully but this is a very nice detailed docstring!

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+
+    Examples
+    --------
```
pheuer, line 321: Can you verify this works if you give it a numpy array of energies? That's likely a common application.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    path, so ``reactivity`` behaves identically for fitted and published
+    reactions.
+
+    Examples
+    --------
+    >>> import astropy.units as u
+    >>> reactivity(10 * u.keV, "D(t,n)A")  # doctest: +SKIP
+    <Quantity 1.13616547e-16 cm3 / s>
+    """
+    if reaction not in rxty_coeffs:
```
pheuer: Since we took a lot of inspiration from that one online notebook, lets include a comment in the code with that link as acknowledgement

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def _in_BH_rxn_ion_temp_range(T, reaction):
+    r"""Return whether ``T`` lies within the reactivity validity range."""
+    rxn = _rxty_coeff(reaction)
+    in_range = (rxn["T_min_keV"] * u.keV <= T) & (rxn["T_max_keV"] * u.keV >= T)
+    return bool(np.all(in_range))
+
+
+def _xs_coeff(r):
```
pheuer: Are these used many places? I think this is a little over broken out and this could be just inline in the xs function, same for reactivity

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    Evaluate the Bosch-Hale Padé approximant for the S-function.
+    """
+    S_vals = rxn["A1"] + E * (
+        rxn["A2"] + E * (rxn["A3"] + E * (rxn["A4"] + E * rxn["A5"]))
+    )
+    S_vals /= 1 + E * (rxn["B1"] + E * (rxn["B2"] + E * (rxn["B3"] + E * rxn["B4"])))
+    return S_vals
+
+
+def _parametrization_formula(S_Vals, energy, rxn):  # B&H Eq (8)
```
pheuer: Also not sure if this should be broken out. The Padre polynomial makes more sense to me

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    r"""
+    Combine the S-function with the Gamow penetrability to get a cross-section.
+    """
+    return S_Vals / (energy * np.exp(rxn["B_G"] / np.sqrt(energy)))
+
+
+@u.quantity_input
+def _BH_cross_section(energy: u.Quantity, reaction: str) -> float:
```
pheuer: I think this could be in the xs function main body

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    Evaluate the :math:`\theta(T)` Padé approximant for the Bosch-Hale
+    reactivity.
+    """
+    theta = T * (r["C2"] + T * (r["C4"] + T * r["C6"]))
+    theta /= 1 + T * (r["C3"] + T * (r["C5"] + T * r["C7"]))
+    theta = 1 - theta
+    return T / theta
+
+
+def _find_xi(Theta, r):
```
pheuer: Just do in the main body

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    r"""
+    Compute the :math:`\xi` factor for the Bosch-Hale reactivity formula.
+    """
+    xi = (r["B_G"]) ** 2
+    xi /= 4 * Theta
+    return (xi) ** (1 / 3)
+
+
+def _find_reactivity(T, r, theta, xi):
```
pheuer: Do in the main fcn

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+    r"""
+    Assemble the Bosch-Hale reactivity from its precomputed pieces.
+    """
+    rcty = np.sqrt(xi / (r["m_r_c2"] * T**3))
+    rcty *= r["C1"] * theta * (np.exp(-3 * xi))
+    return rcty
+
+
+@u.quantity_input
+def _BH_reactivity(ion_temp: u.Quantity, reaction: str) -> float:
```
pheuer: I think this should go in the main function. 

I think you can fit basically all of these in the two main functions + the Padre and rtx fit functions, and that will be easier to read.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,408 @@
...
+
+    with Path.open(physical_path / "bosch_hale_table_vii.json") as f:
+        rxty_coeffs = json.load(f)["reactions"]
+
+    with Path.open(physical_path / "bosch_hale_table_viii.json") as f:
+        table_viii = json.load(f)
+
+
+@validate_quantities
+def cross_section(energy: u.Quantity[u.keV], reaction: str) -> u.Quantity:
```
pheuer: Note this is in CM coordinates. 

BUT, let's throw a wrench in this a little bit. I would like this function to take a 'reference frame' keyword either 'CM' or 'lab' and return the appropriate XS and reactivity! That will require getting the mass in amu of each of the constitute particles from another data file. But it will be very useful!

src/plasmapy/utils/data/xs_pade_polynomial_coefficients.json
```diff
@@ -0,0 +1,146 @@
+{
```
pheuer, line 1: These look good, but not everything here is actually in Bosch and Hale right? Filename is misleading if we add more fits. Let's just have a table for XS and a table for reactivities? You could add a key for each reaction saying where it's from, B&H or ENDF fit or wherever else?

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,362 @@
...
+)
+
+CM3_PER_S = u.cm**3 / u.s
+
+#: The published tables are quoted to four significant figures.
+TABLE_RTOL = 1e-3
+
+
+def _load_table(name):
+    with Path.open(fusion.DATA_DIR / name) as f:
```
pheuer: Use importlib separately here, so you don't depend on this. 

Also DATA_DIR is a public variable, we don't want that to appear in the docs, should only be the two main functions that are public

pheuer (changes requested):
This is very good, especially the detailed tests and docstrings! I'm not convinced the functions need to be so split up, I think it would be easier to read with them condensed, and the split isn't adding much efficiency. Also check to make sure all the variables in the module space are private except the functions!

pheuer:
Also, note your branch is now out of date with PlasmaPy main. To solve that, merge PlasmaPy main INTO your branch, then push the result back up to your branch on the remote.

src/plasmapy/formulary/fusion.py
namurphy: Say, @pheuer, could you travel back to 2017 and give me this advice back when I started working on `plasmapy.particles`? 😅

📝 Relatedly, Python 3.15 is going to have [explicit lazy imports](https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-lazy-imports), like `lazy import antigravity` (or alternatively by defining `__lazy_modules__`).

namurphy (commented):
Thank you for working on this!

src/plasmapy/formulary/fusion.py
seanclark-cell: For this change, I was wondering if defaulting to center_of_mass=True, and if you set it to False it switches to the lab reference frame works instead?

Additionally, in the lab-frame conversion the order of the reactants matters in which particle is the projectile versus the target correct? Do you have any suggestions on the cleanest way to handle that? Or are the reaction keys already written in a standard order that tells us which reactant is which (X(a,b)Y where a is the projectile and X is the target)? If so, should we include way to switch the order of the reactants?

seanclark-cell:
@namurphy @pheuer @JaydenR2305 I was wondering where the best spot to keep the JSON coefficient data files would be? Right now they are in plasmapy/utils/data folder but is there a better place for them?

src/plasmapy/formulary/fusion.py
seanclark-cell: check implementation if it is done the way you were thinking.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+def _load_reactions(name):
+    with as_file(_DATA_DIR) as physical_path, Path.open(physical_path / name) as f:
+        return json.load(f)["reactions"]
+
+
+@validate_quantities
+def fusion_cross_section(
+    energy: u.Quantity[u.keV],
+    reaction: str,
+    out_of_range: str = "raise",
```
pheuer: I think the 'nan' behavior should be the default. I'm not sure if this requires a keyword, it's pretty easy for people to check whether there are any NaN results in the output if they want to avoid that?

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+        - ``"nan"`` — evaluate the in-range energies as usual and return
+          `~numpy.nan` for the out-of-range ones instead of raising. If
+          *every* input energy is out of range the result is entirely
+          NaN and a warning is issued.
+
+    reference_frame : `str`, optional
+        The frame in which ``energy`` is supplied, using the ``X(a,b)Y`` naming
+        of ``reaction`` (target ``X``, projectile ``a``). Must be one of:
+
+        - ``"CM"`` (default) — ``energy`` is the center-of-mass energy and is
```
pheuer, line 148: Maybe note here that target/projectile order doesn't matter for CM frame but does in the lab frame?

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+    to keep the Jacobian columns comparable and avoid an
+    ill-conditioned solve. The resulting :math:`A_i` and :math:`B_j` are
+    stored in the same format as the published coefficients and evaluated
+    through exactly the same code path, so ``fusion_cross_section`` behaves
+    identically for fitted and published reactions.
+
+    Examples
+    --------
+    >>> import astropy.units as u
+    >>> fusion_cross_section(100 * u.keV, "D(t,n)A")  # doctest: +SKIP
```
pheuer: Why skip doctest?

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+    if reference_frame == "lab":
+        nuclide = {
+            "p": "p+",
+            "d": "deuteron",
+            "D": "deuteron",
+            "t": "triton",
+            "T": "triton",
+            "3He": "He-3 2+",
+            "11B": "B-11 5+",
+        }
```
pheuer: Good use of Particles class! This could break if a new reaction is added without updating this dict. Let's get the target and projectile strings, check that they are in this dict and raise an exception if not, before creating the Particle objects..

You could cast the string to lowercase before using it as a dict key to catch D and d at the same time for deuteron, rather than repeating it in the table.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+        target = Particle(nuclide[reaction.split("(", maxsplit=1)[0]])
+        projectile = Particle(nuclide[reaction.split("(")[1].split(",", maxsplit=1)[0]])
+        # Convert lab-frame projectile energy to the CM energy via the mass ratio.
+        E_keV = E_keV * (target.mass / (target.mass + projectile.mass)).value
+
+    scalar = E_keV.ndim == 0
```
pheuer, line 356: Add line comment to explain what this is doing

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,548 @@
...
+            f"all input temperatures are outside the {reaction!r} range "
+            f"({rxn['T_min_keV']} to {rxn['T_max_keV']} keV); returning all NaN",
+            stacklevel=2,
+        )
+
+    result = sv[0] if scalar else sv
+    return result * (u.cm**3 / u.s)
+
+
+def _xs_pade_polynomial(rxn, E):
```
pheuer, line 614: Would be great if these two functions could point to the equation # in the Bosch and Hale paper that you are implementing here

src/plasmapy/utils/data/rxty_pade_polynomial_coefficients.json
```diff
@@ -0,0 +1,149 @@
...
+      "C1": 1.82742007e-19,
+      "C2": 2.37404956,
+      "C3": 2.34404538,
+      "C4": 0.0146796828,
+      "C5": 0.0150195667,
+      "C6": 0.0124176631,
+      "C7": 0.01241633,
+      "T_min_keV": 8.27,
+      "T_max_keV": 100,
+      "max_deviation_percent": 4.97,
```
pheuer, line 84: Define max_deviation_percent somewhere in this file. And what is the pole warning? Remember there needs to be enough information for a new user to come along and add new reactions to this file.

pheuer (commented):
Looking excellent! Here are a few more comments, let's discuss a bit on Friday but otherwise I think this is nearly ready to merge!

src/plasmapy/utils/data/rxty_pade_polynomial_coefficients.json
seanclark-cell: I defined max deviation percent in a Quantities key change, I don't know if this is best option but it seems to fit. 

Additionally, the pole_warning was something I had forgotten was left in there when I was testing to get coefficients. This created a much larger problem because I forgotten that at certain values there were these massive deviations/spikes that were being ignored but I just went back and gave curve_fit better initial guesses to get coefficients that avoided this. Thank you for the catch!

src/plasmapy/formulary/fusion.py
JaydenR2305: I find it strange that B&H doesn't go into detail about the fitting process. The methodology seems really important, since the way the fit is performed can introduce non-physical singularities and a strong dependence on initial parameter guesses (as Sean discovered earlier)

JaydenR2305:
nice work on 100% patch coverage!

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,636 @@
...
+    The cross-section and reactivity coefficients live in separate data
+    files, so this listing is not guaranteed to match
+    `available_reactivity_reactions`.
+
+    Examples
+    --------
+    >>> "D(t,n)A" in available_cross_section_reactions()
+    True
+    """
+    return list(_load_reactions("xs_pade_polynomial_coefficients.json"))
```
pheuer: I'm confused - doesn't this include the entire data structure with the coefficients? Since this function is also called in the clasess and there it returns the coeffs? Don't you need to extract just the keys from the output of `_load_reactions` to just get the reaction names?

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,636 @@
...
+    rxty_coeff = _load_reactions("rxty_pade_polynomial_coefficients.json")
+    if reaction not in rxty_coeff:
+        raise ValueError(
+            f"{reaction!r} is not one of the available reactions: "
+            f"{', '.join(rxty_coeff)}"
+        )
+    rxn = rxty_coeff[reaction]
+
+    T_keV = np.asarray(ion_temp.to(u.keV).value, dtype=float)
+    # True if a single temperature was passed, used to return scalor instead of 1 element array.
```
pheuer: scalor -> scalar

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,636 @@
...
+        )
+
+    result = sv[0] if scalar else sv
+    return result * (u.cm**3 / u.s)
+
+
+# Bosch and Hale equation (9)
+def _xs_pade_polynomial(rxn, E):
+    r"""
+    Evaluate the Bosch-Hale Padé approximant for the S-function.
```
pheuer: approximation?

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,643 @@
...
+
+_TABLE_V_PARAMS = _table_params(_TABLE_V, u.keV, _XS_COEFF)
+_TABLE_VIII_PARAMS = _table_params(_TABLE_VIII, u.keV, _RXTY_COEFF)
+
+
+class TestBoschHaleCrossSection:
+    """Padé parametrization of sigma(E), Bosch & Hale Eq. (8) and Table IV."""
+
+    @pytest.mark.parametrize(("energy", "reaction", "expected_mbarn"), _TABLE_V_PARAMS)
+    def test_reproduces_table_v(self, energy, reaction, expected_mbarn):
```
pheuer, line 83: This test and the following one don't actually call the user function but your internal test function, so they only test the coefficient files not your function. Why not use the actual fusion_cross_section function here to make the comparison? That's what we actually want to test.

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,643 @@
...
+
+class TestBoschHaleReactivity:
+    """Closed-form Maxwellian <sv>(T), Bosch & Hale Eqs. (12)-(14) corrected table VIII values."""
+
+    @pytest.mark.parametrize(
+        ("ion_temp", "reaction", "expected_cm3_per_s"), _TABLE_VIII_PARAMS
+    )
+    def test_reproduces_table_viii(self, ion_temp, reaction, expected_cm3_per_s):
+        """Every verified cell of Table VIII is reproduced to its precision."""
+        sv = _raw_reactivity(ion_temp, reaction)
```
pheuer: Same comment here, these tests should use the actual fusion_reactivity function so we're benchmarking the user function. Otherwise, that function could be broken while this one is fine and we woudlnt' know

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,643 @@
...
+    @pytest.mark.parametrize("reaction", _RXTY_REACTIONS)
+    def test_every_listed_rxty_reaction_evaluates(self, reaction):
+        """Anything advertised must actually work in public function."""
+        assert reaction in fusion.available_reactivity_reactions()
+        r = _RXTY_COEFF[reaction]
+        T = 0.5 * (r["T_min_keV"] + r["T_max_keV"]) * u.keV
+        assert fusion.fusion_reactivity(T, reaction) > 0
+
+
+class TestCrossSectionDispatch:
```
pheuer: Do we need these dispatch tests anymore, since you incorporated everything in the main functions?

pheuer (changes requested):
Very close - I have a few questions, mostly about the tests which I dug into more on this review.

src/plasmapy/formulary/fusion.py
```diff
@@ -0,0 +1,636 @@
...
+) -> u.Quantity[u.m**2]:
+    r"""
+    Calculate the fusion cross section :math:`σ(E)` for a two reactant fusion
+    reaction.
+
+    The cross section is evaluated from the Bosch-Hale Padé
+    parametrization :cite:p:`bosch:1992`. For the reactions tabulated by
+    Bosch and Hale the published coefficients are used; for the remaining
+    reactions the same functional form is fit (Bosch and Hale eq 8 and 9)
+    to ENDF/B evaluated cross sections (see Notes).
```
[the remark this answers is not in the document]
pheuer: Ignore this - copilot is confused about how the coeffs are generated vs. how they are used in the calculation

pheuer:
One good comment from copilot to check:

The documented valid energy range for "3He(3He,2p)A" does not match the coefficient table: xs_pade_polynomial_coefficients.json lists E_min_keV as 5 (not 1.0). Please update the docstring table to match the enforced range.

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+_XS_COEFF = fusion._load_reactions("xs_pade_polynomial_coefficients.json")
+_RXTY_COEFF = fusion._load_reactions("rxty_pade_polynomial_coefficients.json")
+
+#: All reactions with tabulated ENDF cross-sections.
+_XS_REACTIONS = list(_XS_COEFF)
+_RXTY_REACTIONS = list(_RXTY_COEFF)
+_BH_RXTY_REACTIONS = [r for r in _BH_REACTIONS if r in _RXTY_COEFF]
+
+
+def _raw_cross_section(energy, reaction):
```
pheuer: We should just delete these two raw functions and replace any calls to them in the tests to use the actual functions

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        assert_quantity_allclose(
+            sigma,
+            expected_mbarn * u.mbarn,
+            rtol=_TABLE_RTOL,
+            err_msg=f"Table V mismatch for {reaction} at {energy}",
+        )
+
+    @pytest.mark.parametrize("reaction", _XS_REACTIONS)
+    def test_returns_millibarns(self, reaction):
+        sigma = _raw_cross_section(300 * u.keV, reaction)
```
pheuer: replace with actual function

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        assert fusion.fusion_reactivity(ion_temp, reaction).shape == ion_temp.shape
+
+    @pytest.mark.parametrize("reaction", _BH_RXTY_REACTIONS)
+    def test_monotonic_below_the_peak(self, reaction):
+        """
+        <σv> rises monotonically with T until the sigma(E) resonance is passed.
+        D(t,n)α turns over near 64 keV, so the check stops at 50 keV.
+        """
+        ion_temp = np.logspace(np.log10(0.5), np.log10(50), 200) * u.keV
+        sv = _raw_reactivity(ion_temp, reaction).value
```
pheuer: replace with actual function

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        T = 0.5 * (r["T_min_keV"] + r["T_max_keV"]) * u.keV
+        assert fusion.fusion_reactivity(T, reaction) > 0
+
+
+class TestCrossSectionInputValidation:
+    """Public-API contract for ``fusion_cross_section``: unknown reactions,
+    unit coercion, wrong units, and the SI (m^2) output unit."""
+
+    @pytest.mark.parametrize("reaction", _XS_REACTIONS)
+    def test_matches_backend(self, reaction):
```
pheuer: I don't think this is needed since there is no backend now

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        with pytest.raises(u.UnitsError):
+            fusion.fusion_cross_section(100 * u.s, "D(t,n)A")
+
+
+class TestReactivityInputValidation:
+    """Public-API contract for ``fusion_reactivity``: unknown reactions,
+    unit coercion, wrong units, and the SI (m^3/s) output unit."""
+
+    @pytest.mark.parametrize("reaction", _RXTY_REACTIONS)
+    def test_bh_source_matches_backend(self, reaction):
```
pheuer: Same here - not needed with no backend

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        assert v[1] > 0
+
+    @pytest.mark.parametrize("reaction", _RXTY_REACTIONS)
+    def test_in_range_matches_raw_backend(self, reaction):
+        """Masking must not perturb the in-range values."""
+        c = _RXTY_COEFF[reaction]
+        T = np.linspace(c["T_min_keV"], c["T_max_keV"], 5) * u.keV
+        masked = fusion.fusion_reactivity(T, reaction)
+        assert np.all(np.isfinite(masked.value))
+        assert_quantity_allclose(masked, _raw_reactivity(T, reaction), rtol=1e-12)
```
pheuer: replace raw with actual function

tests/formulary/test_fusion.py
```diff
@@ -0,0 +1,632 @@
...
+        assert sv.unit == u.m**3 / u.s
+
+    @pytest.mark.parametrize("reaction", _RXTY_REACTIONS)
+    def test_scalar_in_range_matches_raw_backend(self, reaction):
+        """A scalar in-range value stays scalar and matches the bare formula."""
+        c = _RXTY_COEFF[reaction]
+        mid = 0.5 * (c["T_min_keV"] + c["T_max_keV"]) * u.keV
+        masked = fusion.fusion_reactivity(mid, reaction)
+        assert masked.isscalar
+        assert_quantity_allclose(masked, _raw_reactivity(mid, reaction), rtol=1e-12)
```
pheuer: Replace _raw with actual fcn

pheuer (changes requested):
Found a few more cases where the tests are using the _raw functions where they should use the actual user-facing functions. Let's delete those _raw functions and make sure all of the tests call the actual functions.

seanclark-cell:
> Found a few more cases where the tests are using the _raw functions where they should use the actual user-facing functions. Let's delete those _raw functions and make sure all of the tests call the actual functions.

With that last commit I should have gotten rid of all instances of the raw functionality ("backend"). I was originally trying to have it just for a comparison to test the outputs made sense but it makes more sense to just rely on the Bosch and Hale table recreation for that.

pheuer (approved):
Looks good to me, thanks @seanclark-cell this is a great addition to the package and will be really useful!

PR #3330, opened 2026-07-12
Use sphinx_toolbox.rest_example to de-duplicate reST examples

### LINKED ISSUE #2270: Use `sphinx_toolbox.rest_example` to de-duplicate reST example code blocks with copied output
In our documentation guide, we currently have a few reStructuredText (reST) examples where we do a code-block for the source code, and then copied that source code so that it shows up as rendered docs immediately after.  The problem with this is that we may change the code-block but forget to change the copied code shown immediately after.

The [`sphinx_toolbox.rest_example`](https://sphinx-toolbox.readthedocs.io/en/latest/extensions/rest_example.html) extension to Sphinx would let us use the `.. rest-example::` directive and avoid having to copy & paste the reST examples.  

I don't think we'd need to do this anywhere except the documentation guide.  This is low priority, but would be good to do if we end up doing any revisions of the doc guide.

zaid646:
Closes #2270

Replaces 7 duplicated `code-block` + "this renders as" patterns with `rest-example` directive provided by `sphinx_toolbox`. This eliminates source-of-truth drift between RST source examples and their rendered output.

Changes:
- Adds `sphinx_toolbox.rest_example` to extensions in `docs/conf.py`
- Converts 7 instances in `docs/contributing/doc_guide.rst`

Net: -71 lines, same rendered output, one source of truth.


zaid646:
@namurphy this one is ready to go. Replaced 7 duplicated `code-block` + "renders as" patterns in the doc guide with `rest-example` directive from `sphinx-toolbox` (already a dependency). Eliminates source-of-truth drift. All checks passing.

CITATION.cff
```diff
@@ -781,6 +781,8 @@ authors:
   family-names: Arvind Khakharia
   alias: Mohit-Ak
 
+- given-names: Zaid
+  alias: zaid646
 - given-names: Mike
```
namurphy, line 787:
```suggestion

- given-names: Mike
```

namurphy (approved):
Thank you for submitting this! 🙏🏻 It always make me happy when we can reduce unnecessary duplication.  We can get this merged right away. 

Cross-references: #2387 (where we dropped `sphinx_toolbox`) and #3203 (where we recently added back `sphinx_toolbox`).

PR #3055, opened 2025-07-24
Adding Bdot probe diagnostic and corresponding test

mirfan4906:
### Implementation Details

- Function `compute_bfield` is located in `src/plasmapy/diagnostics/magnetic_pickup_probe.py`.
- The function performs numerical integration of the voltage signal to compute the magnetic field.

### Tests

- Added a corresponding unit test in `tests/diagnostics/test_magnetic_pickup_probe.py`.
- The test inputs a sinusoidal voltage signal and verifies that the computed magnetic field waveform correctly matches the expected integral (within a small numerical tolerance).

PR #3336, opened 2026-07-14
Continuation of B-dot probe implementation from PR #3055 by Maimoona Irfan

JackBonner888:
## Description
Continues #3055 by @mirfan4906, whose work is imported unchanged as the first commit. 
This PR adds `plasmapy.diagnostics.magnetic_pickup_probe,` containing a `compute_bfield()` function that converts the voltage signal of a magnetic pickup (B-dot) probe into magnetic field strength by numerically integrating Faraday's law. I'm currently working with Prof. @dschaffner's group, where this originated, and hope to continue it the rest of the way. 

## Related Issues
Supersedes #3055 
See also #1652

### DIFF
--- a/src/plasmapy/analysis/magnetic_pickup_probe/__init__.py
+++ b/src/plasmapy/analysis/magnetic_pickup_probe/__init__.py
@@ -1,0 +1,10 @@
+"""
+Subpackage containing routines for analyzing magnetic pickup probe
+(Bdot probe) signals.
+"""
+
+__all__ = ["magnetic_field_calculation"]
+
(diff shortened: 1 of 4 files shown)


JackBonner888:
I created an analysis-module version of the B-dot calculation, whose `magnetic_field_calculation()` recovers B(t) from raw NumPy voltage/time arrays. Next step will be developing tests modeled on @mirfan4906's original attempt. As of now, the diagnostics-module is unchanged and left in this draft PR, although it will need to be reworked.

